PR #2072, opened 2020-01-30
Adds EnumerableMap.

MicahZoltu:
I'm not sure what the best way to recommend a library to OZ contracts is, so I figured I would start with a PR and see what happens.  This is code is _strongly_ based on the OZ EnumerableSet.  The biggest usability difference is that you have to instantiate manually and then call `.initialize` rather than doing `EnumerableMap.newMap()`.  This is because struct arrays aren't supported in the ways necessary to make `EnumerableMap.newMap()` work.

The `require` statements are basically invariants to protect against bugs (and provide an early warning during testing) but could be removed as they should never be hit (really they are assertions with useful error messages).

I also changed the array to be 1-indexed, which differs from `EnumerableSet` which has a 0-indexed array and then treats the index values as all off-by-one.  I personally found the code easier to understand with a placeholder value at the first entry position rather than having to shift the indexes by 1.  Since the placeholder is just a 0 value, I suspect there is no meaningful gas difference between the strategies and it is just a matter of style.

The biggest problem here, of course (same as `EnumerableSet`) is that Solidity doesn't contain generics, so the user will need to change the key and value types to match their code.  The nice thing here though is that the key can be any valid mapping key and the value can be anything including structs.  The user only needs to change the type in a few places.  I chose to use `address` and `uint8` because you can do a find/replace as neither type is used outside the key/value types and it is as reasonable of a choice as any.

Fixes #

PR #2151, opened 2020-03-26
API improvements for EnumerableSet

nventuro:
The old code did error-out on out of bounds accesses, but with an invalid opcode instead of a proper revert reason.


nventuro:
In the interest of speed I ended up also doing some internal renaming (values -> keys) on this PR, so that it can be used to build on top of when working on #2072. 

Sorry if you find this too messy. Reviewing commit by commit should be fine though.

nventuro:
This ended up being a mish-mash of unrelated improvements. After experimenting with `EnumerableMap`, I decided that `get(index)` looked weird and would be better suited as `get(key) -> value`, so I renamed it to `at(index)`.

contracts/utils/EnumerableSet.sol
```diff
@@ -20,58 +20,60 @@ pragma solidity ^0.6.0;
...
 
     struct AddressSet {
-        // Position of the value in the `values` array, plus 1 because index 0
-        // means a value is not in the set.
-        mapping (address => uint256) index;
-        address[] values;
+        address[] keys;
+        // Position of the key in the `keys` array, plus 1 because index 0
+        // means a key is not in the set.
+        mapping (address => uint256) indexes;
```
frangio: Why did you reorder the members of the struct? I'm weary of reordering because of the effect on upgradeable contracts.

contracts/utils/EnumerableSet.sol
```diff
@@ -20,58 +20,60 @@ pragma solidity ^0.6.0;
 library EnumerableSet {
 
     struct AddressSet {
-        // Position of the value in the `values` array, plus 1 because index 0
-        // means a value is not in the set.
-        mapping (address => uint256) index;
-        address[] values;
+        address[] keys;
```
frangio: I do not like this renaming to `keys`... They aren't keys in a set, they are values. I also don't understand what this has to do with maps.

contracts/mocks/EnumerableSetMock.sol
```diff
@@ -31,7 +31,7 @@ contract EnumerableSetMock{
         return set.length();
     }
 
-    function get(uint256 index) public view returns (address) {
```
frangio: I didn't understand the reasoning behind this renaming. Can you explain some more?

contracts/utils/EnumerableSet.sol
nventuro: I thought it made the code more readable :man_shrugging: And 3.0 will not be upgrade-compatible with 2.0 anyway, so I thought this was a good time to do it.

contracts/utils/EnumerableSet.sol
nventuro: That was also my original intuition, but I think the consistency with `EnumerableMap` (where key and value have very clear meanings) is worth it.

You can think of the set as a map where the keys have no associated value.

nventuro: It's actually not so much a matter of consistency but rather avoiding confusion between the two.

contracts/mocks/EnumerableSetMock.sol
nventuro: See [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2151#issuecomment-605077802).

contracts/utils/EnumerableSet.sol
frangio: I don't see any possible confusion. When would people be confused or make a mistake because of confusion?

nventuro: A set would have `contains(value)` while a map `contains(key)`, despite them both being pretty much the same thing under the hood.

Something similar happens with `remove(value)` for set and `remove(key)` for map.

nventuro: I've been convinced and renamed to `values`

contracts/utils/EnumerableSet.sol
frangio: Is there a list of storage-breaking changes? I still feel that we should avoid them if possible. But I'm unsure.

contracts/utils/EnumerableSet.sol
```diff
@@ -20,14 +20,15 @@ pragma solidity ^0.6.0;
 library EnumerableSet {
 
     struct AddressSet {
+        address[] values;
```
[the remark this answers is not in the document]
nventuro: We do! Thanks

contracts/utils/EnumerableSet.sol
nventuro: We don't, no.

Is there a reason to avoid them other than allowing for upgrades to happen? Given the extent of the API changes from 2.0 to 3.0 I'd discourage people from doing so.

frangio: Only upgrades.

Issue #1745, opened 2019-05-10
ERC721 "automatic" token URI


abcoathup:
**🧐 Motivation**
Current ERC721 minting with token URI requires manually specifying the URI per token.

To reduce gas cost, the OpenSea example uses concatenation of a base token URI and the tokenID. ([`TradeableERC721Token.sol`](https://github.com/ProjectOpenSea/opensea-creatures/blob/15a27aab9d7ffb9860faab27a8127f8eae9eb297/contracts/TradeableERC721Token.sol))

**📝 Details**

We want to provide an ERC721 extension `ERC721DefaultURI` with a base URI parameter, which overrides `ERC721Metadata`'s `tokenURI()` with the concatenation. To discuss the API please comment on the [forum thread about this](https://forum.zeppelin.solutions/t/add-support-for-erc721-token-uri-concatenation-of-base-token-uri-and-tokenid/566/2).

frangio:
Note that for concatenation of the base URI with the tokenId string we will use `abi.encodePacked`.

logeekal:
As discussed in #1163 , I can take this up.

pointtoken:
@logeekal Will your implementation make the baseTokenURI settable? I think this would be good, in the event that the URL needs to change. Is this code in a branch anywhere?

logeekal:
Hello @pointtoken, I will start implementation today so no code branch as of now.  And I am assuming that it should be settable so that everyone can set there own URI probably in constructor of their contract so that they do not have to do that with every mint operation.

What do you think @abcoathup, In Blockhorses I see that you have hard-coded it, so to be generic I also think it should be settable as well.

abcoathup:
Hi @logeekal 

I agree, I think that a private `_baseTokenURI` state variable should be set in the constructor via a parameter.

There should also be an internal setter function, so tokens inheriting from this contract can create a mechanism to change the _baseTokenURI as required.  

In [PeepethBadges.sol](https://github.com/peepeth/badges/blob/master/contracts/PeepethBadges.sol) we had the private state variable `_baseTokenURI` that was set in the constructor, (though not via a parameter).  We also had a mechanism for the owner to change the baseTokenURI, and I could see some projects wanting this ability to have some Role have the ability to change the baseTokenURI.

pointtoken:
[...]

PR #2160, opened 2020-03-30
Add EnumerableMap, refactor ERC721

nventuro:
Closes #2072.

This PR does a number of things (split across separate commits):
 * Refactor the internals of `EnumerableSet`, to make adding new sets easier
 * Add `EnumerableMap`, following the same approach for generic sets
 * Refactor ERC721 by using these internally

A couple things didn't go very well:
 * Due to issues with the type system, I had to drop `EnumerableSet.enumerate`. I plan on bringing this back once I get feedback from the Solidity team on how to best do it
 * Because of this, I also dropped `ERC721._tokensOfOwner()`

Finally, I added some small new tests to `ERC721`: during the migration I made a small mistake that was not caught by the suite, so I created a new test.


contracts/token/ERC721/ERC721.sol
```diff
@@ -180,16 +168,15 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
+        return _holderTokens[owner].at(index);
     }
 
     /**
      * @dev Gets the total amount of tokens stored by the contract.
      * @return uint256 representing the total amount of tokens
      */
     function totalSupply() public view override returns (uint256) {
-        return _allTokens.length;
+        return _tokenOwners.length();
```
nventuro: This makes me thing that `lenght` might not be the best name. This returns the number of keys, which in this case are tokenIds. Though perhaps what is wrong is the name of the `_tokenOwners` state variable.

nventuro:
Note also that this refactor could also include the approval-related mappings if we extended EnumerableMap/Set to generic structs as opposed to limiting it to 32-byte data chunks.

contracts/mocks/EnumerableMapMock.sol
```diff
@@ -0,0 +1,38 @@
+pragma solidity ^0.6.0;
+
+import "../utils/EnumerableMap.sol";
+
+contract EnumerableMapMock {
+    using EnumerableMap for EnumerableMap.UintToAddressMap;
+
+    event TransactionResult(bool result);
```
frangio: Really bad name. :stuck_out_tongue: What about `WasPresent`...?

test/token/ERC721/ERC721.test.js
```diff
@@ -733,15 +717,15 @@ describe('ERC721', function () {
       describe('when the index is greater than or equal to the total tokens owned by the given address', function () {
         it('reverts', async function () {
           await expectRevert(
-            this.token.tokenOfOwnerByIndex(owner, 2), 'ERC721Enumerable: owner index out of bounds'
+            this.token.tokenOfOwnerByIndex(owner, 2), 'Enumerableset: index out of bounds'
```
frangio: The `s` should be capital `S`, right?

contracts/token/ERC721/ERC721.sol
```diff
@@ -17,21 +18,22 @@ import "../../utils/Counters.sol";
...
+    using EnumerableMap for EnumerableMap.UintToAddressMap;
 
     // Equals to `bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))`
     // which can be also obtained as `IERC721Receiver(0).onERC721Received.selector`
     bytes4 private constant _ERC721_RECEIVED = 0x150b7a02;
 
-    // Mapping from token ID to owner
-    mapping (uint256 => address) private _tokenOwner;
+    // Mapping from holder address to their (enumerable) set of owned tokens
+    mapping (address => EnumerableSet.Uint256Set) private _holderTokens;
```
frangio: Consider the name `_ownedTokens`.

contracts/token/ERC721/ERC721.sol
```diff
@@ -120,10 +110,8 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
      * @return address currently marked as the owner of the given token ID
      */
     function ownerOf(uint256 tokenId) public view override returns (address) {
-        address owner = _tokenOwner[tokenId];
-        require(owner != address(0), "ERC721: owner query for nonexistent token");
-
-        return owner;
+        require(_tokenOwners.contains(tokenId), "ERC721: owner query for nonexistent token");
+        return _tokenOwners.get(tokenId);
```
frangio: Can you try to reduce the number of storage reads?

contracts/token/ERC721/ERC721.sol
frangio: I don't know how to fix this but we can at least add a comment clearing up the potential confusion.

contracts/token/ERC721/ERC721.sol
```diff
@@ -528,81 +495,6 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
-    function _removeTokenFromAllTokensEnumeration(uint256 tokenId) private {
-        // To prevent a gap in the tokens array, we store the last token in the index of the token to delete, and
-        // then delete the last slot (swap and pop).
-
-        uint256 lastTokenIndex = _allTokens.length.sub(1);
-        uint256 tokenIndex = _allTokensIndex[tokenId];
-
-        // When the token to delete is the last token, the swap operation is unnecessary. However, since this occurs so
-        // rarely (when the last minted token is burnt) that we still do the swap here to avoid the gas cost of adding
-        // an 'if' statement (like in _removeTokenFromOwnerEnumeration)
```
frangio: Did we preserve this optimization in the `EnumerableMap`?

contracts/utils/EnumerableMap.sol
```diff
@@ -0,0 +1,182 @@
...
+     */
+    function add(Map storage map, bytes32 key, bytes32 value) private returns (bool) {
+        if (!contains(map, key)) {
+            map._entries.push(MapEntry({ _key: key, _value: value }));
+            // The entry is stored at length-1, but we add 1 to all indexes
+            // and use 0 as a sentinel value
+            map._indexes[key] = map._entries.length;
+            return true;
+        } else {
+            map._entries[map._indexes[key] - 1]._value = value;
```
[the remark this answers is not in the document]
nventuro: It's silly because this is a simple optimization the compiler could make :( But yes, we can improve it at the expense of making the code a bit more fragile.

contracts/token/ERC721/ERC721.sol
nventuro: No, I added it back

contracts/utils/EnumerableMap.sol
```diff
@@ -0,0 +1,182 @@
...
+     * - `key` must be in the map.
+     */
+    function get(Map storage map, bytes32 key) private view returns (bytes32) {
+        require(contains(map, key), "EnumerableMap: nonexistent key");
+        return map._entries[map._indexes[key] - 1]._value; // All indexes are 1-based
+    }
+
+    // UintToAddressMap
+
+    struct UintToAddressMap {
```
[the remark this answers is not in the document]
nventuro: Hm, I didn't notice this. There would be no benefit, no, other than type safety. Would you rather both be named `Uint` or `Uint256`?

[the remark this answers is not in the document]
nventuro: This module will look a bit strange with the different overloads, I'm not sure what's the best way to document generics.

contracts/token/ERC721/ERC721.sol
nventuro: Note that we can remove the `require` and the only change would be the revert reason: `get` already performs that same check.

I thought about adding an overload for `get` (and perhaps `at`) that emits a custom error message instead of the generic ones, similar to `SafeMath.sub`. What do you think?

contracts/mocks/EnumerableMapMock.sol
nventuro: I changed the name to `OperationResult`. The return value is closer to `success`: it returns whether the call actually did anything or not, which for `add` happens when the key was _not_ present.

contracts/token/ERC721/ERC721.sol
frangio: Overload sounds good to me.

contracts/utils/EnumerableSet.sol
```diff
@@ -84,44 +88,121 @@ library EnumerableSet {
...
-    {
+    function contains(Set storage set, bytes32 value) private view returns (bool) {
         return set._indexes[value] != 0;
     }
 
     /**
-     * @dev Returns an array with all values in the set. O(N).
+     * @dev Returns the number of values on the set. O(1).
+     */
+    function length(Set storage set) private view returns (uint256) {
```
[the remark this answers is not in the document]
frangio: Would you mind commenting on this? `length` doesn't sound right to me.

contracts/utils/EnumerableMap.sol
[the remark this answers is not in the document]
frangio: Not sure what type safety you're talking about. I think `Uint256` would be more appropriate... But I don't like the verbosity.

contracts/utils/EnumerableSet.sol
```diff
@@ -18,24 +18,31 @@ pragma solidity ^0.6.0;
...
         // Position of the value in the `values` array, plus 1 because index 0
         // means a value is not in the set.
-        mapping (address => uint256) _indexes;
+        mapping (bytes32 => uint256) _indexes;
     }
 
     /**
      * @dev Add a value to a set. O(1).
      *
      * Returns false if the value was already in the set.
```
frangio, line 42: I think this needs to be improved. I had misunderstood, I thought both `add` and `remove` returned whether the value was present.

```suggestion
     * Returns true if the value was added to the set, that is if it was not already present.
```

contracts/utils/EnumerableSet.sol
```diff
@@ -51,23 +58,27 @@ library EnumerableSet {
      *
      * Returns false if the value was not present in the set.
```
frangio, line 59:
```suggestion
     * Returns true if the value was removed from the set, that is if it was present.
```

contracts/utils/EnumerableSet.sol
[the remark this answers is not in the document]
nventuro: I think this is highly subjective and ultimately decided by context: in C++ for instance `vector` has `size`, as does every other STL data structure, but in Rust `vec` has `len`, as do `HashMap` and `HashSet`. 

I chose `length` because its what Solidity uses for its array types (both `[]` and `bytes`).

frangio:
I think this is good to go. The only thing pending is deciding whether we want `UintToAddressMap` or `Uint256ToAddressMap`.

frangio:
Any news with respect to `enumerate`?

nventuro:
It can be done using inline assembly, but it triggers tons of shadowing warnings due to assembly keywords that clash with function names like `add`. I'll leave that out of scope for this rc, and see if we can get it back before the final release.

KaiRo-at:
Ugh, why are you doing this? I have contracts that intentionally use different (e.g. lighter-gas-cost variants) of ERC721Enumerable and ERC721Metadata and with this refactoring I cannot base them on OpenZeppelin's ERC721 implementation any more but need to copy ERC721.sol with just the inheritance lines changed :(

KaiRo-at:
FWIW, the previous variant of what I'm doing (and would like to do again with OpenZeppelin 3.0) is at https://etherscan.io/address/0x7e789E2dd1340971De0A9bca35b14AC0939Aa330#code in the ERC721EnumerableSimple contract (starting line 534) and the tokenURI() function in line 921 and following, for which I meanwhile created my own variant of the ERC721Metadata contract in newer (as yet unpublished) work. When you create a large number of NFTs (e.g. 150k as we did for Crypto stamp Edition 1), you want minting cost to be as small as possible...

nventuro:
That's very interesting @KaiRo-at, thanks a lot for your feedback!

We intend to support that use case for `tokenURI()` by default (https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1745). The `_mint` optimization is interesting however, I had not heard of such an implementation before.

If I understand your code correctly, what you're basically doing is:
 - disallowing `_burn`
 - making all `tokenId`s sequential

These two together mean that there is no need for a tokenIndex -> tokenId mapping, since the ids and indexes match. You _do_ still need the id -> owner mapping (`ownerOf()`) and to track the total number of tokens (`totalSupply()`), so it's two out of the four relevant storage writes that we'd saving (not counting `tokensOfOwnerByIndex()` related writes).

Now, there's multiple reasons why we decided to merge all ERC721 flavors: to reduce library complexity, make the code easier to read (e.g. by using `EnumerableMap`), and simplify usage of the library. After all, pretty much all ERC721 tokens include both the metadata and enumerable extensions. However, it does prevent these sort of alternative implementations, and we definitely would rather users didn't have to copy-paste parts of the code in order to achieve their goals.

I'm not sure what the best way to proceed regarding this is. On one hand, the requirements for this optimization are rather high (the ability to mint arbitrary `tokenIds` is often critical) and I wouldn't want to drop our usage of `EnumerableMap`, but on the other, failing to provide a way to reuse our code is a big issue.

KaiRo-at:
Interesting to hear about #1745 - that would be great.

For the rest, you understood correctly, we save 2 stores on empty slots on minting as those are a series of collectibles that have sequential IDs anyhow (and cannot be burned by design, which becomes a requirement with this optimization). Even those 2 stores sum up quite a bit if you mint 150k NFTs up-front like we did for Crypto Stamp Edition 1 (and we will do a similar thing again).

That said, arbitrary token IDs of course need to be possible for other tokens, we have use cases ourselves where we need that (and have been using the default enumerable implementation).

In Solidity 0.5 days, I would have just gone and overridden the affected functions and be done (see the `tokenURI()` function I pointed to), but with the requirement of Solidity 0.6 to mark all to-be-overridden functions `virtual`, this kind of flexible pattern has been killed in many cases and I fear we will sadly have to resort to just a copy&modify pattern instead of having only our own changes isolated in our contracts that override/extend OpenZeppelin, which is not your fault at all, but Solidity's (and IMHO a mis-feature on their side).

Unfortunately, the merging of contracts for simplifying the OpenZeppelin library (as nice and clean it looks by itself) exacerbates that problem as it requires me to copy and replace the whole ERC-721 implementation instead of only parts of it. The only thing that could potentially save things somewhat is to mark almost everything virtual so people can override whatever they want, but of course that also makes it much easier for people to shoot themselves in the foot by forgetting to override something that is involved as well.

frangio:
Note that we have indeed decided to mark all non-view functions virtual so it will be possible to override them. See the [forum post](https://forum.openzeppelin.com/t/how-we-plan-to-adopt-solidity-0-6-virtual-functions/2063).

I'm not sure if we should include burn by default though, given that in ERC20 the public burn function is a separate extension.

KaiRo-at:
@frangio, it's enough to have `_burn() internal virtual` in there as `ERC721.sol` has now. That allows to use it where needed or even override it and put restrictions on it (I sometimes prefer to override it with a function that just `revert`s - or do the equivalent in a `_beforeTokenTransfer() override` - just to make it explicit that burning cannot happen).
On the gas-saving stuff, I see that with the `EnumerableMap` for `_tokenOwners`, that looks quite slick design-wise but the `EnumerableMap` implores a gas cost of 3 empty-slot `SSTORE`s on adding any item whereas an unburnable NFT with 0-based sequential tokenIDs only would need 1 there. I completely see the appeal of the `EnumerableMap` but it even makes it hard to copy&modify the contract and remove the overhead there... Hmm, if a contract could override a `using...for` of its parent, I could use a stripped-down `EnumerableMap` that is basically just a mapping with additional functions... But that doesn't work, does it?

nventuro:
Sadly no, libraries don't work like that :/ That's an interesting idea though - I'd bring this up with the Solidity team and see what they think about it, they've been looking at their mechanisms for code reuse recently.

KaiRo-at:
So, I just ran our code through a test, and comparing numbers for a version with my `ERC721EnumerableSimple` ported to OpenZeppelin 3.0 beta.0 (before this change) and a version based on straight ERC721.sol from OpenZeppelin 3.0 RC.1, I see the latter requiring about 40k more gas per token, as I expected.

KaiRo-at:
And with just changing `EnumerableMap`, I can recover pretty much the whole regression, the code below is just for context on how I'm doing it:

```diff
--- openzeppelin-contracts/utils/EnumerableMap.sol      1985-10-26 09:15:00.000000000 +0100
+++ contracts/EnumerableMapSimple.sol   2020-04-14 18:07:43.454981171 +0200
@@ -1,6 +1,6 @@
 pragma solidity ^0.6.0;
 
-library EnumerableMap {
+library EnumerableMapSimple {
     // To implement this library for multiple types with as little code
     // repetition as possible, we write it in terms of a generic Map type with
     // bytes32 keys and values.
@@ -10,18 +10,9 @@
     // This means that we can only create new EnumerableMaps for types that fit
     // in bytes32.
 
-    struct MapEntry {
-        bytes32 _key;
-        bytes32 _value;
-    }
-
     struct Map {
         // Storage of map keys and values
-        MapEntry[] _entries;
-
-        // Position of the entry defined by a key in the `entries` array, plus 1
-        // because index 0 means a key is not in the map.
-        mapping (bytes32 => uint256) _indexes;
+        bytes32[] _entries;
     }
 
     /**
@@ -32,17 +23,14 @@
      * already present.
      */
     function _set(Map storage map, bytes32 key, bytes32 value) private returns (bool) {
-        // We read and store the key's index to prevent multiple reads from the same storage slot
-        uint256 keyIndex = map._indexes[key];
+        uint256 uintKey = uint256(key);
+        require(uintKey <= map._entries.length, "Cannot add entry that is not connected to existing IDs");
 
-        if (keyIndex == 0) { // Equivalent to !contains(map, key)
-            map._entries.push(MapEntry({ _key: key, _value: value }));
-            // The entry is stored at length-1, but we add 1 to all indexes
-            // and use 0 as a sentinel value
-            map._indexes[key] = map._entries.length;
+        if (uintKey == map._entries.length) { // add new entry
+            map._entries.push(value);
             return true;
         } else {
-            map._entries[keyIndex - 1]._value = value;
+            map._entries[uintKey] = value;
             return false;
         }
     }
@@ -52,45 +40,15 @@
      *
      * Returns true if the key was removed from the map, that is if it was present.
      */
-    function _remove(Map storage map, bytes32 key) private returns (bool) {
-        // We read and store the key's index to prevent multiple reads from the same storage slot
-        uint256 keyIndex = map._indexes[key];
-
-        if (keyIndex != 0) { // Equivalent to contains(map, key)
-            // To delete a key-value pair from the _entries array in O(1), we swap the entry to delete with the last one
-            // in the array, and then remove the last entry (sometimes called as 'swap and pop').
-            // This modifies the order of the array, as noted in {at}.
-
-            uint256 toDeleteIndex = keyIndex - 1;
-            uint256 lastIndex = map._entries.length - 1;
-
-            // When the entry to delete is the last one, the swap operation is unnecessary. However, since this occurs
-            // so rarely, we still do the swap anyway to avoid the gas cost of adding an 'if' statement.
-
-            MapEntry storage lastEntry = map._entries[lastIndex];
-
-            // Move the last entry to the index where the entry to delete is
-            map._entries[toDeleteIndex] = lastEntry;
-            // Update the index for the moved entry
-            map._indexes[lastEntry._key] = toDeleteIndex + 1; // All indexes are 1-based
-
-            // Delete the slot where the moved entry was stored
-            map._entries.pop();
-
-            // Delete the index for the deleted slot
-            delete map._indexes[key];
-
-            return true;
-        } else {
-            return false;
-        }
+    function _remove(Map storage /*map*/, bytes32 /*key*/) private pure returns (bool) {
+        revert("No removal supported");
     }
 
     /**
      * @dev Returns true if the key is in the map. O(1).
      */
     function _contains(Map storage map, bytes32 key) private view returns (bool) {
-        return map._indexes[key] != 0;
+        return uint256(key) < map._entries.length;
     }
 
     /**
@@ -112,9 +70,7 @@
     */
     function _at(Map storage map, uint256 index) private view returns (bytes32, bytes32) {
         require(map._entries.length > index, "EnumerableMap: index out of bounds");
-
-        MapEntry storage entry = map._entries[index];
-        return (entry._key, entry._value);
+        return (bytes32(index), map._entries[index]);
     }
 
     /**
@@ -132,9 +88,9 @@
      * @dev Same as {_get}, with a custom error message when `key` is not in the map.
      */
     function _get(Map storage map, bytes32 key, string memory errorMessage) private view returns (bytes32) {
-        uint256 keyIndex = map._indexes[key];
-        require(keyIndex != 0, errorMessage); // Equivalent to contains(map, key)
-        return map._entries[keyIndex - 1]._value; // All indexes are 1-based
+        uint256 uintKey = uint256(key);
+        require(map._entries.length > uintKey, errorMessage); // Equivalent to contains(map, key)
+        return map._entries[uintKey];
     }
 
     // UintToAddressMap
@@ -159,7 +115,7 @@
      *
      * Returns true if the key was removed from the map, that is if it was present.
      */
-    function remove(UintToAddressMap storage map, uint256 key) internal returns (bool) {
+    function remove(UintToAddressMap storage map, uint256 key) internal view returns (bool) {
         return _remove(map._inner, bytes32(key));
     }
 ```

frangio:
@KaiRo-at Can you please create a new issue describing the problems in full? I'm starting to lose track of the context here.

KaiRo-at:
> @KaiRo-at Can you please create a new issue describing the problems in full? I'm starting to lose track of the context here.

Yes, sorry, not the best style to start a discussion on a closed PR. I created #2187 for this issue, summarizing what this all is about (and putting that diff into another comment on there).

PR #2149, opened 2020-03-26
Bundle ERC721 extensions into base contract

nventuro:
Part of #2086.

The goal behind this initiative is to have all optional extensions for tokens baked into the base implementation (that is, for `ERC20` to include `ERC20Detailed`, and `ERC721` to include both `ERC721Metadata` and `ERC721Enumerable`). These optional extensions were often (always?) used along the base implementation, and the [changes in Solidity around inheritance](https://github.com/ethereum/solidity/issues/8141) mean that forcing our users to use multiple inheritance create unnecessary friction.

As a side effect, we should be able to use the newly added `EnumerableSet` and the proposals around `EnumerableMap` (#2072) to improve code readability, gas efficiency and overall security.

Because the tokens are some of the library's biggest contracts, these changes are not trivial. I thought it best to approach this in a step-by-step manner, by:
 * first merging all ERC721 implementations and refactoring the tests (with no changes to the internals of the Solidity code)
 * then refactoring the ERC721 internals
 * then doing the same on ERC20

This PR contains just the first step of the three described above. To make reviewing easier, I split the changes to the contracts code in different commits. The tests refactor is unfortunately hard to review, though I am confident it was performed correctly (I copy-pasted the tests from ERC721.behavior and ERC721Full.test into ERC721.test).

Issue #2086, opened 2020-02-11
OpenZeppelin Contracts v3.0 Pending Changes


nventuro:
This is a non-comprehensive list of changes we want to introduce in the v3.0, the upcoming major release. We will use this opportunity to introduce some breaking changes by removing contracts we don't see much demand for (such as Crowdsales), and improving others we could can be done better (like Roles and Access Control in general).

For more information about the v3.0 release, head to [our roadmap](#2068).

 - [x] Migrate to Solidity v0.6.x [#2080]
 - [x] Design a new access control solution (see the discussion in [this forum post](https://forum.openzeppelin.com/t/redesigning-accesscontrol-for-the-openzeppelin-contracts/2177))
 - [x] Include optional extensions to some of our standard implementations (such as `ERC20Detailed`, possibly `ERC721Metadata` and `ERC721Enumerable`)
 - [x] Review contracts in `drafts` and either remove or promote out of there
 - [x] Make IERC721 `contract`s `interface`s instead (see [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2080#discussion_r377276127))
 - [x] Remove contracts or libraries made obsolete by v0.6, such as `Address.toPayable()` (https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2118)
 - [x] Review naming conventions, including usage of leading underscores in `internal` functions in contracts but not in libraries
 - [x] Consider marking some contracts that should only be used via inheritance `abstract` (such as `Ownable` and `ERC20`)
 - [x] Close all issues on the [v3.0 milestone](https://github.com/OpenZeppelin/openzeppelin-contracts/milestone/8)
 - [x] Perform some [directory and naming cleanup](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2060)
 - [x] Fix documentation website build (requires fixing https://github.com/OpenZeppelin/solidity-docgen/issues/133)
 - [x] Review naming of libraries: singular vs plural, `Strings`, `Counters`, `EnumerableSet`...
 - [x] Review malleability protections in `ECDSA.recover`

frangio:
We should review the situation around `ECDSA.recover` and the measures against malleability. There's been many issues with Ganache because of that change.

gitpusha:
[...]

PR #2161, opened 2020-03-30
Bundle ERC20Detailed

nventuro:
Part of #2086 

This is the ERC20 counterpart of #2149. Like in the case of ERC721, the derived contracts were made `abstract`.

I also had to fix some Create2 tests that relied on ERC20 having no constructor.


contracts/token/ERC20/ERC20.sol
```diff
@@ -37,6 +37,52 @@ contract ERC20 is Context, IERC20 {
...
+    string private _name;
+    string private _symbol;
+    uint8 private _decimals;
+
+    /**
+     * @dev Sets the values for `name`, `symbol`, and `decimals`. All three of
+     * these values are immutable: they can only be set once during
+     * construction.
+     */
+    constructor (string memory name, string memory symbol, uint8 decimals) public {
```
frangio: What do you think about hardcoding `decimals = 18`? There may be some interesting discussion around this in https://github.com/ethereum/EIPs/issues/724.

We could add a `_setDecimals` function as an escape hatch.

contracts/token/ERC20/ERC20Burnable.sol
```diff
@@ -8,7 +8,7 @@ import "./ERC20.sol";
  * tokens and those that they have an allowance for, in a way that can be
  * recognized off-chain (via event analysis).
  */
-contract ERC20Burnable is Context, ERC20 {
+abstract contract ERC20Burnable is Context, ERC20 {
```
frangio: I'm slightly concerned about the user experience around these abstract contracts. Have you played around with using them to see what it feels like, what errors the compiler gives, etc.?

nventuro: Yes, I think error message is reasonable:
```
Compilation errors: 
contracts/mocks/ERC20BurnableMock.sol:5:1: TypeError: Contract "ERC20BurnableMock" should be marked as abstract.
contract ERC20BurnableMock is ERC20, ERC20Burnable {
contracts/token/ERC20/ERC20.sol:49:5: Missing implementation:
    constructor (string memory name, string memory symbol, uint8 decimals) public {
```

It does have an issue issue in that its recommendation is wrong ('contract should be abstract'), but it does point in the direction of the constructor.

contracts/token/ERC20/ERC20.sol
nventuro: We'd be opening the door to `decimals` changing after deployment, but I don't think I'll lose much sleep over that.

contracts/token/ERC20/ERC20Burnable.sol
frangio: I think it's a terrible error message but I agree it can be improved on the compiler side.

contracts/token/ERC20/ERC20.sol
frangio: Unless we make the function "constructor only"/

CHANGELOG.md
```diff
@@ -29,7 +29,7 @@
  * `EnumerableSet`: renamed `get` to `at`. ([#2151](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2151))
  * `ERC165Checker`: functions no longer have a leading underscore. ([#2150](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2150))
  * `ERC20Detailed`: this contract was removed and its functionality merged into `ERC20`. ([#2161](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2161))
- * `ERC20`: added a constructor for `name` and `symbol`, `decimals` default to 18. ([#2161](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2161))
+ * `ERC20`: added a constructor for `name` and `symbol`. `decimals` default to 18. ([#2161](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2161))
```
frangio, line 32:
```suggestion
 * `ERC20`: added a constructor for `name` and `symbol`. `decimals` now defaults to 18. ([#2161](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2161))
```

PR #2162, opened 2020-03-30
Remove 'external' functions

nventuro:
The goal behind making every function overrideable (`virtual`) is to, well, allow for overrides. Often these are 'extensions' of the original function, adding new checks, events, side effects, etc., and therefore require calling into the original:

```solidity
function foo() override {
  super.foo();
  do_more_things();
}
```

However, this is not possible if `foo` is `external`, because `super.foo()` is considered an internal function call by the language (!).

This PR changes all `external` functions and makes them `public`. This creates a slight issue where `external` and `internal` functions had similar names: they will now both be internally-callable, which may lead to hard to detect mistakes. A prime example were `AccessControl`'s `grantRole` and `_grantRole`. After some consideration, we decided to avoid these scenarios, by either renaming the `internal` functions, or removing them where we considered they weren't really necessary.

This PR is best reviewed commit-by-commit.

### DIFF
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -162,7 +162,7 @@ abstract contract AccessControl is Context {
      * event.
      *
      * NOTE: Unlike {grantRole}, this function doesn't perform any checks on the
-     * calling account. Its usage should be restricted to grating the initial
+     * calling account. Its usage should be restricted to granting the initial
      * set of rules during contract construction.
      */
     function _setupRole(bytes32 role, address account) internal virtual {
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -159,13 +159,15 @@ abstract contract AccessControl is Context {
      * @dev Grants `role` to `account`.
      *
      * If `account` had not been already granted `role`, emits a {RoleGranted}
-     * event.
+     * event. Note that unlike {grantRole}, this function doesn't perform any
+     * checks on the calling account.
      *
-     * NOTE: Unlike {grantRole}, this function doesn't perform any checks on the
-     * calling account. Its usage should be restricted to granting the initial
-     * set of rules during contract construction.
+     * Requirements:
+     *
+     * - this function can only be called from a constructor.
      */
     function _setupRole(bytes32 role, address account) internal virtual {
+        require(_isConstructor(), "AccessControl: roles cannot be setup after construction");
         _grantRole(role, account);
     }
 
@@ -186,5 +188,19 @@ abstract contract AccessControl is Context {
         if (_roles[role].members.remove(account)) {
             emit RoleRevoked(role, account, _msgSender());
         }
+    }
+
+    // @dev Returns true if and only if the function is running in the constructor
+    function _isConstructor() private view returns (bool) {
+        // extcodesize checks the size of the code stored in an address, and
+        // address returns the current address. Since the code is still not
+        // deployed when running a constructor, any checks on its code size will
+        // yield zero, making it an effective way to detect if a contract is
+        // under construction or not.
+        address self = address(this);
+        uint256 cs;
+        // solhint-disable-next-line no-inline-assembly
+        assembly { cs := extcodesize(self) }
+        return cs == 0;
     }
 }
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -25,10 +25,7 @@ import "../GSN/Context.sol";
  * }
  * ```
  *
- * Roles can be granted and revoked programatically by calling the `internal`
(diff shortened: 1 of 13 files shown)


contracts/access/AccessControl.sol
```diff
@@ -161,22 +160,13 @@ abstract contract AccessControl is Context {
...
-            emit RoleGranted(role, account, _msgSender());
-        }
-    }
-
-    /**
-     * @dev Revokes `role` from `account`.
      *
-     * If `account` had been granted `role`, emits a {RoleRevoked} event.
+     * NOTE: Unlike {grantRole}, this function doesn't perform any checks on the
+     * calling account. Its usage should be restricted to grating the initial
```
frangio:
```suggestion
     * calling account. Its usage should be restricted to granting the initial
```

contracts/access/AccessControl.sol
```diff
@@ -161,22 +160,13 @@ abstract contract AccessControl is Context {
...
+     * set of rules during contract construction.
      */
-    function _revokeRole(bytes32 role, address account) internal virtual {
-        if (_roles[role].members.remove(account)) {
-            emit RoleRevoked(role, account, _msgSender());
-        }
+    function _setupRole(bytes32 role, address account) internal virtual {
```
frangio: We can restrict this function to work only in the constructor. See [`isConstructror`](https://github.com/OpenZeppelin/openzeppelin-sdk/blob/8bf20f7793ef86548519f8bda1d7eb6341013bbb/packages/lib/contracts/Initializable.sol#L47-L58). This is what I had in mind.

There is a caveat when initialization is done after construction, like an `initialize` function. However, in proxies the initializer runs in the constructor of the proxy as well. In cases other than proxies it should always be possible to have code in the constructor.

test/access/AccessControl.test.js
```diff
@@ -17,6 +17,15 @@ describe('AccessControl', function () {
     this.accessControl = await AccessControlMock.new({ from: admin });
   });
 
+  describe('_setupRole', function () {
+    it('cannot be called outside the constructor', async function () {
```
frangio, line 21: Is there a test that it _can_ be used in the constructor?

contracts/access/AccessControl.sol
```diff
@@ -187,4 +189,18 @@ abstract contract AccessControl is Context {
...
+    function _isConstructor() private view returns (bool) {
+        // extcodesize checks the size of the code stored in an address, and
+        // address returns the current address. Since the code is still not
+        // deployed when running a constructor, any checks on its code size will
+        // yield zero, making it an effective way to detect if a contract is
+        // under construction or not.
+        address self = address(this);
+        uint256 cs;
+        // solhint-disable-next-line no-inline-assembly
+        assembly { cs := extcodesize(self) }
```
frangio, line 203: We should use `extcodehash` because it's cheaper. Actually, shouldn't we just use `Address.isContract`?

test/access/AccessControl.test.js
nventuro: Yes, the mock uses it on construction.

contracts/access/AccessControl.sol
nventuro: Huh that's a great point.

albertocuestacanada:
Hi @nventuro, I didn't have time to reply faster. I've got some concerns about the changes to `AccessControl.sol`.

`_grantRole` and `_revokeRole` are now private, and the only option for an extending contract to grant roles is to use `_setupRole`, which can only be called from a constructor.

With these options, a true dynamic role hierarchy is disallowed. Roles must be known at deployment time. That restricts the library unnecessarily, in my opinion.

There are a number of cases that require roles to be created after the deployment of the contract. The ones I've found are when a multi-user platform is built that allows users to deploy new contracts instances through a factory.

For example, you could have an ERC20 factory that allows users to deploy contracts, and those contracts have some roles associated (minting, for example). You will need a new minter role created with each ERC20 instance. You probably want to have all roles in the same contract so that the same role can be reused for several ERC20 instances.

More especifically, I've used this pattern before to create a tokenized security platform. Different security issuers use the platform to create securities that then are traded in a common market. There is some governance across the platform for which the roles are needed.

Another example where I've done the same is for a supply chain platform. Different companies in the supply chain can tokenize their assets in the platform. Those companies remain in control of their tokenized assets through access control. Each company can create as many roles as they want to partition control over their own assets.

I don't see a problem with `grantRole` being public and `_grantRole` being internal. It's the same pattern used for `ERC20.sol` (`transfer` and `_transfer`) and by now everyone is used to it.

Would it be possible to either make `setupRole` callable outside a constructor, or go back to the internal `_grantRole` and a public `grantRole`?

Issue #2163, opened 2020-04-01
Request for hybrid Detailed-Mintable ERC20 Token


bh2smith:
**🧐 Motivation**
When testing or scripting with smart contracts involving arithmetic between multiple ERC20 tokens (like an Exchange) it can be very useful to have access to a `DetailedMintableERC20` which is both a `MintableERC20` and `DetailedERC20` so that tokens can be minted for deposit, withdraw etc... while having access to relevant values such as number of decimals (for computing exchange rates and such). 
The above described example is already use in both gnosis/dex-contracts [here](https://github.com/gnosis/dex-contracts/blob/master/contracts/test/DeployableERC20Detailed.sol)
and gnosis/dex-liquidity-provision [here](https://github.com/gnosis/dex-liquidity-provision/blob/master/contracts/test/DetailedMintableToken.sol) (soon to be more).

**📝 Details**
This simple feature should amount to the introduction of a new contract that is both Mintable and Detailed ERC20

Example implementations could be either of:

```js
pragma solidity ^0.5.0;

import "./ERC20Detailed.sol";
import "./ERC20Mintable.sol";

contract ERC20DetailedMintable is ERC20Mintable, ERC20Detailed {
    constructor(string memory name, string memory symbol, uint8 decimals) public ERC20Detailed(name, symbol, decimals) {}
}
```

abcoathup:
Hi @bh2smith!  Thanks for the suggestion, it is really appreciated.

The project owner has linked to #2167 which should address this suggestion.

bh2smith:
> The project owner has linked to #2167 which should address this suggestion.

Yes it does thanks. That along with the fact that `ERC20Detailed`has been recently merged into `ERC20` should be exactly what I was hoping for.

PR #2167, opened 2020-04-02
Add 'deploy-ready' token contracts

nventuro:
This adds deploy-ready variants for ERC20 and ERC721. These are burnable (holders can burn their tokens), mintable (minters can mint tokens arbitrarily) and pausable (pausers can pause all token transfers).

They will serve both as an example on how to integrate different OpenZeppelin contracts and use Access Control features, and as simple tokens that can be deployed as-is for quick prototyping. Projects that used to rely on ERC20Mintable for these use cases can now switch the deploy-ready variants.

I decided to give the minter and pauser role to the deployer to make using the contracts simpler: no calls to `grantRole` have to be made. However, because the deployer is also the admin of these roles, new minters and pausers can be created.

Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2163

### DIFF
--- a/contracts/deploy-ready/ERC20DeployReady.sol
+++ b/contracts/deploy-ready/ERC20DeployReady.sol
@@ -1,0 +1,38 @@
+pragma solidity ^0.6.0;
+
+import "../access/AccessControl.sol";
+import "../GSN/Context.sol";
+import "../token/ERC20/ERC20.sol";
+import "../token/ERC20/ERC20Burnable.sol";
+import "../token/ERC20/ERC20Pausable.sol";
+
+contract ERC20DeployReady is Context, AccessControl, ERC20Burnable, ERC20Pausable {
+    bytes32 public constant MINTER_ROLE = keccak256("MINTER_ROLE");
+    bytes32 public constant PAUSER_ROLE = keccak256("PAUSER_ROLE");
+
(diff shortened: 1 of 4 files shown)


abcoathup:
I think the deploy ready version of ERC721 should include auto creation of token URIs.  
Of the few ERC721's I have created they all had auto creation of token URIs.

Would be great to get community input on this.

nventuro:
The base ERC721 will have auto-creation of tokenURIs, but that's a separate PR :)

contracts/deploy-ready/ERC20DeployReady.sol
```diff
@@ -0,0 +1,53 @@
...
+ *  - a minter role that allows for token minting (creation)
+ *  - a pauser role that allows to stop all token transfers
+ *
+ * The account that deploys the contract will be granted the minter role, the
+ * pauser role, and the default admin role, meaning it will be able to grant
+ * both the minter and pauser roles. See {AccessControl} for details.
+ *
+ * This contract serves as an example of how to integrate different {ERC20}
+ * modules and use the {AccessControl} contract to create different permissions.
+ * It can be deployed as-is for quick prototyping or testing.
```
frangio, line 22: It can also be used in production... We shouldn't frame these as simply examples. Not using these should be for intermediate-advanced usage.

PR #2141, opened 2020-03-20
Fix display of array arguments in documentation

frangio:
I don't recall why I added the `pass:normal` macro, but I looked through the generated site and couldn't find any problem caused by removing it. And this fixes the issues with displaying array arguments.

PR #2169, opened 2020-04-03
Add passthrough delimiters to avoid italics from underscores

frangio:
I think this was introduced in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2141.

Some underscores in function/argument names were causing italics to show up.

[https://5e8783f54de1950006a32aef--openzeppelin-contracts-docs.netlify.com/contracts/2.x/api/token/erc20#ERC20-_setupDecimals-uint8-](image)

The fix is to add these `++` delimiters to pass characters through without interpreting them as formatting.


nventuro:
Awesome, thanks!

PR #2170, opened 2020-04-07
Update Contracts docs to make examples compile

abcoathup:
Update Contracts docs to make examples compile


abcoathup:
I went through all the Contracts docs and tried to compile each example. 

What I wasn’t sure about was the order of `ERC20(“”,””) public` so may need ti go through and correct this.

PR #2171, opened 2020-04-08
Fix ReentrancyGuard for Proxy Pattern

BrendanChou:
Allows the ReentrancyGuard to be used in contracts that are proxied. Fixes two problems:
1. The ReentrancyGuard was not able to be used by new contracts that use the proxy pattern.
2. The ReentrancyGuard was not able to be used by old contracts that use the proxy pattern and want to move to new implementation contracts that use this new ReentrancyGuard.

In both cases, the "corrupted" value of the `_notEntered` slot (zero in the case of new contracts, non-zero in the case of existing contracts) is now fixed in the first call to any `nonReentrant` function, making it backwards-compatible with any previous version of `ReentrancyGuard`

Another added benefit is that this should decrease the gas costs of this function as `uint256` values use the entire value of the slot and thus do not have to be checked by bytecode that the storage slot is returning a valid boolean value upon each `SLOAD` or `SSTORE`

I invite anyone to take-on and clean-up this PR as necessary


contracts/utils/ReentrancyGuard.sol
```diff
@@ -17,7 +17,10 @@ pragma solidity ^0.6.0;
  * https://blog.openzeppelin.com/reentrancy-after-istanbul/[Reentrancy After Istanbul].
  */
 contract ReentrancyGuard {
-    bool private _notEntered;
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = uint256(int256(-1));
+    
+    uint256 private _status;
```
frangio, line 23: This is not necessary but will improve the cost of the first execution, and is more correct in my opinion.

```suggestion
    uint256 private _status = _NOT_ENTERED;
```

contracts/utils/ReentrancyGuard.sol
```diff
@@ -17,7 +17,10 @@ pragma solidity ^0.6.0;
  * https://blog.openzeppelin.com/reentrancy-after-istanbul/[Reentrancy After Istanbul].
  */
 contract ReentrancyGuard {
-    bool private _notEntered;
+    uint256 private constant _NOT_ENTERED = 1;
+    uint256 private constant _ENTERED = uint256(int256(-1));
```
frangio, line 21: Is there a reason why you used this number in particular? It should be cheaper to use a smaller number because it can be pushed using `PUSH1` rather than `PUSH32`, which also results in smaller bytecode and is cheaper to deploy.

I think we can safely use the number 2 instead.

nventuro: I believe the reason why @BrendanChou used 0xff..ff is to make this backwards compatible (in a contract upgrade sense), by making sure the value of the reentrancy counter couldn't ever be equal to `_ENTERED`.

nventuro:
It's a sad state of affairs that using the right tools provided by the language, `bool` or even `enums`, result in code much less efficient than the hacky approach.

contracts/utils/ReentrancyGuard.sol
frangio: For that to happen there would have to be a "reentrant" upgrade, right? And in any case any number higher than 1 would have the same "entered" meaning so I think the behavior would remain the same, no?

nventuro: The old ReentrancyGuard had a counter with the number of times it had executed. For `require(_status != _ENTERED)` to pass, you'd want to make sure `_status` (the counter) is always different from `_ENTERED`, hence 0xff..ff.

frangio:
@BrendanChou Let us know if you want to work on this or if we should take over.

BrendanChou:
@frangio go ahead and take it over. I'm less familiar with the standards you all have for code quality and linting

contracts/utils/ReentrancyGuard.sol
nventuro: This is unnecessary: the constructor explicitly sets this value.

contracts/utils/ReentrancyGuard.sol
nventuro: Given that the plan isn't for this to be a storage-compatible upgrade from the v2.x code, I'd say we should go ahead and change these values to 1 and 2 so that they fit in a PUSH1. Can we use `enum`s for this, or will that incur added gas costs? 

I believe added range checks are performed on `enum` assignment, but I'd hope that's not the case when assigning constants.

BrendanChou: Range checks are often performed on enum assignments unfortunately. I'm not 100% sure about it doing it with constants

BrendanChou: As we're already not using `bool` or `enum` for storage, I see little point in making the constants into an enum. My 2 cents

nventuro: Ah true an `enum` would use `uint8` under the hood, silly me. Disregard that.

contracts/utils/ReentrancyGuard.sol
frangio: Right. :facepalm: Is there a reason for doing it in the constructor rather than the state variable declaration?

BrendanChou: I would just do it in whichever is the standard for these libraries. I don't think there is a difference in this case

nventuro: I am personally not a fan of initialization of values outside of constructors in Solidity. Also in the context of upgradeable contracts I like the presence of an explicit constructor as an indicator for the need of an initialization function.

frangio:
Not stale, we'll pick this up soon.

PR #2173, opened 2020-04-09
Updated ERC777 _beforeTokenTransfer hook calls to match ERC20

asselstine:
ERC20 calls _beforeTokenTransfer for both _mint and _burn, but ERC777 did not.

I think the ERC20 behaviour is ideal, as we then have hooks into the entire token lifecycle.

ERC777 has been updated to match.

### DIFF
--- a/contracts/token/ERC777/ERC777.sol
+++ b/contracts/token/ERC777/ERC777.sol
@@ -322,6 +322,8 @@ contract ERC777 is Context, IERC777, IERC20 {
 
         address operator = _msgSender();
 
+        _beforeTokenTransfer(operator, address(0), account, amount);
+
         // Update state variables
         _totalSupply = _totalSupply.add(amount);
         _balances[account] = _balances[account].add(amount);
@@ -381,6 +383,8 @@ contract ERC777 is Context, IERC777, IERC20 {
         require(from != address(0), "ERC777: burn from the zero address");
 
         address operator = _msgSender();
+
+        _beforeTokenTransfer(operator, from, address(0), amount);
 
         _callTokensToSend(operator, from, address(0), amount, data, operatorData);
 
--- a/scripts/gen-nav.js
+++ b/scripts/gen-nav.js
@@ -7,19 +7,19 @@ const startCase = require('lodash.startcase');
 const baseDir = process.argv[2];
 
 const files = proc.execFileSync(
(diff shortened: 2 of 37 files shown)


nventuro:
!! This is a great catch, thank you very much @asselstine! These should indeed affect all operations related to tokens, including minting and burning. I believe I mistakenly assumed that `_move` accounted for those.

In the process I noticed that [ERC20Pausable's documentation](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#ERC20Pausable) isn't accurate in that it doesn't mention this, nor do its tests check for the intended behavior. I'll address all of that while fixing this. Thanks again!

Out of curiosity, how did you find this issue?

asselstine:
I discovered it because of an incorrect assumption, actually.

I did a quick scan of the ERC777 code to see what hooks were available.  I didn't read the code fully.  As soon as I saw the "_beforeTransferToken" hook I just assumed I understood how it was used... but once I started testing my integration I discovered it did not work as intended.

It struck me as super weird that the hook wasn't called, as the Transfer events are fired for mint and burn, so I double checked the ERC20 spec.

It was a good find!

nventuro:
Awesome, thanks! Could you share more about what you're using the hooks for? It may serve as a useful example for our docs, or as inspiration for additions to the library!

asselstine:
Sure; I can share our use case.  The code is still under wraps, but it's for PoolTogether v3.

The hooks allow us to keep a secondary data structure perfectly up-to-date with token balances.  Specifically the [SortitionSumTree](https://github.com/pooltogether/sortition-sum-tree-factory).  This data structure allows us to randomly select a token holder using an integer as a token index, ensuring that the odds of selection are proportional to the balance of each user.

It would be more efficient to use the data structure as the store of balances directly, but keeping it separate from the balances makes the code more reusable and modular.

Additionally, we're looking at transfer restrictions.  If a certain amount of blocks have passed a user will be able to transfer their tokens.

nventuro:
That's very helpful, thanks! Does that data structure allow for O(1) retrieval?

That fits very well with what we had in mind when thinking about hooks. The limitation you mentioned around data redundancy is real, but I worry providing a mechanism to bypass how that works would also increase likelihood of bugs. Hopefully the added gas cost is not so much that this is a big issue for you?

Thanks again!

nventuro:
Closing in favor of #2191.

PR #2174, opened 2020-04-09
provide 'automatic' token URI by appending the token ID to the base URI

### LINKED ISSUE #1745: ERC721 "automatic" token URI

KaiRo-at:
If a base URI is set but no token-specific URI is available, this appends the token ID to the base URI. This should close #1745 🎉.

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -143,14 +143,16 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable 
 
         string memory _tokenURI = _tokenURIs[tokenId];
 
-        // If there is no token-specific URI but a base URI, append the token ID to the base URI.
-        // Otherwise, append the token URI to the base URI.
-        // In both cases, abi.encodePacked is being used to concatenate strings.
-        if (bytes(_tokenURI).length == 0 && bytes(_baseURI).length > 0) {
-            return string(abi.encodePacked(_baseURI, Strings.fromUint256(tokenId)));
-        } else {
+        // If there is no base URI, return the token URI.
+        if (bytes(_baseURI).length == 0) {
+            return _tokenURI;
+        }
+        // If both are set, concatenate the baseURI and tokenURI (via abi.encodePacked).
+        if (bytes(_tokenURI).length > 0) {
             return string(abi.encodePacked(_baseURI, _tokenURI));
         }
+        // If there is a baseURI but no tokenURI, concatenate the tokenID to the baseURI.
+        return string(abi.encodePacked(_baseURI, Strings.fromUint256(tokenId)));
     }
 
     /**
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -133,8 +133,9 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable 
     /**
      * @dev Returns the URI for a given token ID. May return an empty string.
      *
-     * If the token's URI is non-empty and a base URI was set (via
-     * {_setBaseURI}), it will be added to the token ID's URI as a prefix.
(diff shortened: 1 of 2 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -142,19 +143,19 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
 
         string memory _tokenURI = _tokenURIs[tokenId];
 
-        // Even if there is a base URI, it is only appended to non-empty token-specific URIs
-        if (bytes(_tokenURI).length == 0) {
-            return "";
+        // If there is no token-specific URI but a base URI, append the token ID to the base URI.
+        // Otherwise, append the token URI to the base URI.
+        // In both cases, abi.encodePacked is being used to concatenate strings.
+        if (bytes(_tokenURI).length == 0 && bytes(_baseURI).length > 0) {
```
nventuro, line 149: Could this be written as follows? I feel this way we emphasize the fact that setting a base URI is an opt-in for the automatic token URI behavior.

```solidity
// If there is no base URI, return the token URI
if (bytes(_baseURI).length == 0) {
    return _tokenURI;
// Otherwise concatenate the baseURI and tokenURI
} else if (bytes(_tokenURI).length > 0) {
    return string(abi.encodePacked(_baseURI, _tokenURI)); 
// If there is no tokenURI, concatenate the tokenID to the base URI
} else {
    return string(abi.encodePacked(_baseURI, Strings.fromUint256(tokenId))); 
}
```

KaiRo-at: Sure, I'll update the PR tomorrow. And actually, we don't need to use `else` at all, as `return` omits the rest anyhow. Or do you prefer having explicit `else` branches?

nventuro: I have no particular preference over early returns and exhaustive if-else blocks :shrug: The big thing is we're now making it explicit that there's 3 different ways this can work, which was not clear on the original version.

test/token/ERC721/ERC721.test.js
```diff
@@ -92,10 +92,10 @@ describe('ERC721', function () {
         expect(await this.token.tokenURI(firstTokenId)).to.be.equal(newBaseURI + sampleUri);
       });
 
-      it('token URI is empty for tokens with no URI but with base URI', async function () {
+      it('token URI is appended to base URI for tokens with no URI', async function () {
```
nventuro, line 95:
```suggestion
      it('tokenId is appended to base URI for tokens with no URI', async function () {
```

nventuro:
Thanks a lot! 

Would you mind adding a short explanation to `tokenURI` and `baseURI` that goes over what `tokenURI` returns in each case, and how a non-empty `baseURI` changes the behavior of `tokenURI`? I can take care of that otherwise, and then we'll be good to go!

test/token/ERC721/ERC721.test.js
KaiRo-at: Oh, of course. I'm merging this with the other comments changes mentioned below.

KaiRo-at:
OK, I added some updates to the comments, as requested, including the suggestion for the test comment. I hope this is good to go now!

PR #2178, opened 2020-04-11
Add .circleci/config.yml

puchaw:
Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,84 +1,13 @@
-version: 2
-# 2.1 does not yet support local run
-# unless with workaround. For simplicity just use it.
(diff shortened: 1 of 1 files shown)


KaiRo-at:
I smell something weird going on here. My guess would be that this user may be trying something sinister.

abcoathup:
Closing as appears to be in error (potentially spam)

KaiRo-at:
@abcoathup thanks for that. This user also made some confusing comments in #2174 (that's what led me to dig up what else this user has been doing and find this).

frangio:
I've blocked the user from the OpenZeppelin organization.

PR #2130, opened 2020-03-16
Port ERC 1155 branch to Solidity 0.6 (and current master)

KaiRo-at:
This includes the previous work for #1803 my PR #2029 and also makes all the ERC1155 work compile with Solidity 0.6, rebased against current master.

If it's not the right way to submit a PR against master but I should instead do this in a different way, please let me know.

PR #2029, opened 2019-12-19
Add a simple catch-all implementation of the metadata URI interface

KaiRo-at:
As mentioned in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2014#issuecomment-565499083 this is a simple implementation of the metadata URI part of ERC1155.

I'm still a bit confused on how to do tests for OpenZeppelin (this is my first PR to the project), so I have not added tests in this PR for now. Help would be appreciated.

Issue #2003, opened 2019-11-19
Improve extensibility of ERC1155


frangio:
@KaiRo-at suggested that our current contract is not extensible enough due to using `external` functions.

They also suggested to add an internal function that implements transfers to have a central extensibility point. I think this is fine but I would like to see the internal function use `msg.sender` rather than having a `sender` argument like we did in ERC20.

KaiRo-at:
Seems like the [Augur implementation](https://github.com/AugurProject/augur/blob/master/packages/augur-core/source/contracts/libraries/token/ERC1155.sol) does have that internal function structure already. I think its naming of functions is not completely congruent with previous OpenZeppeling token implementations, and it implements some non-standard (but useful) extensions like totalSupply tracking, so it also could not be taken 1:1 but it can inspire for what to do.

KaiRo-at:
This may be mostly bogus now as Solidity 0.6 and OpenZeppelin 3.0 seems to be geared towards WET rather than DRY (due to explicit overrides on Solidity's side and reducing library complexity on the OpenZeppelin side) and any changes that are not strictly extensions to the default OpenZeppelin implementation need copying of the whole contract code for other tokens as well from how I understand it.

nventuro:
One of our top goals is definitively avoiding code repetition and users copy-pasting the library sources, this issue remains relevant.

We've already removed `external` from all other functions in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2162 and introduced _hooks_ as an extensibility point with this very goal in mind (see [the code for ERC777 for example](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.0.0-rc.1/contracts/token/ERC777/ERC777.sol#L405)), so it'd simply be a matter of implementing these on ERC1155: design work should be ready.

KaiRo-at:
Hmm, if enough `virtual`s are spread throughout this and people can override those functions (I forgot for a moment that functions marked `virtual` do not actually require an override), then it may still work and be helpful, yes.

KaiRo-at:
I've been pointed to https://twitter.com/xanderatallah/status/1232124941425881089 today and that makes me wonder if there is a generic way to do at least an `internal` implementation of `_exists()` like we have in ERC721, but that's not really easy the way ERC1155 is built in general.
It can be done in some specific implementations, though.

frangio:
(@KaiRo-at I edited your comment just to add diff highlighting.)

[...]

PR #2185, opened 2020-04-14
Introduce an ERC-1155 _exists() function

KaiRo-at:
This PR is based on a discussion started at https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2003#issuecomment-612109928 and should give the ERC-1155 implementation more compatibility to OpenZeppelin's ERC-721 implementation and also better compatibility to services like OpenSea as pointed to by https://twitter.com/xanderatallah/status/1232124941425881089

The main thing introduced here is an `_exists()` function that is checked in a few places where we'd otherwise return a null-ish default value (`0` or `''`), and to know if a token ID is valid, it has to be registered internally first - either via minting a first token on that ID, or by calling an explicit function.

Tests will fail on this PR for the moment as it bases on both #2130 and #2029 - also, I have not written any tests for this PR itself yet as it's mostly a suggestion on how we could do this in a good way.

### DIFF
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -31,8 +31,8 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     // Used as the URI for all token types by relying on ID substition, e.g. https://token-cdn-domain/{id}.json
     string private _uri;
 
-    // Mapping token ID to that token being registered as existing
-    mapping (uint256 => bool) private _tokenExists;
+    // Mapping token ID to that token being registered as existing (1 for existing, 0 for not existing)
+    mapping (uint256 => uint256) private _tokenExists;
 
     /*
      *     bytes4(keccak256('balanceOf(address,uint256)')) == 0x00fdd58e
@@ -88,7 +88,6 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
      */
     function balanceOf(address account, uint256 id) public view override returns (uint256) {
         require(account != address(0), "ERC1155: balance query for the zero address");
-        require(_exists(id), "ERC1155: balance query for nonexistent token");
         return _balances[id][account];
     }
 
@@ -234,28 +233,29 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     }
 
     /**
-     * @dev Register a token ID so other contract functionality knows this token
-     * actually exists and this ID is valid. Minting will automatically call this.
-     * Also emits an event to flag the existence, see the spec:
+     * @dev Register an `id` so other contract functionality knows this token actually
+     * exists and this ID is valid. Minting will automatically call this.
+     * If `sendEvent` is true, emits an event to flag the existence, see the spec:
      * "To broadcast the existence of a token ID with no initial balance, the contract
      * SHOULD emit the TransferSingle event from 0x0 to 0x0, with the token creator as
      * _operator, and a _value of 0."
-     * @param id uint256 ID of the token to register
+     * Minting will call this without sending an event as a similar event is sent for
+     * the minting itself already.
      */
-    function _registerToken(uint256 id) internal virtual {
-        if (!_tokenExists[id]) {
-            _tokenExists[id] = true;
-            emit TransferSingle(msg.sender, address(0), address(0), id, 0);
+    function _registerToken(uint256 id, bool sendEvent) internal virtual {
+        if (_tokenExists[id] == 0) {
+            _tokenExists[id] = 1;
+            if (sendEvent) {
+                emit TransferSingle(msg.sender, address(0), address(0), id, 0);
+            }
         }
     }
 
     /**
-     * @dev Returns whether the specified token exists. Use {_registerTokenID} to set this flag.
-     * @param id uint256 ID of the token to query the existence of
-     * @return bool whether the token exists
+     * @dev Returns whether the token `id` exists. Use {_registerToken} to set this flag.
      */
     function _exists(uint256 id) internal view returns (bool) {
-        return _tokenExists[id];
+        return _tokenExists[id] != 0;
     }
 
     /**
@@ -277,8 +277,9 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
         _beforeTokenTransfer(operator, address(0), account, _asSingletonArray(id), _asSingletonArray(amount), data);
 
         if (!_exists(id)) {
-            _registerToken(id);
+            _registerToken(id, false);
         }
+
         _balances[id][account] = _balances[id][account].add(amount);
         emit TransferSingle(operator, address(0), account, id, amount);
 
@@ -304,7 +305,7 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
 
         for (uint i = 0; i < ids.length; i++) {
             if (!_exists(ids[i])) {
-                _registerToken(ids[i]);
+                _registerToken(ids[i], false);
             }
             _balances[ids[i]][to] = amounts[i].add(_balances[ids[i]][to]);
         }
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -31,6 +31,9 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     // Used as the URI for all token types by relying on ID substition, e.g. https://token-cdn-domain/{id}.json
     string private _uri;
 
+    // Mapping token ID to that token being registered as existing
+    mapping (uint256 => bool) private _tokenExists;
+
     /*
      *     bytes4(keccak256('balanceOf(address,uint256)')) == 0x00fdd58e
      *     bytes4(keccak256('balanceOfBatch(address[],uint256[])')) == 0x4e1273f4
@@ -85,6 +88,7 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
      */
     function balanceOf(address account, uint256 id) public view override returns (uint256) {
         require(account != address(0), "ERC1155: balance query for the zero address");
+        require(_exists(id), "ERC1155: balance query for nonexistent token");
         return _balances[id][account];
     }
 
@@ -230,6 +234,31 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     }
 
     /**
+     * @dev Register a token ID so other contract functionality knows this token
+     * actually exists and this ID is valid. Minting will automatically call this.
+     * Also emits an event to flag the existence, see the spec:
+     * "To broadcast the existence of a token ID with no initial balance, the contract
+     * SHOULD emit the TransferSingle event from 0x0 to 0x0, with the token creator as
+     * _operator, and a _value of 0."
+     * @param id uint256 ID of the token to register
+     */
+    function _registerToken(uint256 id) internal virtual {
+        if (!_tokenExists[id]) {
+            _tokenExists[id] = true;
+            emit TransferSingle(msg.sender, address(0), address(0), id, 0);
+        }
+    }
+
+    /**
+     * @dev Returns whether the specified token exists. Use {_registerTokenID} to set this flag.
+     * @param id uint256 ID of the token to query the existence of
+     * @return bool whether the token exists
+     */
+    function _exists(uint256 id) internal view returns (bool) {
+        return _tokenExists[id];
+    }
+
+    /**
      * @dev Creates `amount` tokens of token type `id`, and assigns them to `account`.
      *
      * Emits a {TransferSingle} event.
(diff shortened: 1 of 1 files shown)


KaiRo-at:
One thing I somewhat wonder about is if requiring this explicit registration would cause any issues for people who want to use the ERC-1155 implementation from OpenZeppelin.
FWIW, for the project I work on right now, I will use this patch as it should work really well with it.

christopheradams:
I have some reservations about how this matches up with the ERC-1155 spec.

The section on [Enumerating from events] says that "In order to keep storage requirements light for contracts implementing ERC-1155, enumeration (discovering the IDs and values of tokens) must be done using event logs." Baking the token registration storage mapping into the default contact would seem to violate this principle.

Also, the [Metadata Extensions] reads "The uri function MUST NOT be used to check for the existence of a token as it is possible for an implementation to return a valid string even if the token does not exist." I'm afraid clients might intentionally or unintentionally treat the presence or absence of an error generated by `uri()` as a de facto check. Reading the spec, I'm not sure I would expect any of the view functions to ever return errors.

[Enumerating from events]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1155.md#enumerating-from-events
[Metadata Extensions]: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1155.md#metadata-extensions

KaiRo-at:
Hrm, that sounds like it's an unusable standard for most usages then as you only get reliable event logs with an archive node which is really expensive to set up and operate, and services like OpenSea need a good way to check which tokens are valid. Our implementation of the project in use right now will use the patch in this PR and then I probably will advocate for people not using ERC-1155 because of its multiple shortcomings, including those two quotes from the spec.

christopheradams:
Just for the record, I believe there are valid use cases for an ` _exists()` function and a mapping to keep track of it. For example, you need this if you want to guarantee that a token can only be minted once. I'm in favor of seeing a good implementation of this, but I don't think it should be the default (requiring contract storage it may not need), unless the spec changes.

nventuro:
Thanks for thos comments @christopheradams! Here's my interpretation of the spec:

>The uri function MUST NOT be used to check for the existence of a token as it is possible for an implementation to return a valid string even if the token does not exist.

I think this means _callers_ of `uri()` should not use it to check existence: that is, the fact that `uri(id)` returns with no errors doesn't necessarily mean that `id` exists. Nothing prevents us from adding a check ourselves if we add a notion of a token 'existing'.

>In order to keep storage requirements light for contracts implementing ERC-1155, enumeration (discovering the IDs and values of tokens) must be done using event logs.

Again, the specification doesn't forbid implementations from having storage-based enumeration (which would be silly) - rather, it says that enumerating using events only should be possible. I share @KaiRo-at's concerns when it comes to building state from events: both access to archival nodes and building state-aggregating software are not trivial tasks. 

I believe on-chain enumeration (sensibly applied) is a good practice, which is why we started introducing concepts such as [`EnumerableMap`](https://docs.openzeppelin.com/contracts/3.x/api/utils#EnumerableMap) to the library, and use it for [`AccessControl`](https://docs.openzeppelin.com/contracts/3.x/api/access#AccessControl). In that sense, we may even want to go as far as implementing this using [`EnumerableSet`](https://docs.openzeppelin.com/contracts/3.x/api/utils#EnumerableSet), which would let users not only check for existence but also enumerate all tokens in the system.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -193,6 +198,24 @@ contract ERC1155 is ERC165, IERC1155
         _doSafeBatchTransferAcceptanceCheck(msg.sender, from, to, ids, values, data);
     }
 
+    /**
+     * @dev Register a token ID so other contract functionality knows this token
+     * actually exists and this ID is valid. Minting will automatically call this.
+     * @param id uint256 ID of the token to register
+     */
+    function _registerToken(uint256 id) internal virtual {
+        _tokenExists[id] = true;
```
nventuro, line 207: With [EIP2200](https://eips.ethereum.org/EIPS/eip-2200), an `SSTORE` here will cost the same as an `SLOAD` if `_tokenExists[id]` is already set to `true`, so this is already optimal in that sense.

We may want to consider using a 32-byte word instead of a boolean however: see the comments [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2171#pullrequestreview-394923232) (tl;dr: storing booleans is more costly than full-words because of extra checks placed by the compiler).

nventuro: Also, from the 1155 EIP:
>To broadcast the existence of a token ID with no initial balance, the contract SHOULD emit the TransferSingle event from 0x0 to 0x0, with the token creator as _operator, and a _value of 0.

What do you think about emitting such an event here? We should only do it the first time though, so we'd want to introduce an `if` statement anyway, even if not required for efficiency reasons.

AC0DEM0NK3Y:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2185#issuecomment-620866299

Hi. On the above comment, forgive me if I'm wrong it's been a while since I looked at a node but from what I recall you do not need a archive node to get full history of events you need an archive node if you need full history of the state. 
You would need to wait for the events history to be synced also down to the right block on a normal/fast node to get full history for a contract (there are framework functions to query this) but this is way way less time and space than a full/archive node.

So, pulling all balance or uri events is fine from a normal node. Indeed this should be the case if you want to be able to track balances this way too which is way faster than having to hit the node to query.

IIRC the desire to not have something enforced on that front was not to put a gas burden on implementations as it can be inferred from a balance transfer event (either with a "create" style one or an actual transfer) and implementations such as TheSandbox and SkyWeaver did not want to pay this cost.

nventuro:
Thanks for that comment @AC0DEM0NK3Y, it seems I was mistaken when it comes to node types - I believed logs were pruned after some time on regular nodes. Indeed, [Infura seems to store all events](https://blog.infura.io/faster-logs-and-events-e43e2fa13773/) on a traditional database for fast access via the [`eth_getLogs`](https://infura.io/docs/ethereum/json-rpc/eth_getLogs) RPC call.

However, there seems to be many iniciatives to drop this expectation of nodes to keep logs forever. Logs are meant to be used for applications to react to activity on the blockchain, not as a form of inexpensive read-only off-chain storage. See [this great gist](https://gist.github.com/karalabe/60be7bef184c8ec286fc7ee2b35b0b5b#broken-invariants) from one of Geth lead devs on that topic.

All in all, I think this topic will become an issue in the near future and we should prepare for that. Additionally, my point about state-aggregation still stands.

AC0DEM0NK3Y:
Yeah I could see the desire to prune, although I would hope they would keep a mode on that at least allows you to say "sync the history for X,Y and Z contract from block number" it's something I put as a suggestion on parity features a while ago... something like tha would also be desirable for state.

Personally and I think what most serious implentations would do is sync your node event history down once, fill a db with events up to head and then just feed the db as events come in to give balance and uri updates (and flag existance).
Hitting the node for every request seems like it will get quite unwieldy.

KaiRo-at:
@nventuro Ah, thanks for the explanations on the spec, I let myself be misled about the interpretation as well after the comment here but re-reading the statements, i agree with you.
On the event log issue, I has some expectations that event logs may be available on fast-pruning nodes but learned the hard way that they are not - while they are not pruned at least by current Parity/OpenEthereum implementations, you only get the logs for any blocks the node has processed itself, so none from before it loaded a snapshot, and none if it may lag and jump to another snapshot (which is rare but can happen during sync or if it was offline for some reason. Even worse, you just get empty responses and no errors if you query for blocks it missed logs for. So we learned you need an archive node to be sure you have a complete event log. Of course, services like Infura or Etherscan probably operate archive nodes and also may copy the data they need into other databases they can query more efficiently.

AC0DEM0NK3Y:
You have to wait until it has synced the events history to be sure. Being "fully synced" to head block after catchup from a snapshot is not actually fully synced, you are only guaranteed to have events data from snapshot block at that point after which it will switch to history pull along with keeping up with head.
Both parity and geth aren't very good at signalling this, iirc parity had a single number that you'll see increasing in the log that shows where its currently up to.

To be sure you either have to eye/parse out from the log or you can call a function on the node to query where it is up to. Web3 last time I looked at it didn't expose this well either but it was a simple call with ethers iirc.

contracts/token/ERC1155/ERC1155.sol
KaiRo-at: Ah, that is an interesting piece from the spec, I missed that one! I think it definitely would be a good idea to emit that one.
Also, I didn't even think of the double-registering case, thanks for your thoughts on that.
So, should I change `_tokenExists` to a `uint256` (using 0 and 1 for false and true) instead? I find it somewhat sad that a `bool` can end up being more expensive than a 32-byte type...

nventuro: Yes, I'm also not thrilled by that - we run into the same situation [on this other PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2171#issuecomment-615316756).

What do you think about using [`EnumerableSet`](https://docs.openzeppelin.com/contracts/3.x/api/utils#EnumerableSet) to store this data, instead of a plain mapping? As mentioned in the other comments, being able to enumerate this data is both important and hard to do via events. It'd mimic 721's [`tokenByIndex`](https://docs.openzeppelin.com/contracts/3.x/api/token/erc721#IERC721Enumerable-tokenByIndex-uint256-).

KaiRo-at: Well, an `EnumerableSet` mostly makes sense for the reverse indexing, which from what I see is completely unneded here and only costs additional gas. Is there any reason for having this really enumerable? Also, it would not change anything wrt `bool` vs. `uint` as `EnumerableSet` only works on `bytes32` values, potentially in the variant of `UintSet` where they are cast to `uint256` anyhow.

nventuro:
We intend to release v3.1 of Contracts soon, with initial support for ERC1155. Given that `_exists()` is not part of the standard, and we don't yet have consensus on some of the low level details, we intend to leave this improvement for v3.2.

KaiRo-at:
Mr. Bot, this is still relevant, I'll look into it after the releases of OpenZeppelin 3.1 and Crypto stamp 2.

KaiRo-at:
I'll come back to this soon, my dear bot.

frangio:
There seem to be multiple things being discussed here and I think we need to define more clearly what we want and why.

1. Adding an `exists` or `_exists` function.
2. Whether the `view` functions should revert unless a token exists.
3. Whether there should be enumerability.

And I'd add whether any of this should be enabled by default in our `ERC1155`.

Another option would be to make sure `exists` can be implemented by deriving `ERC1155` and that it does not require forking (i.e. using hooks).

---

Based on the tweet by OpenSea, I think adding an `exists` function makes sense. I wonder how much gas overhead it would introduce, but it sounds like most of it would only be incurred during minting, which I think would be acceptable.

Based on the concerns expressed by the EIP document and in the comments in this issue, I'd say the `view` functions should not revert. 

I would oppose enumerability by default, but I would consider it as an optional extension. This whole discussion about enumerability on-chain versus doing it based on events is something that we are often unsure about, though, and I think we need to have a deeper discussion about it.

nventuro:
>Based on the concerns expressed by the EIP document and in the comments in this issue, I'd say the view functions should not revert.

Which functions are these? `balanceOf` and `isApprovedForAll`?

frangio:
And I think also `url()`.

KaiRo-at:
I'm actually starting to look into this again now and hope to have a new PR this week.

KaiRo-at:
OK, I rebased to current `master` and updated the work somewhat, no tests yet, but would be great to know if that is the right direction to go now.

KaiRo-at:
I'm waiting for feedback on the current approach and for time to work on tests.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -31,6 +31,9 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     // Used as the URI for all token types by relying on ID substition, e.g. https://token-cdn-domain/{id}.json
     string private _uri;
 
+    // Mapping token ID to that token being registered as existing (1 for existing, 0 for not existing)
+    mapping (uint256 => uint256) private _tokenExists;
```
frangio, line 35: Why is this `uint256` instead of `bool`?

frangio:
What do you think about making `exists` public?

---

There is a potential issue related to upgradeable ERC1155 contracts. If contract is upgraded from the current ERC1155 version to the one in this PR, `exists` will return incorrect values because the mapping will not be populated with the already existing contracts. What do you think about that? Is there anything that could be done, other than exposing an external function to allow already existing tokens to be registered?

contracts/token/ERC1155/ERC1155.sol
KaiRo-at: That's because of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2185#discussion_r422364288 - do you think that's a bad idea?

KaiRo-at:
> What do you think about making `exists` public?

I mainly followed the pattern we also have in the ERC721 OpenZeppelin contract, leaving it to implementers to expose it publicly. But I'm open to making it public right away.

> There is a potential issue related to upgradeable ERC1155 contracts. If contract is upgraded from the current ERC1155 version to the one in this PR, `exists` will return incorrect values because the mapping will not be populated with the already existing contracts. What do you think about that? Is there anything that could be done, other than exposing an external function to allow already existing tokens to be registered?

I haven't thought about upgradeable contracts before as I usually don't do them, but yes, in this case, I think, yes, they would need to call _registerToken(uint256 id) with the IDs that have been created before.

PR #2186, opened 2020-04-14
Improve ERC20Snapshot documentation

frangio:
Motivated by recent questions in the Telegram group.

### DIFF
--- a/contracts/token/ERC20/ERC20Snapshot.sol
+++ b/contracts/token/ERC20/ERC20Snapshot.sol
@@ -64,7 +64,9 @@ abstract contract ERC20Snapshot is ERC20 {
     event Snapshot(uint256 id);
 
     /**
-     * @dev Creates a new snapshot and returns its snapshot id. Emits a {Snapshot} event that contains the same id.
+     * @dev Creates a new snapshot and returns its snapshot id. 
+     *
+     * Emits a {Snapshot} event that contains the same id.
      */
     function _snapshot() internal virtual returns (uint256) {
         _currentSnapshotId.increment();
--- a/contracts/token/ERC20/ERC20Snapshot.sol
+++ b/contracts/token/ERC20/ERC20Snapshot.sol
@@ -6,16 +6,36 @@ import "../../utils/Counters.sol";
 import "./ERC20.sol";
 
 /**
- * @dev ERC20 token with snapshots.
+ * @dev This contract extends an ERC20 token with a snapshot mechanism. When a snapshot is created, the balances and
+ * total supply at the time are recorded for later access.
  *
- * When a snapshot is made, the balances and total supply at the time of the snapshot are recorded for later
- * access.
+ * This can be used to safely create mechanisms based on token balances such as trustless dividends or weighted voting.
+ * In naive implementations it's possible to perform a "double spend" attack by reusing the same balance from different
+ * accounts. By using snapshots to calculate dividends or voting power, those attacks no longer apply. It can also be
+ * used to create an efficient ERC20 forking mechanism.
  *
- * To make a snapshot, call the {snapshot} function, which will emit the {Snapshot} event and return a snapshot id.
- * To get the total supply from a snapshot, call the function {totalSupplyAt} with the snapshot id.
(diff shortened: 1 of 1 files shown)


frangio:
image

These are shown in the documentation, but I don't think we want them to show, right? I think we need https://github.com/OpenZeppelin/solidity-docgen/issues/175 for this.

contracts/token/ERC20/ERC20Snapshot.sol
```diff
@@ -38,12 +58,13 @@ abstract contract ERC20Snapshot is ERC20 {
...
+    /**
+     * @dev Emitted by {_snapshot} when a snapshot is created. Contains the snapshot id that can be used for retrieval.
+     */
     event Snapshot(uint256 id);
 
     /**
-     * @dev Creates a new snapshot id. Balances are only stored in snapshots on demand: unless a snapshot was taken, a
-     * balance change will not be recorded. This means the extra added cost of storing snapshotted balances is only paid
-     * when required, but is also flexible enough that it allows for e.g. daily snapshots.
+     * @dev Creates a new snapshot and returns its snapshot id. Emits a {Snapshot} event that contains the same id.
```
nventuro, line 67:
```suggestion
     * @dev Creates a new snapshot and returns its snapshot id. 
     *
     * Emits a {Snapshot} event that contains the same id.
```

contracts/token/ERC20/ERC20Snapshot.sol
```diff
@@ -38,12 +58,13 @@ abstract contract ERC20Snapshot is ERC20 {
     // Snapshot ids increase monotonically, with the first value being 1. An id of 0 is invalid.
     Counters.Counter private _currentSnapshotId;
 
+    /**
+     * @dev Emitted by {_snapshot} when a snapshot is created. Contains the snapshot id that can be used for retrieval.
```
nventuro, line 62:
```suggestion
     * @dev Emitted by {_snapshot} when a snapshot identified by `id` is created.
```

nventuro:
Yes, being able to hide overridden functions would be useful.

frangio:
Good idea. Done.

contracts/token/ERC20/ERC20Snapshot.sol
```diff
@@ -38,12 +50,23 @@ abstract contract ERC20Snapshot is ERC20 {
...
+     *
+     * Emits a {Snapshot} event that contains the same id.
+     *
+     * {_snapshot} is internal so that you can decide if and how to expose it externally. If it is exposed, it can be
+     * guarded by something like {AccessControl} or open to the public. The latter is required for implementing certain
+     * trust minimization mechanisms such as forking, but you must consider that it can potentially be used by attackers in
+     * two ways. First, it can be used to increase the cost of retrieval of values from snapshots, although it will grow
+     * logarithmically thus rendering this attack ineffective in the long term. Second, it can be used to target specific
+     * accounts and increase the cost of ERC20 transfers for them, in the ways specified in the Gas Costs secion above.
+     * We haven't measured the actual numbers; if this is something you're interested in please reach out to us.
```
nventuro, line 69:
```suggestion
     * {_snapshot} is `internal`: you must decide how to expose it externally. This can be done both by
     * guarding it with a system such as {AccessControl}, or by leaving it open to the public. 
     *
     * [NOTE]
     * ====
     * While an open way of calling {_snapshot} is required for certain minimization mechanisms
     * such as forking, you must consider that it can potentially be used by attackers in
     * two ways. 
     *
     * First, it can be used to increase the cost of retrieval of values from snapshots, although it will grow
     * logarithmically thus rendering this attack ineffective in the long term. Second, it can be used to target specific
     * accounts and increase the cost of ERC20 transfers for them, in the ways specified in the Gas Costs section above.
     *
     * We haven't measured the actual numbers; if this is something you're interested in please reach out to us.
     * ====
```

nventuro: I tried to reduce the wall-of-textyness: https://deploy-preview-2186--openzeppelin-contracts-docs.netlify.com/contracts/2.x/api/token/erc20#ERC20Snapshot-_snapshot--

PR #2188, opened 2020-04-15
Rename Strings.fromUint256 to Strings.toString

### LINKED ISSUE #2156: utils/String `fromUint256` should probably be `toString`
Supposing we use the common convention of using such utility functions as a library:
```
contract Foo {
  using Strings for uint256;

  function test() returns(string memory) {
    uint i = 42;
    return i.fromUint256();
  }
}
```

The function name here tells us nothing - we already knew it was a `uint256` from the variable type. `toString` would be a much more intuitive name, and line up with what other languages use in the same situation.

frangio:
Fixes #2156

### DIFF
--- a/test/utils/Strings.test.js
+++ b/test/utils/Strings.test.js
@@ -12,15 +12,15 @@ describe('Strings', function () {
 
   describe('from uint256', function () {
     it('converts 0', async function () {
-      expect(await this.strings.fromUint256(0)).to.equal('0');
+      expect(await this.strings.toString(0)).to.equal('0');
     });
 
     it('converts a positive number', async function () {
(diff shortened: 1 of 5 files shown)


test/utils/Strings.test.js
```diff
@@ -12,15 +12,15 @@ describe('Strings', function () {
 
   describe('from uint256', function () {
     it('converts 0', async function () {
-      expect(await this.strings.fromUint256(0)).to.equal('0');
+      expect(await this.strings.toString(0)).to.equal('0');
```
nventuro, line 15: I think this is clashing with the JavaScript `toString` method, which causes the tests to fail. We should use the `contraxt.methods['toString()']` syntax.

PR #2191, opened 2020-04-15
Add missing hook to ERC777, fix relevant documentation

nventuro:
Based on the work by @asselstine in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2173.

This PR adds the missing hook to ERC777, as well as updates the documentation and tests for ERC20Pausable and ERC721Pausable in how they relate to this feature, which was not being properly tested.

### DIFF
--- a/contracts/token/ERC721/ERC721Pausable.sol
+++ b/contracts/token/ERC721/ERC721Pausable.sol
@@ -4,10 +4,20 @@ import "./ERC721.sol";
 import "../../utils/Pausable.sol";
 
 /**
- * @title ERC721 Non-Fungible Pausable token
- * @dev ERC721 modified with pausable transfers.
+ * @dev ERC721 token with pausable token transfers, minting and burning.
+ *
+ * Useful for scenarios such as preventing trades until the end of an evaluation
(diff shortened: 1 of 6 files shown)


contracts/token/ERC721/ERC721Pausable.sol
```diff
@@ -4,10 +4,20 @@ import "./ERC721.sol";
 import "../../utils/Pausable.sol";
 
 /**
- * @title ERC721 Non-Fungible Pausable token
- * @dev ERC721 modified with pausable transfers.
+ * @dev ERC721 token with pausable token transfers, minting and burning.
+ *
+ * Useful for scenarios such as preventing trades until the end of an evaluation
+ * period, or having an emergency switch for freezing all token transfers in the
+ * event of a large bug.
```
frangio, line 11: I feel like we should have a "See {Pausable}" link, and in that page have this kind of explanation.

nventuro: It'd be good to have ideas in Pausable, but I'd still add specific rationale whenever we add Pausable to some other contract.

PR #2192, opened 2020-04-15
ERC721 deploy ready fixes

nventuro:
This fixes two issues with ERC721MinterPauser (our deploy-ready variant):

 - it adds `baseURI` to the constructor, so that the automatic tokenURI mechanism introduced in #2174 can be used (thanks @abcoathup for the suggestion!)
 - it adds a `tokenURI` parameter to `mint`, so that new tokens can have their own URIs

I also updated some of the related documentation.


contracts/deploy-ready/ERC721MinterPauser.sol
```diff
@@ -28,27 +28,32 @@ contract ERC721MinterPauser is Context, AccessControl, ERC721Burnable, ERC721Pau
...
      *
-     * See {ERC721-_mint}.
+     * See {ERC721-_mint} and {ERC721-_setTokenURI}.
      *
      * Requirements:
      *
      * - the caller must have the `MINTER_ROLE`.
      */
-    function mint(address to, uint256 tokenId) public {
+    function mint(address to, uint256 tokenId, string memory tokenURI) public {
```
frangio, line 53: I don't like this one. I think it's important to keep the simple `mint`. We can consider adding a separate `mint` with a `tokenURI` argument, but I don't think it's necessary.

nventuro: Shouldn't we then make it mandatory to use a `baseURI`?

frangio: Not necessarily. Having a `baseURI` argument in the constructor is enough. If someone ones to use an empty string it's their choice to do so.

abcoathup: For deploy ready ERC721 my assumption is that most users will want automatic token URI and will use sequential tokenIDs. I would recommend being opinionated and not having a tokenID parameter in the mint function, so that sequential tokenIDs are generated. Non-sequential tokenIDs is then the exception. 

I think most users will just use the address parameter of mint but having tokenURI gives lots of flexibility.

frangio: It's an interesting idea, but it worries me that it's a bit untested. Also, what if people want the incremental token id feature but not the other things brought in by the deploy ready version? Anyway, I'm just being skeptical but this may actually be a good idea.

abcoathup: OpenSea's example uses an incremental token id with a `mint` only accepting an address, which is where I first saw the idea of automatic token URI.
https://github.com/ProjectOpenSea/opensea-creatures/blob/master/contracts/TradeableERC721Token.sol#L31

nventuro: I think it makes sense to leave this one as simple as possible. Thanks!

contracts/deploy-ready/ERC721MinterPauser.sol
```diff
@@ -21,34 +22,45 @@ import "../token/ERC721/ERC721Pausable.sol";
  * both the minter and pauser roles.
  */
 contract ERC721MinterPauser is Context, AccessControl, ERC721Burnable, ERC721Pausable {
```
frangio, line 24: The addition of auto incremented token ids makes this name no longer accurate, to be honest. I preferred the idea of "deploy ready" where we define whatever that means for each contract.

frangio: `ERC721OutOfTheBox`
`ERC721ReadyToUse`
`ERC721Standard`
`ERC721Essentials`

abcoathup: `ERC721DeployReady`

frangio: ERC721Preset

nventuro: I'd rather keep the name descriptive, we might have in the future multiple variants for each token type. And a non-descriptive name forces you to be on top of the docs to know what those are, I'd rather keep using them simple.

Regarding this one in particular, I'd consider dropping the Pausable bit from it, which is quite unrelated to the other features, and leave it as a mintable 721 with autogenerated IDs and URIs.

abcoathup: I’d like Pausable dropped. Minting with automatic token URIs should be well used. I don’t think there will be much need for pause.

frangio:
Why are we removing Pausable from ERC721 but not from the other two deploy-ready contracts?

contracts/deploy-ready/ERC721MinterAutoID.sol
```diff
@@ -0,0 +1,64 @@
...
+ *  - token ID and URI autogeneration
+ *
+ * This contract uses {AccessControl} to lock the minting functions using the
+ * minter role - head to its documentation for details.
+ *
+ * The account that deploys the contract will be granted the minter role and the
+ * default admin role, meaning it will be able to both mint and grant the minter
+ * role to other accounts.
+ */
+contract ERC721MinterAutoID is Context, AccessControl, ERC721Burnable {
```
frangio, line 23:
```suggestion
contract ERC721MinterAutoId is Context, AccessControl, ERC721Burnable {
```

contracts/presets/ERC721PresetMinterAutoIDPauser.sol
```diff
@@ -5,25 +5,28 @@ import "../GSN/Context.sol";
...
  *
- * The account that deploys the contract will be granted the minter role and the
- * default admin role, meaning it will be able to both mint and grant the minter
- * role to other accounts.
+ * The account that deploys the contract will be granted the minter and pauser
+ * roles, as well as the default admin role, which will let it grant both minter
+ * and pauser roles to aother accounts
  */
-contract ERC721MinterAutoID is Context, AccessControl, ERC721Burnable {
+contract ERC721PresetMinterAutoIdPauser is Context, AccessControl, ERC721Burnable, ERC721Pausable {
```
frangio, line 25: Please don't hate me.

```suggestion
contract ERC721PresetMinterPauserAutoId is Context, AccessControl, ERC721Burnable, ERC721Pausable {
```

PR #2195, opened 2020-04-16
Remove in-constructor requirements

### LINKED ISSUE #2193: AccessControl _grantRole should be internal
**🧐 Motivation**

To enable dynamic role management at runtime, an admin needs to be granted initially. The public function `grantRole()` requires than the admin role exists and there is already a bearer, so child contracts need to first be able to grant it.

Having `_grantRole()` as private blocks this.

**📝 Details**

Example of a Hierarchical structure using AccessControl here: https://github.com/HQ20/contracts/blob/master/contracts/access/Hierarchy.sol

```
pragma solidity ^0.6.0;
import "@openzeppelin/contracts/access/AccessControl.sol";


/**
 * @title Hierarchy
 * @author Alberto Cuesta Canada
 * @notice Implements a dynamic role structure for Roles
 */
contract Hierarchy is AccessControl {
    event AdminRoleSet(bytes32 roleId, bytes32 adminRoleId);

    /// @dev Add `root` as a member of the root role.
    constructor (address root) public {
        _grantRole(DEFAULT_ADMIN_ROLE, root);
    }

    /// @dev Restricted to members of the role passed as a parameter.
    modifier onlyMember(bytes32 roleId) {
        require(hasRole(roleId, msg.sender), "Restricted to members.");
        _;
    }

    /// @dev Create a new role with the specified admin role.
    function addRole(bytes32 roleId, bytes32 adminRoleId)
        public onlyMember(adminRoleId) {
        _setRoleAdmin(roleId, adminRoleId);
        emit AdminRoleSet(roleId, adminRoleId);
    }
}
```

`_grantRole()` in the constructor doesn't compile. Altering to internal will allow it.

nventuro:
The requirement for `_setupRole` and `_setupDecimals` to only be callable inside a constructor was too restrictive for no good reasons: a warning should be enough.

There are scenarios where these may need to be called outside of a constructor, such as during a state migration after a contract code upgrade.

Fixes #2193
Closes #2194

### DIFF
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -164,12 +164,17 @@ abstract contract AccessControl is Context {
      * event. Note that unlike {grantRole}, this function doesn't perform any
      * checks on the calling account.
      *
-     * Requirements:
+     * [WARNING]
+     * ====
+     * This function should only be called from the constructor when setting
+     * up the initial roles for the system.
      *
-     * - this function can only be called from a constructor.
+     * It being callable by other means goes against the spirit of
+     * {AccessControl}, as there would be ways to grant roles not tracked by
+     * {getRoleMemberCount} and {getRoleMember}.
+     * ====
      */
     function _setupRole(bytes32 role, address account) internal virtual {
-        require(!address(this).isContract(), "AccessControl: roles cannot be setup after construction");
         _grantRole(role, account);
     }
 
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -169,9 +169,8 @@ abstract contract AccessControl is Context {
      * This function should only be called from the constructor when setting
      * up the initial roles for the system.
      *
-     * It being callable by other means goes against the spirit of
-     * {AccessControl}, as there would be ways to grant roles not tracked by
-     * {getRoleMemberCount} and {getRoleMember}.
+     * Using this function in any other way is effectively circumventing the admin
+     * system imposed by {AccessControl}.
      * ====
      */
     function _setupRole(bytes32 role, address account) internal virtual {
--- a/contracts/mocks/AccessControlMock.sol
+++ b/contracts/mocks/AccessControlMock.sol
@@ -10,8 +10,4 @@ contract AccessControlMock is AccessControl {
     function setRoleAdmin(bytes32 roleId, bytes32 adminRoleId) public {
         _setRoleAdmin(roleId, adminRoleId);
     }
-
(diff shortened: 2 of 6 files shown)


frangio:
I think this fixes #2193 and closes #2194. Can you check and let them know?

contracts/access/AccessControl.sol
```diff
@@ -164,12 +164,17 @@ abstract contract AccessControl is Context {
...
-     *
-     * - this function can only be called from a constructor.
+     * [WARNING]
+     * ====
+     * This function should only be called from the constructor when setting
+     * up the initial roles for the system.
+     *
+     * It being callable by other means goes against the spirit of
+     * {AccessControl}, as there would be ways to grant roles not tracked by
+     * {getRoleMemberCount} and {getRoleMember}.
```
frangio, line 174: I don't understand this. Can you explain?

nventuro: Say you have the following code:

```solidity
function giveRole(to, role) {
   if (condition) {
      _setupRole(role, to);
   }
}
```

Then there is a way to give roles without any account being its role admin. Part of the idea behind AccessControl is being able to determine who can grant new roles: calling this function outside of a constructor undermines this.

frangio: What I don't understand is what `getRoleMemberCount` and `getRoleMember` have to do with that.

nventuro: Oh, those are the methods that are used to query which accounts have a role, so it's their guarantees that would be broken in a wy.

nventuro: For example, `getRoleMemberCount` returning zero for a given role would normally mean that any role for which that one is admin can no longer be granted, but usage of `_setupRole` changes that.

frangio: Okay. I don't think that point is clear. I'll make a suggestion.

frangio, line 174:
```suggestion
     * Using this function in any other way is effectively circumventing the admin
     * system imposed by {AccessControl}.
```

Putting it this way, I'm getting the feeling that we should rename the function to something like `_unsafeGrantRole`.

frangio: An alternative idea I just had is to allow `_setupRole` only when the member count for the role is zero. In that case though we may need a function to add multiple accounts to the role.

nventuro: I think your wording is better, yes.

The issue with checking for zero members is that may naturally happen during the lifetime of the contract as roles are revoked and renounced.

PR #2196, opened 2020-04-16
Rename libraries to singular

nventuro:
Part of #2086


nventuro:
Performing this renaming causes issues because the `Counter` struct shadows the `Counter` library, which emits a warning. We could rename everything to plural, but then we'd have things like `Addresses.isContract`, which looks strange.

Ultimately, we decided that since the `using` syntax doesn't make sense in either singular nor plural form anyway (`using Strings for string`, `using String for string`), we'd rather not do any renaming at all to reduce the number of breaking changes.

PR #1678, opened 2019-03-13
SafeMath to use Outvars Again

wjmelements:
#### Changes
I measured a regression in add and sub when updating SafeMath from openzeppelin-solidity version 1.9.0 to 2.0.1.
The regression went away when I reverted the outvar change and made `c` an outvar again.
The regression was measured on solc 0.4.24 for an upcoming TrustToken smart contract, and seemed to amount to 5 gas per call.
This patches the regression by switching back to outvars.
Reviewers @frangio @nventuro

PR #2198, opened 2020-04-17
Improve gas costs of SafeMath

moodysalem:
These changes should improve the gas costs

### DIFF
--- a/contracts/math/SafeMath.sol
+++ b/contracts/math/SafeMath.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.6.0;
+pragma solidity ^0.5.0;
 
 /**
  * @dev Wrappers over Solidity's arithmetic operations with added overflow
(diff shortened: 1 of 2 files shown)


nventuro:
Thanks for contributing @moodysalem! 

Unfortunately I'll be closing this PR for the same reasons outlined in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/1678 (see [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/1678#issuecomment-472620753). 

For some information on why Solidity fails to emit optimal code and how they are addressing that, see [here](https://github.com/ethereum/solidity/issues/6075#issuecomment-472622304). It seems like the plans to have Yul as an IR are moving along nicely, which should result in gas improvements across the board for all Solidity users, and make micro-tuning like the one here unnecessary.

PR #2199, opened 2020-04-17
Improve Hooks documentation

nventuro:
This ended up being not just about hooks but extending contracts in general, which I feel is a topic in which we lacked material.

### DIFF
--- a/contracts/utils/EnumerableMap.sol
+++ b/contracts/utils/EnumerableMap.sol
@@ -12,8 +12,6 @@ pragma solidity ^0.6.0;
  * - Entries are enumerated in O(n). No guarantees are made on the ordering.
  *
  * ```
- * import "@openzeppelin/contracts/utils/EnumerableMap.sol";
- *
  * contract Example {
  *     // Add the library methods
  *     using EnumerableMap for EnumerableMap.UintToAddressMap;
--- a/contracts/utils/EnumerableMap.sol
+++ b/contracts/utils/EnumerableMap.sol
@@ -1,5 +1,20 @@
 pragma solidity ^0.6.0;
 
+/**
+ * @dev Library for managing an enumerable variant of Solidity's
+ * https://solidity.readthedocs.io/en/latest/types.html#mapping-types[`mapping`]
+ * type.
+ *
+ * Maps have the following properties:
+ *
+ * - Entries are added, removed, and checked for existence in constant time
+ * (O(1)).
+ * - Entries are enumerated in O(n). No guarantees are made on the ordering.
+ *
+ * As of v3.0.0, only `uint256 -> address` maps are supported.
+ *
+ * Include with `using EnumerableMap for EnumerableMap.UintToAddressMap;`.
+ */
 library EnumerableMap {
     // To implement this library for multiple types with as little code
     // repetition as possible, we write it in terms of a generic Map type with
--- a/contracts/utils/EnumerableMap.sol
+++ b/contracts/utils/EnumerableMap.sol
@@ -11,9 +11,20 @@ pragma solidity ^0.6.0;
  * (O(1)).
  * - Entries are enumerated in O(n). No guarantees are made on the ordering.
  *
- * As of v3.0.0, only `uint256 -> address` maps are supported.
+ * ```
+ * import "@openzeppelin/contracts/utils/EnumerableMap.sol";
  *
- * Include with `using EnumerableMap for EnumerableMap.UintToAddressMap;`.
+ * contract Example {
+ *     // Add the library methods
+ *     using EnumerableMap for EnumerableMap.UintToAddressMap;
+ *
+ *     // Declare a set state variable
+ *     EnumerableMap.UintToAddressMap private myMap;
+ * }
+ * ```
+ *
+ * As of v3.0.0, only maps of type `uint256 -> address` (`UintToAddressMap`) are
+ * supported.
  */
 library EnumerableMap {
     // To implement this library for multiple types with as little code
--- a/contracts/utils/EnumerableSet.sol
+++ b/contracts/utils/EnumerableSet.sol
@@ -12,8 +12,6 @@ pragma solidity ^0.6.0;
  * - Elements are enumerated in O(n). No guarantees are made on the ordering.
  *
  * ```
- * import "@openzeppelin/contracts/utils/EnumerableSet.sol";
- *
  * contract Example {
  *     // Add the library methods
  *     using EnumerableSet for EnumerableSet.AddressSet;
--- a/contracts/utils/EnumerableSet.sol
+++ b/contracts/utils/EnumerableSet.sol
@@ -11,9 +11,10 @@ pragma solidity ^0.6.0;
  * (O(1)).
  * - Elements are enumerated in O(n). No guarantees are made on the ordering.
  *
- * As of v2.5.0, only `address` sets are supported.
+ * As of v3.0.0, `address` and `uint256` sets are supported.
  *
- * Include with `using EnumerableSet for EnumerableSet.AddressSet;`.
+ * Include with `using EnumerableSet for EnumerableSet.AddressSet;` or
+ * `using EnumerableSet for EnumerableSet.UintSet;`.
  *
  * @author Alberto Cuesta Cañada
  */
--- a/contracts/utils/EnumerableSet.sol
+++ b/contracts/utils/EnumerableSet.sol
@@ -11,12 +11,20 @@ pragma solidity ^0.6.0;
  * (O(1)).
  * - Elements are enumerated in O(n). No guarantees are made on the ordering.
  *
- * As of v3.0.0, `address` and `uint256` sets are supported.
+ * ```
+ * import "@openzeppelin/contracts/utils/EnumerableSet.sol";
  *
- * Include with `using EnumerableSet for EnumerableSet.AddressSet;` or
- * `using EnumerableSet for EnumerableSet.UintSet;`.
+ * contract Example {
+ *     // Add the library methods
+ *     using EnumerableSet for EnumerableSet.AddressSet;
  *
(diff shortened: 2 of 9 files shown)


contracts/utils/EnumerableMap.sol
```diff
@@ -1,5 +1,20 @@
...
+ *
+ * Maps have the following properties:
+ *
+ * - Entries are added, removed, and checked for existence in constant time
+ * (O(1)).
+ * - Entries are enumerated in O(n). No guarantees are made on the ordering.
+ *
+ * As of v3.0.0, only `uint256 -> address` maps are supported.
+ *
+ * Include with `using EnumerableMap for EnumerableMap.UintToAddressMap;`.
```
frangio, line 16: I think we should also show how to declare an enumerable map state variable.

contracts/utils/EnumerableSet.sol
```diff
@@ -11,9 +11,10 @@ pragma solidity ^0.6.0;
  * (O(1)).
  * - Elements are enumerated in O(n). No guarantees are made on the ordering.
  *
- * As of v2.5.0, only `address` sets are supported.
+ * As of v3.0.0, `address` and `uint256` sets are supported.
  *
- * Include with `using EnumerableSet for EnumerableSet.AddressSet;`.
+ * Include with `using EnumerableSet for EnumerableSet.AddressSet;` or
+ * `using EnumerableSet for EnumerableSet.UintSet;`.
```
frangio, line 17: Same as above, I think we should show how to declare one as a state variable.

contracts/utils/README.adoc
```diff
@@ -1,25 +1,49 @@
 = Utilities
 
-Miscellaneous contracts containing utility functions, often related to working with different data types.
+Miscellaneous contracts and libraries containing utility functions you can use to improve security, work with new data types, or safely use low-level primitives.
+
+Security tools include:
+
+ * {Pausable}: provides a simple way to halt activity in your contracts (often in reponse to an external threat).
+ * {ReentrancyGuard}: protects you from https://blog.openzeppelin.com/reentrancy-after-istanbul/[attackers performing reentrant calls].
```
frangio, line 8: Hm, reentrant calls could also happen by accident...

```suggestion
 * {ReentrancyGuard}: protects you from https://blog.openzeppelin.com/reentrancy-after-istanbul/[reentrant calls].
```

contracts/utils/README.adoc
```diff
@@ -1,25 +1,49 @@
...
+
+ * {Counters}: a simple way to get a counter that can only be incremented or decremented. Very useful for ID generation, counting contract activity, among others.
+ * {EnumerableMap}: like Solidity's https://solidity.readthedocs.io/en/latest/types.html#mapping-types[`mapping`] type, but with key-value _enumeration_: this will let you know how many entries a mapping has, and iterate over them (which is not possible with `mapping`).
+ * {EnumerableSet}: like {EnumerableMap}, but for https://en.wikipedia.org/wiki/Set_(abstract_data_type)[sets]. Can be used to store privileged accounts, issued IDs, etc.
+
+[NOTE]
+====
+Because Solidity does not support generic types, {EnumerableMap} and {EnumerableSet} are specialized to a limited number of key-value types.
+
+As of v3.0, {EnumerableMap} supports `uint256 -> address` (`UintToAddressMap`), and {EnumerableSet} supports sets of both `address` and `uint256` (`AddressSet` and `UintSet`).
```
frangio, line 22: Sometimes "both" makes it sound like it's both at the same time.

```suggestion
As of v3.0, {EnumerableMap} supports `uint256 -> address` (`UintToAddressMap`), and {EnumerableSet} supports `address` and `uint256` (`AddressSet` and `UintSet`).
```

docs/modules/ROOT/pages/extending-contracts.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+contract ModifiedAccessControl is AccessControl {
+    // Override the revokeRole function
+    function revokeRole(bytes32, address) public override {
+        revert("ModifiedAccessControl: cannot revoke roles");
+    }
+}
+```
+
+The old `revokeRole` is then replaced by our override, and any calls to it will immediately revert. We cannot _remove_ the function from the contract, but reverting on all calls is good enough.
```
frangio, line 33: Does this result in the bytecode for the functionality being omitted from the contract? That would be really interesting for people that are worried about that code lying around, form an attack surface perspective.

docs/modules/ROOT/pages/extending-contracts.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+=== Calling `super`
+
+Sometimes you want to _extend_ a parent's behavior, instead of outright changing it to something else. This is where `super` comes in.
+
+The `super` keyword will let you call functions defined in a parent contract, even if they overridden. This mechanism can be used to add additional checks to a function, emit events, or otherwise add functionality as you see fit.
```
frangio, line 39:
```suggestion
The `super` keyword will let you call functions defined in a parent contract, even if they are overridden. This mechanism can be used to add additional checks to a function, emit events, or otherwise add functionality as you see fit.
```

docs/modules/ROOT/pages/extending-contracts.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+TIP: For more information on how overrides work, head to the https://solidity.readthedocs.io/en/latest/contracts.html#index-17[official Solidity documentation].
```
frangio, line 41: I couldn't find a definitive answer but I'm not sure if "head to" is correct. It sounds really awkward to me.

```suggestion
TIP: For more information on how overrides work, head over to the https://solidity.readthedocs.io/en/latest/contracts.html#index-17[official Solidity documentation].
```

docs/modules/ROOT/pages/extending-contracts.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract ERC20WithSafeTransfer is ERC20 {
+    function _beforeTokenTransfer(address from, address to, uint256 amount)
+        internal virtual override
+    {
+        super._beforeTokenTransfer(from, to, amount);
+
+        require(_validRecipient(to), "ERC20WithSafeTransfer: invalid recipient");
```
frangio, line 89: Why is this after `super`?

docs/modules/ROOT/pages/extending-contracts.adoc
```diff
@@ -0,0 +1,123 @@
...
+
+So far, the only available hook is `_beforeTransferHook`, in all of xref:api:token/ERC20.adoc#ERC20-_beforeTokenTransfer-address-address-uint256-[`ERC20`], xref:api:token/ERC721.adoc#ERC721-_beforeTokenTransfer-address-address-uint256-[ERC721] and xref:api:token/ERC777.adoc#ERC777-_beforeTokenTransfer-address-address-address-uint256-[ERC777]. If you have ideas for new hooks, let us know!
+====
+
+=== Rules of Hooks
+
+There's a few guidelines you should follow when writing code that uses hooks in order to prevent issues. They are very simple, but do make sure you follow them:
+
+1. Whenever you override a parent's hook, re-apply the `virtual` attribute to the hook. That will allow child contracts to add more functionality to the hook.
+2. **Always** call the parent's hook in your override. This will make sure all hooks in the inheritance tree are called: contracts like xref:api:token/ERC20.adoc#ERC20Pausable[`ERC20Pausable`] rely on this behavior.
```
frangio, line 111:
```suggestion
2. **Always** call the parent's hook in your override using `super`. This will make sure all hooks in the inheritance tree are called: contracts like xref:api:token/ERC20.adoc#ERC20Pausable[`ERC20Pausable`] rely on this behavior.
```

docs/modules/ROOT/pages/extending-contracts.adoc
nventuro: No, AFAIK that's not possible. Even if you override it and replace it with an empty function, the selector entry will still be in the jump table.

docs/modules/ROOT/pages/extending-contracts.adoc
nventuro: The order is irrelevant. I always put it before because I find it's easier to comply with the second rule of hooks that way.

docs/modules/ROOT/pages/extending-contracts.adoc
frangio: Yes the selector, but I meant the bytecode for the actual functionality, i.e. the code from the parent that is being "shadowed".

docs/modules/ROOT/pages/extending-contracts.adoc
frangio: I think it's better practice to do requires first. It saves gas.

docs/modules/ROOT/pages/extending-contracts.adoc
nventuro: Ah, yes, of course. The old definition is replaced by the new one (unless `super` is used), so the bytecode is never generated.

docs/modules/ROOT/pages/extending-contracts.adoc
nventuro: You're calling _the hook_, not the actual transfer function. So what happens is:

```
base hook
child 1 hook
child 2 hook
transfer
```

The code for ERC20 is

```
fn transfer() -> {
   _beforeHook();
   doTransfer();
}
```

So any code in the hooks happens before the actual transfer. The order in which hooks execute respective to their place in the inheritance graph (the only thing affected by calling `super.hook` before or after) is irrelevant.

frangio: Oh, right!

Issue #2208, opened 2020-04-21
Add require statements to ERC777 internal functions


nventuro:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2027 made some functions, including `_approve`, `internal`. A [comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2027/files#diff-50790cbc7e9942af7885b70149d10000R412) indicated that doing this required that a commented-out `require` statement be restored, but this was not done.

This leads to some of the arguments of these internal functions not being properly validated, including allowing for `_approve`'s `holder` to be the zero address. We should:
 1. restore this require statement
 2. evaluate which arguments of the other functions made `internal` in that PR should be validated
 3. and backport the fix to v2.5

PR #2027, opened 2019-12-16
Made private methods internal to allow for overriding

MickdeGraaf:
Makes private functions in ERC777 internal instead of private.
 
Making these methods internal allows for overriding and adding of functionality in tokens developers create.

Fixes #2047 

Example usecase: 

In [DDAI](https://ddai.dexwallet.io) interest needs to be claimed on every transfer.
Overriding every functoin like transfer, transferfrom, send and so on would add unnecessary complexity to the code. Thats why I override the _move function.

To prevent people from making unnecessary forks of this repo(like I did) I think its  wise to make all functions internal instead of private.

Issue #1928, opened 2019-10-02
Add Documentation: ERC20Pausable, SafeERC20, TokenTimelock


deadpine:
These contracts' functions don't have explanations:

https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#ERC20Pausable
https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#SafeERC20
https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#TokenTimelock

GuillaumeCz:
According to the documentation, it seems that these function are actualy explained. 
@blancoagostina Can you please provide more details about what's mssing ? ;) 
...or maybe this issue is kind of outdated ?

abcoathup:
Hi @GuillaumeCz,
Thanks for querying.
An example is: 
https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#ERC20Pausable-transfer-address-uint256-

Compared to:
https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#IERC20-transfer-address-uint256-

GuillaumeCz:
Hi @abcoathup 
My bad ! I was looking to the contract's description itself instead of the functions ! 
Is anyone working on this issue ? 
I'm quite interested in having a try ;)

abcoathup:
Hi @GuillaumeCz!  

Thanks for offering to contribute, it is really appreciated.  
OpenZeppelin Contracts 3.0 is currently being developed, so best to wait for the project owner to advise on this.

Please wait until we have discussed this idea before writing any code or submitting a Pull Request, so we can go through the proposed changes beforehand.  We don’t want you to waste your time!

GuillaumeCz:
I would be proud to contribute to this nice project ! 

Please ping me when it's ok on your side ;)

nventuro:
Hey there @GuillaumeCz, thank you very much for the offer!

ERC20Pausable was simplified and now has proper docs, while TokenTimelock was removed from the library in the v3.0 release. If you're up to it, an area where better docs would definitely make a difference is ERC721: while there is some documentation, it is subpar. 

I'd recommend starting with the interfaces (IERC721, IERC721Metadata, IERC721Enumerable) and then going to ERC721, using the docs for ERC20 and ERC777 as guidelines regarding format, phrasing, etc. 

Thanks again!

GuillaumeCz:
I would be glad to do it ! ;)

[...]

PR #2212, opened 2020-04-22
Add missing requirements to ERC777

nventuro:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2208 for the v2.5.0 branch.

https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2027 made `_send`, `_approve`, `_callTokensToSend` and `_callTokensReceived` `internal`. We failed to identify that some of these functions contained arguments that were not validated: when they were `private`, this validation was performed at their callsite. This code was included in the v2.5.0 release.

Then, https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2134 got merged, which made the hook methods `private` again, and removed the `operator` argument from `_send` (the one that was unchecked). This code was published in v3.0.0.

This PR addresses the issues on the v2.5.0 release (and targets that branch). Once merged, v2.5.1 will be released and the branch merged back into `master`, only keeping the fix for `_approve` (the only unprotected function in v3.0.0). That will then become v3.0.1.

Because of this, I only added tests for the `require` statement in `_approve`: the other one will only exist in the v2.5.1 branch.

### DIFF
--- a/contracts/token/ERC777/ERC777.sol
+++ b/contracts/token/ERC777/ERC777.sol
@@ -408,10 +408,13 @@ contract ERC777 is Context, IERC777, IERC20 {
         emit Transfer(from, to, amount);
     }
 
+    /**
+     * @dev See {ERC20-_approve}.
+     *
+     * Note that accounts cannot have allowance issued by their operators.
+     */
     function _approve(address holder, address spender, uint256 value) internal {
-        // TODO: restore this require statement if this function becomes internal, or is called at a new callsite. It is
-        // currently unnecessary.
-        //require(holder != address(0), "ERC777: approve from the zero address");
+        require(holder != address(0), "ERC777: approve from the zero address");
         require(spender != address(0), "ERC777: approve to the zero address");
 
         _allowances[holder][spender] = value;
--- a/contracts/token/ERC777/ERC777.sol
+++ b/contracts/token/ERC777/ERC777.sol
@@ -352,6 +352,7 @@ contract ERC777 is Context, IERC777, IERC20 {
     )
         internal
     {
+        require(operator != address(0), "ERC777: operator is the zero address");
         require(from != address(0), "ERC777: send from the zero address");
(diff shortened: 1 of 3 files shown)


nventuro:
I forgot to add: I don't think we should add any checks to the hook functions. All they do is call into other contracts, which should have their own validations in place. No token contract should call these hooks outside of a token transfer regardless, which is why we made the functions `private` for v3.0.0.

contracts/token/ERC777/ERC777.sol
```diff
@@ -408,10 +409,13 @@ contract ERC777 is Context, IERC777, IERC20 {
         emit Transfer(from, to, amount);
     }
 
+    /**
+     * @dev See {ERC20-_approve}.
+     *
+     * Note that accounts cannot have allowance issued by their operators.
```
frangio, line 415: We're missing the "Requirements" section here. Did you leave it out because it's already in ERC20? I would still put them here I think, especially because the variables are named differently.

contracts/token/ERC777/ERC777.sol
```diff
@@ -352,6 +352,7 @@ contract ERC777 is Context, IERC777, IERC20 {
     )
         internal
     {
```
frangio, line 354: The description for this function can be improved.

contracts/token/ERC777/ERC777.sol
nventuro: I left it out because it's on ERC20 already (the same applies to `transfer`, `transferFrom` and friends).

The doc improvements will be carried over once we merge the 2.5.1 release branch into master, along with the changelog notice, etc.

contracts/token/ERC777/ERC777.sol
nventuro: I added a note in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1928

PR #2213, opened 2020-04-24
Merge v2.5.1 into v3.0.0

### LINKED ISSUE #2208: Add require statements to ERC777 internal functions

nventuro:
Continuation of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2212
Fixes #2208 

This ports the bugfix on the v2.5 branch into the v3.0 branch. Once this is merged, this will be the new v3.0.1 release.


nventuro:
Hm. I don't think there's much of a difference, though I wonder if squashing would make merging a future v2.5.2 harder. I'd think not, but don't really know.

frangio:
Our release guide says not to squash these kinds of merges.

nventuro:
Merge it is!

PR #2214, opened 2020-04-24
Feature/Adding RoleAdminChanged event in AccessControl

### LINKED ISSUE #2210: Add event for AccessControl._setRoleAdmin
Whenever a role's admin is set by [_setRoleAdmin](https://docs.openzeppelin.com/contracts/3.x/api/access#AccessControl-_setRoleAdmin-bytes32-bytes32-), we should emit an event indicating this happened.

Not only is this useful to react to admin changes, it will also help in dynamic role schemes, where a role hierarchy may be set up without any account having the roles.

julianmrodri:
#2210  

Fixes #2210

Adding a new event named "RoleAdminChanged" when the adminRole of a Role is changed. This event is emitted from with the _setRoleAdmin function.


julianmrodri:
I see 2 potential changes worth discussing:

- I put the Declaration of the new event before the existing ones, we can move it to the after the other two events.
- Should we only trigger the event in the case the new Role Admin is different than the existing Role Admin? (this would require to add a validation)

Looking forward for your comments!
.

julianmrodri:
@nventuro @frangio  Would really appreciate your inputs so we can put this puppy to bed! Thanks for your time!

contracts/access/AccessControl.sol
```diff
@@ -52,6 +52,11 @@ abstract contract AccessControl is Context {
 
     bytes32 public constant DEFAULT_ADMIN_ROLE = 0x00;
 
+    /**
+     * @dev Emitted when `adminRole` is set as ``role``'s admin role.
```
nventuro, line 56:
```suggestion
     * @dev Emitted when `adminRole` is set as ``role``'s admin role.
     *
     * `DEFAULT_ADMIN_ROLE` is the starting admin for all roles, despite
     * {RoleAdminChanged} not being emitted signaling this.
```

frangio:
I agree we should include the current admin role in the event as well, similar to the `OwnershipTransferred` event. And I think we should always emit it.

julianmrodri:
Sounds good! Thanks! Will implement the change to add the existing Admin Role and let you know.

julianmrodri:
@nventuro @frangio  I have implemented the changes you suggested. 

- I added the "_previousAdminRole_" to the event. I followed the same naming you are using in `OwnershipTransferred`.  I had to emit the event before updating the value now (the same is being done in 'OwnershipTransferred'). 
- Regarding the tests I had to break one statement into 3 lines due to **lint** issues (max length). 

Let me know what you think thanks for your support!

julianmrodri:
Thanks for your time and patience! for sure!

PR #2220, opened 2020-04-29
Fix SafeERC20.safeApprove

### LINKED ISSUE #2219: SafeERC20.safeApprove() Has unnecessary and unsecure added behavior
SafeERC20.safeApprove() should not have the added functionality of checking the existing allowance of the token. It should simply call `token.approve()`.

The purpose of this added code is to check that a user is either setting their allowance to zero, or setting it from zero. This is so that the user doesn't try to re-set their allowance from one non-zero number to another non-zero number where they can get front-run by the approved address using their allowance in-between those two transactions. However, this code doesn't actually solve the problem as the approval will still pass if the sandwich attack uses the rest of the remaining allowance.

Such behavior lends itself to bugs, unnecessary gas usage, and a false sense of security. In addition, there is no provided method in SafeERC20 that replicates the behavior of `approve` for tokens that are not ERC20 compliant, making this problem even more frustrating.

SafeERC20 already has `safeIncreaseAllowance` and `safeDecreaseAllowance` to solve this problem. Neither of which can be front-run.

Lastly, this allowance issue should not even be solved within this file. The intention of this file should solely be to wrap non-compliant ERC20 functions which was the original intention of this file.

```
 * @dev Wrappers around ERC20 operations that throw on failure (when the token
 * contract returns false). Tokens that return no value (and instead revert or
 * throw on failure) are also supported, non-reverting calls are assumed to be
 * successful.
```

BrendanChou:
Fixes #2219

### DIFF
--- a/contracts/token/ERC20/SafeERC20.sol
+++ b/contracts/token/ERC20/SafeERC20.sol
@@ -25,14 +25,12 @@ library SafeERC20 {
         _callOptionalReturn(token, abi.encodeWithSelector(token.transferFrom.selector, from, to, value));
     }
 
+    /**
+     * @dev safeApprove should only be called when setting an initial allowance,
+     * or when resetting it to zero. To increase and decrease it, use
+     * 'safeIncreaseAllowance' and 'safeDecreaseAllowance'.
+     */
     function safeApprove(IERC20 token, address spender, uint256 value) internal {
-        // safeApprove should only be called when setting an initial allowance,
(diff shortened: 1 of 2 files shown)


nventuro:
Closing in favor of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2268 as per the discussion in #2219. Thank you very much for your help @BrendanChou!

PR #2225, opened 2020-05-06
Making IERC721Receiver an interface

### LINKED ISSUE #2217: IERC721Receiver should be an 'interface'
Since I saw that IERC721 was converted from an 'abstract contract' to an 'interface', I tried to include IERC721Receiver in an 'interface' of mine (with OpenZeppelin 3.0.0) and was surprised that it has not been converted as well. This should probably be done.

julianmrodri:
#2217 

Fixes #2217 

Converting IERC721Receiver into an interface as it was missed in a previous PR and was still defined as an **abstract contract**.

### DIFF
--- a/contracts/token/ERC721/IERC721Receiver.sol
+++ b/contracts/token/ERC721/IERC721Receiver.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.6.2;
+pragma solidity ^0.6.0;
 
 /**
  * @title ERC721 token receiver interface
--- a/contracts/token/ERC721/IERC721Receiver.sol
+++ b/contracts/token/ERC721/IERC721Receiver.sol
@@ -1,11 +1,11 @@
-pragma solidity ^0.6.0;
+pragma solidity ^0.6.2;
 
 /**
  * @title ERC721 token receiver interface
  * @dev Interface for any contract that wants to support safeTransfers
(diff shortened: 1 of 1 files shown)


julianmrodri:
@nventuro @frangio  Ready to review when you have time. Thanks!

contracts/token/ERC721/IERC721Receiver.sol
```diff
@@ -1,11 +1,11 @@
-pragma solidity ^0.6.0;
+pragma solidity ^0.6.2;
```
frangio, line 1: I don't think this bump is necessary, is it? We want to keep the previous one to avoid a breaking change for our users.

julianmrodri: @frangio  I did it for consistency as all the other interfaces for ERC721 are in that version (^0.6.2) but can certainly remove (it would be different than the other Interfaces). Just let me know! Thanks!

frangio: Hm, I've forgot why the other interfaces have that pragma. But we should still keep this one as it is now if it's not necessary to bump.

nventuro: It's because the other ones inherit from each other, and interface inheritance was only added in 0.6.2.

PR #2229, opened 2020-05-08
Make Context abstract

frangio:
Other instances where we could do the same if we moved state variable initialization outside of the constructor:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3f2382278a238547b95716772f806dddcbccc7f9/contracts/utils/ReentrancyGuard.sol#L22-L30

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3f2382278a238547b95716772f806dddcbccc7f9/contracts/utils/Pausable.sol#L30-L32

### DIFF
--- a/contracts/GSN/Context.sol
+++ b/contracts/GSN/Context.sol
@@ -10,11 +10,7 @@ pragma solidity ^0.6.0;
  *
  * This contract is only required for intermediate, library-like contracts.
  */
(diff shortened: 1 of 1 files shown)


nventuro:
This is related to https://github.com/ethereum/solidity/issues/8162, `abstract` should be required.

Issue #2234, opened 2020-05-14
SPDX license identifiers missing


wbt:
As of today (Solidity 0.6.8, from [this PR](https://github.com/ethereum/solidity/pull/8907)), Solidity issues warnings on all OpenZeppelin (and other) contracts for lack of an "[SPDX-License-Identfier](https://solidity.readthedocs.io/en/v0.6.8/layout-of-source-files.html#spdx-license-identifier)" code comment at the top of the file. 

A proposed fix would be to add
`//SPDX-License-Identifier: MIT`
to the start of every contract file.

wbt:
Any chance of putting out a patch release with this in it?

nventuro:
That's a great suggestion @wbt, thanks! I just published [v3.0.2](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/CHANGELOG.md#302-2020-06-08) with that change, you should be able to install it from npm now and not get any warnings.

PR #2235, opened 2020-05-14
Add SPX license identifier

nventuro:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2234

Not having this identifier emits a warning on Solidity ^0.6.8.


nventuro:
I don't think so. I was just mimicking how license text is ussually displayed in source code.

Issue #2206, opened 2020-04-20
EIP-2612: Add ERC20 permit() function


nventuro:
There's been some discussion about an ERC20 extension with the `permit` function, popularized by MakerDAO's Dai token (see the code [here](https://github.com/makerdao/dss/blob/44330065999621834b08de1edf3b962f6bbd74c6/src/dai.sol#L118-L140)). 

This recently became even more relevant in light of hacks revolving ERC777 and reentrancy, since `permit` could help solve some of the UX issues that drive people to using ERC777 improperly. It looks like [a proper ERC might be in the works](https://twitter.com/MartinLundfall/status/1251880329264201730): we should make sure we're compliant if that ends up happening. In all likelihood the spec will be identical to the Dai code, however.

[Link to EIP-2612](https://eips.ethereum.org/EIPS/eip-2612)

miohtama:
I agree that `permit()` needs its own ERC.

However, the ERC-777 solves UX issues better than `permit()`. `permit()` requires all wallets to have a special flow to sign + send. In ERC-777 we can just `send()` tokens to the smart contract. ERC-777 t is semantically more correct and also from the user perspective much easier to understand. 

I think it is a separate discussion if people are using ERC-777 improperly or not, and maybe we should not go in there on Github.

miohtama:
For the underlying issue of re-entrance:

* Have Solidity to prevent re-entrance by defaut (like we changed with payable in the past) 

* Have EVM level change that prevents re-entrance by default

Currently the reason why these hacks happen are "bad defaults". Re-entrance should be opt-in, not opt-out.

nventuro:
Yes, I wasn't suggesting ERC777 should be removed or avoided (on the contrary, I think it opens the door to many interesting possibilities), but it does require a deeper understanding of the EVM and the Ethereum ecosystem, partly due to the 'bad defaults' you've mentioned. 

In that sense, `permit` would let people get a better UX than plain ERC20 offers without having to worry about the dangers of misusing it.

abcoathup:
EIP: https://github.com/ethereum/EIPs/issues/2613

mds1:
> In all likelihood the spec will be identical to the Dai code, however.

I don't think this is necessarily true. There's two potential improvements I've seen discussed, which I'm sure you're aware of, but figured I'd mention anyway:

[...]

Issue #1810, opened 2019-06-26
EIP191: Signature validation in solidity


jMyles:
Hey friends!  For the past few months, we at @NuCypher have been finding all the right places to insert effort in order to achieve basic signing and verifying of arbitrary payloads across wallets and dapps.  Last year, we sponsored a gitcoin bounty in order to close EIPs 191 and 712, which each deal with signing and verifying using wallet keys.

Now that [EIP191](https://github.com/ethereum/EIPs/issues/191) has landed, we are looking for the best place to put the verification logic in solidity.

**🧐 Motivation**

On-chain verification of EIP191 signatures is clearly an essential piece of this puzzle, and we are trying to head off any new boilerplate solidity being sprayed all over every dapp codebase in the world.

**📝 Details**

My colleagues @szotov and @cygnusv have already done most (perhaps all) of the legwork, here:

https://github.com/nucypher/nucypher/blame/6cdfcf033270371660e4cb5b1dcebd169574e31c/nucypher/blockchain/eth/sol/source/contracts/lib/SignatureVerifier.sol

(Note: This PR is subject to a subsequent rebase and thus might not be available at that URL.  In that event, you can see it at this gist: https://gist.github.com/jMyles/9526dee62674718ab97b93793ac45bfa)

So what do you think?  Is OpenZeppelin the right place for the canonical version of this logic?

cygnusv:
Correct link to `SignatureVerifier` with EIP191 logic:
https://github.com/nucypher/nucypher/blame/4ea10a651ddf5452cd187d3070e5259133aea4ff/nucypher/blockchain/eth/sol/source/contracts/lib/SignatureVerifier.sol

nventuro:
Hey there Justin, great talking to you again!

EIP191 definitely sounds like something we'd be interested in: we currently have a somewhat barebones [`ECDSA`](https://docs.openzeppelin.org/v2.3.0/api/cryptography#ecdsa) library to handle address recovery, but while working on projects with different signing needs, we've also come to the conclusion that some standard signing specification is required. Thank you for suggesting this!

Regarding the EIP itself, you mentioned it 'has landed', but the EIP [is listed as a draft](https://eips.ethereum.org/EIPS/eip-191), and I didn't find any activity related to it since April 2018 - is this happening somewhere else that I'm not seeing?

[...]

PR #2237, opened 2020-05-15
Add ERC20 Permit (EIP-2612)

nventuro:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2206

This adds support for the ERC20 Permit extension, tentatively called ERC2612 and based on the discussion in https://github.com/ethereum/EIPs/pull/2612. The EIP is still in draft status, ~we'll need to monitor it for any changes, and will likely not want to publish this code until it is more mature / finalized~ so it will go in the drafts directory.

This implementation accounts for `ChainID` changing (i.e. a hard fork) while at the same time being as gas efficient as possible.

The code and documentation is ready, but the tests are not,

#### To do

- [x] Add tests
- [x] Move to drafts directory
- [x] Rebase once #2418 is merged
- [x] Review contract name: `ERC20Permit` vs `ERC2612Permit` or `ERC2612`


frangio:
Previous discussion about EIP712 in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1810.

nventuro:
The Uniswap tests might be a good starting place to write the tests for this. We can also look into using [`eth_signTypedData`](https://github.com/trufflesuite/ganache-core/pull/149).

contracts/token/ERC20/ERC20Permit.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    function _updateDomainSeparator() private returns (bytes32) {
+        uint256 chainID = _chainID();
+
+        bytes32 newDomainSeparator = keccak256(
+            abi.encode(
+                keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
+                keccak256(bytes(name())),
+                keccak256(bytes("1")), // Version
```
k06a, line 81: It seems we should use pure internal method `version()` or `permitVersion()` for this.

contracts/token/ERC20/ERC20Permit.sol
```diff
@@ -0,0 +1,109 @@
...
+    function nonces(address owner) public view override returns (uint256) {
+        return _nonces[owner].current();
+    }
+
+    function _updateDomainSeparator() private returns (bytes32) {
+        uint256 chainID = _chainID();
+
+        bytes32 newDomainSeparator = keccak256(
+            abi.encode(
+                keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
```
rstormsf, line 79: this could a hardcoded constant

contracts/token/ERC20/ERC20Permit.sol
nventuro: Are you suggesting an internal setter for users to set their own version field? I'm actually not sure what all the use cases for `version` are, if you have insights on this that'd be extremely helpful.

That said we do need to provide a way for either the version or domain separator to be queried, as noted [here](version).

contracts/token/ERC20/ERC20Permit.sol
nventuro: Good catch, yes. Intuitively I'd expect this `immutable` would reduce deployment costs (because we need to store less data), though I'd like to measure that first.

contracts/token/ERC20/ERC20Permit.sol
k06a: I believe version makes sense when behaviour changes without address change (upgradablity?). So I mean version method could be overrided by user, or maybe should be even abstract.

nventuro: I see. It'd be great to have some examples to show in the docs.

Regarding implementation, we could have a setter and a getter, where the setter also calls `_updateDomainSeparator` and not increase gas costs of `permit` that way.

frangio:
While we still have to work on the tests for this feature, it's kind of on hold because the ERC is in Draft status.

dmihal:
Stumbled upon this while working on my [eth-permit](https://www.npmjs.com/package/eth-permit) library

Might be useful for your unit tests

frangio:
@dmihal Looks like it will be useful, thank you!

gorgos:
With a [few modifications](https://github.com/soliditylabs/ERC20-Permit/blob/main/patches/eth-permit%2B0.1.7.patch), I successfully used the `eth-permit` library to run some [basic tests](https://github.com/soliditylabs/ERC20-Permit/blob/main/test/ERC20Permit.test.js).

dmihal:
@gorgos the new version of eth-permit should work correctly.

And feel free to add PRs to https://github.com/dmihal/eth-permit 🙂

frangio:
Is anyone familiar with the finalization status of this EIP?

contracts/token/ERC20/ERC20Permit.sol
```diff
@@ -0,0 +1,109 @@
...
+    using Counters for Counters.Counter;
+
+    mapping (address => Counters.Counter) private _nonces;
+
+    bytes32 private immutable _PERMIT_TYPEHASH = keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
+
+    // Mapping of ChainID to domain separators. This is a very gas efficient way
+    // to not recalculate the domain separator on every call, while still
+    // automatically detecting ChainID changes.
+    mapping (uint256 => bytes32) private _domainSeparators;
```
bingen, line 26: What’s the gas benefit? I don’t see it. I have done some tests on Remix and it seemed to me that it was slightly more expensive. Maybe I’m missing something.
This is what I tried on Remix, and the execution cost of the first one was 1156 while the second one was 810. Besides the SSTORE used in the constructor.

```
pragma solidity 0.6.11;


contract DomainSeparator1 {
    string constant internal NAME = "LUSD Stablecoin";
    string constant internal VERSION = "1";

    mapping (uint256 => bytes32) private _domainSeparators;

    constructor() public {
        _updateDomainSeparator();        
    }

    function domainSeparator() external returns (bytes32) {
        bytes32 domainSeparator_ = _domainSeparators[_chainID()];
        if (domainSeparator_ != 0x00) {
            return domainSeparator_;
        } else {
            return _updateDomainSeparator();
        }
    }

    function _updateDomainSeparator() private returns (bytes32 domainSeparator_) {
        uint256 chainID = _chainID();
        domainSeparator_ = keccak256(abi.encode( 
            keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)'),
            keccak256(bytes(NAME)), keccak256(bytes(VERSION)), chainID, address(this)));
            
        _domainSeparators[chainID] = domainSeparator_;
    }

    function _chainID() private pure returns (uint256 chainID) {
        assembly {
            chainID := chainid()
        }
    }
}


contract DomainSeparator2 {
    string constant internal NAME = "LUSD Stablecoin";
    string constant internal VERSION = "1";

    mapping (uint256 => bytes32) private _domainSeparators;
    
    function domainSeparator() external view returns (bytes32) {
        uint256 chainID = _chainID();
        return keccak256(abi.encode( 
            keccak256('EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)'),
            keccak256(bytes(NAME)), keccak256(bytes(VERSION)), chainID, address(this)));
        
    }

    function _chainID() private pure returns (uint256 chainID) {
        assembly {
            chainID := chainid()
        }
    }
}
```

frangio: @bingen Thanks for bringing this up. It does seem that the version that doesn't use storage is cheaper.

@nventuro Has something change in the costs of storage that this is now cheaper?

rstormsf: I think you can remove `_domainSeparators ` in `DomainSeparator2`

frangio: I've done some tests and concluded that the storage based approach in this PR is indeed inferior, and I've arrived at a better solution using a combination of `immutable` and `keccak` as fallback for when chain id has changed.

https://github.com/frangio/domain-separator-gas-report/blob/311306300546de7a021b856fb2ff1723d5833a34/contracts/Test.sol#L86-L92

bingen: > I think you can remove _domainSeparators in DomainSeparator2

Yes, indeed, thanks, it was just a quick test in Remix, but I don’t think it’s relevant for the gas comparison.

contracts/token/ERC20/ERC20Permit.sol
frangio: @k06a In this case though the semantics of the signature are defined by the EIP, so I don't see a reason for `version` to change.

frangio:
@spalladino @mverzilli The code is ready for review. I still need to review the documentation a little bit. If you have any opinions on that front let me know.

test/drafts/ERC20Permit.test.js
```diff
@@ -0,0 +1,100 @@
...
+      const signature = ethSigUtil.signTypedMessage(wallet.getPrivateKey(), { data });
+      const { v, r, s } = fromRpcSig(signature);
+
+      const receipt = await this.token.permit(owner, spender, value, deadline, v, r, s);
+
+      expect(await this.token.nonces(owner)).to.be.bignumber.equal('1');
+      expect(await this.token.allowance(owner, spender)).to.be.bignumber.equal(value);
+    });
+
+    it('rejects other signature', async function () {
```
spalladino, line 79: I'd add another failing test that tries sending the same signature twice

test/drafts/ERC20Permit.test.js
```diff
@@ -0,0 +1,100 @@
...
+        this.token.permit(owner, spender, value, deadline, v, r, s),
+        'ERC20Permit: invalid signature',
+      );
+    });
+
+    it('rejects expired permit', async function () {
+      const deadline = (await time.latest()) - time.duration.weeks(1);
+
+      await expectRevert(
+        this.token.permit(owner, spender, value, deadline, 0, ZERO_BYTES32, ZERO_BYTES32),
```
spalladino, line 95: I know it's not needed because the deadline check is performed first, but I'd still include the signature here instead of zeros.

spalladino (approved):
LGTM!

contracts/drafts/IERC2612Permit.sol
```diff
@@ -0,0 +1,53 @@
...
+     * https://eips.ethereum.org/EIPS/eip-2612#specification[relevant EIP
+     * section].
+     */
+    function permit(address owner, address spender, uint256 amount, uint256 deadline, uint8 v, bytes32 r, bytes32 s) external;
+
+    /**
+     * @dev Returns the current ERC2612 nonce for `owner`. This value must be
+     * included whenever a signature is generated for {permit}.
+     *
+     * Every successful call to {permit} increases ``owner``'s nonce by one. This
```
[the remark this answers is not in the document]
frangio: This is intentional. The single backticks only work on word boundaries (however that is defined) and the apostrophe breaks that, double backticks don't require word boundaries. (See difference between [constrained and unconstrained](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#formatted-text) formatting.)

contracts/drafts/IERC2612Permit.sol
```diff
@@ -0,0 +1,53 @@
...
+     * Emits an {Approval} event.
+     *
+     * Requirements:
+     *
+     * - `owner` cannot be the zero address.
+     * - `spender` cannot be the zero address.
+     * - `deadline` must be a timestamp in the future.
+     * - `v`, `r` and `s` must be a valid `secp256k1` signature from `owner`
+     * over the EIP712-formatted function arguments.
+     * - the signature must use ``owner``'s current nonce (see {nonces}).
```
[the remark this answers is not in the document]
frangio: Yes, it's intentional. The single backticks only work on word boundaries (however that is defined) and the apostrophe breaks that, double backticks don't require word boundaries. (See difference between [constrained and unconstrained](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#formatted-text) formatting.)

test/drafts/ERC20Permit.test.js
frangio: It's also not needed because we check that the revert reason corresponds to the deadline error. Do you still think the signature should be valid?

spalladino: True. But if it's not too much effort, I'd still do it, so it decouples the test from the internals of the implementation. Otherwise, just by swapping the order of the checks in the implementation this test breaks. I know, I'm looking for _the hair of the egg_ (what's a good English translation for that one?), so feel free to merge without it!

contracts/drafts/IERC2612Permit.sol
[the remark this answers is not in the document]
spalladino: Wow, I always learn something new

mverzilli (approved):

PR #2244, opened 2020-05-24
Decrease CREATE2 address computation gas usage from 623 to 539

k06a:
Small fix

### DIFF
--- a/contracts/utils/Create2.sol
+++ b/contracts/utils/Create2.sol
@@ -54,6 +54,6 @@ library Create2 {
         bytes32 _data = keccak256(
             abi.encodePacked(bytes1(0xff), deployer, salt, bytecodeHash)
         );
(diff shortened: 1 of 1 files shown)


frangio:
Does the `address` cast to the necessary masking?

k06a:
@frangio, yep, sure.

nventuro:
I run a small test to confirm this:

```solidity
contract TestA {
    function foo(uint256 a) public pure returns (uint256) {
        address b = address(a);
        return uint256(b);
    }
}
```

When called with `115792089237316195423570985008687907853269984665640564039457584007913129639935` (2^255 - 1), it returns `1461501637330902918203684832716283019655932542975` (2^160 - 1).

k06a:
@frangio 👌

PR #2246, opened 2020-05-25
Update SafeERC20.sol

Ro5s:
tweak intro dev comment to reflect more typical use case, wrapping an `IERC20` reference for a deposit contract using others' ERC-20 tokens. 

I might be mistaken ~ but typically SafeERC20 won't be used in conjunction with ERC20 for OZ token deployments?


frangio:
Thank you!

PR #2257, opened 2020-06-01
Modify token presets "mint" function as virtual to allow extending

EvilJordan:
1. Discussion (one-sided at the time of this request) of this PR can be found here: https://forum.openzeppelin.com/t/erc721presetminterpauserautoid-mint-function-and-tokenid/2780/8

2. The preset `mint` function is not `override`-able by parent contracts as it is not marked as `virtual`. Since `mint` is not part of the ERC721 or ERC20 standard, and the ability to user `super` and OZ's outstanding presets exists, why not allow for more flexibility?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Improvements
  * `ReentrancyGuard`: reduced overhead of using the `nonReentrant` modifier. ([#2171](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2171))
  * `AccessControl`: added a `RoleAdminChanged` event to `_setAdminRole`. ([#2214](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2214))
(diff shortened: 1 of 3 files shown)


frangio:
This must have been an oversight. I believe all functions in these contracts should be virtual, [like the rest of the library](https://forum.openzeppelin.com/t/how-we-plan-to-adopt-solidity-0-6-virtual-functions/2063). @nventuro can you confirm this?

nventuro:
Yes, indeed. I did a quick search and it seems no other contracts have their functions missing the `virtual` attribute, though I did add it to some more functions in the presets.

I'd like to mention however that part of the intent behind the presets is to use them without having to modify them. If you find that you need to customize their behavior, it might be a better idea to use the preset as a basis to write your own custom code that interfaces directly with the base contracts (`ERC721`, `ERC721Pausable`, etc.).

EvilJordan:
> I'd like to mention however that part of the intent behind the presets is to use them without having to modify them.

Totally, and that's what I've done in my implementation, but had thought it strange this wasn't even an _option_. Thanks for the attention, and great work!!

PR #2261, opened 2020-06-01
Update Test Helpers to 0.5.6

frangio:


test/token/ERC777/ERC777.test.js
```diff
@@ -60,8 +60,8 @@ describe('ERC777', function () {
       });
     });
 
-    it.skip('does not emit AuthorizedOperator events for default operators', async function () {
-      expectEvent.not.inConstructor(this.token, 'AuthorizedOperator'); // This helper needs to be implemented
+    it('does not emit AuthorizedOperator events for default operators', async function () {
+      expectEvent.notEmitted.inConstruction(this.token, 'AuthorizedOperator');
```
nventuro, line 64:
```suggestion
      await expectEvent.notEmitted.inConstruction(this.token, 'AuthorizedOperator');
```

According to [the docs](https://docs.openzeppelin.com/test-helpers/0.5/api#inconstruction) this function returns a promise (because it needs to fetch a tx receipt from the deployment tx hash).

PR #2264, opened 2020-06-04
feat: add wrapper function for low level calls

### LINKED ISSUE #2260: Add a .call wrapper with Solidity function call checks
Some contracts need to perform function calls that can't be expressed as Solidity functions and so must resort to low level `.call`. In most of these cases Solidity function call semantics are still desired but it's tricky to get all the details right. There's a couple of instances of this in this library: [`ERC721._checkOnERC721Received`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/394987f36527f16f85efdfbe9c944da629077917/contracts/token/ERC721/ERC721.sol#L501) and [`SafeERC20`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/394987f36527f16f85efdfbe9c944da629077917/contracts/token/ERC20/SafeERC20.sol#L69).

We should provide a function that implements these details, possibly in the `Address` library. This function has to:

1. Verify the target address is a contract. (`Address.isContract`)
2. Use `.call` with a specified `bytes data`, and require the call to succeed, otherwise bubbling up the revert reason or using a default revert reason. (See the implementation of this in `ERC721`.)
3. Return the returndata.

This function should then be used in the library in the places I linked above.

We have to come up with a name. `Address.checkedCall` comes to mind, but I'd like to see other options.

julianmrodri:
#2260 

Fixes #2260 

This PR creates a function `Address.functionCall` which can be used to wrap low level calls that need to be performed on other Smart contracts. .

It also modifies the `ERC721.sol` and `SafeERC20.sol`contracts to use this new function instead of performing directly the low level call. Tests for these contracts needed to be updated to match the new revert messages in certain situations.


contracts/utils/Address.sol
```diff
@@ -57,4 +57,31 @@ library Address {
...
+        (bool success, bytes memory returndata) = target.call(data);
+        if (!success) {
+            if (returndata.length > 0) {
+                // solhint-disable-next-line no-inline-assembly
+                assembly {
+                    let returndata_size := mload(returndata)
+                    revert(add(32, returndata), returndata_size)
+                }
+            } else {
+                revert("Address: low-level call failed");
```
frangio, line 81: What we can do is receive this string as an argument to `functionCall`, and have an override without the third argument that uses this default reason.

frangio:
This is looking great!

contracts/utils/Address.sol
nventuro: Yes, this is what I was thinking. This would be the same approach as in [SafeMath.sub](https://docs.openzeppelin.com/contracts/3.x/api/math#SafeMath-sub-uint256-uint256-string-).

julianmrodri: Great. I'll implement that.

julianmrodri:
Thanks!  I think I might need some guidance on the Tests to include. Which kind of `calls` should I mock?  Do you have any preference on what contract to use and what samples to use for the Unit tests?

nventuro:
We might want to also add a `value` field for ETH transfers.

>Thanks! I think I might need some guidance on the Tests to include. 

Off the top of my head, interesting scenarios are the following (thought we might not need to test them all):

- call to a contract that implements the called function and returns some data
- call to a contract that implements the called function but reverts
- call to a contract that implements the called function but runs out of gas
- call to a contract that doesn't implement the function (and reverts)
- call to a non-contract

If we add a `value` field, we might also want to test functions that are not `payable`.

julianmrodri:
@nventuro @frangio I've added the error message parameter so now we have the two options for `functionCall` (with or without the custom message). I also added some of the tests. 

I would like you to review the TESTS to see if this is the correct direction. I had to make too many decisions and is better for you to validate if this is OK. Feel free to let me know if this is not the correct way of doing this!

So far I covered:

* Call to a contract that implements the called function successfully
* Call to a contract that implements the called function but reverts
* Call to a contract that doesn't implement the function (and reverts)
* Call to a non-contract

I was **not** able to test the case of a function that returns data with this approach. It is quite challenging as I get the **Transaction** info and not the **returned value**, to eventually decode. Do you know how this can be achieved? Thanks!

contracts/utils/Address.sol
```diff
@@ -57,4 +57,42 @@ library Address {
...
+     * provided by the low-level call. Uses `errorMessage` as default revert message.
+     
+     * The call is not executed if the target address is not a contract.
+     */
+    function functionCall(address target, bytes memory data, string memory errorMessage) internal returns (bytes memory) {
+        require(isContract(target), "Address: call to non-contract");
+
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory returndata) = target.call(data);
+        if (!success) {
```
nventuro, line 84: Can we have this be `if (success)` and swap the if and else blocks? I find the code more readable that way.

test/token/ERC20/SafeERC20.test.js
```diff
@@ -15,7 +15,7 @@ describe('SafeERC20', function () {
       this.wrapper = await SafeERC20Wrapper.new(hasNoCode);
     });
 
-    shouldRevertOnAllCalls('SafeERC20: call to non-contract');
+    shouldRevertOnAllCalls('Address: call to non-contract');
```
nventuro, line 18: @frangio are we ok with changing revert reasons in non-major releases?

(I think we are)

contracts/mocks/AddressImpl.sol
```diff
@@ -13,6 +13,10 @@ contract AddressImpl {
         Address.sendValue(receiver, amount);
     }
 
+    function functionCall(address target, bytes calldata data) external returns (bytes memory) {
+        return Address.functionCall(target, data);
```
nventuro, line 17: Because this function is not `view`, accessing the returned data is not possible. The easiest way to test this is to emit an event including it. See for example the tests for `EnumerableSet`.

nventuro: @julianmrodri this is what you were asking about

contracts/utils/Address.sol
julianmrodri: Yes I dont see why not. I took that from the original implementation but we can swap.

contracts/mocks/AddressImpl.sol
julianmrodri: Thanks @nventuro will look into this. The rest of the tests are kind of OK-ish?

test/token/ERC20/SafeERC20.test.js
frangio: Hm not possible to do that kind of thing dynamically. I don't think revert reasons should be stable, it's too limiting for us. I'm not 100% sure, but I think it should be ok to change it.

contracts/utils/Address.sol
```diff
@@ -57,4 +57,42 @@ library Address {
...
+    function functionCall(address target, bytes memory data) internal returns (bytes memory) {
+      return functionCall(target, data, "Address: low-level call failed");
+    }
+
+    /**
+     * @dev Replacement for Solidity's low-level `call`: performs a low-level
+     * call with `data` to the target address `target`. Returns the `returndata`
+     * provided by the low-level call. Uses `errorMessage` as default revert message.
+     
+     * The call is not executed if the target address is not a contract.
```
frangio, line 77: The calls should also mention that the function will bubble up the revert if the function call reverts.

contracts/utils/Address.sol
```diff
@@ -57,4 +57,42 @@ library Address {
         (bool success, ) = recipient.call{ value: amount }("");
         require(success, "Address: unable to send value, recipient may have reverted");
     }
+
+    /**
+     * @dev Replacement for Solidity's low-level `call`: performs a low-level
```
frangio, line 62: Instead of repeating the same docs here, we would put a link to the other one ("See `functionCall`") and only explain what's different about this one (in this case the default revert reason).

I'm not sure how we can automatically generate a link to a function with an overload like this one. @nventuro do you remember?

test/token/ERC20/SafeERC20.test.js
nventuro: If we _really_ wanted to keep the old message, we could have an extra `Address.isContract()` check on `SafeERC20` and have that fail first, but I agree with @frangio here and we can leave as-is.

contracts/utils/Address.sol
nventuro: What I do is just run the docsite manually and copy-paste the identifier. I can add this myself on top of this PR later on.

julianmrodri: Yes. I repeated the docs because I saw it was done a similar approach in the [div](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/math/SafeMath.sol#L91-L125) example. But I agree better not to repeat.

nventuro: Ah that's ineteresting, thanks. We should improve those.

nventuro:
By the way I [asked the Solidity dev team on gitter](https://gitter.im/ethereum/solidity-dev?at=5edfb465013105125a245b86) if they think our checks are correct, and they gave us green light:

>>>
nventuro
hello! I am trying to replicate Solidity function call semantics using call, and wanted to double-check I'm not missing anything. This is what I'm currently doing:

    check using extcodehash that the recipient account has code, revert otherwise
    perform the call
    if success, abi.decode the returndata
    if failure, check the returndata for a revert reason (lenght > 0). if available, revert with that, otherwise revert with a generic reason

Is that all?

chriseth
@nventuro looks good!

nventuro:
@julianmrodri no need to apologize! Your help is greatly appreciated :)

nventuro:
For some reason the links are not working (see https://deploy-preview-2264--openzeppelin-contracts-docs.netlify.app/contracts/3.x/api/utils#Address-functionCall-address-bytes-string-).

julianmrodri:
Wow you took this to the next level. Awesome!

contracts/utils/Address.sol
```diff
@@ -59,30 +59,69 @@ library Address {
...
     }
 
     /**
-     * @dev Replacement for Solidity's low-level `call`: performs a low-level
-     * call with `data` to the target address `target`. Returns the `returndata`
-     * provided by the low-level call. Uses `errorMessage` as default revert message.
-     
-     * The call is not executed if the target address is not a contract.
+     * @dev Same as {Address-functionCall-address-bytes-}, but with
+     * `errorMessage` as a custom revert reason when `target` reverts.
```
frangio, line 80: I think this conveys the idea a bit better.

```suggestion
     * `errorMessage` as a fallback revert reason when `target` reverts.
```

contracts/utils/Address.sol
```diff
@@ -59,30 +59,69 @@ library Address {
...
+     * - the calling contract must have an ETH balance of at least `value`.
+     * - calling `target` with `data` must not revert.
+     */
+    function functionCallWithValue(address target, bytes memory data, uint256 value) internal returns (bytes memory) {
+        return functionCallWithValue(target, data, value, "Address: low-level call with value failed");
+    }
+
+    /**
+     * @dev Same as {Address-functionCallWithValue-address-bytes-uint256-}, but
+     * with `errorMessage` as a custom revert reason when `target` reverts.
```
frangio, line 106:
```suggestion
     * with `errorMessage` as a fallback revert reason when `target` reverts.
```

contracts/utils/Address.sol
```diff
@@ -59,30 +59,69 @@ library Address {
...
+     */
+    function functionCallWithValue(address target, bytes memory data, uint256 value, string memory errorMessage) internal returns (bytes memory) {
+        require(address(this).balance >= value, "Address: insufficient balance for call");
```
frangio, line 109: This is interesting, does Solidity do this check too?

nventuro: No, the `call` reverts, so `success` is false, and you get the generic `low level call failed` revert reason. I added this check since that balance check is extremely cheap: https://eips.ethereum.org/EIPS/eip-1884#specification

test/utils/Address.test.js
```diff
@@ -145,4 +189,86 @@ describe('Address', function () {
...
+          inputs: [],
+        }, []);
+
+        await expectRevert(
+          this.mock.functionCallWithValue(this.contractRecipient.address, abiEncodedCall, amount),
+          'Address: insufficient balance for call'
+        );
+      });
+
+      it('calls the requested function with existing value', async function () {
```
frangio, line 229: We're not asserting that the target contract received the requested value.

contracts/utils/Address.sol
frangio: I wouldn't call that "extremely cheap"!! In fact, it's the same as the base cost of `CALL`, which should the dominating cost of `functionCall`. Adding this check increases that 2x. So I have a strong preference towards removing it.

frangio: Hm, okay when transferring value the cost of a call increases by 9000 gas, so it would not be a 2x increase, but I'd still call it significant.

In any case, I think we should remove the check so as to be closer to Solidity semantics.

nventuro: `balance` costs 700 gas, but `selfbalance` (what that expression compiles into) costs 5 gas. That's what I was linking, `selfbalance` was introduced in that EIP. Sorry if this caused confusion.

test/utils/Address.test.js
nventuro: Great catch, thanks!

contracts/utils/Address.sol
frangio: Ah, my bad!

test/utils/Address.test.js
```diff
@@ -233,9 +233,13 @@ describe('Address', function () {
           inputs: [],
         }, []);
 
+        const tracker = await balance.tracker(this.contractRecipient.address);
```
frangio, line 236: :heart_eyes:

julianmrodri:
Congrats guys! Awesome job. Glad to be able to be part of this. 👍

nventuro:
Thanks a lot to you both!

PR #2266, opened 2020-06-05
fix wrong version

x5engine:
Fix because this break other contracts like AccessControl when compiling

image


abcoathup:
Hi @x5engine,

Address.sol uses "0.6.2" due to using notation only introduced in 0.6.2 as the previous notation was deprecated in 0.6.4.

See the following in the forum for details:
https://forum.openzeppelin.com/t/why-does-address-sol-in-openzeppelin-contracts-use-pragma-0-6-2/3016

I am closing this PR for this reason.

x5engine:
Thank you Andrew, I totally got it.

PR #2275, opened 2020-06-11
Fix docs links to overloaded functions

frangio:


frangio:
Seems to be fixed.

nventuro:
Looks like the table of contents is now correct, but the anchors are still not available to be linked to (see https://5ee28577a907ed0007830f9a--openzeppelin-contracts-docs.netlify.app/contracts/3.x/api/utils#Address-functionCall-address-bytes-string-).

frangio:
See my last commit. You need to do something like that.

PR #2289, opened 2020-06-22
added transfer event

GeorgeGoodall:
the contract had called the transfer event that didn't exist. causing a "Returned values aren't valid, did it run Out of Gas?" error


GeorgeGoodall:
> Why add 'Transfer' event here if it is already in the IERC721?

My mistake, I have closed the pr

Issue #2189, opened 2020-04-15
Upgrade solidity-coverage to ^0.7.2 to support Solidity v0.6


frangio:

nventuro:
Ideally this will be handled by a new version of [OpenZeppelin Test Environment](docs.openzeppelin.com/test-environment/), which has added this feature to its [Q2 2020 roadmap](https://github.com/OpenZeppelin/openzeppelin-test-environment/issues/73).

PR #2291, opened 2020-06-25
Re-enable coverage analysis

ylv-io:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2189

Adds support for `solidity-coverage` using Test Environment library 🍾


scripts/coverage.sh
```diff
@@ -1,25 +0,0 @@
...
-
-# The allFiredEvents file is created inside coverageEnv, but solidity-coverage
-# expects it to be at the top level. We create a symlink to fix this
-ln -s coverageEnv/allFiredEvents allFiredEvents
-
-OZ_TEST_ENV_COVERAGE=true npx solidity-coverage || log "Test run failed"
-
-if [ "$CI" = true ]; then
-  curl -s https://codecov.io/bash | bash -s -- -C "$CIRCLE_SHA1"
-fi
```
frangio, line 25: We have to add this snippet in `coverage.js` to get coverage reports in PRs.

test/utils/Address.test.js
```diff
@@ -138,6 +140,7 @@ describe('Address', function () {
       });
 
       it('reverts when the called function runs out of gas', async function () {
+        if (coverage) { return this.skip(); }
```
frangio, line 143: What's the reason we have to do this? Can you add a comment explaining that here?

nventuro: This test attempts to make the execution run out of gas, but the coverage EVM sets a ridiculously high gas limit. In practice, this ends up timeouting and causing all other tests to fail due to the hang in Ganache.

ylv-io: It is due to coverage setting block gas limit to `0xffffffffff` which cause `mockFunctionOutOfGas` function to crash Ganache and the subsequent tests before running out of gas.
```
  function mockFunctionOutOfGas() public payable {
        for (uint256 i = 0; ; ++i) {
            _array.push(i);
        }
    }
```

scripts/coverage.js
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env node
 
+const { execSync } = require('child_process');
 const { runCoverage } = require('@openzeppelin/test-environment');
 
-runCoverage(['mocks'], 'npm run compile', './node_modules/.bin/mocha --exit --timeout 10000 --recursive');
+runCoverage(['mocks'], 'npm run compile', './node_modules/.bin/mocha --exit --timeout 10000 --recursive'.split(' '));
+
+if (process.env.CI) {
+  execSync('curl -s https://codecov.io/bash | bash -s -- -C "$CIRCLE_SHA1"');
```
frangio, line 9: This is so we see the logs in CI.

```suggestion
  execSync('curl -s https://codecov.io/bash | bash -s -- -C "$CIRCLE_SHA1"', { stdio: 'inherit' });
```

nventuro:
We're getting a strange error on the CI run:

```
TypeError: testCommand.split is not a function
 
    at runCoverage (/home/circleci/project/node_modules/@openzeppelin/test-environment/src/coverage.ts:56:37)
 
    at Object.<anonymous> (/home/circleci/project/scripts/coverage.js:6:1)
```

I see no call to `split` [on line 56](https://github.com/OpenZeppelin/openzeppelin-test-environment/blob/feature/solidity-coverage/src/coverage.ts#L56). I'm not sure what's up with that.

ylv-io:
It is because it can't realize that `test-env` has changed. Once @frangio approves `test-env` PR we'll have a proper dependency and it will work.

PR #2292, opened 2020-06-26
Fixes broken link in ERC20 comment.

MicahZoltu:
Hard newlines are the devil, which was part of the problem, but even if I removed the hard newline from the middle of the link it still didn't work.  Removing the `[...]` from the end of the link seems to have resolved the issue, which is what this commit does.

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -15,8 +15,7 @@ import "../../utils/Address.sol";
  * For a generic mechanism see {ERC20PresetMinterPauser}.
  *
  * TIP: For a detailed writeup see our guide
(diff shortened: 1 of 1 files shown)


frangio:
Thanks for the PR @MicahZoltu!

I see the link as working [on the site](https://docs.openzeppelin.com/contracts/3.x/api/token/erc20#ERC20). Where do you see it broken?

nventuro:
Ah, perhaps that tripped you. This is an [asciidoc link](https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#links) (formatted as `link[display text]`), which is what we use to generate the documentation website. The source code comments are not meant to be read by people, but instead help generate the content of https://docs.openzeppelin.com/contracts

PR #2294, opened 2020-06-28
Update ReentrancyGuard.sol with minor gas cost saving

Ro5s:
set non-entered uint256 value to 0 rather than 1 (gas savings?)


nventuro:
Hello @Ro5s, thanks for contributing to OpenZeppelin Contracts!

Unfortunately, setting back to 0 will not change gas costs. Understanding how this works is not trivial though, and requires reading up on both [EIP2200](https://eips.ethereum.org/EIPS/eip-2200) and [EIP1283](https://eips.ethereum.org/EIPS/eip-1283) (which defines the terms used in 2200).

The only difference between setting back to zero or non-zero is the former will use more gas but have an equally large refund. For the reasons laid in the comment in the source right above your change, we want to keep this low to a) increase likelihood of the full refund being effective, and b) reduce upfront required gas costs. The current values perform better in this regard, so I'm closing the PR.

Thanks!

PR #2296, opened 2020-06-28
Add enumerate() back into EnumerableSet

### LINKED ISSUE #2166: Restore EnumerableSet.enumerate
This function was removed in #2160 due to issues with casting arrays of different types.

This can be achieved using assembly:

```solidity
bytes32[] x = ...;
uint[] ret;
assembly { ret := x }
```

However, creating an assembly block brings all assembly keywords into scope, which causes shadowing warnings to be emitted by the compiler due to the existence of functions with names such as `add` in the same scope.

We should devise a way to bring `enumerate` back without these warnings being triggered.

### LINKED ISSUE #2295: It's not an enumerable set if you can't enumerate it.
Realistically, all removing the enumeration does is push the loop to the implementation level and adds unnecessary internal calls.

thegostep:
Fixes #2166

### DIFF
--- a/contracts/utils/EnumerableSet.sol
+++ b/contracts/utils/EnumerableSet.sol
@@ -158,6 +158,21 @@ library EnumerableSet {
         return _remove(set._inner, bytes32(uint256(value)));
     }
 
+	/**
+	 * @dev Returns an array with all values in the set. O(N).
+	 * Note that there are no guarantees on the ordering of values inside the
+	 * array, and it may change when more values are added or removed.
+	 * WARNING: This function may run out of gas on large sets: use {length} and
+	 * {at} instead in these cases.
+	 */
+	function enumerate(AddressSet storage set) internal view returns (address[] memory) {
+		address[] memory output = new address[](set._inner._values.length);
+		for (uint256 i; i < set._inner._values.length; i++) {
+			output[i] = address(uint256(set._inner._values[i]));
+		}
+		return output;
+	}
+
     /**
      * @dev Returns true if the value is in the set. O(1).
      */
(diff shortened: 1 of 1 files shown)


nventuro:
Thanks @thegostep! This would indeed make such an external function easier to implement. However, I worry about unbounded array access: even if these are `view` functions, `eth_call` is still [gas-limited](https://infura.io/docs/ethereum/json-rpc/eth-call).

What do you think about doing pagination by providing a query size? It has the disadvantage that you need to query the set size beforehand, but I think this is one of the better ways to make this scale.

Using the high level getters the code might look something like:

```solidity
function enumerate(Set set, uint256 startIndex, uint256 pageSize) internal view returns (uint256[] memory) {
    uint256[] memory result;
    require(startIndex + pageSize <= set.length(), "Read index out of bounds");

    for (uint256 i = 0; i < pageSize; ++i) {
        result[i] = set.at(i + startIndex);
    }

    return result;
}
```

thegostep:
Why not all 3?

It's usually what I do in my projects:
https://github.com/erasureprotocol/erasure-protocol/blob/be41efc93cdc177b22f7b5a6dac0bb8d63aeb88c/contracts/modules/Registry.sol#L176-L223

thegostep:
Removing functionality because there is a chance of running out of gas is not a good approach IMO.

Realistically, most use cases will not reach that edge case and the ones that do have all they need to mitigate it.

frangio:
>the ones that do have all they need to mitigate it.

How is this so? A smart contract that uses a function that returns an array has a vulnerability that potentially can't be mitigated. We don't want to provide a primitive that could put users in that position. 

While you're free to include all 3 variants in your own contracts, because you know you won't misuse them, we think we have to be more careful and avoid providing interfaces that are easy to misuse.

thegostep:
ERC20 is easy to misuse 😂

frangio:
@nventuro Should we close this issue in favor of a pagination or indexing approach?

> It has the disadvantage that you need to query the set size beforehand, but I think this is one of the better ways to make this scale.

I don't think this is true by the way. It should be possible to query for pages until you get one that is smaller than the size you requested, then you know you've reached the end.

nventuro:
>Should we close this issue in favor of a pagination or indexing approach?

What do you mean by 'pagination or indexing'? Whether the index is the actual item index or the page number? Page number seems more sensible, and yes I'd go for that.

>I don't think this is true by the way. It should be possible to query for pages until you get one that is smaller than the size you requested, then you know you've reached the end.

Good point.

frangio:
Yes, I mean what you said.

PR #2300, opened 2020-06-29
Add readme notices with link to docs site

frangio:
This PR adds a notice at the top of each contract README with a link to the docs site. It should be hidden in the published site.

### DIFF
--- a/contracts/GSN/README.adoc
+++ b/contracts/GSN/README.adoc
@@ -1,7 +1,7 @@
 = Gas Station Network (GSN)
 
 [.readme-notice]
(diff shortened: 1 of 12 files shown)


nventuro:
Cool! What makes these not render on the site?

I'd consider dropping the period at the end of the line to make copy/pasting the link easier.

frangio:
Yeah that's a good idea. It's CSS `display: none;` introduced in https://github.com/OpenZeppelin/docs.openzeppelin.com/commit/26c720c6c9cef6b2cd4aa6e1544964024dd0d092.

PR #2305, opened 2020-07-05
Adding of a "onlyAdminOf" modifier into AccessControl 

frankthequeen:
Added "onlyAdminOf" modifier to merge duplicated check used in "grantRole" and "revokeRole" functions.
I used it into a personal project and I noticed it could be useful for further functions also.

Updated test to merge error messages.

### DIFF
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -89,6 +89,14 @@ abstract contract AccessControl is Context {
     }
 
     /**
+     * @dev Throws if called by any account without the ``role``'s adminRole.
+     */
+    modifier onlyAdminOf(bytes32 role) {
+        require(hasRole(_roles[role].adminRole, _msgSender()), "AccessControl: caller must be an admin of the role");
+        _;
+    }
+
+    /**
      * @dev Returns the number of accounts that have `role`. Can be used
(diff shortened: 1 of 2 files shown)


nventuro:
Hello @frankthequeen, thanks for contributing!

A small issue with using modifiers like this is that we'd lose the unique revert reasons, which we're fond of. For the library's internals, I don't mind this small duplication given this and the readability of the original version.

That said, we could add a modifier to the library if it makes using it simpler. Do you have a preference for modifiers over writing your own `require` statements?

frankthequeen:
Hi Nicolás,
I understand your concern about the unified revert message, feel free to reject my commit.

In fact I suggested to add a modifier because I used this contract customizing it a lot. Just for example, I added functions to manage DEFAULT_ADMIN_ROLE and I assigned them to "onlyOwner" then I created few utility functions to manage others roles and I assigned them to "onlyAdminOf(role)". In short, for the use I made of it, the new modifier was very useful to me. But I understand that mine may be a specific case.

frangio:
We're going to keep `AccessControl` without a modifier for now, but if more people feel that they would benefit from a modifier like this let us know and we will reconsider it.

PR #2311, opened 2020-07-13
feat: use extcodesize for isContract to reduce gas

### LINKED ISSUE #2310: Consider switching Address.isContract back to extcodesize
See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/1802#issuecomment-657668876 by @juli for more details.

julianmrodri:
#2310

Fixes #2310 

* Restablishes `extcodesize` to check if an address `isContract` as this now consumes less gas than `extcodehash`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.1.1 (Unreleased)
+
+### Improvements
+ * `Address.isContract`: switched from `extcodehash` to `extcodesize` for less gas usage. ([#2311](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2311))
+
 ## 3.1.0 (2020-06-23)
 
 ### New features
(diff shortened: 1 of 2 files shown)


julianmrodri:
@nventuro @frangio  just implemented the change so we have it there. I will run some verifications to confirm less gas is consumed with this method.

julianmrodri:
Execution costs (tested on Remix):
`extcodehash` 
1) If address is EOA:  1079 gas
2) If address is contract: 1099 gas

With `extcodesize`:
1) if address is EOA: 1015 gas
2) if address is contract: 1015 gas

So there is a reduction in the execution costs  by using `extcodesize`. Whether it iss significant I will leave it up to you guys, would really appreciate to understand your rationale.

frangio:
Thanks @julianmrodri! I wouldn't call the reduction significant but it's still a welcome change!

Can you add a small changelog entry?

julianmrodri:
Sure will do right now.

PR #2315, opened 2020-07-19
Remove redundant else and improve readability

MrClottom:
Small code change to improve readability. This PR does not change any functionality.

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -121,21 +121,18 @@ library Address {
 
         // solhint-disable-next-line avoid-low-level-calls
         (bool success, bytes memory returndata) = target.call{ value: weiValue }(data);
-        if (success) {
+        if (success)
             return returndata;
-        } else {
-            // Look for revert reason and bubble it up if present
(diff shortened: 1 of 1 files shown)


MrClottom:
The tests that fail are apparently failing because of a timeout, don't know how my change could cause that, does anyone know why this could be happening?

frangio:
Sorry for that timeout @MrClottom. It's a flakiness in our test suite that we haven't been able to track down yet.

Using early returns versus exlicit else branches is a matter of style, and we prefer the style that we're currently using, so I'm going to reject these changes. But thank you for the suggestion and the PR!

MrClottom:
What if the explicit `else` after the return were to be kept but the `if (condition) else revert` were replaced with a `require` as suggested?

frangio:
No, I'm sorry.

MrClottom:
@frangio Ok but why? Isn't that the whole point of the revert function? The inner if else just hurts readability, is there some additional functionality that I'm missing? I can open a separate issue with the only one change if it's preferred.

PR #2319, opened 2020-07-29
Migrate contracts to Solidity 0.7

### LINKED ISSUE #2318: Support for Solidity 0.7.0
**🧐 Motivation**
I know [solc 0.7.0 was only released yesterday](https://github.com/ethereum/solidity/releases/tag/v0.7.0) however logging it here for consideration.

**📝 Details**
Upgrade contracts to solidity 0.7.0 which overall doesn't have as many breaking changes as 0.6.0 so shouldn't be that invasive. Mostly it is maintenance updates and switching certain compiler warnings to errors.

elenadimitrova:
Fixes #2318 

All contracts were migrated to ^0.7.0. Changes include:

- Reference `Safemath` explicitly as `using A for B` syntax no longer affects derived contracts
- Remove `internal`/`public` declaration on constructors

`solhint` currently throws `Error: Constructors have to be declared either "public" or "internal"` probably due to unsupported solidity version.


frangio:
Thank you so much @elenadimitrova!

We still haven't figured out how we're going to fit support for Solidity 0.7 in our versioning scheme, since the pragma bump is a breaking change. We may want to continue providing support and new features for Solidity 0.6. So I'm merging this to a new branch while we figure that out, and it also serves to wait until the linter and [our docs generator](https://github.com/OpenZeppelin/solidity-docgen) are updated for 0.7.

PR #2325, opened 2020-08-09
Fixing link for "How to to implement supply mechanisms" in ERC20.sol [currently will open broken link in vs code]

moose-code:
When trying to open this link in VS code, it directs to a broken link as the [How part is included in the URL.

Simple fix, move it to the next line.

Fixes #


julianmrodri:
Nice catch, I was able to confirme the behavior. 
image

frangio:
Thank you @moose-code. Unfortunately this change breaks our docs site rendering. We have to prioritize the site before how it renders in an editor, especially because it's highly dependent on the editor used. If anyone has ideas on how to fix it for both scenarios let us know.

Issue #1493, opened 2018-11-09
ERC20.sol doesn't allow transfer to zero address, not standard compliant


c-o-l-o-r:
The [ERC-20 Token Standard](https://eips.ethereum.org/EIPS/eip-20) describes the following for `transfer`:
> `transfer`
> Transfers `_value` amount of tokens to address `_to`, and MUST fire the Transfer event. The function SHOULD throw if the `_from` account balance does not have enough tokens to spend.

The OpenZeppelin ERC20.sol contract adds an additional requirement to this standard by [blocking transfers](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/token/ERC20/ERC20.sol#L156) to `0x0000000000000000000000000000000000000000`. This opinionated decision should be removed from the base ERC20 implementation. The ERC20Burnable.sol contract is not a solution to this, because it is not compliant with the base standard and will not be interoperable with other contracts that accept any arbitrary ERC20.

As of right now, this token is not compatible with contracts that burn their tokens by transferring them to the zero address.

cleanunicorn:
As @zmitton suggested, keeping the `require(to != address(0))` in place will prevent bad software that cause empty arg or improper address parsing from sending the funds into the void.

However there is a need to burn tokens, that could be achieved by calling a `burn()` function which can remove the funds from the user balance and emit a `Transfer()` event.

Now there's a discussion where the destination of the `Transfer()` event should be. It could either be `address(0)` or (another suggestion is) [0xdead](https://etherscan.io/address/0x000000000000000000000000000000000000dead), making it specific about sending the funds into the void and decreasing the likelihood of a software bug.

frangio:
@c-o-l-o-r Thank you for expressing this concern. Sorry for not replying to it earlier.

OpenZeppelin contracts are opinionated precisely because we don't want the burden of making these decisions to be on the developers. If one of our decisions turns out to be a bad one, we don't want it to be in the library at all, not even as an opt-in thing.

Disallowing a send to the address `0` is compliant with the ERC20 spec, and it's widespread behavior across deployed tokens. The OpenZeppelin implementation is even listed in the spec itself.

I understand that it sucks to not have a standard way to burn tokens in ERC20, but this is a feature that is missing in the spec and not a problem in the implementations.

PR #2326, opened 2020-08-11
ERC20: Prevent transfer to the token contract

dmihal:
There's been a bit of discussion on Twitter today about preventing token transfers to the token contract, after a user lost about $35k worth of YFI.

https://twitter.com/dmihal/status/1293128845994995713
https://twitter.com/nanexcool/status/1293224689146486787

I threw together this PR to open the discussion on blocking transfers to the token contract.

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -209,6 +209,7 @@ contract ERC20 is Context, IERC20 {
     function _transfer(address sender, address recipient, uint256 amount) internal virtual {
         require(sender != address(0), "ERC20: transfer from the zero address");
         require(recipient != address(0), "ERC20: transfer to the zero address");
+        require(recipient != address(this), "ERC20: transfer to the token contract");
 
         _beforeTokenTransfer(sender, recipient, amount);
 


tinchoabbate:
Isn't this too opinionated ? Why enforce it at the ERC20 contract level when anyone inheriting from ERC20 could just do:

~~~solidity
contract Token is ERC20 {
    constructor() public ERC20("Token", "Token") {}
    function _transfer(address sender, address recipient, uint256 amount) internal override {
	    
	require(recipient != address(this), "ERC20: transfer to the token contract");
		
	super._transfer(sender, recipient, amount);
   }
}
~~~

and programmatically prevent the user error you mention ?

Today contracts inheriting from ERC20 might be actually using deposited tokens in the same token contract for other additional features, and this proposed change would break them.

dmihal:
@tinchoabbate 

Quoting @frangio from another thread

> OpenZeppelin contracts are opinionated precisely because we don't want the burden of making these decisions to be on the developers. If one of our decisions turns out to be a bad one, we don't want it to be in the library at all, not even as an opt-in thing.

https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1493#issuecomment-511084760

julianmrodri:
I think this was before the latest release in which it was decided by OZ to keep the `ERC20` contract as unopinionated as possible, and use the [preset contracts](https://docs.openzeppelin.com/contracts/3.x/api/presets) for opinionated, ready-to deploy implementations. 

We actually added a note in the `README.md`, you can check it [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/README.adoc).

IMHO this is too restrictive and opinionated to include it in the base standard contract. It is also as mentioned a breaking change for existing implementations.

frangio:
Thank you for the suggestion @dmihal.

I agree with @tinchoabbate on this one. Since there could be legitimate use cases for the token contract holding balance, this is a decision that we can't make for the developer.

PR #1739, opened 2019-05-07
Hardcode ERC777 granularity to 1, remove tests.

nventuro:
As discussed in https://github.com/OpenZeppelin/openzeppelin-solidity/pull/1735, the `ERC777` contract will have the granularity hardcoded to 1: we'll be providing a mixin down the line (if there's demand for it) that enables support for arbitrary granularities.

PR #2327, opened 2020-08-12
Fix function state mutability warning

zemse:
PR #1739 hardcodes ERC777 granularity to 1, this causes the `solc` compiler generates the following warning:
```
WARNING
  @openzeppelin/contracts/token/ERC777/ERC777.sol:116:5: Warning: Function state mutability can be restricted to pure
      function granularity() public view override returns (uint256) {
```

This PR fixes the above warning by changing the state mutability of granularity function from `view` to `pure`.

### DIFF
--- a/contracts/token/ERC777/ERC777.sol
+++ b/contracts/token/ERC777/ERC777.sol
@@ -113,7 +113,7 @@ contract ERC777 is Context, IERC777, IERC20 {
      *
      * This implementation always returns `1`.
      */
-    function granularity() public view override returns (uint256) {
+    function granularity() public pure override returns (uint256) {
         return 1;
     }
 


zemse:
This gives error in `0.6`, but in the newly released `0.7` first language feature, it is mentioned that `pure` can override `view`. Source: https://github.com/ethereum/solidity/releases/tag/v0.7.0

As seen the linter test is failing, looks like it also needs to be upgraded. I'll take a look if I can do that.

fvictorio:
I'm about to publish a version of solhint with support for 0.7. Should I test it in the `OpenZeppelin:solc-0.7` branch?

frangio:
@fvictorio Yes, please! Let me know how it goes.

fvictorio:
@frangio Just published solhint@3.2.0, that works fine in that branch (there's a bazillion warnings though). I tried to make a PR upgrading it but `node-gyp` is messing with me :cry:

frangio:
@fvictorio Thanks! Seems to work well. Just updated out 0.7 branch.

Aivolt1 (commented):
/

PR #2328, opened 2020-08-14
Fix ERC20Snapshot#_beforeTokenTransfer during mint

sunhurts:
## Fixes ERC20Snapshot#_beforeTokenTransfer

In ERC20Snapshot contract we have to update snapshots for the addresses involved in every transfer before the balances are updated. 
For Example: 
* before `transfer()`: we should update account snapshots for both `to` and `from` addresses 
* before `mint()`: we should update account snapshot for `to` address
* before `burn()`: we should update account snapshot for `from` address

This is implemented correctly on v3.1.0 but it overrides `_transfer`, `_mint` and `_burn` functions. In the master branch, this is done by using the `_beforeTokenTransfer` hook which is cleaner, but a small logic error crept in. Where this function updates the account snapshot of the from address during both mint and burn calls which is not correct. 

Updated `_beforeTokenTransfer` function
```
function _beforeTokenTransfer(address from, address to, uint256 amount) internal virtual override {

      if (from == address(0)) {
        // mint
        _updateAccountSnapshot(to);
        _updateTotalSupplySnapshot();
      } else if (to == address(0)) {
        // burn
        _updateAccountSnapshot(from);
        _updateTotalSupplySnapshot();
      } else {
        // transfer
        _updateAccountSnapshot(from);
        _updateAccountSnapshot(to);
      }
    }
```

### DIFF
--- a/contracts/token/ERC20/ERC20Snapshot.sol
+++ b/contracts/token/ERC20/ERC20Snapshot.sol
@@ -110,8 +110,12 @@ abstract contract ERC20Snapshot is ERC20 {
     function _beforeTokenTransfer(address from, address to, uint256 amount) internal virtual override {
       super._beforeTokenTransfer(from, to, amount);
 
-      if (from == address(0) || to == address(0)) {
-        // mint or burn
+      if (from == address(0)) {
+        // mint
+        _updateAccountSnapshot(to);
+        _updateTotalSupplySnapshot();
+      } else if (to == address(0)) {
+        // burn
         _updateAccountSnapshot(from);
         _updateTotalSupplySnapshot();
       } else {


julianmrodri:
Good catch. Makes a lot of sense!

frangio:
Merging to get this on master but we should add a test for this error.

julianmrodri:
@frangio  I took a deeper look at why this was working fine despite of the error, because it was quite surprising for me. 

The reason is that there is some sort of `try-catch` in the code, that in the case you don't have an account balance as part of the Snapshot, it returns the `current` balance:

`return snapshotted ? value : balanceOf(account);`

You can check the code [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/6d5a73815de843267c75f9134c4c11243fa895d8/contracts/token/ERC20/ERC20Snapshot.sol#L95)

So the scenario we have in the **test** file, it only checks on the latest Snapshot, which ends up defaulting to the current balance and that makes it always end in **success**.  

 If you are OK I can take care of this and add some more complex scenarios to **REALLY** validate that it is taking proper snapshots.

frangio:
Right! That is definitely the reason.

>If you are OK I can take care of this and add some more complex scenarios to REALLY validate that it is taking proper snapshots. I think just including one more iteration of snapshot with change values would be enough.

Sounds great! I agree that including a third snapshot would work.

julianmrodri:
Really sorry for this pal!! Was a tricky one.

frangio:
Don't worry about it the fault is on the reviewer (me).

sunhurts:
Hi @frangio @julianmrodri, thanks for the quick turn around! I was building a project that uses the ERC20Snapshot contract and saw that it overrides `_mint`, `_burn` and `_transfer` functions. So, thought I would create a PR to update this and use the `_beforeTokenTransfer` function instead. Then I checked the master branch and noticed that it was already implemented, but was slightly different from what I had. When I looked closer it was missing the `_mint()` condition, so created this PR instead. Cheers guys, have a great weekend.

julianmrodri:
@frangio I looked deeper into this. The contract is handling well the additional snapshots even if they are changes, so that path does not help us to test the error we had before. 

The reason is that the `transfer` operation is already creating a snapshot entry and the correct account balance is returned and managed after that point.  The way I found to force the issue  was to simply comment the `transfer` operation, and leave just for example the `mint` and `burn` operations. This would have made the  tests fail and we would have detected the issue.

One simple solution is to use the `other` account for the `mint` operation, so we ensure each operation (`transfer`, `mint`, `burn`)  is impacting a different account, so all need to be properly updated and processed independently. 

So we would be replacing this line:

`await this.token.mint(recipient, new BN('50'));`
with:
`await this.token.mint(other, new BN('50'));`

The line that needs to be changed is [this](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7d48d79b532f9f79a9984924ff94c0bb97da5980/test/token/ERC20/ERC20Snapshot.test.js#L137)

If we do this simple change we can capture issues on the `_beforeTokenTransfer` hook for each operation (independently)

To sum up, the contract is working really **well**. I tested with multiple Snapshots, changing balances and checking previous snapshots and is behaving really well. Then I realised this `transfer` operation was actually creating the needed snapshot and "covering" for the subsequent wrong `mint` update.

Let me know if you need additional explanations or examples, or something looks odd here. Thanks!

PR #2333, opened 2020-08-20
Add functionStaticCall and functionDelegateCall methods to Address library

k06a:
Consider having `functionStaticCall` and `functionDelegateCall` methods in `Address` library.


frangio:
@k06a Do you have any comments on the topic of avoiding the repeated code?

k06a:
@frangio we could aggregate only `call` and `delegatecall` since they are both non-view methods :(

frangio:
Ah that's a good point.

I think we should be able to put the if conditional inside a separate private function that can be reused though. Could you try that out in this PR?

k06a:
@frangio did, can't make it work with function pointers because of `view` modifiers.

frangio:
@k06a This PR is great but you seem to have deleted the natspec documentation. :smile: Let me know if you can add it back, along with similar docs for the new functions, or if you want me to take care of it.

Thank you!

k06a:
@frangio sorry about this, could someone help to restore it? You could simply push your own commits to my branch since I opened PR - it is allowed by github default settings :)

frangio:
No problem, I'll take care of it.

PR #2334, opened 2020-08-21
Fix/improve test description #1157

### LINKED ISSUE #1157: Improve test suite descriptions
The description of each tests (i.e. the text in the `describe`, `context` and `it` blocks) is a bit inconsistent, we should move towards a more standard structure.

I suggest:
 * `contract`(<ContractName>)
 * `describe`(<feature/function>)
 * `context`(<context (!), e.g: once deployed, in refund period, when disabled, etc>)
 * `it`(<verb in present simple, e.g. 'accepts payments', 'allows owner to transfer ownership' // avoid using 'should do x', go for 'does x' instead>)
    * Describe what you're testing, not what you're trying to prove: e.g. if testing a transaction fails from a non-owner account, write 'rejects non owners', instead of 'requires an owner account'. 
    * Use the terms 'null address' for address 0, 'zero-valued' for txs that don't transfer ether, and 'null [role]' for a role assigned to the null address (e.g. 'allows setting a null owner') 

This will allow us to write test chains that read very naturally, e.g.
`PausableToken transfer when paused reverts`

dibi91:
1157

Fixes #1157

This PR is for improve tests descriptions

### DIFF
--- a/test/access/Ownable.test.js
+++ b/test/access/Ownable.test.js
@@ -35,7 +35,7 @@ describe('Ownable', function () {
     it('guards ownership against stuck state', async function () {
       await expectRevert(
         this.ownable.transferOwnership(ZERO_ADDRESS, { from: owner }),
-        'Ownable: new owner is null address'
+        'Ownable: new owner is the zero address'
       );
     });
   });
--- a/test/access/Ownable.test.js
+++ b/test/access/Ownable.test.js
@@ -48,7 +48,7 @@ describe('Ownable', function () {
       expect(await this.ownable.owner()).to.equal(ZERO_ADDRESS);
     });
 
-    it('prevent non-owners from renouncement', async function () {
+    it('prevents non-owners from renouncement', async function () {
       await expectRevert(
         this.ownable.renounceOwnership({ from: other }),
         'Ownable: caller is not the owner'
--- a/test/access/Ownable.test.js
+++ b/test/access/Ownable.test.js
@@ -13,42 +13,46 @@ describe('Ownable', function () {
     this.ownable = await Ownable.new({ from: owner });
   });
 
-  it('should have an owner', async function () {
+  it('has an owner', async function () {
     expect(await this.ownable.owner()).to.equal(owner);
   });
 
-  it('changes owner after transfer', async function () {
-    const receipt = await this.ownable.transferOwnership(other, { from: owner });
-    expectEvent(receipt, 'OwnershipTransferred');
+  describe('transfer ownership', function () {
+    it('changes owner after transfer', async function () {
+      const receipt = await this.ownable.transferOwnership(other, { from: owner });
+      expectEvent(receipt, 'OwnershipTransferred');
 
-    expect(await this.ownable.owner()).to.equal(other);
+      expect(await this.ownable.owner()).to.equal(other);
+    });
+
+    it('prevents non-owners from transferring', async function () {
+      await expectRevert(
+        this.ownable.transferOwnership(other, { from: other }),
+        'Ownable: caller is not the owner'
+      );
+    });
+
+    it('guards ownership against stuck state', async function () {
+      await expectRevert(
+        this.ownable.transferOwnership(ZERO_ADDRESS, { from: owner }),
+        'Ownable: new owner is null address'
+      );
+    });
   });
 
-  it('should prevent non-owners from transferring', async function () {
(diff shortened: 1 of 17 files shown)


test/access/Ownable.test.js
```diff
@@ -13,42 +13,46 @@ describe('Ownable', function () {
...
-  });
+      expect(await this.ownable.owner()).to.equal(ZERO_ADDRESS);
+    });
 
-  it('should prevent non-owners from renouncement', async function () {
-    await expectRevert(
-      this.ownable.renounceOwnership({ from: other }),
-      'Ownable: caller is not the owner'
-    );
+    it('prevent non-owners from renouncement', async function () {
```
frangio, line 51:
```suggestion
    it('prevents non-owners from renouncement', async function () {
```

test/payment/PaymentSplitter.test.js
```diff
@@ -54,45 +54,48 @@ describe('PaymentSplitter', function () {
...
+      });
     });
 
-    it('should throw if non-payee want to claim', async function () {
-      await send.ether(payer1, this.contract.address, amount);
-      await expectRevert(this.contract.release(nonpayee1),
-        'PaymentSplitter: account has no shares'
-      );
+    describe('release', async function () {
+      it('does revert if no funds to claim', async function () {
```
frangio, line 85:
```suggestion
      it('reverts if no funds to claim', async function () {
```

test/payment/PaymentSplitter.test.js
```diff
@@ -54,45 +54,48 @@ describe('PaymentSplitter', function () {
...
+        await expectRevert(this.contract.release(payee1),
+          'PaymentSplitter: account is not due payment'
+        );
+      });
+      it('does revert if non-payee want to claim', async function () {
```
frangio, line 90:
```suggestion
      it('reverts if non-payee want to claim', async function () {
```

test/utils/Counters.test.js
```diff
@@ -9,23 +9,26 @@ describe('Counters', function () {
   beforeEach(async function () {
     this.counter = await CountersImpl.new();
   });
-
-  it('starts at zero', async function () {
-    expect(await this.counter.current()).to.be.bignumber.equal('0');
+  context('once deployed', function () {
```
frangio, line 12: We should only add a `once deployed` context for contracts where we have some tests "before" deployment, for example if we have to test that deployment fails because of invalid constructor arguments.

I think all of the `once deployed` contexts you added in this PR should be removed. It's okay if an `it` is not inside a `context`.

test/utils/Create2.test.js
```diff
@@ -23,71 +23,82 @@ describe('Create2', function () {
...
+    });
   });
 
-  it('should deploy a ERC20Mock with correct balances', async function () {
-    const offChainComputed = computeCreate2Address(saltHex, constructorByteCode, this.factory.address);
-
-    await this.factory.deploy(0, saltHex, constructorByteCode);
+  describe('deployERC1820Implementer', function () {
+    context('once deployed', function () {
+      it('deploys a ERC1820Implementer from inline assembly code', async function () {
```
frangio, line 48: This test should go in the `deploy` describe block.

dibi91:
Thanks a lot for your reviews, will apply the fixes asap.

dibi91:
Left context in `EnumerableMap` and `EnumerableSet` tests when testing `remove`, I only removed the 'once deployed' part, it's that correct?
Sorry if I've made some mistake/mess with commits and this pull request, this is my first pull request ever :sweat_smile:

frangio:
> Left context in `EnumerableMap` and `EnumerableSet` tests when testing `remove`, I only removed the 'once deployed' part, it's that correct?

It would be best to remove the context there too. Context is useful when we need to apply the same `beforeEach` block to a number of tests. Otherwise, I prefer to keep it simpler with less levels of nesting.

dibi91:
> > Left context in `EnumerableMap` and `EnumerableSet` tests when testing `remove`, I only removed the 'once deployed' part, it's that correct?
> 
> It would be best to remove the context there too. Context is useful when we need to apply the same `beforeEach` block to a number of tests. Otherwise, I prefer to keep it simpler with less levels of nesting.

Ok, I get it, now it's more clear to me.
Removed and committed, thank you really much.

dibi91:
Thank you! will try to help more💪

PR #2337, opened 2020-08-24
Update README.md

jithin-zachariah:
changed the package import path

Fixes #


frangio:
Hi @jithin-zachariah, the package in the readme is already correct. We also publish under `openzeppelin-solidity` but that is an older package name and `@openzeppelin/contracts` is the preferred one.

Thank you.

PR #2343, opened 2020-09-01
Fix typos

tinchoabbate:
PR to fix a number of typos in the contracts' documentation.


abcoathup:
Hi @tinchoabbate,

What did you use to check for typos?  
I used a spell check extension in VS Code and found a different set (with some overlap) of typos:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2338

PR #2345, opened 2020-09-02
Adapt proxies to Contracts conventions

frangio:
- Renames `_willFallback` to `_beforeFallback`, in line with the existing hooks we have like `_beforeTokenTransfer`.
- Makes private the low level setters `_setImplementation` (`_upgradeTo` remains internal) and `_setAdmin`.
- Makes `Initializable` an abstract contract, to signal that it's not meant to be deployed on its own.

### DIFF
--- a/contracts/proxy/TransparentUpgradeableProxy.sol
+++ b/contracts/proxy/TransparentUpgradeableProxy.sol
@@ -92,7 +92,9 @@ contract TransparentUpgradeableProxy is UpgradeableProxy {
      * NOTE: Only the admin can call this function. See {ProxyAdmin-changeProxyAdmin}.
      */
     function changeAdmin(address newAdmin) external ifAdmin {
-        _changeAdmin(newAdmin);
+        require(newAdmin != address(0), "TransparentUpgradeableProxy: new admin is the zero address");
+        emit AdminChanged(_admin(), newAdmin);
+        _setAdmin(newAdmin);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


contracts/proxy/TransparentUpgradeableProxy.sol
```diff
@@ -131,6 +129,17 @@ contract TransparentUpgradeableProxy is UpgradeableProxy {
         }
     }
 
+    /**
+     * @dev Changes the admin of the proxy.
+     * 
+     * Emits an {AdminChanged} event.
+     */
+    function _changeAdmin(address newAdmin) internal {
+        require(newAdmin != address(0), "TransparentUpgradeableProxy: new admin is the zero address");
```
spalladino, line 138: I would have expected the require to be part of the external function, not the internal, eg in case someone wants to implement a `renounceAdmin` function and need to make a `_changeAdmin(0)` call. But if this is in line with the conventions in Contracts, let's go ahead.

spalladino (approved):

contracts/proxy/TransparentUpgradeableProxy.sol
frangio: Okay I took a better look at this and `Owable` only provides the public `transferOwnership` which will check that the sender is the current owner. And `AccessControl` provides `_setupRole` internally _as an exception_ bypassing the role admin because pragmatically it can really useful to set up roles in the constructor.

So if there is a convention it is that this function should not exist internally, and the contract would force this change to go through the admin.

frangio: To follow this convention I think the right thing to do is to have no internal way to change the admin. I'm comfortable with this, we can reconsider if someone asks for it, but I highly doubt it would happen.

spalladino (approved):
Let's go ahead and favor consistency then. We will eventually have a good talk about extensibility vs footguns for the whole project, and we can address this particular case as an instance of that bigger discussion!

frangio:
Agree!

PR #2354, opened 2020-09-09
Add TimelockController

Amxx:
A contract to enable timelocked maintenance operations. Particularly usefull when EOA or multisigs are in charge of smart contract updates.

This PR includes:
- Contract
- Testing
- Documentation

Many design characteristics have been discussed on Slack before arriving to this result, but I'll be able to discuss them again here and eventually change some details.


contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+     */
+    function isOperationDone(bytes32 id) public view returns (bool done) {
+        return _timestamps[id] == _DONE_TIMESTAMP;
+    }
+
+    /**
+     * @dev Returns the timestamp at with an operation becomes ready (0 for
+     * unset operations, 1 for done operations).
+     */
+    function viewTimestamp(bytes32 id) public view returns (uint256 timestamp) {
```
frangio, line 110: For getters we use the naming convention `getTimestamp`.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    /**
+     * @dev Modifier to make a function callable only by a certain role. In
+     * addition to checking the sender's role, `address(0)` 's role is also
+     * considered. Granting a role to `address(0)` is equivalent to enabling
+     * this role for everyone.
+     */
+    modifier onlyRole(bytes32 role) {
+        require(hasRole(role, _msgSender()) || hasRole(role, address(0)), "TimelockController: sender requiers permission");
```
frangio, line 75: I'm not sure how I feel about this extra semantics on top of `AccessControl`.

I think this should be either built into `AccessControl`, or maybe an additional contract that extends `AccessControl` to add this functionality. (Or some other way to implement opening a role to the public.)

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+     * unset operations, 1 for done operations).
+     */
+    function viewTimestamp(bytes32 id) public view returns (uint256 timestamp) {
+        return _timestamps[id];
+    }
+
+    /**
+     * @dev Returns the minimum delay for an operation to become valid.
+     */
+    function viewMinDelay() public view returns (uint256 duration) {
```
frangio, line 117: Same as above, `getMinDelay`.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+    function hashOperation(address target, uint256 value, bytes calldata data, bytes32 predecessor, bytes32 salt) public pure returns (bytes32 hash) {
+        return keccak256(abi.encode(target, value, data, predecessor, salt));
+    }
+
+    /**
+     * @dev Returns the identifier of an operation containing a batch of
+     * transactions.
+     */
+    function hashOperationBatch(address[] calldata targets, uint256[] calldata values, bytes[] calldata datas, bytes32 predecessor, bytes32 salt) public pure returns (bytes32 hash) {
+        return keccak256(abi.encode(targets, values, datas, predecessor, salt));
```
frangio, line 134: Is there any way in which the output of `hashOperationBatch` could have a collision with some output of `hashOperation`? That is, are the images of `keccak256(abi.encode(target, value, data, predecessor, salt)` and the batch `abi.encode(targets, values, datas, predecessor, salt)` disjoint?

I don't remember exactly how arrays are encoded, but if I'm not mistaken the only inputs that produce the same encoding in this case are with `target = 0, value = 0, data = ""` and `targets = [], values = [], data = []`, which doesn't look like it could be a problem.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+ *
+ * This contract is designed to be self administered, meaning it should be its
+ * own administrator. The proposer (resp executor) role is in charge of
+ * proposing (resp executing) operation. A common use case is to position this
+ * {TimelockController} as the owner of a smart contract, with a multisig or a
+ * DAO as the sole proposer. Once at least one proposer and one executor have
+ * been appointed, self-administration can be enable using the `makeLive`
+ * function.
+ */
+contract TimelockController is AccessControl {
```
frangio, line 25: By convention all of our non-view functions should be `virtual`. (See [How we plan to adopt Solidity 0.6 virtual functions](https://forum.openzeppelin.com/t/how-we-plan-to-adopt-solidity-0-6-virtual-functions/2063).)

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+        _schedule(id, delay);
+        for (uint256 i = 0; i < targets.length; ++i) {
+            emit CallScheduled(id, i, targets[i], values[i], datas[i], predecessor);
+        }
+    }
+
+    /**
+     * @dev Schedule an operation that is to becomes valid after a given delay.
+     */
+    function _schedule(bytes32 id, uint256 delay) internal {
```
frangio, line 175: I would consider making this function private. In general we take a conservative approach and keep things private until clear use cases for extending/overriding present. There is some criticism about this approach, and we have our doubts too (maybe we should be more permissive about everything), but we should stick to the approach until we have a discussion about it and decide to change it.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * @dev Execute a transaction.
+     *
+     * Emits a {CallExecuted} event.
+     */
+    function _call(bytes32 id, uint256 index, address target, uint256 value, bytes calldata data) internal returns (bool) {
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success,) = target.call{value: value}(data);
+        require(success, "TimelockController: underlying transaction reverted");
```
frangio, line 248: I would like to use [`Address.functionCall`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/09d437c1f9f68c4c242db8e489cf3ab882b7a1f3/contracts/utils/Address.sol#L79) here, which handles the revert nicely by forwarding a revert reason if there is one. However, it will revert if the target is not a contract, so it wouldn't work for simple value transfers.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+     *
+     * Emits one {RoleGranted} and one {RoleRevoked} event.
+     *
+     * Requirements:
+     *
+     * - the caller must be the only address with `ADMIN_ROLE`.
+     * - there must be at least one account with 'PROPOSER_ROLE' and one with
+     *   'EXECUTOR_ROLE'.
+     */
+    function makeLive() external /* onlyRole(ADMIN_ROLE) */ {
```
frangio, line 277: I've been thinking about this function and I think we should remove it. Although we think the pattern makes sense, the name `makeLive` implies that we see it as a requirement for a timelock to be "right". But I think that will depend on the team. If a team wants to implement this pattern, they are free to do so, but we shouldn't force them.

Maybe the timelock should have `ADMIN_ROLE` by default assigned in the constructor, and the pattern would be reduced to renouncing the role from the deployer contract.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * @dev Emitted when the minimum delay for future operations is modified.
+     */
+    event MinDelayChange(uint256 newDuration, uint256 oldDuration);
+
+    /**
+     * @dev Initializes the contract with a given `minDelay`.
+     */
+    constructor(uint256 minDelay) public {
```
frangio, line 58: Maybe to make setup easier/cheaper proposers and executors should be assigned in the constructor by receiving arrays of addresses.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * @dev Emitted when operation `id` is cancelled.
+     */
+    event Cancelled(bytes32 indexed id);
+
+    /**
+     * @dev Emitted when the minimum delay for future operations is modified.
+     */
+    event MinDelayChange(uint256 newDuration, uint256 oldDuration);
```
frangio, line 53: The arguments are in the opposite order that we usually use. See [`Ownable.OwnershipTransferred`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/09d437c1f9f68c4c242db8e489cf3ab882b7a1f3/contracts/access/Ownable.sol#L21).

frangio (commented):
Nice work @Amxx! This is a first part of the review.

contracts/access/TimelockController.sol
frangio: As discussed offline to avoid this collision we want to require that the arrays have at least one element.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+    bytes32 public constant EXECUTOR_ROLE = keccak256("EXECUTOR_ROLE");
+    uint256 internal constant _DONE_TIMESTAMP = uint256(1);
+
+    mapping(bytes32 => uint256) private _timestamps;
+    uint256 private _minDelay;
+
+    /**
+     * @dev Emitted when a call is scheduled as part of operation `id`.
+     */
+    event CallScheduled(bytes32 indexed id, uint256 indexed index, address target, uint256 value, bytes data, bytes32 predecessor);
```
frangio, line 38: I think this event should also contain the `delay` number.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * @dev Cancel an operation.
+     *
+     * Requirements:
+     *
+     * - the caller must have the 'proposer' role.
+     */
+    function cancel(bytes32 id) public onlyRole(PROPOSER_ROLE) {
+        require(!isOperationDone(id), "TimelockController: operation is already executed");
```
frangio, line 204: I'd use `isOperationPending` instead. If i'm not mistaken the only difference is whether the function works on already unset operations?

```suggestion
        require(isOperationPending(id), "TimelockController: operation is already executed");
```

contracts/access/TimelockController.sol
Amxx: After tests, I noticed that the example above does not cause collision:

```
web3.eth.abi.encodeParameters([
	'address',
	'uint256',
	'bytes',
	'uint256',
	'bytes32',
], [
	constants.ZERO_ADDRESS,
	0,
	'0x',
	constants.ZERO_BYTES32,
	constants.ZERO_BYTES32,
])
→ 0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
web3.eth.abi.encodeParameters([
	'address[]',
	'uint256[]',
	'bytes[]',
	'uint256',
	'bytes32',
], [
	[],
	[],
	[],
	constants.ZERO_BYTES32,
	constants.ZERO_BYTES32,
])
→ 0x00000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+     *
+     * Emits a {CallExecuted} event.
+     *
+     * Requirements:
+     *
+     * - the caller must have the 'executor' role.
+     */
+    function execute(address target, uint256 value, bytes calldata data, bytes32 predecessor, bytes32 salt) public payable onlyRole(EXECUTOR_ROLE) {
+        bytes32 id = hashOperation(target, value, data, predecessor, salt);
+        _execute(id, predecessor);
```
frangio, line 221: The operation should be marked as executed only after all calls are made, as protection against potential reentrancy.

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+     * Requirements:
+     *
+     * - the caller must have the 'executor' role.
+     */
+    function executeBatch(address[] calldata targets, uint256[] calldata values, bytes[] calldata datas, bytes32 predecessor, bytes32 salt) public payable onlyRole(EXECUTOR_ROLE) {
+        require(targets.length == values.length, "TimelockController: length missmatch");
+        require(targets.length == datas.length, "TimelockController: length missmatch");
+
+        bytes32 id = hashOperationBatch(targets, values, datas, predecessor, salt);
+        _execute(id, predecessor);
```
frangio, line 239: Same as above: the operation should be marked as executed only after all calls are made, as protection against potential reentrancy.

contracts/access/TimelockController.sol
Amxx: I think the ability to send value to EOA is essential, has the timelock might receive ETH as part of maintenance operations. (some contract might send value to the owner ... of might send funds back to the owner in case of selfdestruct...)

So yes for `Address.functionCall` unless it reverts when calling EOAs

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,294 @@
...
+        require(predecessor == bytes32(0) || isOperationDone(predecessor), "TimelockController: missing dependency");
+        _timestamps[id] = _DONE_TIMESTAMP;
+    }
+
+    /**
+     * @dev Execute a transaction.
+     *
+     * Emits a {CallExecuted} event.
+     */
+    function _call(bytes32 id, uint256 index, address target, uint256 value, bytes calldata data) internal returns (bool) {
```
frangio: Return value seems unused. 

```suggestion
    function _call(bytes32 id, uint256 index, address target, uint256 value, bytes calldata data) internal {
```

frangio (commented):
Some more code comments. (Sorry for the delay.)

contracts/access/TimelockController.sol
frangio: Yes, I agree. I edited my comment after your reply but before seeing it. My conclusion:

>Maybe we can leave this to be improved by a subsequent PR.

contracts/access/TimelockController.sol
frangio: Oh I see, that makes sense. We were thinking of the way arrays are encoded in Solidity storage or memory, which is different from the ABI encoding.

frangio: After some tinkering I came close to a collision but found that for the `bytes` value in the non-batch encoding there will always be a `0x0...0a0` in the third word, whereas in the batch encoding that word will be `0x0...0e0` or greater. Since those two words will always be different, there can't be a collision. In this case we can keep the code as is, I also don't see the need to add the length restriction.

contracts/access/TimelockController.sol
Amxx: `makeLive` has been removed

contracts/access/TimelockController.sol
Amxx: done

contracts/access/TimelockController.sol
Amxx: changed

contracts/access/TimelockController.sol
Amxx: delay or timestamp ?

contracts/access/TimelockController.sol
Amxx: yes, that is the only difference.

contracts/access/TimelockController.sol
Amxx: resolved

contracts/access/TimelockController.sol
Amxx: Reentrancy trying to cancel the current operations are not an issue since the would cause the tx to revert, regarding of the order.

Similarly, an operation trying to re-execute itself would lead to a revert. If the marking is done before the call, the "inner" execute would revert, which might eventually get caught. If the marking is done after the call, the "outer" execute would revert, thus reverting everything. So for re-executing reentrancy, I'd prefer marking before running.

There is a reentrancy issue if a batch operation tries to execute its child before all the call in the batch have been called. This is possible since the batch would be marked as done before all calls have been done. This is potentially dangerous but is very difficult to do:
- The father operation needs to include the execution of the child operation, so it must be part of the data that produces the id
- The child operation's id depend on its predecessor's ...
This hash loop can be broken by storing the child operation id in an execution proxy that would be called by the father operation.

Have I missed other reentrancy attacks?

Amxx: Drawback with checking after the call is that is the check is to fail, gas has already been consumed

contracts/access/TimelockController.sol
frangio: Maybe timestamp is the more useful number.

contracts/access/TimelockController.sol
frangio: What about only checking before the call, then re-checking and marking as executed after the call?

Yes I think the only danger is with batch operations as you mentioned. It's true that an attack would be difficult to put together but I don't think it justifies not putting in place the proper protections.

contracts/access/README.adoc
```diff
@@ -5,8 +5,94 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/contracts/
...
+
+TIP: The three arrays must have the same length. They contain the details of the calls, just like in a simple operation.
+
+=== Operation lifecycle
+
+Timelocked operations are identified by a unique id (their hash) and follow a specific lifecycle:
+
+`Unset` -> `Pending` -> `Pending` + `Ready` -> `Done`
+
+* By calling xref:api:access.adoc#TimelockController-schedule-address-uint256-bytes-bytes32-bytes32-uint256-[`schedule`] (or xref:api:access.adoc#TimelockController-scheduleBatch-address---uint256---bytes---bytes32-bytes32-uint256-[`scheduleBatch`]), a proposer moves the operation from the `Unset` to the `Pending` state. This starts a timer that must be longer than the minimum delay. The timer expires at a timestamp accessible through the xref:api:access.adoc#TimelockController-getTimestamp-bytes32-[`getTimestamp`] methods.
```
frangio, line 63:
```suggestion
* By calling xref:api:access.adoc#TimelockController-schedule-address-uint256-bytes-bytes32-bytes32-uint256-[`schedule`] (or xref:api:access.adoc#TimelockController-scheduleBatch-address---uint256---bytes---bytes32-bytes32-uint256-[`scheduleBatch`]), a proposer moves the operation from the `Unset` to the `Pending` state. This starts a timer that must be longer than the minimum delay. The timer expires at a timestamp accessible through the xref:api:access.adoc#TimelockController-getTimestamp-bytes32-[`getTimestamp`] method.
```

contracts/access/README.adoc
```diff
@@ -5,8 +5,94 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/contracts/
...
+* Once the timer expires, the operation gets the `Ready` state. At this point, it can be executed.
```
frangio, line 64: May be worth being explicit that this transition is automatic:

```suggestion
* Once the timer expires, the operation automatically gets the `Ready` state. At this point, it can be executed.
```

contracts/access/README.adoc
```diff
@@ -5,8 +5,94 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/contracts/
...
+
+* xref:api:access.adoc#TimelockController-isOperationPending-bytes32-[`isOperationPending(bytes32)`]
+* xref:api:access.adoc#TimelockController-isOperationReady-bytes32-[`isOperationReady(bytes32)`]
+* xref:api:access.adoc#TimelockController-isOperationDone-bytes32-[`isOperationDone(bytes32)`]
+
+=== Roles
+
+==== Admin
+
+The admins are in charge of managing proposers and executors. For the timelock to be self-governed, this role should only be given to the timelock itself. Upon deployment, both the timelock and the deployer have this role. After further configuration and testing, the deployer should renounce this role such that all further maintenance operations have to go through the timelock process.
```
frangio, line 78:
```suggestion
The admins are in charge of managing proposers and executors. For the timelock to be self-governed, this role should only be given to the timelock itself. Upon deployment, both the timelock and the deployer have this role. After further configuration and testing, the deployer can renounce this role such that all further maintenance operations have to go through the timelock process.
```

contracts/access/README.adoc
```diff
@@ -5,8 +5,94 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/contracts/
...
+
+This role is identified by the *PROPOSER_ROLE* value: `0xb09aa5aeb3702cfd50b6b62bc4532604938f21248a27a1d5ca736082b6819cc1`
+
+==== Executor
+
+The executors are in charge of executing the operations scheduled by the proposers once the timelock expires. Logic dictates that multisig or DAO that are proposers should also be executors in order to guarantee operations that have been scheduled will eventually be executed. However, having additional executor can reduce the cost (the executing transaction does not require validation by the multisig or DAO that proposed it), while ensuring whoever is in charge of execution cannot trigger actions that have not been scheduled by the proposers.
+
+This role is identified by the *EXECUTOR_ROLE* value: `0xd8aa0f3194971a2a116679f7c2090f6939c8d4e01a2a8d7e41d55e5351469e63`
+
+==== Warning
```
frangio, line 96:
```suggestion
```

contracts/access/README.adoc
```diff
@@ -5,8 +5,94 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/contracts/
...
+
+WARNING: A live contract without at least one proposer and one executor is locked. Make sure these roles are filled by reliable entities before the deployer renounces its administrative rights in favour of the timelock contract itself. See the {AccessControl} documentation to learn more about role management.
```
frangio, line 98: This got me thinking, if this is so dangerous maybe we should override `revokeRole`/`renounceRole` so that the contract can never enter this state for these specific roles...

frangio (commented):
Great work!

contracts/access/TimelockController.sol
Amxx: The event is in a transaction. The transaction has a block number. The block contains a timestamp.

In TheGraph it is as easy as doing `event.block.timestamp`

contracts/access/TimelockController.sol
Amxx: Double-checking saves gas for reverts, but increases for successful operations :/

Last commits include:
1. Checking the predecessor before doing the call
2. Checking pre-execution after the call
3. Marking completion after the call

1+3 prevent ordering error (particularly with batch)
2+3 prevents reentrancy on the same operation

test/access/TimelockController.test.js
```diff
@@ -0,0 +1,1089 @@
...
+          const receipt = await this.timelock.schedule(
+            this.operation.target,
+            this.operation.value,
+            this.operation.data,
+            this.operation.predecessor,
+            this.operation.salt,
+            MINDELAY,
+            { from: proposer }
+          );
+          expectEvent(receipt, 'CallScheduled', {
```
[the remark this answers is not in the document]
Amxx: addressed in 96e0122

contracts/access/TimelockController.sol
Amxx: That is for the current timestamp. If you are asking is the operation timestamp, this can either be recomputed (block timestamp + delay) or by querying the Timelock `getTimestamp` function with the id present in the event:

https://github.com/Amxx/openzeppelin-contracts-subgraphs/blob/master/templates/TimelockController/src/TimelockController.ts#L40

frangio: Looking at the JSON RPC interface: if you want to get the timestamp based on the delay contained in an event, you need an extra RPC call to get the timestamp for the event's blockHash. But conversely, if the event contains the timestamp, the same extra RPC call is required to compute the delay.

I think it's reasonable that interfaces will want to display both values, and the event should only contain one of them. Let's keep it as is and emit only the delay.

contracts/access/TimelockController.sol
frangio: Perfect!

contracts/access/README.adoc
```diff
@@ -5,8 +5,94 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/contracts/
...
+* *Predecessor*: An (optional) dependency between operations. An operation can depend on another operation (its predecessor), forcing the execution order of these two operations.
+* *Role*:
+** *Proposer:* An address (smart contract or EOA) that is in charge of scheduling (and cancelling) operations.
+** *executor:* An address (smart contract or EOA) that is in charge of executing operations.
+
+=== Operation structure
+
+Operation executed by the xref:api:access.adoc#TimelockController[`TimelockControler`] can contain one or multiple subsequent calls. Depending on whether you need to multiple calls to be executed atomically, you can either use simple or batched operations.
+
+==== Simple operations contain:
```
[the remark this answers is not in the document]
Amxx: addressed

contracts/access/TimelockController.sol
```diff
@@ -0,0 +1,281 @@
...
+ *
+ * By default, this contract is self administered, meaning administration tasks
+ * have to go through the timelock process. The proposer (resp executor) role
+ * is in charge of proposing (resp executing) operations. A common use case is
+ * to position this {TimelockController} as the owner of a smart contract, with
+ * a multisig or a DAO as the sole proposer.
+ */
+contract TimelockController is AccessControl {
+
+    bytes32 public constant ADMIN_ROLE = keccak256("ADMIN_ROLE");
```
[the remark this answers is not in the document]
Amxx: addressed

docs/modules/ROOT/pages/access-control.adoc
```diff
@@ -185,3 +185,35 @@ for (let i = 0; i < minterCount; ++i) {
...
+
+The first step in configuring the xref:api:access.adoc#TimelockController[`TimelockControler`] is to assign at least one proposer and one executor. These can be assigned during construction or later by anyone with the administrator role. These roles are not exclusive, meaning an account can have both roles.
+
+Roles are managed using the xref:api:access.adoc#AccessControl[`AccessControl`] interface and the `bytes32` values for each role are accessible through the `ADMIN_ROLE`, `PROPOSER_ROLE` and `EXECUTOR_ROLE` constants.
+
+TIP: Giving the proposer or executor role to `address(0)` opens access to anyone. This features, while potentially useful for testing and in some cases for the executor role, is dangerous and should be used with caution.
+
+At this point, with both a proposer and an executor assigned, the timelock can perform operations. The next step is for the administrator to renounce its administrative privileges and leave the proposers and executors in charge of the timelock.
```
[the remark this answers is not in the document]
Amxx: addressed

Amxx (commented):
addressing review cd0b97c

frangio (approved):
Thank you!

PR #2357, opened 2020-09-12
Remove unused `Address` library in ERC20.sol

Ro5s:
The current implementation of ERC20.sol in the OZ contracts seems to have an unused import of the Address library. This PR removes the import statement and header reference 'using Address for address.'


frangio (approved):
Thank you @Ro5s!

PR #2359, opened 2020-09-14
Minor syntax change on IERC721Receiver.sol to conform to OZ format

Ro5s:
Slight nit on spacing to reduce # of lines // conform to other OZ function syntax


frangio (approved):
Thanks!

PR #2362, opened 2020-09-16
Remove unused ganache-cli dependency

frangio:


frangio:
Seems like this is used by Test Environment for coverage.

PR #2371, opened 2020-09-23
adds Proxies factory library

### LINKED ISSUE #2339: Provide a proxy factory library
We should provide functionality to make it simple to create proxy instances at runtime. Creating a single proxy instance is simple using the proxy contracts provided, but it's worth providing something on top of that that implements a common pattern when the implementation contract needs to be reused. This pattern should be simple and not opinionated.

The proposed design is a library with a `ProxyFactory` "object" that would be used as in the following snippet.

```solidity
contract MyFactory {
    using Proxies for Proxies.ProxyFactory;

    // This struct contains the address of the implementation that is
    // used for the proxies it creates.
    struct Proxies.ProxyFactory factory;

    constructor() public {
        // Initializes the factory struct by deploying the implementation
        // and storing its address.
        factory = Proxies.newFactory(FooUpgradeable.creationCode);
    }

    // initializerData is the encoded call to the initializer function.
    function deployInstance(address admin, bytes initializerData) public {
        factory.deploy(admin, initializerData);
    }
}
```

In this case the factory creates instances of `TransparentUpgradeableProxy`, but the same pattern could be offered for other kinds of proxies. We should think about this as it will affect naming.

julianmrodri:
2339

Fixes #2339 

Implements a Proxy Factory Library names `Proxies` which allows to deploy multiple proxy instances when the implementation contract needs to be reused. 

This library can be utilized from a `Factory` contract which should provide the `creationCode` for the Implementation contract to deploy, the proxy `admin` address and the encoded `initializerData`. 

This PR is in DRAFT but wanted to check if its going in the right direction and what additional functionality we should include.

Pending tasks:
* Mockup Factory Contract
* Tests
* Docs

### DIFF
--- a/contracts/proxy/Proxies.sol
+++ b/contracts/proxy/Proxies.sol
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.6.0;
+
+import "./TransparentUpgradeableProxy.sol";
+
+/**
+ * @dev Proxy Factory library to create multiple proxy instances at runtime
+ *
+ * These functions implement a simple Factory pattern that can be used to deploy
+ * multiple proxy instances when the implementation contract needs to be reused. 
+ */
+library Proxies {
+  // Contains the address of the implementation used for the proxies it creates.
+  struct ProxyFactory { address implementation; }
+
+  /**
+   * @dev Returns the ProxyFactory struct with the impementation contract address.
+   * 
+   * Initializes the factory struct by deploying the implementation contract and
+   * storing its address.
+  */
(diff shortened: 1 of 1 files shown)


julianmrodri:
@frangio Here I send an initial bare bones proposal for the Proxy Factory library named `Proxies`

Let me know if this is in the right direction and what additional functions we might want to include.

I do have some tests and mockups done that can be uploaded relatively soon if we want to proceed. Thanks!!

frangio:
Following up from #2339.

>>Im not happy at all about importing the Proxy code in the library. 

What do you mean here?

frangio:
This PR is a good start, but given that the functionality is relatively simple and easy to replicate by developers, we're going to prioritize other work for now.

PR #2376, opened 2020-10-12
Replace `if (...) revert()` with `require(...)`

poma:
`if (...) revert()` pattern is outdated and should be replaced with `require`


frangio (approved):
Thanks @poma!

PR #2380, opened 2020-10-16
Update SafeERC20.sol - using const bytecode instead of selector

ductm54:
Fixes #
Using const bytecode instead of selector so it save some gas in SafeTransfer


frangio:
Thank you for the suggestion @ductm54.

I tried this out and Solidity inlines the constant and produces exactly the same code when using `.selector`.

PR #2383, opened 2020-10-17
Make contracts abstract if they had internal constructors

psykeeper:
In commit https://github.com/OpenZeppelin/openzeppelin-contracts/commit/04fc35707d97e925168b3f46694b606d12e667fd the `Ownable` contract constructor had its `internal` visibility removed to comply with breaking changes from the upgrade to solidity 0.7.0, however, the purpose of `internal` visibility was to restrict the `Ownable` contract from being deployed. As per the [docs](https://docs.openzeppelin.com/contracts/3.x/api/access#Ownable):

> This module is used through inheritance. 

Adding `abstract` will maintain this requirement.

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -15,7 +15,7 @@ import "../GSN/Context.sol";
  * `onlyOwner`, which can be applied to your functions to restrict their use to
  * the owner.
  */
-contract Ownable is Context {
+abstract contract Ownable is Context {
     address private _owner;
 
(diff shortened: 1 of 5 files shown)


frangio:
Thank you @psykeeper!

Would you mind doing the same for the other contracts that had `internal` removed in the same PR? I think there is a couple more.

psykeeper:
@frangio okay, it seems this PR now covers all of the changes that need to be made

frangio (approved):
Yes. Thanks!

PR #2394, opened 2020-10-22
Add Strings.toHex()

k06a:
Proposing converting `uint256`, `address`, `bool`, and `bytes32` to a hexadecimal string.

Gonna add tests and docs in case of the PR will be considered as useful :)

### DIFF
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -31,4 +31,32 @@ library Strings {
         }
         return string(buffer);
     }
+
+    function toHex(bool flag) internal pure returns(string memory) {
+        return toHex(abi.encodePacked(flag));
+    }
+
+    function toHex(address account) internal pure returns(string memory) {
+        return toHex(abi.encodePacked(account));
+    }
+
+    function toHex(uint256 value) internal pure returns(string memory) {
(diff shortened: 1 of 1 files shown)


k06a:
I use this mostly for debugging, and believe this is more compact and efficient way to present numbers. To debug sometimes I use:
```solidity
revert(toHex(value));
```

frangio:
I think this can be useful but we would need tests!

TomasFlow (approved):

TomasFlow (approved):

k06a:
@frangio do you think we also need non-padded version for `uint256`? To print `123` as `0x7B` instead of `0x000000000000000000000000000000000000000000000000000000000000007B`.

frangio:
@k06a Which version is most useful in your experience? I would prefer to add a single version.

k06a:
@frangio proposed implementation is very compact, I would prefer to keep it:

image

frangio:
If we were to implement trimming of zero-padding I would do it in `toHex(bytes)`, so `toHex(uint256)` etc. would remain the same.

`toHex(bytes)` is already not so compact, but I don't know how many lines would have to be added.

frangio:
Closing this in favor of #2504

PR #2395, opened 2020-10-23
Add BytesSet

nventuro:
This adds a (trivial) set for type `bytes32`. `bytes32` is very commonly used for identifiers, since it is the result of a hash operation, and unlike `uint256` doesn't have arithmetic operators.


frangio (commented):
Thanks!

I think we should change the name to `Bytes32Set`, since `bytes` is the type for dynamic length byte arrays in Solidity. (Note that `UintSet` doesn't have this problem because `uint` and `uint256` are synonyms.)

frangio (approved):
LGTM. Thanks!

PR #2396, opened 2020-10-25
Fix/solc 0.7.4 warnings #2391

### LINKED ISSUE #2391: Warning: This declaration shadows an existing declaration.
There are lot of `Warning: This declaration shadows an existing declaration.`.

Is there a way to silence these warnings.

I'm using solc 0.7.4 

```
Compilation warnings: 
contracts/token/ERC20/ERC20.sol:57:18: Warning: This declaration shadows an existing declaration.
    constructor (string memory name, string memory symbol) {
                 ^----------------^
contracts/token/ERC20/ERC20.sol:66:5: The shadowed declaration is here:
    function name() public view returns (string memory) {

```

zemse:
Fixes #2391

This PR aims to remove plenty of warnings that show up when compiling contracts with solc `0.7.4`.

There were three kinds of warnings:

1. declaration shadowing
```
WARNING
  contracts/proxy/TransparentUpgradeableProxy.sol:33:33: Warning: This declaration shadows an existing declaration.
      constructor(address _logic, address _admin, bytes memory _data) payable UpgradeableProxy(_logic, _data) {
                                  ^------------^
  contracts/proxy/TransparentUpgradeableProxy.sol:126:5: The shadowed declaration is here:
      function _admin() internal view returns (address adm) {
```

2. return variable
```
WARNING
  contracts/proxy/TransparentUpgradeableProxy.sol:70:48: Warning: Unnamed return variable can remain unassigned. Add an explicit return with value to all non-reverting code paths or name the variable.
      function admin() external ifAdmin returns (address) {
                                                 ^-----^
```

3. state mutability
```
WARNING
  contracts/mocks/GSNRecipientMock.sol:14:5: Warning: Function state mutability can be restricted to pure
      function acceptRelayedCall(address, address, bytes calldata, uint256, uint256, uint256, uint256, bytes calldata, uint256)
```

1. To prevent declaration shadowing, contract method arguments are suffixed with an underscore. Given that `name` naming is generally used for public vars or functions and `_name` naming for private vars or functions. Just to mention, this convention is already present in OpenZeppelin contracts and many other projects use this convention. Also, many devs use underscore prepending to function args to prevent shadowing but since solc 0.7.4, the compiler highlights if that shadows functions too (which were missed before). So having an alternate convention for arguments easily solves the problem. Inspired from [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e5fbbda9bac49039847a7ed20c1d966766ecc64a/contracts/token/ERC20/ERC20.sol#L287), the convention of having `name_` naming for function arguments to prevent shadowing is taken in this PR in the first commit.

2. The second commit tries to silence this warning.

3. The third commit simply changes `view` to `pure`.

Please let me know if this is fine or any changes that need to be done.

### DIFF
--- a/contracts/GSN/GSNRecipientERC20Fee.sol
+++ b/contracts/GSN/GSNRecipientERC20Fee.sol
@@ -86,6 +86,8 @@ contract GSNRecipientERC20Fee is GSNRecipient {
 
         // The maximum token charge is pre-charged from the user
         _token.safeTransferFrom(from, address(this), maxPossibleCharge);
+
+        return 0;
     }
 
     /**
--- a/contracts/introspection/IERC1820Registry.sol
+++ b/contracts/introspection/IERC1820Registry.sol
@@ -59,7 +59,7 @@ interface IERC1820Registry {
      * queried for support, unless `implementer` is the caller. See
      * {IERC1820Implementer-canImplementInterfaceForAddress}.
      */
-    function setInterfaceImplementer(address account, bytes32 interfaceHash, address implementer) external;
+    function setInterfaceImplementer(address account, bytes32 _interfaceHash, address implementer) external;
 
     /**
      * @dev Returns the implementer of `interfaceHash` for `account`. If no such
@@ -70,7 +70,7 @@ interface IERC1820Registry {
      *
      * `account` being the zero address is an alias for the caller's address.
      */
-    function getInterfaceImplementer(address account, bytes32 interfaceHash) external view returns (address);
+    function getInterfaceImplementer(address account, bytes32 _interfaceHash) external view returns (address);
 
     /**
      * @dev Returns the interface hash for an `interfaceName`, as defined in the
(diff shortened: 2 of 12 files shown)


frangio:
Thank you @zemse!

I'm not very fond of the underscore suffix. I know we use it in a few places, but so far it has only been when the underscore prefix was not available due to shadowing. Rather than adopting this convention for all arguments, I would prefer to continue with our current convention:

1. arguments by default are not underscored,
2. if the name isn't available due to shadowing (and a different name is not reasonable), use prefix underscore,
3. if prefix underscore isn't available due to shadowing, use suffix underscore.

There may be some issues with this approach but I would rather we solved the warnings as simply as possible first, then discuss the issues with this convention and if necessary come up with a new one.

Would you like to make the changes I suggested?

zemse:
I get the point. This is something which I could discuss before actually making the change.

> Would you like to make the changes I suggested?

Yeah, likely by EOD.

zemse:
@frangio I've removed previous commit and added new commit based on the followed conventions. Can you please have a look?

frangio (approved):
Awesome work, thank you!

PR #2397, opened 2020-10-28
Migrate to Hardhat

frangio:
We recently [announced](https://forum.openzeppelin.com/t/building-for-interoperability-why-we-re-focusing-on-upgrades-plugins/4088) we will be focusing our upgradeability efforts around Contracts and the [Upgrades plugins](https://github.com/OpenZeppelin/openzeppelin-upgrades), and halting development of the OpenZeppelin CLI.

Thus, this PR migrates this project to use Hardhat, until recently known as Buidler. (We will actually be using Buidler until solidity-coverage upgrades, see https://github.com/sc-forks/solidity-coverage/issues/541.)


buidler.config.js
```diff
@@ -0,0 +1,23 @@
+usePlugin('solidity-coverage');
+usePlugin('@nomiclabs/buidler-truffle5');
+
+extendEnvironment(env => {
+  const { contract } = env;
+  env.contract = function (name, body) {
+    // remove the default account from the accounts list used in tests, in order
+    // to protect tests against accidentally passing due to the contract
+    // deployer being used subsequently as function caller
+    contract(name, accounts => body(accounts.slice(1)));
```
nventuro, line 10: I love this. @alcuadrado make this the default behavior pretty please

nventuro (approved):
:construction: :construction_worker_woman: :construction_worker_woman: :construction:

buidler.config.js
alcuadrado: I'm not sure I understand this. Is it like the `contract` thing in truffle?

nventuro: This, this is a truffle plugin. I was referring to the feature of hiding the default account however.

frangio: It's kind of opinionated, and I suppose the goal of the Truffle plugins is to have as much compatibility as possible.

Issue #2377, opened 2020-10-15
Add ERC20 and ERC777 presets with fixed supply


abcoathup:
**🧐 Motivation**
Create simple ERC20 and ERC777 presets with fixed supply.
The benefits are:
* Easy to deploy (including from Remix).
* No access control (for minting/pausing), and hence no governance required.
* Can be verified easily on public networks and then are verified for all.

**📝 Details**

Community members are creating their own simple ERC20 tokens, so it would be good if they could use a Preset from Contracts which wouldn't require any code changes.

To verify contracts as multi-file it is best to use Buidler Etherscan plugin (Truffle plugin is getting updated).  This can be a challenge for community members deploying from Remix.  A preset contract once verified on a network would be verified for all subsequent deployments.

The Simple ERC777 token example in the forum has 4K+ views so there is potential interest in an ERC777 preset too.
https://forum.openzeppelin.com/t/simple-erc777-token-example/746

Potential presets could look like the following:

## Draft ERC20PresetFixedSupply 
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.6.2;

import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
import "@openzeppelin/contracts/token/ERC20/ERC20Burnable.sol";

contract ERC20PresetFixedSupply is ERC20, ERC20Burnable {
    constructor(
        string memory name,
        string memory symbol,
        uint256 initialSupply
    ) public ERC20(name, symbol) {
        _mint(msg.sender, initialSupply);
    }
}
```

## Draft ERC777PresetFixedSupply 
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.6.2;

import "@openzeppelin/contracts/token/ERC777/ERC777.sol";

contract ERC777PresetFixedSupply is ERC777 {
    constructor(
        string memory name,
        string memory symbol,
        uint256 initialSupply,
        address[] memory defaultOperators
    ) public ERC777(name, symbol, defaultOperators) {
        _mint(msg.sender, initialSupply, "", "");
    }
}
```

ashwinYardi:
Hi @frangio @abcoathup ! Have submitted the [PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2399) .

Can you please take a look whenever possible and let me know if anything needs to be changed / added? Thanks!

frangio:
[...]

PR #2399, opened 2020-11-03
Add ERC20 and ERC777 fixed supply presets #2377

ashwinYardi:
Open issue: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2377

Additions #

1. Added ERC20 preset contract with with fixed initial supply.
2. Added ERC777 preset contract with fixed initial supply.
3. Added tests for both the presets.


contracts/presets/ERC20PresetFixedSupply.sol
```diff
@@ -0,0 +1,21 @@
...
+ * Tip: Contract is preconfigured with fixed initial supply.
+ * minting / pausing functionality is not supported. This removes the need of access control and thereby governance. 
+ */
+contract ERC20PresetFixedSupply is ERC20, ERC20Burnable {
+    constructor(
+        string memory name,
+        string memory symbol,
+        uint256 initialSupply
+    ) public ERC20(name, symbol) {
+        _mint(msg.sender, initialSupply);
```
frangio, line 19: I think it would be better to use an address in the constructor arguments instead of `msg.sender`. This improves "composability", for example it allows you to use a generic factory to deploy the contract and assign the initial balance to a separate account.

In hindsight this may be an issue with the current roles presets, since they assign roles to the deployer account.

test/presets/ERC777PresetFixedSupply.js
```diff
@@ -0,0 +1,53 @@
+const { BN } = require('@openzeppelin/test-helpers');
+
+const { expect } = require('chai');
+
+const ERC777PresetFixedSupply = artifacts.require('ERC777PresetFixedSupply');
+
+contract('ERC777', function (accounts) {
```
frangio, line 7:
```suggestion
contract('ERC777PresetFixedSupply', function (accounts) {
```

contracts/presets/ERC20PresetFixedSupply.sol
```diff
@@ -0,0 +1,21 @@
...
+import "../token/ERC20/ERC20.sol";
+import "../token/ERC20/ERC20Burnable.sol";
+
+/**
+ * @dev Extension of {ERC20} and {ERC20Burnable} with fixed initialSupply.
+ *
+ * Tip: Contract is preconfigured with fixed initial supply.
+ * minting / pausing functionality is not supported. This removes the need of access control and thereby governance. 
+ */
+contract ERC20PresetFixedSupply is ERC20, ERC20Burnable {
```
frangio, line 13: What was the reason for including `ERC20Burnable`?

test/presets/ERC777PresetFixedSupply.js
```diff
@@ -0,0 +1,53 @@
...
+  const [defaultOperatorA, defaultOperatorB, anyone] = accounts;
+
+  const initialSupply = new BN('10000');
+  const name = 'ERC777Preset';
+  const symbol = '777P';
+
+  const defaultOperators = [defaultOperatorA, defaultOperatorB];
+
+  context('with default operators', function () {
```
frangio, line 16: You should be testing the balance of the initial account.

frangio (commented):
Thanks @ashwinYardi!

contracts/presets/ERC20PresetFixedSupply.sol
ashwinYardi: Got it! Replacing msg.sender by an argument 'owner' of type address.

contracts/presets/ERC20PresetFixedSupply.sol
ashwinYardi: Burn functionality would enable the use of this preset for the purposes of deflationary tokens, correct?

Skyge: Emmm, I am confused the can I just write `contract ERC20PresetFixedSupply is ERC20Burnable`, cause `ERC20Burnable` is derived from `ERC20`

ashwinYardi: ahh, yes you are right! Thanks @Skyge

ashwinYardi:
@frangio @Skyge Some info about the testcases included: 

Since the base contracts used in above presets, already have a full fledged test cases to cover the entire behavior, Only bare minimum test cases which help in establishing the high level behavior of preset contracts have been added. Again, I referred to test cases written for already added presets and it made sense to proceed this way. 

Thanks and please let me know if there is anything missing or needs to be corrected, would be happy to do it immediately so that we can get this meged! ( PS. Sorry for not mentioning this info about testcases on PR's description ).

frangio (commented):
@ashwinYardi Sorry for taking so long to come back to this!

Thank you for explaining the rationale for what to test. I didn't remember the other presets followed the same approach to testing.

I'd like to see a few more changes in the tests before we merge this:

- Remove the `context` and `describe` "wrappers", since they don't serve any purpose here.
- Remove tests that we can assume correct from the base ERC777 contract:
    - returns a granularity of 1
    - returns 18 when decimals is called
- ERC20: add a test for the total supply

I also think the documentation needs a little bit of work. But this is something we can do if you prefer. If you'd like to take a stab at it you can take a look at the docs comments for the existing presets and follow a similar format. The docs for the constructor are missing too.

In order for the presets to show in the site, you need to add their names in `contracts/presets/README.adoc`.

ashwinYardi:
Hi @frangio ! Thanks a lot for the feedback. 

- [x] Removed the unused context and describe "wrappers"
- [x] Removed the suggested tests that can be assumed correct from base implementation
- [x] Added a test for total supply in ERC20 preset
- [x] Added a test for decrementing the total supply when tokens are burnt in ERC20 preset
- [x] Gave a shot at improving documentation
- [x] Added required entries in `contracts/presets/README.adoc`

frangio (approved):
Great work @ashwinYardi! I made a few small changes to the docs, added a changelog entry, and I realized I prefer to keep the `defaultOperators` argument in the same place as the `ERC777` constructor so I changed that as well.

PR #2401, opened 2020-11-06
Fix ERC1155 URI example

abcoathup:
The current ERC1155 example has an invalid URI for ID substitution.
```solidity
ERC1155("https://game.example/api/item/{1}.json")
```

This should be
```solidity
ERC1155("https://game.example/api/item/{id}.json")
```


frangio (approved):
Thank you!

PR #2403, opened 2020-11-10
Add guide about OZ Contracts Upgrade Safe

frangio:
Note for anyone reading early: the package mentioned in this guide has not been released yet.

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -1,5 +1,8 @@
 * xref:index.adoc[Overview]
-* xref:upgrade-safe.adoc[Upgrade Safety]
+* xref:extending-contracts.adoc[Extending Contracts]
+* xref:upgrade-safe.adoc[Using with Upgrades]
+
+* xref:releases-stability.adoc[Releases & Stability]
 
 * xref:access-control.adoc[Access Control]
 
@@ -13,8 +16,4 @@
 * xref:gsn.adoc[Gas Station Network]
 ** xref:gsn-strategies.adoc[Strategies]
 
-* xref:extending-contracts.adoc[Extending Contracts]
-
 * xref:utilities.adoc[Utilities]
-
-* xref:releases-stability.adoc[Releases & Stability]
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -1,4 +1,5 @@
 * xref:index.adoc[Overview]
+* xref:upgrade-safe.adoc[Upgrade Safety]
 
 * xref:access-control.adoc[Access Control]
 
--- a/docs/modules/ROOT/pages/upgrade-safe.adoc
+++ b/docs/modules/ROOT/pages/upgrade-safe.adoc
@@ -35,9 +35,19 @@ Constructors are replaced by internal initializer functions following the naming
      }
 ```
 
-[CAUTION]
-====
-Use with multiple inheritance requires special care. Initializer functions are not linearized by the compiler like constructors. Because of this, each `+__{ContractName}_init+` function embeds the linearized calls to all parent initializers. As a consequence, calling two of these `init` functions can potentially initialize the same contract twice.
+CAUTION: Use with multiple inheritance requires special attention. See the section below titled <<multiple-inheritance>>.
+
+== Further Notes
+
+[[multiple-inheritance]]
+=== Multiple Inheritance
+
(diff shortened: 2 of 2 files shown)


docs/modules/ROOT/nav.adoc
```diff
@@ -1,4 +1,5 @@
 * xref:index.adoc[Overview]
+* xref:upgrade-safe.adoc[Upgrade Safety]
```
abcoathup, line 2: We could change the navigation order to have all the general guides be at the top.
We could even have the pages be sub pages of Overview.
* Overview
  * Extending Contracts
  * Upgrade Safety
  * Releases & Stability

docs/modules/ROOT/pages/upgrade-safe.adoc
```diff
@@ -0,0 +1,43 @@
+= Upgrade Safety
+
+If your contract is going to be deployed with upgradeability, such as using the OpenZeppelin Upgrades Plugins, you will need to use the Upgrade Safe variant of OpenZeppelin Contracts.
```
abcoathup, line 3: Add link to Upgrades Plugins
```suggestion
If your contract is going to be deployed with upgradeability, such as using the xref:upgrades-plugins::index.adoc[OpenZeppelin Upgrades Plugins], you will need to use the Upgrade Safe variant of OpenZeppelin Contracts.
```

docs/modules/ROOT/pages/upgrade-safe.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+This variant is available as separate package called `@openzeppelin/contracts-upgrade-safe`, which is hosted in the repository https://github.com/OpenZeppelin/openzeppelin-contracts-upgrade-safe[OpenZeppelin/openzeppelin-contracts-upgrade-safe].
```
abcoathup, line 5:
```suggestion
This variant is available as a separate package called `@openzeppelin/contracts-upgrade-safe`, which is hosted in the repository https://github.com/OpenZeppelin/openzeppelin-contracts-upgrade-safe[OpenZeppelin/openzeppelin-contracts-upgrade-safe].
```

docs/modules/ROOT/pages/upgrade-safe.adoc
```diff
@@ -0,0 +1,43 @@
...
+
+It follows all of the rules for xref:upgrades-plugins::writing-upgradeable.adoc[Writing Upgradeable Contracts]: constructors are replaced by initializer functions, state variables are initialized in initializer functions, and we additionally make sure there are no storage incompatibilities across versions.
```
abcoathup, line 7: We may have storage incompatibilities, so some upgrades won't be possible e.g. 2.x to 3.x
```suggestion
It follows all of the rules for xref:upgrades-plugins::writing-upgradeable.adoc[Writing Upgradeable Contracts]: constructors are replaced by initializer functions, state variables are initialized in initializer functions, and we additionally check for storage incompatibilities across versions.
```

docs/modules/ROOT/pages/upgrade-safe.adoc
```diff
@@ -0,0 +1,43 @@
...
+```diff
+-    constructor() ERC721("MyCollectible", "MCO") public {
++    function initialize() initializer public {
++        __ERC721_init("MyCollectible", "MCO");
+     }
+```
+
+[CAUTION]
+====
+Use with multiple inheritance requires special care. Initializer functions are not linearized by the compiler like constructors. Because of thus, each `+__{ContractName}_init+` function embeds the linearized calls to all parent initializers. As a consequence, calling two of these `init` functions can potentially initialize the same contract twice.
```
abcoathup, line 40:
```suggestion
Use with multiple inheritance requires special care. Initializer functions are not linearized by the compiler like constructors. Because of this, each `+__{ContractName}_init+` function embeds the linearized calls to all parent initializers. As a consequence, calling two of these `init` functions can potentially initialize the same contract twice.
```

abcoathup, line 40: We should add a section explaining what `__gap` is used for:

>__gap is an empty storage gap put in each Upgrade Safe contract. This allows in future versions new storage variables to be added to a contract.
>It isn't safe to simply add a storage variable because it "shifts down" all of the storage variables down the inheritance chain. This makes the storage layouts incompatible, (see: https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable#modifying-your-contracts). The number is calculated so that the size of the storage used by a contract always adds up to the same number (e.g. 100).

abcoathup, line 40: We should add a section on using in the same project as contracts extending from OpenZeppelin Contracts.

abcoathup (changes requested):

PR #2411, opened 2020-11-20
Add a beacon proxy contract

frangio:
This PR adds a beacon proxy inspired by Dharma's. The code was initially developed by @f8n in https://github.com/f8n/openzeppelin-sdk/pull/6. I've changed a few things on top of that code:

- Renamed contracts and rewrote revert reasons to conform to our conventions.
- Edited docs and wrote what was missing.
- Used `Address.functionDelegateCall` instead of custom code.
- Adapted tests to run in our test suite.


contracts/proxy/README.adoc
```diff
@@ -9,6 +9,8 @@ The abstract {Proxy} contract implements the core delegation functionality. If t
 
 Upgradeability is implemented in the {UpgradeableProxy} contract, although it provides only an internal upgrade interface. For an upgrade interface exposed externally to an admin, we provide {TransparentUpgradeableProxy}. Both of these contracts use the storage slots specified in https://eips.ethereum.org/EIPS/eip-1967[EIP1967] to avoid clashes with the storage of the implementation contract behind the proxy.
 
+An alternative upgradeability mechanism is provided in <<UpgradeableBeacon>>. This is a pattern that was popularized by Dharma, whereby the proxy contract doesn't hold the implementation address in storage like {UpgradeableProxy}, but the address of a {UpgradeableBeacon} contract, which is where the implementation address is actually stored and retrieved from. The reason a project may want to use this pattern is that multiple proxies can be upgraded in a single transaction if they all share the same beacon.
```
spalladino, line 12:
```suggestion
An alternative upgradeability mechanism is provided in <<UpgradeableBeacon>>. This pattern, popularized by Dharma, allows multiple proxies to be upgraded to a different implementation in a single transaction. In this pattern, the proxy contract doesn't hold the implementation address in storage like {UpgradeableProxy}, but the address of a {UpgradeableBeacon} contract, which is where the implementation address is actually stored and retrieved from. The `upgrade` operations that change the implementation contract address are then sent to the beacon instead of to the proxy contract, and all proxies that follow that beacon are automatically upgraded.
```

Rephrased to move the motivation to the beginning of the paragraph.

contracts/proxy/UpgradeableBeacon.sol
```diff
@@ -0,0 +1,64 @@
...
+     * - msg.sender must be the owner of the contract.
+     * - `newImplementation` must be a contract.
+     */
+    function upgradeTo(address newImplementation) public onlyOwner {
+        _setImplementation(newImplementation);
+        emit Upgraded(newImplementation);
+    }
+
+    /**
+     * @dev Sets the
```
spalladino, line 54:
```suggestion
     * @dev Sets the implementation contract address for this beacon
```

test/proxy/BeaconProxy.test.js
```diff
@@ -0,0 +1,493 @@
...
+
+        expect(value).to.equal(true);
+      });
+    });
+
+    describe('upgrade', function () {
+      describe('when the sender is the admin', function () {
+        const from = proxyCreator;
+
+        describe('when the given implementation is different from the current one', function () {
```
spalladino, line 293: What if it is the same?

test/proxy/BeaconProxy.test.js
```diff
@@ -0,0 +1,493 @@
...
+      });
+
+      it('UpgradeableBeacon has the correct implementation', async function () {
+        const implementation = await this.beacon.implementation();
+        expect(implementation).to.be.equal(this.instance3.address);
+      });
+
+      it('should remove function from proxy 1', async function () {
+        const instance2 = await new Implementation2(this.proxy1.address);
+        await expectRevert.unspecified(instance2.getValue());
```
spalladino, line 484: Let's add a test for a successful operation. I don't feel confident that an upgrade succeeded if the only thing we're testing is an unspecified revert.

spalladino (commented):
LGTM @frangio! I added a few comments only.

test/proxy/BeaconProxy.test.js
frangio: Well, it will be upgraded too. :slightly_smiling_face: I don't really see the need to test these two scenarios separately. What do you think?

spalladino: Neither do I, but then it doesn't make sense to have this `describe` block if we are not intending to test this as a separate path.

spalladino (approved):
Great work with the test suite, looks much better now!

PR #2415, opened 2020-11-27
Update TimelockController docs

nventuro:
Hi! I noticed there was some documentation missing regarding usage of `updateDelay`: this PR tries to improve this.

I also think it'd be good to expand on the docs on the importance of the minimum delay (the entire notion of the timelock is useless without it), but I didn't add anything to that effect.

### DIFF
--- a/contracts/access/TimelockController.sol
+++ b/contracts/access/TimelockController.sol
@@ -277,8 +277,8 @@ contract TimelockController is AccessControl {
      *
      * Requirements:
      *
-     * - the caller must be the timelock itself. This can only be achieved by executing an operation
-     * where the timelock is the target and the data is the ABI-encoded call to this function.
+     * - the caller must be the timelock itself. This can only be achieved by scheduling and later executing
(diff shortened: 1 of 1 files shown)


contracts/access/TimelockController.sol
```diff
@@ -269,9 +271,14 @@ contract TimelockController is AccessControl {
...
     /**
-     * @dev Changes the timelock duration for future operations.
+     * @dev Changes the minimum timelock duration for future operations.
      *
      * Emits a {MinDelayChange} event.
+     *
+     * Requirements:
+     *
+     * - the caller must be the timelock itself. This can only be achieved by executing an operation
+     * where the timelock is the target and the data is the ABI-encoded call to this function.
```
frangio, line 281:
```suggestion
     * - the caller must be the timelock itself. This can only be achieved by scheduling and later executing
     * an operation where the timelock is the target and the data is the ABI-encoded call to this function.
```

frangio (commented):
Good catch. Thank you!

Issue #2417, opened 2020-11-30
Remove Visibility from abstract contracts


pitieu:
According to latest [changes since 0.7.0](https://docs.soliditylang.org/en/v0.7.0/070-breaking-changes.html#functions-and-events)

> Visibility (public / external) is not needed for constructors anymore: To prevent a contract from being created, it can be marked abstract. This makes the visibility concept for constructors obsolete.

Removing Visibility from abstract contracts would fix some warning messages I get when compiling using latest v0.7.5

frangio:
Hi @pitieu, this is a known issue and we haven't removed the warning in the package because it's a breaking change for users of 0.6.

As I mentioned in the [announcement](https://forum.openzeppelin.com/t/openzeppelin-contracts-3-3/4804):
> This release can be used with both Solidity 0.6 and 0.7. However, developers who compile with 0.7 will see warnings originating in the constructors of the library. This is caused by a new warning in Solidity 0.7, that we can't get rid of without causing a breaking change for people who continue using version 0.6 of the compiler.
>
>**In order to compile without warnings, you can install the `solc-0.7` tag on npm: `@openzeppelin/contracts@solc-0.7`.**

We hope to resolve this definitively soon.

PR #2420, opened 2020-12-01
Update ERC20.sol

jdmedeiros:
As of v0.7.0 visibility (public / external) is not needed for constructors anymore: To prevent a contract from being created, it can be marked abstract. This makes the visibility concept for constructors obsolete. Ref.: https://docs.soliditylang.org/en/v0.7.0/070-breaking-changes.html#functions-and-events

Fixes #

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -52,7 +52,7 @@ contract ERC20 is Context, IERC20 {
      * All three of these values are immutable: they can only be set once during
      * construction.
      */
-    constructor (string memory name_, string memory symbol_) public {
+    constructor (string memory name_, string memory symbol_) {
         _name = name_;
         _symbol = symbol_;
         _decimals = 18;


frangio:
Thank you for the suggestion @jdmedeiros. Please take a look at my comment in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2417#issuecomment-735809899:

>This is a known issue and we haven't removed the warning in the package because it's a breaking change for users of 0.6.
>
>As I mentioned in the [announcement](https://forum.openzeppelin.com/t/openzeppelin-contracts-3-3/4804):
>> This release can be used with both Solidity 0.6 and 0.7. However, developers who compile with 0.7 will see warnings originating in the constructors of the library. This is caused by a new warning in Solidity 0.7, that we can't get rid of without causing a breaking change for people who continue using version 0.6 of the compiler.
>>
>>**In order to compile without warnings, you can install the `solc-0.7` tag on npm: `@openzeppelin/contracts@solc-0.7`.**
>
>We hope to resolve this definitively soon.

PR #2424, opened 2020-12-03
add ERC1271 wrapper

### LINKED ISSUE #1104: Facilities for smart accounts or identity contracts
Smart accounts came up in https://github.com/OpenZeppelin/openzeppelin-solidity/issues/1005. They came up in the context of `SignatureBouncer` because the contract is designed to work with ECDSA signatures by externally owned accounts, and thus doesn't support smart accounts. I proposed that instead of changing `SignatureBouncer` we first look into generalizing `ECRecovery` or building an intermediate abstraction.

It seems to me like we need a function `isSignedBy(bytes message, address account, bytes signature) returns (bool)` that does ECDSA verification for EOAs, and calls `account.isValidSignature(signature, message)` for contracts.

I started going through all the identity ERCs to see if there is a function like `isValidSignature` already proposed... but there are a lot of related ERCs and I got a bit lost. I don't think I saw anything like this function though. I'd appreciate some thoughts from someone more interiorized in this space. Pinging @PhABC since he brought it up initially.

I'm nore sure what facilities other than this we could provide.

thegostep:
Fixes #1104

### DIFF
--- a/contracts/access/ERC1271.sol
+++ b/contracts/access/ERC1271.sol
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity >=0.6.0 <0.8.0;
+
+import {ECDSA} from "../cryptography/ECDSA.sol";
+import {Ownable} from "./Ownable.sol";
+
+interface IERC1271 {
+    function isValidSignature(bytes32 _messageHash, bytes memory _signature)
+        external
+        view
+        returns (bytes4 magicValue);
+}
+
+/// @title ERC1271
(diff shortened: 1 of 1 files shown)


frangio:
Thanks @thegostep. We accidentally opened a duplicate PR about the same thing which is more fully formed. See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2459. Closing this in favor of it.

alfredolopez80:
the Best implementation of this Standard Base on Open Zeppelin Library is in the Smart Contract ver 4.1
https://docs.openzeppelin.com/contracts/4.x/api/utils#SignatureChecker

PR #2438, opened 2020-12-14
Use bytes1 and avoid using the obsolete byte type

axic:
`byte` is (and always has been) an alias of `bytes1`, but it provides no benefit over it (perhaps just confusion) and `utils/Create2.sol` already follows this pattern.


frangio (approved):
Thanks!

axic:
Thanks, that was extreme quick to merge! :wink:

PR #2439, opened 2020-12-14
Introduce explicit conversions between addresses and numbers.

axic:
This is valid on <0.8.0, but will be required on >=0.8.0.

Co-authored-by: cameel <cameel2@gmail.com>


frangio (approved):
Thanks!

PR #2442, opened 2020-12-17
Update contracts to support Solidity 0.8.x

### LINKED ISSUE #2441: Support for Solidity 0.8.x
**🧐 Motivation**
Solidity 0.8.0 has been [released](https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/), and I think it would be a good idea to have openzeppelin-contracts supporting it.

**📝 Details**
Updating the current contracts to support being compiled with solc-0.8 should be pretty straightforward since most breaking changes can be easily addressed.

andrekorol:
Fixes #2441 

Contracts' `solidity` version pragmas were updated to support Solidity 0.8.x.  

Changes include:  

* Updated explicit type conversions that conform with the new restrictions.
* Functions that rely on `chainid` in inline assembly are now `view` instead of `pure`.

**Important Note:** Some tests are still not passing because since 0.8.0 failing assertions and other internal checks like division by zero or arithmetic overflow do not use the invalid opcode, but instead the revert opcode. And I believe that this is something worth discussing how to handle before making any changes to the current test cases.


contracts/access/TimelockController.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity >=0.6.9 <0.8.0;
+pragma solidity >=0.6.9 <0.9.0;
 pragma experimental ABIEncoderV2;
```
thegostep, line 4: can remove `pragma experimental ABIEncoderV2`

contracts/GSN/GSNRecipientERC20Fee.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity >=0.6.0 <0.8.0;
+pragma solidity >=0.6.0 <0.9.0;
 
 import "./GSNRecipient.sol";
 import "../math/SafeMath.sol";
```
aaronovz1, line 6: SafeMath could now be deprecated, right?

contracts/access/TimelockController.sol
andrekorol: I left it so that contracts compiled with solc versions older than 0.8.x still use v2 of the ABI encoder.

contracts/GSN/GSNRecipientERC20Fee.sol
andrekorol: In practice, yes. But I would not remove it here since it's a huge part of the `@openzeppelin/contracts` package and a lot of projects still depend on it.

frangio:
Thanks @andrekorol! We in fact want to remove all uses of `SafeMath` from the library now that checked arithmetic is built into Solidity.

We have to discuss whether we want to remove `SafeMath` entirely, though. It may make sense to deprecate it but still keep the interface for a while for backwards compatibility, and removing the manual checks instead just relying on Solidity's checks. There are some subtleties like the custom error messages that are produced by some of the operations, but that has been reported recently to increase gas cost unnecessarily so I think we will want to remove it.

These changes will mean that Solidity 0.8 support will happen in a new major version of the library. We plan to release a final version of the current major before that happens, so we will merge the migration to Solidity 0.8 into a separate branch.

So I will ask, if you can, to change pragmas to `pragma solidity ^0.8.0;`, remove uses of `SafeMath`, and also to resolve conflicts because I merged a related PR that we had pending since before this one.

andrekorol:
@frangio I can make and commit those changes, no problem! I agree with Solidity 0.8 support coming in a new major version of the library, I don't think it makes much sense to have all the changes you mentioned in a new minor version.

andrekorol:
@frangio I've made the necessary changes. Some tests are still not passing because they expect  custom error messages instead of just `VM Exception while processing transaction: revert`.  

I still didn't make any change to the tests though, I wanted to confirm if the correct approach would be to change the expected custom error messages to the default `VM Exception while processing transaction: revert`.  

Also, the [Strings](https://github.com/andrekorol/openzeppelin-contracts/blob/fix/solc-0.8-%232441/contracts/utils/Strings.sol) utility library seems to be broken now, and I'm still not sure why, the only thing changed was the version pragma. I suppose it's something to do with the new restrictions on type conversions...

andrekorol:
So, apparently the `Strings` library wasn't working because it was not possible to explicitly convert from uint256 to bytes1.  

But with 1abf010190ee067dcee058487b4dd8ab09e201e4, it's working again.

contracts/mocks/EIP712External.sol
```diff
@@ -22,7 +22,7 @@ contract EIP712External is EIP712 {
         require(recoveredSigner == signer);
     }
 
-    function getChainId() external pure returns (uint256 chainId) {
+    function getChainId() external view returns (uint256 chainId) {
         // solhint-disable-next-line no-inline-assembly
         assembly {
             chainId := chainid()
```
axic, line 28: 0.8.0 provides this as a language feature: `block.chainid`

andrekorol: Solved on 768d92ca9c5755af864d90a51ffa2531e334d36d.

lephleg:
Great work, @andrekorol! Thanks for your efforts on this.
Please let me know if you require a spare hand implementing or reviewing anything particular.

frangio:
> I still didn't make any change to the tests though, I wanted to confirm if the correct approach would be to change the expected custom error messages to the default `VM Exception while processing transaction: revert`.

The test helpers have an `expectRevert.unspecified` function that you can use.

contracts/drafts/EIP712.sol
```diff
@@ -96,10 +96,7 @@ abstract contract EIP712 {
...
     }
 
-    function _getChainId() private pure returns (uint256 chainId) {
-        // solhint-disable-next-line no-inline-assembly
-        assembly {
-            chainId := chainid()
-        }
+    function _getChainId() private view returns (uint256 chainId) {
+        chainId = block.chainid;
     }
```
frangio, line 101: We can remove this function now and just use the built in.

contracts/mocks/EIP712External.sol
```diff
@@ -22,10 +22,7 @@ contract EIP712External is EIP712 {
...
     }
 
-    function getChainId() external pure returns (uint256 chainId) {
-        // solhint-disable-next-line no-inline-assembly
-        assembly {
-            chainId := chainid()
-        }
+    function getChainId() external view returns (uint256 chainId) {
+        chainId = block.chainid;
     }
```
frangio, line 27: Same here.

contracts/mocks/ERC20PermitMock.sol
```diff
@@ -10,14 +10,11 @@ contract ERC20PermitMock is ERC20Permit {
...
     }
 
-    function getChainId() external pure returns (uint256 chainId) {
-        // solhint-disable-next-line no-inline-assembly
-        assembly {
-            chainId := chainid()
-        }
+    function getChainId() external view returns (uint256 chainId) {
+        chainId = block.chainid;
     }
```
frangio, line 19: Same here.

frangio (commented):
Thanks @andrekorol. Most of the changes look good.

I'm now having doubts about whether we should do regarding the custom revert reasons on subtraction overflow. Maybe we should keep them. What do other people think about this? Should we keep custom revert reasons such as "insufficient balance"?

Can you also change `SafeMath` to remove our manual checks and rely only on the built in checked arithmetic?

thegostep:
> Should we keep custom revert reasons such as "insufficient balance"?

I generally prefer explicit revert messages now that it is convenient to test for them with waffle.

aaronovz1:
> > Should we keep custom revert reasons such as "insufficient balance"?
> 
> I generally prefer explicit revert messages now that it is convenient to test for them with waffle.

Would custom messages undo the gas savings from using built in arithmetic checker? This is the best part of 0.8 IMO - we wouldn't want to lose that benefit.

frangio:
We have to run some tests and measure the impact. I don't know if there is any special magic in the built in arithmetic checker that would make it more efficient than manual checks.

andrekorol:
> Can you also change `SafeMath` to remove our manual checks and rely only on the built in checked arithmetic?

For sure, I'm on it!

aaronovz1:
> We have to run some tests and measure the impact. I don't know if there is any special magic in the built in arithmetic checker that would make it more efficient than manual checks.

I would have thought it would be cheaper since there are less operations being performed and no extra `require` statement. Perhaps it ends up being the same bytecode in the end :man_shrugging:

andrekorol:
> Great work, @andrekorol! Thanks for your efforts on this.
> Please let me know if you require a spare hand implementing or reviewing anything particular.

@lephleg Help would be appreciated in fixing tests that are failing under coverage (solidity-coverage/ganache) 😄

andrekorol:
> Should we keep custom revert reasons such as "insufficient balance"?

One way to retain custom revert reasons would be to keep functions with different signatures like `sub(uint256 a, uint256 b)` and `sub(uint256 a, uint256 b, string memory errorMessage)`, where in the one with the errorMessage manual checks are performed as before, and the actual operation is performed inside an `unchecked` block.

lephleg:
> @lephleg Help would be appreciated in fixing tests that are failing under coverage (solidity-coverage/ganache) 😄

At first glance it seems there's a memory leak introduced in node. Tricky one, I'll try to give it a shot with clear head tomorrow.

axic:
> Can you also change SafeMath to remove our manual checks and rely only on the built in checked arithmetic?

@frangio does it actually make sense to keep SafeMath as a weird wrapper? It adds no benefit, and the "special error message" versions are removed in this PR.

Perhaps it is better biting the bullet by removing it, and that could trigger users to review the new semantics and use them properly (checked vs. unchecked). And also be aware of the gas implications.

andrekorol:
> > Can you also change SafeMath to remove our manual checks and rely only on the built in checked arithmetic?
> 
> 
> 
> @frangio does it actually make sense to keep SafeMath as a weird wrapper? It adds no benefit, and the "special error message" versions are removed in this PR.
> 
> 
> 
> Perhaps it is better biting the bullet by removing it, and that could trigger users to review the new semantics and use them properly (checked vs. unchecked). And also be aware of the gas implications.

As I've said in the comment above, it's possible to revert those changes that removed the "special error messages" and have SafeMath overload functions with two different signatures. One would be just that weird wrapper, and the other would perform the manual checks with custom error messages. The actual arithmetic operation would be wrapped in an unchecked block to avoid wasting gas with double-checking.

hantuzun:
Removing SafeMath is tedious but a straightforward refactoring, but its benefit to the ecosystem would worth the effort. I'm against keeping it as a wrapper.

test/token/ERC1155/ERC1155.behavior.js
```diff
@@ -208,7 +208,7 @@ function shouldBehaveLikeERC1155 ([minter, firstTokenHolder, secondTokenHolder,
       });
 
       it('reverts when transferring more than balance', async function () {
-        await expectRevert(
+        await expectRevert.unspecified(
```
axic, line 211: Should https://github.com/OpenZeppelin/openzeppelin-test-helpers/issues/149 be implemented, tests like these would not lose precision as much as in this current PR.

andrekorol:
@hantuzun Thanks, happy holidays to you as well!!

Should I remove SafeMath and the related tests, including cases like the one @axic mentioned on his last review?

axic:
Just to be clear, I am not part of OpenZeppelin, so my suggestions are only suggestions, not an official review 😃

andrekorol:
No problem, thanks for the review anyway!

andrekorol:
@frangio is there a consensus on whether SafeMath should be removed? 

Do you think setting up a poll before making any further changes is a good idea?

Amxx (commented):
Everything looks good to me.

I was expecting a change of the proxies, using the new `fallback (bytes calldata _input) external [payable] returns (bytes memory _output)` signature, but considering the assembly part is in the `_delegate` internal function, It might do more harm then good to embrace this change.

Amxx:
My grain of salt:
I think the current version of `SafeMath` is good, and we should keep it. Its just a wrapper around the solidity built-in op, so it should be optimized trivially ... and it makes old code more easily portable to 0.8.

Also, it can be used to force/enable safemath inside `unchecked {}` blocks:
````
contract Test {
    uint256 constant internal x = 0;
    uint256 constant internal y = 1;
    function a() public pure returns(uint256) { return x - y; }
    function b() public pure returns(uint256) { unchecked { return x - y; } }
    function c() public pure returns(uint256) { return SafeMath.sub(x, y); }
    function d() public pure returns(uint256) { unchecked { return SafeMath.sub(x, y); } }
}
```

Amxx:
Note for the future: Some "low level" functions, such as `Array::findUpperBound` or `String::toString` could probably benefit from using `unchecked` to reduce gas cost.

frangio:
Let's keep `SafeMath` in this PR and discuss its removal separately. I've created an issue dedicated to it, please participate! Go to https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2465.

We have to take a look at what's causing the coverage job to fail. Otherwise the PR seems good to go. Thanks for working on this @andrekorol!

andrekorol:
It was my pleasure!

I think the coverage job failing has something to do with trufflesuite/ganache-core/issues/515.

It should be fixed in [ganache@3.0.0-internal.0](https://github.com/trufflesuite/ganache-core/releases/tag/v3.0.0-internal.0), but I'll take a closer look at it to see if that's really the case.

andrekorol:
Coverage was failing due to an out of memory error, which could be solved by either updating Node's version here to any version >= 14: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/318c4b44eaba2d745ed4a6381c43e03edc53634d/.github/workflows/test.yml#L35

or by increasing the heap memory as I did in 6a30dcc.

Now the coverage job is failing because of six different test cases that are not passing. Two of them (`domain separator`) I assume to be related to that chainId issue.

frangio:
I think you have debugged it correctly, it's likely to be the chainId issue. We may have to migrate to Hardhat since the latest hardhat-coverage plugin uses Hardhat EVM and will probably be fine. We can do that in a separate PR. (@Amxx If you have your Hardhat migration almost ready, I think it would be great to make a PR for the `solc-0.8` branch where this is being merged.)

Something else we will have to do afterwards is review all of the `expectRevert.unspecified` in tests and add back the revert reasons. We will be able to bring them back with the functionality we're adding in #2462.

Thank you everyone! Please remember this is going in the `solc-0.8` branch, and we will do our best to release it as soon as possible as version 4.0 of the library. We are now preparing 3.4 as the last release of the 3.x line.

frangio:
A correction: while doing a last review of the changes in this PR I found that the chain id issue in Ganache already had a workaround and it was being removed here. It fixed the coverage for me locally and hopefully the CI will agree.

PR #2443, opened 2020-12-18
Adjust for chainid becoming view in 0.8.0.

axic:
Borrowed the mutability warning silencing hack from other parts of OpenZeppelin.

I think it makes sense to add this change to pre-0.8.0, just to ensure users of these contracts are aware of the semantic changes related to chainid's mutability. This is practically a bug fix, see https://github.com/ethereum/solidity/issues/10090

### DIFF
--- a/contracts/drafts/EIP712.sol
+++ b/contracts/drafts/EIP712.sol
@@ -96,7 +96,8 @@ abstract contract EIP712 {
         return keccak256(abi.encodePacked("\x19\x01", _domainSeparatorV4(), structHash));
     }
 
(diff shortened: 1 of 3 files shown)


axic:
I am not sure what the plans are regarding zeppelin releases, but if there's a release before switching to 0.8.0, I think this PR should be part of it.

axic:
Checked the actual invocations, which are already view, so I guess in practice this produces no externally visible change. Probably safe to merge then :wink:

frangio (approved):
Thanks! Agree on releasing this in our last release for 0.6/0.7.

PR #2445, opened 2020-12-23
Rename permit arguments to match the spec

ZumZoom:
[EIP-2612](https://eips.ethereum.org/EIPS/eip-2612) has the following interface for permit:
`function permit(address owner, address spender, uint value, uint deadline, uint8 v, bytes32 r, bytes32 s)`

Note that third argument is called `value` which is also reflected in `_PERMIT_TYPEHASH`.

It is probably better to follow the spec and use the same argument names as in EIP and in the typehash.


frangio (approved):
Thanks @ZumZoom. We generally don't follow the exact parameter names found in EIPs, but given that this one also shows up in the type hash I think it makes sense.

PR #2446, opened 2020-12-23
Add available since notice to TimelockController docs

nventuro:
I was wondering why I couldn't find the contracts, until I noticed I was using the wrong release. This should be in the docs, not on the release page (see e.g. the [ERC1155 docs](https://docs.openzeppelin.com/contracts/3.x/api/token/erc1155#IERC1155)).


frangio (approved):
Yes we forgot about this, thanks!

PR #2449, opened 2021-01-04
Add ERC1167 library (minimal proxy)

### LINKED ISSUE #2437: Minimal proxy factory
EIP-1167 specifies a [Minimal Proxy Contract](https://eips.ethereum.org/EIPS/eip-1167). It's just a bytecode blob and it would be convenient to wrap it in a factory contract or library. A few people have asked for this.

Consider also including functionality to use create2 for deterministic addresses. See the [`ProxyFactory`](https://github.com/OpenZeppelin/openzeppelin-sdk/blob/7d96de7248ae2e7e81a743513ccc617a2e6bba21/packages/lib/contracts/upgradeability/ProxyFactory.sol) contract in OpenZeppelin SDK for reference.

Amxx:
Proposal, to be discussed, for an ERC1167 ProxyFactory.

Fixes #2437

### DIFF
--- a/contracts/drafts/Clones.sol
+++ b/contracts/drafts/Clones.sol
@@ -4,7 +4,7 @@ pragma solidity >=0.6.0 <0.8.0;
 
 /**
  * @dev https://eips.ethereum.org/EIPS/eip-1167[EIP 1167] is a standard for
- * deploying minimal clone contracts.
+ * deploying minimal proxy contracts, also known as "clones".
  *
  * This contract provide tooling to deploy proxies following the EIP 1167
  * proposed bytecode. This is possible using both create and create2.
--- a/contracts/drafts/Clones.sol
+++ b/contracts/drafts/Clones.sol
@@ -9,8 +9,7 @@ pragma solidity >=0.6.0 <0.8.0;
  * This contract provide tooling to deploy proxies following the EIP 1167
  * proposed bytecode. This is possible using both create and create2.
  */
-library Clones
-{
+library Clones {
     function clone(address master) internal returns (address instance) {
         // solhint-disable-next-line no-inline-assembly
         assembly {
--- a/contracts/drafts/Clones.sol
+++ b/contracts/drafts/Clones.sol
@@ -6,6 +6,9 @@ pragma solidity >=0.6.0 <0.8.0;
  * @dev https://eips.ethereum.org/EIPS/eip-1167[EIP 1167] is a standard for
  * deploying minimal proxy contracts, also known as "clones".
  *
+ * > To simply and cheaply clone contract functionality in an immutable way, this standard specifie
+ * > a minimal bytecode implementation that delegates all calls to a known, fixed address.
+ *
  * This contract provide tooling to deploy proxies following the EIP 1167
  * proposed bytecode. This is possible using both create and create2.
  */
--- a/contracts/drafts/Clones.sol
+++ b/contracts/drafts/Clones.sol
@@ -24,7 +24,7 @@ library Clones {
         }
     }
 
-    function clone2(address master, bytes32 salt) internal returns (address instance) {
+    function cloneDeterministic(address master, bytes32 salt) internal returns (address instance) {
         // solhint-disable-next-line no-inline-assembly
         assembly {
             let ptr := mload(0x40)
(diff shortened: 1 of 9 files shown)


contracts/drafts/Clones.sol
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.6.0 <0.8.0;
+
+/**
+ * @dev https://eips.ethereum.org/EIPS/eip-1167[EIP 1167] is a standard for
+ * deploying minimal clone contracts.
```
frangio, line 7:
```suggestion
 * deploying minimal proxy contracts, also known as "clones".
```

Following the name of the EIP.

contracts/drafts/Clones.sol
```diff
@@ -0,0 +1,53 @@
...
+ *
```
frangio, line 8: Copied for the EIP.

```suggestion
 *
 * > To simply and cheaply clone contract functionality in an immutable way, this standard specifie
 * > a minimal bytecode implementation that delegates all calls to a known, fixed address.
 *
```

contracts/drafts/Clones.sol
```diff
@@ -0,0 +1,53 @@
...
+        assembly {
+            let ptr := mload(0x40)
+            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
+            mstore(add(ptr, 0x14), shl(0x60, master))
+            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+            instance := create(0, ptr, 0x37)
+        }
+    }
+
+    function clone2(address master, bytes32 salt) internal returns (address instance) {
```
frangio, line 24: I would prefer a name like the following:

```suggestion
    function cloneDeterministic(address master, bytes32 salt) internal returns (address instance) {
```

contracts/drafts/Clones.sol
```diff
@@ -0,0 +1,53 @@
...
+            let ptr := mload(0x40)
+            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
+            mstore(add(ptr, 0x14), shl(0x60, master))
+            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+            instance := create2(0, ptr, 0x37, salt)
+        }
+        require(instance != address(0), "ERC1167: create2 failed");
+    }
+
+    function predict2(address master, bytes32 salt, address deployer) internal pure returns (address predicted) {
```
frangio, line 36: And here maybe `predictAddress`.

frangio (commented):
The library looks good, we need some tests for it. The EIP is Final so we can move it to the `proxy` directory.

contracts/drafts/Clones.sol
Amxx: What about `predictDeterministicAddress` ?

frangio: Sounds good!

Amxx:
> The library looks good, we need some tests for it. The EIP is Final so we can move it to the `proxy` directory.

I moved this to contracts/proxy and added tests

contracts/proxy/Clones.sol
```diff
@@ -13,6 +13,11 @@ pragma solidity >=0.6.0 <0.8.0;
  * proposed bytecode. This is possible using both create and create2.
  */
 library Clones {
+    /**
+     * @dev Deploys and return the address of a clone that mimics the behaviour of `master`.
+     *
+     * This function uses the create opcode, which should never revert.
```
frangio, line 19: I just realized that this could potentially revert in an out of gas situation. I guess this means we should revert if `instance` is 0.

Amxx: In out of gas, would the instance be set to 0 (which we should check) ? I was under the impression that the call would just revert on its own

frangio:
I moved your content around in the docs to keep it either at the top of the proxies page or inside the contract description.

I know that we have reservations with this content organization but let's stick to the current thing until we review it.

frangio (approved):
I think this looks good!

PR #2450, opened 2021-01-04
Adding support for short signature (eip-2098) in cryptography/ECDSA

Amxx:
[EIP-2098](https://eips.ethereum.org/EIPS/eip-2098) proposes a way to use short, 64 bytes long, signatures instead of the traditional 65 bytes long ones.

This PR updates cryptography/ECDSA to support both 65 (r,s,v) and 64(r,vs) bytes signatures.

### DIFF
--- a/contracts/cryptography/ECDSA.sol
+++ b/contracts/cryptography/ECDSA.sol
@@ -24,23 +24,34 @@ library ECDSA {
      * be too long), and then calling {toEthSignedMessageHash} on it.
      */
     function recover(bytes32 hash, bytes memory signature) internal pure returns (address) {
-        // Check the signature length
-        if (signature.length != 65) {
-            revert("ECDSA: invalid signature length");
-        }
-
         // Divide the signature in r, s and v variables
         bytes32 r;
         bytes32 s;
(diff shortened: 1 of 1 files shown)


frangio:
My concern about this is that the EIP is in Draft status. Draft EIPs are problematic for us because their API can change breaking backwards compatibility, so in the past we have either rejected these or published them in the `drafts` directory. We have a `drafts` directory now, with the additional requirement that the EIPs we include there should have been audited (spec and implementation) and used in production. This serves to filter the EIPs that we will accept to an amount that is manageable with our resources, and also has stronger guarantees to be sound, useful, and secure.

So we have to run this tests for this EIP. I'm not familiar with it. Do you know if it's used in production and if it has been audited?

Amxx:
I know that ether includes tooling for it, but as @ricmoo is the ERC proposer ... I am not aware of any app using it in production, I am not aware of any audit.

Just looking at the math, it is easy to produce a valid long signature from a valid short signature. So if this format was malleable, then so would the long format (is is assumed not to be).

I'll point @ricmoo to here, see if he has more information.

PR #2453, opened 2021-01-05
Move Context from GSN to utils directory

Amxx:
Proposal to move the `Context` contract to the `contracts/utils` folder


frangio:
To provide some context: we want to move this file out of the GSN directory to make it clear that Context itself has nothing to do with GSN. It enables opting in to GSN if a developer wants to do so.

frangio (commented):
Thanks!

For backwards compatibility we need to keep a file `contracts/GSN/Context.sol` that just imports the file from the new location:
```solidity

pragma solidity >=0.6.0 <0.8.0;

import '../utils/Context.sol';
```

Issue #2455, opened 2021-01-06
PaymentSplitter.release uses address.transfer


Amxx:
PaymentSplitter.release relies on the built in transfer function to sent Eth to the payee. This is bad practice, since some address need more then 2300 gas to receive funds, and should be replaced by Address:sendValue

frangio:
Fixed by #2456.

PR #2456, opened 2021-01-06
Use Address.sendValue for PaymentSplitter

Amxx:
Fix for issue #2455


frangio:
This made me notice that `sendValue` does not bubble up the revert reason. We should fix that as well.

PR #2459, opened 2021-01-07
Tooling to verify signatures with support for ERC1271

Amxx:
SignatureChecker is a library for seamless verification of signature with support for both EOA (ECDSA) and smart contract wallets (ERC1271).

While still being a draft, this ERC is supported by smart contract wallets such as Argent.

I believe this feature could greatly improve the compatibility of dapps that relies message signatures with such wallets. In particular since Argent is now able to sign message (including ERC721 typedData) through WalletConnect.

### DIFF
--- a/contracts/interfaces/IERC1271.sol
+++ b/contracts/interfaces/IERC1271.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity >=0.6.0 <0.8.0;
+pragma solidity >=0.6.0 <0.9.0;
 
 interface IERC1271 {
   /**
--- a/contracts/interfaces/IERC1271.sol
+++ b/contracts/interfaces/IERC1271.sol
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.6.0 <0.8.0;
+
+interface IERC1271 {
+  /**
+   * @dev Should return whether the signature provided is valid for the provided data
+   * @param hash      Hash of the data to be signed
+   * @param signature Signature byte array associated with _data
+   */
+  function isValidSignature(bytes32 hash, bytes memory signature) external view returns (bytes4 magicValue);
+}
--- a/contracts/mocks/ERC1271WalletMock.sol
+++ b/contracts/mocks/ERC1271WalletMock.sol
@@ -1,13 +1,13 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity >=0.6.0 <0.8.0;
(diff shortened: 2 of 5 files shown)


frangio:
One thing I would like to see if we ship this, is documentation outlining how EIP-1271 signatures can be different from normal cryptographic signatures. In particular, the property of non-repudiation is compromised on-chain because a signature that was valid at some point can stop being valid at a later point, according to the "signing" contract. Of course off-chain one could look at the history of the chain but other contracts can't do that.

This is a limitation of the EIP and we should simply document it.

thegostep:
here is my implementation if useful:

```solidity
// SPDX-License-Identifier: GPL-3.0-only
pragma solidity 0.7.6;

import {ECDSA} from "@openzeppelin/contracts/cryptography/ECDSA.sol";
import {Address} from "@openzeppelin/contracts/utils/Address.sol";

interface IERC1271 {
    function isValidSignature(bytes32 _messageHash, bytes memory _signature)
        external
        view
        returns (bytes4 magicValue);
}

/// @title ERC1271
/// @notice Module for ERC1271 compatibility
/// @dev Security contact: dev-support@ampleforth.org
abstract contract ERC1271 is IERC1271 {
    using ECDSA for bytes32;
    using Address for address;

    // Valid magic value bytes4(keccak256("isValidSignature(bytes32,bytes)")
    bytes4 internal constant VALID_SIG = IERC1271.isValidSignature.selector;
    // Invalid magic value
    bytes4 internal constant INVALID_SIG = bytes4(0);

    modifier onlyValidSignature(bytes32 permissionHash, bytes memory signature) {
        require(
            isValidSignature(permissionHash, signature) == VALID_SIG,
            "ERC1271: Invalid signature"
        );
        _;
    }

    function _getOwner() internal view virtual returns (address owner);

    function isValidSignature(bytes32 permissionHash, bytes memory signature)
        public
        view
        override
        returns (bytes4)
    {
        address owner = _getOwner();
        if (owner.isContract()) {
            try IERC1271(owner).isValidSignature(permissionHash, signature) returns (
                bytes4 returnVal
            ) {
                return returnVal;
            } catch {
                return INVALID_SIG;
            }
        } else {
            address signer = permissionHash.recover(signature);
            return signer == owner ? VALID_SIG : INVALID_SIG;
        }
    }
}
```

MatthiasEgli-PwC:
What happened to this PR @Amxx ?

Amxx:
This targeted a branch that got deleted when solc 0.8 support was added to master. Will reopen a new one after 4.0 preview is published.

Is it something you'd like to see in the repo / you'd use ?

PR #2462, opened 2021-01-08
Refactor SafeMath to avoid memory leaks

### LINKED ISSUE #2413: The `SafeMath.div` function needlessly allocates memory
The `SafeMath.div` function needlessly allocates memory.

**💻 Environment**

All environments.

**📝 Details**

Calling `SafeMath.div(uint256, uint256)` allocates memory in order to pass a revert reason to `SafeMath.div(uint256, uint256, string memory)`, even if the operation would not revert. This causes an addition 64-byte allocation per call for the revert reason string `length || paddedContents`.

While in it of itself, this is a minor issue, it is worth noting that the other `SafeMath` methods do not have the same problem, and that in a loop, the gas cost of `SafeMath.div` grows quadratically (because of memory costs) instead of linearly with the number of iterations as one would naturally expect.

**🔢 Code to reproduce bug**

The easiest way to reproduce this is to use the following code in Remix and calling `doThings` with varying lengths of `things` input:
```solidity
// SPDX-License-Identifier: CC0-1.0
pragma solidity ^0.6.0;

import "https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v3.2.0/contracts/math/SafeMath.sol";

contract SafeMathDivMemoryTest {
    using SafeMath for uint256;
    
    function doManyThings(uint256[] calldata things) external pure returns (uint256[] memory result, uint256 mem) {
        result = new uint256[](things.length);
        
        assembly {
            mem := mload(0x40)
        }
        for (uint256 i = 0; i < things.length; i++) {
            result[i] = doThing(things[i]);
        }
        assembly {
            mem := sub(mload(0x40), mem)
        }
    }
    
    function doThing(uint256 thing) internal pure returns (uint256) {
        // return thing / 2;
        return thing.div(2);
    }
}
```

The call to `doThing` allocates 64 bytes per call, because of the `div` call. Replacing `thing.div(2)` with `thing / 2` reduces memory allocation to 0 as expected.

Amxx:
FIxes #2413, remove the string allocation in SafeMath  default functions: caused unrecovered memory allocation even for successfull calls

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
  * `Context`: moved from `contracts/GSN` to `contracts/utils`. ([#2453](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2453))
  * `PaymentSplitter`: replace usage of `.transfer()` with `Address.sendValue` for improved compatibility with smart wallets. ([#2455](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2455))
  * `UpgradeableProxy`: bubble revert reasons from initialization calls. ([#2454](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2454))
- * `SafeMath`: fix a memory allocation issue by adding new `SafeMath.tryXXX(uint,uint)→(bool,uint)` function. `SafeMath.XXX(uint,uint,string)→(uint)` are not deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
+ * `SafeMath`: fix a memory allocation issue by adding new `SafeMath.tryOp(uint,uint)→(bool,uint)` functions. `SafeMath.op(uint,uint,string)→uint` are now deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
  * `EnumerableMap`: fix a memory allocation issue by adding new `EnumerableMap.tryGet(uint)→(bool,address)` function. `SafeMath.get(uint)→(string)` is now deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
 
 ## 3.3.0 (2020-11-26)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,11 @@
  * `ERC20Permit`: added an implementation of the ERC20 permit extension for gasless token approvals. ([#2237](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2237))
  * Presets: added token presets with preminted fixed supply `ERC20PresetFixedSupply` and `ERC777PresetFixedSupply`. ([#2399](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2399))
  * `Address`: added `functionDelegateCall`, similar to the existing `functionCall`. ([#2333](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2333))
- * `Context`: moved from `contracts/GSN` to `contracts/utils`. ([#2453](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2453)) 
+ * `Context`: moved from `contracts/GSN` to `contracts/utils`. ([#2453](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2453))
  * `PaymentSplitter`: replace usage of `.transfer()` with `Address.sendValue` for improved compatibility with smart wallets. ([#2455](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2455))
- * `UpgradeableProxy`: bubble revert reasons from initialization calls. ([#2454](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2454)) 
+ * `UpgradeableProxy`: bubble revert reasons from initialization calls. ([#2454](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2454))
+ * `SafeMath`: fix a memory allocation issue by adding new `SafeMath.tryXXX(uint,uint)→(bool,uint)` function. `SafeMath.XXX(uint,uint,string)→(uint)` are not deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
+ * `EnumerableMap`: fix a memory allocation issue by adding new `EnumerableMap.tryGet(uint)→(bool,address)` function. `SafeMath.get(uint)→(string)` is now deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
 
 ## 3.3.0 (2020-11-26)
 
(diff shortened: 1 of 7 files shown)


frangio:
After discussing offline we agreed on the following:

- We will deprecate the existing functions like `sub(uint, uint, string)` because they leak memory. (This is adding a comment in the source code and in the release notes.)
- We will implement `sub(uint, uint)` by directly calling `require` so it does not leak memory.
- We will add a new family of functions that return a pair `(bool, uint)` that can be used to protect against overflow with custom revert reasons, and use these internally in the library.

CHANGELOG.md
```diff
@@ -7,9 +7,11 @@
...
  * `Address`: added `functionDelegateCall`, similar to the existing `functionCall`. ([#2333](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2333))
- * `Context`: moved from `contracts/GSN` to `contracts/utils`. ([#2453](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2453)) 
+ * `Context`: moved from `contracts/GSN` to `contracts/utils`. ([#2453](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2453))
  * `PaymentSplitter`: replace usage of `.transfer()` with `Address.sendValue` for improved compatibility with smart wallets. ([#2455](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2455))
- * `UpgradeableProxy`: bubble revert reasons from initialization calls. ([#2454](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2454)) 
+ * `UpgradeableProxy`: bubble revert reasons from initialization calls. ([#2454](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2454))
+ * `SafeMath`: fix a memory allocation issue by adding new `SafeMath.tryXXX(uint,uint)→(bool,uint)` function. `SafeMath.XXX(uint,uint,string)→(uint)` are not deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
```
frangio, line 13: "XXX" is a little weird. :sweat_smile: 

```suggestion
 * `SafeMath`: fix a memory allocation issue by adding new `SafeMath.tryOp(uint,uint)→(bool,uint)` functions. `SafeMath.op(uint,uint,string)→uint` are now deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
```

CHANGELOG.md
```diff
@@ -7,9 +7,11 @@
...
+ * `EnumerableMap`: fix a memory allocation issue by adding new `EnumerableMap.tryGet(uint)→(bool,address)` function. `SafeMath.get(uint)→(string)` is now deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
```
frangio, line 14:
```suggestion
 * `EnumerableMap`: fix a memory allocation issue by adding new `EnumerableMap.tryGet(uint)→(bool,address)` functions. `EnumerableMap.get(uint)→string` is now deprecated. ([#2462](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2462))
```

contracts/math/SafeMath.sol
```diff
@@ -101,48 +123,65 @@ library SafeMath {
...
+    function mod(uint256 a, uint256 b) internal pure returns (uint256) {
+        require(b > 0, "SafeMath: modulo by zero");
+        return a % b;
+    }
 
-        return c;
+    /**
+     * @notice DEPRECATED: causes memory leak
+     * @dev Returns the subtraction of two unsigned integers, reverting with custom message on
+     * overflow (when the result is negative).
```
frangio, line 150: `@notice` is not shown in our docs site at the moment, only `@dev`.

```suggestion
     * @dev Returns the subtraction of two unsigned integers, reverting with custom message on
     * overflow (when the result is negative).
     *
     * CAUTION: This function is deprecated because it requires allocating memory for the error
     * message unnecessarily. For custom revert reasons use {trySub}.
```

contracts/math/SafeMath.sol
```diff
@@ -101,48 +123,65 @@ library SafeMath {
...
+        require(b <= a, errorMessage);
+        return a - b;
     }
 
     /**
-     * @dev Returns the remainder of dividing two unsigned integers. (unsigned integer modulo),
-     * Reverts when dividing by zero.
+     * @notice DEPRECATED: causes memory leak
+     * @dev Returns the integer division of two unsigned integers, reverting with custom message on
+     * division by zero. The result is rounded towards zero.
```
frangio, line 166:
```suggestion
     * @notice DEPRECATED: causes memory leak
     * @dev Returns the integer division of two unsigned integers, reverting with custom message on
     * division by zero. The result is rounded towards zero.
     *
     * CAUTION: This function is deprecated because it requires allocating memory for the error
     * message unnecessarily. For custom revert reasons use {tryDiv}.
```

contracts/math/SafeMath.sol
```diff
@@ -101,48 +123,65 @@ library SafeMath {
...
      */
-    function mod(uint256 a, uint256 b) internal pure returns (uint256) {
-        return mod(a, b, "SafeMath: modulo by zero");
+    function div(uint256 a, uint256 b, string memory errorMessage) internal pure returns (uint256) {
+        require(b > 0, errorMessage);
+        return a / b;
     }
 
     /**
+     * @notice DEPRECATED: causes memory leak
```
frangio, line 182:
```suggestion
```

contracts/math/SafeMath.sol
```diff
@@ -101,48 +123,65 @@ library SafeMath {
...
      * @dev Returns the remainder of dividing two unsigned integers. (unsigned integer modulo),
-     * Reverts with custom message when dividing by zero.
+     * reverting with custom message when dividing by zero.
```
frangio, line 184:
```suggestion
     * reverting with custom message when dividing by zero.
     *
     * CAUTION: This function is deprecated because it requires allocating memory for the error
     * message unnecessarily. For custom revert reasons use {tryMod}.
```

test/math/SafeMath.test.js
```diff
@@ -5,142 +5,396 @@ const { expect } = require('chai');
 
 const SafeMathMock = artifacts.require('SafeMathMock');
 
+async function deepExpect (value, expected) {
```
frangio, line 8: There is technically a problem with this function. It iterates using for..in, but recurses if the value has `Symbol.iterator`. The thing is `Symbol.iterator` is used for for..of loops, whereas for..in iterates over `Object.keys`.

IMO we should have a simpler function to compare the kinds of objects returned by the try functions. `expectEqualWithFlag` and no recursion. Plus `testCommutativeIterable` -> `testCommutativeWithFlag`.

frangio (commented):
We have to add the same deprecation notice for `EnumerableSet.get`. We use it in `ERC721.ownerOf` so while we're at it we should probably change that too.

frangio (approved):
Awesome work thanks!

PR #2464, opened 2021-01-08
Adding an internal _setOwner to Ownable

### LINKED ISSUE #2463: Adding internal function to set the owner of Ownable
Some pattern require an Ownable contract to change its owner under specific condition. This would for example be necessary to implement an `OwnableClaimable` contract for #2369 on top of the existing `Ownable`.

However, the current implmentation of `Ownable` can only be transfered using the public `transferOwnership`, which has the modifier onlyOwner.

I believe it would be better to have two function, one internal that does the change, and one public that gives restricted access to the internal.

```
abstract contract Ownable is Context {
    ...
    function renounceOwnership() public virtual onlyOwner {
        _setOwnership(address(0));
    }

    function transferOwnership(address newOwner) public virtual onlyOwner {
        require(newOwner != address(0), "Ownable: new owner is the zero address");
        _setOwnership(newOwner);
    }

    function _setOwnership(address newOwner) internal {
        emit OwnershipTransferred(_owner, newOwner);
        _owner = newOwner;
    }
}
```

Amxx:
Fixes #2463

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -23,10 +23,8 @@ abstract contract Ownable is Context {
     /**
      * @dev Initializes the contract setting the deployer as the initial owner.
      */
-    constructor () {
-        address msgSender = _msgSender();
-        _owner = msgSender;
-        emit OwnershipTransferred(address(0), msgSender);
+    constructor () internal {
+        _transferOwnership(_msgSender());
     }
 
     /**
@@ -52,7 +50,7 @@ abstract contract Ownable is Context {
      * thereby removing any functionality that is only available to the owner.
      */
     function renounceOwnership() public virtual onlyOwner {
-        _setOwner(address(0));
+        _transferOwnership(address(0));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


contracts/access/Ownable.sol
```diff
@@ -62,6 +59,14 @@ abstract contract Ownable is Context {
...
     function transferOwnership(address newOwner) public virtual onlyOwner {
         require(newOwner != address(0), "Ownable: new owner is the zero address");
+        _transferOwnership(newOwner);
+    }
+
+    /**
+     * @dev Internal function that forces an ownership change. Can be used to
+     * implement custom ownership management logic in childs contracts.
+     */
+    function _transferOwnership(address newOwner) internal virtual {
```
frangio, line 69: We try to avoid situations where a single character in the function name changes the semantics of the function in a significant way. In this case it changes whether the `msg.sender` is required to be the owner so it would be better to use another name.

Maybe `_resetOwner` or `_resetOwnership`, what do you think?

Amxx: the `_transferOwnership` used to exist, but I get your remark. 
What about `_setOwner` ?

frangio: Yes it used to exist, what I mentioned is a newer guideline we came up with. :smiley: 

`_setOwner` sounds fine, though I wonder if this sets a precedent for defining internal setters...

I think I want to put this PR on hold, finish up 3.4 and come back to this afterwards. I don't actually remember people asking for an internal owner setter, except perhaps the Claimable thing that you mentioned. Honestly in my opinion `Ownable` is so trivial that I think `Claimable` should be its own contract instead of derived from `Ownable`.

Amxx: That would indeed set a precedent for internal setters.

Having a `Claimable` extension of `Ownable` would make sens, except this extensions would not be able to set `_owner`. It would therefore have to duplicate code from `Ownable`. I think discoverability would suffer from this.

While I'm ok to discuss that after 3.4, I think having this feature in the 3.x branch would improve usability of this branch for users not willing to move to 0.8.0.

PR #2466, opened 2021-01-08
Implement ERC1404 (extention of ERC20)

Amxx:
Implements https://github.com/ethereum/eips/issues/1404

This EIP is used by [https://etherscan.io/address/erlc.iexec.eth](https://etherscan.io/address/erlc.iexec.eth), and could potentially be usefull to building permissionned tokens (in conjonction with `AccessControl`?)


frangio:
The usual comment about Draft ERCs: we need to check if an audited implementation of the ERC has been deployed to production, then we need to include it in the drafts directory.

PR #2468, opened 2021-01-11
Make non-view functions virtual

### LINKED ISSUE #2414: Make ERC777 non-view methods virtual
Make ERC777's [`send()`](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/token/ERC777/ERC777Upgradeable.sol#L149), [`transfer()`](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/token/ERC777/ERC777Upgradeable.sol#L161), [`burn()`](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/token/ERC777/ERC777Upgradeable.sol#L180), etc `virtual`.

**📝 Details**

As [announced in the forum](https://forum.openzeppelin.com/t/how-we-plan-to-adopt-solidity-0-6-virtual-functions/2063/7), all non-view functions should be made virtual so they can be overriden. I haven't checked if there's any other contracts missing this feature.

Amxx:
Some non-view / non-pure functions are missing the virtual keyword. This PR fixes that.

I also took the occasion to reorder some keyword and fixe some indentation for consistency sake.

Fixes #2414

### DIFF
--- a/contracts/proxy/TransparentUpgradeableProxy.sol
+++ b/contracts/proxy/TransparentUpgradeableProxy.sol
@@ -91,7 +91,7 @@ contract TransparentUpgradeableProxy is UpgradeableProxy {
      *
      * NOTE: Only the admin can call this function. See {ProxyAdmin-changeProxyAdmin}.
      */
-    function changeAdmin(address newAdmin) external ifAdmin {
+    function changeAdmin(address newAdmin) external virtual ifAdmin {
         require(newAdmin != address(0), "TransparentUpgradeableProxy: new admin is the zero address");
         emit AdminChanged(_admin(), newAdmin);
         _setAdmin(newAdmin);
@@ -102,7 +102,7 @@ contract TransparentUpgradeableProxy is UpgradeableProxy {
      *
      * NOTE: Only the admin can call this function. See {ProxyAdmin-upgrade}.
      */
-    function upgradeTo(address newImplementation) external ifAdmin {
+    function upgradeTo(address newImplementation) external virtual ifAdmin {
         _upgradeTo(newImplementation);
     }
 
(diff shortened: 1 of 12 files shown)


contracts/proxy/TransparentUpgradeableProxy.sol
```diff
@@ -67,7 +67,7 @@ contract TransparentUpgradeableProxy is UpgradeableProxy {
      * https://eth.wiki/json-rpc/API#eth_getstorageat[`eth_getStorageAt`] RPC call.
      * `0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103`
      */
-    function admin() external ifAdmin returns (address admin_) {
+    function admin() external virtual ifAdmin returns (address admin_) {
```
frangio, line 70: This and `implementation` are meant to be used as `view` functions, but Solidity doesn't allow to mark them as `view` because of the `ifAdmin` modifier. But given that they are intended as `view` I think I would remove the `virtual` modifier.

contracts/utils/Address.sol
```diff
@@ -34,6 +34,18 @@ library Address {
         return size > 0;
     }
 
+    /**
+     * @dev Returns true if and only if the function is running in the constructor
+     */
+    function isConstructor() internal view returns (bool) {
```
frangio, line 40: This is my only issue with this PR. This function is not really about addresses, it's more about the "lifecycle" of a contract, and it happens to use `Address.isContract` in its implementation.

I would prefer to keep `isConstructor` where it was originally, but using `Address.isContract` instead of the assembly it was using before.

frangio (commented):
Great work thanks!

frangio (approved):
Thanks!

PR #2469, opened 2021-01-12
Add ERC165Checker.getSupportedInterfaces

### LINKED ISSUE #2429: Proposal: ERC165Checker - getSupportedInterfaces
**🧐 Motivation**

ERC165 allows you to check which interfaces are implemented on a contract. OZ has a library `ERC165Checker.sol` with helper methods for checking if some contract implements an interface. First, it checks that ERC165 is implemented and then it checks if the interface is implemented. For batch checking interfaces, there is a separate method `supportsAllInterfaces` which checks if ERC165 is implemented only once, and then checks each interface to ensure they are implemented. Here is the relevant existing code:

```solidity
    /**
     * @dev Returns true if `account` supports the interface defined by
     * `interfaceId`. Support for {IERC165} itself is queried automatically.
     *
     * See {IERC165-supportsInterface}.
     */
    function supportsInterface(address account, bytes4 interfaceId) internal view returns (bool) {
        // query support of both ERC165 as per the spec and support of _interfaceId
        return supportsERC165(account) &&
            _supportsERC165Interface(account, interfaceId);
    }

    /**
     * @dev Returns true if `account` supports all the interfaces defined in
     * `interfaceIds`. Support for {IERC165} itself is queried automatically.
     *
     * Batch-querying can lead to gas savings by skipping repeated checks for
     * {IERC165} support.
     *
     * See {IERC165-supportsInterface}.
     */
    function supportsAllInterfaces(address account, bytes4[] memory interfaceIds) internal view returns (bool) {
        // query support of ERC165 itself
        if (!supportsERC165(account)) {
            return false;
        }

        // query support of each interface in _interfaceIds
        for (uint256 i = 0; i < interfaceIds.length; i++) {
            if (!_supportsERC165Interface(account, interfaceIds[i])) {
                return false;
            }
        }

        // all interfaces supported
        return true;
    }

```

However, this batching mechanism reduces the result to a single boolean value "true" or "false". The proposal is to replicate this behaviour but for each interface return a separate true/false value as sometimes you anticipate some interfaces will be implemented but not all and can tailor your logic around the interfaces supported. The benefit of the helper method vs looping over `supportsInterface` is that redundant checks to `supportsERC165` aren't made.

**📝 Details**

Here is the code I've written which is pretty self-explanatory. If ERC165 isnt implemented on the contract then no checks can be made and an empty array is returned, else a boolean array is returned corresponding to the interface ids passed.

```solidity
/**
     * @dev Returns a boolean array where each value corresponds to the
     * interfaces passed in and whether they're supported or not. This allows
     * you to batch check interfaces for a contract where your expectation
     * is that some interfaces may not be supported.
     *
     * See {IERC165-supportsInterface}.
     */
    function getSupportedInterfaces(address account, bytes4[] memory interfaceIds) public view returns (bool[] memory) {
        // query support of ERC165 itself
        if (!supportsERC165(account)) {
            //return an empty array to signify no checks could be made...
            return new bool[](0);
        }

        // an array of booleans corresponding to interfaceIds and whether they're supported or not
        bool[] memory interfaceIdsSupported = new bool[](interfaceIds.length);

        // query support of each interface in interfaceIds
        for (uint256 i = 0; i < interfaceIds.length; i++) {
            if (!_supportsERC165Interface(account, interfaceIds[i])) {
                interfaceIdsSupported[i] = false;
            } else {
                interfaceIdsSupported[i] = true;
            }
        }

        return interfaceIdsSupported;
    }
```

If this proposal is good and OZ is happy to go ahead with it, I can write tests and PR in.

conspyrosy:
Fixes #2429 

See the issue, explained there.

Tests work. Linter giving me line errors in windows

### DIFF
--- a/contracts/introspection/ERC165Checker.sol
+++ b/contracts/introspection/ERC165Checker.sol
@@ -48,7 +48,7 @@ library ERC165Checker {
      *
      * See {IERC165-supportsInterface}.
      */
-    function getSupportedInterfaces(address account, bytes4[] memory interfaceIds) public view returns (bool[] memory) {
+    function getSupportedInterfaces(address account, bytes4[] memory interfaceIds) internal view returns (bool[] memory) {
         // query support of ERC165 itself
         if (!supportsERC165(account)) {
             //return an array with false values
--- a/contracts/introspection/ERC165Checker.sol
+++ b/contracts/introspection/ERC165Checker.sol
@@ -49,28 +49,17 @@ library ERC165Checker {
      * See {IERC165-supportsInterface}.
      */
     function getSupportedInterfaces(address account, bytes4[] memory interfaceIds) internal view returns (bool[] memory) {
-        // query support of ERC165 itself
-        if (!supportsERC165(account)) {
-            //return an array with false values
-            bool[] memory supportedInterfaces = new bool[](interfaceIds.length);
-            for(uint256 i = 0; i < interfaceIds.length; i++) {
-                supportedInterfaces[i] = false;
-            }
-            return supportedInterfaces;
-        }
-
         // an array of booleans corresponding to interfaceIds and whether they're supported or not
         bool[] memory interfaceIdsSupported = new bool[](interfaceIds.length);
 
-        // query support of each interface in interfaceIds
-        for (uint256 i = 0; i < interfaceIds.length; i++) {
-            if (!_supportsERC165Interface(account, interfaceIds[i])) {
-                interfaceIdsSupported[i] = false;
-            } else {
-                interfaceIdsSupported[i] = true;
+        // query support of ERC165 itself
+        if (supportsERC165(account)) {
+            // query support of each interface in interfaceIds
+            for (uint256 i = 0; i < interfaceIds.length; i++) {
+                interfaceIdsSupported[i] = _supportsERC165Interface(account, interfaceIds[i]);
             }
         }
-        
+
         return interfaceIdsSupported;
     }
 
--- a/contracts/introspection/ERC165Checker.sol
+++ b/contracts/introspection/ERC165Checker.sol
@@ -41,6 +41,36 @@ library ERC165Checker {
     }
 
     /**
+     * @dev Returns a boolean array where each value corresponds to the
+     * interfaces passed in and whether they're supported or not. This allows
+     * you to batch check interfaces for a contract where your expectation
+     * is that some interfaces may not be supported.
+     *
+     * See {IERC165-supportsInterface}.
+     */
+    function getSupportedInterfaces(address account, bytes4[] memory interfaceIds) public view returns (bool[] memory) {
+        // query support of ERC165 itself
+        if (!supportsERC165(account)) {
+            //return an empty array to signify no checks...
+            return new bool[](0);
+        }
+
+        // an array of booleans corresponding to interfaceIds and whether they're supported or not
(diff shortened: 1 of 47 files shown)


contracts/introspection/ERC165Checker.sol
```diff
@@ -40,6 +40,40 @@ library ERC165Checker {
...
+            }
+            return supportedInterfaces;
+        }
+
+        // an array of booleans corresponding to interfaceIds and whether they're supported or not
+        bool[] memory interfaceIdsSupported = new bool[](interfaceIds.length);
+
+        // query support of each interface in interfaceIds
+        for (uint256 i = 0; i < interfaceIds.length; i++) {
+            if (!_supportsERC165Interface(account, interfaceIds[i])) {
```
Amxx, line 67: I this doing
```
for (uint256 i = 0; i < interfaceIds.length; i++) {
    interfaceIdsSupported[i] = _supportsERC165Interface(account, interfaceIds[i]);
}
```
would be both clearer and possibly more gas efficiant

test/introspection/ERC165Checker.test.js
```diff
@@ -33,9 +33,10 @@ contract('ERC165Checker', function (accounts) {
       expect(supported).to.equal(false);
     });
 
-    it('does not support mock interface via supportsAllInterfaces', async function () {
-      const supported = await this.mock.supportsAllInterfaces(this.target.address, [DUMMY_ID]);
-      expect(supported).to.equal(false);
+    it('does not support mock interface via getSupportedInterfaces', async function () {
```
Amxx, line 36: Please don't remove the existing test and add the new one in a separate `it` just like you did for the other contexts

Amxx (changes requested):

Amxx:
Not that by default, newly allocated arrays should contain 0 (@frangio can you confirme this?)

If so the getSupportedInterface can be shorten to:
```
function getSupportedInterfaces(address account, bytes4[] memory interfaceIds) internal view returns (bool[] memory) {
    // an array of booleans corresponding to interfaceIds and whether they're supported or not
    bool[] memory interfaceIdsSupported = new bool[](interfaceIds.length);
    
    // query support of ERC165 itself
    if (supportsERC165(account)) {
        // query support of each interface in interfaceIds
        for (uint256 i = 0; i < interfaceIds.length; ++i) {
            interfaceIdsSupported[i] = _supportsERC165Interface(account, interfaceIds[i]);
        }
    }
    
    return interfaceIdsSupported;
}
```

frangio:
Thanks @conspyrosy!

> I do believe that, by default, newly allocated arrays should contain 0 (@frangio can you confirme this?), which in boolean terms means false.

I'm pretty sure this is the case but I couldn't find it in the documentation for Solidity. I've asked the team to confirm.

frangio:
The reply from the Solidity team was:

>at the Solidity level, all newly allocated data structures are guaranteed to be initialized properly. At the assembly level, you cannot rely on 0x40 pointing to zeroed out memory

In summary, yes we can just initialize a new array.

conspyrosy:
I've only just got round to checking on this, looks like some changes have been made which look good 👍 Let me know if this needs anything from me.

frangio:
Hi @conspyrosy, no problem. Sorry that we jumped in, we want to ship this in the next release. :slightly_smiling_face:

frangio (approved):

PR #2470, opened 2021-01-15
Update ERC721.sol

smarteasy:
Fixes #


smarteasy:
For consistency in coding style.

frangio:
Thank you @smarteasy. I think these particular variable names were following our current convention. Take a look at https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2396#issuecomment-717488049.

PR #2471, opened 2021-01-15
Update Ownable.sol

akarshb02:
Fixes #


Amxx:
Hello,
Thank you for contributing this PR. Can you explain what issue you are trying to fix ?
`Ownable`'s contrastructor in internal because `Ownable` is an abstract contract, which should not be used alone, but as a building block when creating bigger, administered, contracts.
For more information about usage of the `Ownable` contract, [please have a look at our documentation](https://docs.openzeppelin.com/contracts/3.x/access-control)

PR #2472, opened 2021-01-18
Argument missing in approve method.

rahulct17:
the recipient argument was missing in the  approve message

Fixes #

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -131,7 +131,7 @@ contract ERC20 is Context, IERC20 {
      *
      * - `spender` cannot be the zero address.
      */
-    function approve(address spender, uint256 amount) public virtual override returns (bool) {
+    function approve(address spender,address recipient, uint256 amount) public virtual override returns (bool) {
         _approve(_msgSender(), spender, amount);
         return true;
     }


Amxx:
Hello @rahulct17  and thank you for your PR.

Could you please explain what you are trying to fix? The approve function, as currently implemented, correspond to the [ERC20 standard](https://eips.ethereum.org/EIPS/eip-20). Furthermore, the parameter you are adding is unused.

rahulct17:
When run in remix ide ...the approve function has some error showing it is
not an override function...

So when we pass the argument (address recipient) it runs without any errors

Amxx:
The ERC20 interface is standard and should not be changed. It you have an issue using this contract in Remix, it would be much better to find the origine of the issue and fix it. Would you be able to provide a piece of code that, when put into remix, reproduces your issue ?

rahulct17:
Please see the attachment

frangio:
@rahulct17 We didn't receive an attachment here in the GitHub pull request.

If Remix is showing an error related to overrides, then there is something else wrong in your setup.

rahulct17:
Please provide your email  address so I can explain you about the issue

frangio:
For support please visit https://forum.openzeppelin.com and create a thread to discuss your issue.

PR #2473, opened 2021-01-18
Make view and pure functions virtual

### LINKED ISSUE #2154: Virtual view functions
Please mark `tokenURI` in ERC721 `virtual`.

As it's not `virtual`, it can't be overridden by inheriting contracts. Making this virtual would allow contracts to implement programmatic generation of token URIs without using storage slots unnecessarily.

Amxx:
Fixes #2154

Overriding many of these functions is risky, and need proper documentation/warning.

### DIFF
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -84,7 +84,7 @@ abstract contract AccessControl is Context {
     /**
      * @dev Returns `true` if `account` has been granted `role`.
      */
-    function hasRole(bytes32 role, address account) public view virtual returns (bool) {
+    function hasRole(bytes32 role, address account) public view returns (bool) {
         return _roles[role].members.contains(account);
     }
 
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -92,7 +92,7 @@ abstract contract AccessControl is Context {
      * @dev Returns the number of accounts that have `role`. Can be used
      * together with {getRoleMember} to enumerate all bearers of a role.
      */
-    function getRoleMemberCount(bytes32 role) public view virtual returns (uint256) {
+    function getRoleMemberCount(bytes32 role) public view returns (uint256) {
         return _roles[role].members.length();
     }
 
@@ -108,7 +108,7 @@ abstract contract AccessControl is Context {
      * https://forum.openzeppelin.com/t/iterating-over-elements-on-enumerableset-in-openzeppelin-contracts/2296[forum post]
      * for more information.
      */
-    function getRoleMember(bytes32 role, uint256 index) public view virtual returns (address) {
+    function getRoleMember(bytes32 role, uint256 index) public view returns (address) {
         return _roles[role].members.at(index);
     }
 
@@ -118,7 +118,7 @@ abstract contract AccessControl is Context {
      *
      * To change a role's admin, use {_setRoleAdmin}.
      */
-    function getRoleAdmin(bytes32 role) public view virtual returns (bytes32) {
+    function getRoleAdmin(bytes32 role) public view returns (bytes32) {
         return _roles[role].adminRole;
     }
 
@@ -133,7 +133,7 @@ abstract contract AccessControl is Context {
      * - the caller must have ``role``'s admin role.
      */
     function grantRole(bytes32 role, address account) public virtual {
-        require(hasRole(getRoleAdmin(role), _msgSender()), "AccessControl: sender must be an admin to grant");
+        require(hasRole(_roles[role].adminRole, _msgSender()), "AccessControl: sender must be an admin to grant");
 
         _grantRole(role, account);
     }
@@ -148,7 +148,7 @@ abstract contract AccessControl is Context {
      * - the caller must have ``role``'s admin role.
      */
     function revokeRole(bytes32 role, address account) public virtual {
-        require(hasRole(getRoleAdmin(role), _msgSender()), "AccessControl: sender must be an admin to revoke");
+        require(hasRole(_roles[role].adminRole, _msgSender()), "AccessControl: sender must be an admin to revoke");
 
         _revokeRole(role, account);
     }
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -84,7 +84,7 @@ abstract contract AccessControl is Context {
     /**
      * @dev Returns `true` if `account` has been granted `role`.
      */
-    function hasRole(bytes32 role, address account) public view returns (bool) {
+    function hasRole(bytes32 role, address account) public view virtual returns (bool) {
         return _roles[role].members.contains(account);
     }
 
@@ -92,7 +92,7 @@ abstract contract AccessControl is Context {
      * @dev Returns the number of accounts that have `role`. Can be used
      * together with {getRoleMember} to enumerate all bearers of a role.
      */
-    function getRoleMemberCount(bytes32 role) public view returns (uint256) {
+    function getRoleMemberCount(bytes32 role) public view virtual returns (uint256) {
         return _roles[role].members.length();
     }
 
@@ -108,7 +108,7 @@ abstract contract AccessControl is Context {
      * https://forum.openzeppelin.com/t/iterating-over-elements-on-enumerableset-in-openzeppelin-contracts/2296[forum post]
      * for more information.
      */
-    function getRoleMember(bytes32 role, uint256 index) public view returns (address) {
+    function getRoleMember(bytes32 role, uint256 index) public view virtual returns (address) {
         return _roles[role].members.at(index);
     }
 
@@ -118,7 +118,7 @@ abstract contract AccessControl is Context {
      *
      * To change a role's admin, use {_setRoleAdmin}.
      */
-    function getRoleAdmin(bytes32 role) public view returns (bytes32) {
+    function getRoleAdmin(bytes32 role) public view virtual returns (bytes32) {
         return _roles[role].adminRole;
     }
 
@@ -133,7 +133,7 @@ abstract contract AccessControl is Context {
      * - the caller must have ``role``'s admin role.
      */
     function grantRole(bytes32 role, address account) public virtual {
-        require(hasRole(_roles[role].adminRole, _msgSender()), "AccessControl: sender must be an admin to grant");
+        require(hasRole(getRoleAdmin(role), _msgSender()), "AccessControl: sender must be an admin to grant");
 
         _grantRole(role, account);
     }
@@ -148,7 +148,7 @@ abstract contract AccessControl is Context {
      * - the caller must have ``role``'s admin role.
      */
     function revokeRole(bytes32 role, address account) public virtual {
-        require(hasRole(_roles[role].adminRole, _msgSender()), "AccessControl: sender must be an admin to revoke");
+        require(hasRole(getRoleAdmin(role), _msgSender()), "AccessControl: sender must be an admin to revoke");
 
         _revokeRole(role, account);
     }
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -32,7 +32,7 @@ abstract contract Ownable is Context {
     /**
      * @dev Returns the address of the current owner.
      */
-    function owner() public view returns (address) {
+    function owner() public view virtual returns (address) {
         return _owner;
     }
 
@@ -40,7 +40,7 @@ abstract contract Ownable is Context {
      * @dev Throws if called by any account other than the owner.
      */
     modifier onlyOwner() {
-        require(_owner == _msgSender(), "Ownable: caller is not the owner");
+        require(owner() == _msgSender(), "Ownable: caller is not the owner");
         _;
     }
 
--- a/contracts/access/TimelockController.sol
+++ b/contracts/access/TimelockController.sol
@@ -94,32 +94,41 @@ contract TimelockController is AccessControl {
     receive() external payable {}
 
     /**
+     * @dev Returns whether an id correspond to a registered operation. This
+     * includes both Pending, Ready and Done operations.
+     */
+    function isOperation(bytes32 id) public view virtual returns (bool pending) {
+        return getTimestamp(id) > 0;
+    }
+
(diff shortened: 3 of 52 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -155,29 +155,29 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
-    function baseURI() public view returns (string memory) {
+    function baseURI() public view virtual returns (string memory) {
         return _baseURI;
     }
 
     /**
      * @dev See {IERC721Enumerable-tokenOfOwnerByIndex}.
      */
-    function tokenOfOwnerByIndex(address owner, uint256 index) public view override returns (uint256) {
+    function tokenOfOwnerByIndex(address owner, uint256 index) public view virtual override returns (uint256) {
```
frangio, line 165: This feels like it shouldn't be virtual.

contracts/token/ERC721/ERC721.sol
```diff
@@ -155,29 +155,29 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
+    function totalSupply() public view virtual override returns (uint256) {
         // _tokenOwners are indexed by tokenIds, so .length() returns the number of tokenIds
         return _tokenOwners.length();
     }
 
     /**
      * @dev See {IERC721Enumerable-tokenByIndex}.
      */
-    function tokenByIndex(uint256 index) public view override returns (uint256) {
+    function tokenByIndex(uint256 index) public view virtual override returns (uint256) {
```
frangio, line 180: This feels like it shouldn't be virtual either"

contracts/utils/Pausable.sol
```diff
@@ -36,7 +36,7 @@ abstract contract Pausable is Context {
     /**
      * @dev Returns true if the contract is paused, and false otherwise.
      */
-    function paused() public view returns (bool) {
+    function paused() public view virtual returns (bool) {
```
frangio, line 39: The issue with making this virtual is that the Paused/Unpaused events stop being reliable. What do you think?

contracts/token/ERC777/ERC777.sol
```diff
@@ -115,21 +115,21 @@ contract ERC777 is Context, IERC777, IERC20 {
      *
      * This implementation always returns `1`.
      */
-    function granularity() public view override returns (uint256) {
+    function granularity() public view virtual override returns (uint256) {
```
frangio, line 118: I think this is okay because there needs to be a way to implement tokens with different granularity, but those contracts should also at the same time override transfer to make sure it reverts if the amount is not a multiple, so this should be documented.

Alternatively we add that check ourselves but that will cost gas, I don't think the compiler will optimize it.

contracts/token/ERC777/ERC777.sol
```diff
@@ -115,21 +115,21 @@ contract ERC777 is Context, IERC777, IERC20 {
...
         return 1;
     }
 
     /**
      * @dev See {IERC777-totalSupply}.
      */
-    function totalSupply() public view override(IERC20, IERC777) returns (uint256) {
+    function totalSupply() public view virtual override(IERC20, IERC777) returns (uint256) {
```
frangio, line 125: Not virtual, same arguments as ERC20.

contracts/token/ERC777/ERC777.sol
```diff
@@ -115,21 +115,21 @@ contract ERC777 is Context, IERC777, IERC20 {
...
         return _totalSupply;
     }
 
     /**
      * @dev Returns the amount of tokens owned by an account (`tokenHolder`).
      */
-    function balanceOf(address tokenHolder) public view override(IERC20, IERC777) returns (uint256) {
+    function balanceOf(address tokenHolder) public view virtual override(IERC20, IERC777) returns (uint256) {
```
frangio, line 132: Not virtual, same arguments as ERC20.

contracts/token/ERC777/ERC777.sol
```diff
@@ -176,7 +176,7 @@ contract ERC777 is Context, IERC777, IERC20 {
     /**
      * @dev See {IERC777-isOperatorFor}.
      */
-    function isOperatorFor(address operator, address tokenHolder) public view override returns (bool) {
+    function isOperatorFor(address operator, address tokenHolder) public view virtual override returns (bool) {
```
frangio, line 179: For this one and `defaultOperators`... I can imagine people might want to override this with a more efficient implementation that doesn't use storage, or something. But ERC777 specifies that the default operators should never change, so this is something people would have to know if they override them. I guess they shouldn't expect to be able to override a function specified in the ERC without checking that their semantics are compliant with the spec.

contracts/token/ERC721/ERC721.sol
```diff
@@ -112,28 +112,28 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
     /**
      * @dev See {IERC721-ownerOf}.
      */
-    function ownerOf(uint256 tokenId) public view override returns (address) {
+    function ownerOf(uint256 tokenId) public view virtual override returns (address) {
```
frangio, line 115: Not sure how I feel about this one.

contracts/token/ERC721/ERC721.sol
```diff
@@ -103,7 +103,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
     /**
      * @dev See {IERC721-balanceOf}.
      */
-    function balanceOf(address owner) public view override returns (uint256) {
+    function balanceOf(address owner) public view virtual override returns (uint256) {
```
frangio, line 106: Not sure how I feel about this one either.

frangio (commented):
These are all my opinions but I got tired of writing "I think...". :stuck_out_tongue:

contracts/access/AccessControl.sol
```diff
@@ -84,15 +84,15 @@ abstract contract AccessControl is Context {
     /**
      * @dev Returns `true` if `account` has been granted `role`.
      */
-    function hasRole(bytes32 role, address account) public view returns (bool) {
+    function hasRole(bytes32 role, address account) public view virtual returns (bool) {
```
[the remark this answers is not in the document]
Amxx: I think it is a similar function to ERC721 `isApprovedOrOwner` in the sens that is easily composable. Also I see real usecases to compose that.

contracts/access/TimelockController.sol
```diff
@@ -93,33 +93,42 @@ contract TimelockController is AccessControl {
      */
     receive() external payable {}
 
+    /**
+     * @dev Returns whether an id correspond to a registered operation. This
+     * includes both Pending, Ready and Done operations.
+     */
+    function isOperation(bytes32 id) public view virtual returns (bool pending) {
```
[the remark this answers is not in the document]
Amxx: I thought that if someone was to overload the isOperationPending / isOperationReady / isOperationDone in a way that would impact the contract's orchastration, then it would be good to have this virtual function, as the corresponding logic would likely also have overriden for changes to the other functions to ben consistent

contracts/payment/escrow/RefundEscrow.sol
```diff
@@ -79,15 +79,15 @@ contract RefundEscrow is ConditionalEscrow {
      * @dev Withdraws the beneficiary's funds.
      */
     function beneficiaryWithdraw() public virtual {
-        require(_state == State.Closed, "RefundEscrow: beneficiary can only withdraw while closed");
-        _beneficiary.transfer(address(this).balance);
+        require(state() == State.Closed, "RefundEscrow: beneficiary can only withdraw while closed");
+        Address.sendValue(payable(beneficiary()), address(this).balance);
```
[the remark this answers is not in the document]
Amxx: see #2480

contracts/utils/Pausable.sol
Amxx: Thats is true of all virtual functions that trigger events. Following this argument, `ERC20._transfer` should not be made virtual.

I think that overriding (non-hook) functions requiers a good understanding of the contract being overridden.

This question is more about the usage we expect from pausable:
- should pausable be a mechnism that expects explicit calls to the `_pause` & `_unpause` functions, which trigger event, so that the pausable status can be follow from events. This is the logic we want if we see pausable a a tool for defender.
- should pausable be a building blocks that provide modifier that can be used in your contract and are compatible with any pausing/unpausing pattern. There is a default one (relying on storage, events, and internal functions), but this is just an example and this can be changed to any logic (like a "pause on sunday" logic) that will not necessarily trigger events / be compatible with defender.

I prefer the second option, but that is just my point of view

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -129,7 +129,7 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     /**
      * @dev See {IERC1155-isApprovedForAll}.
      */
-    function isApprovedForAll(address account, address operator) public view override returns (bool) {
+    function isApprovedForAll(address account, address operator) public view virtual override returns (bool) {
```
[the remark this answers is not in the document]
Amxx: Same discussion as for `Pausable.paused()`

contracts/access/Ownable.sol
```diff
@@ -32,15 +32,15 @@ abstract contract Ownable is Context {
     /**
      * @dev Returns the address of the current owner.
      */
-    function owner() public view returns (address) {
+    function owner() public view virtual returns (address) {
```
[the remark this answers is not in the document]
Amxx: I lean on the side of:
- people are eventually going to request it
- we can make it virtual without breaking consistency.
So why not do it ?

contracts/token/ERC721/ERC721.sol
Amxx: So,

This function reverts when to token doesn't exist (not sure if that is part of the protocole). So the usecase that returns address(0) if expirect actually make sens: "the token exist, but isn't owned by anyone". Still there would be an issue if someone was to transfer an expired token, since that would trigger events (and hook) with `from = address(0)`, and it would try to remove the tokenid from the `_holderTokens[0]` set.

So if this is made virtual, it should not be used as a getter when doing `_transfer` or `_burn` (nor in _approve, which would technically still be triggered if other function are overloaded to give approval to tokens owner by 0).

This is a tricky one. Not as obviously breaking as balanceOf, but still sensitive. The thing is that, unlike balanceOf (and provided we don't use this getter in the place described before), a user could overload that in a way that would keep the registry "mostly" standard without having to overload other functions.

If we have it virtual, we should have a big disclaimer. If we have it non-virtual, we need a clear argument as to why not (this was requested)

contracts/token/ERC777/ERC777.sol
Amxx: Not sure what you are trying to say. Keeping it virtual with a reminder of the expected behaviour & ERC specifications ? Not having it virtual ?

contracts/drafts/ERC20Permit.sol
```diff
@@ -62,7 +62,7 @@ abstract contract ERC20Permit is ERC20, IERC20Permit, EIP712 {
     /**
      * @dev See {IERC20Permit-nonces}.
      */
-    function nonces(address owner) public view override returns (uint256) {
+    function nonces(address owner) public view virtual override returns (uint256) {
```
frangio, line 65: What do you think about this one given that we also do `_nonces[owner].increment()`?

contracts/access/Ownable.sol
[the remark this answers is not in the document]
frangio: Yeah, sounds good.

contracts/token/ERC777/ERC777.sol
frangio: Sorry I rambled a bit. I was unsure but I think if it can be easily made virtual in a consistent way then we should do it, and people should be familiar with the spec requirements if they ever override it.

contracts/token/ERC721/ERC721.sol
frangio: >So the usecase that returns address(0) if expired

Note that the [example](https://github.com/ensdomains/ethregistrar/blob/75b4515b33ff06f4e5e980c36a9a5214e7d1bd4c/contracts/BaseRegistrarImplementation.sol#L46-L49) the user shared was:

```solidity
    function ownerOf(uint256 tokenId) public view returns (address) {
        require(expiries[tokenId] > now);
        return super.ownerOf(tokenId);
    }
```

So it actually reverts if expired. I don't really think reverting here means anything for the purposes of ERC721 semantics. The token may have an owner but the contract might still allow someone else to reclaim it if it has expired. So if I were to design the contract I wouldn't worry about `ownerOf`... But in any case, it's true that we could internally use a private `_ownerOf` function (or read directly from storage), and allow people if they want to override the public function. I don't see an issue with that.

We could even do the same thing for `ERC20.balanceOf`... People can override the public function if they want but we internally keep reading from and writing to storage. If they want to change those semantics they can override `_transfer` or the before hook.

contracts/utils/Pausable.sol
frangio: >Following this argument, ERC20._transfer should not be made virtual.

Why not? In ERC20 it's not possible to make a transfer without emitting the event, even if `_transfer` is virtual.

But the argument does apply to not making `ERC20.balanceOf` virtual. And `ERC721.ownerOf`... But that would be basically against the spec so people would know not to do it.

 `Pausable` is not specified, so we would have to document the semantics. But it feels impossible to override `paused` in a way that preserves the semantics that events fully track pausing and unpausing.

Amxx: > Pausable is not specified, so we would have to document the semantics. But it feels impossible to override paused in a way that preserves the semantics that events fully track pausing and unpausing.

True. I just think that the semantics are not essential in this case (not an ERC?) and the intent to pause function through a modifier is the key element, regardless of changes being notified by events.
The change should be notified when "triggered", but I don't have an issue with pausable patterns where the change is not "triggered", but a consequence of an external element (time, other contract)

PR #2482, opened 2021-01-25
Use trySub to avoid memory allocation

Amxx:
Use the new `SafeMath.trySub(uint256,uint256)` functions to avoid memory leaks caused by `SafeMath.sub(uint256,uint256,string)`

#### PR Checklist

- [x] Tests
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -157,7 +157,10 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
 
         _beforeTokenTransfer(operator, from, to, _asSingletonArray(id), _asSingletonArray(amount), data);
 
-        _balances[id][from] = _balances[id][from].sub(amount, "ERC1155: insufficient balance for transfer");
+        (bool success, uint256 newBalance) = _balances[id][from].trySub(amount);
+        require(success, "ERC1155: insufficient balance for transfer");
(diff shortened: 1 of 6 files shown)


frangio (commented):
Looks good, but we want to see the gas difference first.

Amxx:
> Looks good, but we want to see the gas difference first.

I tried doing a `transferFrom(<me>,<me>,0)` with both versions:

- *master*: 
  - transaction cost: 34451
  - execution cost: 10235
- "this PR":
  - transaction cost: 34429
  - execution cost: 10213
(Using solidity 0.6.12 with optimizations).

So there is a benefit, but i'm not really sure it's worth the decrease in code readability.

Amxx:
Just to be safe about memory growing in size, and the effect only being visible when many calls are performed, I used this tool:
```
contract TESTER {
    function testTransferFrom(IERC20 token, uint256 iterations) public {
        for (uint256 i = 0; i < iterations; ++i) token.transferFrom(address(this), address(this), 0);
    }
}
```

Running the `testTransferFrom` with 100 iterations used:
- 1194948 gas with the master
- 1192748 gas with this PR

Conclusion: we save exactly 22gas per run of transferFrom, ie 11gas per changes from `.sub(uint,uint,string)` to `.trySub(uint,uint)`. I don't really think its worth it.

frangio:
Great experiment. Agree we should prioritize clarity of the code here.

PR #2487, opened 2021-01-28
Adopt new Solidity features interfaceId, try/catch, keccak constants

Amxx:
This is a purely internal changes, that only modifies the way some `private constant` variable are computed.

- Using type(T).interfaceId automatically computes ERC165 interface id from the corresponding Interface, removing the need to hardcode it
- Hardcoding some constant values that were hashed at compilation

- [x] Tests

### DIFF
--- a/contracts/access/TimelockController.sol
+++ b/contracts/access/TimelockController.sol
@@ -20,21 +20,9 @@ import "./AccessControl.sol";
  * _Available since v3.3._
  */
 contract TimelockController is AccessControl {
-    // We inline the result of the following hashes because Solidity doesn't resolve them at compile time.
-    // See https://github.com/ethereum/solidity/issues/4024.
-
-    // keccak256("TIMELOCK_ADMIN_ROLE");
-    bytes32 public constant TIMELOCK_ADMIN_ROLE =
-        0xd8aa0f3194971a2a116679f7c2090f6939c8d4e01a2a8d7e41d55e5351469e63;
-
-    // keccak256("PROPOSER_ROLE");
-    bytes32 public constant PROPOSER_ROLE =
-        0xb09aa5aeb3702cfd50b6b62bc4532604938f21248a27a1d5ca736082b6819cc1;
-
-    // keccak256("EXECUTOR_ROLE");
-    bytes32 public constant EXECUTOR_ROLE =
-        0x29ddb589b1fb5fc7cf394961c1adf5f8c6454761adf795e67fe149f658abe895;
-
+    bytes32 public constant TIMELOCK_ADMIN_ROLE = keccak256("TIMELOCK_ADMIN_ROLE");
+    bytes32 public constant PROPOSER_ROLE = keccak256("PROPOSER_ROLE");
+    bytes32 public constant EXECUTOR_ROLE = keccak256("EXECUTOR_ROLE");
     uint256 private constant _DONE_TIMESTAMP = uint256(1);
 
     mapping(bytes32 => uint256) private _timestamps;
--- a/contracts/access/TimelockController.sol
+++ b/contracts/access/TimelockController.sol
@@ -20,11 +20,22 @@ import "./AccessControl.sol";
  * _Available since v3.3._
  */
 contract TimelockController is AccessControl {
+    // We inline the result of the following hashes because Solidity doesn't resolve them at compile time.
+    // See https://github.com/ethereum/solidity/issues/4024.
 
-    bytes32 public constant TIMELOCK_ADMIN_ROLE = keccak256("TIMELOCK_ADMIN_ROLE");
-    bytes32 public constant PROPOSER_ROLE = keccak256("PROPOSER_ROLE");
-    bytes32 public constant EXECUTOR_ROLE = keccak256("EXECUTOR_ROLE");
-    uint256 internal constant _DONE_TIMESTAMP = uint256(1);
+    // keccak256("TIMELOCK_ADMIN_ROLE");
+    bytes32 public constant TIMELOCK_ADMIN_ROLE =
+        0xd8aa0f3194971a2a116679f7c2090f6939c8d4e01a2a8d7e41d55e5351469e63;
+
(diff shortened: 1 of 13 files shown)


contracts/access/TimelockController.sol
```diff
@@ -20,11 +20,10 @@ import "./AccessControl.sol";
  * _Available since v3.3._
  */
 contract TimelockController is AccessControl {
-
     bytes32 public constant TIMELOCK_ADMIN_ROLE = keccak256("TIMELOCK_ADMIN_ROLE");
     bytes32 public constant PROPOSER_ROLE = keccak256("PROPOSER_ROLE");
     bytes32 public constant EXECUTOR_ROLE = keccak256("EXECUTOR_ROLE");
-    uint256 internal constant _DONE_TIMESTAMP = uint256(1);
+    uint256 private constant _DONE_TIMESTAMP = uint256(1);
```
frangio, line 26: If this is unrelated let's keep it out of this PR.

contracts/introspection/ERC165Checker.sol
```diff
@@ -101,29 +95,8 @@ library ERC165Checker {
...
-    function _callERC165SupportsInterface(address account, bytes4 interfaceId)
-        private
-        view
-        returns (bool, bool)
-    {
-        bytes memory encodedParams = abi.encodeWithSelector(_INTERFACE_ID_ERC165, interfaceId);
-        (bool success, bytes memory result) = account.staticcall{ gas: 30000 }(encodedParams);
-        if (result.length < 32) return (false, false);
-        return (success, abi.decode(result, (bool)));
+        (bool success, bytes memory result) = account.staticcall{ gas: 30000 }(abi.encodeWithSelector(IERC165(account).supportsInterface.selector, interfaceId));
```
frangio, line 98: Is there a memory allocation benefit to inlining the `abi.encodeWithSelector` expression in the argument list? Otherwise it was clearer as a separate variable.

contracts/introspection/ERC165Checker.sol
```diff
@@ -10,22 +12,14 @@ pragma solidity ^0.8.0;
...
     /**
      * @dev Returns true if `account` supports the {IERC165} interface,
      */
     function supportsERC165(address account) internal view returns (bool) {
         // Any contract that implements ERC165 must explicitly indicate support of
         // InterfaceId_ERC165 and explicitly indicate non-support of InterfaceId_Invalid
-        return _supportsERC165Interface(account, _INTERFACE_ID_ERC165) &&
-            !_supportsERC165Interface(account, _INTERFACE_ID_INVALID);
+        return _supportsERC165Interface(account, type(IERC165).interfaceId) &&
+            !_supportsERC165Interface(account, 0xffffffff);
```
frangio, line 22: I think it was better to have a name for `0xffffffff`. I would keep the constant variable.

contracts/token/ERC721/ERC721.sol
```diff
@@ -433,18 +395,17 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Enumerable
...
         private returns (bool)
     {
-        if (!to.isContract()) {
+        if (to.isContract()) {
+            try IERC721Receiver(to).onERC721Received(_msgSender(), from, tokenId, _data) returns (bytes4 retval) {
+                return retval == IERC721Receiver(to).onERC721Received.selector;
+            } catch Error (string memory reason) {
+                revert(reason);
+            } catch {
+                revert("ERC721: transfer to non ERC721Receiver implementer");
```
frangio, line 404: We need something like this to catch every revert reason, including others that would match `Panic(uint256)`, etc.
But the problem with my suggestion below is that I think `revert` wraps it with an `Error(string)` selector. We need to check if this is true, and if so we may need to use assembly `revert`.

```suggestion
            } catch (bytes memory reason) {
                if (reason.length == 0) {
                    revert("ERC721: transfer to non ERC721Receiver implementer");
                } else {
                    revert(reason);
                }
            }
```

frangio (commented):
This PR has a lot of good things but it goes beyond ERC165, so can you change the title of the PR?

contracts/token/ERC721/ERC721.sol
Amxx: When the revert calldata contains a "revert reason" string, this is catched by 
```
} catch Error (string memory reason) {
        revert(reason);
}
```
and forwarded correctly.

Catching Panics in a specific way only makes sens if we want to send a specific revert message for them. I'll check that Panic are catched by the genereic `} catch {`, but if yes, the implementation in the PR should be good. It does pass all tests we have, including the ones where the receiver reverts with a message

Amxx: the question is, what error message to we expect to get when:
- the receiver panics
- the receiver reverts without a reason

Amxx: I'm not sure we have any way to distinguish "the receiver reverts without a reason" and "the receiver does not implement `onERC721Received`.

Question is then: what to do when the receiver panics

Amxx: FYI, with master implementation, if the receiver panics, we end up with a revert that matched `expectRevert.unspecified`. Not sure if that contains the panic code, but I personnaly believe that even if it does, it is less expressive then an error that says the erc721receiver caused a panic

test/token/ERC721/ERC721.test.js
```diff
@@ -324,7 +324,7 @@ contract('ERC721', function (accounts) {
 
           describe('to a valid receiver contract', function () {
             beforeEach(async function () {
-              this.receiver = await ERC721ReceiverMock.new(RECEIVER_MAGIC_VALUE, false);
+              this.receiver = await ERC721ReceiverMock.new(RECEIVER_MAGIC_VALUE, 0);
```
frangio, line 327: It would be better to define constants for `0`, `1`, ..., so we don't need to consult what they mean.

frangio (approved):
This is good to go after my comment about the constants in tests is addressed.

PR #2491, opened 2021-01-29
re-enabling safemath revert reasons in ERC20, ERC777 and ERC1155

Amxx:
This PR adds revert message on some overflow errors in ERC20 and ERC1155. These revert message used to be part of 3.x and were removed during the transition to 0.8.0's built in safemath.

- [x] Tests


Amxx:
Adding the revert messages back to ERC777 as well.

PR #2500, opened 2021-02-02
Add revert string to Counter decrement overflow

Amxx:
This PR adds revert message an underflow error in Counter. There used to be a default safemath revert message in 3.x which was removed during the transition to 0.8.0's built in safemath.

#### PR Checklist

- [x] Tests
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/Counters.sol
+++ b/contracts/utils/Counters.sol
@@ -23,10 +23,16 @@ library Counters {
     }
 
     function increment(Counter storage counter) internal {
-        counter._value += 1;
+        unchecked {
+            counter._value += 1;
+        }
     }
 
     function decrement(Counter storage counter) internal {
-        counter._value = counter._value - 1;
+        unchecked {
+            uint256 value_ = counter._value;
+            require(value_ > 0, "Counter: decrement underflow");
+            counter._value = value_ - 1;
+        }
     }
 }
--- a/contracts/utils/Counters.sol
+++ b/contracts/utils/Counters.sol
@@ -30,9 +30,9 @@ library Counters {
 
     function decrement(Counter storage counter) internal {
         unchecked {
-            uint256 value_ = counter._value;
-            require(value_ > 0, "Counter: decrement underflow");
-            counter._value = value_ - 1;
+            uint256 value = counter._value;
+            require(value > 0, "Counter: decrement underflow");
+            counter._value = value - 1;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


contracts/utils/Counters.sol
```diff
@@ -23,10 +23,16 @@ library Counters {
...
+        unchecked {
+            counter._value += 1;
+        }
     }
 
     function decrement(Counter storage counter) internal {
-        counter._value = counter._value - 1;
+        unchecked {
+            uint256 value_ = counter._value;
+            require(value_ > 0, "Counter: decrement underflow");
```
frangio, line 34: We should keep unchecked to as few lines as possible.

```suggestion
        uint256 value_ = counter._value;
        require(value_ > 0, "Counter: decrement underflow");
        unchecked {
```

contracts/utils/Counters.sol
```diff
@@ -23,10 +23,16 @@ library Counters {
...
-        counter._value += 1;
+        unchecked {
+            counter._value += 1;
+        }
     }
 
     function decrement(Counter storage counter) internal {
-        counter._value = counter._value - 1;
+        unchecked {
+            uint256 value_ = counter._value;
```
frangio, line 33: Like in the tokens PR, I think we should remove the underscore suffix.

contracts/utils/Counters.sol
```diff
@@ -23,10 +23,16 @@ library Counters {
...
-        counter._value += 1;
+        unchecked {
+            counter._value += 1;
+        }
     }
 
     function decrement(Counter storage counter) internal {
-        counter._value = counter._value - 1;
+        uint256 value = counter._value;
+        require(value > 0, "Counter: decrement underflow");
```
frangio, line 33: Sorry, one last comment I didn't notice before. I don't know if this is the revert reason we had before but this should be "overflow". Underflow is for floating point numbers only.

Amxx: The previous error was the safemath one. I believed it would be better to have a more explicit one.

Do you want `Counter: decrement overflow` or would you prefer something different ?

frangio: "decrement overflow" sounds good to me.

Amxx: without the `Counter:` prefix ?

frangio: No, sorry, `Counter: decrement overflow`.

frangio (approved):
LGTM. Thanks!

PR #2501, opened 2021-02-02
Add support for eth-gas-reporter

Amxx:
Adding support for the hardhat-gas-reporter package.
Report is produced using npm run report-gas

#### PR Checklist

- [ ] Documentation
- [ ] Changelog entry


.github/workflows/test.yml
```diff
@@ -24,8 +24,19 @@ jobs:
...
-      - run: npm run test
+      - name: lint
+        run: npm run lint
+      - name: test
+        run: npm run test
+      - name: eth-gas-report
+        run: npm run test
+        env:
+          REPORT_GAS: 1
+          CC_SECRET: ${{ secrets.CC_SECRET }}
```
cgewecke, line 35: I think you'll also need to run an `npx codechecks` command separately after the tests complete. This retrieves historical data from the codechecks backend to construct the comparison diff and triggers the check in GH actions.

@Amxx Were you able to get a CC_SECRET in the private beta at codechecks?

cf: [codechecks docs at eth-gas-reporter][1]

[1]: https://github.com/cgewecke/eth-gas-reporter/blob/master/docs/codechecks.md#setup

Amxx: thx for the help. I'm trying to get that working and don't really understand everything yet

Amxx: Running `npx codechecks` gives:
```Error occured:  Provider should never be in fork mode and not in PR mode!```

Do you have any idea ho to fix that ?

cgewecke: Mmm. Not sure. The codechecks GH Actions integration might not work well with PRs from forks...

Will investigate.

cgewecke: @Amxx Not certain but this error might be a quirk caused by adding codechecks to the PR after the PR has already been opened. There's a comment in the GH Actions code at [codechecks/monorepo][1] which says:

> // github actions have explicit event for PR creation but we cant use it since it fired only once
> // in future we need to build a mechanism to get pull request information using API since it's impossible to get from the inside the PR

You could test this hypothesis by re-opening as different PR? Or merging into a staging branch? 


[1]: https://github.com/codechecks/monorepo/blob/ba52ad88a50898bf4ad246fd16ce5f7f7ada4034/packages/client/src/ci-providers/GithubActions.ts#L4-L5

Amxx: We are going to merge a version that generate report without CI integration for now. Will open an issue to do that in a following PR.

frangio:
I don't know why but I can't see the full report in the checks logs. :frowning_face: I think it's only showing a portion. Can you see the full report there?

Amxx:
No, I can't ... maybe the logs are truncated ?

Amxx:
locally, the result is good.

frangio:
I've separated the printing of the gas report in GitHub Actions into a separate step because the previous setup was truncating it. Should be good now.

frangio (approved):

PR #2504, opened 2021-02-02
Add a Strings.toHexString function

### LINKED ISSUE #2489: Add a ".toHexString" in the string library
**🧐 Motivation**
In some cases, such as when building revert message, it can be usefull to cast integers to strings. String.toString already does that in a "decimal" way. There is currently no option to cast uint to string in an "hex" format

**📝 Details**
Adding a second function in the `utils/Strings` library, named `toHexString` that would do that "encoding"

Amxx:
Fixes #2489

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -40,18 +40,26 @@ library Strings {
             return "0x00";
         }
         uint256 temp = value;
-        uint256 digits = 2;
+        uint256 length = 0;
         while (temp != 0) {
-            digits += 2;
+            length++;
             temp >>= 8;
         }
-        bytes memory buffer = new bytes(digits);
+        return toHexString(value, length);
+    }
+
+    /**
+     * @dev Converts a `uint256` to its ASCII `hexadecimal string` representation with fixed length.
+     */
+    function toHexString(uint256 value, uint256 length) internal pure returns (string memory) {
+        bytes memory buffer = new bytes(2 * length + 2);
         buffer[0] = "0";
         buffer[1] = "x";
-        while (value != 0) {
-            buffer[--digits] = alphabet[value & 0xf];
+        for (uint256 i = 2 * length + 1; i > 1; --i) {
+            buffer[i] = alphabet[value & 0xf];
             value >>= 4;
         }
         return string(buffer);
     }
+
 }
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -26,7 +26,8 @@ library Strings {
         }
         bytes memory buffer = new bytes(digits);
         while (value != 0) {
-            buffer[--digits] = bytes1(uint8(48 + uint256(value % 10)));
+            digits -= 1;
+            buffer[digits] = bytes1(uint8(48 + uint256(value % 10)));
             value /= 10;
         }
         return string(buffer);
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -6,7 +6,7 @@ pragma solidity ^0.8.0;
  * @dev String operations.
  */
 library Strings {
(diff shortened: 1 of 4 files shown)


contracts/utils/Strings.sol
```diff
@@ -6,6 +6,8 @@ pragma solidity ^0.8.0;
  * @dev String operations.
  */
 library Strings {
+    bytes16 private constant alphabet = 0x30313233343536373839616263646566;
```
frangio, line 9: This seems to work and is clearer.

```suggestion
    bytes16 private constant alphabet = "0123456789abcdef";
```

contracts/utils/Strings.sol
```diff
@@ -6,6 +6,8 @@ pragma solidity ^0.8.0;
...
+
     /**
      * @dev Converts a `uint256` to its ASCII `string` representation.
```
frangio, line 12:
```suggestion
     * @dev Converts a `uint256` to its ASCII `string` decimal representation.
```

contracts/utils/Strings.sol
```diff
@@ -23,12 +25,41 @@ library Strings {
...
-        temp = value;
+        while (value != 0) {
+            buffer[--digits] = bytes1(uint8(48 + uint256(value % 10)));
+            value /= 10;
+        }
+        return string(buffer);
+    }
+
+    /**
+     * @dev Converts a `uint256` to its ASCII `hexadecimal string` representation.
```
frangio, line 36:
```suggestion
     * @dev Converts a `uint256` to its ASCII `string` hexadecimal representation.
```

contracts/utils/Strings.sol
```diff
@@ -23,12 +25,41 @@ library Strings {
...
-            buffer[--index] = bytes1(uint8(48 + uint256(temp % 10)));
-            temp /= 10;
+            length++;
+            temp >>= 8;
+        }
+        return toHexString(value, length);
+    }
+
+    /**
+     * @dev Converts a `uint256` to its ASCII `hexadecimal string` representation with fixed length.
```
frangio, line 52:
```suggestion
     * @dev Converts a `uint256` to its ASCII `string` hexadecimal representation with fixed length.
```

contracts/utils/Strings.sol
```diff
@@ -23,12 +25,41 @@ library Strings {
             temp /= 10;
         }
         bytes memory buffer = new bytes(digits);
-        uint256 index = digits;
-        temp = value;
+        while (value != 0) {
+            buffer[--digits] = bytes1(uint8(48 + uint256(value % 10)));
```
frangio, line 29: I think this use of `--` is confusing for many people. I don't find it intuitive either.

```suggestion
            digits -= 1;
            buffer[digits] = bytes1(uint8(48 + uint256(value % 10)));
```

contracts/utils/Strings.sol
```diff
@@ -23,12 +25,41 @@ library Strings {
...
+     * @dev Converts a `uint256` to its ASCII `hexadecimal string` representation with fixed length.
+     */
+    function toHexString(uint256 value, uint256 length) internal pure returns (string memory) {
+        bytes memory buffer = new bytes(2 * length + 2);
+        buffer[0] = "0";
+        buffer[1] = "x";
+        for (uint256 i = 2 * length + 1; i > 1; --i) {
+            buffer[i] = alphabet[value & 0xf];
+            value >>= 4;
         }
```
frangio, line 61: What do you think of adding something like this?

```suggestion
        }
        require(value == 0, "Strings: hex length insufficient");
```

Amxx: Not sure.

frangio (approved):
LGTM

PR #2505, opened 2021-02-02
 Refactor ERC165 to use function overriding instead of storage.

Amxx:
ERC165: Uses virtual/override on the supportsInterface function instead of mapping

This reduces the storage footprint of ERC165 compliant contract, thus reducing the gas cost of deploying and configuring such contracts.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/mocks/ERC165Mock.sol
```diff
@@ -2,9 +2,9 @@
 
 pragma solidity ^0.8.0;
 
-import "../introspection/ERC165.sol";
+import "../introspection/ERC165Storage.sol";
 
-contract ERC165Mock is ERC165 {
+contract ERC165Mock is ERC165Storage {
```
frangio, line 7: This should now be `ERC165StorageMock`, and the test should be `ERC165Storage.test.js`, with `contract('ERC165Storage', ...`.

frangio (commented):
Looking great.

test/introspection/ERC165.test.js
```diff
@@ -9,6 +9,12 @@ contract('ERC165', function (accounts) {
     this.mock = await ERC165Mock.new();
   });
 
+  it('register interface', async function () {
```
frangio, line 12: I think this test should be removed since it pertains `ERC165Storage` and not plain `ERC165`.

contracts/mocks/ERC165Mock.sol
```diff
@@ -2,9 +2,9 @@
```
frangio, line 2: I think this entire file can be deleted.

frangio:
I've implemented the changes I suggested above.

contracts/mocks/ERC165Mock.sol
Amxx: But then `ERC165Storage` should not be abstract, right ?

frangio (approved):
LGTM!

PR #2508, opened 2021-02-04
Minimal support for ERC2771 (GSNv2)

### LINKED ISSUE #2258: GSNv2 support
**🧐 Motivation**
[GSNv2](https://docs.opengsn.org/learn/index.html) is under development by the OpenGSN team, and it's very similar to GSNv1 regarding recipient contract implementation. We could add support for both v2 and v1 with some minor changes.

**📝 Details**
GSNv2 moves the responsibility of deciding whether to pay for a meta-tx (ie `acceptRelayedCall`, along with `pre/post`) out of the Recipient into a [Paymaster contract](https://docs.opengsn.org/contracts/index.html#paymaster). The Recipient now only needs to implement the base `_msgSender` and `_msgData` helpers, which should use the injected address at the end of the CALLDATA (instead of msg.sender) if the call [comes from a **TrustedForwarder**](https://docs.opengsn.org/contracts/index.html#trusted_forwarder).

The TrustedForwarder has the responsibilities of validating the user signature, managing its nonce, appending the signer address to the CALLDATA, and forwarding the call. As with the RelayHub in GSNv1, the Recipient should assume they are receiving a meta-tx if it's sent by the TrustedForwarder. And as with the RelayHub, the TrustedForwarder can be a singleton, since it has no app-specific logic.

This approach is described in detail in [this ERC to be published](https://docs.google.com/document/d/1EnLof0KCtdnrqp07MN6m3TQWDKVrjYMKttL9yNW2pY4/edit?usp=sharing). Note that `getHubAddr` was changed to [`isTrustedForwarder`](https://github.com/opengsn/gsn/blob/master/contracts/interfaces/IRelayRecipient.sol#L15) in the Recipient internal interface.

To sum up, we need to:
- Refactor the _msgSender and _msgData methods out of [GSNRecipient](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/GSN/GSNRecipient.sol) into a MetaTransactionRecipient (preferrably including the ERC number once published).
- Add a TrustedForwarder implementation to Contracts so devs can easily set up a local meta-tx setup, and optionally include addresses to the singleton deployments.
- Clarify that GSNRecipient is only valid for GSNv1, since the [Paymaster interface](https://github.com/opengsn/gsn/blob/master/contracts/interfaces/IPaymaster.sol) has changed in v2.

Amxx:
This PR includes a ERC2771 compatible BaseRelayRecipient and a compatible, ERC712 based, relayer.

Fixes #2258

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


contracts/GSNv2/BaseRelayRecipient.sol
```diff
@@ -0,0 +1,37 @@
...
+        _trustedForwarder = trustedForwarder;
+    }
+
+    function isTrustedForwarder(address forwarder) public view virtual returns(bool) {
+        return forwarder == _trustedForwarder;
+    }
+
+    function _msgSender() internal view virtual override returns (address sender) {
+        if (isTrustedForwarder(msg.sender)) {
+            // return abi.decode(abi.encodePacked(bytes12(0), msg.data[msg.data.length-20:]), (address));
```
frangio, line 23:
```suggestion
            // The assembly code is more direct than the Solidity version using `abi.decode`.
```

contracts/GSNv2/BaseRelayRecipient.sol
```diff
@@ -0,0 +1,37 @@
...
+            assembly { sender := shr(96, calldataload(sub(calldatasize(), 20))) }
+        } else {
+            return Context._msgSender();
```
frangio, line 26: Using `super` feels like the right thing here. Let me know if you disagree though.

```suggestion
            return super._msgSender();
```

contracts/GSNv2/BaseRelayRecipient.sol
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+
+    function _msgData() internal view virtual override returns (bytes calldata) {
+        if (isTrustedForwarder(msg.sender)) {
+            return msg.data[:msg.data.length-20];
+        } else {
+            return Context._msgData();
```
frangio, line 34: As above.

```suggestion
            return super._msgData();
```

contracts/GSNv2/MinimalForwarder.sol
```diff
@@ -0,0 +1,57 @@
...
+            req.value,
+            req.gas,
+            req.nonce,
+            keccak256(req.data)
+        ))).recover(signature);
+        return _nonces[req.from] == req.nonce && signer == req.from;
+    }
+
+    function execute(ForwardRequest calldata req, bytes calldata signature) public payable returns (bool, bytes memory) {
+        require(verify(req, signature));
```
frangio, line 47:
```suggestion
        require(verify(req, signature), "MinimalForwarder: signature does not match request");
```

frangio (commented):
I don't know how to evaluate whether this `TrustedForwarder` implementation is compliant with the GSNv2 ERCs. I could only find a PR for an EIP that has been closed due to inactivity: https://github.com/ethereum/EIPs/pull/2770. In this draft there is something weird about extending the `ForwardRequest` type that I don't understand where it fits in the GSNv2 picture. The signature for the `verify` and `execute` functions are different from what we have here.

Do you know if these two functions got changed at some point?

Amxx:
> I don't know how to evaluate whether this `TrustedForwarder` implementation is compliant with the GSNv2 ERCs. I could only find a PR for an EIP that has been closed due to inactivity: [ethereum/EIPs#2770](https://github.com/ethereum/EIPs/pull/2770). In this draft there is something weird about extending the `ForwardRequest` type that I don't understand where it fits in the GSNv2 picture. The signature for the `verify` and `execute` functions are different from what we have here.
> 
> Do you know if these two functions got changed at some point?

AFAIK, there is no requierements about what the forwarder should or should not do. I went with a ERC712 metatx relayer, but anything is possible. My understanding is that the standard focus on the data encoding between the forwarder and the recipient. How the forwarder works, and how it verifies ownership before encoding data is not specified.

frangio:
There is at least a requirement for the `TrustedForwarder` to follow an interface that the RelayHub will use, which if I understand correctly consists of `verify` and `execute`. However, these two functions are not specified in the EIP that was merged. We should report this to the OpenGSN team to at least get some clarity, because otherwise we would be merging a feature that we can't be sure is spec-compliant.

frangio:
Actually, there is no EIP but I suppose the current de facto spec is the OpenGSN repo and it seems we're implementing a different `Forwarder` interface. Take a look at [`IForwarder.sol`](https://github.com/opengsn/forwarder/blob/bf3447b2e54c407a6e02e0bd2af858fe71e13711/contracts/IForwarder.sol).

forshtat:
We will get a verification audit of GSN v2 smart contracts, including Forwarder, in a very short time. I don't expect Forwarder to be affected, but still, after that I will update the ERCs with current interfaces.

In the meanwhile we accepted a proposal to add optional expiration date to the base `ForwardRequest` to prevent eternally valid signed messages (not on master yet):
https://github.com/opengsn/gsn/commit/5d562f86cdc31c98e9dd44a5631eb0bf42a1b4cc

frangio:
Thanks for the update @forshtat!

---

@Amxx Given that ERC2771 can be used independently of the GSN we should consider renaming the `GSNv2` directory. Perhaps something associated to meta-transactions would be a good name. One option is `metatx`.

contracts/metatx/BaseRelayRecipient.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "../utils/Context.sol";
+
+/*
+ * @dev Context variant with ERC2771 support.
+ */
+abstract contract BaseRelayRecipient is Context {
```
frangio, line 10: We want to change this name. Some candidates are:

- `MetaTxContext`
- `ForwarderContext`
- `ForwardedContext`
- `ERC2771Context`
- `RelayedContext`

contracts/metatx/MinimalForwarder.sol
```diff
@@ -0,0 +1,57 @@
...
+        return _nonces[req.from] == req.nonce && signer == req.from;
+    }
+
+    function execute(ForwardRequest calldata req, bytes calldata signature) public payable returns (bool, bytes memory) {
+        require(verify(req, signature), "MinimalForwarder: signature does not match request");
+        _nonces[req.from] = req.nonce + 1;
+
+        // solhint-disable-next-line avoid-low-level-calls
+        (bool success, bytes memory returndata) = req.to.call{gas: req.gas, value: req.value}(abi.encodePacked(req.data, req.from));
+        // Check gas: https://ronan.eth.link/blog/ethereum-gas-dangers/
```
frangio, line 52:
```suggestion
        // Validate that the relayer has sent enough gas for the call.
        // See https://ronan.eth.link/blog/ethereum-gas-dangers/
```

contracts/metatx/MinimalForwarder.sol
```diff
@@ -0,0 +1,57 @@
...
+        assert(gasleft() > req.gas / 63);
```
frangio, line 53: Do you understand why this is 63 and not 64?

Amxx: Lets note gas before call: `X`. After call we have `gasleft() = X - req.gas`

We know that at most `63X/64` are being forwarded so we need to ensure that `req.gas < 63X/64`
This becomes:
- `64req.gas < 63X`
- `req.gas < 63(X-req.gas)`
- `req.gas < 63gasleft()`
- `req.gas/63 < gasleft()`

Amxx: We check after, so that the check itself doesn't reduce the gas that should go in the call.

contracts/metatx/BaseRelayRecipient.sol
Amxx: I don't like `ForwarderContext` (might confuse people in thinking this is part of the forwarder)
No strong opinions for the others

contracts/metatx/MinimalForwarder.sol
Amxx: This is all documented in ronan's article (link in the comments)

frangio: Thanks, that was exactly the math I wanted to see. I hadn't seen it in Ronan's article.

frangio (approved):
Thank you!

PR #2512, opened 2021-02-05
Remove enumerable from AccessControl and add an AccessControlEnumerable extension

Amxx:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2139

EnumerableSet and EnumerableMap heavily rely on sstore & sload, which are very expensive operations. At the same time, the Enumerable aspect can be tracked offchain using tools like thegraph.

Developer should have the choice between having this enumerable functionality (and pay the gas price), or drop it if onchain access to these information is sufficient.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry


frangio:
Before making this decision we need to know what is the effect of having enumerability in AccessControl. What operations are affected and by how much.

For example, it seems that `hasRole` using `EnumerableMap` is only doing a single `sload`, which is the optimum.

Granting and revoking roles will be a lot more expensive with enumerability but these operations are not done often so it may be okay if they are not optimized for that.

Lastly, we also need to know the effect on contract size.

Amxx:
Cost of deploying the AccessControlMock (from eth-gas-reporter):
master: 580730
non-enumerable: 376083

Grant role: 
master: 84860
non-enumerable: 46762

Amxx:
Latest commit fixes #2139

contracts/access/AccessControlEnumerable.sol
```diff
@@ -0,0 +1,87 @@
...
+ * accounts that have a role's admin role can call {grantRole} and {revokeRole}.
+ *
+ * By default, the admin role for all roles is `DEFAULT_ADMIN_ROLE`, which means
+ * that only accounts with this role will be able to grant or revoke other
+ * roles. More complex role relationships can be created by using
+ * {_setRoleAdmin}.
+ *
+ * WARNING: The `DEFAULT_ADMIN_ROLE` is also its own admin: it has permission to
+ * grant and revoke this role. Extra precautions should be taken to secure
+ * accounts that have been granted it.
```
frangio, line 41: This can't be the same as `AccessControl`.

```suggestion
 * @dev Extension of {AccessControl} that allows enumerating the members of each role.
```

contracts/access/AccessControlEnumerable2.sol
```diff
@@ -0,0 +1,87 @@
...
+ * By default, the admin role for all roles is `DEFAULT_ADMIN_ROLE`, which means
+ * that only accounts with this role will be able to grant or revoke other
+ * roles. More complex role relationships can be created by using
+ * {_setRoleAdmin}.
+ *
+ * WARNING: The `DEFAULT_ADMIN_ROLE` is also its own admin: it has permission to
+ * grant and revoke this role. Extra precautions should be taken to secure
+ * accounts that have been granted it.
+ */
+abstract contract AccessControlEnumerable2 is AccessControl {
```
frangio, line 43: I don't recall if we discussed something about this one. Three variants of the same contract feels like too much.

Can we remove this from this PR and discuss after the preview release whether we want to include this other kind of enumerability into `AccessControlEnumerable`?

contracts/access/AccessControl.sol
```diff
@@ -42,18 +40,11 @@ import "../utils/Context.sol";
...
 abstract contract AccessControl is Context {
-    using EnumerableSet for EnumerableSet.AddressSet;
-    using Address for address;
-
-    struct RoleData {
-        EnumerableSet.AddressSet members;
-        bytes32 adminRole;
-    }
-
-    mapping (bytes32 => RoleData) private _roles;
```
frangio, line 53: Was there a reason to remove the `RoleData` struct and replace this mapping with the two separate mappings below? I would've kept it, just changing `members` into a mapping.

contracts/access/AccessControl.sol
```diff
@@ -133,7 +100,7 @@ abstract contract AccessControl is Context {
      * - the caller must have ``role``'s admin role.
      */
     function grantRole(bytes32 role, address account) public virtual {
-        require(hasRole(_roles[role].adminRole, _msgSender()), "AccessControl: sender must be an admin to grant");
+        require(AccessControl.hasRole(AccessControl.getRoleAdmin(role), _msgSender()), "AccessControl: sender must be an admin to grant");
```
frangio, line 103: What is the reason to use the fully qualified `AccessControl` functions?

If we're making these functions virtual I don't see a reason not to respect their overrides here.

The same goes for the rest of this contract.

Amxx: I'd say consistency with other contract. In ERC721, we decided that view function should be virtual BUT that overloading them should not affect the parent behavior (ownerOf for example).

I don't see why we should use fully qualified functions in ERC721 but not in AccessControl

frangio: Because ERC721 has more complicated internal invariants, and numbers have to add up properly while booleans like `hasRole` are trivial for us to support overriding. I know this doesn't feel like a consistent rule but I think it's an acceptable compromise.

contracts/access/AccessControl.sol
```diff
@@ -2,13 +2,11 @@
...
 pragma solidity ^0.8.0;
 
-import "../utils/EnumerableSet.sol";
-import "../utils/Address.sol";
 import "../utils/Context.sol";
 
 /**
  * @dev Contract module that allows children to implement role-based access
- * control mechanisms.
+ * control mechanisms. Light, not enumerable, version.
```
frangio, line 9:
```suggestion
 * control mechanisms. This is a lightweight version that doesn't allow enumerating role
 * members except through off-chain means by accessing the contract event logs. Some
 * applications may benefit from on-chain enumerability, for those cases see
 * {AccessControlEnumerable}.
```

frangio (dismissed):
Looks good! I left an improvement for the docs.

We also need to add `{{AccessControlEnumerable}}` in `access/README.adoc`.

frangio (dismissed):

frangio (approved):

PR #2516, opened 2021-02-08
Add a VTableProxy

Amxx:
Prototype of VTable based proxy. This proxy redirect to different module, depending on the `msg.sig` and on an internal bytes4→address mapping. VTable also include an owner for update, which is separate from any Ownable inheritance of the modules/implementations.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Closing for cleanup since we don't plan to merge this soon.

PR #2517, opened 2021-02-08
Add an interface folder that lists common interfaces

Amxx:
Includes:

- ERC20: ERC-20 Token Standard (final)
- ERC165: ERC-165 Standard Interface Detection (final)
- ERC721: ERC-721 Non-Fungible Token Standard (final) → includes IERC721Enumerable and IERC721Metadata
- ERC777: ERC777 Token Standard (final)
- ERC1155: ERC-1155 Multi Token Standard (final) → includes IERC1155MetadataURI, IERC1155Receiver
- ERC1271: Standard Signature Validation Method for Contracts (last call)
- ERC1363: ERC-1363 Payable Token (final) → should be implemented as an ERC20 extension
- ERC1820: Pseudo-introspection Registry Contract (final)
- ERC2612: permit – 712-signed approvals (draft) → base of ERC20Permit
- ERC2981: NFT Royalty Standard (final)
- ERC3156: Flash Loans (final)

Eventually, all of https://eips.ethereum.org/erc should be considered

#### PR Checklist

- [x] Documentation
- [x] Changelog entry


frangio:
Since this is not a breaking change we will discuss this after 4.0 or at least after we've finished work on the breaking changes.

frangio:
What should we do with the interface we have elsewhere? Should we deduplicate them?

Amxx:
> What should we do with the interface we have elsewhere? Should we deduplicate them?

We just cannot move them for obvious backward compatibility reasons. Yet, I think they should be in here. So my approach has been to link them using `import` statements. The other option I see, is to move the interface file here, and have the links where they originally were ... but my guess is you won't like it.

Any preference/opinion ?

frangio:
Oh sorry I missed that there were `import` statements to the existing interfaces.

Well what are we going to do for future ERCs we implement? Are we going to add two different files for each new ERC interface? It seems like we should instead add all new interfaces in the interfaces directory, and import them from there.

Amxx:
> It seems like we should instead add all new interfaces in the interfaces directory

Yes. This is what I did for ERC1271 in signaturechecker

frangio:
So given that any future interfaces are going to be added in the interfaces directory, I would move the current interfaces over there for consistency, and place imports in the current locations for backwards compatibility.

frangio:
@Amxx Please see my comment above.

frangio (approved):
LGTM!

PR #2518, opened 2021-02-09
Improve gas efficiency of EnumerableMap

Amxx:
**EnumerableSet**:

- use of inline assembly to optimise `add`
- use of inline assembly to optimise `remove`

**EnumerableMap:**

- Using an EnumerableSet for keys management + simple `key => value` mapping
- Optimizes for key lookup rather then index lookup

**Gas report**
*Note: gas report doesn't look to be such a good indication. tests on remix show the assembly to be more efficient, but its difficult to quantify.*

|Function|Min|Max|Average|
|-|-|-|-|
|`EnumerableSetMock.add`|23630 → 23630|85403 → 85403|65740 → 65740|
|`EnumerableSetMock.remove`|22460 → 21450|26666 → 25656|25193 → 24471|
|`EnumerableMapMock.set`|25809 → 25038|105902 → 106023|87441 → 87400|
|`EnumerableMapMock.remove`|23592 → 24023|32502 → 28229|29053 → 26577|

#### PR Checklist

- [x] Tests
- [x] Changelog entry


contracts/utils/EnumerableSet.sol
```diff
@@ -53,10 +53,19 @@ library EnumerableSet {
      */
     function _add(Set storage set, bytes32 value) private returns (bool) {
         if (!_contains(set, value)) {
-            set._values.push(value);
-            // The value is stored at length-1, but we add 1 to all indexes
-            // and use 0 as a sentinel value
-            set._indexes[value] = set._values.length;
+            uint256 last;
+            assembly {
```
frangio, line 57: I feel too uneasy about this assembly block. I prefer to have an extra `sload`.

If we want to insist on the optimization, we should use as little assembly as possible:

```solidity
    function push(bytes32[] storage array, bytes32 value) internal returns (uint256) {
        uint arraySlot;
        uint[1] storage arrayLength;
        assembly {
            arraySlot := array.slot
            arrayLength.slot := array.slot
        }
        uint arrayLastIndex = arrayLength[0];
        arrayLength[0] += 1;
        uint arrayLastSlot = arrayLastIndex + uint256(keccak256(abi.encodePacked(arraySlot)));
        bytes32[1] storage arrayLast;
        assembly {
            arrayLast.slot := arrayLastSlot
        }
        arrayLast[0] = value;
        return arrayLastIndex;
    }
```

However, I still worry about the clarity of all this.

contracts/utils/EnumerableSet.sol
```diff
@@ -74,27 +83,21 @@ library EnumerableSet {
...
-            set._values[toDeleteIndex] = lastvalue;
-            // Update the index for the moved value
-            set._indexes[lastvalue] = toDeleteIndex + 1; // All indexes are 1-based
-
-            // Delete the slot where the moved value was stored
-            set._values.pop();
-
-            // Delete the index for the deleted slot
+            bytes32 tmp;
+            assembly {
```
frangio, line 88: Same concerns as above.

frangio (commented):
I have concerns about the use of assembly but the changes to `EnumerableMap` look good.

contracts/utils/EnumerableMap.sol
```diff
@@ -173,9 +131,9 @@ library EnumerableMap {
      * message unnecessarily. For custom revert reasons use {_tryGet}.
      */
     function _get(Map storage map, bytes32 key, string memory errorMessage) private view returns (bytes32) {
```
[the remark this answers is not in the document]
Amxx: Well, they are private function in a library, so there is no real cost to keep them. I thought that between deprecating and removing, we should let a least one major version ... and there were just deprecated in 3.4.

contracts/utils/EnumerableSet.sol
```diff
@@ -53,10 +53,19 @@ library EnumerableSet {
      */
     function _add(Set storage set, bytes32 value) private returns (bool) {
         if (!_contains(set, value)) {
-            set._values.push(value);
-            // The value is stored at length-1, but we add 1 to all indexes
-            // and use 0 as a sentinel value
-            set._indexes[value] = set._values.length;
+            uint256 last;
+            assembly {
+                let p := sload(0x40)
```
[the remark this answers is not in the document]
Amxx: good catch, it frightening that this did not cause any critical failure in the tests

contracts/utils/EnumerableSet.sol
Amxx: yes, clarity is not great

frangio: Then I think we should keep the previous code even if there's a redundat `SLOAD` of the length.

frangio:
This is good but needs a changelog.

frangio (approved):
Looks good. Thank you!

PR #2532, opened 2021-02-19
Add tooling to verify signatures with support for ERC1271

### LINKED ISSUE #2535: Tools for ERC1271 signature verification
**🧐 Motivation**
Smart wallet are becoming more and more common. Recovery of signature is supported by contracts through the ECDSA library, but this only works for EOA. Smart Wallet rely on ERC1271 to perform signature verification, but this is not currently support by openzeppelin/contracts 

**📝 Details**
Add a `SignatureChecker` library, that given an expected signer (address), a hash (bytes32) and a signature (bytes), will transparently support traditional EOA signature, or resolve to ERC1271 if the expected signer is a smart contract.

Amxx:
SignatureChecker is a library for seamless verification of signatures with support for both EOA (ECDSA) and smart contract wallets (ERC1271).

While still being a draft, this ERC is supported by smart contract wallets such as Argent.

I believe this feature could greatly improve the compatibility of dapps that relies on message signatures with such wallets. In particular since Argent is now able to sign messages (including ERC721 typedData) through WalletConnect.

Continuation of #2459
FIxes #2535 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.6.0 <0.9.0;
+
+import "./ECDSA.sol";
+import "../Address.sol";
+import "../../interfaces/IERC1271.sol";
+
+library SignatureChecker {
+    function isValidSignature(address signer, bytes32 hash, bytes memory signature) internal view returns (bool) {
```
frangio, line 10: As we discussed offline, ERC1271 signatures that are valid at some point can become invalid in the future, and this is an important difference when compared to cryptographic signatures, which have the non-repudiation property.

I proposed renaming this function to something that highlights this fact. Options:

- `isCurrentlyValidSignature`
- `isValidSignatureNow`

Amxx: I agreed with that comment until I realized that the IERC1271 function is `isValidSignature`. Basically, this is where I got the name from (even though they are actually 2 different functions signature.

I think it would have made sense to rename the IERC1271 function back in the days before it got adopted/implemented by smart wallets. But since this is the underlying ERC name, I think we should match it somehow (`checkIsValidSignature` ?)

frangio: I don't see why the library name should necessarily be the same as the ERC function name, since it's a utility library and not an implementation of the ERc.

contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.6.0 <0.9.0;
+
+import "./ECDSA.sol";
+import "../Address.sol";
+import "../../interfaces/IERC1271.sol";
+
+library SignatureChecker {
```
frangio, line 9: Missing docs! Remember adding in README.adoc as well.

frangio, line 9: Missing docs! We should make particular emphasis on why the function is called "...Now".

contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -0,0 +1,21 @@
...
+    function isValidSignatureNow(address signer, bytes32 hash, bytes memory signature) internal view returns (bool) {
+        if (Address.isContract(signer)) {
+            try IERC1271(signer).isValidSignature(hash, signature) returns (bytes4 magicValue) {
+                return magicValue == IERC1271(signer).isValidSignature.selector;
+            } catch {
+                return false;
+            }
+        } else {
+            return ECDSA.recover(hash, signature) == signer;
```
frangio, line 18: From our recommendation in `ECDSA`:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5ecd2b850c2acb52b94d8374ac8d22cb543e71c8/contracts/utils/cryptography/ECDSA.sol#L20-L24

We should be providing some other link in the docs, but there is more info in [this article](https://jimmysong.medium.com/faketoshis-nonsense-signature-8700a44536b5).

So, is this function really safe? Isn't it vulnerable to crafting `signature` and `hash` values that will recover to the `signer` but which do not correspond to a real ECDSA signature?

contracts/interfaces/IERC1271.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.6.0 <0.9.0;
+
+interface IERC1271 {
```
frangio, line 5: Missing docs. Should contain a link to the EIP page.

contracts/utils/cryptography/SignatureChecker.sol
Amxx: > So, is this function really safe? Isn't it vulnerable to crafting signature and hash values that will recover to the signer but which do not correspond to a real ECDSA signature?

Our own ECDSA.recover doesn't enforce that. You can technically use it to recover a hash that was not properly formated. The same goes for a smart wallet 1271 implementation (the code in `isValidSignature`). In both cases, we cannot do anything about it (the hash might be 191, or 712, or anything not yet standardized. The best we can do is adding the warning you mentioned in the interface for 1271.

frangio: Ok, for some reason I thought this API was more vulnerable to misuse, but now that I'm looking at it again it I can see that it's the same.

PR #2556, opened 2021-03-03
Separate "production" and "testing" mode

Amxx:
Using the MODE=PROD envvar will not compile the mocks, and enable optimisations. Mocks are now longer part of the 'contract' folder.

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


hardhat.config.js
```diff
@@ -11,16 +11,20 @@ for (const f of fs.readdirSync(path.join(__dirname, 'hardhat'))) {
 }
 
 const enableGasReport = !!process.env.ENABLE_GAS_REPORT;
+const enableProduction = process.env.MODE === 'PROD';
```
frangio, line 14:
```suggestion
const enableProduction = process.env.COMPILE_MODE === 'production';
```

Rationale: Using the value `'production'` which is very common with stuff like Webpack. They use the variable `NODE_ENV` but I don't think that's very accurate for this case. I changed to `COMPILE_MODE` to be a little more specific.

hardhat.config.js
```diff
@@ -11,16 +11,20 @@ for (const f of fs.readdirSync(path.join(__dirname, 'hardhat'))) {
...
 const enableGasReport = !!process.env.ENABLE_GAS_REPORT;
+const enableProduction = process.env.COMPILE_MODE === 'production';
 
 /**
  * @type import('hardhat/config').HardhatUserConfig
  */
 module.exports = {
+  paths: {
+    sources: enableProduction ? 'contracts' : 'test/mocks',
+  },
```
frangio, line 22: @cgewecke We tried to do this to remove `mocks` from our `contracts` directory, but as a consequence solidity-would only measure coverage in the mocks directory and ignoring the `contracts` directory.

I assume this is the same mechanism to avoid measuring coverage of dependencies in `node_modules`. Is there any way that we could make this setup work?

frangio:
Closing since this setup is incompatible with coverage.

PR #2557, opened 2021-03-03
Enable optimizations when publishing package

Amxx:
Enable solidity optimisations through COMPILE_MODE=production.

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio (approved):
Thanks! I think this will work.

PR #2565, opened 2021-03-05
Add an internal _useNonce(address) function in ERC20Permit

### LINKED ISSUE #2564: Add an internal _useNonce(address) function in ERC20Permit
**🧐 Motivation**
The ERC20Permit contract is great in that it provide a building block which is both ERC20 & ERC712 compliant. It also include some features, like nonce management, which are not by default in ERC712 but are required to do proper signature checking in the permit function.

However, when wanting to expand the ERC712 compliant features beyound the simple `permit` function, it is not possible to increment the counter (private slot). This means that if I wanted to add a `transferFromBySig` or any other feature using ERC712, I would have to use an independent nonce. This is really not great.

**📝 Details**
A simple, yet effective, solution is to add an internal function that lets you "consumed" a nonce:

```
function _useNonce(address owner) internal  virtual returns (uint256 current) {
    Counters.Counter storage nonce = _nonces[owner];
    current = nonce.current();
    nonce.increment();
}
```

Since this function only increment, it doesn't create any security issue that would allow signed messages to be replayed.

Amxx:
Fixes #2564

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
  * `IERC20Metadata`: add a new extended interface that includes the optional `name()`, `symbol()` and `decimals()` functions. ([#2561](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2561))
  * `ERC777`: make reception acquirement optional in `_mint`. ([#2552](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2552))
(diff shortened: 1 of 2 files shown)


frangio:
`nonces` should be made virtual now that it is possible to override `_useNonce`.

frangio (approved):
LGTM!

PR #2568, opened 2021-03-06
Make some private functions internal to allow the developpement of "withSignature" functions (like permit)

### LINKED ISSUE #2567: Make some private functions internal to allow the development of "withSignature" functions (like permit)
**🧐 Motivation**
It seems to be a global trend to allow some "priviledged" operation to be performed by a third party provided they can show a valid signature by the legitimate account. For example, the `ERC20Permit` contract allows anyone to forward a signed message (ERC712) that will set token approval. The comp governance token offers a similar feature to delegate vote using a signature.

This mechanism is very powerfull, and we can expect more of these "with signature" function to be implemented.

As `ERC20Permit` shows, this is easy to do when an internal function is available (`_approve`), and can be called when the signature is verified. This is something that is widely available in the token smart contract.

However, some of these functions are either non-existing, or private, which prevent the implementation of such "with signature" functions. In particular I can think of:

- `AccessControl.grantRole`: The `msgSender()` must be admin, and the underlying function `_grantRole` is private
- `AccessControl.revokeRole`: The `msgSender()` must be admin, and the underlying function `_revokeRole` is private
- `AccessControl.renounceRole`: The `msgSender()` must be the account, and the underlying function `_revokeRole` is private
- `Ownable.transferOwnership`: The `msgSender()` must be owner, and there is no equivalent `_transferOwnership` internal function

**📝 Details**
In order to improve the possibility to write such wrappers, I believe:
- `_grantRole`, `_revokeRole` and `_revokeRole` should be internal
- There should be an internal `Ownable._transferOwnership` that changes the owner and emit an event, without doing any msgSender() check.

Amxx:
Fixes #2567

#### PR Checklist
- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio:
Being discussed in #2567.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+ * `Ownable`: add `_transferOwnership` to Ownable. ([#2568](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/#2568))
+ * `AccessControl`: make some private functions internal. ([#2568](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/#2568))
```
frangio, line 6: Should specify which functions.

contracts/access/AccessControl.sol
```diff
@@ -194,14 +194,14 @@ abstract contract AccessControl is Context, IAccessControl, ERC165 {
...
-    function _grantRole(bytes32 role, address account) private {
+    function _grantRole(bytes32 role, address account) internal virtual {
         if (!hasRole(role, account)) {
             _roles[role].members[account] = true;
             emit RoleGranted(role, account, _msgSender());
         }
     }
 
-    function _revokeRole(bytes32 role, address account) private {
+    function _revokeRole(bytes32 role, address account) internal virtual {
```
frangio, line 204: Grant and revoke need documentation.

frangio:
Sorry to come back to this but I have one remaining concern. At some point we thought it was not such a good idea to have a single underscore determine such different semantics. I think we should try to have different names... the name `_setupRole` was pretty good in that regard I think. We can have `_setupOwner`, but I don't know how we would handle revoke, unless we do `_setupRole(..., false)`.

Amxx:
I personally think that a prefix underscore is clearly a mark of "this is internal, with no access check".

We already use this pattern in a lot of contracts, particularly `transfer`/`_transfer` and `approve`/`_approve`.
I think the logic is clear, and the discoverability is way better that way than if we use a different name.

frangio (approved):

Mehduzza420:
How can I retrieve my account since it has been severely compromised and I cannot access my account to transfer to my fiat wallet or financial institution

Amxx:
> How can I retrieve my account since it has been severely compromised and I cannot access my account to transfer to my fiat wallet or financial institution

Hello @Mehduzza420 

This is not the place for this question. You should try your wallet's community channel.

Mehduzza420:
OK thank you

Mehduzza420:
> I personally think that a prefix underscore is clearly a mark of "this is internal, with no access check".
> 
> We already use this pattern in a lot of contracts, particularly `transfer` vs `_transfer` and `approve` vs `_approve`.
> I think the logic is clear, and the discoverability is way better that way than if we use a different name.

Hello I have been trying to ask for help regarding the same issue... I have been severely hacked and been contacted by several fake team members from circle regarding my transfer

Mehduzza420:
Hello the problem I have with this pull request is that how can pas with no checks and is it possible to have two owners on the pull request?  Or is it possible to have one being an actor or doing a PR on behalf of the owner?

Mehduzza420:
Hello the problem I have with this pull request is that how can pas with no checks and is it possible to have two owners on the pull request?  Or is it possible to have one being an actor or doing a PR on behalf of the owner?

PR #2576, opened 2021-03-08
Make ERC1155.uri public

### LINKED ISSUE #2574: Make ERC1155 uri() function public instead external
uri() function in ERC1155.sol is declared external and subclasses overriding uri() cannot call super.uri().

**💻 Environment**

hardhat

**📝 Details**

When overwriting uri() implemented here:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/ERC1155.sol#L58

it is not possible to call super.uri() because of the external "modifier". uri() is the only way to access the _uri private member in ERC1155.sol which is set in ctor.

**🔢 Code to reproduce bug**

contract test is ERC1155 {
  function uri(uint256 tokenId) external view virtual override returns (string memory) {
    // return URI for special cases
    //fallback:
    return super.uri().  <- compilation failed because uri() can only be called from external
  }

Amxx:
Fixes #2574

#### PR Checklist
- [x] Changelog entry


frangio (approved):
Thanks!

PR #2582, opened 2021-03-12
Add support for EIP2098 "short signatures" in the ECDSA library

### LINKED ISSUE #2539: Add support for EIP2098 "short signatures" in the ECDSA library
**🧐 Motivation**
Secp256k1 signatures produced by EOA to sign transaction and messages are usually encode on 65 bytes:
- R: 32 bytes
- S: 32 bytes
- V: 1 bytes.

However, the V parameter can be reduced to only one single bit, and the S parameter only requiers 255 bits to be encoded (due to parity constraints). This result in the signature space fitting inside 64 bytes! [EIP2098](https://eips.ethereum.org/EIPS/eip-2098) describes how to pack the R,S,V values into a shorter R,VS format tat would take less space and thus me less expensive to transmit/store

Ethers.js already include tooling to produce such signatures.

**📝 Details**
EIP details are available [here](https://eips.ethereum.org/EIPS/eip-2098).
Implementation in the ECDSA only requires to accept 65 & 64 bytes long signatures, split them into R,S,V values, and call the recover method. Since the changes that only in the "bytes unpacking" part of the function, and the actual cryptography check done in ECDSA.recover are not modified or bypassed, there should be no security concern. 

```
function recover(bytes32 hash, bytes memory signature) internal pure returns (address) {
    // Divide the signature in r, s and v variables
    bytes32 r;
    bytes32 s;
    uint8 v;

    // Check the signature length
    // - case 65: r,s,v signature (standard)
    // - case 64: r,vs signature (cf https://eips.ethereum.org/EIPS/eip-2098)
    if (signature.length == 65) {
        // ecrecover takes the signature parameters, and the only way to get them
        // currently is to use assembly.
        assembly {
            r := mload(add(signature, 0x20))
            s := mload(add(signature, 0x40))
            v := byte(0, mload(add(signature, 0x60)))
        }
    } else if (signature.length == 64) {
        // ecrecover takes the signature parameters, and the only way to get them
        // currently is to use assembly.
        assembly {
            r := mload(add(signature, 0x20))
            s := and(mload(add(signature, 0x40)), 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
            v := add(shr(7, byte(0, mload(add(signature, 0x40)))), 27)
        }
    } else {
        revert("ECDSA: invalid signature length");
    }

    return recover(hash, v, r, s);
}
```

Amxx:
Fixes #2539

#### PR Checklist
- [x] Tests
- [x] Changelog entry

### DIFF
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -49,7 +49,7 @@ library ECDSA {
                 let vs := mload(add(signature, 0x40))
                 r := mload(add(signature, 0x20))
                 s := and(vs, 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
-                v := add(shr(vs, 255), 27)
+                v := add(shr(255, vs), 27)
             }
         } else {
             revert("ECDSA: invalid signature length");
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -49,7 +49,7 @@ library ECDSA {
                 let vs := mload(add(signature, 0x40))
                 r := mload(add(signature, 0x20))
                 s := and(vs, 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
-                v := add(shr(7, byte(0, vs)), 27)
+                v := add(shr(vs, 255), 27)
             }
         } else {
             revert("ECDSA: invalid signature length");
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -24,23 +24,34 @@ library ECDSA {
      * be too long), and then calling {toEthSignedMessageHash} on it.
      */
     function recover(bytes32 hash, bytes memory signature) internal pure returns (address) {
-        // Check the signature length
-        if (signature.length != 65) {
-            revert("ECDSA: invalid signature length");
-        }
-
         // Divide the signature in r, s and v variables
         bytes32 r;
         bytes32 s;
         uint8 v;
 
-        // ecrecover takes the signature parameters, and the only way to get them
-        // currently is to use assembly.
-        // solhint-disable-next-line no-inline-assembly
-        assembly {
-            r := mload(add(signature, 0x20))
-            s := mload(add(signature, 0x40))
-            v := byte(0, mload(add(signature, 0x60)))
+        // Check the signature length
+        // - case 65: r,s,v signature (standard)
+        // - case 64: r,vs signature (cf https://eips.ethereum.org/EIPS/eip-2098)
+        if (signature.length == 65) {
+            // ecrecover takes the signature parameters, and the only way to get them
+            // currently is to use assembly.
+            // solhint-disable-next-line no-inline-assembly
+            assembly {
+                r := mload(add(signature, 0x20))
+                s := mload(add(signature, 0x40))
+                v := byte(0, mload(add(signature, 0x60)))
+            }
+        } else if (signature.length == 64) {
+            // ecrecover takes the signature parameters, and the only way to get them
+            // currently is to use assembly.
+            // solhint-disable-next-line no-inline-assembly
+            assembly {
+                r := mload(add(signature, 0x20))
+                s := and(mload(add(_sign, 0x40)), 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
+                v := add(shr(7, byte(0, mload(add(_sign, 0x40)))), 27)
+            }
+        } else {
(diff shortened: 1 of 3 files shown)


frangio:
Waiting for a reply about the EIP becoming Final. https://github.com/ethereum/EIPs/issues/2440

test/utils/cryptography/ECDSA.test.js
```diff
@@ -8,6 +8,22 @@ const ECDSAMock = artifacts.require('ECDSAMock');
 const TEST_MESSAGE = web3.utils.sha3('OpenZeppelin');
 const WRONG_MESSAGE = web3.utils.sha3('Nope');
 
+function to2098Format (signature) {
+  const long = web3.utils.hexToBytes(signature);
+  expect(long.length).to.be.equal(65);
+  const short = long.slice(0, 64);
+  short[32] |= (long[64] % 27) << 7;
```
frangio, line 15:
```suggestion
  short[32] |= (long[64] % 27) << 7; // set the first bit of the 32nd byte to the v parity bit
```

test/utils/cryptography/ECDSA.test.js
```diff
@@ -8,6 +8,22 @@ const ECDSAMock = artifacts.require('ECDSAMock');
...
+  return web3.utils.bytesToHex(short);
+}
+
+function from2098Format (signature) {
+  const short = web3.utils.hexToBytes(signature);
+  expect(short.length).to.be.equal(64);
+  short.push((short[32] >> 7) + 27);
+  short[32] &= (1 << 7) - 1;
```
frangio, line 23:
```suggestion
  short[32] &= (1 << 7) - 1; // zero out the first bit of 1 the 32nd byte
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -24,23 +24,35 @@ library ECDSA {
...
+            }
+        } else if (signature.length == 64) {
+            // ecrecover takes the signature parameters, and the only way to get them
+            // currently is to use assembly.
+            // solhint-disable-next-line no-inline-assembly
+            assembly {
+                let vs := mload(add(signature, 0x40))
+                r := mload(add(signature, 0x20))
+                s := and(vs, 0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
+                v := add(shr(7, byte(0, vs)), 27)
```
frangio, line 52: Cheaper and closer to the sample Python code in EIP-2098.

```suggestion
                v := add(shr(vs, 255), 27)
```

test/utils/cryptography/ECDSA.test.js
```diff
@@ -61,6 +77,15 @@ contract('ECDSA', function (accounts) {
           await expectRevert(this.ecdsa.recover(TEST_MESSAGE, signature), 'ECDSA: invalid signature \'v\' value');
         });
       });
+
+      context('short 2098 format', function () {
```
frangio, line 81:
```suggestion
      context('short EIP2098 format', function () {
```

test/utils/cryptography/ECDSA.test.js
```diff
@@ -93,6 +118,15 @@ contract('ECDSA', function (accounts) {
...
```
frangio, line 122:
```suggestion
      context('short EIP2098 format', function () {
```

test/utils/cryptography/ECDSA.test.js
```diff
@@ -61,6 +77,15 @@ contract('ECDSA', function (accounts) {
...
+        it('works', async function () {
```
frangio, line 82: A lot of the tests in this file have `it('works'` and `it('reverts'` as the only tests in a describe block. This is really unnecessary, we should be writing the test directly as in: `it('recovers EIP2098 format', ...` and `it('reverts with ...`.

Can you clean this up while we're working on this file?

frangio:
We've decided to move forward with this PR even though the corresponding EIP is a Draft because we don't see how it could change in a backwards incompatible way, and it seems to be on the path to be Finalized.

frangio (commented):
Ready to merge after my comments here are addressed.

test/utils/cryptography/ECDSA.test.js
Amxx: Lets discuss that in sync. You know I'm bad with names, and its not just about the new tests, its also about the old ones...

frangio: I'm ok with changing to `works with <name of describe>` and `reverts with <name of describe>`. If you don't feel comfortable with it let's move ahead and merge if this is ready and we can talk about the test titles afterwards.

frangio:
There should be some documentation about the format of the `signature` argument. We don't even have documentation for the long format...

frangio (approved):
Thank you!

PR #2583, opened 2021-03-12
add ERC20SnapshotBlockNumber

### LINKED ISSUE #2460: ERC20Snapshot: add getting balanceOfAt using block number
**🧐 Motivation**

Add mechanism to get ERC20Snapshot `balanceOfAt` using block number, as current implementation uses snapshot ID which isn't usable unless snapshot ID is known by contracts.

This would allow onchain governance projects to more easily use ERC20Snapshot.

**📝 Details**

The current implementation of `ERC20Snapshot` exposes a function [`balanceOfAt`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/90ed1af972299070f51bf4665a85da56ac4d355e/contracts/token/ERC20/ERC20Snapshot.sol#L92) that requires a `snapshotId` instead of [`_blockNumber`](https://github.com/Giveth/minime/blob/ea04d950eea153a04c51fa510b068b9dded390cb/contracts/MiniMeToken.sol#L283) as in the original `minime` token. As a consequence, these snapshots are not accessible by smart contracts which have no idea about the snapshot ID. Furthermore these external contracts also don't have a way of looking for them as the `_accountBalanceSnapshots` are private.

Since this would enable a larger amount of on-chain governance projects to leverage this feature I would suggest to add a function which implements the binary search and takes a block number similar to minime.

Initially raised in the forum: https://forum.openzeppelin.com/t/erc20-snapshot-not-usable-for-smart-contracts/5310

Amxx:
Fixes #2460

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Closing in favor of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2632.

PR #2586, opened 2021-03-12
Add tests to extend code coverage

Amxx:
Add tests that cover some uncovered lines.

#### PR Checklist
- [x] Tests


Amxx:
The remaining coverage holes are:

- `proxy/Clones.sol:33`: cannot happen, create should never fail (unlike create2)
- `proxy/transparent/ ProxyAdmin.sol`: need testing when ProxyAdmin targets an invalid proxy
- `utils/cryptography/ draft-EIP712.sol:68`: need to change the chained to simulate work, not possible with current tooling.

frangio (approved):
Thank you!

PR #2597, opened 2021-03-16
Make overide of ERC721.isApprovedForAll affect the ERC721 logic

### LINKED ISSUE #2594: Improve ERC721 design for OpenSea proxy compatibility
**🧐 Motivation**

OpenSea encourages ERC721 developer to include an additional "proxies" approval to smoothen the user experience. This is made available in a [dedicated repo](https://github.com/ProjectOpenSea/opensea-creatures/blob/master/contracts/ERC721Tradable.sol#L64-L79) that is unfortunately still using solidity 0.5.0.

**📝 Details**

The ERC721 tradable mechanism is implemented through a hardcoded `proxyRegistryAddress`

```
contract ProxyRegistry {
    mapping(address => OwnableDelegateProxy) public proxies;
}
```

that is used when checking approval for all:

```
function isApprovedForAll(address owner, address operator)
    public
    view
    returns (bool)
{
    // Whitelist OpenSea proxy contract for easy trading.
    ProxyRegistry proxyRegistry = ProxyRegistry(proxyRegistryAddress);
    if (address(proxyRegistry.proxies(owner)) == operator) {
        return true;
    }
     return super.isApprovedForAll(owner, operator);
}
```

Still, reproducing this in our latest version (3.4.0 and 4.0) is made difficult by the restriction we enforce on virtual overloading:
- in addition to being public, `isApprovedForAll` is used in 2 functions (`approve` and `_isApprovedOrOwner`). However, in both cases, we force the use on the base implementation, meaning that overloading `isApprovedForAll` has no effect on the contract behavior. (this create a "gap" between what the contract advertizes, and what it actually does, which I see as an issue)
- we could also overload `_isApprovedOrOwner`, but this requires extra `ownerOf` calls, meaning unnecessary sload :/
- even with both functions overloaded, `approve` would still not behave correctly, and would also need to be overloaded. Due to its design, this last overloading cannot rely on a super call, and must thus the entirely rewritten.

We can see that, while this update was almost trivial on early versions of contracts, the inheritance restrictions that are currently part of the codebase are a pain to work with, and are more likely to cause issues due to inconsistent behavior between functions than to actually protect end-users. Even advanced users can have a hard time realizing the complexity of it, particularly if they are used to import contracts and not have a deep look at it.
Hopefully, they have good exhaustive tests and see that before deploying.

My proposition would be, at least for the `isApprovedForAll`, to use the overloaded version of the public function for the internal logic. `ownerOf` is complex since it returns an address that is supposed to match the internal book keeping, but `isApprovedForAll` is just a boolean flag that IMO should be easily overloaded.

Amxx:
Fixes #2594

#### PR Checklist

- [x] Tests
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -113,7 +113,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         address owner = ERC721.ownerOf(tokenId);
         require(to != owner, "ERC721: approval to current owner");
 
-        require(_msgSender() == owner || ERC721.isApprovedForAll(owner, _msgSender()),
+        require(_msgSender() == owner || isApprovedForAll(owner, _msgSender()),
             "ERC721: approve caller is not owner nor approved for all"
         );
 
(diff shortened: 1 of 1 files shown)


frangio (approved):
LGTM!

PR #2599, opened 2021-03-19
Draft: Add return values to Counters

k06a:
Idea to improve usability of `Counters`. Let's discuss if it is a good idea.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/Counters.sol
+++ b/contracts/utils/Counters.sol
@@ -22,17 +22,32 @@ library Counters {
         return counter._value;
     }
 
-    function increment(Counter storage counter) internal {
+    function increment(Counter storage counter) internal returns (uint256) {
         unchecked {
-            counter._value += 1;
+            return counter._value++;
+        }
+    }
+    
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @k06a, and thanks for the proposal.

I really like the idea of having `increment` and `decrement` return uint256 value of the counter. On the other hand, I'm not sure about the `incremented` and `decremented`. I think these two might be confusing to some user.

What would you think of moving forward with the return values for increment and decrement first (this could go in the next release), and maybe open an issue to discuss the additional functions.

k06a:
@Amxx so you're ok with these methods returning previous state?

Amxx:
I understand that this would be the lost usefull value to return... but at the same time I'm afraid this might be confusing...

frangio:
Hey @k06a, we discussed this a little more and we think there is too much potential for confusion between `increment` and `incremented`, that we don't think this would be a good idea.

The pricing changes in Berlin also make this a less important optimization.

PR #2605, opened 2021-03-22
EnumerableSet: Skip SSTORE On Last Item

wjmelements:
#### Changes
Reverses a bad tradeoff in the design of EnumerableSet.
Now even more efficient to remove from the end.
The JUMPI is >23 (DUPx, DUPx, PUSHx, EQ, JUMPI, JUMPDEST).
The redundant SSTOREs after EIP-2929 is >3256 = 2x(SSTORE_SLOAD_WARM, SLOAD_WARM, 5xDUPx, JUMPI, LT).
So there would need to be at least 142 items for skipping this check to be optimal, on average.
Of course solidity is less optimal than that but we can estimate with future compiler improvements.
But there is the pattern of removing all items to consider: it is not necessarily true that the items will be removed randomly.
In fact, one [major use](https://github.com/airswap/airswap-protocols/blob/7a6ca7aa1d22e4c335b17e54a506d634fd177a6d/source/registry/contracts/Registry.sol#L70) of the EnumerableSet is to track and remove all items in a mapping.
I believe the prior cost analysis missed solidity's boundary checks and supposed random removal.
Reviewers @nventuro


wjmelements:
An alternative might be to supply another method that removes the last item.

frangio (approved):
Thanks a lot for this suggestion @wjmelements. Sorry we took a while to get to it, last few weeks have been busy for us.

I definitely agree that this is a much beter tradeoff. We totally missed the mark on this.

frangio (approved):

PR #2606, opened 2021-03-22
EnumerableSet: Remove Boundary Check in _at

wjmelements:
#### Changes
This check is redundant because solidity does it for you without your consent, at great gas cost.
If you knew all of the things solidity does to arrays you would never use them.
But I'm not asking you to replace all of your solidity arrays.
Instead, complain to the compiler team, who doesn't trust you.
Reviewer @amxx

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
  * Enumerables: Improve gas cost of removal in `EnumerableSet` and `EnumerableMap`.
- * Enumerables: Improve gas cost of lookup `EnumerableSet` and `EnumerableMap`.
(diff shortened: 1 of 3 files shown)


wjmelements:
You're basically paying 1000 gas for a cleaner error message you are never expecting to see.

wjmelements:
Here is the difference, from the tests (which I will take the time to repair if you agree that this is a very expensive error message):
```
       Wrong kind of exception received
      + expected - actual

      -VM Exception while processing transaction: revert 
      +EnumerableSet: index out of bounds
```

As you can see, solidity asserts in this case, and that is fine.

Amxx (approved):

PR #2607, opened 2021-03-22
EnumerableSet: Clear

wjmelements:
#### Changes
It should be cheaper to remove all items.
It won't always be possible to do this because of the block gas limit, but when it is this should be much cheaper than sequential removal.
Reviewers @nventuro

### DIFF
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -206,6 +206,12 @@ library EnumerableSet {
         return _at(set._inner, index);
     }
 
+    /**
+     * @dev Removes all elements from the set. O(n).
+     *
+     * Note that this can run out of gas for large sets.
+     * First use `remove` to reduce the size of a large set in this case.
+     */
     function clear(Bytes32Set storage set) internal {
         return _clear(set._inner);
     }
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -66,7 +66,7 @@ library EnumerableSet {
     function _clear(Set storage set) private {
         uint256 last = set._values.length;
         while (last --> 0) {
-            delete set.indexes[set._values[last]];
+            delete set._indexes[set._values[last]];
         }
         set._values.length = 0;
     }
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -63,6 +63,12 @@ library EnumerableSet {
         }
     }
 
+    /**
+     * @dev Removes all elements from the set. O(n).
+     *
+     * Note that this can run out of gas for large sets.
+     * First use `remove` to reduce the size of a large set in this case.
+     */
     function _clear(Set storage set) private {
         uint256 last = set._values.length;
         unchecked {
@@ -70,7 +76,11 @@ library EnumerableSet {
                 delete set._indexes[set._values[last]];
             }
         }
-        delete set._values;
+        // Only set the _values.length to zero
+        // No need to clear _values from state: they will be overridden on push
+        assembly {
+            sstore(set.slot, 0)
+        }
     }
 
     /**
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -254,6 +254,12 @@ library EnumerableSet {
         return address(uint160(uint256(_at(set._inner, index))));
     }
 
+    /**
+     * @dev Removes all elements from the set. O(n).
+     *
+     * Note that this can run out of gas for large sets.
+     * First use `remove` to reduce the size of a large set in this case.
+     */
     function clear(AddressSet storage set) internal {
         return _clear(set._inner);
     }
@@ -312,6 +318,12 @@ library EnumerableSet {
         return uint256(_at(set._inner, index));
     }
 
+    /**
+     * @dev Removes all elements from the set. O(n).
+     *
+     * Note that this can run out of gas for large sets.
+     * First use `remove` to reduce the size of a large set in this case.
+     */
     function clear(UintSet storage set) internal {
         return _clear(set._inner);
     }
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -63,6 +63,14 @@ library EnumerableSet {
         }
     }
 
+    function _clear(Set storage set) private {
+        uint256 last = set._values.length;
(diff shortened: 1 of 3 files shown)


wjmelements:
```
TypeError: Member "length" is read-only and cannot be used to resize arrays.
  --> contracts/utils/structs/EnumerableSet.sol:71:9:
   |
71 |         set._values.length = 0;
   |         ^^^^^^^^^^^^^^^^^^
```
I hate that they removed this functionality. It was so good; it zeroed all of the elements too.

wjmelements:
If this feature will be accepted, I will write tests for it.

Amxx:
Hello @wjmelements, and that you for your many PR addressing gas usage! While we didn't found ourselves in a situation where this feature was required, it would be a nice to have that user might find handy.

I'm worried about the "loop" part, and the fact that a smart contract relying on this feature might be in trouble if the set includes too many elements for the clear to git in one transaction.

> ```
> TypeError: Member "length" is read-only and cannot be used to resize arrays.
>   --> contracts/utils/structs/EnumerableSet.sol:71:9:
>    |
> 71 |         set._values.length = 0;
>    |         ^^^^^^^^^^^^^^^^^^
> ```
> 
> I hate that they removed this functionality. It was so good; it zeroed all of the elements too.

While zeroing the elements is a nice addition, I don't think it is strictly necessary. Setting the length to 0, even without the elements being erased, should maintain the safe behavior of the set. You would not get the refund associated with erasing the elements, but then putting new data in will also be cheaper.
While solidity prevents it, I do believe it can still be done in assembly. What do you think?

wjmelements:
> While we didn't found ourselves in a situation where this feature was required, it would be a nice to have that user might find handy.

https://github.com/airswap/airswap-protocols/blob/7a6ca7aa1d22e4c335b17e54a506d634fd177a6d/source/registry/contracts/Registry.sol#L64

> relying on this feature might be in trouble if the set includes too many elements for the clear to git in one transaction.

In my case there is [another way to remove individual elements](https://github.com/airswap/airswap-protocols/blob/7a6ca7aa1d22e4c335b17e54a506d634fd177a6d/source/registry/contracts/Registry.sol#L48) that we can fall back to. Can recommend that via docs.

> You would not get the refund associated with erasing the elements, but then putting new data in will also be cheaper.
> While solidity prevents it, I do believe it can still be done in assembly. What do you think?

Fine by me, especially if they go forward with [removing the refund incentive in London](https://eips.ethereum.org/EIPS/eip-3403).

wjmelements:
> While solidity prevents it, I do believe it can still be done in assembly. What do you think?

I started this but I am getting an error I don't see a way to disable. I've looked through the solidity source code and found no such flag. The sage advice on the stack exchange is to rename the functions that collide. I suggest renaming `add` to `insert`. @Amxx does that sound good to you?

```
Warning: Variable is shadowed in inline assembly by an instruction of the same name
   --> contracts/utils/structs/EnumerableSet.sol:159:5:
    |
159 |     function add(Bytes32Set storage set, bytes32 value) internal returns (bool) {
    |     ^ (Relevant source part starts here and spans across multiple lines).


Warning: Variable is shadowed in inline assembly by an instruction of the same name
   --> contracts/utils/structs/EnumerableSet.sol:217:5:
    |
217 |     function add(AddressSet storage set, address value) internal returns (bool) {
    |     ^ (Relevant source part starts here and spans across multiple lines).


Warning: Variable is shadowed in inline assembly by an instruction of the same name
   --> contracts/utils/structs/EnumerableSet.sol:275:5:
    |
275 |     function add(UintSet storage set, uint256 value) internal returns (bool) {
    |     ^ (Relevant source part starts here and spans across multiple lines).
```

Amxx:
We should definitely NOT rename any functions. This would be a braking change in the library and we want to avoid that at all cost.

Also, this warning was removed since solc 0.8.2. Can you confirm by running the code against 0.8.2 or 0.8.3 ?

wjmelements:
> We should definitely NOT rename any functions. This would be a breaking change in the library and we want to avoid that at all cost.
> 
> Also, this warning was removed since solc 0.8.2. Can you confirm by running the code against 0.8.2 or 0.8.3 ?

Yes, that does address the warning. Should I commit the change in `hardhat.config.js` or leave it as-is?

wjmelements:
> Yes, that does address the warning. Should I commit the change in `hardhat.config.js` or leave it as-is?

@Amxx My intuition is that I should leave it as-is because it doesn't break any of the checks and other patch-related issues may go unnoticed despite the intended compiler versions to support, but the counter-points are that there are no other build warnings, and end-users may see the warning and file issues.

wjmelements:
I see the solc version is bumped in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2612 so I will merge master after that merges.

contracts/utils/structs/EnumerableSet.sol
```diff
@@ -63,6 +63,26 @@ library EnumerableSet {
...
+    /**
+     * @dev Removes all elements from the set. O(n).
+     *
+     * Note that this can run out of gas for large sets.
+     * First use `remove` to reduce the size of a large set in this case.
+     */
+    function _clear(Set storage set) private {
+        uint256 last = set._values.length;
+        unchecked {
+            while (last --> 0) {
```
eMarchenko, line 75: tbh, I've spent 15 minutes figuring out what is this `-->` syntax. Eventually my colleague saved me as he guessed it is `last-- > 0`. I wasn't even aware that it is possible. While this language feature is cool and I really like the way it looks, it seems to be too advanced for the most popular contract library.

wjmelements: This is the best way to loop to zero inclusive on an unsigned variable. I've always thought it more clear when written `-->`, but I would rearrange the whitespace upon reviewer request.

wjmelements:
ptal @Amxx

frangio:
We've been discussing this and we don't feel comfortable including function that loops over an array of unbounded length, so we're not going to merge it.

We evaluated using assembly to set the length to zero while leaving the old set values in storage, but this seems to still require a loop over the indices array. It may be possible to get it working with some more tricks, we encourage people to try this out and let us know if they find a clean way to implement set clear without any loops.

A potential way to implement "clearing" is to throw away the set and use a new one, by keeping a mapping (or an array?) of sets, where bumping the index to the currently used set is equivalent to clearing. This will incur an additional storage load.

contracts/utils/structs/EnumerableSet.sol
wjmelements: Additional matter of interest: decrement inside solidity loop conditions doesn't seem to work at all in 0.8.0. I am concerned at how the tests still passed.

PR #2609, opened 2021-03-23
Add modifier & internal function with standard revert message in AccessControl

### LINKED ISSUE #2601: Hint at missing role in revert reason for AccessControl rejection
It would be useful to notify clients of what role is necessary to call a function protected by `AccessControl`. We can include this information in the revert reason in a standard and machine-parseable format.

Note that we don't have a built in function or modifier to check if an account has a role. Users are currently encouraged to use `require(hasRole(MINTER_ROLE, msg.sender))`, so we would need to add this new modifier or function in order to standardize the revert reason hint.

Amxx:
Fixes #2601

#### PR Checklist
- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio (dismissed):

frangio (approved):
Good to go!

PR #2612, opened 2021-03-25
improve hardhat config

Amxx:
Current version enables `hardhat-gas-reporter` even when not requested, and without properly enabling optimization.

This PR uses yargs to clarify options in the hardhat config. This will be useful for future expansions of the config (add flags for optimism support ?)


frangio:
>without properly enabling optimization

What do you mean here?

Can you add the `yargs` dev dependency?

Amxx:
Currently doing `yarn test` enables gas reporter (it shouldn't) without enabling optimisation. That is why I think it should be enabled/disabled at the package require level. That will prevent people looking at unoptimized gas report.

I'm adding yargs

frangio (approved):
Thanks!

PR #2618, opened 2021-03-28
add payable to ERC721 interface functions

### LINKED ISSUE #2610: ERC721 payable transferFrom
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/378531b70906c906461fa14e5631f0569b938257/contracts/token/ERC721/IERC721.sol#L70

Hello,

I'm trying to implement royalty in ERC721 token. The idea is to have a `wantToSell` method to set a price for a token, if you want to allow someone to buy it, and a `transferFrom payable` method to allow buyer to call it providing enough value ETH.
The royalty is distributed to the token contract owner and token ID author in all subsequent re-sells.

The problem is that the `transferFrom` function declared without the `payable` modifier and the solidity compiler complains if I try to override the `transferFrom` function adding the `payable` modifier myself.

The standard declares the `transferFrom` method as a `payable` https://eips.ethereum.org/EIPS/eip-721

```solidity
function transferFrom(address _from, address _to, uint256 _tokenId) external payable;
``` 

1. Is there some reason why openzeppelin implementation deviate from the standard in this regard?
2. Is there some workaround I can use to implement the royalty functionality in the current openzeppelin implementation?
3. Is there a better way to implement the royalty functionality that would not suffer from this issue?

sunnya97:
Fixes #2610 

The official EIP 721 states that `transferFrom`, `safeTransferFrom`, and `approve` are payable.  However, this is not implemented in the open-zeppelin interfaces.

>  A reason why we did not make them payable in our implementation is to avoid funds getting stuck since there is no mechanism to withdraw any fund that might be sent by mistake

In order to maintain this, the sample ERC721.sol implementation is also switched to payable (so it can be easily imported), but requires `msg.value ==  0` in the function to prevent stuck funds unless explicitly overridden.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -108,7 +108,10 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
     /**
      * @dev See {IERC721-approve}.
      */
-    function approve(address to, uint256 tokenId) public virtual override {
+    function approve(address to, uint256 tokenId) public payable virtual override {
+        
+        
+
         address owner = ERC721.ownerOf(tokenId);
         require(to != owner, "ERC721: approval to current owner");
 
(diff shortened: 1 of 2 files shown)


frangio:
See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2610#issuecomment-819006037.

PR #2619, opened 2021-03-29
Improve ECDSA tests and docs

### LINKED ISSUE #1667: Improve ECDSA tests
https://github.com/OpenZeppelin/openzeppelin-solidity/pull/1622 introduced some changes to the `ECDSA` contract, and made it clear that our test suite for that particular contract is somewhat lacking. We should refactor it so that:

- it's easy for users to understand how to use the contract (i.e. how to generate signatures e.g. using `web3.eth.sign`)
- all cases are properly covered (all combinations of valid and invalid `s` and `v` values)

### LINKED ISSUE #2190: Document workarounds for ECDSA.recover with Ganache
In 2.2.0 we introduced protections against malleability of ECDSA signatures. More details can be found in https://github.com/OpenZeppelin/openzeppelin-solidity/pull/1622.

One unfortunate side effect of this has been that `ECDSA.recover` stopped working with Ganache-generated signatures, because they use the signatures parameters that the library rejects. This seems to be a problem with Ganache only. There seems to be plans to fix it in the next major version of Ganache, with an unknown ETA. The relevant issues are https://github.com/trufflesuite/ganache-core/issues/451, https://github.com/trufflesuite/ganache-core/issues/556.

We evaluated rolling back the changes to ECDSA in Contracts 3.0. Given that we think malleability protection makes sense in general, that the issue only affects the current version of Ganache, and that there are workarounds, we decided to keep the current behavior and instead improve the documentation by including the workarounds that are required to make it work in Ganache, and more details about the issue to understand it.

The workaround can be found in [`test/helpers/sign.js: fixSignature`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b6513f6ad7f8a31c2bbfd9a8e0a85e78ea5b0df5/test/helpers/sign.js#L9-L19). We should move it to test-helpers to make it available as a complement to the documentation.

### LINKED ISSUE #2600: Improve Cryptography test clarity to showcase how signatures are generated
**🧐 Motivation**
Many people learn how some of the more exotic features of OpenZeppelin contracts work by reading the tests. Having clear tests has always been a hallmark of the OpenZeppelin library. 

In the Cryptography tests (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/cryptography/ECDSA.test.js) the signature used is pre-generated. This opens an uncertainty for the reader who perhaps doesn't understand how to properly generate this signature. 

**📝 Details**

I would like to ask that in the Cryptography tests (and other tests if this idea applies) that we don't use precomputed values, and instead, show the computation of these values as part of the tests so that users understand how to replicate the results on their own, and how to design Javascript code that will properly pass the tests. 

See for [example](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/09e302384a3c77be1b482b18a4e30f88630cfd66/test/utils/cryptography/ECDSA.test.js#L37): 

The pre-computation of this signature leaves the reader unsure how to properly create a signature on their own. 

[Here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/09e302384a3c77be1b482b18a4e30f88630cfd66/test/utils/cryptography/ECDSA.test.js#L102) in the same test file, we have a different form of signature which is unexplained a "high-s" value. 

Precomputing "magic numbers" without an explanation as to how or why this value was obtained makes the tests opaque to those unfamiliar with the subject. 

IMHO the OpenZeppelin tests should not only test the smart contract code, but also show the user HOW to test the smart contract code. 

The accompanying OpenZeppelin [Documentation](https://docs.openzeppelin.com/contracts/3.x/api/cryptography) similarly does not describe how to create signatures so the user is left unsure how the tests created these values.

Amxx:
Fixes #2600
Fixes #2190
Fixes #1667

#### PR Checklist
- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -22,6 +22,10 @@ library ECDSA {
      * recover to arbitrary addresses for non-hashed data. A safe way to ensure
      * this is by receiving a hash of the original message (which may otherwise
      * be too long), and then calling {toEthSignedMessageHash} on it.
+     *
+     * Documentation for signature generation:
+     * - with https://web3js.readthedocs.io/en/v1.3.4/web3-eth-accounts.html#sign[Web3.js]
+     * - with https://docs.ethers.io/v5/api/signer/#Signer-signMessage[ethers]
      */
     function recover(bytes32 hash, bytes memory signature) internal pure returns (address) {
         // Check the signature length
--- a/test/helpers/sign.js
+++ b/test/helpers/sign.js
@@ -4,23 +4,6 @@ function toEthSignedMessageHash (messageHex) {
   return web3.utils.sha3(Buffer.concat([prefix, messageBuffer]));
 }
 
-function fixSignature (signature) {
-  // in geth its always 27/28, in ganache its 0/1. Change to 27/28 to prevent
-  // signature malleability if version is 0/1
-  // see https://github.com/ethereum/go-ethereum/blob/v1.8.23/internal/ethapi/api.go#L465
-  let v = parseInt(signature.slice(130, 132), 16);
-  if (v < 27) {
-    v += 27;
-  }
-  const vHex = v.toString(16);
-  return signature.slice(0, 130) + vHex;
-}
-
-// signs message in node (ganache auto-applies "Ethereum Signed Message" prefix)
-async function signMessage (signer, messageHex = '0x') {
-  return fixSignature(await web3.eth.sign(messageHex, signer));
-};
-
 /**
  * Create a signer between a contract and a signer for a voucher of method, args, and redeemer
  * Note that `method` is the web3 method, not the truffle-contract method
@@ -55,12 +38,10 @@ const getSignFor = (contract, signer) => (redeemer, methodName, methodArgs = [])
 
   // return the signature of the "Ethereum Signed Message" hash of the hash of `parts`
   const messageHex = web3.utils.soliditySha3(...parts);
-  return signMessage(signer, messageHex);
+  return web3.eth.sign(messageHex, signer);
 };
 
 module.exports = {
-  signMessage,
   toEthSignedMessageHash,
-  fixSignature,
   getSignFor,
 };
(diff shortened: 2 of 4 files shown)


l2ichshop2 (approved):

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -22,6 +22,10 @@ library ECDSA {
      * recover to arbitrary addresses for non-hashed data. A safe way to ensure
      * this is by receiving a hash of the original message (which may otherwise
      * be too long), and then calling {toEthSignedMessageHash} on it.
+     *
+     * Documentation for signature generation:
```
frangio, line 26: It would be nicer to show the code inline here. "A client would generate signatures by using the signing methods in Web3.js or Ethers.js, as shown below:" ...

frangio, line 26: Can you take a look at https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2190 as well? I think this is a good opportunity to review that issue, see if it's still valid with the latest version of Ganache, and document the workarounds. We're using the `fixSignature` helper in our tests (we actually may no longer need this under Hardhat...), but we should also think about what our users will need to do in their own tests and how we could make it easier for them.

frangio:
Note that this helps but does not really fix #2600. That issue is about removing the magic signature constants inlined in the test and replace them with code that generates the signatures, like the one in #2535.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5ecd2b850c2acb52b94d8374ac8d22cb543e71c8/test/utils/cryptography/ECDSA.test.js#L26

frangio:
See also https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1667. This PR should close that issue.

contracts/utils/cryptography/ECDSA.sol
Amxx: You'll like 04d4b09 :)

frangio (approved):

PR #2623, opened 2021-04-02
Refactor ERC1155 transfers to use internal functions #2622

### LINKED ISSUE #2622: ERC1155 transfers should have internal function that doesn't check approvals
**🧐 Motivation**
I'm trying to implement a "signed transfer" function for an ERC1155 token (see https://etherscan.io/address/0x478059f577Cb739f0AF0f37A365DD069Ba5BD8Be#code line 2018 ff. for an ERC721 variant) - any user can call this function but needs a fitting signature to actually perform that transfer. In ERC721, this can easily be done by calling the internal `_safeTransfer()` function after checking that all the conditions for the transfer (signature, etc.) are met - but in ERC1155, the same functions that perform the actual transfer also check approval and so with OpenZeppelin 4.0.0, it's hard to implement this (there is a hack I could use but it's ugly). There may be other cases as well where you may want to transfer a token gated by other means than the traditional/standard approval process.
As an overriding contract is unable to access or manipulate the balances directly due to those being private to the OpenZeppelin implementation, it's also not an option to completely copy what the existing functions do into the overriding contract (and I would consider that a hack as well if it were possible).

**📝 Details**
To solve this problem, the ERC1155 transfer functions (`safeTransferFrom`, `safeBatchTransferFrom`) should call additional internal functions that does the actual transfer operation and doesn't check approvals, following the same pattern as the OpenZeppelin ERC721 contract does. Note that this is just a code structuring question, as this would not be available to external callers directly, only to overriding contracts based on the OpenZeppelin implementation.

I actually have a PR for this upcoming.

KaiRo-at:
Fixes #2622 (which I just created)

This moves the actual functionality of ERC1155 transfers into internal functions. I copied and adapted the function documentation from IERC1155 as it slightly differs by not checking for approvals, I'm not 100% sure if that's the right way to do that.

Test changes should not be needed, I think, because ERC721 also doesn't explicitly test those internal functions, only the actually exposed ones.
I see a "changelog" entry in the checklist for the PR, should this change go into the changelog at all, as small and invisible as it is for the majority of people?

### DIFF
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -130,24 +130,11 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
         virtual
         override
     {
-        require(to != address(0), "ERC1155: transfer to the zero address");
         require(
             from == _msgSender() || isApprovedForAll(from, _msgSender()),
             "ERC1155: caller is not owner nor approved"
         );
-
-        address operator = _msgSender();
-
-        _beforeTokenTransfer(operator, from, to, _asSingletonArray(id), _asSingletonArray(amount), data);
-
-        uint256 fromBalance = _balances[id][from];
-        require(fromBalance >= amount, "ERC1155: insufficient balance for transfer");
-        _balances[id][from] = fromBalance - amount;
-        _balances[id][to] += amount;
-
-        emit TransferSingle(operator, from, to, id, amount);
-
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @KaiRo-at, and thank you for your Issue and your PR.

I completely understand you use case. In many cases, overloading isApprovedForAll could work, but using that trick here would involve unnecessary stores, which is bad.

We'll discuss and review that next week, but AFAIK, I think your proposal is solid.

Amxx:
sligtly modified and merge as #2636

PR #2625, opened 2021-04-07
Improve transferFrom security

aybehrouz:
`transferFrom` function updates `_allowances` data structure after calling `_transfer` function:

```solidity
function transferFrom(address sender, address recipient, uint256 amount) public virtual override returns (bool) {
    _transfer(sender, recipient, amount);

    uint256 currentAllowance = _allowances[sender][_msgSender()];
    require(currentAllowance >= amount, "ERC20: transfer amount exceeds allowance");
    _approve(sender, _msgSender(), currentAllowance - amount);

    return true;
}
```

This can easily lead to *re-entrance* vulnerabilities, when the user overrides `_transfer` method. As recommended, we should finish all internal work (i.e. state changes) first, and only then we should call the (possibly) external function.

Consider the following scenario: 

A user wants to implement an exchange like logic via inheritance of OpenZeppelin's ERC20 contract. He wants to implement a logic that enables conversion of his token to Ether via sending tokens to his ERC20 contract address. To do so, he overrides the `_transfer` method:

```solidity
function _transfer(address sender, address recipient, uint256 amount) internal override {
    super._transfer(sender, recipient, amount);
    if (recipient == address(this)) {
        uint amountToSend = convert(amount);
        (bool success, ) = msg.sender.call{value: amount}(""); 
        require(success);
    }
}
```
In case of calls to normal `transfer`, the re-entrance to this function will only result the sender to convert more tokens to ETH. This behavior can not be considered a security flaw. But in case of `transferFrom` situation is different. Because `_allowances` is updated after the call to `_transfer`, an attacker who has a limited allowance amount `x` by calling `transferFrom`, can use that allowance to send an arbitrary amount of user's tokens to the ERC20 contract address and receive the corresponding amount of ETH to his account. 

By changing the implementation of `transferFrom` as I've proposed, this vulnerability will not be created in the user code and the attacker will not be able to convert more than `x` amount of user's tokens.  

Fixes #???? 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -146,11 +146,11 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
      * `amount`.
      */
     function transferFrom(address sender, address recipient, uint256 amount) public virtual override returns (bool) {
-        _transfer(sender, recipient, amount);
-
         uint256 currentAllowance = _allowances[sender][_msgSender()];
         require(currentAllowance >= amount, "ERC20: transfer amount exceeds allowance");
         _approve(sender, _msgSender(), currentAllowance - amount);
+        
+        _transfer(sender, recipient, amount);
 
         return true;
     }


Amxx:
Hello @aybehrouz and thank you for your concern about security.
 
While it is true that overloading `_transfer` could allow  reeintrancy with an allowance that is not updated, you are missing the part of the call where the allowance is updated. Lets unwrap such an attack with one level of reintrance:

```
→ transferFrom
  → _transfer
    → super._transfer
    → receive + reintrancy
      → tranferFrom
        → _transfer
          → super._transfer
          → receive (no reintrancy)
        → check allowance
        → _approve (decrement)
  → check
  → _approve (decrement)
```

If the 2 transferFrom add up to an amount that is bigger then the allowance, then the check allowance will cause a revert. In the end, the allowance will be respected !

Amxx:
Also, this as already been discussed (I don't really remember where, but it was discussed that the most common cause of revert was lack of funds (and not lack of allowance, which is often set to `uint1)` by dapps.

Thus doing `_transfer` before `_approve` help save gas (by reverting sooner) for the average call

aybehrouz:
Thank you for your explanation. I just pushed the functions into the call stack and forgot to pop them! lol

PR #2626, opened 2021-04-09
Fix merge issue in CHANGELOG.md

svpplynd3mand:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Openzeppelin contracts is attached to being a light project and It doesn't have any dependencies.
Not only is truffle one of the biggest package in the ecosystem, but it is not used here at all. Compilation an test is done with hardhat.

Not sure if that is a tool PR, but it looks like it.

PR #2632, opened 2021-04-19
ERC20 extension for governance tokens (vote delegation and snapshots)

Amxx:
The goal is to offer an ERC20 extension that can be included in a token to make it suitable for on-chain governance with vote delegation. This extension is meant to be COMP-compatible.

- [x] check compatibility against current comp tokens

#### PR Checklist
- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,169 @@
...
+
+    mapping (address => address) private _delegates;
+    mapping (address => Checkpoint[]) private _checkpoints;
+
+    function checkpoints(address account, uint32 pos) external view virtual override returns (Checkpoint memory) {
+        return _checkpoints[account][pos];
+    }
+
+    function numCheckpoints(address account) external view virtual override returns (uint32) {
+        return uint32(_checkpoints[account].length);
```
frangio, line 35: We should use `SafeCast` when downcasting.

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,169 @@
...
+    function delegates(address account) public view virtual override returns (address) {
+        return _delegates[account];
+    }
+
+    /**
+     * @dev Example: This enables autodelegation, makes each transfer more expensive but doesn't require user to
+     * delegate to themselves. Can be usefull for tokens useds exclusivelly for governance, such as voting wrappers of
+     * pre-existing ERC20.
+     */
+    // function delegates(address account) public view override returns (address) {
```
frangio, line 50: We should remove this comment.

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,169 @@
...
+        while (low < high) {
+            uint256 mid = Math.average(low, high);
+            if (ckpts[mid].fromBlock > blockNumber) {
+                high = mid;
+            } else {
+                low = mid + 1;
+            }
+        }
+
+        return low == 0 ? 0 : ckpts[low - 1].votes;
```
frangio, line 88: This doesn't look right to me. What happens when you query for the first checkpointed block? I think this returns zero.

Line 84 should be `low = mid` and here we should return `ckpts[low]` (unless...).

`high` is the exclusive upper bound of the range that contains the right checkpoint, and `low` is an inclusive lower bound. 

Please confirm. Tests for this scenario would be good.

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,169 @@
...
+            if (dstRep != address(0)) {
+                uint256 dstRepNum = _checkpoints[dstRep].length;
+                uint256 dstRepOld = dstRepNum == 0 ? 0 : _checkpoints[dstRep][dstRepNum - 1].votes;
+                uint256 dstRepNew = dstRepOld + amount;
+                _writeCheckpoint(dstRep, dstRepNum, dstRepOld, dstRepNew);
+            }
+        }
+    }
+
+    function _writeCheckpoint(address delegatee, uint256 pos, uint256 oldWeight, uint256 newWeight) private {
```
frangio, line 153: Should use `SafeCast` for downcasts.

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,169 @@
...
+        _moveDelegates(currentDelegate, delegatee, delegatorBalance);
+    }
+
+    function _moveDelegates(address srcRep, address dstRep, uint256 amount) private {
+        if (srcRep != dstRep && amount > 0) {
+            if (srcRep != address(0)) {
+                uint256 srcRepNum = _checkpoints[srcRep].length;
+                uint256 srcRepOld = srcRepNum == 0 ? 0 : _checkpoints[srcRep][srcRepNum - 1].votes;
+                uint256 srcRepNew = srcRepOld - amount;
+                _writeCheckpoint(srcRep, srcRepNum, srcRepOld, srcRepNew);
```
frangio, line 141: How could we make `_writeCheckpoint()` more abstract and avoid duplicating the other two lines?

contracts/token/ERC20/extensions/ERC20Snapshot.sol
```diff
@@ -79,11 +83,18 @@ abstract contract ERC20Snapshot is ERC20 {
     function _snapshot() internal virtual returns (uint256) {
         _currentSnapshotId.increment();
```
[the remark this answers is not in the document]
Amxx: Its an internal, state changing, function ... so for consistency sake it should be virtual

contracts/token/ERC20/extensions/draft-IERC20Votes.sol
```diff
@@ -0,0 +1,24 @@
...
+
+pragma solidity ^0.8.0;
+
+import "../IERC20.sol";
+
+interface IERC20Votes is IERC20 {
+    struct Checkpoint {
+        uint32  fromBlock;
+        uint224 votes;
+    }
```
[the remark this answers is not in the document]
Amxx: Its used as the return type of one of the functions

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
Amxx: I honestly like it that way. I'm not a fan of adding a boolean to distinguish add/sub

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
Amxx: We have this test, which I believe covers it:

      it('returns the number of checkpoints for a delegate', async function () {
        await this.token.transfer(recipient, '100', { from: holder }); //give an account a few tokens for readability
        expect(await this.token.numCheckpoints(other1)).to.be.bignumber.equal('0');

        const t1 = await this.token.delegate(other1, { from: recipient });
        expect(await this.token.numCheckpoints(other1)).to.be.bignumber.equal('1');

        const t2 = await this.token.transfer(other2, 10, { from: recipient });
        expect(await this.token.numCheckpoints(other1)).to.be.bignumber.equal('2');

        const t3 = await this.token.transfer(other2, 10, { from: recipient });
        expect(await this.token.numCheckpoints(other1)).to.be.bignumber.equal('3');

        const t4 = await this.token.transfer(recipient, 20, { from: holder });
        expect(await this.token.numCheckpoints(other1)).to.be.bignumber.equal('4');

        expect(await this.token.checkpoints(other1, 0)).to.be.deep.equal([ t1.receipt.blockNumber.toString(), '100' ]);
        expect(await this.token.checkpoints(other1, 1)).to.be.deep.equal([ t2.receipt.blockNumber.toString(), '90' ]);
        expect(await this.token.checkpoints(other1, 2)).to.be.deep.equal([ t3.receipt.blockNumber.toString(), '80' ]);
        expect(await this.token.checkpoints(other1, 3)).to.be.deep.equal([ t4.receipt.blockNumber.toString(), '100' ]);
      });```

Amxx: walkthrough, assuming `length = 4` and `ckpts[0] = blocknumber` 

- low = 0, high = 4, mid = 2 → update high
- low = 0, high = 2, mid = 1 → update high
- low = 0, high = 1, mid = 0 → update low
- low = 1, high = 1, loop stops
- we read ckpts[low-1] → good!

Amxx: For low to be 0 after the loop end, you'd need low = high = 0, meaning that there are either no checkpoints at all, or for all checkpoints, including ckpts[0], `ckpts[mid].fromBlock > blockNumber` ... this means that all checkpoints are older then the request, and thus returning 0 is the right thing to do

Amxx: Also, if we do low = mid, we can get stuck in infinite loops:
low = 0, high = 1, mid = 0 ... doing low = mid will loop forever

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,160 @@
...
+        while (low < high) {
+            uint256 mid = Math.average(low, high);
+            if (ckpts[mid].fromBlock > blockNumber) {
+                high = mid;
+            } else {
+                low = mid + 1;
+            }
+        }
+
+        return low == 0 ? 0 : ckpts[low - 1].votes;
```
frangio, line 79: This makes more sense to me.

```suggestion
        // In the loop below, `high` is always greater than the index of the block number we're
        // looking for. At the end of the loop, it is the least of the such indices, thus we return the
        // checkpoint at index high - 1.
        uint256 high = ckpts.length;
        uint256 low = 0;
        while (low < high) {
            uint256 mid = Math.average(low, high);
            if (ckpts[mid].fromBlock > blockNumber) {
                high = mid;
            } else {
                low = mid;
            }
        }

        return high == 0 ? 0 : ckpts[high - 1].votes;
```

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,166 @@
...
+          _checkpoints[delegatee].push(Checkpoint({
+              fromBlock: SafeCast.toUint32(block.number),
+              votes: SafeCast.toUint224(newWeight)
+          }));
+      }
+
+      emit DelegateVotesChanged(delegatee, oldWeight, newWeight);
+    }
+
+    function _beforeTokenTransfer(address from, address to, uint256 amount) internal virtual override {
```
frangio, line 164: We should check that the total supply doesn't exceed uint224. Either here or in `_mint`.

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
```diff
@@ -0,0 +1,166 @@
...
+        uint256 delegatorBalance = balanceOf(delegator);
+        _delegates[delegator] = delegatee;
+
+        emit DelegateChanged(delegator, currentDelegate, delegatee);
+
+        _moveDelegates(currentDelegate, delegatee, delegatorBalance);
+    }
+
+    function _moveDelegates(address srcRep, address dstRep, uint256 amount) private {
+        if (srcRep != dstRep && amount > 0) {
```
frangio, line 134: I think the `Rep` variable names are confusing. The "rep" concept isn't used anywhere else in the code. 

Also `_moveDelegates` may be better named `_moveVotingPower`?

Amxx: Names come from Compound. I agree they are not great, if you have better ideas I'm all hears

contracts/token/ERC20/extensions/draft-ERC20Votes.sol
Amxx: I'd definitelly do it in _mint

Amxx: Implemented, I'm worried however that it might interfere with ERC20Capped

frangio (approved):
LGTM

PR #2633, opened 2021-04-19
Wrapper extension for ERC20 token 

Amxx:
#### PR Checklist
- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/token/ERC20/extensions/draft-ERC20Wrapper.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "../ERC20.sol";
+import "../utils/SafeERC20.sol";
+
+abstract contract ERC20Wrapper is ERC20 {
```
frangio, line 8: Does this have to be abstract? It sounds useful on its own.

contracts/token/ERC20/extensions/draft-ERC20Wrapper.sol
```diff
@@ -0,0 +1,26 @@
...
+    IERC20 immutable public underlying;
+
+    constructor(IERC20 underlyingToken) {
+        underlying = underlyingToken;
+    }
+
+    function depositFor(address account, uint256 amount) public virtual returns (bool) {
```
frangio, line 15: What do you think of adding a `depositByPermit` function for when the underling token supports ERC20Permit? Although this would add to bytecode size unnecessarily if the underlying token doesn't support it.

Amxx: That is a possibility. I was also considering being a 1363 receiver, so the deposit can be triggered by erc1363 transfer methods

RonSherfey (approved):
Change Request 1.0

contracts/token/ERC20/extensions/draft-ERC20Wrapper.sol
frangio: How do we deal with all of these variants? It wouldn't be right to include them if they aren't going to be used, but it doesn't feel worth it to make several variants using inheritance...

Amxx: We shouldn't put everything in. I'm not sure what to chose. We should discuss that.

contracts/token/ERC20/extensions/draft-ERC20Wrapper.sol
Amxx: For it to be non-abstract, the constructor needs to take a name and symbol and pass them through. I'm not a big fan of the idea. I'd prefer a preset, or support in the wizard.

RonSherfey (changes requested):
Reviewed changes 1.0

contracts/token/ERC20/extensions/draft-ERC20Wrapper.sol
Amxx: Would you want something like this
```
    function depositByPermit(address owner, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s) public virtual returns (bool) {
        IERC20Permit(underlying).permit(owner, address(this), value, deadline, v, r, s);
        SafeERC20.safeTransferFrom(underlying, owner, address(this), value);
        _mint(owner, value);
        return true;
    }
```
Or would you assume the permit was done in a separate tx, and have a signature that protects that would allow the recipient of the wrapped token to be different then the owner of the underlying asset ?

test/token/ERC20/extensions/ERC20Wrapper.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+  it('has underlying', async function () {
+    expect(await this.token.underlying()).to.be.bignumber.equal(this.underlying.address);
+  });
+
+  describe('deposit', function () {
+    it('valid', async function () {
+      await this.underlying.approve(this.token.address, initialSupply, { from: initialHolder });
+      const { tx } = await this.token.depositFor(initialHolder, initialSupply, { from: initialHolder });
+      expectEvent.inTransaction(tx, this.underlying, 'Transfer', {
```
frangio, line 43: It shouldn't be necessary to get the inner `tx` here, you can just write:

```
const receipt = await this.token.depositFor(...);
expectEvent(receipt, this.underlying,  'Transfer', {
```

test/token/ERC20/extensions/ERC20Wrapper.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+    it('missing balance', async function () {
+      await expectRevert(
+        this.token.withdrawTo(initialHolder, MAX_UINT256, { from: initialHolder }),
+        'ERC20: burn amount exceeds balance',
+      );
+    });
+
+    it('valid', async function () {
+      const { tx } = await this.token.withdrawTo(initialHolder, initialSupply, { from: initialHolder });
```
frangio, line 100: Can we test this with a value smaller than `initialSupply`?

test/token/ERC20/extensions/ERC20Wrapper.test.js
Amxx: There are two transfer event emitted by this tx, one in the underlying token and one in the wrapper. I found it way cleaner to check both events using the `expectEvent.inTransaction` mechanism, which lets me specify the sender. That way I can isolate both event and check them properly

test/token/ERC20/extensions/ERC20Wrapper.test.js
Amxx: sure

frangio (approved):
LGTM!

Issue #2639, opened 2021-04-23
Ownable should support setting the initial owner through its constructor


wighawag:
**🧐 Motivation**
Currently Ownable force you to set the initial owner to be the msg sender. This brings a dependence on the account that deploy the contract, which is often not the same as the expected owner of the contract. For security it is often better to ensure the account making deployment do not have any responsibility.

it also brings issue when you need to deploy the contract through a factory where the factory become the defacto owner of the contract.

While contracts using `Ownable` could call `transferOwnership` this would trigger another event.

It would be better if Ownable constructor had a `owner` parameter so contract using Ownable can decide whose address is going to be the initial owner

Amxx:
Hello @wighawag 

This feature has already been requested multiple times. In a perfect world, solidity would support optional arguments, and we would be able to use the provided value or fallback to initializing using msg.sender. Unfortunately, overloading constructor is not yet a thing.

While the change you ask for is easy to do, and make a lot of sense, it would also be a breaking change. Any contract that uses Ownable right now (and there are many!) would not be compatible with the newer version, and this would require users to change their code. We want to avoid this, and we certainly cannot do such a change in a minor version.

We might possibly consider this for whenever we release contracts 5.0, but until then, calling `transferOwnership` in the constructor is the way to go. While it is true that there are 2 events, the 2 are consistent with the ownership logic, so they should not be to confusing for an outside observer.

wighawag:
Hi @Amxx 

That's a shame we cannot have it sooner :) I have actually complained secretly about the issue for years :D

I understand the backward compatibility requirement though, thanks for explaning.

My main use case for now is ProxyAdmin that I have to reimplement so I can specify the onwer at deployment time. it would great in version 5.0 that no assumption was made as to the role of msg.sender and that every contract that extends Ownable also allow users to pass the owner in their constructor transitively.

frangio:
This is a duplicate of https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2402 (closed). See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2402#issuecomment-780877304, although it has been summarized by @Amxx above.

Is it possible for you to inherit `ProxyAdmin` and use `transferOwnership` in the constructor, without reimplementing the entire contract?

wighawag:
@frangio yes this is possible but feel like a waste (even if small), it would also emit the OwnershipTransfer event twice.

[...]

Issue #2402, opened 2020-11-09
Assigning the ownership in the constructor creates issues for factory patterns


thegostep:
This creates additional 5k deployment cost to update the ownership as a part of the deployment. I recommend using an initializer function which accepts an arbitrary address as the owner.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ec8efd52b4b802042c151e1e7b8be2785ab7cd37/contracts/access/Ownable.sol#L26-L30

frangio:
I agree that using `msg.sender` in the constructor was a mistake. As far as I can tell there is nothing we can do in a backwards compatible way, unless Solidity supports optional arguments at some point.

Amxx:
That is not exactly what you are asking for, but if you plan to have your contract inherit from `Ownable` AND have an initializer, maybe the initializer could use the internal function proposed in #2464

thegostep:
Nice - that's also a much needed improvement but I would still request removing setting ownership in the constructor

Amxx:
Have you considered using OwnableUpgradeable from https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable ?

frangio:
We've come to the following conclusion. We would have slightly preferred to have had the `owner` constructor argument since the beginning, but at this point we don't want to make the breaking change in the existing contracts.

The upcoming storage cost changes in the Berlin hard fork, if I'm reading them correctly, seem to drastically reduce the cost of the extra `SSTORE` caused by having to use an internal setter in the constructor body after the default value of `msg.sender` has already been set.

PR #2647, opened 2021-04-27
Ownable should support setting the initial owner through its constructor #2639

Manik-Jain:
Fixes #???? 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


Amxx:
This has already been discussed in #2639 and #2402.
Variables are private for a reason, which is that there are script rules on how can perform a change. Additionally, we need to guaranty that no change happens without an event be emitted.

We have a discussion here #2567 about providing an internal setter. Please contribute to the discussion there.

PR #2651, opened 2021-04-30
ERC20 : Increase Allowance should cater to max value

### LINKED ISSUE #2650: ERC20 : Increase Allowance should cater to max value
increaseAllowance function in ERC20 expect uint256 value, which has no check on the upper bounds. Although the failure returns a false, there should be an early exit with a descriptive message

**💻 Environment**

**📝 Details**
Version : latest
IDE : Remix

addAllowance  function in ERC20 expect uint256 value, which has no check on the upper bounds. Although the failure returns a false, there should be an early exit with a descriptive message

**🔢 Code to reproduce bug**

increaseAllowance(spender, type(uint).max);

Manik-Jain:
Fixes #2650 

1. Spender address should be valid
2. addedValue should be less than maximum value of uint, and sender should be having enough funds in their account to carry out the increase function

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry


frangio:
Thank you @Manik-Jain. We're currently relying on Solidity overflow checks for addition overflow.

See for example `mint`.
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0d40f705a7d4a42ff622ae3a0e1a90305fc5b93e/contracts/token/ERC20/ERC20.sol#L240

PR #2218, opened 2020-04-29
Documentation/erc721 contracts

GuillaumeCz:
Improvements / completions on the ERC721 contracts. 

To do: 
- [x] IERC721Receiver
- [x] ERC721Holder

PR #2653, opened 2021-05-01
Fix `ERC721._beforeTokenTransfer` docs

wchargin:
Fixes a regression introduced in ecf0725dd1bb (#2218). As written, the
docs are contradictory: they document what happens when `from` is zero
and what happens when `to` is zero, then go on to state that neither of
these can be zero. What's meant is what was written in a previous
version of the code: that `from` and `to` are never *both* zero.

wchargin-branch: erc721-beforetokentransfer-docs


Amxx (approved):

frangio:
Thank you @wchargin!

PR #2667, opened 2021-05-11
Fix whitespace in ERC777 

iskanderandrews:
### What did I fix? 
1. Fix a typo in the **ERC777.sol** contract.
2. Fix a typo in the **IERC777Sender.sol** contract
3. **ERC777.sol:** Changing the granularity() function modifier to pure, since this function doesn't need access to any of the contract's states, it's just returning 1.
 
#### PR Checklist

- [ ] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC777/ERC777.sol
+++ b/contracts/token/ERC777/ERC777.sol
@@ -103,7 +103,7 @@ contract ERC777 is Context, IERC777, IERC20 {
      *
      * This implementation always returns `1`.
      */
-    function granularity() public view virtual override returns (uint256) {
+    function granularity() public pure virtual override returns (uint256) {
         return 1;
     }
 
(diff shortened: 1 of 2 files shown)


contracts/token/ERC777/ERC777.sol
```diff
@@ -103,7 +103,7 @@ contract ERC777 is Context, IERC777, IERC20 {
      *
      * This implementation always returns `1`.
      */
-    function granularity() public view virtual override returns (uint256) {
+    function granularity() public pure virtual override returns (uint256) {
```
frangio, line 106: Granularity is defined as `view` so that a child contract can override this function with an implementation that reads from storage. As far as I can tell there is no benefit to use `pure` here.

iskanderandrews: That makes sense.

frangio (approved):
Thanks!

Issue #2665, opened 2021-05-09
Slightly reduce ERC721 deployment gas cost


vladyan18:
**🧐 Motivation**

[ERC721.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/86694813099ba566f227b7d7a46c950baa364b64/contracts/token/ERC721/ERC721.sol#L113)

Changing the order of arguments of first `require`  in  `approve` function can slightly reduce fuel costs for contract deploy.

**📝 Details**

In function

    /**
     * @dev See {IERC721-approve}.
     */
    function approve(address to, uint256 tokenId) public virtual override {
        address owner = ERC721.ownerOf(tokenId);
        require(to != owner, "ERC721: approval to current owner");

        require(_msgSender() == owner || isApprovedForAll(owner, _msgSender()),
            "ERC721: approve caller is not owner nor approved for all"
        );

        _approve(to, tokenId);
    }

Changing the line 

        require(to != owner, "ERC721: approval to current owner");

to

        require(owner != to, "ERC721: approval to current owner");

may lead to a minor decrease in gas cost at deployment. I think the reason is more efficient work with stack.
This does not affect the gas consumption at function call.

Amxx:
Wow @vladyan18, this is an amazing find.

That is great to know. We'll obviously have to double-check that, and to try to understand it better.
How did you notice that?

frangio:
How much of an improvement is it? This should also be reported to the Solidity repository as an opportunity for the optimizer.

vladyan18:
Now i tried to check it in Kovan network and found no improvement(
I relied on gas estimation from truffle test. It was about ~12 gas difference. So it seems that the optimizer is handling this actually. 

But instead of this i found another interesting case [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/86694813099ba566f227b7d7a46c950baa364b64/contracts/token/ERC721/extensions/ERC721Enumerable.sol#L140-L158):

Change [from](https://kovan.etherscan.io/tx/0x87767947223f8af3c911b88ef966f5244065e58dd8ac677176daee50d65bd625):

    function _removeTokenFromAllTokensEnumeration(uint256 tokenId) private {
        // To prevent a gap in the tokens array, we store the last token in the index of the token to delete, and
        // then delete the last slot (swap and pop).

        uint256 lastTokenIndex = _allTokens.length - 1;
[...]

PR #2669, opened 2021-05-11
Tokens: wrap safe substractions in uchecked block

vladyan18:
Discussed in #2665

Reduces gas consumption at deployment and calls.

Just in case ([from](https://docs.soliditylang.org/en/v0.8.0/control-structures.html#checked-or-unchecked-arithmetic)):

> The setting only affects the statements that are syntactically inside the block. Functions called from within an unchecked block do not inherit the property.

#### PR Checklist

- [x] Changelog entry


Amxx (dismissed):
LGTM

PR #2672, opened 2021-05-14
Governor

Amxx:
Adding code for a modular governance system

#### PR Checklist
- [x] Tests
- [x] Documentation
- [ ] Changelog entry


contracts/governance/Governor.sol
```diff
@@ -20,73 +25,39 @@ abstract contract Governor is IGovernor, EIP712, Context {
...
 
+    /**
+     * @dev Sets the value for {name} and {version}
+     */
     constructor(string memory name_, string memory version_)
     EIP712(name_, version_)
     {
-        name = name_;
-        version = version_;
+        _name = name_;
```
nya-elimuai, line 40: Isn't it better to use `_name = name` or `name = _name`? (to avoid the amount of underscores)

Amxx: - `name` is the external viewing function (part of the interface)
- `_name` is the storage slot storing it (private to avoid inheriting contract to accidentally write to it)
- `name_` is the constructor parameter.

All tree have different names to avoid the collision warning.

contracts/utils/Time.sol
```diff
@@ -0,0 +1,56 @@
...
+
+    function setDeadline(Timer storage timer, uint256 timestamp) internal {
+        timer._deadline = timestamp;
+    }
+
+    function reset(Timer storage timer) internal {
+        setDeadline(timer, uint256(ReservedTimestamps.Unset));
+    }
+
+    function lock(Timer storage timer) internal {
```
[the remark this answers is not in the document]
Amxx: The difference between lock and reset is that lock identifies the special without the need for an additional memory slot.

We could use 2 boolean slots, for executed and canceled ... but I think this would be more expensive ... we could also add 2 different states in Timer, like I shown in this issue: https://github.com/ethereum/solidity/issues/11469#issuecomment-853746865

contracts/utils/Time.sol
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    function isPending(Timer memory timer) internal view returns (bool) {
+        // solhint-disable-next-line not-rely-on-time
+        return getDeadline(timer) > block.timestamp;
+    }
+
+    function isExpired(Timer memory timer) internal view returns (bool) {
+        // solhint-disable-next-line not-rely-on-time
+        return uint256(ReservedTimestamps.length) <= getDeadline(timer) && getDeadline(timer) <= block.timestamp;
```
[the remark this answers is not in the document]
Amxx: In taht context, Timer is "memory", not "storage" ... so I'm expecting the improvement to be marginal

contracts/governance/IGovernor.sol
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * @dev Name of the governor instance (used in building the ERC712 domain separator).
+     */
+    function name() external view virtual returns (string memory);
+
+    /**
+     * @dev Version of the governor instance (used in building the ERC712 domain separator).
+     */
+    function version() external view virtual returns (string memory);
```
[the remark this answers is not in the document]
Amxx: EIP712 doesn't strictly need them.

I used not to have them, but I felt like you wanted `name` to be part of the interface ... and if name is felt version should be too. But technically, we can remove them

contracts/governance/IGovernor.sol
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * @dev Proposal snapshot: block number used to retrieve user's votes and quorum.
+     */
+    function proposalSnapshot(uint256 proposalId) public view virtual returns (uint256);
+
+    /**
+     * @dev Proposal weight: amont of votes already casted.
+     */
+    function proposalWeight(uint256 proposalId) public view virtual returns (uint256);
```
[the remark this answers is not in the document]
Amxx: >If someone wants to implement Bravo-style quorum, they have to define proposalWeight = forVotes. Would this be correct?

yes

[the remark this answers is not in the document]
Amxx: I we change the doc toward "total number of votes that count towards quorum", would that resolve this issue ?

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,303 @@
...
+        _proposals[proposalId].timer.lock();
+        _proposals[proposalId].canceled = true;
+        emit ProposalCanceled(proposalId);
+    }
+
+    /**
+     * @dev Internal execute mechanism: verifies that a proposal is successfull, lock the timelock to prevent
+     * re-execution, and calls the {_calls} internal function.
+     *
+     * Note: does not emit any events. Events are part of the public function so they can be customized. In particular
```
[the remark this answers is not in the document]
Amxx: I really don't want to replicate the "ready to execute logic" in queue. Technically queueing is a form of execution...

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,303 @@
...
+        address target,
+        uint256 value,
+        bytes memory data
+    )
+        internal virtual
+    {
+        if (data.length == 0) {
+            Address.sendValue(payable(target), value);
+        } else {
+            Address.functionCallWithValue(target, data, value);
```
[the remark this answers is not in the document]
Amxx: `functionCallWithValue` reverts if the target is not a contract (see line 116 of Address.sol). I personally think this is terrible design!

Also, I don't like using low level calls here, when we have a library for that (which does `_verifyCallResult`)

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,303 @@
...
+     */
+    function execute(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 salt
+    )
+        public payable virtual override returns (uint256 proposalId)
+    {
+        proposalId = _execute(targets, values, calldatas, salt);
```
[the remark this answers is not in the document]
Amxx: I have no strong opinion for that

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,303 @@
...
+        bytes32 salt
+    )
+        internal virtual returns (uint256 proposalId)
+    {
+        proposalId = hashProposal(targets, values, calldatas, salt);
+
+        Proposal storage proposal = _proposals[proposalId];
+        require(proposal.timer.isExpired(), "Governance: proposal not ready");
+        require(proposalWeight(proposalId) >= quorum(proposal.snapshot), "Governance: quorum not reached");
+        require(_voteSuccess(proposalId), "Governance: required score not reached");
```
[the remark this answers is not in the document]
Amxx: IMO, having many storage slots increasses the cost and the complexity. If everything thing is in the same slot, you are sure not to miss checking something else

contracts/governance/Governor.sol
```diff
@@ -0,0 +1,303 @@
...
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 salt
+    )
+        internal virtual returns (uint256 proposalId)
+    {
+        proposalId = hashProposal(targets, values, calldatas, salt);
+        _proposals[proposalId].timer.lock();
+        _proposals[proposalId].canceled = true;
+        emit ProposalCanceled(proposalId);
```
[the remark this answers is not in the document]
Amxx: I'm missing a `require(_proposals[proposalId].timer.isActive())`. Nice catch

contracts/governance/extensions/GovernorWithToken.sol
```diff
@@ -0,0 +1,26 @@
...
+
+import "../IGovernor.sol";
+import "../../token/ERC20/extensions/IERC20Votes.sol";
+
+/**
+ * @dev Extension of {Governor} for voting weight extraction from a Comp or {ERC20Votes} token.
+ *
+ * _Available since v4.2._
+ */
+abstract contract GovernorWithToken is IGovernor {
```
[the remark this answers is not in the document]
Amxx: > What alternatives do you see for us to ship an extension with easy to set up quorum based on % of total supply?

Not sure I understand you question. AFAIK, we can easily make something like 
```
abstract contract GovernorWithTokenQurom is GovernorWithToken {
    uint256 private ratio;
    
    constructor(uint256 _ratio) { ratio = _ratio }
    
    function quorumRatio() external view returns (uint256) { return ratio; }

    function quorum(uint256 blockNumber) public view virtual returns (uint256) {
        return IERC20Votes(token).getPriorTotalSupply(blockNumber) * ratio / 10**18: 
    }
}
```

contracts/governance/extensions/GovernorTimelockExternal.sol
```diff
@@ -0,0 +1,163 @@
...
+            salt,
+            _timelock.getMinDelay()
+        );
+    }
+
+    /**
+     * @dev Public endpoint to update the underlying timelock instance. Restricted to the timelock itself, so updates
+     * must be proposed, scheduled and executed using the {Governor} workflow.
+     */
+    function updateTimelock(address newTimelock) external virtual {
```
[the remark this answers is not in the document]
Amxx: hooo, I'll make that public then

contracts/governance/extensions/GovernorTimelockExternal.sol
```diff
@@ -0,0 +1,163 @@
...
+     */
+    function queue(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 salt
+    )
+        public virtual override returns (uint256 proposalId)
+    {
+        proposalId = _execute(targets, values, calldatas, salt);
```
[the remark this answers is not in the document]
Amxx: if we rename `_execute` to something like `_process` would that be good ?

I don't really like calling `state()` function, because it makes it more difficult to control storage access that can become expensive)

contracts/governance/extensions/GovernorTimelockExternal.sol
```diff
@@ -0,0 +1,163 @@
...
+     * @dev Set the timelock.
+     */
+    constructor(address timelock_) {
+        _updateTimelock(timelock_);
+    }
+
+    /**
+     * @dev Overriden version of the {Governor-state} function with added support for the `Queued` status.
+     */
+    function state(uint256 proposalId) public view virtual override returns (ProposalState) {
```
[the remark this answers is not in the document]
Amxx: good point !

contracts/governance/extensions/GovernorTimelockExternal.sol
```diff
@@ -0,0 +1,163 @@
...
+ * the assets and permissions must be attached to the {TimelockController}. Any asset sent to the {Governor} will be
+ * inaccessible.
+ *
+ * _Available since v4.2._
+ */
+abstract contract GovernorTimelockExternal is IGovernorTimelock, Governor {
+    using Time for Time.Timer;
+
+    TimelockController private _timelock;
+    mapping (uint256 => Time.Timer) private _executionTimers;
```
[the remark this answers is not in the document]
Amxx: We probaly can, try to do that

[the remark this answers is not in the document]
Amxx: One thing is that we would still need a `proposalId` to `timelockProposalId` mapping, because they don't match

contracts/governance/Governor.sol
```diff
@@ -4,208 +4,224 @@ pragma solidity ^0.8.0;
...
     /**
-     * @dev See {IGovernor-version}.
+     * @dev See {IGovernor-vestion}.
+     */
+    function version() public view virtual override returns (string memory) {
+        return "1";
+    }
+
+    /**
+     * @dev See {IGovernor-hashProposal}.
```
frangio, line 63: Should we also document the default hash function in case people want to recreate it outside the contract?

contracts/governance/Governor.sol
```diff
@@ -4,208 +4,224 @@ pragma solidity ^0.8.0;
...
+        } else if (proposal.voteEnd.isPending()) {
+            return ProposalState.Active;
+        } else if (proposal.voteEnd.isExpired()) {
+            return
+                _quorumReached(proposalId) && _voteSuccess(proposalId)
+                    ? ProposalState.Succeeded
+                    : ProposalState.Defeated;
         } else {
-            return ProposalState.Executed;
+            revert("Governor::state: invalid proposal id");
```
frangio, line 94: I think we should use the revert reason convention we've used so far which would just be `Governor: invalid proposal id`.

I think "unknown" proposal id would be better by the way.

contracts/governance/Governor.sol
```diff
@@ -4,208 +4,224 @@ pragma solidity ^0.8.0;
...
+     * @dev See {IGovernor-votingPeriod}
      */
-    function hasVoted(uint256 proposalId, address account) public view virtual override returns (bool) {
-        return _votes[proposalId][account];
-    }
+    function votingPeriod() public view virtual override returns (uint64);
 
     /**
-     * @dev See {IGovernor-hashProposal}.
+     * @dev See {IGovernor-quorum}
```
frangio, line 127: Would be great to include a couple of guidelines or ideas for overriding this.

contracts/governance/extensions/GovernorWithERC20Votes.sol
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @dev Extension of {Governor} for voting weight extraction from a Comp or {ERC20Votes} token.
+ *
+ * _Available since v4.3._
+ */
+abstract contract GovernorWithERC20Votes is IGovernor {
+    ERC20Votes public immutable token;
+
+    constructor(address token_) {
```
frangio, line 16: Now that this contract is not COMP-compatible I definitely think we should add a `quorum` argument. We basically need to express a fraction of the past total supply. We had discussed that using percentages was not necessarily sufficient if someone wants to use a fractional percentage (e.g. 30.5%). We also considered parts per thousand. We could also use 18 decimals just like ETH and most token balances.

I don't know if any of this is worth it. I just had a look at several projects and they all use relatively large numbers, above 25%, except Uniswap at 4%.

It may be better for us to stick to percentages, numbers from 0 to 100, and if someone needs more granularity they can fork this module which is very simple.

contracts/proxy/Proxy.sol
```diff
@@ -63,15 +63,15 @@ abstract contract Proxy {
      * @dev Fallback function that delegates calls to the address returned by `_implementation()`. Will run if no other
      * function in the contract matches the call data.
      */
-    fallback() external payable virtual {
+    fallback() external payable {
```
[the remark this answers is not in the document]
Amxx: THis is an old "prettier" related issue that got lost in the code. Thanks for cathing it !

contracts/governance/Governor.sol
Amxx: I did follow this https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/GovernorAlpha.sol#L227
Revert message compatibility might not be important, in that case I'd be ok to use `Governor: unknown proposal id` instead

contracts/governance/IGovernor.sol
```diff
@@ -2,135 +2,150 @@
...
     /**
      * @dev Emitted when a vote is casted.
      *
      * Note: `support` values should be seen as buckets. There interpretation depends on the voting module used.
      */
-    event VoteCast(address indexed voter, uint256 indexed proposalId, uint8 support, uint256 weight);
+    event VoteCast(address indexed voter, uint256 proposalId, uint8 support, uint256 weight, string reason);
 
     /**
+     * @notice module:core
```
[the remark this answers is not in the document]
Amxx: There are a lot of abstract functons defined here. The idea was to give some ideas about which module is expected to implement them. This @notice might not be clear, I'll have to redact it better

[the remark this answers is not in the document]
frangio: I like the idea, but `@notice` is meant as a notice to the end user according to NatSpec. So I would just find a way to include the info in the normal `@dev` par.

contracts/governance/Governor.sol
frangio: I wouldn't worry about revert reason compatibility, we plan to change them to custom errors soon anyway.

contracts/governance/Governor.sol
```diff
@@ -252,52 +308,39 @@ abstract contract Governor is IGovernor, EIP712, Context {
...
-        internal virtual returns (uint256 weight)
-    {
-        Proposal storage proposal = _proposals[proposalId];
-        require(proposal.timer.isPending(), "Governance: vote not currently active");
-
-        weight = getVotes(account, proposal.snapshot);
+        uint8 support,
+        string memory reason
+    ) internal virtual returns (uint256) {
+        ProposalCore memory proposal = _proposals[proposalId];
```
frangio, line 314: I don't think we need to load the whole thing to memory.

```suggestion
        ProposalCore storage proposal = _proposals[proposalId];
```

contracts/governance/Governor.sol
```diff
@@ -4,200 +4,233 @@ pragma solidity ^0.8.0;
...
+        _proposals[proposalId].executed = true;
 
-        Proposal storage proposal = _proposals[proposalId];
-        require(proposal.timer.isUnset(), "Governance: proposal already exists");
+        for (uint256 i = 0; i < targets.length; ++i) {
+            if (calldatas[i].length == 0) {
+                Address.sendValue(payable(targets[i]), values[i]);
+            } else {
+                Address.functionCallWithValue(targets[i], calldatas[i], values[i]);
+            }
```
frangio, line 224: Re-raising a previous comment in case it gets lost.

I would like us to use `.call`. I think we agreed to make `Address._verifyCallResult` an internal function instead of a private helper, so we can reuse it here.

contracts/governance/IGovernor.sol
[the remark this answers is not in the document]
frangio: I would suggest to document these in the description of the contract. Something like:

> A **vote counting module** needs to implement `_pushVote`, `_quorumReached`, and `_voteSuccess`.

contracts/governance/extensions/GovernorVotingSimple.sol
```diff
@@ -19,63 +19,77 @@ abstract contract GovernorVotingSimple is IGovernor {
...
 
     /**
-     * @dev See {IGovernor-proposalWeight}.
+     * @dev Accessor to the internal vote counts.
      */
-    function proposalWeight(uint256 proposalId) public view virtual override returns (uint256) {
-        return _votings[proposalId].againstVotes
-            + _votings[proposalId].forVotes
-            + _votings[proposalId].abstainVotes;
+    function proposalVotes(uint256 proposalId)
```
frangio, line 41: This function is not a part of any interface. Do we have a use case in mind? I'm considering alternatives like `proposalVotes(uint256 proposalId, uint8 support)` to make it more generic so that other voting modules could define it as well, and that it could be declared in the core.

contracts/governance/extensions/GovernorVotingSimple.sol
```diff
@@ -19,63 +19,77 @@ abstract contract GovernorVotingSimple is IGovernor {
         Abstain
     }
 
-    struct Voting {
+    struct Receipt {
```
frangio, line 22: I would not use the term "receipt" for this. Because in Bravo there is one proposal per account (per proposal). Something like `ProposalVotes` sounds more appropriate to me.

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * @dev See {IGovernorCompatibilityBravo-quorumVotes}.
+     */
+    function quorumVotes() public view virtual override returns (uint256) {
+        return quorum(block.number);
+    }
+
+    // ==================================================== Voting ====================================================
```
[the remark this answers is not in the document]
Amxx: `VotingSimple`'s `_pushVote` does not store everything needed by the compatibility layer.

If we were inheriting `VotingSimple`, then we would need to override `_pushVote` to store additional information. In order to have the receipt structure (from compound) already available, I want to store everything like compound does. considering that the voting it the most gas-critical part, we would have to remove the storage that is done by `VotingSimple`'s `_pushVote` ... and very quickly we are overriding everything in the `VotingSimple` module (since storage is somewhere else).

It felt way simpler to not inherit it, and just include a voting module in the compatibility layer.

contracts/governance/compatibility/IGovernorCompatibilityBravo.sol
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * @dev Interface extension that adds missing functions to the {Governor} core to provide `GovernorBravo` compatibility.
+ *
+ * _Available since v4.3._
+ */
+interface IGovernorCompatibilityBravo is IGovernor {
+    /**
+     * @dev Proposal structure from Compound Governor Bravo
+     */
+    struct Proposal {
```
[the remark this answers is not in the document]
Amxx: This causes an issue. If you put the mapping in the structure, then you get an error when compiling the accessor

```function proposals(uint256 proposalId) public view virtual override returns (Proposal memory)```

it tells you that a proposal that contains a mapping cannot live in memory.

File: contracts/governance/compatibility/IGovernorCompatibilityBravo.sol [hunk 28]
```diff
@@ -0,0 +1,110 @@
...
+        uint96 votes;
+    }
+
+    /**
+     * @dev Part of the Governor Bravo's interface: _"The delay before voting on a proposal may take place, once*
+     * proposed, in blocks"_.
+     *
+     * Note: already defined in {IGovernor}
+     */
+    // function votingDelay() external view returns (uint256);
```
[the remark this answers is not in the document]
Amxx: there are part of `GovernorBravo` so I felt like the should be here, but they are also part of `Governor`. Having them defined in both is a nightmare that I'll have to fix.

contracts/governance/extensions/GovernorVotingSimple.sol
Amxx: honestly, I mostly needed it for testing, and I noticed their was no way to access this data (including from a inheriting contract like a mock) if an accessor was not available.

I thought that having some access would be great ... but yes, it felt module-specific. Not sure how to make it generic. Will think about it.

contracts/governance/compatibility/IGovernorCompatibilityBravo.sol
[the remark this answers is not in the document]
frangio: I understand we may have reasons to remove it, but does it have any observable consequence of the kind I suggested?

contracts/governance/extensions/GovernorVotingSimple.sol
frangio: What about `proposalVotes(uint256 proposalId, uint8 support) returns (uint256)`?

contracts/governance/extensions/GovernorTimelockController.sol
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        return proposalId;
+    }
+
+    /**
+     * @dev Address through which the governor executes action. In this case, the timelock.
+     */
+    function _executor() internal view virtual override returns (address) {
+        return timelock();
```
frangio, line 136: We're reading the state variable directly elsewhere.

```suggestion
        return address(_timelock);
```

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -0,0 +1,251 @@
...
+        Timers.Timestamp timer;
+        bool executed;
+    }
+
+    ICompoundTimelock private _timelock;
+
+    mapping(uint256 => ProposalTimelock) private _proposalTimelocks;
+
+    /**
+     * @dev Emitted when the minimum delay for future operations is modified.
```
frangio, line 88: Is this documentation correct?

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -0,0 +1,251 @@
...
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
+
+        require(state(proposalId) == ProposalState.Succeeded, "Governor: proposal not successfull");
+
+        uint256 eta = block.timestamp + _timelock.delay();
+        _proposalTimelocks[proposalId].timer.setDeadline(SafeCast.toUint64(eta));
```
frangio, line 156: `SafeCast.toUint64(eta))` -> `eta.toUint64()`

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -0,0 +1,251 @@
...
+     * @dev Overloaded execute function that run the already queued proposal through the timelock.
+     */
+    function execute(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public payable virtual override(IGovernor, Governor) returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
+        Address.sendValue(payable(_timelock), msg.value);
```
frangio, line 180: We implemented something slightly different here than what Bravo does. They send `values[i]` with each call to `executeTransaction`. I feel we should do the same, what do you think?

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        return proposalId;
+    }
+
+    /**
+     * @dev Address through which the governor executes action. In this case, the timelock.
+     */
+    function _executor() internal view virtual override returns (address) {
+        return timelock();
```
frangio, line 221:
```suggestion
        return address(_timelock);
```

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    /**
+     * @dev Accept admin right over the timelock.
+     */
+    // solhint-disable-next-line private-vars-leading-underscore
+    function __acceptAdmin() public {
```
frangio, line 228: This is restricted to the guardian in GovernorAlpha. What is our reasoning to keep it open?

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -0,0 +1,251 @@
...
+        require(
+            _timelock.pendingAdmin() != address(0),
+            "GovernorTimelockCompound: old timelock must be transfered before update"
+        );
+        _updateTimelock(newTimelock);
+    }
+
+    function _updateTimelock(address newTimelock) private {
+        emit TimelockChange(address(_timelock), newTimelock);
+        _timelock = ICompoundTimelock(payable(newTimelock));
```
frangio, line 249: If `newTimelock` needs to be payable we can declare that directly in the argument type, here and in the constructor as well.

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -0,0 +1,267 @@
...
+        string[] signatures;
+        bytes[] calldatas;
+        uint256 forVotes;
+        uint256 againstVotes;
+        uint256 abstainVotes;
+        mapping(address => Receipt) receipts;
+        bytes32 descriptionHash;
+    }
+
+    mapping(uint256 => ProposalDetails) internal _proposalDetails;
```
frangio, line 36:
```suggestion
    mapping(uint256 => ProposalDetails) private _proposalDetails;
```

contracts/governance/compatibility/IGovernorCompatibilityBravo.sol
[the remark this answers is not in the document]
Amxx: When looking at the ABI, there might be an issue with the return type indeed, but not the one we expected.

This is the ABI entry from a governor alpha/bravo:
```
{
  "constant":true,
  "inputs":[
    {"internalType":"uint256","name":"","type":"uint256"}
  ],
  "name":"proposals",
  "outputs":[
    {"internalType":"uint256","name":"id","type":"uint256"},
    {"internalType":"address","name":"proposer","type":"address"},
    {"internalType":"uint256","name":"eta","type":"uint256"},
    {"internalType":"uint256","name":"startBlock","type":"uint256"},
    {"internalType":"uint256","name":"endBlock","type":"uint256"},
    {"internalType":"uint256","name":"forVotes","type":"uint256"},
    {"internalType":"uint256","name":"againstVotes","type":"uint256"},
    {"internalType":"bool","name":"canceled","type":"bool"},
    {"internalType":"bool","name":"executed","type":"bool"}
  ],
  "payable":false,
  "stateMutability":"view",
  "type":"function"
},
```

and this is the one for our compatibility layer

```
{
  "inputs":[
    {"internalType":"uint256","name":"proposalId","type":"uint256"}
  ],
  "name":"proposals",
  "outputs":[
    {
      "components":[
        {"internalType":"uint256","name":"id","type":"uint256"},
        {"internalType":"address","name":"proposer","type":"address"},
        {"internalType":"uint256","name":"eta","type":"uint256"},
        {"internalType":"address[]","name":"targets","type":"address[]"},
        {"internalType":"uint256[]","name":"values","type":"uint256[]"},
        {"internalType":"string[]","name":"signatures","type":"string[]"},
        {"internalType":"bytes[]","name":"calldatas","type":"bytes[]"},
        {"internalType":"uint256","name":"startBlock","type":"uint256"},
        {"internalType":"uint256","name":"endBlock","type":"uint256"},
        {"internalType":"uint256","name":"forVotes","type":"uint256"},
        {"internalType":"uint256","name":"againstVotes","type":"uint256"},
        {"internalType":"uint256","name":"abstainVotes","type":"uint256"},
        {"internalType":"bool","name":"canceled","type":"bool"},
        {"internalType":"bool","name":"executed","type":"bool"}
      ],
      "internalType":"struct IGovernorCompound.Proposal",
      "name":"",
      "type":"tuple"
    }
  ],
  "stateMutability":"view",
  "type":"function"
}```

It appears that the default accessor added by solidity with `mapping (uint => Proposal) public proposals;` doesn't just remove the mapping, it also removes the array accesses. That is why they also have the `getActions` function that would otherwise look like a duplicate:

```
function getActions(uint proposalId) public view returns (address[] memory targets, uint[] memory values, string[] memory signatures, bytes[] memory calldatas)
```

I'm going to change our implementation of `proposals` to fix this newly found incompatibility

File: contracts/governance/extensions/GovernorVotingSimple.sol (hunk 24 above)
Comment (Amxx) [reply contracts/governance/extensions/GovernorVotingSimple.sol]: I can imagine module were this wouldn't make sense/be possible to return.


```
abstract contract GovernorCountingScore is Governor {
    struct Voting {
        uint256 supply;
        uint256 score;
        mapping (address => bool) hasVoted;
    }

    mapping (uint256 => Voting) private _votings;
    
    function maxScore() public view virtual returns (uint8) {
        return 100;
    }

    function requiredScore() public view virtual returns (uint8) {
        return 50;
    }

    function proposalWeight(uint256 proposalId) public view virtual override returns (uint256) {
        return _votings[proposalId].supply;
    }

    function proposalScore(uint256 proposalId) public view virtual returns (uint256) {
        return _votings[proposalId].score;
    }

    function hasVoted(uint256 proposalId, address account) public view virtual override returns (bool) {
        return _votings[proposalId].hasVoted[account];
    }

    function _quorumReached(uint256 proposalId) internal view virtual override returns (bool) {
        ProposalVote storage proposalvote = _proposalVotes[proposalId];

        return quorum(proposalSnapshot(proposalId)) < _votings[proposalId].supply;
    }

    function _voteSucceeded(uint256 proposalId) internal view virtual override returns (bool) {
        return _votings[proposalId].supply.requiredScore() < _votings[proposalId].score;
    }

    function _countVote(
        uint256 proposalId,
        address account,
        uint8 support,
        uint256 weight
    ) internal virtual override {
        require(support <= maxScore(), "ScoreVoting: voting over the maximum score");

        Voting storage voting = _votings[proposalId];

        require(!voting.hasVoted[account], "GovernorVotingSimple: vote already casted");
        voting.hasVoted[account] = true;
        voting.supply += weight;
        voting.score  += weight * support;
    }
}
```

File: contracts/governance/compatibility/GovernorCompatibilityBravo.sol [hunk 37]
```diff
@@ -0,0 +1,267 @@
...
+    // public for hooking
+    function proposalEta(uint256 proposalId) public view virtual override returns (uint256);
+
+    // public for hooking
+    function queue(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256);
```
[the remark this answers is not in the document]
Amxx: interface vs abstract contract is a difficult choice.

Making the IGovernor abstract contracts instead of interfaces means we can declare public an internal function in them. This reduces the need for this kind of abstract public hooks indeed.

The downside is that, if someone wants to write an interface of their own, they can't have it inherit from an abstract contract.

contracts/governance/extensions/GovernorTimelockCompound.sol
Amxx: AFAIK, it is safe.

- This operation will only work if the timelock is ready to accept ownership transfer.
- This doesn't modify the timelock address, which was updated by governance

The only case where this has an exact is if the timelock variable points to an instance that is pending ownership transfer. In that case the Governor cannot execute anything until the ownership is accepted. The ownership transfer was proposed by the (still current) timelock owner. I really see no reason why this would be restricted. In fact, I believe it is essential that no-one be able to sensor this operation.

contracts/governance/extensions/GovernorTimelockCompound.sol
Amxx: I don't really like having payable instead of address. It create a discrepancy between the function declaration and its selector (still computed using address). Also, no payment is actually going to happen here, so I think it's better not to create confusion in the eye of whoever is calling the `updateTimelock` function.

Amxx:
TODO: run compound governor tests against our compatibility layer

contracts/governance/extensions/GovernorCountingSimple.sol
```diff
@@ -0,0 +1,99 @@
...
+        )
+    {
+        ProposalVote storage proposalvote = _proposalVotes[proposalId];
+        return (proposalvote.againstVotes, proposalvote.forVotes, proposalvote.abstainVotes);
+    }
+
+    /**
+     * @dev See {Governor-_quorumReached}.
+     */
+    function _quorumReached(uint256 proposalId) internal view virtual override returns (bool) {
```
frangio, line 58: What do you think about turning this into `_quorumReached(uint256 proposalId, uint256 quorumVotes) returns (bool)`? Where `quorumVotes` is the quorum for the proposal, passed in by the core.

1. Every definition of this function otherwise would have to get `quorum(proposalSnapshot(proposalId))`.
2. Changing the function signature in this way makes the counting modules more self-contained.

contracts/governance/extensions/GovernorVotesQuorumFractional.sol
```diff
@@ -0,0 +1,46 @@
...
+
+import "./GovernorVotes.sol";
+
+/**
+ * @dev Extension of {Governor} for voting weight extraction from an {ERC20Votes} token and a quorum expressed as a
+ * fraction of the total supply.
+ *
+ * _Available since v4.3._
+ */
+abstract contract GovernorVotesQuorumFractional is GovernorVotes {
```
frangio, line 13: Suggested renaming:

- `QuorumFractional` -> `QuorumFraction`
- "quorum ratio" -> "quorum numerator"
- "quorum ratio max" -> "quorum denominator"

I'd get rid of the term "ratio" because I don't think it has an obvious interpretation. (The ratio of what to what?)

Did you consider a single getter returning a tuple (numerator, denominator)? It sounds like a reasonable way to represent a rational number.

contracts/governance/extensions/GovernorVotesQuorumFractional.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    function quorumRatioMax() public view virtual returns (uint256) {
+        return 100;
+    }
+
+    function quorum(uint256 blockNumber) public view virtual override returns (uint256) {
+        return (token.getPastTotalSupply(blockNumber) * quorumRatio()) / quorumRatioMax();
+    }
+
+    function updateQuorumRatio(uint256 newQuorumRatio) external virtual onlyGovernance() {
```
frangio, line 34: Does Prettier have an opinion about this? I prefer no parentheses.

```suggestion
    function updateQuorumRatio(uint256 newQuorumRatio) external virtual onlyGovernance {
```

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -0,0 +1,246 @@
...
+     * For security reason, the timelock must be handed over to another admin before setting up a new one. The two
+     * operations (hand over the timelock) and do the update can be batched in a single proposal.
+     *
+     * Note that if the timelock admin has been handed over in a previous operation, we refuse updates made through the
+     * timelock if admin of the timelock has already been accepted and the operation is executed outside the scope of
+     * governance.
+     */
+    function updateTimelock(ICompoundTimelock newTimelock) external virtual onlyGovernance {
+        require(
+            _timelock.admin() == address(this) && _timelock.pendingAdmin() != address(0),
```
frangio, line 236: We said we probably want to remove this check and potentially add a check in `_updateTimelock` to somehow validate the new timelock. The second point is out of the concern that someone might mix up a Compound and OpenZeppelin Timelock with the opposite Governor module, although I recall we discussed this at the very beginning and I don't remember what we said at the time.

Unclear if we want to add this check on the new timelock because it imposes a deployment order, so we have to think what a deployment process would look like.

contracts/governance/extensions/GovernorTimelockController.sol
```diff
@@ -0,0 +1,151 @@
...
+ * delay, enforced by the {TimelockController} to all successfull proposal (in addition to the voting duration). The
+ * {Governor} needs the proposer (an ideally the executor) roles for the {Governor} to work properly.
+ *
+ * Using this model means the proposal will be operated by the {TimelockController} and not by the {Governor}. Thus,
+ * the assets and permissions must be attached to the {TimelockController}. Any asset sent to the {Governor} will be
+ * inaccessible.
+ *
+ * _Available since v4.3._
+ */
+abstract contract GovernorTimelockController is IGovernorTimelock, Governor {
```
frangio, line 20: What do you think about the name `GovernorTimelockControl`? I think the current name is confusing in relation to `TimelockController` itself.

contracts/governance/extensions/GovernorTimelockController.sol
```diff
@@ -0,0 +1,151 @@
...
+    TimelockController private _timelock;
+    mapping(uint256 => bytes32) private _ids;
```
frangio, line 22: Maybe `_timelockIds` would be a more descriptive name.

contracts/governance/extensions/GovernorTimelockController.sol
```diff
@@ -0,0 +1,151 @@
...
+     */
+    function queue(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
+
+        require(state(proposalId) == ProposalState.Succeeded, "Governor: proposal not successfull");
```
frangio, line 88: Can you check for this typo in the entire PR?

```suggestion
        require(state(proposalId) == ProposalState.Succeeded, "Governor: proposal not successful");
```

contracts/governance/extensions/GovernorTimelockController.sol
```diff
@@ -44,16 +44,19 @@ abstract contract GovernorTimelockController is IGovernorTimelock, Governor {
...
-                    ? ProposalState.Executed
-                    : proposalState
-                : proposalState;
+        ProposalState status = super.state(proposalId);
+
+        if (status != ProposalState.Succeeded) {
+            return status;
+        }
+
+        // core tracks execution, so we just have to check if successfull proposal have been queued.
```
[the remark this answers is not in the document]
Amxx: Right, I forgot that ...

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -0,0 +1,276 @@
...
+     */
+    function getReceipt(uint256 proposalId, address voter) public view virtual override returns (Receipt memory) {
+        return _proposalDetails[proposalId].receipts[voter];
+    }
+
+    /**
+     * @dev See {IGovernorCompatibilityBravo-quorumVotes}.
+     */
+    function quorumVotes() public view virtual override returns (uint256) {
+        return quorum(block.number);
```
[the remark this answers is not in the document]
Amxx: the actual quorum checking is done in `_quorumReached` using 

```
return quorum(proposalSnapshot(proposalId)) < details.forVotes;
```

The only test we have with GovernorCompatilibityBravo are with a Comp compatible token, and a fixed quorum. 

```
    function quorum(uint256) public pure override(IGovernor, Governor) returns (uint256) {
        return 0;
    }
    ```

File: contracts/governance/compatibility/GovernorCompatibilityBravo.sol [hunk 47]
```diff
@@ -0,0 +1,276 @@
...
+    ) internal virtual override {
+        ProposalDetails storage details = _proposalDetails[proposalId];
+        Receipt storage receipt = details.receipts[account];
+
+        require(!receipt.hasVoted, "GovernorCompatibilityBravo: vote already casted");
+        receipt.hasVoted = true;
+        receipt.support = support;
+        receipt.votes = SafeCast.toUint96(weight);
+
+        if (support == 0) {
```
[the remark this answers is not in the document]
Amxx: I think I did that to match compound's reverting behaviour. If we use an enum, invalid values would revert in a way that we don't control, and we wouldn't return a revert message like governor bravo does.

I guess that is not something we want to keep

contracts/governance/extensions/GovernorCountingSimple.sol
```diff
@@ -0,0 +1,99 @@
...
+        ProposalVote storage proposalvote = _proposalVotes[proposalId];
+
+        return
+            quorum(proposalSnapshot(proposalId)) <
```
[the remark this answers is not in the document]
Amxx: I hesitated, and was afraid that if the quorum was 0, and no votes were casted at all, a proposal could be seen as successful. Fortunatelly, the success requires STRICTLY more for votes then against votes, so we can do that.

I'm not expecting this to have any real impact in the real world

contracts/governance/extensions/GovernorVotesQuorumFractional.sol
Amxx: I prefer parentheses, and I don't think prettier cares.

Amxx:
Following you comment I did change the compatibility layer's propose function to use a strict inequality
```
        require(
            getVotes(msg.sender, block.number - 1) > proposalThreshold(),
            "GovernorCompatibilityBravo: proposer votes below proposal threshold"
        );
```

I really don't like the new behavior, because even which a `proposalThreshold()` being 0, proposer still need to have some voting power, and for that they need to delegate votes. I personally believe that non-strict comparison would be better here!

contracts/governance/extensions/GovernorCountingSimple.sol
Amxx: I'm not sure it make things clearer, I'll have to think about it.

frangio:
The point about the threshold being 0 is a good one. I'm ok with using `>=` I don't think this impacts compatibility in any meaningful way.

contracts/governance/extensions/GovernorCountingSimple.sol
```diff
@@ -0,0 +1,101 @@
...
+
+import "../Governor.sol";
+
+/**
+ * @dev Extension of {Governor} for simple, 3 options, vote counting.
+ *
+ * _Available since v4.3._
+ */
+abstract contract GovernorCountingSimple is Governor {
+    string public constant COUNTING_MODULE_NAME = type(GovernorCountingSimple).name;
```
frangio, line 13: I think this should be in `IGovernor`, because it should be defined by every counting module, including `GovernorCompatibilityBravo`.

I don't like the name so much because I think is not so much about the "name" but about the way it operates.

My suggestion: `COUNTING_MODE = "ternary"`.

I'm trying to understand what this value should describe exactly. Definitely the accepted `support` values and their meaning.

What about the quorum calculation? Should this describe whether it counts only For votes (as in Bravo), or all votes, or something else?

Amxx: > What about the quorum calculation? Should this describe whether it counts only For votes (as in Bravo), or all votes, or something else?

I think yes

frangio: Our decision. Variable name `COUNTING_MODE`. Value is a URL-encoded key value pair.

For the Bravo compatibility module we'll use `COUNTING_MODE = "support=bravo&quorum=bravo"`.

For CountingSimple we'll use `COUNTING_MODE = "support=bravo&quorum=for,abstain"` assuming we change it to be `forVotes + abstainVotes > quorumVotes`.

contracts/governance/extensions/GovernorTimelockCompound.sol
frangio: Decided we want to remove this check.

frangio (approved):

PR #2679, opened 2021-05-17
Explain how baseURI is used for building the tokenURI

spalladino:
I couldn't find on the docs how baseURI was used. I suspected it was the concatenation, but had to go to the source to confirm.


frangio (approved):
Thanks!

PR #2681, opened 2021-05-19
Add ceiling division operation to the `Math.sol` library

### LINKED ISSUE #2680: Add ceiling division to math library
**🧐 Motivation**

This FR proposes adding a `ceilDiv` method to the `Math.sol` library for performing division that rounds up instead of truncating (standard integer division).

**📝 Details**

Ceiling division is fairly easy to implement, and can either be done with:
- `(a + b - 1) / b`
- `a / b + uint(a % b != 0)`

The second implementation is probably desired, as it does not overflow for large `a` and `b` where the result would fit in a 256-bit integer.

nlordell:
Fixes #2680

This PR adds a `ceilDiv` method to the `Math.sol` library. It uses the `a / b + uint(a % b != 0)` method for the reasons described in the issue.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


Amxx (dismissed):
LGTM

frangio (approved):

frangio (approved):

PR #2682, opened 2021-05-19
Update ERC20Burnable.sol

aress31:
`burnFrom` returns `true` to stay consistent with `transferFrom`

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Thank you @aress31. While we value consistency, we prefer to avoid the boolean return value if it's not providing any information and is not required by the ERC.

PR #2683, opened 2021-05-19
Update ERC20.sol

aress31:
Fix `transferFrom` allowance should be checked before the `_transfer`

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Also, while we do appreciate PRs, we appreciate even more when issues are raised and discussed before a PR is made :)

aress31:
Fair enough, didn't know this issue was already raised, my bad... Still for consistency I would still go with this option even if it means slightly higher gas fees, but that is just me :)

aress31:
That's good to know. Thanks for the answer!

PR #2685, opened 2021-05-20
Reserve unchecked for when guarded by explicit require

frangio:
#2669 introduced `unchecked` blocks for operations that are guaranteed not to overflow. Some of the operations wrapped un `unchecked` only guarantee this through indirect invariants, and we should continue to use checked arithmetic for those. This PR reverts #2669 in those cases, leaving `unchecked` only for the cases when there is an explicit require that guarantees they won't overflow.

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -267,8 +267,8 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
         require(accountBalance >= amount, "ERC20: burn amount exceeds balance");
         unchecked {
             _balances[account] = accountBalance - amount;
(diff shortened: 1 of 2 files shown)


Amxx:
I don't think I agree with that move. Sure the guaranty is indirect, but I still believe the use of private storage is enough to guaranty these invariants.
Its true the gas-saving are really minor, but I also thing the security improvement are null. Only auditability is improved.

I'd even go as far as to say that some + operation can be made unchecked. For example `_balances[account] += amount;` on line 246 of ERC20.sol doesn't need the checking. Checking the totalSupply overflow is enough.

Amxx (approved):

PR #2686, opened 2021-05-20
Add SignedMath with math utilities for signed integers

rotcivegaf:
* [x]  Tests
* [ ] Documentation
* [ ] Changelog entry


Amxx:
Hello @rotcivegaf 

I'm sorry we haven't reached out yet. This PR resurfaced as we are planning for the next release and we would love to merge it. Still some small issues must be addressed:

- In average, the function can overflow. If bot values are `type(int256).max`, then the function will revert despite the average fitting in an int256. I encourage you to have a look at `Math.average` to see how this can be addressed.

- In ceilDiv, you are using a int8 in a place where its not necessary. I would rather use int256 to avoid unnecessary implicit cast.

Can you have a go at this ?

rotcivegaf:
Sorry for the delay, I was very busy

In the second point, when I remove the cast, I get this error from the compiler:
  `TypeError: Type uint8 is not implicitly convertible to expected type int256.`

The other option I thought of is to replace that line with:
```
int256 r;
if(a % b != 0) r = 1;
```

I work in the first point, in this days

Amxx:
Noted.

I think this point is less controversial, as it should not affect the produce bytecode
I think I would personally do

```
        int256 r = a % b == 0 ? int256(0) : int256(1);
```

rotcivegaf:
In the ceilDiv signed average function, I think your idea its good, I commited

In the average signed average function, I try do with this way:

```solidity
    function average(int256 a, int256 b) internal pure returns (int256) {
          return (a / 2) + (b / 2) + ((a % 2 + b % 2) / 2);
    }
```

But when calculate with an even signed number and an odd number, the return number is one(1) integer greater than the expect, this cause because this part of the equation (a % 2 + b % 2), I think the problem is because we have the number 0

I worked in this ugly solution, but I still thinking in another better

```solidity
    function average(int256 a, int256 b) internal pure returns (int256) {
        if (((a < 0 && b > 0) && -a < b) || ((b < 0 && a > 0) && -b < a))
            return (a / 2) + (b / 2) + (((a % 2 + b % 2) - 1) / 2);
        else
            return (a / 2) + (b / 2) + ((a % 2 + b % 2) / 2);
    }
```

frangio:
[This formula](https://stackoverflow.com/a/19110944/667959) on StackOverflow looks really neat:

```
int avg = (a & b) + (a ^ b) / 2;
```

Amxx:
> [This formula](https://stackoverflow.com/a/19110944/667959) on StackOverflow looks really neat:
> 
> ```
> int avg = (a & b) + (a ^ b) / 2;
> ```

should we use that for Unisgned Math ? I guess it would be more gas efficient

rotcivegaf:
Well with unsigned values its more complicated we have the number 0, and this a problem
This implementation `int avg = (a & b) + (a ^ b) / 2;` dont pass this tests:
  - one even signed number and one odd signed number
  - one even signed number and one odd number

But I use this idea in Math.sol to make another PR

frangio:
We want a nicer definition of `div` in order to merge this. We think it should be possible to find one by tweaking the proof for unsigned div formula:
image

frangio:
I keep trying to adapt the formula in an elegant way but I can't find one.

The issue is that the formula seems to round towards negative infinity and we want it to round to zero.

Amxx:
I've refactored the `average` code. 

We still need a changelog entry and some documentation. I'll work on that tomorrow

contracts/utils/math/SignedMath.sol
```diff
@@ -0,0 +1,48 @@
...
+     * towards zero.
+     */
+    function average(int256 a, int256 b) internal pure returns (int256) {
+        // (a + b) / 2 can overflow, and the unsigned formula doesn't simply translate to signed integers
+        int256 base = (a & b) + (a ^ b) / 2;
+        if ((a < 0 && b < 0) || ((a < 0 || b < 0) && (a + b > 0))) {
+            return base + ((a ^ b) % 2);
+        } else {
+            return base;
+        }
```
frangio, line 34: From the book Hacker's Delight, it seems we can say:

```
    function average(int a, int b) public pure returns (int x) {
        x = (a & b) + ((a ^ b) >> 1);
        x += int(uint(x) >> 255) & (a ^ b);
    }
```

test/utils/math/SignedMath.test.js
```diff
@@ -0,0 +1,213 @@
...
+      expect(await this.math.min(max, min)).to.be.bignumber.equal(min);
+    });
+  });
+
+  describe('average', function () {
+    function bnAverage (a, b) {
+      return a.add(b).divn(2);
+    }
+
+    it('is correctly calculated with two odd numbers', async function () {
```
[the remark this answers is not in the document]
frangio: Alternative, put all cases in a single test, automatically generate all combinations using iteration, and include the values in the assertion message.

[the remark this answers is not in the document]
Amxx: I like the idea of generating all combinaison of (meaningfull) values

contracts/utils/math/SignedMath.sol
```diff
@@ -0,0 +1,44 @@
...
+    function min(int256 a, int256 b) internal pure returns (int256) {
+        return a < b ? a : b;
+    }
+
+    /**
+     * @dev Returns the average of two signed numbers. The result is rounded
+     * towards zero.
+     */
+    function average(int256 a, int256 b) internal pure returns (int256) {
+        // (a + b) / 2 can overflow, and the unsigned formula doesn't simply translate to signed integers
```
frangio, line 28:
```suggestion
     * @dev Returns the average of two signed numbers without overflow.
     * The result is rounded towards zero.
     */
    function average(int256 a, int256 b) internal pure returns (int256) {
        // Formula from the book "Hacker's Delight"
```

contracts/utils/math/SignedMath.sol
```diff
@@ -0,0 +1,44 @@
...
+        // Formula from the book "Hacker's Delight"
+        int256 x = (a & b) + ((a ^ b) >> 1);
+        return x + (int256(uint256(x) >> 255) & (a ^ b));
+    }
+
+    /**
+     * @dev Returns the ceiling of the division of two signed numbers.
+     *
+     * This differs from standard division with `/` in that it rounds up instead
+     * of rounding down.
```
frangio, line 37: The language of "rounding up" vs "down" was copied from unsigned math but we need to be careful here. Are we sure we're rounding up? Or are we rounding away from zero, i.e. towards infinity of the same sign?

Amxx: This is "rounding toward infinity of the same sign instead of rounding toward zero".

Maybe ceilDiv is not the right name. Maybe we don't want this function at all, and we want both a real ceilDiv and a floorDiv

frangio (approved):

frangio:
@Amxx I've removed `ceilDiv` since it's not proper "ceiling". Let's add it back if someone requests something like it.

Please merge if it looks good.

rotcivegaf:
I'm afk, but I want to congratulate them
Great work!

PR #2687, opened 2021-05-20
Add optimization for average on Math

rotcivegaf:


Amxx:
Hello @rotcivegaf.

Before erasing code, you should read what you are erasing. In particular, you just propose to remove a comment:
```// (a + b) / 2 can overflow, so we distribute.```
 that explains what we should not do what you propose we do.
 
 Lets imagine that `a = 2**255 - 3` and `b = 2**255-1` both are valid `uint256` and their average (`2**255-2`) is also a valid `uint256`. Yet you code will overflow and throw an error.

PR #2697, opened 2021-05-27
Solidity linting using prettier

Amxx:
Use `prettier-plugin-solidity` to check and fix solidity style.


frangio (commented):
Reasonably good but there are some issues. I've opened issues in the repository and will put this PR on hold for now.

frangio (approved):

PR #2704, opened 2021-06-01
Clones, shorten proxy creation bytecode one byte

### LINKED ISSUE #2689: The `Clones` contract could be further optimized to reduce the cost of cloning
**🧐 Motivation**

While inspecting the OpenZeppelin `Clones` contract, I noticed that the deployment portion of the bytecode seemed like it could be further optimized. Looking to see if this had been discussed elsewhere, I ran across an [OpenZeppelin article](https://blog.openzeppelin.com/deep-dive-into-the-minimal-proxy-contract/) that took note of the same possibility for optimization!

It seems possible to drop 1 byte from this portion of the bytecode.

**📝 Details**
In the `clone` function of the `Clone` contract there is this block of assembly:

```
let ptr := mload(0x40)
mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
mstore(add(ptr, 0x14), shl(0x60, master))
mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
instance := create(0, ptr, 0x37)
```

The bytes that are responsible for getting our `clone` runtime code onto the blockchain are here:

```
0x3d602d80600a3d3981f3
```

This sequence starts by using a `RETURNDATASIZE` to push a `0` onto the stack (3d), so that we can duplicate it (81) later, just before the return (f3). However, by the end of sequence, we have this left over `0` on the stack. If we removed the initial `RETURNDATASIZE` and, instead, just used it later in place of the DUP2 -- we could save a byte. Users would realize the resultant gas savings with every single clone operation.

Since we're shifting things left by one byte from the start, we need to shift many of the relevant pointers by one byte later in the code.

The optimized, functional assembly block would now look like:

```
let ptr := mload(0x40)
mstore(ptr, 0x602d8060093d393df3363d3d373d3d3d363d7300000000000000000000000000)
mstore(add(ptr, 0x13), shl(0x60, master))
mstore(add(ptr, 0x27), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
instance := create(0, ptr, 0x36)
```

And, of course, analogous changes need to be made to the other functions to keep the entire library on the same page.

Unless there is something I am overlooking that would make us prefer the current implementation - I can submit a PR for this. Is there any reason not to implement this?

CallMeGwei:
Fixes #2689  

#### PR Checklist

As the deployed contract bytecode is identical to the previous implementation, tests and documentation should not require any updates. An entry was added to the changelog.

- [ ] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
  * Tokens: Wrap definitely safe subtractions in `unchecked` blocks. 
  * `Math`: Add a `ceilDiv` method for performing ceiling division.
  * `ERC1155Supply`: add a new `ERC1155` extension that keeps track of the totalSupply of each tokenId. ([#2593](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2593))
+ * `Clones`: remove an unnecessary byte from the proxy creation bytecode; saves three gas every clone.
 
  ### Breaking Changes
  
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -25,10 +25,10 @@ library Clones {
         // solhint-disable-next-line no-inline-assembly
         assembly {
             let ptr := mload(0x40)
-            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
-            mstore(add(ptr, 0x14), shl(0x60, implementation))
-            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
-            instance := create(0, ptr, 0x37)
+            mstore(ptr, 0x602d8060093d393df3363d3d373d3d3d363d7300000000000000000000000000)
+            mstore(add(ptr, 0x13), shl(0x60, implementation))
+            mstore(add(ptr, 0x27), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+            instance := create(0, ptr, 0x36)
         }
         require(instance != address(0), "ERC1167: create failed");
     }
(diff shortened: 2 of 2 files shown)


Amxx:
Thank you a lot for this PR, it is very valuable to evaluate gas savings that can be achieved.

I'm not good enough at reading bytecode to do an in-depts analysis, but from what I can measure, this code change saves between 3 and 15 gas per run. It also saves ~650gas when deploying a contract that uses this library. While these improvement are nice to have, they also come with a drawback, that is entirely on me: backward compatibility.

If we do this change, we are changing the way deterministic clones' addresses are generated. This is an issue for two reasons:

- Some contracts might want to use `predictDeterministicAddress(address,bytes32,address)` to determine the address of a clone produced by some other contract. However, if both contracts do not run the same version of the library, there will be inconsistencies. I consider this risk to be low.

 - Some contracts might be upgradeable. Upgrading them to a new version that uses the new library that has this PR will be undetected by our protection mechanism, and will break assumptions about the uniqueness of a clone with a given salt (on top of breaking predictability). I believe this is very serious issue. 
 
 I'm really sorry for the time you spent on this PR. Your contribution is really appreciated, but I'm personally not comfortable with merging it considering the balance between the (somehow low) gas-saving and the fact we could be breaking things.

Dylan-Kerler:
@CallMeGwei I think an additional change is needed in the second portion of the byte code:

```javascript
0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000
//                                              ^^ This jumpdest (2b) should be changed to (2a) because of the 1 byte left offset
```

@Amxx Also I think breaking things is how progress is made. If we are worried about backwards compatibility nothing would ever move forward. Look at apple and the USBC change or even the solidity compiler and their v0.8 change. If you're not willing to break things to move forward then you become outdated and eventually someone else will create a new product to replace you. 

Any team who is serious about using solidity should have already locked down their library versions so I don't think the risk of your second point is that great.

frangio:
We are willing to break things to move forward, as we've done in the past, in a structured and not so frequent way, and for good enough reasons.

Due to the pace of breaking changes in the Solidity compiler, a lot of users remain of past versions unable to benefit from the features in the more recent versions. See the list of [recently verified contracts](https://etherscan.io/contractsVerified) on Etherscan for reference.

Amxx:
The contract we provide comes in an upgradeable flavor. This means that developers using one version of contracts may (and they actually do) want to upgrade to use the newest module we have for them. When they do such as upgrade, they expect things that are unmodified to continue working. ERC20 balances should not disappear, events should be emitted in the same way, and predictable behavior such as the address of clones deployed through create2 should remain consistent.

PR #2707, opened 2021-06-04
Add an Enumerable extension to ERC1155

### LINKED ISSUE #2705: [ERC1155] Provide IERC1155Enumerable interface to manipulate tokens of specified account
**🧐 Motivation**
Support token enumeration of **specified account** in ERC1155 standard contract.

**📝 Details**
Our dapp want to list tokens in a pagination view for a specified ERC1155 contract. So, we want 2 enumeration APIs as following:
1. `tokenCountOf(address account) public view returns (uint256)`.
2. `tokensOfByIndex(address account, uint256 offset, uint256 limit) public view returns (uint256[] memeory)`.

We strongly recommend to define such functions as a ERC1155 extension methods, so as to develop NFT exhibition alike dapp much easiler.

Moreover, NFT images are usually required on UI page, so batch API to fetch URIs of multiple tokens is really helpful to reduce the times of RPC call (`eth_call`).

Amxx:
Fixes #2705

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,15 @@
  * Enumerables: Improve gas cost of removal in `EnumerableSet` and `EnumerableMap`.
  * Enumerables: Improve gas cost of lookup in `EnumerableSet` and `EnumerableMap`.
  * `Counter`: add a reset method. ([#2678](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2678))
- * Tokens: Wrap definitely safe subtractions in `unchecked` blocks. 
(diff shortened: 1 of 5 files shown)


frangio:
I'm currently against this extension because people generally want enumerability in order to list tokens off-chain, but we should recommend a different approach that doesn't incur gas costs for users.

Amxx:
While it's true most request didn't been this to be on-chain, and are better served by an off-chain solution, I still believe there might be use-cases where is code is useful.

I'm marking this PR "on hold" for now

PR #2710, opened 2021-06-07
Add a BitMap struct

Amxx:
Native bitmap are really storage inneficient:

`mapping(X => bool)` will use a new storage slot for each entry
`bool[]` will use an underlying `bytes1[]` array, which only truly uses 32 bits per 256 bits storage slot.

This PR include a dense bitmap, that will save on gas when entries are consecutive. This design is used by Uniswap [merkle-distributor](https://github.com/Uniswap/merkle-distributor/blob/master/contracts/MerkleDistributor.sol) to reduce the gas cost of writing to new (clean) storage slot by a factor 8 over what bool[] would have required.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


contracts/utils/structs/BitMap.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Library for managing uint256 to bool mapping in a compact and efficient way, providing the keys are sequential.
+ * Largelly inspired by Uniswap's https://github.com/Uniswap/merkle-distributor/blob/master/contracts/MerkleDistributor.sol[merkle-distributor].
+ */
+library BitMap {
+    struct UintBitMap {
```
frangio, line 9: Do we plan to have any non-`Uint` variants of this? Can we just call it `BitMap`?

contracts/utils/structs/BitMap.sol
```diff
@@ -0,0 +1,32 @@
...
+        uint256 pos = index % 256;
+        uint256 word = bitmap._data[bucket];
+        uint256 mask = (1 << pos);
+        return word & mask == mask;
+    }
+
+    function set(UintBitMap storage bitmap, uint256 index) internal {
+        uint256 bucket = index / 256;
+        uint256 pos = index % 256;
+        bitmap._data[bucket] = bitmap._data[bucket] | (1 << pos);
```
frangio, line 24: Any reason not to use this?

```suggestion
        bitmap._data[bucket] |= (1 << pos);
```

contracts/utils/structs/BitMap.sol
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    function unset(UintBitMap storage bitmap, uint256 index) internal {
+        uint256 bucket = index / 256;
+        uint256 pos = index % 256;
+        bitmap._data[bucket] = bitmap._data[bucket] & ~(1 << pos);
```
frangio, line 30:
```suggestion
        bitmap._data[bucket] &= ~(1 << pos);
```

contracts/utils/structs/BitMap.sol
```diff
@@ -0,0 +1,32 @@
...
+    struct UintBitMap {
+        mapping(uint256 => uint256) _data;
+    }
+
+    function get(UintBitMap storage bitmap, uint256 index) internal view returns (bool) {
+        uint256 bucket = index / 256;
+        uint256 pos = index % 256;
+        uint256 word = bitmap._data[bucket];
+        uint256 mask = (1 << pos);
+        return word & mask == mask;
```
frangio, line 18: I'm guessing this will be a little more efficient, since `mask` doesn't have to be reused it needs less stack manipualtion. It also personally makes more sense to me.

```suggestion
        return word & mask != 0;
```

frangio (commented):
Code and tests look good to me.

contracts/utils/structs/BitMap.sol
Amxx: Having a struct with the same name as the library its in causes warning message.

frangio: Looking at our other libraries, they are usually plural, so we should name the library `BitMaps` and the struct `BitMap`. This avoids the warning message.

frangio (dismissed):
Looks good, though I wonder if we also want a `set(BitMap, uint256, bool)` function. Could be useful in some cases.

frangio (approved):

PR #2712, opened 2021-06-10
Use hardhat-exposed plugin to remove most mocks

frangio:
We are looking into removing mock contracts to make maintainability easier. The [hardhat-exposed](https://github.com/frangio/hardhat-exposed) plugin I built automatically creates "mock" contracts that 1) expose internal functions as external functions, and 2) add a constructor with the necessary parameters to make the contract concrete (unless it has unimplemented functions).

The "exposed" contracts have an `X` prefix, and the automatic external functions have an `x` prefix. I would've liked to use a `$` prefix for both but this is currently not supported by Ethers so I went with `x`, although since our test suite doesn't use Ethers we could go with `$` if I make the plugin configurable.

Some mocks need to remain but  we can make them simpler by leaving it to the plugin to expose the internal functions. For more detail refer to the readme of [hardhat-exposed](https://github.com/frangio/hardhat-exposed).

---

In this PR I also included the removal of `chainId` functions in the mock contracts. We don't need this anymore because solidity-coverage runs in Hardhat EVM now.


frangio:
Compilation times double with this plugin. Currently it increases in this repo from 5 seconds to 11 or so. This will likely go down once we remove most of the manual mocks.

test/metatx/ERC2771Context.test.js
```diff
@@ -22,7 +23,7 @@ contract('ERC2771Context', function (accounts) {
     this.domain = {
       name,
       version,
-      chainId: await web3.eth.getChainId(),
```
Amxx, line 25: what is wrong with `web3.eth.getChainId()`, and why do we need an helper to do the rpc call when web3 does it right ?

frangio:
Comments by @Amxx:

> In some cases we want to test a modifier (see `AccessControlMock`)
> Overloaded functions (like in `SafeMath`) are difficult test 
> How to test `StorageSlot`

PR #2713, opened 2021-06-11
add a recover(bytes32,bytes32,bytes32) that follows eip2098

Amxx:
ECDSA currently supports the following checks

- `recover(bytes32 hash, bytes signature)`
- `recover(bytes32 hash, bytes32 r, bytes32 v, uint8 s)`

The first instance supports both 65 bytes and 64 bytes (short) signatures.

This PR add a third option:

- `recover(bytes32 hash, bytes32 r, bytes32 vs)`

which can be called if with a short signature directly, and is used by the `recover(bytes32 hash, bytes signature)` variant when signatures are 64 bytes long.

This is already covered by the existing tests

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -28,15 +28,13 @@ library ECDSA {
      * - with https://docs.ethers.io/v5/api/signer/#Signer-signMessage[ethers]
      */
     function recover(bytes32 hash, bytes memory signature) internal pure returns (address) {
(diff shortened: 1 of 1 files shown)


frangio (approved):
It's getting complicated to communicate which one people should use.

I'm thinking maybe all overloads should be presented in the docs site as different signatures for the same function. Just an idea to consider in the future.

PR #2716, opened 2021-06-13
UnSafeMath upgrade

jessielesbian:
Please use my gas optimizations.


Amxx:
Hello @jessielesbian.

We did consider doing these optimizations. They are guaranteed safe by contract-wide invariants but not by local "in-function" checks. There is debate either these contract-wide invariants should justify the use of unchecked here. particularly since they can be broken by some implementation that would tamper with the storage slots.

Have you measured the gas savings, and are they affecting your application?

frangio:
Closing because of the reasons given on the last comment.

PR #2727, opened 2021-06-17
Check inheritance tree consistency

Amxx:
In order to build contracts out of modules, it must be possible to linearize the inheritance tree.
One way to check that this will always be possible is to ensure that there exists an ordering of the contract that matches the inheritance patterns.

This PR adds a script that verifies such an ordering exist by building a graph of dependency relative orders and verifying that this graph is acyclic.

This script can be run with `npm run test:inheritance`, and is added to the CI workflow

Note that, when contracts are modified and recompiled, the old manifest in `artifacts/build-info` must be deleted.


frangio (approved):
LGTM. Thanks!

PR #2739, opened 2021-06-24
Allow Ownable to work with meta-trasnactions (ERC2771Context)

drortirosh:
The following code fails to compile:

```javascript
  import "@openzeppelin/contracts/metatx/ERC2771Context.sol";
  import "@openzeppelin/contracts/access/Ownable.sol";

  contract MyContract is ERC2771Context, Ownable {
  
    constructor(address forwarder_) 
      ERC2771Context(forwarder_) {
    }
```
The error: 
```javascript
TypeError: Immutable variables cannot be read during contract creation time, which means they cannot be read in the constructor or any function or modifier called from it.
  --> @openzeppelin/contracts/metatx/ERC2771Context.sol:18:29:
   |
18 |         return forwarder == _trustedForwarder;
   |                             ^^^^^^^^^^^^^^^^^
```

The reason for this problem, is that `Ownable` attempt to set the owner, and uses `_msgSender()` in its constructor.
However, this fails to compile, since it is forbidden to access members from a constructor code (the `trustedForwarder` address)
Luckily, it is OK to use the normal `msg.sender` in this case, since a constructor call can't be called directly through meta-tx anyway..

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -25,7 +25,11 @@ abstract contract Ownable is Context {
      * @dev Initializes the contract setting the deployer as the initial owner.
      */
     constructor() {
-        _setOwner(_msgSender());
+        // this is one of the only places that _msgSender() should NOT be used:
+        // a constructor can't be called directly through meta-transactions.
+    	// and also _msgSender() references a member (_trustedForwrader) - which is forbidden
+    	// from a constructor call..
+        _setOwner(msg.sender);
     }
 
     /**


Amxx:
WOW, thank you for this finding @drortirosh.

This is a serious issue that we will try to address quickly!

monkeydluffyto (approved):

PR #2742, opened 2021-06-27
changed possibly wrong comma to decimal

sudo-vaibhav:
Fixes #???? 

There was a comma in one of the comments which seemed like a typo and should be a decimal sign instead according to me.

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [ ] Changelog entry


frangio (approved):
Thanks! This wasn't necessarily a typo, other languages use comma for decimals, but we should stick to the English convention.

Issue #1214, opened 2018-08-16
Make TokenVesting more aligned with usual requirements


frangio:
@alcuadrado mentioned to me that our `TokenVesting` contract is not aligned with the usual requirements for vesting, so some potential users have had to write their own implementations.

For example, our contract does a "continuous" release of tokens, whereas the usual requirement is for tokens to be released e.g. every month.

We should review the contrct to conform to what users need.

I'm a bit concerned that this will be a breaking change but we won't be able to get it done for v2.0. So perhaps we should look into adding any parameters that we need now, and implement support for them later. For example, for the issue I described above we could add a "granularity" parameter in the constructor (problaby a bad name) and for now require that it is always zero. Just a thought.

nventuro:
What if we kept it out of the 2.0 release, since we know we're going to revamp it? It could be then later added in a minor release.

frangio:
@rstormsf We actually want to keep the current model where there is one vesting contract deployed per schedule per beneficiary, and Colony's `Vesting` is quite the opposite since it is a central contract that is used for all beneficiaries. Do you see other problems with our current `TokenVesting` that Colony's contract solves?

The continuous release is the biggest issue with our current offering. We need the possibility to make it non-continuous because that's the way actual paper-based vesting contracts are signed. Those may, for example, vest monthly, i.e. release only at the end of each month.

The frequency is one of the parameters that we have to add, but we need to do some more research on the typical vesting schedules used and try to support them. There may be other parameters necessary.

frangio:
@rstormsf Because it better sandboxes each investor's funds. It's also a simpler implementation, resulting in less attack surface.

nventuro:
#1040 by @ninjaahhh proposed graded vesting, where the amount of tokens released during each period doesn't remain the same (in their project, it increased as time went by). We may want to consider adding this to our implementation.

thedon-chris:
As a token contract owner, I would like to vest tokens to recipients on a “monthly” & non-linear basis, so that tokens are released on a controlled monthly interval (_as opposed to token holders withdrawing at their leisure_)

djenning90:
I have something I'd like to share that might be of interest to this group.

[...]

PR #2748, opened 2021-07-01
Add a VestingWallet 

Amxx:
Add a vesting wallet with a default linear schedule. The vesting schedule can be customized by overriding a single function.

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


contracts/finance/VestingWallet.sol
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @dev Amont of token already released
+     */
+    function released(address token) public view returns (uint256) {
+        return _released[token];
+    }
+
+    /**
+     * @dev Release the tokens that have already vested.
```
frangio, line 84:
```suggestion
     * @dev Release the tokens that have already vested.
     *
     * Emits a {TokensReleased} event.
```

We used to have these notes where we emit an event. We haven't been doing a good job of adding them for new contracts lately. Let's try to start adding them. (Though it would be nice if we could automate it with `solidity-docgen`. :thinking:)

contracts/finance/VestingWallet.sol
```diff
@@ -0,0 +1,112 @@
...
+        _;
+    }
+
+    /**
+     * @dev Set the beneficiary, start timestamp and vesting duration of the vesting wallet.
+     */
+    constructor(
+        address initialBeneficiary,
+        uint256 initialStart,
+        uint256 initialDuration
```
frangio, line 40: Sorry I don't think these are good because it can mislead into thinking the values could change later. I would either keep the underscore suffixes or go with `beneficiaryAddress`, `startTimestamp`, `durationSeconds`.

test/finance/VestingWallet.test.js
```diff
@@ -0,0 +1,118 @@
...
+    this.start = (await time.latest()).addn(3600); // in 1 hour
+    this.token = await ERC20VotesMock.new('Name', 'Symbol');
+    this.vesting = await VestingWallet.new(beneficiary, this.start, duration);
+    await this.token.mint(this.vesting.address, amount);
+
+    this.schedule = Array(256).fill()
+      .map((_, i) => web3.utils.toBN(i).mul(duration).divn(224).add(this.start))
+      .map(timestamp => ({
+        timestamp,
+        vested: min(amount.mul(timestamp.sub(this.start)).div(duration), amount),
```
frangio, line 25:
```suggestion
        vested: min(amount, amount.mul(timestamp.sub(this.start)).div(duration)),
```

I think this is easier to read.

contracts/finance/VestingWallet.sol
```diff
@@ -0,0 +1,112 @@
...
+ * @dev This contract handles the vesting of ERC20 tokens for a given beneficiary. Custody of multiple tokens can be
+ * given to this contract, which will release the token to the beneficiary following a given vesting schedule. The
+ * vesting scheduled is customizable through the `vestedAmount(address,uint256)` function.
+ *
+ * Any token transferred to this contract will follow the vesting schedule as if they were locked from the beginning.
+ * Consequently, if the vesting has already started, any amount of tokens sent to this contract will (at least partly)
+ * be immediately releasable.
+ *
+ * While tokens are locked, the beneficiary still has the ability to delegate the voting power potentially associated
+ * with these tokens.
```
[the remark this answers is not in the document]
Amxx: the delegation part ... maybe

I have the feeling that more people will want it and forget than not want it. I'd love to have more opinions about that

contracts/finance/VestingWallet.sol
```diff
@@ -0,0 +1,112 @@
...
+            return _historicalBalance(token);
+        } else {
+            return (_historicalBalance(token) * (timestamp - start())) / duration();
+        }
+    }
+
+    /**
+     * @dev Calculates the historical balance (current balance + already released balance).
+     */
+    function _historicalBalance(address token) internal view returns (uint256) {
```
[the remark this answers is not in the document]
Amxx: I is private in Fortify, but overloading contract will want to access it.

frangio:
We should check if the contract we're proposing supports the use cases listed in the comments in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1214.

Amxx:
> We should check if the contract we're proposing supports the use cases listed in the comments in #1214.

- "Graded vesting" can be implemented by overriding the release function with any shape/curve
- Monthly” & non-linear basis can also be implemented by overriding the release function. I have code for that, which is not part of the "small" PR but that we used to test.

contracts/finance/vesting/ERC20VestingVoting.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../token/ERC20/extensions/ERC20Votes.sol";
+import "./ERC20VestingWallet.sol";
+
+/**
+ * @title ERC20VestingWalletVotes
```
frangio, line 8:
```suggestion
 * @title ERC20VestingWalletVoting
```

test/finance/vesting/ERC20VestingWallet.test.js
```diff
@@ -0,0 +1,87 @@
+const { constants, expectEvent, expectRevert, time } = require('@openzeppelin/test-helpers');
+const { expect } = require('chai');
+
+const ERC20VotesMock = artifacts.require('ERC20VotesMock');
+const ERC20VestingWallet = artifacts.require('ERC20VestingWallet');
+
+const min = (...args) => args.slice(1).reduce((x, y) => x.lt(y) ? x : y, args[0]);
+
+contract('ERC20VestingWallet', function (accounts) {
```
frangio, line 9: What do you think about defining a behavior instead of repeating the tests?

frangio: Bump!

Amxx: I gave it a try yesterday, and it was more difficult then I expected. Schedules a regenerated in beforeEach with every new instance, and I wanted to make the behaviour schedule-agnostic, through a generator function.

frangio: It sounds like we should be able to make it generic with a function `start x duration x timestamp -> number`.

contracts/finance/VestingWallet.sol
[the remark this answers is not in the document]
noobshow: More would definitely want it, than not...

noobshow:
Hi @Amxx 
Great work on this code...

I need something like this and seen it hasn't been published and many tests are even failing(probably for some known reasons)...

I want to know if this is usable at its current stage?

Would appreciate some info if there's any help needed to move it forward too...

Thanks

frangio:
We've decided to offer a single `VestingWallet` for both ERC20 and native token. The `release` function should operate the native token if the token address is a specific value to represent it, possibly `0x00...00` or `0xee...ee`. Personally I'm leaning towards `0x0` because the "e" doesn't represent native token in other networks, and because the zero address is easily available as a constant in Ethers.js. Additionally there should be a separate release function for the native token, ideally with a different name so it doesn't cause function overloading.

NeoXtreem:
@frangio What's the planned release date for this `VestingWallet`?

PR #2754, opened 2021-07-06
Use private variable instead of immutable to allow calls to _msgSender in constructor.

Amxx:
Immutable _trustedForwarder causes issues when _msgSender() is called in constructors

Fixes #2739

#### PR Checklist

- [x] Tests
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+ * `ERC2771Context`: use private variable from storage to store the forwarder address. Fixes issues where `_msgSender()` was not callable from constructors. ([#2754](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2754))
+
 ## 4.2.0 (2021-06-30)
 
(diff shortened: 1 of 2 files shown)


frangio (approved):

drortirosh:
Sorry, but I think this solution is sub-optimal, compared to the solution in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2739/files
- the `_trustedForwarder` is a member modified only by the constructor. there is no need to make it `private` (which also costs more.)
- the issue is not with the `ERC2771Context` at all (which indeed, works with any other use-case, e.g. with ERC20, AccessControl ) but with the `Ownable` contract: `_msgSender()` is completely meaningless in the constructor code, as it can't be called directly via a meta-transaction call.

frangio:
@drortirosh Thanks for the feedback. The reasons why we went with this change instead are:

1) The problem extends beyond `Ownable`. Contracts like `AccessControl` are also affected.
2) In some of those cases, `_msgSender` is not used directly in the constructor but indirectly through an internal function, and this becomes a lot harder to work around.
2) Other variants of `Context._msgSender` could technically make sense in the constructor, so we prefer keeping it there.

Additionally, the Solidity team has hinted that this limitation could soon be lifted at the compiler level, and we plan to add `immutable` back then. https://github.com/ethereum/solidity/issues/10463#issuecomment-872280433

PR #2767, opened 2021-07-14
Update and rename ReentrancyGuard.sol to MultiLevelReentrancyGuard.sol

sudiptab2100:
There might be multiple functions that needs to be `nonReentrant` for calls to themselves only, but using one modifier for every nonReentrant functions can pause the working of other functions when one is working. 
To solve the problem, this solution is proposed a multi level Reentrancy Guard. In this implementation, `nonReentrant` modifiers will have a parameter called `level`. This way non reentrant functions with different level can be executed concurrently & non reentrant functions that should not be working concurrently can be put in same level.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/security/MultiLevelReentrancyGuard.sol
+++ b/contracts/security/MultiLevelReentrancyGuard.sol
@@ -3,20 +3,15 @@
 pragma solidity ^0.8.0;
 
 /**
- * @dev Contract module that helps prevent reentrant calls to a function.
+ * @dev Contract module that helps prevent reentrant calls to functions.
  *
  * Inheriting from `ReentrancyGuard` will make the {nonReentrant} modifier
  * available, which can be applied to functions to make sure there are no nested
  * (reentrant) calls to them.
  *
- * Note that because there is a single `nonReentrant` guard, functions marked as
- * `nonReentrant` may not call one another. This can be worked around by making
- * those functions `private`, and then adding `external` `nonReentrant` entry
- * points to them.
+ * Note: functions marked as `nonReentrant` with same `level`
+ * may not call one another. 
  *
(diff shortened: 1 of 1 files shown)


Thajuiceman69 (approved):

Amxx:
Hello @sudiptab2100, and thank you for the proposal.

In the future, please raise an issue before proposing a PR. It is always better to discuss the feature before puting time into the code.

In its current form, your proposal has 2 big issues:

- It breaks backward compatibility with contract that expect te modifier to not have any argument.
- Using values 1 and 2 instead of 1 and 2 is very bad for gas usage. we want te state change to be as cheap as possible, and it is cheaper to update dirty storage (2 x 5k gas) then it is to set it from clean state (21k gas) and then cleaning with a refund.

I'd love to know more about the usecase. In any case, I would keep the old modifier, and I would make sure the gas usage are optimized

Amxx (changes requested):
- Keep original modifier in the code (new features can be in a different modifier)
- Keep `uint256 private constant _NOT_ENTERED = 1;` and `uint256 private constant _ENTERED = 2;` for gas efficiency

Thajuiceman69:
Thanks

Amxx:
Also, keep in mind that we want to keep storage compatibility in case user want to upgrade their contract.
Here, actual storage content is not really important, but we would still want to keep the `uint256 status` slot to avoid the upgrade plugins detecting what appears like an inconsistency

frangio:
@Thajuiceman69 Can you share an example use case to motivate this feature so we can decide if we want to include it?

Issue #2390, opened 2020-10-19
Add 'paginated' getters for EnumerableSet and Map


nventuro:
The current API only provides access via `.length()` and `.at()`. It'd be great to have a `.list(start, end)` that returned a _partial_ view of the set's elements.

The reasoning behind making this a non-complete view is to let the caller decide how much gas they want to spend. Calls to `view` functions from off-chain sources (i.e. `eth_call`) are _also_ [capped in gas](https://infura.io/docs/ethereum/json-rpc/eth-call#eth_call), and a large enough set can cause them to fail.

The implementation should be quite simple:
```solidity
function list(
    EnumerableSet.AddressSet storage set,
    uint256 start,
    uint256 end
) internal view returns (address[] memory) {
    address[] memory elements = new address[](end - start);

    for (uint256 i = 0; i < elements.length; ++i) {
        elements[i] = set.at(i + start);
    }

    return elements;
}
```

abcoathup:
Hi @nventuro !  Thanks for the suggestion, it is really appreciated.

The project owner will review your suggestion as soon as they can. 

Please wait until we have discussed this idea before writing any code or submitting a Pull Request, so we can go through the design beforehand.  We don’t want you to waste your time!

ducquangkstn:
I would like to have some kind of listAll function
```solidity
function listAll(
    EnumerableSet.AddressSet storage set
) internal view returns (address[] memory);
```

Amxx:
Hello @ducquangkstn 

This was already requested many times, but the issue remains the same. AddressSet, and all the other sets, use a `bytes32[]` based underlying storage. This means that the values are stored as a `bytes32[]`. Unfortunatelly, it is not possible to cast a `bytes32[]` to a `address[]` easily. This casting would require creating a new array, and casting all the elements one by one.

This would very gas expensive to do it onchain, and it feels like gettings elements on request using the `.at()` function is a better option, with no hidden costs. If you really need the array of values, you can access the bytes32[]  version, but you'll have to process it accordingly.

nventuro:
>Unfortunatelly, it is not possible to cast a bytes32[] to a address[] easily

Actually I think this is not that difficult using inline assembly:

```
pragma solidity ^0.8.0;

contract A {
    bytes32[] private _foo;
    
    function foo() internal view returns (bytes32[] storage) {
        return _foo;
    }
    
    function fooAsAddress() internal view returns (address[] storage) {
[...]

PR #2768, opened 2021-07-14
Add values() functions to EnumerableSets

Amxx:
Address request from #2390

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -130,6 +130,13 @@ library EnumerableSet {
         return set._values[index];
     }
 
+    /**
+     * @dev Return the entier set in an array
+     */
+    function _values(Set storage set) private view returns (bytes32[] memory) {
+        return set._values;
+    }
+
     // Bytes32Set
 
     struct Bytes32Set {
@@ -188,7 +195,7 @@ library EnumerableSet {
      * @dev Return the entier set in an array
      */
     function values(Bytes32Set storage set) internal view returns (bytes32[] memory) {
-        return set._inner._values;
+        return _values(set._inner);
     }
 
     // AddressSet
@@ -249,11 +256,11 @@ library EnumerableSet {
      * @dev Return the entier set in an array
      */
     function values(AddressSet storage set) internal view returns (address[] memory) {
-        bytes32[] storage store = set._inner._values;
-        address[] storage result;
+        bytes32[] memory store = _values(set._inner);
+        address[] memory result;
 
         assembly {
-            result.slot := store.slot
+            result := store
         }
 
         return result;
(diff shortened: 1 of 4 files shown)


contracts/utils/structs/EnumerableSet.sol
```diff
@@ -184,6 +184,13 @@ library EnumerableSet {
         return _at(set._inner, index);
     }
 
+    /**
+     * @dev Return the entier set in an array
+     */
+    function values(Bytes32Set storage set) internal view returns (bytes32[] memory) {
+        return set._inner._values;
+    }
```
nventuro, line 192: Note that returning a `memory` array implies reading the entire thing from storage and copying it - not sure if that was the original intent

nventuro: Also, shouldn't this be implemented for `Set` as `Set._values()`, and call that here?

contracts/utils/structs/EnumerableSet.sol
```diff
@@ -238,6 +245,20 @@ library EnumerableSet {
...
+     */
+    function values(AddressSet storage set) internal view returns (address[] memory) {
+        bytes32[] storage store = set._inner._values;
+        address[] storage result;
+
+        assembly {
+            result.slot := store.slot
+        }
+
+        return result;
```
nventuro, line 259: If you're returning `memory` arrays and not storage ones, this is how you'd cast a memory array:

```suggestion
        bytes32[] memory store = _values();
        address[] memory result;

        assembly {
            result := store
        }

        return result;
```

This is essentially a pointer copy.

contracts/utils/structs/EnumerableSet.sol
Amxx: Returning a storage pointer will potentially encourage the users to write/modify it, which would break the structure. We really want to avoid this.

I'll consider adding a `Set._values()`. For it to make sense it should return a storage pointer otherwize specialized versions will not be able to do the slot trick for casting ... but as long as it is private, I'd be ok with that.

contracts/utils/structs/EnumerableSet.sol
Amxx: I didn't know you could do a memory array cast like that, definitelly sounds like the right solution !

contracts/utils/structs/EnumerableSet.sol
```diff
@@ -130,6 +130,13 @@ library EnumerableSet {
         return set._values[index];
     }
 
+    /**
+     * @dev Return the entier set in an array
```
frangio, line 134: We should update the documentation to warn people to only use this for off-chain querying and be mindful that it can cause out of gas errors.

Also note the typo in "entier" which should be "entire".

frangio (approved):

PR #2781, opened 2021-07-19
More conservative audit language

### LINKED ISSUE #2780: Language about audit is unclear
The documentation makes statements about OpenZeppelin Contracts being "audited". These statements are inaccurate at best or they may possibly be called disingenuous.

Here are facts that contributors to this project understand, using plain language, and not possibly accessible to mere *users* of the Contracts project:

1. An audit was performed by a reputable firm of the 2.0.0 product version on October 2018.
2. Nobody should use OpenZeppelin Contracts 2.0.0. At a minimum, this is confirmed by the https://docs.soliditylang.org/en/latest/, and I hope it is not necessary for me to provide further detail here.
3. The audit, the OpenZeppelin Contracts project, and the OpenZeppelin company all confirm that as soon as you touch anything, the audit becomes inapplicable.
4. The audits that were performed in the past are only useful to historians and researchers. They have zero attestation of quality or safety on the current state of the project. (Which of course, has been touched since 2.0.0.)

So the following changes should be made.

1. Old audits are dangerous and should be removed from repository. They are historical, and can live permanently in the git history.
2. If for marketing reasons it is desirable to mention these past audits, they must be discussed in the historical context.
3. Never mention "full audit". This implies there is some other kind of audit, which there isn't.

Other notes are outside the scope of this issue:

- Design guideline D6 says "We will perform regular code audits on major releases, and hire security professionals to provide independent review"
- Readme / security says "OpenZeppelin is meant to provide tested and community-audited code"
- The last audit was in 2018 when Bitcoin cost USD 6,000

These three statements seem out of synchronization.

fulldecent:
Fixes #2780

This corrects language around "audit" as mentioned throughout the project.

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [ ] Changelog entry


frangio:
I've removed the misleading comment in the `index.adoc` file (part of the documentation site) in https://github.com/OpenZeppelin/openzeppelin-contracts/commit/56b3b7a4879fee4addcf77f70d519b7c8635309f. I disagree with the rest of the changes in this PR so I have not applied them. Will elaborate in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2780.

fulldecent:
Notes here are fair. The largest issue was addressed and a note about which versions are supported (separate) does separate what was audited versus what is current. Thanks.

PR #2794, opened 2021-07-28
Improve Governor

frangio:
1. Changes interfaces to abstract contracts because it simplifies the mandatory overrides for users.
2. Ignore Against votes when counting votes for quorum to remove the disincentive to vote Against.
3. Add a `cancel` function to the Bravo compatibility module, implementing the same behavior as GovernorBravo.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -137,26 +137,6 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
     }
 
     /**
-     * @dev See {IGovernor-votingDelay}
-     */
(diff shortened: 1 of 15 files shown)


Amxx (dismissed):
All good !

Amxx:
I'm not sure about 
```
it('claims support [skip-on-coverage]', async function () {
    expect(await this.contractUnderTest.supportsInterface(interfaceId)).to.equal(true);
});
```

While I compeltelly get the fact that the "use less then 30k gas" test can fail in coverage mode, this one is just checking the view function. It should not be affected by gas costs

PR #2797, opened 2021-07-29
Avoid safe math operations in BitMap

k06a:
Checking if it would cost less gas

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry


Amxx (approved):

0racl3z (commented):
Thanks!

PR #2798, opened 2021-07-30
Remove all underbar prefix for function parameters

fulldecent:
Underbar prefix in variables names has a reserved meaning: private variables.

This is important because switching private to non-private is a big deal and that will require you to touch every part of code that references.

---

This PR removes all use of underbar prefix from function parameters.

If a some mangling is required (i.e. because that name is already in use elsewhere), prefer to use an underbar suffix. This is currently how ERC721 is implemented here and it looks fine.

---

Related thread to recommend this rule for everybody is at https://github.com/protofire/solhint/issues/300

### DIFF
--- a/contracts/mocks/DummyImplementation.sol
+++ b/contracts/mocks/DummyImplementation.sol
@@ -19,12 +19,12 @@ contract DummyImplementation {
         value = 100;
     }
 
-    function initializeNonPayableWithValue(uint256 _value) public {
-        value = _value;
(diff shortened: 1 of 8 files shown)


Frenzoid (approved):
For those that came from a background in C & C++, underscore prefixes are a convention for naming private variables.

Also in Python, the PEP 8 convention mentions that a variable or method starting with a single underscore is intended for internal use.

frangio:
We don't think this is an issue. Most of the contracts in this PR are also only testing artifacts.

PR #2802, opened 2021-07-31
Update tests for RFC 2606

fulldecent:
RFC 2606 gives us domains we can use as an example. Let's use those rather than recommending some specific unrelated company.


frangio (approved):
Thanks!

PR #2803, opened 2021-07-31
Update tests for RFC 2606 NOT DUPE

fulldecent:
RFC 2606 gives us domains we can use as an example. Let's use those rather than recommending some specific unrelated company. 

PART 2. Sorry 2 commits


frangio:
Cherry picked into https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2803.

PR #2806, opened 2021-08-01
Move fourthTokenId to top

fulldecent:
Some implementations care about which token IDs are used, so moving this stray "fourthTokenID" up next to the other ones so that can be edited together if needed.


frangio (approved):
Thanks!

frangio (approved):

PR #2815, opened 2021-08-07
Extend MerkleProof to support unsorted Merkle tree

k06a:
Let's discuss and if we would decide to have this functionality then would add tests and docs.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/cryptography/MerkleProof.sol
+++ b/contracts/utils/cryptography/MerkleProof.sol
@@ -23,10 +23,10 @@ library MerkleProof {
         bytes32 root,
         bytes32 leaf
     ) internal pure returns (bool) {
-        (bool success, uint256 index) = verifyAndRecoverIndex(proof, root, leaf);
+        (bool success, ) = verifyAndRecoverIndex(proof, root, leaf);
         return success;
     }
-    
+
     /**
      * @dev Returns true if a `leaf` can be proved to be a part of a Merkle tree
      * defined by `root`. For this, a `proof` must be provided, containing
@@ -57,7 +57,7 @@ library MerkleProof {
         // Check if the computed hash (root) is equal to the provided root
         return (computedHash == root, index);
     }
-    
+
     /**
      * @dev Returns true if a `leaf` can be proved to be a part of a Merkle tree
      * defined by `root`. For this, a `proof` must be provided, containing
--- a/contracts/utils/cryptography/MerkleProof.sol
+++ b/contracts/utils/cryptography/MerkleProof.sol
@@ -40,4 +40,36 @@ library MerkleProof {
         // Check if the computed hash (root) is equal to the provided root
         return computedHash == root;
     }
+    
+    /**
+     * @dev Returns true if a `leaf` can be proved to be a part of a Merkle tree
+     * defined by `root`. For this, a `proof` must be provided, containing
+     * sibling hashes on the branch from the leaf to the root of the tree. Each
+     * pair of leaves and each pair of pre-images are assumed to be sorted.
+     * Instead of assuming pair of leaves and pair of pre-images are being sorted,
+     * this method relies on leaf index to recover merkle path directions.
+     */
+    function verify(
+        bytes32[] memory proof,
+        bytes32 root,
+        bytes32 leaf,
+        uint256 leadIndex
+    ) internal pure returns (bool) {
+        bytes32 computedHash = leaf;
+
+        for (uint256 i = 0; i < proof.length; i++) {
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @k06a 

Usually, we would discuss that in an issue before starting a PR. Can you give more details about the use case? Have you faced a situation where sorting the Merkle tree was an issue? While it is not a situation that we might ever face, this technically adds a restriction on the height of the tree.

I'd love to know more about Merkle tree management libraries, which might justify this alternate version.

k06a:
@Amxx index-based merle trees could be useful for bit invalidation. In case of Sorted Merkle Tree it is also possible to recover index by restoring index bit on each operation.

Also what is interesting is to use 128 bits for hashes to 2x reduce calldata, but solidity implementation costs much more somehow for me. Only using Solidity assembly I was able to have 2x gas cheaper implementation.

k06a:
@Amxx please check index recovery technique I just submitted. It could be used for bit invalidation, what Uniswap had in their [Merkle Distributor](https://github.com/Uniswap/merkle-distributor/blob/c3255bfa2b684594ecd562cacd7664b0f18330bf/contracts/MerkleDistributor.sol#L13).

Amxx:
> @Amxx please check index recovery technique I just submitted. It could be useful for the bit invalidation, what Uniswap had in their [Merkle Distributor](https://github.com/Uniswap/merkle-distributor/blob/c3255bfa2b684594ecd562cacd7664b0f18330bf/contracts/MerkleDistributor.sol#L13).

Uniswap uses our Merkle proof library, without the additional feature you are proposing. They are also using a bitmap structure, [which we also provide](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/BitMaps.sol)

I'm not sure what in uniswap would have benefited from unsorted merkle trees

k06a:
@Amxx I meant they we need to include index into nodes, but unique index could be reconstructed during proof validation.my PR now proposes 2 different things: index reconstruction and unsorted merkle proofs validation.

Amxx:
These are 2 different features, that I think should be discussed independently, and have their own PR. This is why we usually recommend filling issues first.

I think index reconstruction is nice indeed. @frangio, any opinion ?

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -23,14 +23,63 @@ library MerkleProof {
...
         for (uint256 i = 0; i < proof.length; i++) {
             bytes32 proofElement = proof[i];
 
             if (computedHash <= proofElement) {
                 // Hash(current computed hash + current element of the proof)
                 computedHash = keccak256(abi.encodePacked(computedHash, proofElement));
+            } else {
+                // Hash(current element of the proof + current computed hash)
+                computedHash = keccak256(abi.encodePacked(proofElement, computedHash));
+                index |= 1 << i;
```
Amxx, line 53: Note: this indexing me logic doesn't guarantee uniqueness of indexes because proofs are processed from the leaf up to the root, and not all branches have equal length.

Example:
```
   Root
   /  \
  X    C
 / \
A   B
```

Path from B to root is :
- right for B to X →  `index |= 1 << 0`
- left from X to root  → no change
in the end, `index = 1`

Path from C to root:
- right from C to root  →  `index |= 1 << 0`
in the end, `index = 1`

This can be fixed by using `index |= 1 << (proof.length -1)` or using the solution proposed in #2841

Amxx: AFAIK, there is no simple way to retrieve the leaf index with the information we have here.

k06a: @Amxx I see, will think of it.

k06a: The only solution I see here is to add one more argument: `treeHeight`

k06a: It could and should be stored in `immutable` paired with `merkleRoot`

Amxx:
This PR includes features requests:
- Being able to verify proofs for "unsorted" trees by providing an index that acts as a "path" to properly rebuild the root
- Being able to rebuild a similar index when verifying proofs in "sorted" trees.

We understand the values in both proposals, but before committing to maintaining a codebase, we want to make sure that the index we use has the right mathematical properties, and is supported by tooling. So far, it is not clear is a specific construct is "standard", and we want to investigate further before we commit to merging any code.

We are closing this PR in favor of an issue where the properties of these objects can be discussed.

PR #2821, opened 2021-08-13
Update ERC20PresetFixedSupply.sol

phyBrackets:
Fixes #???? 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry


mavish302:
change the version of
 pragma solidity ^0.8.0;
to 
pragma solidity >=0.5.0 <0.9.0;

Amxx:
Hello @phyBrackets 

The preset you are talking about depends on many contracts, who also have a pragma set to `^0.8.0`.
While we would love to support more versions of solidity, there are breaking changes which prevent use from targeting older version.

I would recommend that you use solidity 0.8.3 or higher in your project. If you are still on solidity 0.6.x or 0.7.x and are not ready to upgrade, then you should use a 3.x version of openzeppelin contracts (3.4.2 or 3.4.2-solc-0.7)

Issue #2786, opened 2021-07-23
Return the current `_value` when incrementing or decrementing a `Counter`


hexcowboy:
**🧐 Motivation**
Instead of running a call to `counter.increment()` or `counter.decrement()` and then a subsequent call to `counter.current()`, wouldn't it be simpler to just return the count along with the relative call?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Counters.sol

**📝 Details**

Currently the interface is:

```solidity
current(struct Counters.Counter counter) → uint256 internal
increment(struct Counters.Counter counter) internal
decrement(struct Counters.Counter counter) internal
```

but I'm proposing it's changed to

```solidity
current(struct Counters.Counter counter) → uint256 internal
increment(struct Counters.Counter counter) → uint256 internal
decrement(struct Counters.Counter counter) → uint256 internal
```
This could be implemented like so:

```diff
    function increment(Counter storage counter) internal returns(uint256) {
        unchecked {
            counter._value += 1;
        }
+        return counter._value;
    }
```

Amxx:
Hello @hexcowboy and thank you for raising that issue.

This was already discussed as part of #2599. At the time we didn't feel comfortable with the proposition of the PR with added 2 functions with close names. Your approach seems more conservative and is backward compatible.

frangio:
My issue with #2599 was that it is not obvious in the interface whether the return value is pre or post increment. I would oppose this suggestion for the same reason.

tab00:
Intuitively we'd expect the return value to be the post-incremented one, i.e. the end result of calling the `counter.increment()` function.

frangio:
The problem is that often that's not really the useful value. Imagine that your counter is a nonce, you generally want it to start at 0, so you'd want the pre-increment value.

frangio:
I support something _like this_, but `returnAndIncrement` is a confusing name, as "returning" is always the last thing done in a function. 

We should brainstorm other options. What about `getCurrentAndIncrement` together with `incrementAndGet`? Hoping to hear other options as well though.

frangio:
A way to sidestep this issue would be to return both pre and post values from the function:

```
function increment(Counter storage counter) returns (uint previousValue, uint currentValue);
```

Although this sidesteps the issue of naming, it has an issue that return parameters are not really named so the callee needs to know the order. I'd imagine `previous, current` is the intuitive order though.

elawbek:
[...]

PR #2827, opened 2021-08-18
#2786 try to emit the current value via an event

phyBrackets:
Fixes #???? 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


phyBrackets:
#2786 try to return value via an event

PR #2834, opened 2021-08-23
Add an internal _setApprovalForAll function (721 & 1155)

### LINKED ISSUE #2832: Add method to modify _operatorApprovals from derived contracts
**🧐 Motivation**
It is impossible to extend ERC-1155 in a way that the derived contract is accessing the `_operatorApprovals`, because the mapping is `private` and there is no internal method to modify it. Thus, only the ERC1155 contract itself can set or unset approvals. This makes extensions like an ERC1155Permit impossible as they need a way to modify the approval mapping. The proposal is to add an internal function, which can be used in contracts extending ERC1155 to modify `_operatorApprovals`.

**📝 Details**
Contracts extending ERC1155 need a way to modify `_operatorApprovals`. There are two options:
1. Change `_operatorApprovals` visibility from `private` to `internal`
2. Add an `internal` function to modify the mapping.

The proposal is for option two as it aligns well with ERC20, which has a similar setup: `private _allowances`, which can be modfied in derived contracts with `_approve(owner, spender, amount)`. 

The function to be added to ERC1155 is this:

```
function _setApprovalForAll(address owner, address operator, bool approved) 
   internal virtual {
  require(owner != operator, "ERC1155: setting approval status for self");
  require(owner != address(0), "ERC1155: approve from the zero address");
  require(operator != address(0), "ERC1155: approve to the zero address");

  _operatorApprovals[owner][operator] = approved;
  emit ApprovalForAll(owner, operator, approved);
}
```
The main difference to the existing `public setApprovalForAll(address operator, bool approved)` is that it allows to add approvals for other owners than `msg.sender`, which is necessary if we want to create an ERC1155Permit that would set approvals based on signatures.

Amxx:
Fixes #2832

#### PR Checklist
- [ ] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -372,7 +372,11 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
      *
      * Emits a {ApprovalForAll} event.
      */
-    function _setApprovalForAll(address owner, address operator, bool approved) internal virtual {
+    function _setApprovalForAll(
+        address owner,
+        address operator,
+        bool approved
+    ) internal virtual {
         // TODO: add sanity checks?
         _operatorApprovals[owner][operator] = approved;
         emit ApprovalForAll(owner, operator, approved);
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -100,7 +100,6 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
      * @dev See {IERC1155-setApprovalForAll}.
      */
     function setApprovalForAll(address operator, bool approved) public virtual override {
-        require(_msgSender() != operator, "ERC1155: setting approval status for self");
         _setApprovalForAll(_msgSender(), operator, approved);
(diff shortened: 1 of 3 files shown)


contracts/token/ERC1155/ERC1155.sol
```diff
@@ -369,6 +367,21 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
+     * @dev Approve `operator` to operate on all of `owner` tokens
+     *
+     * Emits a {ApprovalForAll} event.
+     */
+    function _setApprovalForAll(
+        address owner,
+        address operator,
+        bool approved
+    ) internal virtual {
+        // TODO: add sanity checks?
```
frangio, line 380: What checks did you have in mind?

Amxx: possibly `require(owner != operator)` ... but I was not really sure

frangio: I agree with adding that require statement. We should move the one in the external function into the internal function.

Note: This PR is also pending a changelog entry.

frangio (dismissed):
LGTM!

PR #2837, opened 2021-08-25
remove redundant impl

armatrix:
IERC20Metadata is already impl IERC20

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Thank you @armatrix. However, we prefer the explicit code we have now.

PR #2838, opened 2021-08-30
Support london with coverage using initialBaseFee = 0

Amxx:
Do not revert to berlin for coverage, instead, just set initialBaseGasFee to 0.

Will be handy if we ever need to use opcodes introduced after berlin in contracts.


frangio (approved):
LGTM!

PR #2841, opened 2021-08-31
Add a MerkleProof.processProof utility function

### LINKED ISSUE #2840: MerkleProof rebuild root.
Sometimes it can be usefull to rebuild the root given a leaf and proof, but not check it matches a known value.

For example, one might have a mapping of valid roots `mapping(bytes32 => bool) private _valid` and may want to run

```
require(_valid[MerkleProof.getRoot(leaf, proof)], "proof doesn't match a valid merkle tree");
```

Amxx:
Fixes #2840

#### PR Checklist

- [x] Tests (covered by existing tests)
- [x] Changelog entry

### DIFF
--- a/contracts/utils/cryptography/MerkleProof.sol
+++ b/contracts/utils/cryptography/MerkleProof.sol
@@ -23,6 +23,23 @@ library MerkleProof {
         bytes32 root,
         bytes32 leaf
     ) internal pure returns (bool) {
+        // Check if the computed hash (root) is equal to the provided root
+        return processProof(proof, leaf) == root;
+    }
+
+    /**
+     * @dev Returns the rebuild hash of node obtained by traversing a Merklee
+     * tree up from `leaf` using `proof`. If `proof` proof is valid, in the send
+     * that it contains the hashes of the sibling on the branch from the leaf to
+     * the root, then this will return the hash of the root of the tree. In
+     * processing the proof we assume the pairs of leafs & pre-images are sorted.
+     *
+     * _Available since v4.4._
+     */
+    function processProof(
+        bytes32[] memory proof,
+        bytes32 leaf
+    ) internal pure returns (bytes32) {
         bytes32 computedHash = leaf;
 
         for (uint256 i = 0; i < proof.length; i++) {
@@ -37,7 +54,6 @@ library MerkleProof {
             }
         }
 
-        // Check if the computed hash (root) is equal to the provided root
-        return computedHash == root;
+        return computedHash;
     }
 }


contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -23,21 +23,36 @@ library MerkleProof {
...
+        // Check if the computed hash (root) is equal to the provided root
+        (bytes32 computedHash, ) = processProof(proof, leaf);
+        return computedHash == root;
+    }
 
+    /**
+     * @dev Returns the rebuild hash of node obtained by traversing a Merklee
+     * tree up from `leaf` using `proof`. If `proof` proof is valid, in the send
+     * that it contains the hashes of the sibling on the branch from the leaf to
+     * the root, then this will return the hash of the root of the tree. In
```
frangio, line 35:
```suggestion
     * @dev Returns the rebuilt hash of nodes obtained by traversing a Merklee
     * tree up from `leaf` using `proof`. If `proof` is valid, in the sense
     * that it contains the hashes of the siblings on the branch from the leaf to
     * the root, then this will return the hash of the root of the tree. In
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -23,21 +23,36 @@ library MerkleProof {
...
+     *
+     * _Available since v4.4._
+     */
+    function processProof(bytes32[] memory proof, bytes32 leaf) internal pure returns (bytes32, uint256) {
+        bytes32 computedHash = leaf;
+        uint256 index = 0;
         for (uint256 i = 0; i < proof.length; i++) {
             bytes32 proofElement = proof[i];
-
+            index <<= 1;
```
frangio, line 45: Can we do `index * 2` for overflow protection?

```suggestion
            index *= 2;
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -23,21 +23,36 @@ library MerkleProof {
...
         bytes32 leaf
     ) internal pure returns (bool) {
-        bytes32 computedHash = leaf;
+        // Check if the computed hash (root) is equal to the provided root
+        (bytes32 computedHash, ) = processProof(proof, leaf);
+        return computedHash == root;
+    }
 
+    /**
+     * @dev Returns the rebuild hash of node obtained by traversing a Merklee
```
frangio, line 32: Can you document the `index` value that is returned? I don't really understand it either. :smiley:

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+ * `MerkleProof`: add a processProof function that returns the rebuilt root hash given a leaf and a proof. ([#2841](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2841))
```
frangio, line 5:
```suggestion
 * `MerkleProof`: add a `processProof` function that returns the rebuilt root hash given a leaf and a proof. ([#2841](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2841))
```

contracts/utils/cryptography/MerkleProof.sol
Amxx: I'm mostly sure (need a proof) of uniqueness, but I have counter-examples showing non-ordering and non-sequentiality (mostly because of the "branches" being has order rather than branch direction)

Amxx: It can be useful to mark a leaf-proof combo as consumed in case a leaf hash is present multiple times in the tree.

PR #2849, opened 2021-09-06
Allow Governor + CompoundTimelock to manage native tokens (eth) in and out of the timelock contract.

Amxx:
[In GovernorBravo (and Alpha), the execute function must send the value attached to the proposal to the timelock](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/GovernorBravoDelegate.sol#L146)

This means two things:

- The executor must give away all the Eth needed to run the proposal
- Any Eth present on the timelock cannot be moved away from it (such proposal would require the executor to provide an equal amount, resulting in a null result from the timelock point of view).

This means that governor + compound timelock cannot be used to manage big funds in Eth.

This PR moves away from compound behavior to enable such usecases.

Note: this issue does NOT affect OZ Governor + TimelockController combo

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -56,6 +56,13 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
     }
 
     /**
+     * @dev Function to receive ETH that will be handled by the governor (disabled if executor is a third party contract)
+     */
+    receive() external payable virtual {
+        require(_executor() == address(this));
+    }
+
+    /**
      * @dev See {IERC165-supportsInterface}.
      */
     function supportsInterface(bytes4 interfaceId) public view virtual override(IERC165, ERC165) returns (bool) {
--- a/contracts/governance/extensions/GovernorTimelockCompound.sol
+++ b/contracts/governance/extensions/GovernorTimelockCompound.sol
@@ -177,8 +177,9 @@ abstract contract GovernorTimelockCompound is IGovernorTimelock, Governor {
     ) internal virtual override {
         uint256 eta = proposalEta(proposalId);
         require(eta > 0, "GovernorTimelockCompound: proposal not yet queued");
+        Address.sendValue(payable(_timelock), msg.value);
         for (uint256 i = 0; i < targets.length; ++i) {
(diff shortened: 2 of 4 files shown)


tarrencev:
Thanks for helping look into this. I've confirmed that our [integration test](https://github.com/dopedao/contracts/pull/13/files#diff-8e7f52af79e9b36979f80291eebd5a35a8862618dd237aec030597830c38b41cR88) passes with this change.

frangio:
I don't think this change is ideal because it is different from the behavior of GovernorBravo in an important way: ETH sent to the Governor will be locked in it. I think this is a more general problem we should tackle for ERC20 sent to the Governor as well, although that is not a new problem with respect to GovernorBravo. I would not merge this change as is without a way to retrieve those funds.

An alternative change is to send `Math.min(this.balance, values[i])`. I think this results in a way to execute proposals that can use both the funds in the Governor and in the Timelock.

Amxx:
What about the difference of behavior between our governors depending on the timelock module used? To me they are more problematic.

Also, the governor should not have a receive function if there is a timelock module (or if there is one it should revert)

contracts/governance/Governor.sol
```diff
@@ -55,6 +55,13 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
         _name = name_;
     }
 
+    /**
+     * @dev Function to receive ETH that will be handled by the governor (disabled if executor is a third party contract)
```
frangio, line 59:
```suggestion
     * @dev Function to receive ETH that will be handled by the governor. Should be disabled if executor is a third party contract, such as a timelock.
```

contracts/governance/Governor.sol
```diff
@@ -55,6 +55,13 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
...
+     */
+    receive() external payable virtual {
```
frangio, line 61: But we are missing the overrides in the timelock modules, right?

Amxx: Not needed, the require takes care of that... hence the @dev message

frangio (commented):
Looks good to me!

frangio (approved):
Looks good to me!

PR #2851, opened 2021-09-09
Add a 'verifyWithIndex' function

Arachnid:
I was going to open an issue to discuss this, but figured it was just as quick to give an example PR for feedback.

It's trivial for the merkle proof verification function to compute and return an index when it verifies the proof. This is particularly useful for the common case of wanting to track when airdrops have been claimed using a bitmap; the returned index value can be used to specify the bit in the bitmap to set.

### DIFF
--- a/contracts/utils/cryptography/MerkleProof.sol
+++ b/contracts/utils/cryptography/MerkleProof.sol
@@ -14,18 +14,21 @@ pragma solidity ^0.8.0;
 library MerkleProof {
     /**
      * @dev Returns true if a `leaf` can be proved to be a part of a Merkle tree
-     * defined by `root`. For this, a `proof` must be provided, containing
-     * sibling hashes on the branch from the leaf to the root of the tree. Each
-     * pair of leaves and each pair of pre-images are assumed to be sorted.
+     * defined by `root`, along with the element's index in the tree.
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @Arachnid, and thank you for your PR

The changes proposed add a `verifyWithIndex(bytes32[],bytes32,bytes32) returns (bool,uint256)`, which is nice, but renames the existing function. I think it is essential to keep the existing signature `verify(bytes32[],bytes32,bytes32) returns (bool)` for backward compatibility.

Note that there is already a PR affecting `MerkleProof.sol` #2841. If you are ok with it, I think we should merge the two.

Amxx:
Also, I don't really think your indexes are the ones I would expect

- The first leaf from the left is 0,
- The second one is 2*(n-1) with n the depth of the tree

Arachnid:
#2841 looks like it's pretty much a superset of this, so I'm happy to close it in favor of that!

Arachnid:
Personally I would've had it revert on overflow, and document that it only supports trees that are at most 256 levels deep, and that indexes will be unique but not necessarily sequential. I think this is still a useful primitive under those contraints.

The proposed workaround of supplying the index as part of the leaf is quite inelegant, as it requires providing redundant information.

PR #2852, opened 2021-09-09
Fix EIP712 for delegatecalls

k06a:
Some tokens do delegate their implementations, this fix would potential double spends.

Let's discuss.

### DIFF
--- a/contracts/utils/cryptography/draft-EIP712.sol
+++ b/contracts/utils/cryptography/draft-EIP712.sol
@@ -29,6 +29,7 @@ abstract contract EIP712 {
     // invalidate the cached domain separator if the chain id changes.
     bytes32 private immutable _CACHED_DOMAIN_SEPARATOR;
     uint256 private immutable _CACHED_CHAIN_ID;
+    address private immutable _CACHED_THIS;
 
     bytes32 private immutable _HASHED_NAME;
     bytes32 private immutable _HASHED_VERSION;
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @k06a and thank you for your PR.

I think you found out something serious. We'll get back to your asap

frangio:
@k06a Have you seen tokens that do delegation outside the context of upgradeability?

It is nevera good idea to use this package with upgradeable contracts and I don't think we should include this fix because it would be validating that use.

k06a:
USDC token have permit and is proxy contract.

k06a:
I would prefer to have this check as extra security to avoid misusage. Misusage could lead to a double spend.

frangio:
After some thought I think we should go ahead with this. What convinced me is that we have a discrepancy in the "this" value that is used in each branch of the if statement (cached vs recreated).

Note that this slightly changes the behavior and could be considered backwards incompatible, but the cases that would be affected would have had the domain separator change anyway.

As always, note that upgradeable contracts should be using [the Upgradeable version of this contract](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/release-v4.3/contracts/utils/cryptography/draft-EIP712Upgradeable.sol).

---

A changelog entry is needed before merging.

Amxx (dismissed):

frangio:
There is something wrong with Codecov.

PR #2854, opened 2021-09-09
Update typehash string formatting

k06a:
Hope you'll like it, because this required to modify https://github.com/prettier-solidity/prettier-plugin-solidity/pull/575

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -26,7 +26,13 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
     using SafeCast for uint256;
     using Timers for Timers.BlockNumber;
 
-    bytes32 public constant BALLOT_TYPEHASH = keccak256("Ballot(uint256 proposalId,uint8 support)");
+    bytes32 public constant BALLOT_TYPEHASH =
+        keccak256(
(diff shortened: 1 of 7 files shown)


Amxx:
Hello @k06a and thank you for yet another great proposal.

I didn't know that `keccak` includes an implicit `abi.encodePacked`/`bytes.concat`.

k06a:
@Amxx this is called [Multi-line string literal](https://docs.soliditylang.org/en/v0.8.7/grammar.html?highlight=literal#a4.SolidityParser.stringLiteral). These three strings are identical in terms of compiled source code.
```
"Ballot(uint256 proposalId,uint8 support)"
```
```
"Ballot(" "uint256 proposalId," "uint8 support" ")"
```
```
"Ballot("
"uint256 proposalId,"
"uint8 support"
")"
```

k06a:
This PR is not refactoring but reformatting

k06a:
So WDYT?

frangio:
Hey so I think we're actually not convinced that this formatting looks better. Personally I find the one-line version easier to read, and I think it could come in handy to be able to copy-paste it into JavaScript where the multiline string wouldn't be accepted. Sorry!

PR #2855, opened 2021-09-10
Add notice about location of funds in Governor+Timelock setup

frangio:


Amxx (commented):
I think Governor is missing a receive function, so cannot easily receive ether.
(we may need to add that to the compound timelock module ... but I'd rather "fix" the module)

Amxx (approved):

PR #2856, opened 2021-09-13
Add MultiParty contract standard with implementations

saurabhsanthosh:
**Add Multiparty contract standard with implementations.**

This is a new standard which allows developers to create contracts which can be managed by multiple parties in a democratic and secure way.

Discussions :  

- [Ethereum Magicians Forum](https://ethereum-magicians.org/t/contract-standard-for-co-owning-assets/6877)
- [Ethereum EIP Git](https://github.com/ethereum/EIPs/pull/3742)

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/multiparty/IMultiParty.sol
+++ b/contracts/multiparty/IMultiParty.sol
@@ -1,0 +1,113 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Interface of the Multiparty contract standard.
+ * This contract standard can be used if a smart contract has to be managed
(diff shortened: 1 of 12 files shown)


Amxx:
Hello @saurabhsanthosh 

There are many implementations of multisig contracts through the ecosystem, each with their own supporters and tooling. I don't feel it is our role, at OpenZeppelin to endorse any particular interface, and maintain a compatible implementation.

We could possibly rethink that if an EIP standardizes a widely accepted interface. ERC725v1 (now known as ERC734) tried that. ERC1077 also tried adding meta-tx on top of it. Both were short-lived.

PR #2860, opened 2021-09-14
Create ERC721Full.sol

Aalian2241:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Hello @Aalian2241 

Thank you for your proposal, but the fact is, we are currently moving away from presets and grouping contracts like the one you propose. We believe that using and advertizing [the wizard](https://wizard.openzeppelin.com/) is a way better option.

PR #2863, opened 2021-09-15
Show friendly tips when the relayer has not sent enough gas for the c…

TheStarBoys:
Show friendly tips in MinimalForwarder.sol

Otherwise, when the relayer has not sent enough gas for the call, just show `Error: Returned error: VM Exception while processing transaction: revert`. It's not a friendly tips and hard to find out the point that occurs `revert`.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/metatx/MinimalForwarder.sol
+++ b/contracts/metatx/MinimalForwarder.sol
@@ -51,7 +51,7 @@ contract MinimalForwarder is EIP712 {
         );
         // Validate that the relayer has sent enough gas for the call.
         // See https://ronan.eth.link/blog/ethereum-gas-dangers/
-        assert(gasleft() > req.gas / 63);
+        require(gasleft() > req.gas / 63, "MinimalForwarder: the relayer has not sent enough gas for the call");
 
         return (success, returndata);
     }


Amxx:
Hello @TheStarBoys 

While you proposal could help in some cases, it would break the relaying system. Relayed calls can be nested, and it is important that if one fails because of gas requirements this behavior if bubbled up to the other calls.

This is discussed in [this post](https://ronan.eth.link/blog/ethereum-gas-dangers/), that I encourage you to read and share

Amxx:
Edit: [direct ipfs link for when ipns resolution fails](https://ipfs.io/ipfs/QmbbYTGTeot9ic4hVrsvnvVuHw4b5P7F5SeMSNX9TYPGjY/blog/ethereum-gas-dangers/)

TheStarBoys:
> Hello @TheStarBoys
> 
> While your proposal could help in some cases, it would break the relaying system. Relayed calls can be nested, and it is important that if one fails because of gas requirements this behavior if bubbled up to the other calls. A revert would not achieve that like an assert does.
> 
> This is discussed in [this post](https://ronan.eth.link/blog/ethereum-gas-dangers/), that I encourage you to read and share

Thanks for your comments and recommendations. It's helpful for me.

PR #2865, opened 2021-09-16
#890: Add ECDSA#toEthSignedMessage for bytes type

### LINKED ISSUE #890: Add ECRecovery#ethSignedMessage for bytes type
## 🎉 Description

The `ECRecovery.sol` function should probably have a function like

```
function ethSignedMessage(bytes s) pure internal returns (bytes32) {
  return keccak256(
    "\x19Ethereum Signed Message:\n",
    uintToBytes(s.length),
    s);
}
```

Where `uintToBytes` is implemented from an ideally MIT-licensed project with appropriate attribution.

- [x] 📈 This is a feature request.

td-bn:
Fixes #890 
Issue [link](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/890)

The PR adds a function toEthSignedMessage(bytes) to the ECDSA contract, that works for any length of a bytes array.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/mocks/ECDSAMock.sol
+++ b/contracts/mocks/ECDSAMock.sol
@@ -6,6 +6,7 @@ import "../utils/cryptography/ECDSA.sol";
 
 contract ECDSAMock {
     using ECDSA for bytes32;
+    using ECDSA for bytes;
 
     function recover(bytes32 hash, bytes memory signature) public pure returns (address) {
         return hash.recover(signature);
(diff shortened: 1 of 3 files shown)


Amxx:
Hello @td-bn , and thank you for this PR!

Do you think it would be possible to add a test with a message that is not 32 bytes long?

frangio:
Please add a test in the section that says `using web3.eth.sign`, so we can confirm that this function is doing the same that `eth_sign` is doing under the hood.

Also, the function name of the new function should be `toEthSignedMessageHash`, not `toEthSignedMessage`. The "hash" part of the name refers to the output, not to the argument. We want to overload the current function so that it also supports `bytes` arguments.

Issue #987, opened 2018-06-08
Feedback From Open Zeppelin About ERC-173 Contract Ownership Standard


mudgen:
A new proposal/draft for the contract interface for owning and transferring ownership of contracts was recently written and merged with the ethereum eip project. The draft is here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-173.md

What is Open Zepplin's response/feedback about this?

Also, can Open Zepplin help get the word out about contract ownership becoming standardized? Get the word out about this EIP draft so it can get the feedback it needs for standardization.

frangio:
Hi there, @mudgen.

An important piece of feedback is that the semantics of the functions and events don't seem to be specified at all.

mudgen:
@frangio thank you for your comment.  I noticed that the @notice for the `transferOwnership()` function was wrong so I fixed that and will be in the next pull request. 

The notice for the  `transferOwnership()` function now says, "Set the address of the new owner of the contract".  The notice for the `owner()` function says, "Get the address of the owner " Are more semantics needed? If so can you give me more idea of the semantics that should be included?

Implementations of functions `owner()` and `transferOwnership()` are unspecified because they can be implemented how people want to implement them. But the standard does link to Open Zepplin's implementation of the standard as an example.

frangio:
Closing as discussion should be happening in https://github.com/ethereum/EIPs/issues/173.

PR #2874, opened 2021-09-21
Support ERC173

axic:
Perhaps it is a bit premature to open this, but it felt easier to discuss it this way and not in an issue. The last discussion took place on #987. Opened a question I have here: https://github.com/ethereum/EIPs/issues/173#issuecomment-924097327

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -2,6 +2,7 @@
 
 pragma solidity ^0.8.0;
 
+import "./IERC173.sol";
 import "../utils/Context.sol";
 
 /**
@@ -16,10 +17,8 @@ import "../utils/Context.sol";
  * `onlyOwner`, which can be applied to your functions to restrict their use to
  * the owner.
  */
-abstract contract Ownable is Context {
+abstract contract Ownable is Context, IERC173 {
     address private _owner;
-
-    event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
 
     /**
      * @dev Initializes the contract setting the deployer as the initial owner.
@@ -31,7 +30,7 @@ abstract contract Ownable is Context {
     /**
      * @dev Returns the address of the current owner.
      */
-    function owner() public view virtual returns (address) {
+    function owner() public view virtual override returns (address) {
         return _owner;
     }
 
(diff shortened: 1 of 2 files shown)


contracts/access/Ownable.sol
```diff
@@ -16,11 +17,9 @@ import "../utils/Context.sol";
  * `onlyOwner`, which can be applied to your functions to restrict their use to
  * the owner.
  */
-abstract contract Ownable is Context {
+abstract contract Ownable is Context, IERC173 {
```
axic, line 20: Also the whole point of this would be to implement `supportsInterface` here, but not sure how welcome is it to add it (similar to how it is done in `token/ERC721/ERC721.sol` for example).

Amxx:
Hello @axic 

This is something I considered during the recent introduction of the `interfaces` folder. This was ruled out because ERC173 is not final, hence [its message asking about the status](https://github.com/ethereum/EIPs/issues/173#issuecomment-890918007).

When the ERC moves to final we will do this change (but with the `IERC173.sol` file in the `interfaces` folder)

frangio:
I'm skeptical about the usefulness of ERC165 in this context. Will reconsider the PR if ERC173 moves forward.

niccolopetti:
On July 9 2022 ERC173 has finally moved to final status (see commit https://github.com/ethereum/EIPs/commit/110aff008b86bcc905998df9cb9bfd5a7efa666a )

I think it's time to reopen discussion about this, and imho make oz adhere to the new standard

Amxx:
There are two things:

- ERC173 doesn't include the renounceOwnership function, and instead accepts transferring to 0. It be a breaking change on our end to comply with that.
- Adding ERC165 would cause a lot of inheritance conflicts around the `supportInterface` function. This could be painful to devs.

Therefore, I would suggest that if we include that in our codebase, it is seen as a breaking change and is scheduled for v5.0.0

KrakenLynch:
This needs to get integrated in open-zeppelin as now, opensea will only support royalty for contracts that has implemented EIP-173

frangio:
@KrakenLynch Can you share a reference? Where did you see that?

Amxx:
should we keep `renounceOwnership` for backward compatibility (in addition to transfer to 0 being authorized) ?

PR #2884, opened 2021-09-25
Add Base64 library to utils

### LINKED ISSUE #2859: Feature: Include Base64 library in utils
**🧐 Motivation**
Recently, with the NFT rising and on-chain metadata protocols such as Uniswap and Loot (including all of its derivatives), Base64 encoder has become a widely used library, but there's no further documentation about it, and everybody seems to be copying the usage from the base Uniswap and Loot contracts:

- Uniswap: https://github.com/Uniswap/uniswap-v3-periphery/blob/main/contracts/libraries/NFTDescriptor.sol#L66
- Loot: https://etherscan.io/address/0xff9c1b15b16263c61d017ee9f65c50e4ae0113d7#code#L1609

Even though it seems to be working correctly, it might be needed to include in Open Zeppelin as one of the available libraries, so there is an official reference for it.

Currently, there is this [Base64 outdated experiment](https://github.com/OpenZeppelin/solidity-jwt/blob/master/contracts/Base64.sol), and [this is the implementation](https://github.com/Brechtpd/base64#readme) that most of the projects are using up to this date

**📝 Details**

- Include @Brechtpd implementation into `contracts/utils` folder
- Add documentation and tests for them

Edit: Base64 data prepended with its media type [is considered a valid URI](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs). This library will enable NFTs on-chain `tokenURIs`

I can help if there's somebody who can guide me through the process

ernestognw:
Fixes #2859 

#### Description

As previously discussed on #2859, seems like Base64 has been used as a way to construct Safe-URL `tokenURIs` for `ERC721`. This PR aims to provide a standard library to perform those operations through Open Zeppelin utils.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

##### Notes
I haven't been able to run docs locally to start creating the documentation for this, and according to the checklist, I don't know how to update the Changelog entry. Would be helpful to have some guide from one of the team members

### DIFF
--- a/contracts/mocks/Base64Mock.sol
+++ b/contracts/mocks/Base64Mock.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "../utils/Base64.sol";
+
+contract Base64Mock {
+    function encode(bytes memory value) external pure returns (string memory) {
+        return Base64.encode(value);
+    }
+}
--- a/contracts/utils/Base64.sol
+++ b/contracts/utils/Base64.sol
@@ -1,0 +1,102 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Provides a set of functions to operate with Base64 strings.
+ */
+library Base64 {
+  /**
+   * @dev Base64 Encoding/Decoding Table
+   */
+    string internal constant _TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+    /**
+     * @dev Converts a `bytes` to its Bytes64 `string` representation.
+     * Inspired by Brecht Devos (Brechtpd) implementation - MIT licence
+     * https://github.com/Brechtpd/base64/blob/e78d9fd951e7b0977ddca77d92dc85183770daf4/base64.sol
+     */
+    function encode(bytes memory data) internal pure returns (string memory) {
+        if (data.length == 0) return "";
+
+        // Loads the table into memory
+        string memory table = _TABLE;
+
+        // Encoding takes 3 bytes chunks of binary data from `bytes` data parameter
+        // and split into 4 numbers of 6 bits.
+        // The final Base64 length should be `bytes` data length multiplied by 4/3 rounded up
+        // - `data.length + 2`  -> Round up 
+        // - `/ 3`              -> Number of 3-bytes chunks
+        // - `4 *`              -> 4 characters for each chunk
+        uint256 encodedLen = 4 * ((data.length + 2) / 3);
+
+        // Add some extra buffer at the end required for the writing
+        string memory result = new string(encodedLen + 32);
+
+        assembly {
+            // Store the actual result length in memory
+            mstore(result, encodedLen)
+
+            // Prepare the lookup table
+            let tablePtr := add(table, 1)
+
+            // Prepare input pointer
+            let dataPtr := data
+            let endPtr := add(dataPtr, mload(data))
+
+            // Prepare result pointer, jump over length
+            let resultPtr := add(result, 32)
+
+            // Run over the input, 3 bytes at a time
+            for { } lt(dataPtr, endPtr) { } {
+                // Advance 3 bytes
+                dataPtr := add(dataPtr, 3)
+                let input := mload(dataPtr)
+
+                // To write each character, shift the 3 bytes (18 bits) chunk
+                // 4 times in blocks of 6 bits for each character (18, 12, 6, 0)
+                // and apply logical AND with 0x3F which is the number of
+                // the previous character in the ASCII table prior to the Base64 Table
+                // The result is then added to the table to get the character to write,
+                // and finally write it in the result pointer but with a left shift
+                // of 256 (1 byte) - 8 (1 ASCII char) = 248 bits
+
+                mstore(
+                  resultPtr, 
+                  shl(248, mload(add(tablePtr, and(shr(18, input), 0x3F))))
+                )
+                resultPtr := add(resultPtr, 1) // Advance
+                
+                mstore(
+                  resultPtr, 
+                  shl(248, mload(add(tablePtr, and(shr(12, input), 0x3F))))
+                )
+                resultPtr := add(resultPtr, 1) // Advance
+                
+                mstore(
+                  resultPtr,
+                  shl(248, mload(add(tablePtr, and(shr(6, input), 0x3F))))
+                )
+                resultPtr := add(resultPtr, 1) // Advance
+                
+                mstore(
+                  resultPtr,
+                  shl(248, mload(add(tablePtr, and(input, 0x3F))))
+                )
+                resultPtr := add(resultPtr, 1) // Advance
+            }
+
+            // When data `bytes` is not exactly 3 bytes long
+            // it is padded with `=` characters at the beginning
+            switch mod(mload(data), 3)
+            case 1 {
+                mstore(sub(resultPtr, 2), shl(240, 0x3d3d))
+            }
+            case 2 {
+                mstore(sub(resultPtr, 1), shl(248, 0x3d))
+            }
+        }
+
+        return result;
+    }
+}
--- a/contracts/utils/Base64.sol
+++ b/contracts/utils/Base64.sol
@@ -87,7 +87,7 @@ library Base64 {
             }
 
             // When data `bytes` is not exactly 3 bytes long
-            // it is padded with `=` characters at the beginning
+            // it is padded with `=` characters at the end
             switch mod(mload(data), 3)
             case 1 {
                 mstore(sub(resultPtr, 2), shl(240, 0x3d3d))
--- a/contracts/utils/Base64.sol
+++ b/contracts/utils/Base64.sol
@@ -6,17 +6,19 @@ pragma solidity ^0.8.0;
  * @dev Provides a set of functions to operate with Base64 strings.
  */
 library Base64 {
-  /**
-   * @dev Base64 Encoding/Decoding Table
-   */
+    /**
+     * @dev Base64 Encoding/Decoding Table
+     */
     string internal constant _TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
     /**
      * @dev Converts a `bytes` to its Bytes64 `string` representation.
-     * Inspired by Brecht Devos (Brechtpd) implementation - MIT licence
-     * https://github.com/Brechtpd/base64/blob/e78d9fd951e7b0977ddca77d92dc85183770daf4/base64.sol
      */
     function encode(bytes memory data) internal pure returns (string memory) {
+        /**
+         * Inspired by Brecht Devos (Brechtpd) implementation - MIT licence
+         * https://github.com/Brechtpd/base64/blob/e78d9fd951e7b0977ddca77d92dc85183770daf4/base64.sol
+         */
         if (data.length == 0) return "";
 
         // Loads the table into memory
@@ -25,7 +27,7 @@ library Base64 {
         // Encoding takes 3 bytes chunks of binary data from `bytes` data parameter
         // and split into 4 numbers of 6 bits.
         // The final Base64 length should be `bytes` data length multiplied by 4/3 rounded up
-        // - `data.length + 2`  -> Round up 
+        // - `data.length + 2`  -> Round up
         // - `/ 3`              -> Number of 3-bytes chunks
         // - `4 *`              -> 4 characters for each chunk
         uint256 encodedLen = 4 * ((data.length + 2) / 3);
@@ -48,7 +50,11 @@ library Base64 {
             let resultPtr := add(result, 32)
 
             // Run over the input, 3 bytes at a time
-            for { } lt(dataPtr, endPtr) { } {
+            for {
+
+            } lt(dataPtr, endPtr) {
+
+            } {
                 // Advance 3 bytes
                 dataPtr := add(dataPtr, 3)
                 let input := mload(dataPtr)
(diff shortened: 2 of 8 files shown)


contracts/mocks/Base64Mock.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "../utils/Base64.sol";
+
+contract Base64Mock {
```
vince0656, line 7: just my 2 cents but as an alternative, you can have the following line at the top of the contract:
`using Base64 for bytes;`

and then in your encode:
`value.encode();`

Amxx:
Hello @ernestognw, thank you very much for this PR, and sorry we didn't answer it sooner.

This is really something interesting that we would love to merge. I addition to actually providing code and tests, you have done a great job documenting the changes.

One of the reasons for the late answer, is we have been reluctant to dive into the assembly code. The tests are showing it works, but we are still not really comfortable with it. Our opinion is that this code would not be paid for, and just be used as part of "view" operation. Thus it might not be necessary to make it as efficient as possible, and maybe we would prefer having more easily understandable code.

Have you tried writing it in pure solidity, without any assembly ?

cameel:
> Have you tried writing it in pure solidity, without any assembly ?

I just stumbled upon this PR and decided to give it a try myself to see how the new optimizer in the compiler would fare with fare against that hand-optimized code and see what we could improve.

Here's a version of the library in pure Solidity (**WARNING: not tested at all, might be buggy**):

```solidity
library Base64 {
    bytes internal constant _TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";

    function encode(bytes memory data) internal pure returns (string memory) {
        if (data.length == 0) return "";

        bytes memory table = _TABLE;
        uint256 encodedLen = 4 * ((data.length + 2) / 3);

        bytes memory result = new bytes(encodedLen);

        unchecked {
            uint resultPtr = 32;

            for (uint dataPtr = 0; dataPtr < data.length; dataPtr += 3) {
                uint24 input =
                    uint24(uint8(data[dataPtr    ])) << 16 +
                    uint24(uint8(data[dataPtr + 1])) <<  8 +
                    uint24(uint8(data[dataPtr + 2]));

                result[resultPtr++] = table[((input >> 18) & 0x3F)];
                result[resultPtr++] = table[((input >> 12) & 0x3F)];
                result[resultPtr++] = table[((input >>  6) & 0x3F)];
                result[resultPtr++] = table[( input        & 0x3F)];
            }

            if (data.length % 3 == 2) {
                result[resultPtr - 2] = 0x3d;
                result[resultPtr - 1] = 0x3d;
            }
            else if (data.length % 3 == 1)
                result[resultPtr - 1] = 0x3d;
        }

        return string(result);
    }
}
```
You can see the resulting assembly at the end of this post. Unfortunately looks like there are still a few things that the optimizer could do but does not:
- **Inlining**. The most visible difference is that there's a bunch of extra functions compared to the hand-optimized assembly. The inliner is currently a little too conservative due to the risk of running into "stack too deep" errors (inlining usually results in more local variables). That's actually something I'm working on right now - we have recently improved the EVM code generation from Yul and now we're looking into relaxing these rules a little.
    - In addition to removing jumps if all these functions got inlined, it would allow many other optimizations to be applied more broadly. For example there's some unnecessary shifting and masking due to conversions and cleaning of unused bits that could be removed if the code was not in separate functions.
- **Array length checks**. Looks like they were not optimized out in this case.
- **Redundant `mstore8` instructions**. In the final `switch`, the hand-optimized assembly is a tiny bit more efficient because it can write two bytes using a single `mstore` instruction while the compiler does two `mstore8` instead. Note however that the inline assembly version pays for this by wasting 32 bytes of memory in the `result` array (the Solidity version does not).

For now as you can see, the hand-optimized version should be cheaper. I think I'll add this code as a test case for the optimizer though so it should improve in future versions.

### Pure Solidity version passed through Yul optimizer
This is the result from running `solc --optimize --ir-optimized --debug-info none <input file>` on the library and the `Base64Mock` contract from the PR. I included only the functions that are different or not present in the inline assembly version, the whole code was a bit longer.
```solidity
function memory_array_index_access_bytes(baseRef, index) -> addr
{
    if iszero(lt(index, mload(baseRef)))
    {
        mstore(0, shl(224, 0x4e487b71))
        mstore(4, 0x32)
        revert(0, 0x24)
    }
    addr := add(add(baseRef, index), 32)
}
function read_from_memoryt_bytes1(ptr) -> returnValue
{
    returnValue := and(mload(ptr), shl(248, 255))
}
function convert_bytes1_to_uint8(value) -> converted
{ converted := shr(248, value) }
function convert_uint8_to_uint24(value) -> converted
{ converted := and(value, 0xff) }
function wrapping_add_uint24_1076(y) -> sum
{
    sum := and(add(16, y), 0xffffff)
}
function wrapping_add_uint24(y) -> sum
{
    sum := and(add(8, y), 0xffffff)
}
function shift_left_uint24_uint24(value, bits) -> result
{
    let _1 := 0xffffff
    result := and(shl(and(bits, _1), and(value, _1)), _1)
}
function fun_encode(var_data_mpos) -> var__mpos
{
    if iszero(mload(var_data_mpos))
    {
        var__mpos := allocate_memory_array_string()
        leave
    }
    let expr_35_mpos := copy_literal_to_memory_84d8a590de33e00cbdc16e1f28c3506f5ec15c599fab9a6a4bcd575cc2f110ce()
    let expr_mpos := allocate_and_zero_memory_array_bytes(checked_mul_uint256(checked_div_uint256(checked_add_uint256(mload(var_data_mpos)))))
    let var_resultPtr := 0x20
    let var_dataPtr := 0x00
    for { }
    0x01
    {
        var_dataPtr := add(var_dataPtr, 0x03)
    }
    {
        if iszero(lt(var_dataPtr, mload(var_data_mpos))) { break }
        let expr := convert_uint8_to_uint24(convert_bytes1_to_uint8(read_from_memoryt_bytes1(memory_array_index_access_bytes(var_data_mpos, var_dataPtr))))
        let _1 := 0x01
        let expr_1 := shift_left_uint24_uint24(expr, wrapping_add_uint24_1076(convert_uint8_to_uint24(convert_bytes1_to_uint8(read_from_memoryt_bytes1(memory_array_index_access_bytes(var_data_mpos, add(var_dataPtr, _1)))))))
        let _2 := 0x02
        let expr_2 := shift_left_uint24_uint24(expr_1, wrapping_add_uint24(convert_uint8_to_uint24(convert_bytes1_to_uint8(read_from_memoryt_bytes1(memory_array_index_access_bytes(var_data_mpos, add(var_dataPtr, _2)))))))
        let _3 := 0x3f
        mstore8(memory_array_index_access_bytes(expr_mpos, var_resultPtr), byte(0x00, read_from_memoryt_bytes1(memory_array_index_access_bytes(expr_35_mpos, and(shr(18, expr_2), _3)))))
        mstore8(memory_array_index_access_bytes(expr_mpos, add(var_resultPtr, _1)), byte(0x00, read_from_memoryt_bytes1(memory_array_index_access_bytes(expr_35_mpos, and(shr(12, expr_2), _3)))))
        mstore8(memory_array_index_access_bytes(expr_mpos, add(var_resultPtr, _2)), byte(0x00, read_from_memoryt_bytes1(memory_array_index_access_bytes(expr_35_mpos, and(shr(6, expr_2), _3)))))
        let _4 := read_from_memoryt_bytes1(memory_array_index_access_bytes(expr_35_mpos, and(expr_2, _3)))
        let _5 := add(var_resultPtr, 0x03)
        var_resultPtr := add(var_resultPtr, 0x04)
        mstore8(memory_array_index_access_bytes(expr_mpos, _5), byte(0x00, _4))
    }
    let _6 := mod(mload(var_data_mpos), 0x03)
    switch eq(_6, 0x02)
    case 0 {
        if eq(_6, 0x01)
        {
            mstore8(memory_array_index_access_bytes(expr_mpos, add(var_resultPtr, not(0))), 0x3d)
        }
    }
    default {
        mstore8(memory_array_index_access_bytes(expr_mpos, add(var_resultPtr, not(1))), 0x3d)
        mstore8(memory_array_index_access_bytes(expr_mpos, add(var_resultPtr, not(0))), 0x3d)
    }
    var__mpos := expr_mpos
}
```
### Inline assembly version passed through Yul optimizer
```solidity
function checked_add_uint256_794(x) -> sum
{
    if gt(x, not(32)) { panic_error_0x11() }
    sum := add(x, 0x20)
}
function fun_encode(var_data_mpos) -> var_mpos
{
    if iszero(mload(var_data_mpos))
    {
        var_mpos := allocate_memory_array_string()
        leave
    }
    let expr_37_mpos := copy_literal_to_memory_84d8a590de33e00cbdc16e1f28c3506f5ec15c599fab9a6a4bcd575cc2f110ce()
    let expr := checked_mul_uint256(checked_div_uint256(checked_add_uint256(mload(var_data_mpos))))
    let expr_mpos := allocate_and_zero_memory_array_string(checked_add_uint256_794(expr))
    mstore(expr_mpos, expr)
    let usr$dataPtr := var_data_mpos
    let usr$endPtr := add(var_data_mpos, mload(var_data_mpos))
    let usr$resultPtr := add(expr_mpos, 0x20)
    for { } lt(usr$dataPtr, usr$endPtr) { }
    {
        let _1 := 0x03
        usr$dataPtr := add(usr$dataPtr, _1)
        let usr$input := mload(usr$dataPtr)
        let _2 := 1
        let _3 := 0x3F
        let _4 := mload(add(add(expr_37_mpos, and(shr(18, usr$input), _3)), _2))
        let _5 := 248
        mstore(usr$resultPtr, shl(_5, _4))
        mstore(add(usr$resultPtr, _2), shl(_5, mload(add(add(expr_37_mpos, and(shr(12, usr$input), _3)), _2))))
        mstore(add(usr$resultPtr, 0x02), shl(_5, mload(add(add(expr_37_mpos, and(shr(6, usr$input), _3)), _2))))
        mstore(add(usr$resultPtr, _1), shl(_5, mload(add(add(expr_37_mpos, and(usr$input, _3)), _2))))
        usr$resultPtr := add(usr$resultPtr, 0x04)
    }
    switch mod(mload(var_data_mpos), 0x03)
    case 1 {
        mstore(add(usr$resultPtr, not(1)), shl(240, 15677))
    }
    case 2 {
        mstore(add(usr$resultPtr, not(0)), shl(248, 61))
    }
    var_mpos := expr_mpos
}
```

Amxx:
Note I'm trying to evaluate the different implementation. It's been long overdue, but we would want this to be part of our next release.

I'm afraid to say that @cameel's implementation is not correct. It just doesn't return the expected value. I'll be trying to fix it so I can then accurately measure the impact on gas cost (if that was ever called as part of a transaction) and deployment cost.

Amxx:
There is a major issue with the solidity implementation, and that is that, if the data length is not a multiple of 3, the last iteration of the loop will fail. Reading `data[dataPtr + 1]` or `data[dataPtr + 2]` will fall outside of the array, and even with unchecked, this causes a faillure

Amxx:
Here is a working solidity implementation. Even with optimisation is is SIGNIFICANTLY more expensive to run:

```
library Base64V2 {
    bytes private constant TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";

        function encode(bytes memory data) public pure returns (string memory) {
        if (data.length == 0) return "";

        bytes memory table     = TABLE;
        bytes memory result    = new bytes(4 * ((data.length + 2) / 3));
        uint256      resultPtr = 0;

        for (uint256 dataPtr = 0; dataPtr < data.length; dataPtr += 3) {
            uint24 chunk = (                            (uint24(uint8(data[dataPtr + 0])) << 16))
                         + (dataPtr + 1 < data.length ? (uint24(uint8(data[dataPtr + 1])) <<  8) : 0)
                         + (dataPtr + 2 < data.length ? (uint24(uint8(data[dataPtr + 2]))      ) : 0);
            
            result[resultPtr++] = table[uint8(chunk >> 18) & 0x3f];
            result[resultPtr++] = table[uint8(chunk >> 12) & 0x3f];
            result[resultPtr++] = table[uint8(chunk >>  6) & 0x3f];
            result[resultPtr++] = table[uint8(chunk      ) & 0x3f];
        }

        if (data.length % 3 == 1) {
            result[--resultPtr] = 0x3d;
            result[--resultPtr] = 0x3d;
        }
        else if (data.length % 3 == 2) {
            result[--resultPtr] = 0x3d;
        }

        return (string(result));
    }
}
```

Amxx:
The bytecode for the solidity version is 77.2% larger. This increases the deployment cost by 156k gas.

I'll go for the assembly any days. Its actually pretty clear / simple when you known of base64 encoding works

contracts/utils/Base64.sol
```diff
@@ -0,0 +1,96 @@
...
+        // Encoding takes 3 bytes chunks of binary data from `bytes` data parameter
+        // and split into 4 numbers of 6 bits.
+        // The final Base64 length should be `bytes` data length multiplied by 4/3 rounded up
+        // - `data.length + 2`  -> Round up
+        // - `/ 3`              -> Number of 3-bytes chunks
+        // - `4 *`              -> 4 characters for each chunk
+        uint256 encodedLen = 4 * ((data.length + 2) / 3);
+
+        // Add some extra buffer at the end required for the writing
+        string memory result = new string(encodedLen + 32);
```
Amxx, line 36: Why do we need the additional space for ?
the padding equals already fit into the encodedLen

cameel:
> I'm afraid to say that @cameel's implementation is not correct.

Thanks for fixing it. Like I said, my version was not tested at all. I was mostly interested in a rough comparison of the generated bytecode to get a good picture of whether going to assembly here is worth it here or not and for that purpose it was good enough.

> Even with optimisation is is SIGNIFICANTLY more expensive to run:

Some of it is because your version does not have the `unchecked` block. It's not something I'd recommend in general but might be a better middle-ground before going straight to assembly.

But yeah, it's not the only thing that makes it more expensive. I'm also pretty disappointed by stuff that was not optimized out. In the current version of the compiler the Solidity code is much more expensive. Just wanted to let you know that many of missed optimization opportunities here are things we're actively working on right now so it's not going to stay like this indefinitely.

contracts/utils/Base64.sol
cameel: I think it's because the code operates on whole 32-byte slots rather than single bytes. It uses `mstore` instead of `mstore8` so when setting the value for a byte it also writes over 31 bytes after it.

Amxx: then lets use mstore8 !

Amxx (approved):

Amxx (approved):

Drjacky:
Are we still need something to get this PR merged? 🤔

Amxx:
@Drjacky We still need some internal review. This should make it into 4.5

frangio (approved):
This looks good to me.

ernestognw:
> Hello @ernestognw, thank you very much for this PR, and sorry we didn't answer it sooner.

I'm happy it's getting merged soon. To be honest, the optimization discussion went out of my scope, so thanks @cameel for providing extra examples and doing some tests.

Let me know if there's anything else I can do. Otherwise here's a lovely meme:

simpsons-leonardo-nimoy

PR #2892, opened 2021-09-30
Delay the Pending state until strictly after proposal.voteStart

### LINKED ISSUE #2886: Bug: Governor state
for example: when startBlock=100000

A have 200 COMP 
B no COMP
when block.number=10000
OpenZeepelin  state is Active (is error)
Compound state is Pending  (is OK)

A castVote. and transfer to B. B is also castVote.  in block.number=10000

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/1b27c13096d6e4389d62e7b0766a1db53fbb3f1b/contracts/governance/Governor.sol#L111-L130

https://github.com/compound-finance/compound-protocol/blob/a6251ecc30adb8dfa904e2231e9a9fdb4eea78be/contracts/Governance/GovernorAlpha.sol#L226-L246

Amxx:
Delay the Pending state until strictly after proposal.voteStart to avoid issue with potential invalid implementations of the voting token

Fixes #2886

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -115,7 +115,8 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
             return ProposalState.Executed;
         } else if (proposal.canceled) {
             return ProposalState.Canceled;
-        } else if (proposal.voteStart.isPending()) {
+        } else if (proposal.voteStart.getDeadline() >= block.number) {
+            // prevent voting until voteStart is passed, so that getVotes never tries to get value for the current block.
             return ProposalState.Pending;
         } else if (proposal.voteEnd.isPending()) {
             return ProposalState.Active;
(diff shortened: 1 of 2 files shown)


contracts/governance/Governor.sol
```diff
@@ -115,7 +115,8 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
             return ProposalState.Executed;
         } else if (proposal.canceled) {
             return ProposalState.Canceled;
-        } else if (proposal.voteStart.isPending()) {
+        } else if (proposal.voteStart.getDeadline() >= block.number) {
+            // prevent voting until voteStart is passed, so that getVotes never tries to get value for the current block.
             return ProposalState.Pending;
         } else if (proposal.voteEnd.isPending()) {
```
frangio, line 121: Note: Also change end so that we align this with Bravo.

frangio:
Missing a changelog entry. "Delayed Governor vote start and end by one block"?

PR #2910, opened 2021-10-12
Prevent transfers to the zero address

ghost:
The SafeERC20 wrapper library should revert on transfers to the zero address to prevent funds from being accidentally lost and to maintain consistency with OpenZeppelin's [ERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol) implementation.


frangio:
We discussed this but we don't think SafeERC20 should be this opinionated.

If the underlying token supports transfer to 0 then we don't think we should disallow that.

PR #2915, opened 2021-10-16
updated docs for governor votes, 0 is against not for

PatrickAlphaC:
The docs were outdated. Took me a few hours to find out what was going wrong. In both the compound and oz versions of counting votes, 0 is against, but in the docs, it said 0 was for. 

Examples:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/56d4063e928d963a96e376f2dc71831cfecf5214/contracts/governance/IGovernor.sol#L76

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/56d4063e928d963a96e376f2dc71831cfecf5214/contracts/governance/compatibility/GovernorCompatibilityBravo.sol#L31

#### PR Checklist

- [x ] Tests
- [x ] Documentation
- [x ] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
- * `IGovernor`: Corrected documentation for what vote support stands for. 0 is against and 1 is for. 
+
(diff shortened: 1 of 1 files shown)


PatrickAlphaC:
I almost didn't and then I saw the checklist.

I chose to err on the side of caution, but good to know!

PR #2917, opened 2021-10-18
Re-enable immutable forwarder in ERC2771Context

### LINKED ISSUE #2905: Add back immutable in ERC2771Context
We had to remove use of immutable since the variable could not be read during the constructor, causing all sorts of issues as pointed out in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2739.

Since Solidity 0.8.8 this should no longer be a problem (though we should confirm with a test).

Amxx:
Fixes #2905 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/metatx/ERC2771Context.sol
+++ b/contracts/metatx/ERC2771Context.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts v4.3.2 (metatx/ERC2771Context.sol)
 
-pragma solidity ^0.8.8;
+pragma solidity ^0.8.9;
 
 import "../utils/Context.sol";
 
--- a/contracts/metatx/ERC2771Context.sol
+++ b/contracts/metatx/ERC2771Context.sol
@@ -9,6 +9,7 @@ import "../utils/Context.sol";
  * @dev Context variant with ERC2771 support.
  */
 abstract contract ERC2771Context is Context {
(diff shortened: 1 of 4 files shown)


contracts/metatx/ERC2771Context.sol
```diff
@@ -1,15 +1,15 @@
...
+pragma solidity ^0.8.9;
 
 import "../utils/Context.sol";
 
 /**
  * @dev Context variant with ERC2771 support.
  */
 abstract contract ERC2771Context is Context {
-    address private _trustedForwarder;
+    address private immutable _trustedForwarder;
```
frangio, line 12: Do we want this to be immutable in Upgradeable Contracts?

Amxx: I think so.

frangio, line 12:
```suggestion
    /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
    address private immutable _trustedForwarder;
```

frangio:
Note: Missing changelog entry.

frangio (approved):

frangio:
I want to wait to merge once we've finished the 4.4 release, because I realized there will be some issues around the immutable variable and the transpiler.

frangio:
Should be good now for the transpiler I think.

PR #2926, opened 2021-10-28
Add a relay mechanism in the governor

Amxx:
#### PR Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -348,8 +348,10 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
     }
 
     /**
-     * @dev Relaying mechanism allowing the acting governor (possibly a timelock) to extract value and token sent to
-     * the governor by mistake.
+     * @dev Relays a transaction or function call to an arbitrary target. In cases where the governance executor
+     * is some contract other than the governor itself, like when using a timelock, this function can be invoked
+     * in a governance proposal to recover tokens or Ether that was sent to the governor contract by mistake.
+     * Note that if the executor is simply the governor itself, use of `relay` is redundant.
      */
     function relay(
         address target,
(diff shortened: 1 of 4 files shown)


contracts/governance/Governor.sol
```diff
@@ -347,6 +347,18 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
         return weight;
     }
 
+    /**
+     * @dev Relaying mechanism allowing the acting governor (possibly a timelock) to extract value and token sent to
```
frangio, line 351:
```suggestion
     * @dev Relaying mechanism allowing the acting governor (possibly a timelock) to extract value and tokens sent to
```

contracts/governance/Governor.sol
```diff
@@ -347,6 +347,18 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
...
+     * the governor by mistake.
```
frangio, line 352:
```suggestion
     * @dev Relays a transaction or function call to an arbitrary target. In cases where the governance executor
     * is some contract other than the governor itself, like when using a timelock, this function can be invoked
     * in a governance proposal to recover tokens or Ether that was sent to the governor itself by mistake.
     * Note that if the executor is simply the governor itself, use of `relay` is redundant.
```

frangio:
Does it make sense to add a test for plain `Governor` without a timelock?

test/governance/extensions/GovernorTimelockCompound.test.js
```diff
@@ -328,6 +328,57 @@ contract('GovernorTimelockCompound', function (accounts) {
     runGovernorWorkflow();
   });
 
+  describe('relay', function () {
```
frangio, line 331: Does it make sense to add a test for plain Governor without a timelock?

Amxx: The Governor without a timelock already acts by itself. using relay would add an indirection level and not add anything. So it doesn't add anything to the non-timelock flavors ... but we can still test it (and test its protected)

frangio (dismissed):
LGTM

frangio (approved):

PR #2935, opened 2021-11-02
Improve docs wording on constructors and proxies

parseb:
Fixes https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/445 

> _Since **a proxied contract can't have a constructor**, it's common to move constructor logic to an
> external initializer function, usually called initialize._

As outlined in the issue, proxied contract(s) retain their natural ability to utilize a constructor.
The current formulation negates this language level constant, namely, that all contracts can have a constructor.
  
**Replaced with:**

>  Since **proxied contracts do not make use of a constructor**, it's common to move constructor logic to an
>  external initializer function, usually called `initialize`.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/proxy/utils/Initializable.sol
+++ b/contracts/proxy/utils/Initializable.sol
@@ -5,7 +5,7 @@ pragma solidity ^0.8.0;
 
 /**
  * @dev This is a base contract to aid in writing upgradeable contracts, or any kind of contract that will be deployed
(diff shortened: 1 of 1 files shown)


frangio (approved):
Thank you!

PR #2944, opened 2021-11-03
Add ERC721Votes for NFT-based governance

### LINKED ISSUE #2873: Add the possibility to include ERC721 token voting within OZ Governor module
**🧐 Motivation**
As stated in the [governance introductory article](https://blog.openzeppelin.com/governor-smart-contract/):
> OpenZeppelin Contracts contain the most common requirements out of the box. Beyond these, OpenZeppelin Governor could potentially be configured with ERC721 voting power so certain NFT owners can participate in governance when specified.

I was wondering if such possibility is already explored somewhere or not yet...
I imagine the customization and usecases could be infinite, but what would imply to have a basic version of the OZ Governor that uses ERC721 tokens to create proposals, delegate and vote?

This issue is a continuation of the https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2868 one.
In [this thread](https://forum.openzeppelin.com/t/using-governor-with-erc721/15824) I explain a little bit more the usecase I'm working on.

**📝 Details**
Following @Amxx suggestions I drafted 3 new contracts:
- [GovernorERC721Votes](https://github.com/ivanminutillo/openzeppelin-contracts/blob/master/contracts/governance/extensions/GovernorERC721Votes.sol): A new governance flavor extension for voting weight extraction from an ERC721Votes token. 
- [ERC721Votes](https://github.com/ivanminutillo/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Votes.sol): A clone of ERC20Votes, with the same functions - included `getPastVotes` - but that inherits `NFTPermit` instead of `draft-ERC20Permit`
- [NFTPermit](https://github.com/ivanminutillo/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/NFTPermit.sol): A draft implementation of Permit for the ERC721 token (taken from [here](https://gist.github.com/Amxx/e3c87476093a6b27d9271b2e54b35292))

As I am not really proficient with solidity, I expect there's lot of devil in the details when implementing even the most trivial things, so I was wondering if I'm totally out of track or not?

JulissaDantes:
Fixes #2873  

Governance with NFT-backed votes. An ERC721Votes extension of ERC721 and a GovernorVotesERC721 module for OZ Governor contract. The contracts is be analogous to ERC20Votes and GovernorVotes, but for NFTs.

#### PR Checklist

- [X] Tests
- [X] Documentation
- [x] Changelog entry

#### Suggested Next steps

So far this implementation account for 1 Token = 1 vote, if a `_getVotingPower` internal virtual function is created, taking the token Id and returning the voting power of such token, that would allow for differently weighted token(you could call it token rarity or token value) to be used as voting power.


contracts/token/ERC721/extensions/ERC721Votes.sol
```diff
@@ -0,0 +1,294 @@
...
+     * @dev Move voting power when tokens are transferred.
+     *
+     * Emits a {DelegateVotesChanged} event.
+     */
+    function _afterTokenTransfer(
+        address from,
+        address to,
+        uint256 tokenId
+    ) internal virtual override{
+        _moveVotingPower(delegates(from), delegates(to), 1);
```
frangio, line 209: It's very important to include a call to `super._afterTokenTransfer` here. See [Rules of Hooks](https://docs.openzeppelin.com/contracts/4.x/extending-contracts#rules_of_hooks).

contracts/token/ERC721/extensions/ERC721Votes.sol
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * @dev Move voting power when tokens are transferred.
+     *
+     * Emits a {DelegateVotesChanged} event.
+     */
+    function _afterTokenTransfer(
+        address from,
+        address to,
+        uint256 tokenId
+    ) internal virtual override{
```
frangio, line 208:
```suggestion
    ) internal virtual override {
```

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -0,0 +1,557 @@
...
+
+contract('ERC721Votes', function (accounts) {
+  const [ holder, recipient, holderDelegatee, recipientDelegatee, other1, other2 ] = accounts;
+  const NFT1 = new BN('10');
+  const NFT2 = new BN('20');
+  const NFT3 = new BN('30');
+  const name = 'My Token';
+  const symbol = 'MTKN';
+  const version = '1';
+  const initalTokenId = new BN('10000000000000000000000000');
```
frangio, line 64: This variable name doesn't make a lot of sense to me. It seems to be just another one of NFT1,2,3.

I think the name "NFT0" would be better.

frangio (commented):
Thank you, great work!

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
```diff
@@ -284,6 +284,11 @@ abstract contract ERC721Votes is ERC721, EIP712 {
         return _domainSeparatorV4();
     }
 
+    /**
+    * @dev Returns token voting power
+    */
+    function _getVotingPower(uint tokenId) internal virtual returns(uint256) {}
```
frangio, line 290: This will default to 0! We need to add `return 1;` as a default.

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,7 +343,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
 
         emit Transfer(from, to, tokenId);
 
-        _afterTokenTransfer(from, to, tokenId);
+        _afterTokenTransfer(from, to);
```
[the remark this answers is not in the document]
JulissaDantes: It wasnt being used at this point and was causing the lint error that didnt allow to test, but the next commits will include it for the `_getVotingPower` usage.

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
JulissaDantes: Solved on 29061447c0e3372d5cd3e8b7ff918f34d918e8e2

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
```diff
@@ -44,7 +45,8 @@ abstract contract ERC721Votes is ERC721, EIP712 {
      * @dev Initializes the {EIP712} domain separator using the `name` parameter, and setting `version` to `"1"`.
      *
      * It's a good idea to use the same `name` that is defined as the ERC721 token name.
-    */
+     */
+    constructor(string memory name, string memory symbol) ERC721(name, symbol) EIP712(name, "1") {}
```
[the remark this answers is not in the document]
JulissaDantes: If I omit the ERC721 constructor declaration I get a `missing implementation`of that constructor error, do you know a workaround about it?

frangio:
Note regarding next steps: There are some challenges around implementing `delegate` because we need to know the voting power of an account even before it sets a delegate. So we might need to keep track of that incrementally, which will introduce some overhead to transfers. We should try to minimize that overhead.

byterose:
[LibGovernor](https://github.com/byterose/nffeels-contracts/blob/main/contracts/libraries/LibGovernor.sol)
[LibERC721](https://github.com/byterose/nffeels-contracts/blob/main/contracts/libraries/LibERC721.sol)

I implemented erc721 voting with different weights. Since I was working with a diamond I moved the whole logic into a library, though the function names are the same as in the oz implementation,

Maybe this helps as well..

frangio:
We will resume this PR next week.

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -382,13 +382,12 @@ contract('ERC721Votes', function (accounts) {
     describe('numCheckpoints', function () {
       it('returns the number of checkpoints for a delegate', async function () {
 
-        await this.token.transferFrom(holder, recipient, NFT1, { from: holder }); //give an account two tokens for readability
+        await this.token.transferFrom(holder, recipient, NFT1, { from: holder }); //give an account two tokens for readability
```
Amxx, line 385: There are some trailing whitespaces on line 385 and 388. Does that pass the linter ? IMO it shouldn't

JulissaDantes: Solved with commit 51002f9

contracts/governance/extensions/GovernorVotesERC721.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts v4.3.2 (governance/extensions/ERC721GovernorVotesERC721.sol)
+
+pragma solidity ^0.8.0;
+
+import "../Governor.sol";
+import "../../token/ERC721/extensions/draft-ERC721Votes.sol";
+import "../../utils/math/Math.sol";
```
Amxx, line 8: It looks like Math.sol is not needed here (same comment apply to `governance/extensions/GovernorVotes.sol`)

contracts/governance/extensions/GovernorVotesERC721.sol
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @dev Extension of {Governor} for voting weight extraction from an {ERC721Votes} token.
+ *
+ * _Available since v4.5._
+ */
+abstract contract GovernorVotesERC721 is Governor {
```
Amxx, line 15: AFAIK, this module produces the exact same result as GovernorVotesERC20, because the `getPastVotes` method is shared. Maybe we should not have 2 identical modules

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts v4.4.0 (token/ERC20/extensions/ERC20Votes.sol)
+// OpenZeppelin Contracts v4.3.2 (token/ERC20/extensions/ERC20Votes.sol)
```
Amxx, line 2: This should not be touched.

The scripts will overwrite this when we release anyway, but it would be better not to change it.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ## 4.4.0 (2021-11-25)
 
+## Unreleased
+ * `Voting`: Create library to be use for ERC721 and ERC1155 voting ([#2944](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2944))
```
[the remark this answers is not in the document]
JulissaDantes: Commit acd9a00 fixed it

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
```diff
@@ -0,0 +1,225 @@
...
+ * Enabling self-delegation can easily be done by overriding the {delegates} function. Keep in mind however that this
+ * will significantly increase the base gas cost of transfers.
+ *
+ * _Available since v4.5._
+ */
+abstract contract ERC721Votes is ERC721, EIP712 {
+    using Counters for Counters.Counter;
+    using Voting for Voting.Votes;
+
+    uint256 _totalVotingPower;
```
Amxx, line 34: What do we need that for, isn't that a duplicate of 
```
_votes._totalCheckpoints[_votes._totalCheckpoints.length].value
```
?

Amxx: My bad, I just realized this is an equivalent of totalSupply().

I'm not a big fan of this for two reasons:
- it is a duplicate of something present in ERC721Enumerable
- it is actually expensive to maintain
- it protects against something that, IMO, will never happen.

Unlike ERC20, ERC721 are minted one at a time, for it to come into action, we would need to function to be called 2**224 time. There is not enough energy in our universe for that to be remotely possible.

I suggest we drop this storage slot, and the corresponding checks. @frangio ?

Amxx:
I really think `_afterTokenTransfer` should not only be called by `_transfer`, but also by `_mint` and `_burn`. Future extensions will expect that hook to be called for all transfers, including minting and burning ones.

this would mean that it would no longer be right to put the `_votes.transfer` into the `_afterTokenTransfer` hook, but rather in an overloaded version of `_transfer`, like I did here: https://github.com/Amxx/amxx-hre/blob/master/contracts/extensions/ERC721Votes.sol

Amxx:
I also think we should be super careful about the ordering of updates (before/after the super.call), because of `onERC721Received` issues (reintrancy ?)

contracts/governance/extensions/GovernorVotesERC721.sol
JulissaDantes: Wouldnt it serve as a guide for users trying to implment Governance with ERCVotes? Or should we include some documentation somewhere instead?

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
JulissaDantes: The first approach was to inherit from `ERC721Enumerable` but it added unnecesarry overhead, since that not the only thing in that contract, and that's why we needed to keep it as a new storage variable.

Amxx: But do we? It's only used for checking that minting if not going over 2**224, which is never going to happen in this univers

JulissaDantes:
> I really think `_afterTokenTransfer` should not only be called by `_transfer`, but also by `_mint` and `_burn`. Future extensions will expect that hook to be called for all transfers, including minting and burning ones.
> 
> this would mean that it would no longer be right to put the `_votes.transfer` into the `_afterTokenTransfer` hook, but rather in an overloaded version of `_transfer`, like I did here: https://github.com/Amxx/amxx-hre/blob/master/contracts/extensions/ERC721Votes.sol

Since the `_afterTokenTransfer` just moves voting power and I proceeded with the overload of the transfer function, which does the voting power rearrangement on the `_votes.transfer` call, I removed the `_afterTokenTransfer` from the contract no need to call it after mit and burn since both mint and burn move the voting power as well on their `_votes.mint` and `_votes.burn` calls.

contracts/token/ERC721/ERC721.sol
```diff
@@ -423,8 +423,8 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         address to,
         uint256 tokenId
     ) internal virtual {}
-
-    /**
+     
+     /**
```
Amxx: What IDE are you using? Can you configure it to remove trailing whitespaces when saving?
this part of the commit is really strange

JulissaDantes: I'm using VScode, just turned on the setting, the trailing space should not be there now.

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
```diff
@@ -27,112 +25,12 @@ import "../../../utils/cryptography/draft-EIP712.sol";
...
-     * @dev Delegate votes from the sender to `delegatee`.
-     */
-    function delegate(address delegatee) public virtual {
-        _delegate(_msgSender(), delegatee);
-    }
-
+abstract contract ERC721Votes is ERC721, Votes {
     /**
-     * @dev Delegates votes from signer to `delegatee`
+     * @dev Returns total amount of votes at given position.
```
frangio, line 30: "Position" should be "block number", and note that the function argument is "timestamp" and should also be "blockNumber".

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./Context.sol";
+import "./Counters.sol";
+import "./Checkpoints.sol";
+import "./cryptography/draft-EIP712.sol";
+
+/**
+ * @dev Voting operations.
```
frangio, line 10: We need more extensive documentation that explains what the contract is and how it's used. Important points to mention: when using this module, the derived contract must implement `_getDelegatorVotes`, and can use `_moveVotingPower` when a delegator's voting power is changed.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+abstract contract Votes is Context, EIP712 {
+    using Checkpoints for Checkpoints.History;
+    using Counters for Counters.Counter;
+
+    bytes32 private constant _DELEGATION_TYPEHASH =
+        keccak256("Delegation(address delegatee,uint256 nonce,uint256 expiry)");
+    mapping(address => address) _delegation;
+    mapping(address => Checkpoints.History) _userCheckpoints;
+    mapping(address => Counters.Counter) private _nonces;
+    Checkpoints.History _totalCheckpoints;
```
frangio, line 21: Variables need to be private.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+     * @dev Returns total amount of votes for account.
+     */
+    function getVotes(address account) public view returns (uint256) {
+        return _userCheckpoints[account].latest();
+    }
+
+    /**
+     * @dev Returns total amount of votes at given position.
+     */
+    function _getPastVotes(address account, uint256 timestamp) internal view returns (uint256) {
```
frangio, line 43: Why is there a combination of public (`getVotes`) and internal (`_getPastVotes`) functions? I think they should all be public functions.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+
+    /**
+     * @dev Emitted when a token transfer or delegate change results in changes to an account's voting power.
+     */
+    event DelegateVotesChanged(address indexed delegate, uint256 previousBalance, uint256 newBalance);
+
+    /**
+     * @dev Returns total amount of votes for account.
+     */
+    function getVotes(address account) public view returns (uint256) {
```
frangio, line 36: All functions need to be marked `virtual`.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+     * @dev Get number of checkpoints for `account` including delegation.
+     */
+    function _getTotalAccountVotes(address account) internal view returns (uint256) {
+        return _userCheckpoints[account].length();
+    }
+
+    /**
+     * @dev Delegate votes from the sender to `delegatee`.
+     */
+    function delegate(address delegatee) public virtual {
```
frangio, line 88: Can you reorder functions so that `delegate` and `delegateBySig` are one after the other?

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+
+        _moveVotingPower(oldDelegation, newDelegation, balance);
+    }
+
+    /**
+     * @dev Moves voting power.
+     */
+    function _moveVotingPower(
+        address src,
+        address dst,
```
frangio, line 123: Let's rename this to `from` and `to` which are more common throughout the codebase.

docs/modules/ROOT/pages/governance.adoc
```diff
@@ -129,19 +129,20 @@ If your project requires The voting power of each account in our governance setu
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.2;
 
-import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
-import "@openzeppelin/contracts/token/ERC721/extensions/draft-ERC721Votes.sol";
+import "openzeppelin-solidity/contracts/token/ERC721/ERC721.sol";
+import "openzeppelin-solidity/contracts/token/ERC721/extensions/draft-ERC721Votes.sol";
```
frangio, line 133: This was fine it should be `@openzeppelin/contracts`.

docs/modules/ROOT/pages/governance.adoc
```diff
@@ -129,19 +129,20 @@ If your project requires The voting power of each account in our governance setu
...
 
 contract MyToken is ERC721Votes {
-    constructor() ERC721Votes("MyToken", "MTK"){}
+    constructor(string memory name, string memory symbol) ERC721(name, symbol) EIP712(name, "1") {}
```
frangio, line 136:
```suggestion
    constructor() ERC721("MyToken", "MTK") EIP712("MyToken", "1") {}
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -342,6 +342,8 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         _owners[tokenId] = to;
 
         emit Transfer(from, to, tokenId);
+
+        _afterTokenTransfer(from, to, tokenId);
```
frangio, line 346: We still need to add `_afterTokenTransfer` to `_mint` and `_burn`.

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
```diff
@@ -0,0 +1,80 @@
...
+     */
+    function getPastVotes(address account, uint256 timestamp) external view returns (uint256) {
+        return _getPastVotes(account, timestamp);
+    }
+
+    /**
+     * @dev Maximum token supply. Defaults to `type(uint224).max` (2^224^ - 1).
+     */
+    function _maxSupply() internal view virtual returns (uint224) {
```
frangio, line 39: We no longer use this, right? We can remove this function I think.

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
```diff
@@ -0,0 +1,80 @@
...
+        return type(uint224).max;
+    }
+
+    /**
+     * @dev Snapshots the totalSupply after it has been increased.
+     */
+    function _mint(address account, uint256 tokenId) internal virtual override {
+        super._mint(account, tokenId);
+        _afterTokenTransfer(address(0), account, tokenId);
```
frangio, line 48: `_afterTokenTransfer` needs to be inserted in mint and burn in the core `ERC721`, not here. Take a look at `_beforeTokenTransfer` for comparison.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+ * @dev Voting operations.
+ */
+abstract contract Votes is Context, EIP712 {
+    using Checkpoints for Checkpoints.History;
+    using Counters for Counters.Counter;
+
+    bytes32 private constant _DELEGATION_TYPEHASH =
+        keccak256("Delegation(address delegatee,uint256 nonce,uint256 expiry)");
+    mapping(address => address) _delegation;
+    mapping(address => Checkpoints.History) _userCheckpoints;
```
frangio, line 19: I think this should be renamed to `_delegateCheckpoints`.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+        return _delegation[account];
+    }
+
+    /**
+     * @dev Change delegation for `delegator` to `delegatee`.
+     *
+     * Emits events {DelegateChanged} and {DelegateVotesChanged}.
+     */
+    function _delegate(
+        address account,
```
frangio, line 106: `delegator` is a better variable name.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+        address newDelegation,
+        uint256 balance
```
frangio, line 108: We should remove the `balance` argument. It should always be `_getDelegatorVotes(delegator)`, we should use this expression in the function body.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function DOMAIN_SEPARATOR() external view returns (bytes32) {
+        return _domainSeparatorV4();
+    }
+
+    /**
+     * @dev Returns the balance of the delegator account
+     */
+    function _getDelegatorVotes(address) internal virtual returns (uint256);
```
frangio, line 209: I think we should use the term "voting power" instead of "votes" here. In that way it naturally looks like a counterpart to `moveVotingPower`.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+    /**
+     * @dev Moves voting power.
+     */
+    function _moveVotingPower(
+        address src,
+        address dst,
+        uint256 amount
+    ) internal {
+        if (src != dst && amount > 0) {
+            if (dst == address(0) && src != address(0)) {
```
Amxx, line 127: that is micro optimization, but if `dst == address(0)` then we know that `src != address(0)` ... otherwize we should not be here anyway ( we know that `src != dst`)

I'd do
```
if (src == address(0)) {
    _totalCheckpoints.push(_add, amount);
}
if (dst == address(0)) {
    _totalCheckpoints.push(_subtract, amount);
}
```

contracts/utils/Votes.sol
JulissaDantes: which one?

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /**
+     * @dev Moves voting power.
+     */
+    function _moveVotingPower(
+        address src,
+        address dst,
+        uint256 amount
+    ) internal {
```
Amxx, line 125: I'd make this private

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,210 @@
...
+     */
+    function delegate(address delegatee) public virtual {
+        address delegator = _msgSender();
+        _delegate(delegator, delegatee, _getDelegatorVotes(delegator));
+    }
+
+    /**
+     * @dev Returns account delegation.
+     */
+    function delegates(address account) public view returns (address) {
```
Amxx, line 96: This (and many other functions) should be virtual

contracts/utils/Votes.sol
frangio: No I think this needs to be internal, it's the main interface interface of this contract. Otherwise how does ERC721Votes trigger moving of votes on transfer?

Amxx: Right, I thought we would be exposing a `_afterTokenTransfer` but that would lead to confusion between tokenId and amount ...

contracts/utils/Checkpoints.sol
```diff
@@ -0,0 +1,87 @@
...
+    function length(History storage self) internal view returns (uint256) {
+        return self._checkpoints.length;
+    }
+
+    /**
+     * @dev Returns checkpoints at given position.
+     */
+    function at(History storage self, uint256 pos) internal view returns (Checkpoint memory) {
+        return self._checkpoints[pos];
+    }
```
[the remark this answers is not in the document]
Amxx: I had this functions because ERC20Votes has them ... and my guess is ERC20Votes has them for testing only (which I think is bad)

contracts/utils/Checkpoints.sol
```diff
@@ -0,0 +1,87 @@
...
+
+import "./math/Math.sol";
+import "./math/SafeCast.sol";
+
+/**
+ * @dev Checkpoints operations.
+ */
+library Checkpoints {
+    struct Checkpoint {
+        uint32 index;
```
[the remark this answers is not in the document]
Amxx: The initial idea was to make this agnostic, by having the `block.number` or `block.timestamp` abstracted. But we are not going to do that anytime soon, so lets change it.

contracts/utils/Votes.sol
Amxx: I still think we should have an intermediary function between this and the token hooks. I think it would be easy for token hooks to do

```
function _afterTokenTransfer(address from, address to, uint256 amount) internal virtual override {
    super._afterTokenTransfer(from, to, amount);
    _moveVotingPower(from, to, amount);
}
```
instead of
```
function _afterTokenTransfer(address from, address to, uint256 amount) internal virtual override {
    super._afterTokenTransfer(from, to, amount);
    _moveVotingPower(delegate(from), delegate(to), amount);
}
```

That is why I would make `_moveVotingPower` private and add a 
```
function _transferVotingPower(address from, address to, uint256 amount) internal virtual {
    _moveVotingPower(delegate(from), delegate(to), amount);
}
```

contracts/utils/Votes.sol
JulissaDantes: `getVotes` returns the amount of votes for an account, and `_getPastVotes` returns the total amount on a blockNumber.

contracts/utils/Votes.sol
JulissaDantes: fixed on `f054441f`

contracts/utils/Votes.sol
JulissaDantes: Fixed `f054441f`

contracts/utils/Votes.sol
JulissaDantes: Fixed `f054441f`

contracts/utils/Votes.sol
JulissaDantes: Fixed `f054441f`

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
JulissaDantes: Fixed `f054441f`

contracts/utils/Votes.sol
JulissaDantes: Fixed `0c8b601d`

contracts/governance/README.adoc
```diff
@@ -22,6 +22,8 @@ Votes modules determine the source of voting power, and sometimes quorum number.
 
 * {GovernorVotes}: Extracts voting weight from an {ERC20Votes} token.
 
+* {GovernorVotesERC721}: Extracts voting weight from an {ERC721Votes} token.
```
Amxx, line 25: We have GovernorVotes, so I guess this is not needed, is it ?

contracts/governance/README.adoc
```diff
@@ -66,6 +68,8 @@ NOTE: Functions of the `Governor` contract do not include access control. If you
 
 {{GovernorVotesQuorumFraction}}
 
+{{GovernorVotesERC721}}
```
Amxx, line 71: Same here

contracts/interfaces/IVotes.sol
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * @dev Delegate votes from the sender to `delegatee`.
+     */
+    function delegate(address delegatee) external;
+
+    /**
+     * @dev Returns account delegation.
+     */
+    function delegates(address account) external view returns (address);
```
Amxx, line 39: IMO this should be before delegate, so we don't interleave view and non-view functions

contracts/utils/Checkpoints.sol
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    struct History {
+        Checkpoint[] _checkpoints;
+    }
+
+    /**
+     * @dev Returns checkpoints length.
+     */
+    function length(History storage self) internal view returns (uint256) {
```
Amxx, line 23: If we don't have an `at()` function, we don't need that either, do we ?

contracts/mocks/CheckpointsImpl.sol
```diff
@@ -0,0 +1,27 @@
...
+pragma solidity ^0.8.0;
+
+import "../utils/Checkpoints.sol";
+
+contract CheckpointsImpl {
+    using Checkpoints for Checkpoints.History;
+
+    Checkpoints.History private _totalCheckpoints;
+
+    function length() public view returns (uint256) {
```
Amxx, line 12: I think this can be removed (see comment in the lib)

contracts/governance/README.adoc
JulissaDantes: You are right

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,206 @@
...
+ */
+abstract contract Votes is Context, EIP712, IVotes {
+    using Checkpoints for Checkpoints.History;
+    using Counters for Counters.Counter;
+
+    bytes32 private constant _DELEGATION_TYPEHASH =
+        keccak256("Delegation(address delegatee,uint256 nonce,uint256 expiry)");
+    mapping(address => address) private _delegateCheckpoints;
+    mapping(address => Checkpoints.History) private _userCheckpoints;
+    mapping(address => Counters.Counter) private _nonces;
```
Amxx, line 33: I'd prefer seeing that line after the _totalCheckpoints so that the 3 slots used for delegation:
```
mapping(address => address) private _delegate;
mapping(address => Checkpoints.History) private _userCheckpoints;
Checkpoints.History private _totalCheckpoints;
```

are consecutive.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,206 @@
...
+     * @dev Returns total amount of votes.
+     */
+    function _getTotalVotes() internal view virtual returns (uint256) {
+        return _totalCheckpoints.latest();
+    }
+
+    /**
+     * @dev Get number of checkpoints for `account` including delegation.
+     */
+    function _getTotalAccountVotes(address account) internal view virtual returns (uint256) {
```
Amxx, line 83: is that needed ?

contracts/mocks/GovernorVoteMock.sol
```diff
@@ -3,10 +3,10 @@
 pragma solidity ^0.8.0;
 
 import "../governance/extensions/GovernorCountingSimple.sol";
-import "../governance/extensions/GovernorVotesERC721.sol";
+import "../governance/extensions/GovernorVotes.sol";
 
-contract GovernorERC721Mock is GovernorVotesERC721, GovernorCountingSimple {
-    constructor(string memory name_, ERC721Votes token_) Governor(name_) GovernorVotesERC721(token_) {}
+contract GovernorVoteMock is GovernorVotes, GovernorCountingSimple {
```
frangio, line 8: Please change `GovernorVoteMock` to `GovernorVotesMocks` (with an `s`). Just trying to keep naming convention consistent.

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,206 @@
...
+     */
+    function delegate(address delegatee) public virtual override {
+        address delegator = _msgSender();
+        _delegate(delegator, delegatee);
+    }
+
+    /**
+     * @dev Returns account delegation.
+     */
+    function delegates(address account) public view virtual override returns (address) {
```
Amxx, line 98: Just like for the interface, I'd put that with the other view function, before `delegate(address)`

contracts/utils/Votes.sol
JulissaDantes: No, Im deleting it. realized thats the only function calling length

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,206 @@
...
+        return _delegateCheckpoints[account];
+    }
+
+    /**
+     * @dev Change delegation for `delegator` to `delegatee`.
+     *
+     * Emits events {DelegateChanged} and {DelegateVotesChanged}.
+     */
+    function _delegate(address delegator, address newDelegation) internal virtual {
```
Amxx, line 107: I'd put that after `delegateBySig` 

with my previous comment, that would result in the following:
```
function delegates(address) public view virtual override returns (address); // public view
function delegate(address) public virtual override;
function delegateBySig(address, uint256, uint256, uint8 v, bytes32 r, bytes32 s) public virtual override;
function _delegate(address, address) internal virtual;
```

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -25,7 +26,7 @@ import "../../../utils/cryptography/ECDSA.sol";
  *
  * _Available since v4.2._
  */
-abstract contract ERC20Votes is ERC20Permit {
+abstract contract ERC20Votes is ERC20Permit, IVotes {
```
frangio, line 29: As a general guideline, we tend to put interfaces before contracts in the inheritance list:

```suggestion
abstract contract ERC20Votes is IVotes, ERC20Permit {
```

docs/modules/ROOT/pages/governance.adoc
```diff
@@ -119,18 +123,52 @@ contract MyToken is ERC20, ERC20Permit, ERC20Votes, ERC20Wrapper {
...
+
+import "openzeppelin/contracts/token/ERC721/ERC721.sol";
+import "openzeppelin/contracts/token/ERC721/extensions/draft-ERC721Votes.sol";
+
+contract MyToken is ERC721Votes {
+    constructor() ERC721("MyToken", "MTK") EIP712("MyToken", "1") {}
+
+    // The functions below are overrides required by Solidity.
+
+    function _afterTokenTransfer(
```
Amxx, line 140: Why override if only one parent implements them ? I'm a bit lost here

contracts/utils/Votes.sol
```diff
@@ -21,7 +22,7 @@ import "./cryptography/draft-EIP712.sol";
  * When using this module, the derived contract must implement {_getDelegatorVotingPower}, and can use {_moveVotingPower}
  * when a delegator's voting power is changed.
  */
-abstract contract Votes is Context, EIP712 {
+abstract contract Votes is Context, EIP712, IVotes {
```
frangio, line 25: This is the exception, Context and Initializable go before interfaces:

```suggestion
abstract contract Votes is Context, IVotes, EIP712 {
```

contracts/token/ERC721/extensions/draft-ERC721Votes.sol
```diff
@@ -0,0 +1,49 @@
...
+// OpenZeppelin Contracts v4.4.0 (token/ERC721/extensions/draft-ERC721Votes.sol)
+
+pragma solidity ^0.8.0;
+
+import "../ERC721.sol";
+import "../../../utils/Votes.sol";
+import "../../../utils/math/Math.sol";
+import "../../../utils/Checkpoints.sol";
+import "../../../utils/math/SafeCast.sol";
+import "../../../utils/cryptography/ECDSA.sol";
```
Amxx, line 11: A lot of these imports are not needed. AFAIK, only ERC721 and Votes are needed.

contracts/mocks/VotingMock.sol
```diff
@@ -0,0 +1,33 @@
...
+
+    function getTotalVotes() public view returns (uint256) {
+        return _getTotalVotes();
+    }
+
+    function getTotalAccountVotes(address account) public view returns (uint256) {
+        return _getTotalAccountVotes(account);
+    }
+
+    function getTotalVotesAt(uint256 timestamp) public view returns (uint256) {
```
Amxx, line 18: Why have a public view for something that is already public view ? Also, this is not a timestamp, its a block number

contracts/utils/Checkpoints.sol
```diff
@@ -17,13 +17,6 @@ library Checkpoints {
...
 
-    /**
-     * @dev Returns checkpoints length.
-     */
-    function length(History storage self) internal view returns (uint256) {
-        return self._checkpoints.length;
-    }
-
     /**
      * @dev Returns total amount of checkpoints.
```
frangio, line 21:
```suggestion
     * @dev Returns the value in the latest checkpoint.
```

contracts/utils/Votes.sol
```diff
@@ -0,0 +1,206 @@
...
+ * When using this module, the derived contract must implement {_getDelegatorVotingPower}, and can use {_moveVotingPower}
+ * when a delegator's voting power is changed.
+ */
+abstract contract Votes is Context, EIP712, IVotes {
+    using Checkpoints for Checkpoints.History;
+    using Counters for Counters.Counter;
+
+    bytes32 private constant _DELEGATION_TYPEHASH =
+        keccak256("Delegation(address delegatee,uint256 nonce,uint256 expiry)");
+    mapping(address => address) private _delegateCheckpoints;
```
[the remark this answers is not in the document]
JulissaDantes: I updated the names

contracts/governance/utils/Votes.sol
```diff
@@ -0,0 +1,214 @@
...
+     */
+    function getPastTotalSupply(uint256 blockNumber) public view virtual override returns (uint256) {
+        require(blockNumber < block.number, "ERC721Votes: block not yet mined");
+        return _totalCheckpoints.getAtBlock(blockNumber);
+    }
+
+    /**
+     * @dev Returns total amount of votes.
+     */
+    function _getTotalVotes() internal view virtual returns (uint256) {
```
Amxx, line 78: I think this should be renamed `_getTotalSupply`

It would be more consistent with the other function
```
getVotes(address)
getPastVotes(address, uint256)
getPastTotalSupply(uint256)
```

contracts/governance/utils/Votes.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @dev Returns total amount of votes at given blockNumber.
+     */
+    function getPastVotes(address account, uint256 blockNumber) public view virtual override returns (uint256) {
+        return _delegateCheckpoints[account].getAtBlock(blockNumber);
+    }
+
+    /**
+     * @dev Retrieve the `totalVotingPower` at the end of `blockNumber`. Note, this value is the sum of all balances.
```
Amxx, line 63: `totalVotingPower` → `totalSupply`

contracts/governance/utils/Votes.sol
```diff
@@ -0,0 +1,214 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../utils/Context.sol";
+import "../../utils/Counters.sol";
+import "../../utils/Checkpoints.sol";
+import "./IVotes.sol";
```
Amxx, line 7: might be my OCD speaking here, But I'd prefer to have the imports more neatly ordered. IMO, `IVotes` should be after `draft-EIP721`.

contracts/governance/utils/Votes.sol
```diff
@@ -0,0 +1,214 @@
...
+     * @dev Subtracts two numbers.
+     */
+    function _subtract(uint256 a, uint256 b) private pure returns (uint256) {
+        return a - b;
+    }
+
+    /**
+     * @dev "Consume a nonce": return the current value and increment.
+     *
+     * _Available since v4.1._
```
Amxx, line 187: I don't think this should be here

contracts/governance/utils/Votes.sol
```diff
@@ -0,0 +1,214 @@
...
+ * power can be queried through {getVotes}.
+ *
+ * By default, token balance does not account for voting power. This makes transfers cheaper. The downside is that it
+ * requires users to delegate to themselves in order to activate checkpoints and have their voting power tracked.
+ * Enabling self-delegation can easily be done by overriding the {delegates} function. Keep in mind however that this
+ * will significantly increase the base gas cost of transfers.
+ *
+ * When using this module, the derived contract must implement {_getDelegatorVotingPower}, and can use {_moveVotingPower}
+ * when a delegator's voting power is changed.
+ */
```
Amxx, line 24: we may want an "available since 4.5" thing here

Amxx:
It might be a bit late to change that, but I'm always worried about the nonce thing.

What is another module wants to use signature and nonces later, do we require that it imports Votes or do we risk a clash in names ?

Would it make sens to have a 

```
contract withNonce {
     
    mapping(address => Counters.Counter) private _nonces;
    
    /**
     * @dev "Consume a nonce": return the current value and increment.
     */
    function _useNonce(address owner) internal virtual returns (uint256 current) {
        Counters.Counter storage nonce = _nonces[owner];
        current = nonce.current();
        nonce.increment();
    }

    /**
     * @dev Returns an address nonce.
     */
    function nonces(address owner) public view virtual returns (uint256) {
        return _nonces[owner].current();
    }
}
```

Votes could inherit that, and so could other modules.

contracts/governance/utils/Votes.sol
JulissaDantes: Resolved by `b171720c`

contracts/governance/utils/Votes.sol
JulissaDantes: Resolved by `b171720c`

contracts/governance/utils/Votes.sol
JulissaDantes: Resolved by `b171720c`

test/utils/Checkpoints.test.js
```diff
@@ -0,0 +1,31 @@
...
+    expect(await this.checkpoint.latest()).to.be.bignumber.equal('3');
+  });
+
+  it('calls getAtBlock', async function () {
+    expect(await this.checkpoint.getAtBlock(this.tx1.receipt.blockNumber - 1)).to.be.bignumber.equal('0');
+    expect(await this.checkpoint.getAtBlock(this.tx2.receipt.blockNumber - 1)).to.be.bignumber.equal('1');
+    expect(await this.checkpoint.getAtBlock(this.tx3.receipt.blockNumber - 1)).to.be.bignumber.equal('2');
+  });
+
+  it('reverts if block number >= current block', async function () {
```
frangio, line 25: This is checking > current block. To check = current block we need to add a function in `CheckpointsImpl` that calls `getAtBlock(..., block.number)`.

test/utils/Checkpoints.test.js
```diff
@@ -0,0 +1,31 @@
...
+const { expect } = require('chai');
+
+const CheckpointsImpl = artifacts.require('CheckpointsImpl');
+
+contract('Checkpoints', function (accounts) {
+  beforeEach(async function () {
+    this.checkpoint = await CheckpointsImpl.new();
+    this.tx1 = await this.checkpoint.push(1);
+    this.tx2 = await this.checkpoint.push(2);
+    this.tx3 = await this.checkpoint.push(3);
```
frangio, line 12: We need to check the behavior of `getAtBlock` for a block that did not have a checkpoint stored. For this we should add a line somewhere between pushes using `time.advanceBlock` from Test Helpers, and then add a new case to the `getAtBlock` test below.

test/utils/Checkpoints.test.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  it('calls latest', async function () {
```
frangio, line 15: We need to test `latest` when there are no checkpoints.

test/governance/utils/Votes.test.js
```diff
@@ -0,0 +1,41 @@
+const { expectRevert } = require('@openzeppelin/test-helpers');
+
+const { expect } = require('chai');
+
+const Votes = artifacts.require('VotesMock');
+
+contract('Voting', function (accounts) {
```
frangio, line 7: We are missing tests for a lot of the functionality and a lot of cases. As guidelines: we should have tests for all non-private functions, and for each of them there should be a test for every `require` on top of testing the basic functionality.

In this particular case most things are really covered in the tests for ERC721Votes. But they should be tested here as well. I think we should have a Votes "behavior" that we can use both for Votes on its own as well as ERC721Votes, and possibly also ERC20Votes. Maybe you can take the current tests from ERC721Votes and wrap them in a function that we can use in all those places? At least in Votes and ERC721Votes.

test/utils/Checkpoints.test.js
JulissaDantes: There is a function on `CheckpointsImpl ` that calls getAtBlock, what we need is to add another `expect` sending the current blocknumber.

frangio (approved):
LGTM. Thank you!

PR #2946, opened 2021-11-04
AccessControlEnumerable: Hook into the internal function

### LINKED ISSUE #938: A logical error in assertRevert.js
## 🎉 Description

- [x] 🐛 This is a bug report.
- [ ] 📈 This is a feature request.

## 💻 Environment

Agnostic (platform-independent)

## 📝 Details

If the `promise` is completed successfully, then you throw an exception with an error message containing `"revert"`.
You then immediately catch this exception, and since its error message contains `"revert"`, the `assert` completes successfully and no exception is thrown.

The thing to understand is that the Mocha framework reports a failure only when the tested scenario (`it`) throws an exception.
When that happens, the Mocha framework catches this exception and reports failure on the specific test at hand.
When you throw an exception and immediately catch it, the Mocha framework is not even aware of it.

For example, let's analyze a scenario in which the `promise` completes successfully, hence the test **should fail**:

- In the `try` part, the `promise` completes successfully
- In the `try` part, you throw an exception with `"revert"` in its error message
- In the `catch` part, you immediately catch that exception
- In the `catch` part, the `assert` completes successfully because the error message contains `"revert"`
- No exception is thrown, and the test **does not fail**

In order to understand the logical confusion, keep in mind that there are two different failure scenarios which you need to detect and handle:

1. No error has occurred (the `promise` has completed successfully)
2. An error without `"revert"` has occurred (the `promise` has failed but not because of `require`)

In your code, I believe that you detect the first scenario correctly, but you handle it incorrectly.

I therefore suggest the following alternative:

    try {
        await promise;
        throw null;
    }
    catch (error) {
        assert(error, `Expected an error but did not get one`);
        assert(error.message.includes("revert"), `Expected an error containing "revert" but got "${error.message}" instead`);
    }

Amxx:
Since `AccessControl` now expauses `_grantRole` and `_reboveRole` as internal function, `AccessControlEnumerable` should hook into that.

Fixes #938

#### PR Checklist
- [x] Tests
- [x] Changelog entry

### DIFF
--- a/contracts/access/AccessControlEnumerable.sol
+++ b/contracts/access/AccessControlEnumerable.sol
@@ -47,34 +47,18 @@ abstract contract AccessControlEnumerable is IAccessControlEnumerable, AccessCon
     }
 
     /**
-     * @dev Overload {grantRole} to track enumerable memberships
+     * @dev Overload {_grantRole} to track enumerable memberships
      */
-    function grantRole(bytes32 role, address account) public virtual override(AccessControl, IAccessControl) {
-        super.grantRole(role, account);
+    function _grantRole(bytes32 role, address account) internal virtual override {
+        super._grantRole(role, account);
         _roleMembers[role].add(account);
     }
 
     /**
-     * @dev Overload {revokeRole} to track enumerable memberships
+     * @dev Overload {_revokeRole} to track enumerable memberships
      */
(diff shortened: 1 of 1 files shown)


frangio (approved):
Thanks!

frangio:
Cherry picked onto 4.4.

PR #2956, opened 2021-11-08
Fix ERC1155 supply tracking

Amxx:
... to avoid data inconsistency during onReceive callbacks

We will publish more information about this issue soon.

#### PR Checklist

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 
 ## 4.3.3
 
- * `ERC1155Supply`: Handle "totalSupply" changes by hooking into `_beforeTokenTransfer` to ensure consistency of balances and supplies during `IERC1155Receiver.onERC1155Received` calls.
+ * `ERC1155Supply`: Handle `totalSupply` changes by hooking into `_beforeTokenTransfer` to ensure consistency of balances and supply during `IERC1155Receiver.onERC1155Received` calls.
 
 ## 4.3.2 (2021-09-14)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@
  * `Governor`: enable receiving Ether when a Timelock contract is not used. ([#2748](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2748))
  * `GovernorTimelockCompound`: fix ability to use Ether stored in the Timelock contract. ([#2748](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2748))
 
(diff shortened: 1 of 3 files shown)


Amxx:
This should be cherrypicked into release-4.3 and published as 4.3.3

frangio (approved):
LGTM. Thanks!

musecrypto:
Hi all, my project is using the old version of ERC1155Supply, can you explain the issue @Amxx ?

Amxx:
Hello @musecrypto 

The issue is a possible inconsistency in the value returned by totalsupply DURING a mint.

If you mint a token, the receiver is a smart contract, and the receiver implements `onERC1155Receive`, then this receiver is called with the balance already updated, but with the totalsupply not yet updated.

This is only an issue if both:
- minting isn't protected
- you totalsupply for things like voting

musecrypto:
Right, my apologies! I diffed this and mistakenly thought I had the new version and this update was to the old one. Thank you for explaining, that absolutely makes sense.

realisation:
> Hello @musecrypto
> 
> The issue is a possible inconsistency in the value returned by totalsupply DURING a mint.
> 
> If you mint a token, the receiver is a smart contract, and the receiver implements `onERC1155Receive`, then this receiver is called with the balance already updated, but with the totalsupply not yet updated.
> 
> This is only an issue if both:
> 
> * minting isn't protected
> * you totalsupply for things like voting

I wouldn't minimize this problem if I were you. This was just pointed out to me in an audit we were doing with CodeArena. If this had made it into our codebase, it would have been an absolute cataclysm.

We were using it in a way you'd expect to use it - using the industry standard smart contracts penned by OZ, top security firm - we were minting ERC1155's in response to user input. Does that mean the minting was unpermissioned? Absolutely not. And then we were using the total supply of the ERC1155 on unwind to calculate critical accounting information. 

If this had made it into our codebase, it would have been an ungodly, maybe even mortal frustration. Luckily it was an expense in gas and would have been removed on account of that, not to mention our audit. 

This stands as a serious message to any protocol developer out there. Absolutely never trust anything. Verify, always.

I just hope that OZ brings greater rigor to their smart contract audits. This is a standard re-entrancy bug.

musecrypto:
@realisation, yeah I realized this was a large issue once an auditing colleague send it to me as well to look out for.

We forked OZ to remove the mint receipt hook and have literally everything locked away with reentrancy guards, so luckily no damage here. I'm glad to hear the issue was caught before going into prod for you!

PR #2962, opened 2021-11-11
[5.0.0 Breaking change] Using user defined value type for timers

Amxx:
The structures used in the timers library prevent storage packing (in particular, this affects Governor's Proposals).

This has to effect that makes it a breaking change:

- Storage incompatibility (data is now packed) in the Governor contract.
- New timers being "value type", makes it impossible to write "setDeadline" and "reset" methods.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -25,6 +25,7 @@ import "./IGovernor.sol";
  */
 abstract contract Governor is Context, ERC165, EIP712, IGovernor {
     using SafeCast for uint256;
+    using Timers for uint64;
     using Timers for Timers.BlockNumber;
 
     bytes32 public constant BALLOT_TYPEHASH = keccak256("Ballot(uint256 proposalId,uint8 support)");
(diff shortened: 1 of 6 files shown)


frangio:
This needs to be rebased on `next-5.0`.

frangio:
@Amxx Do we still want to do this or is your current belief that it's better to use `uint48` types instead of introducing the timer abstraction?

Amxx:
In 5.0 we definitelly want to move away from struct-based timers has this prevents proper packing. 

This leaves us with some options:
- using UDVT (as shown in the PR)
  - the underlying type should probably be changed from uint64 to uint48
  - we'll continue to have both timestamp & blocknumber
- remove the Timers library altogether

A third option might be to add an abstract "withClock" contract, that implement EIP-5805's `clock()`, a UDVT, and the internal function to compare that UDVT to the clock
(I'm not sure about this third option, need more toughts)

frangio:
Closing per #4062.

PR #2973, opened 2021-11-15
Add a governor module to protect against late quorum 

Amxx:
This PR adds a new governor module that fixes an issue described by @Arachnid.

In cases where voters' participation is low, and quorum is far from being reached, voters might not be incentivized to participate for a proposal that they don't think will reach the quorum anyway. This would allow a "whale" voter to take over the vote at the end of the voting period. The whale would have enough tokens to reach the quorum and impose its result, leaving no time for voters to react.

This module ensures a minimal voting delay after quorum is reached. If quorum is reached late, then the proposal deadline will be extended to ensure this minimal post-quorum voting duration.

#### PR Checklist
- [x] Tests
- [x] Documentation
- [x] Changelog entry


Amxx:
Note: if proposal details become packed as part of a breaking release (proposed in #2962), commit #ff0d213 should be reverted for gas optimization.

Arachnid (dismissed):

contracts/governance/Governor.sol
```diff
@@ -110,24 +110,34 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
      * @dev See {IGovernor-state}.
      */
     function state(uint256 proposalId) public view virtual override returns (ProposalState) {
-        ProposalCore memory proposal = _proposals[proposalId];
+        ProposalCore storage proposal = _proposals[proposalId];
```
Amxx, line 113: Note: this is more gas effective in the current format, where proposal is not packed and uses 3 slots. If proposal are packed in an upcoming breaking change, then this should be changed back to memory.

contracts/governance/extensions/GovernorExtendedVoting.sol
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * @dev Extension of {Governor} for extending voting past quorum being reached.
+ *
+ * _Available since v4.X._
+ */
+abstract contract GovernorExtendedVoting is Governor {
+    using SafeCast for uint256;
+    using Timers for Timers.BlockNumber;
+
+    uint64 private _votingDelayExtension;
```
frangio, line 18: "Delay" is not the right term. `votingDelay` measures delay between proposal creation and start of the vote.

The voting "period" what is extended by this module.

contracts/governance/extensions/GovernorExtendedVoting.sol
```diff
@@ -0,0 +1,91 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts vX.X.X (governance/extensions/GovernorExtendedVoting.sol)
```
frangio, line 2: Just so you know, this is added automatically so it's not necessary to include it manually in a PR.

contracts/governance/Governor.sol
```diff
@@ -110,24 +110,34 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
...
-            return
-                _quorumReached(proposalId) && _voteSucceeded(proposalId)
-                    ? ProposalState.Succeeded
-                    : ProposalState.Defeated;
-        } else {
-            revert("Governor: unknown proposal id");
         }
+
+        return
+            _quorumReached(proposalId) && _voteSucceeded(proposalId) ? ProposalState.Succeeded : ProposalState.Defeated;
```
frangio, line 140: Personally, I think an `if-else` here would make the function look cleaner.

contracts/governance/extensions/GovernorExtendedVoting.sol
```diff
@@ -0,0 +1,91 @@
...
+    function votingDelayExtension() public view virtual returns (uint64) {
+        return _votingDelayExtension;
+    }
+
+    /**
+     * @dev Update the voting delay extension duration. This operation can only be performed through a governance proposal.
+     *
+     * Emits a {VotingDelayExtensionSet} event.
+     */
+    function setVotingDelayExtension(uint64 newVotingDelayExtension) public onlyGovernance {
```
frangio, line 78: This should be virtual.

contracts/governance/extensions/GovernorExtendedVoting.sol
```diff
@@ -0,0 +1,91 @@
...
+     */
+    function _castVote(
+        uint256 proposalId,
+        address account,
+        uint8 support,
+        string memory reason
+    ) internal virtual override returns (uint256) {
+        uint256 result = super._castVote(proposalId, account, support, reason);
+
+        Timers.BlockNumber storage extension = _extendedVoteEnd[proposalId];
```
frangio, line 51: We're using three different names for essentially the same concept: extension, extended vote end, extended deadline.

I wish we could find a more unified naming scheme.

Amxx: I tried renaming variables and contract to make everything clearer. Let me know what you think

byterose:
Very good work!

Lucid36 (approved):

Lucid36 (dismissed):

Lucid36 (approved):

Lucid36 (approved):

PR #2975, opened 2021-11-15
Fixed confusing require() string

robriks:
Line 332 require() error string:
 "... transfer of token that is not own"
is confusing but easily fixed to:
"... transfer of token that is not owned by 'from' "

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Thanks for the suggestion @robriks. I agree the current string is quite confusing...

One of the constraints we have is that we want these strings to be as short as possible because they add to deployment cost.

I also don't think it's a good idea to include the `'from'` name in the string as it's not clear it refers to the argument.

I want to propose `ERC721: transfer from incorrect owner`. What do you think?

robriks:
That sounds perfect! Very clear and concise, fixing the confusion.

Happy holidays!

frangio (approved):

Issue #2981, opened 2021-11-21
Add Function `Math.abs`


barakman:
**🧐 Motivation**
A function which returns the absolute (and obviously unsigned) value of a signed value.

**📝 Details**

Here is the full work (as I am not authorized to post a Pull Request on your repository):

**File `Math.sol`:**
```
    /**
     * @dev Returns the absolute unsigned value of a signed value.
     */
    function abs(int256 n) internal pure returns (uint256) {
        unchecked {
            // must be unchecked in order to support `n = type(int256).min`
            return uint256(n >= 0 ? n : -n);
        }
    }
```
**File `MathMock.sol`:**
```
    function abs(int256 n) public pure returns (uint256) {
        return Math.abs(n);
    }
```
**File `Math.test.js`:**
```
const { MAX_UINT256, MAX_INT256, MIN_INT256 } = constants;
...
  describe('abs', function () {
    for (const n of [MIN_INT256, MIN_INT256.addn(1), new BN('-1'), new BN('0'), new BN('1'), MAX_INT256.subn(1), MAX_INT256]) {
      it(`correctly computes the absolute value of ${n}`, async function () {
        expect(await this.math.abs(n)).to.be.bignumber.equal(n.abs());
      });
    }
  });
...
```

Amxx:
Hello @barakman and thank you for this request.

I'm not puzzled about what the return type should be. On the one end returning `int256` would allow doing something like `int256 y = -x.abs();` but on the other hand it would causes overflow when doing `type(int256).min.abs()`

For the record, I'm not sure you implementation would work with `type(int256).min`, this would have to be tested, or documented.

barakman:
@Amxx:
1. It doesn't make sense to return an `int256` for a value which is **always non-negative**.
2. It is not even possible, because `abs(-2**255) == `2**255`, which cannot be represented with `int256`.
3. Everything is tested, including `abs(type(int256).min)`; please have a look at the (JS) code that I posted.

Thanks

barakman:
@Amxx: IMHO, function `abs` should not be used for signed arithmetic **by its very definition**.

With regards to the specific example that you've given:
```
int256 y = -x.abs();
```
The correct thing to do would be:
```
using SafeCast for uint256;
int256 y = -x.abs().toInt256();
```
Which will revert for `x == abs(type(int256).min)` and work correctly for all other values.

[...]

PR #2984, opened 2021-11-23
A function which returns the absolute value of a signed value

barakman:
Following feature request #2981, adding a function which returns the absolute (and obviously unsigned) value of a signed value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * `GovernorTimelockControl`: improve the `state()` function to have it reflect cases where a proposal has been canceled directly on the timelock. ([#2977](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2977))
+* `Math`: add a `abs(int256)` function that returns the unsigned, absolute, value of a signed value. ([#2984](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2984))
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


Amxx:
Thank you @barakman for starting a PR.

This will need some documentation and a Changelog entry.

barakman:
> Thank you @barakman for starting a PR.
> 
> This will need some documentation and a Changelog entry.

I can update the CHANGE log, but what other documentation do you want (and where)?

Amxx:
Documentation is in `docs/modules/ROOT/pages/utilities.adoc` ... but I am now realizing that the Math part was not upgraded when we moved to 0.8.0 and safemath became native. Lets not touch that, we'll have to rework it later.

Just a changelog entry, fixing the lint test, and its good to merge !

barakman:
> Documentation is in `docs/modules/ROOT/pages/utilities.adoc` ... but I am now realizing that the Math part was not upgraded when we moved to 0.8.0 and safemath became native. Lets not touch that, we'll have to rework it later.
> 
> Just a changelog entry, fixing the lint test, and its good to merge !

BTW, I'm 100% sure that `Math` is the most adequate library for this function, but it seemed more adequate than `SafeMath` and `SignedSafeMath`.

I guess that `SignedMath` would be the most adequate (if you such this library).

Amxx:
The utilities.adoc page is about everything under `contract/utils`. 

I agree `.abs()` should be in Math.sol. `SafeMath` and `SignedSafeMath` are about overflow protection, while `Math` is about math primitives, which I believe `abs()` to be one.

Amxx (approved):

frangio:
@Amxx When we merged this we didn't have `SignedMath`, but now we do. Should we move this function over there? (It hasn't been released yet so this is not a breaking change.)

PR #2994, opened 2021-12-02
Address: explain dangers of isContract

Ivshti:
This explains in detail why `isContract` should be avoided a security check.

I've been seeing  more and more teams in crypto implement checks that depend on `isContract` to guard against flash loan attacks, reentrancies, among other things

Obviously this breaks composability and completely breaks support for smart wallets like Gnosis Safe, Argent, Ambire and others.

I believe the direct way in this is written is necessary, because figuring out that you should NOT do this from the current disclaimer requires some thought. Seeing that this is becoming a pattern (pppular projects like Anyswap do this and say it protects them from flashloans), I think stating it directly is the best way to go.

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -15,6 +15,12 @@ library Address {
      * It is unsafe to assume that an address for which this function returns
      * false is an externally-owned account (EOA) and not a contract.
      *
+     * **You cannot rely on `isContract` to protect against flash loan attacks!**
+     * Using `isContract` to intentionally prevent calls from contracts should be avoided, as it:
+     * 1) breaks composability
+     * 2) breaks support for smart wallets like Gnosis Safe where every user is a contract
+     * 3) can be circumvented by calling from a contract constructor
+     *
      * Among others, `isContract` will return false for the following
      * types of addresses:
      *


Amxx:
Hello @Ivshti and thank you for this PR

While I agree with you warning, I think it should be rephrased from

> You cannot rely on `isContract` to protect against flash loan attacks!

to

> You shouldn't rely on `isContract` to protect against flash loan attacks!

Would you agree with that formulation?

Ivshti:
@Amxx yeah, I actually think it's better that way

frangio:
@Amxx Why though? As far as I can tell `isContract` is not an effective protection against flash loans, since it can be circumvented. So the message is that you _can't_ rely on it.

contracts/utils/Address.sol
```diff
@@ -15,6 +15,12 @@ library Address {
      * It is unsafe to assume that an address for which this function returns
      * false is an externally-owned account (EOA) and not a contract.
      *
+     * **You cannot rely on `isContract` to protect against flash loan attacks!**
+     * Using `isContract` to intentionally prevent calls from contracts should be avoided, as it:
+     * 1) breaks composability
+     * 2) breaks support for smart wallets like Gnosis Safe where every user is a contract
+     * 3) can be circumvented by calling from a contract constructor
```
frangio, line 22: A comment about formatting: not sure if you meant this to be a numbered list in the docs site, but in the way it's written now it gets rendered inline in a paragraph ([see](https://deploy-preview-2994--openzeppelin-contracts-docs.netlify.app/contracts/4.x/api/utils#Address-isContract-address-)).

In order to make it a numbered list you need to add a newline after the colon and then use `1.` rather than `1)`.

Amxx:
While they can be circumvented, that is not very easy to do, and I'd argue that even if that was an effective solution, you should still not want to use that for the other reasons.

frangio:
Merged into https://github.com/OpenZeppelin/openzeppelin-contracts/commit/04109f8bee373de4c7522f9ca5df7886fdc07bf3 with slightly different wording.

Thank you @Ivshti!

PR #2996, opened 2021-12-03
Reduce revert reason size in SafeCast

### LINKED ISSUE #2590: Reduce Error Message size in SafeCast.sol
**🧐 Motivation**
SafeCast error messages are very large (even throwing a linter warning on default solhint settings.

Because Solidity is an error message hog, and I don't believe it does error message mapping yet, best to reduce the error message size to something more reasonable.

** Proposed **

`SafeCast: value doesn\'t fit in 128 bits`

to:

`SafeCast: value > 128 bits`

frangio:
Fixes #2590 

Posting as a first draft, let me know if you like these or think others would be clearer.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/math/SafeCast.sol
+++ b/contracts/utils/math/SafeCast.sol
@@ -30,7 +30,7 @@ library SafeCast {
      * - input must fit into 224 bits
      */
     function toUint224(uint256 value) internal pure returns (uint224) {
-        require(value <= type(uint224).max, "SafeCast: value doesn't fit in 224 bits");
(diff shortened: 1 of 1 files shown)


Amxx:
Don't we want custom errors?

frangio:
Yes, but Web3.js/Ganache still doesn't support them... In the meantime we can reduce the size.

Amxx:
This is deprecated by the new procedural generation.

PR #2998, opened 2021-12-04
Improve the calculation speed of function ceilDiv in library Math

gzliudan:
`(a > 0) ? ((a - 1) / b + 1) : 0` should be faster than `a / b + (a % b == 0 ? 0 : 1)` because of not use `a % b`

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -38,7 +38,7 @@ library Math {
      */
     function ceilDiv(uint256 a, uint256 b) internal pure returns (uint256) {
         // (a + b - 1) / b can overflow on addition, so we distribute.
-        return a / b + (a % b == 0 ? 0 : 1);
+        return (a > 0) ? ((a - 1) / b + 1): 0;
     }
 
     /**


Amxx:
Hello @gzliudan 

Please note that a modulo cost only 5 gas, just the same as a multiplication or a division.

If I try to recap the cost for each version:

- Curent:
  - division: 5 gas
  - addition: 3 gas
  - modulo:   5 gas
  - isZero:   3 gas
  → 16 + cost of ternary

- Proposal:
  - gt: 3 gas 
  - substraction: 3 gas
  - division: 5 gas
  - addition: 3 gas
  → 14 + cost of ternary 
  
  It's almost exactly the same ... except that we should not forget stack operations. isZero is an opcode, but your gt requires pushing 0 on the stack (that is more gas for your proposal 3 gas)
  
  However, those numbers don't account for the eventual swap (3 gas each)
  
  My conclusion: 
  - It is not obvious that your proposal is cheaper (we would have to test that further)
  - If it is cheaper, the change would really be marginal
  - I believe the current version is easier to understand, but that is very personal
 
In such a situation I think we should use whatever is easier to read/understand. @frangio do you have any preference ?

frangio:
Agree with all of @Amxx's points, and I also think the current version is more readable.

PR #3003, opened 2021-12-06
Allow to override usage of ERC20 storage

k06a:
Allow inherited smart contract to utilize 1 global storage slot and 1 storage slot per user

#### Example

```solidity
contract MyPool is ERC20 {
    function balanceOf(address account) public view override returns(uint256) {
        return _shrinkBits(super.balanceOf(account));
    }

    function _setBalance(address account, uint256 amount) internal override returns(uint256) {
        require(amount < 0x00000000FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF);
        return (super.balanceOf(account) & 0xFFFFFFFF00000000000000000000000000000000000000000000000000000000) | _shrinkBits(amount);
    }

    function _shrinkBits(uint256 amount) internal view returns(uint256) {
        return amount & 0x00000000FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
    }
}
```

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -228,12 +228,12 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
 
         _beforeTokenTransfer(sender, recipient, amount);
 
-        uint256 senderBalance = _balances[sender];
+        uint256 senderBalance = balanceOf(sender);
         require(senderBalance >= amount, "ERC20: transfer amount exceeds balance");
         unchecked {
-            _balances[sender] = senderBalance - amount;
+            _setBalance(sender, senderBalance - amount);
         }
-        _balances[recipient] += amount;
+        _setBalance(recipient, balanceOf(recipient) + amount);
 
         emit Transfer(sender, recipient, amount);
 
@@ -254,8 +254,8 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
 
         _beforeTokenTransfer(address(0), account, amount);
 
-        _totalSupply += amount;
-        _balances[account] += amount;
+        _setTotalSupply(totalSupply() + amount);
+        _setBalance(account, balanceOf(account) + amount);
         emit Transfer(address(0), account, amount);
 
         _afterTokenTransfer(address(0), account, amount);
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @k06a 

This is something we considered in the past, particularly when we made all view functions virtual. Making `balanceOf()` virtual was a huge concern because we don't exactly know what level of consistency we can/should guarantee if someone ever decides to override it. 

For the record, we think that is very dangerous, and often a bad idea to override such function.

We did consider having virtual getters and setters, and not relying on storage anywhere except for these setters and getters. One of the reason that decided against that is that:

- we didn't really see any reasonable use cases
- we are really concerned that people using these setters would cause some devs to bypass the `Transfer` events, which would cause significant consistency issues between the on-chain data and off-chain aggregators.

Can you tell us more about your use case?

It looks like you want to put additional data in the balance slot of users. While it makes sense from a storage-effectiveness point of view, I think it really makes the code more complex, more susceptible to bugs, and less readable. This is not usually the direction we go for.

k06a:
@Amxx thanks for the detailed response. We are looking into a possibility to store few more bytes per user in a relatively cheap place. But also trying to avoid reimplementing ERC20 for sure!

Amxx:
I'm really puzzled.

On one hand, I see how a `_setBalance` could be useful to your usecase. If done carefully it can indeed be super powerful.

But I'm also thinking of all the other less experienced users that would possibly use that function to mint / burn without emitting the proper events ... which is something we REALLY don't want.

@frangio, any opinion on that ?

k06a:
I have one more idea of how you can expose ERC20 behavior in 2 ways without code duplication: same as now and for more sophisticated usage. I will show an example soon.

k06a:
@Amxx please take a look: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3037

frangio:
@k06a @Amxx Let's move this discussion to a dedicated issue and think about other implementation strategies: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3038

PR #3006, opened 2021-12-07
Update initializer modifier to prevent reentrancy during initialization

Amxx:
This issue was identified and reported by @chaitinblockchain.

It is currently possible for `initializer()`-protected functions to be executed twice, if this happens in the same transaction. For this to happen, either one call has to be a subcall the other, or both call have to be subcalls of a common `initializer()`-protected function. 

This can particularly be dangerous is the initialization is **not** part of the proxy construction, and reentrancy is possible by executing an external call to an untrusted address.

This PR is a move toward fixing this potential issue, by forbidding `initializer()`-protected function to be nested when the contract is already constructed and re-entrance is possible.

**⚠ This change is breaking in some cases. ⚠**

This change require changes to the upgradeable contracts:
- During the transpilation process, the `__XXX_init` function must use the new `onlyInitializing()` modifier instead of `initializer()`

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
looking at the old mocks, and what they have become, I really think it was better before ... despite re-entrancy potential :/

contracts/proxy/utils/Initializable.sol
```diff
@@ -41,11 +43,21 @@ abstract contract Initializable {
      */
     bool private _initializing;
 
+    /**
+     * @dev Modifier to protect (internal) functions that should only be executed as subcalls of an initializing
+     * function (protected with the {initializer} modifier). This modifier is designed to be used in upgradeable
+     * contracts for the constructor-replacing internal functions.
+     */
+    modifier onlyInitializing() {
```
frangio, line 51: The goal is that we will update our Upgradeable Contracts to use `onlyInitializing` in place of `initializer`?

CHANGELOG.md
```diff
@@ -2,11 +2,28 @@
...
+ * `Math`: add a `abs(int256)` method that returns the unsigned absolute value of a signed value. ([#2984](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2984))
+ * Preset contracts are now deprecated in favor of [Contracts Wizard](https://wizard.openzeppelin.com). ([#2986](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2986))
  * `Governor`: add a relay function to help recover assets sent to a governor that is not its own executor (e.g. when using a timelock). ([#2926](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2926))
  * `GovernorExtendedVoting`: add new module to ensure a minimum voting duration is available after the quorum is reached. ([#2973](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2973))
+ * `Initializable`: change the existing `initializer` modifier and add a new `onlyInitializing` modifier to prevent reentrancy risk. ([#3006](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3006))
+
+### Breaking changes
```
frangio, line 12: Please review this section.

contracts/proxy/utils/Initializable.sol
Amxx: Yes, I think its essential to support minimal proxy, and for that we need to use `onlyInitializing`

CHANGELOG.md
Amxx: LGTM

Amxx: LGTM

frangio (approved):

slendermaan:
Hello everyone, I'm sorry to bother you. I'm a little concerned about whether this issue will be included in the Security advisories section of the openzepplin library.

frangio:
Hi @slendermaan yes once we release. We had to postpone the release until tomorrow.

hacker-DOM:
> It is currently possible for `initializer()`-protected functions to be executed twice, if this happens in the same transaction. For this to happen, either one call has to be a subcall the other, or both call have to be subcalls of a common `initializer()`-protected function.
> 
> This can particularly be dangerous is the initialization is **not** part of the proxy construction, and reentrancy is possible by executing an external call to an untrusted address.

Hey folks, is there any chance you could provide a minimal example for the bug? I'm trying very hard but the English explanation is so ambiguous it's hard to come up with an example. Thanks!

hacker-DOM:
I tried to parse the text again, and it seems to me like the situation you are talking about is something like:

```solidity
contract D {
    bool private _initializing;
    bool private _initialized;

    modifier initializer() {
        require(_initializing || !_initialized, "Initializable: contract is already initialized");

        bool isTopLevelCall = !_initializing;
        if (isTopLevelCall) {
            _initializing = true;
            _initialized = true;
        }
        _;
        if (isTopLevelCall) {
            _initializing = false;
        }
    }

    function init1() public initializer {
        init2();
        init2(); // <-- doesn't revert
    }

    function init2() public initializer {}
}
```

Would this be the correct way to do it now? It still doesn't revert.

```solidity
contract E {
    bool private _initializing;
    bool private _initialized;

    modifier initializer() {
        require(_initializing ? _isConstructor() : !_initialized, "Initializable: contract is already initialized");

        bool isTopLevelCall = !_initializing;
        if (isTopLevelCall) {
            _initializing = true;
            _initialized = true;
        }
        _;
        if (isTopLevelCall) {
            _initializing = false;
        }
    }

    function init1() public initializer {
        init2();
        init2(); // <-- still doesn't revert
    }

    function init2() public onlyInitializing {} // <-- changed this per documentation

    modifier onlyInitializing() {
        require(_initializing, "Initializable: contract is not initializing");
        _;
    }

    function _isConstructor() private view returns (bool) {
        return !isContract(address(this));
    }
    
    function isContract(address a) public view returns (bool) {
        return a.code.length > 0;
    }
}
```

hacker-DOM:
Ok I get it nvm.

paulrberg:
Hey guys, I stumbled upon this PR after reading the [comments](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/20380ee699406211ae88c4edd11b0af87c8abdd6/contracts/proxy/utils/Initializable.sol#L50-L52) added in the latest implementation of `Initializable.sol`. I recently upgraded to the latest version of `@openzeppelin/contracts-upgradeable` and that has made some of our tests to fail - it looks like we were indeed executing two initializers, one of which was a subcall of the other.

I'm trying to wrap my head around this, though:

> This can particularly be dangerous is the initialization is **not** part of the proxy construction, and reentrancy is possible by executing an external call to an untrusted address.

@Amxx, could you share an example for when the previous behavior would have been problematic?

Amxx:
Let's say you have:
- A proxy (P)
- A logic contract (L) that includes an `initialize` function. This function is protected by the `initializer` modifier
- An external contract (E)

Now the flow:
- Step 1: P is constructed, and set to point to L, but not initialized.
- Step 2: We call the initializer in P
  - 2.A: this is expected to do the initialization, and set the and register that it has been done, so it cannot be initialized
  - 2.B: (the issue is here) the initialization might call (E). We are still in the initialization phase.
  - 2.C: E can call `initialize` on P, with different settings. We are still in the initialization phase.
  - 2.D: `initializer` accepts that call, because we are in the initialization phase.
  - 2.E: P possibly gets reinitialized, with E controlled parameters.
  
By spliting the `initializer` modifier (for external functions) and the `onlyInitializating` (for internal functions) we make sure that the external initializer call can still trigger the internal ones, but 2.D will throw.

If the initializer is done during construction, then we don't have this threat because E would not be able to call P (because P is not yet constructed)

paulrberg:
Very interesting. But I think this was low-risk, wasn't it?

Firstly, most upgradeable proxy deployers take care of initialization atomically - I think this is what [hardhat-upgrades](https://docs.openzeppelin.com/upgrades-plugins/1.x/hardhat-upgrades) does.

Secondly, I haven't seen many contracts making calls to external contracts during construction. The vast majority of contracts just call some internal function inherited from other contracts.

paulrberg:
> Have you been able to fix your issue? You should probably use `onlyInitializing` instead of `initializer`.

Yeah that's exactly what I did. Thanks for confirming!

Rassska:
> 

But how does initialization invokes E? Why it should make a call into some external contract which is not under our control?

frangio:
>Why it should make a call into some external contract which is not under our control?

This may be part of the logic of the initializer body for some reason. To be clear: the `initializer` modifier would never initiate this external call on its own.

PR #3011, opened 2021-12-10
Added `normalizeUpperASCII()` and `normalizeLowerASCII()` functions to "\utils\Strings.sol"

CollinMcKinney:
Adds `normalizeUpperASCII()` and `normalizeLowerASCII()` functions to Strings.sol

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts v4.4.0 (utils/Strings.sol)
+// OpenZeppelin Contracts v4.4.1 (utils/Strings.sol)
 
 pragma solidity ^0.8.0;
 
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -75,6 +75,7 @@ library Strings {
             require(buffer[i] & 0x80 == 0, "Basic ASCII string required."); // 0x80-bit == non-ASCII
             if (buffer[i] >= "a" && buffer[i] <= "z") {
                 buffer[i] ^= 0x20; // removes "lower-case bit"
+                assert(buffer[i] >= "A" && buffer[i] <= "Z");
             }
         }
         return string(buffer);
@@ -90,6 +91,7 @@ library Strings {
             require(buffer[i] & 0x80 == 0, "Basic ASCII string required."); // 0x80-bit == non-ASCII
             if (buffer[i] >= "A" && buffer[i] <= "Z") {
                 buffer[i] |= 0x20; // adds "lower-case bit"
+                assert(buffer[i] >= "a" && buffer[i] <= "z");
             }
         }
         return string(buffer);
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -79,9 +79,7 @@ library Strings {
             if (buffer[i] >= "a" && buffer[i] <= "z") {
                 // remove the lower-case bit. 0xDF is the bit-wise complement ('~') of
                 // the lower-case bit mask (0x20).
-                buffer[i] &= 0xDF;
-                // assert invariance in conversion to upper-case.
-                assert(buffer[i] >= "A" && buffer[i] <= "Z");
+                buffer[i] &= ~bytes1(0x20);
             }
         }
         // return the upper-case normalized, basic-ASCII string result.
@@ -102,8 +100,6 @@ library Strings {
             if (buffer[i] >= "A" && buffer[i] <= "Z") {
                 // set the lower-case bit (0x20).
                 buffer[i] |= 0x20;
-                // assert invariance in conversion to lower-case.
-                assert((buffer[i] >= "a") && (buffer[i] <= "z"));
             }
         }
         // return the lower-case normalized, basic-ASCII string result.
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -72,7 +72,7 @@ library Strings {
     function normalizeUpperASCII(string calldata value) internal pure returns (string memory) {
         bytes memory buffer = bytes(value);
         for (uint256 i = 0; i < buffer.length; i++) {
-            require(buffer[i] & 0x80 == 0); // If the 1000_0000 bit is set in a char, string is non-ASCII
+            require(buffer[i] & 0x80 == 0, "Basic ASCII string required."); // 0x80-bit == non-ASCII
             if (buffer[i] >= "a" && buffer[i] <= "z") {
                 buffer[i] ^= 0x20; // removes "lower-case bit"
             }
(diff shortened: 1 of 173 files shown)


Amxx:
Hello @CollinMcKinney and that you for this PR. 

While I'm not seeing immediate use-cases to these function, this could be a nice addition to the library! 
I would however refactor it a bit to avoid bytes1 / uint8 conversion.

```
    function toUpper(string calldata value) external pure returns (string memory) {
        bytes memory buffer = bytes(value);
        for (uint256 i = 0; i < buffer.length; i++) {
            if (buffer[i] >= "a" && buffer[i] <= "z") {
                buffer[i] ^= 0x20;
            }
        }
        return string(buffer);
    }

    function toLower(string calldata value) external pure returns (string memory) {
        bytes memory buffer = bytes(value);
        for (uint256 i = 0; i < buffer.length; i++) {
            if (buffer[i] >= "A" && buffer[i] <= "Z") {
                buffer[i] |= 0x20;
            }
        }
        return string(buffer);
    }
    ```

**Comment by CollinMcKinney** [comment]
> Hello @CollinMcKinney and that you for this PR.
> 
> While I'm not seeing immediate use-cases to these function, this could be a nice addition to the library! I would however refactor it a bit to avoid bytes1 / uint8 conversion.
> 
> ```
>     function toUpper(string calldata value) external pure returns (string memory) {
>         bytes memory buffer = bytes(value);
>         for (uint256 i = 0; i < buffer.length; ++i) {
>             if (buffer[i] >= "a" && buffer[i] <= "z") {
>                 buffer[i] ^= 0x20;
>             }
>         }
>         return string(buffer);
>     }
> 
>     function toLower(string calldata value) external pure returns (string memory) {
>         bytes memory buffer = bytes(value);
>         for (uint256 i = 0; i < buffer.length; ++i) {
>             if (buffer[i] >= "A" && buffer[i] <= "Z") {
>                 buffer[i] |= 0x20;
>             }
>         }
>         return string(buffer);
>     }
> ```

Thanks for this suggestion. I've added it into the pull-request.

**Comment by CollinMcKinney** [comment]
I actually can't think of many specific use-cases where these functions *should* be called through a contract. In most applications that I can theorize, ToUpper and ToLower functions should be run by the client/front-end. Maybe there are some cases where the owner may want to deploy a contract, have several third parties write their own front-ends, and still have control over string-casing on the owner's end.

**Comment by frangio** [comment]
I assume this does not work generally for UTF8 strings so it wouldn't be reliable for case insensitive operations. Fully Unicode lowercasing is probably not a good fit for a smart contract though. I agree that this should be done at the front-end as much as possible.

**Comment by CollinMcKinney** [comment]
> I assume this does not work generally for UTF8 strings so it wouldn't be reliable for case insensitive operations. Fully Unicode lowercasing is probably not a good fit for a smart contract though. I agree that this should be done at the front-end as much as possible.

It should work for UTF8 just fine. When a single byte, 0-127 values decode the same as ASCII. When it's a multi-byte char, each byte has a 1 in the highest bit, which means each byte is 128 or greater. 128 is outside of the range that A-z encode from, so there will be no collisions.

**Comment by frangio** [comment]
Sorry I wasn't clear enough. What I meant is that it only works for ASCII characters and not for other Unicode characters that may be present in the string, for example accented characters, which wouldn't be converted from lowercase to uppercase and I think this would be an unexpected result.

**Comment by CollinMcKinney** [comment]
> Sorry I wasn't clear enough. What I meant is that it only works for ASCII characters and not for other Unicode characters that may be present in the string, for example accented characters, which wouldn't be converted from lowercase to uppercase and I think this would be an unexpected result.

Oh ok yeah you're right, this wouldn't work with capitalization of the Turkish i's for example.

File: contracts/utils/Strings.sol [hunk 1]
```diff
@@ -64,4 +64,36 @@ library Strings {
...
+    /**
+     * @dev Converts each character ('a' to 'z') of a `string` to it's upper-case form ('A' to 'Z').
+     * Requires parameter string `value` to be comprised purely of basic-ASCII characters.
+     */
+    function normalizeUpperASCII(string calldata value) internal pure returns (string memory) {
+        bytes memory buffer = bytes(value);
+        for (uint256 i = 0; i < buffer.length; i++) {
+            require(buffer[i] & 0x80 == 0, "Basic ASCII string required."); // 0x80-bit on is non-ASCII.
+            if (buffer[i] >= "a" && buffer[i] <= "z") { // only work on lower-case characters.
+                buffer[i] &= 0x1F; // mask-off the lower-case bit.
```
Amxx, line 77: This is not good. What you want to do is turn of the 6th bit (that is work 32)
Is is done by doing `^= 0x20`

`&= 0x1F` is a completely different operation that makes no sens here.

contracts/utils/Strings.sol
```diff
@@ -64,4 +64,36 @@ library Strings {
...
+                assert(buffer[i] >= "A" && buffer[i] <= "Z"); // assert invariance in change to uppercase.
```
Amxx, line 78: This is guaranteed by math ... why check it? it brings nothing and adds bytecode.

CollinMcKinney: It's to protect against (potential) bugs in the math. Afaik, this is how assert() is meant to be used - at the end to validate invariance in code / results. This helps guarantee the function always returns a normalized string. I think since the scope of this function is to normalize strings, there's some value in having this guard in place. I don't know how much bytecode it adds, I assumed it wouldn't be too much.

contracts/utils/Strings.sol
CollinMcKinney: ^= toggles bits. If somehow an uppercase later made it through the separate lines of code above, it would produce a lowercase character instead of uppercase.

&= allows us to take all the bits, *except* the casing bit. Though I do agree &= 0x1F is incorrect, as that also relies on the `require()` above it to exist. I think `&= 0xDF` would be more reasonable.

Amxx: How do you expect an uppercase letter to make it through here? Under what circumstances would an uppercase pass `buffer[i] >= "a" && buffer[i] <= "z"` ???

contracts/utils/Strings.sol
Amxx: I do agree with your philosophy behind the usage of assert, I just don't think it applies here. IMO the math is simple enough that we can be certain this will never be triggered... And if there was still a doubt, I would rather use formal verification than shipping extra bytecode.

contracts/utils/Strings.sol
Amxx: `&= 0xDF` sounds good !

contracts/utils/Strings.sol
CollinMcKinney: I'm not opposed to removing the assert() if that's the consensus.

contracts/utils/Strings.sol
CollinMcKinney: > How do you expect an uppercase letter to make it through here? Under what circumstances would an uppercase pass `buffer[i] >= "a" && buffer[i] <= "z"` ???

Under possible future circumstances where that line of code becomes bugged.

CollinMcKinney: > `&= 0xDF` sounds good !

It's now been changed to `&= ~bytes1(0x20)` (which equates to 0xDF).

frangio:
Apologies for the back and forth on this @CollinMcKinney, but after some discussion we're not willing to merge this without a clear use case, and it seems for now we all agree that upper/lowercasing should be done in the frontend. Case insensitive string comparisons are potentially interesting, but concrete use cases are also not clear, and most importantly I would argue ASCII-only support is not good enough for case insensitivity, although for specific use cases it might be so again it goes back to missing concrete use cases that can guide development.

PR #3015, opened 2021-12-11
Add documentation for delegateBySig (#2927)

### LINKED ISSUE #2927: Document use of delegateBySig
We don't seem to have documentation for the format of the message that needs to be signed.

hack3r-0m:
Fixes #2927  

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

image

### DIFF
--- a/contracts/token/ERC20/extensions/ERC20Votes.sol
+++ b/contracts/token/ERC20/extensions/ERC20Votes.sol
@@ -131,6 +131,26 @@ abstract contract ERC20Votes is IVotes, ERC20Permit {
 
     /**
      * @dev Delegates votes from signer to `delegatee`
+     *
+     * Types for creating EIP712 signature(v, r, s) are as below
+     * ```
+     * {
+     *     EIP712Domain: [
+     *         { name: 'name', type: 'string' },
+     *         { name: 'chainId', type: 'uint256' },
+     *         { name: 'verifyingContract', type: 'address' },
+     *     ],
+     *     Delegation: [
+     *         { name: 'delegatee', type: 'address' },
+     *         { name: 'nonce', type: 'uint256' },
+     *         { name: 'expiry', type: 'uint256' }
+     *     ]
+     * }
+     * ```
+     * 
+     * `primaryType` is `Delegation`
+     * 
+     * reference: https://medium.com/compound-finance/delegation-and-voting-with-eip-712-signatures-a636c9dfec5e
      */
     function delegateBySig(
         address delegatee,


hack3r-0m:
is CHANGELOG entry required for documentation edit?

Amxx:
@hack3r-0m I don't think we need a changelog entry for that. We do however need to document that the domain separator includes a `version` field. Without that, your documentation is actually incorrect.

Maybe it would also be nice to add example of code to produce these signature.

hack3r-0m:
thanks @Amxx, i have documented `version`, regarding example, how comprehensive it should be? should it use `ethereum-js` utils or `ethers` or `web3`? similiar to https://gist.github.com/ajb413/a5466a61558c0897491c378ce6682ad0#file-delegate-sign-js with version in domain seperator should work?

Amxx:
In my opinion, the more the better.

I do believe that web3 doesn't support that natively.

I would definitely include ethers's `Wallet._signTypedData`

```javascript
const signature = await wallet._signTypedData({
    chainId: 5,
    name: 'Mytoken',
    verifyingContract: '0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC',
    version: '1.0.0',
}, {
    // No need for Domain definition with ethers
    Delegation: [
        { name: 'delegatee', type: 'address' },
        { name: 'nonce', type: 'uint256' },
        { name: 'expiry', type: 'uint256' },
    ],
}, {
    delegatee: <some address>,
    nonce: <some nonce>,
    deadline: <some deadline>,
});
```

I would also include the jsonrpc way
```javascript
const msgParams = JSON.stringify({
    domain: {
        chainId: 5,
        name: 'Mytoken',
        verifyingContract: '0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC',
        version: '1.0.0',
    },
    types: {
        EIP712Domain: [
             { name: 'name', type: 'string' },
             { name: 'version', type: 'string' },
             { name: 'chainId', type: 'uint256' },
             { name: 'verifyingContract', type: 'address' },
        ],
        Delegation: [
            { name: 'delegatee', type: 'address' },
            { name: 'nonce', type: 'uint256' },
            { name: 'expiry', type: 'uint256' },
        ],
    },
    message: {
        delegatee: <some address>,
        nonce: <some nonce>,
        deadline: <some deadline>,
    },
    primaryType: 'Delegation',
  });
const signature = await provider.send('eth_signTypedData_v4', [ from, msgParams ]);
```

hack3r-0m:
thanks, @Amxx, version should be `1.0.0` or `1` while adding examples into docs? 
since https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/draft-ERC20Permit.sol#L36 uses `1` while the one you mentioned uses `1.0.0`

[official eip](https://eips.ethereum.org/EIPS/eip-712#specification-of-the-eth_signtypeddata-json-rpc) too does have any clarification whether it should be a semantic version or not.

Amxx:
You are right, our code uses "1" ... then so should the examples !

frangio:
Thank you @hack3r-0m! I agree we should say in the documentation that `version` is hardcoded to `"1"`.

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -131,6 +131,30 @@ abstract contract ERC20Votes is IVotes, ERC20Permit {
...
+     *         { name: 'nonce', type: 'uint256' },
+     *         { name: 'expiry', type: 'uint256' }
+     *     ]
+     * }
+     * ```
+     *
+     * `name` and `version` should match values passed in contract's constructor.
+     *
+     * NOTE: https://medium.com/compound-finance/delegation-and-voting-with-eip-712-signatures-a636c9dfec5e[Compound's delegateBySig implementation] does not include `version` in `EIP712Domain`.
+     * However, the way OpenZeppelin's EIP712 implementation computes domain seperator, `version` is required.
```
frangio, line 155:
```suggestion
     * However, the way OpenZeppelin's EIP712 implementation computes domain separator, `version` is required.
```

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -131,6 +131,30 @@ abstract contract ERC20Votes is IVotes, ERC20Permit {
...
+     *
+     * `primaryType` is `Delegation`
```
frangio, line 157:
```suggestion
     * The message's primary type is `Delegation`.
```

hack3r-0m:
I have also replaced `deadline` from the above example with `expiry` to match the type's attribute.

image

frangio:
This is great content but it's too much to include in the Solidity file itself... I'm trying to figure out where else we can put it.

ernestognw:
Closing as it was never merged to avoid the huge amount of documentation in the contract itself. Today I don't think this is super relevant since it's managed by standard libraries.

PR #3016, opened 2021-12-12
Fix typo

RogerPodacter:


Amxx (approved):
thank you @RogerPodacter for making that PR

Issue #2620, opened 2021-03-29
Deprecate ERC777 implementation


k06a:
**🧐 Motivation**

There is an opinion that ERC777 is over-engineered and is a bad practice to follow. Moreover it introduces bad abstractions to rely on and requires very important checks to be implemented by every integrator.

Let's switch to EIP2612 (Permit) to deprecate dangerous infinite approve behavior and make it mainstream ASAP: https://github.com/OpenZeppelin/openzeppelin-contracts/contracts/token/ERC20/extensions/draft-ERC20Permit.sol

**📝 Details**

Do we need to collect list of issues? Starting here:

1. Whole idea of avoiding spam tokens by having hook function is wrong. It is not possible to protect from spam tokens because developers of these tokens will always modify their tokens to allow spamming – just ignore them.
2. Token fallback concept is wrong because the callback is being called from token smart contract and there is no way to verify who was the original `msg.sender`. There are so many possible ways to abuse `ERC777Receiver`. The only way to solve it I see – work with whitelisted tokens only – DeFi deserves better approach. Imagine you have DEX and wanna deposit token and do custom action (swap):

    ```solidity
    function tokensReceived(
        address operator,
        address from,
        address to,
        uint256 amount,
        bytes calldata userData,
        bytes calldata operatorData
    ) external override {
        address token = msg.sender;
        balances[token][operator] += amount;
        _performSwap(operator, operatorData); // <- `operator` is not trustworthy, due `msg.sender` can be malicious smart contract
    }

    function _performSwap(address user, bytes memory data) internal {
        // ...
    }
    ```

Amxx:
Hello @k06a,
As a developper working with ERC777 during hackathon, I can only agree that this standard creates a lot of frustration ... and can result in very gas-extensive transactions.

However, deprecating contract breaks backward compatibility. This is not something we would consider doing during a "minor" release. The latest major release was v4.0, less then a month ago. Thus, we wouldn't move with anything like that before a while.

Still, I'd love to build a case for/against ERC777 (and other contract that might not be relevant). This will help us taking decision whenever we fell like moving to the next major version.

rstormsf:
I definitely support this, so new solidity developers won't inherit this broken token design pattern.

k06a:
@Amxx thanks for your support, added 2 cases to the [OP](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2620#issue-843492356).

frangio:
[...]

PR #3017, opened 2021-12-12
ERC1363

Amxx:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3736

[ERC1363](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1363.md) is marked final

I think we should implement it. It would be an important move toward adoption of these mechanisms that can both improve the UX (only one tx needed instead of 2) and save a lot of gas during contract interaction (using approval added 2 sstore and 2 events: one when approval is set and one when transferFrom is called.)

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


Amxx:
AFAIK, we cannot go further with coverage...

frangio:
There are many EIPs for `transferAndCall` functionality. Why should we implement this one? I can see that the EIP is Final but it looks like there was zero discussion about it, or I can't find it. Has an implementation been audited and deployed to production?

Amxx:
IMO, this "transfer and call" mechanism is really interesting, and some projects could really benefit from it. In particular, it can be used to avoid the approval + transferFrom workflow, which costs 2 sstores and 2 Approval events.

AFAIK, there are not that many other options. 
- ["ERC223"](https://github.com/ethereum/EIPs/issues/223) is just an open issue (no proper merged ERC document) and is not backward compatible with ERC20.
- ["ERC667"](https://github.com/ethereum/EIPs/issues/677) is just an open issue (no proper merged ERC document).
- [ERC1363](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1363.md) is a properly formatted ERC, and is complete in terms of features (while remaining simple/clear).
 
An I missing other standardization efforts?

My understanding is that this feature can be used both 
- within a web3app, to have multiple contracts interact
- across apps, to improve interoperability.

The second point would be great in the long run, but I'd say my main fear is that projects would need the first point and will either not use it, for lack of knowledge of this pattern (making their workflow overly expensive/complex) or use a non-standard version which will work for them in the short run, but make point 2 much harder to achieve.

IMO, we need to push toward more standardization, and one way to do that is by providing a safe & effective implementation.

[Note: I've opened this discussion in sept 2020, and the discussion was not that rich :/](https://ethereum-magicians.org/t/about-677-and-other-functions-on-top-of-erc20/4589)

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,88 @@
...
+
+import "../ERC20.sol";
+import "../../../interfaces/IERC1363.sol";
+import "../../../interfaces/IERC1363Receiver.sol";
+import "../../../interfaces/IERC1363Spender.sol";
+import "../../../utils/introspection/ERC165.sol";
+
+abstract contract ERC1363 is IERC1363, ERC20, ERC165 {
+    function supportsInterface(bytes4 interfaceId) public view virtual override(IERC165, ERC165) returns (bool) {
+        return interfaceId == type(IERC1363).interfaceId || super.supportsInterface(interfaceId);
```
Amxx, line 12: Maybe also advertise IERC20 and IERC20Metadata

frangio:
I honestly don't see anything in the EIP that strongly indicates `transferAndCall` should revert when the destination is not a contract. @lukehutch You mentioned that the spec says transferring to EOAs is not allowed (https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2620#issuecomment-1156422782: "search for MUST implement") but I don't see it. It says "A contract that wants to accept token payments via `transferAndCall` or `transferFromAndCall` MUST implement the following interface" but this is only making a statement about contracts, it's not making a statement about EOAs.

What can be read as ambiguous is that the intention is to "execute code" on the receiver, and the wording that `transferAndCall` should transfer "and then call `onTransferReceived` on receiver". The ambiguity is that `CALL` at the EVM level succeeds for EOAs, but a "call" in Solidity fails if the target is not a contract.

I think the EIP is the ultimate source of truth. @vittominacori's [implementation](https://github.com/vittominacori/erc1363-payable-token) isn't linked in the EIP so arguably it's not even a "reference" implementation, and it doesn't claim to be so in the repository (says it's "an implementation").

All that said, I'm slightly leaning towards requiring the receiver to be a contract, because I think the general expectation for a "call" is that it will actually execute code. The fact that `CALL` silently succeeds when the target is not a contract is generally seen as a bad quirk of the EVM.

vittominacori:
@frangio my own implementation is not linked because of the reviewer said to remove external link to repo and test cases (also if I've seen many EIP referencing their own implementation).

However maybe I missed explicit reference to EOA address but it was clear for me to refer only contracts by saying that "contracts MUST implement..." and that `transferAndCall`, `transferFromAndCall` and `approveAndCall` method is a call to a method and not the `CALL`.

Will try to update the EIP as they say "A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications." in order to clarify this missing declaration.

lukehutch:
@frangio Exactly, the EIP does not mention EOAs at all, or even hint at what should be done for EOA recipients or spenders. And the statement that you found is what I was referring to. It is simply ambiguous about the EOA case (by omission), as I pointed out several times in the issue about deprecating ERC777.

Because it is ambiguous, implementations should not all decide differently about what should be done in this case. I think @Amxx did the right thing by asking @vittominacori what his intention was in this case, and he made that intention very clear in the comment that I quoted: he believes ERC20 APIs should be use to send to EOAs, since ERC1363 is a proper superset of ERC20 -- therefore, ERC1363 should revert if trying to send to EOAs. And I don't think the reference implementation is irrelevant, since it proves what @vittominacori's intent really was in the spec (even though the implementation is not linked from the spec -- it used to be, but had to be removed at the request of a reviewer).

Note that ERC1363's reference implementation refusing to send to EOAs is different from ERC777 and ERC4524, both of which allow sending to EOAs through their ERC20-superset APIs, by actual definition in the spec. I think there is room for a sender notification API that refuses to send to EOAs.

>  a "call" in Solidity fails if the target is not a contract

Correct if "call" is passed a non-empty ABI encoding of a function call. I don't think it matters that `transferAndCall` has a different connotation at the EVM level and the Solidity/Yul level, because almost nobody working in Solidity deals with EVM nuances. The Solidity "call" behavior should be the most relevant here.

Amxx:
In solidity, doing

`(bool success,) = target.call(anything)`

Will be true if target is an EOA. Just like sending à transaction with any data field to an EOA will not revert.

lukehutch:
Correct, to prevent sending to EOA you have to do

```
require(account.code.length > 0, "Can't send to EOA");
```

(Although that also prevents sending to contracts whose constructors have not finished executing)

frangio:
I meant external function calls like `target.foo()` revert if the target is not a contract.

lukehutch:
> I meant external function calls like target.foo() revert if the target is not a contract.

Not true, if the target implements a fallback function. This is actually a huge potential security problem that affects hundreds of millions of dollars locked up in deployed contracts.

https://media.dedaub.com/phantom-functions-and-the-billion-dollar-no-op-c56f062ae49f

This can even affect ERC777, because there is no return value for the sender and recipient notification interface.

ERC1363 and ERC4524 solve this by requiring their notification hooks to return their own function selector.

Amxx:
> > I meant external function calls like target.foo() revert if the target is not a contract.
> 
> Not true, if the target implements a fallback function. This is actually a huge potential security problem that affects hundreds of millions of dollars locked up in deployed contracts.

How can a "not a contract" have a fallback function?

When doing `target.foo()`, under the hood solidity does something like
```
require(target.isContract);
(bool success, bytes memory returndata) = target.call(abi.encodeCall(target.foo, ()));
require(success);
return abi.decode(returndata, <foo return type>);
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+    }
+
+    /**
+     * @dev Internal function to invoke {IERC1363Spender-onApprovalReceived} on a target address.
+     * The call is not executed if the target address is not a contract.
+     */
+    function _checkOnApprovalReceived(
+        address owner,
```
vittominacori, line 124: Why passing owner here and not using msg.sender?

vittominacori:
Sorry @Amxx you stopped working here and removed it from milestone because you don't like where the discussion pointed?

So as the discussion said that it must revert on EOA it is not useful to be included because you don't agree?

I will open my [PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3487) again if the EIP will be updated with the EOA revert statement as you leave it incomplete.

lukehutch:
@vittominacori What is the chance that the EIP PR will be accepted? I'm not familiar with how difficult a process it is to get errata accepted.

vittominacori:
@lukehutch 

Reviewer is asking to another one if we should remove the external links. Nothing about the errata right now but I think it should be accepted because it is really simple https://github.com/ethereum/EIPs/pull/5167

On guidelines they say: *"Final - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications."*. This should be the case.

Amxx:
> Sorry @Amxx you stopped working here and removed it from milestone because you don't like where the discussion pointed?

One of the important points of OZ is that we try to not be opinionated. Releasing contracts as part of our library can lead to adoption in the sense that we may be seen as advocates/proponents of their usage. I believe this is partly why ERC1363 was not added earlier: the demand is not there (I don't think anyone ever asked us to provide it).

I personally believed that these features would benefit the ecosystem and that as ERC777 is losing traction, we could play a role in pushing for more adoption of ERC1363. This is an opinionated move, that I try to motivate by the fact I believe it would be for the better.

The last few days showed me that there is definitely no consensus on what behavior people needed, and what standard to follow. I now believe that promoting ERC1363 (and the receiver/spender interfaces) might not be in the best interest of the community given that we may see another standard being proposed, with incompatible receiver interfaces.

I'm not abandoning the idea of releasing it, but I want to be careful about how this evolves. Given that we are in the middle of 4.7's RC, the next release (4.8) is 1~1.5 months away. There is still time for the dust to settle, and this getting into 4.8. However, if there is still confusion, we should not consider this to be so important that it delays 4.8.

frangio:
I really don't think the behavior of the function with EOA receivers is such an important issue.

The important bit in this EIP is the addition of a callback for contracts. I discussed with @Amxx earlier today and he's now come to this as well.

I personally kind of appreciate the simplicity of being sure that the callback always runs.

In summary: we will merge an implementation of EIP-1363 that reverts on `transferAndCall` to an EOA receiver.

contracts/token/ERC20/extensions/ERC1363.sol
Amxx: It is in our best practices that the internal and private function should be context-free, and that the public/external interface should extract the context and pass it through arguments.

Amxx:
Note the test were broken by master going from 0.8.9 to 0.8.13.

I noticed that the behavior changes from 0.8.9 to 0.8.10:

Considering this code:
```
        try IERC1363Receiver(to).onTransferReceived(_msgSender(), from, value, data) returns (bytes4 retval) {
            return retval == IERC1363Receiver.onTransferReceived.selector;
            if (reason.length == 0) {
                revert("ERC1363: transfer to non ERC1363Receiver implementer");
            } else {
                assembly {
                    revert(add(32, reason), mload(reason))
                }
            }
        }
```

- with 0.8.9, if the `to` had no code, the call reverted with `function call to a non-contract account`
- with 0.8.10, it looks like the call is performed, and the returns data is not the right length, leading to `function returned an unexpected amount of data`

Amxx:
It's part of 0.8.10 changelog:
> - Code Generator: Skip existence check for external contract if return data is expected. In this case, the ABI decoder will revert if the contract does not exist.

contracts/token/ERC20/extensions/ERC1363.sol
vittominacori: Ok, so should `_msgSender()` also to be removed from there?
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/eadcad5e6945462b3e1d056ec770353207ed57c5/contracts/token/ERC20/extensions/ERC1363.sol#L105

Amxx: yes

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,143 @@
...
+     */
+    function _checkOnTransferReceived(
+        address operator,
+        address from,
+        address to,
+        uint256 value,
+        bytes memory data
+    ) private returns (bool) {
+        try IERC1363Receiver(to).onTransferReceived(operator, from, value, data) returns (bytes4 retval) {
+            return retval == IERC1363Receiver.onTransferReceived.selector;
```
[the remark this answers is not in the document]
Amxx: I copied the approach we use in the ERC721 checks, for consistency through the code base, but I guess putting the require here would be ok

frangio:
In #3525 it was pointed out that the interface should be moved from the interfaces directory and be located next to the implementation. What do you think?

vittominacori:
Yes, so in order to be similar to any other OZ interface in #3525 I moved them into token folder and then linked in `interfaces` folder like the ERC20. 
I see that here the interface is all written into the interfaces folder 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/08dd234f57f0372260422ac85281a21d70adac50/contracts/interfaces/IERC1363.sol

Also I created different mock files instead of having all into a single file like here https://github.com/OpenZeppelin/openzeppelin-contracts/blob/08dd234f57f0372260422ac85281a21d70adac50/contracts/mocks/ERC1363Mock.sol

Amxx:
Again, I think 1363 should be just like 3156 or 4626, with the ERC interface in contracts/interfaces. I don't see why 1363 would be handled differently than 4626. 
The erc20/extension folder has enough files already, no need to make things more difficult 

The mock fonder is messy. IMO it's worth putting all the 1363 related mocks in a single file.

vittominacori:
Hi guys, is there any work to be done to merge this PR?

radeksvarz:
Reading the whole discussion here - regarding the adoption, it is like chicken-egg problem. People will start using that once the 1363 is available in the OZ wizard.

I can see there are only foundry-tests blocking the merge now?

Amxx:
> I can see there are only foundry-tests blocking the merge now?

We only do foundry tests when we want fuzzing. IMO fuzzing is not needed here.

AFAIK, the blocker is a formal decision on what to do when targetting EOA. It seems that different people have different interpretations of the ERC document, and that the original idea of the ERC author might not match the ERC document. In that case I'd personnally follow the ERC document, but that is not resolved.

Amxx:
> I propose there were both implementations

That is the best way to make sure people are confused, don't know which one to use (possibly don't understand the difference) ... In the end you'll just have no standard behavior, and no reliability on what to expect from a contract.

Either we come to an agreement on what to provide or we don't provide anything (so far its been the later)

vittominacori:
> I really don't think the behavior of the function with EOA receivers is such an important issue.
> 
> The important bit in this EIP is the addition of a callback for contracts. I discussed with @Amxx earlier today and he's now come to this as well.
> 
> I personally kind of appreciate the simplicity of being sure that the callback always runs.
> 
> In summary: we will merge an implementation of EIP-1363 that reverts on `transferAndCall` to an EOA receiver.

I don't want to open the discussion again but above we agreed to revert to EOA receiver.
I just want to remember that the EIP says that 

```
A contract that wants to accept token payments via transferAndCall or transferFromAndCall MUST implement the following interface...
```

Yes, it doesn't say nothing about EOA addresses but the original purpose was to give a way to use transfer with a receiver that can make a callback.

Why should someone want to call a method on a receiver and expecting it is done but it doesn't as it is an EOA?
To transfer to EOA you can use the standard ERC20 transfer as ERC1363 extends it.

The `*AndCall` methods expect to have the call executed. My vision as EIP author.

Here my implementation https://github.com/vittominacori/erc1363-payable-token
I've PR this over OZ but it was now obsolete. I could PR again using the latest OZ version.
Just check if it should go into `ERC20/extensions` or in other position.

Amxx:
The `transfer*AndCall` function say
> Transfer tokens from `msg.sender` to another address and then call `onTransferReceived` on receiver

Below it says
> A contract that wants to accept token payments via transferAndCall or transferFromAndCall MUST implement the following interface...

IMO it is clear that if the receiver is an EOA the specs is to call and not expect anything coming back

vittominacori:
Yes I'm ok that the EIP could be confusing but for me (as author) the following concept is clear.

If I call the `*AndCall` on an EOA it is impossible that the magic value is returned so I don't want that tokens go to the EOA without doing nothing else. Tokens can be lost and the desired action will never be executed.

If I use `*AndCall` (and we should force that behavior via throwing for EOA) the receiver MUST do the callback.
Otherwise we don't need about this and we can simply use `transfer`.

bcxbb:
What is the status of this ERC? Anything I can do to contribute to finalizing it?

vittominacori:
Once v5 will be ready, I will be happy to close this PR and open a fresh new one with the new guidelines.

freefilm010:
D Rey has sent you an email via Gmail confidential mode:

[image: Gmail logo]Re: [OpenZeppelin/openzeppelin-contracts] ERC1363 (PR #3017) <https://confidential-mail.google.com/msg/ANo0PrUmFG9I-qkOXvf01aZWOLrOapWtkg-ZildCR53eV9T_6YJXMm8z_kH1vcQCFwwmDO42TQiI_8_StM5wcZoCdkxh9-D8e6Yj0ANmd3r-7MgQtFNKiDDaGdtQ_bGDMBhz9x5KKaSh1T6VoB6zhsC0pDxHNK3ZqVmFTDDI6SBuZ46W7f7AuduBwadMqvJAgBXWkL6MVg==>

This message was sent on Jul 17, 2023 at 6:25:56 PM PDT
You can open it by clicking the link below. This link will only work for ***@***.***

View the email <https://confidential-mail.google.com/msg/ANo0PrUmFG9I-qkOXvf01aZWOLrOapWtkg-ZildCR53eV9T_6YJXMm8z_kH1vcQCFwwmDO42TQiI_8_StM5wcZoCdkxh9-D8e6Yj0ANmd3r-7MgQtFNKiDDaGdtQ_bGDMBhz9x5KKaSh1T6VoB6zhsC0pDxHNK3ZqVmFTDDI6SBuZ46W7f7AuduBwadMqvJAgBXWkL6MVg==>

Gmail confidential mode gives you more control over the messages you send. The sender may have chosen to set an expiration time, disable printing or forwarding, or track access to this message. Learn more <https://support.google.com/mail/answer/7674059>
Gmail: Email by Google
Use is subject to the Google Privacy Policy <https://myaccount.google.com/privacypolicy?hl=en>
Google LLC, 1600 Amphitheatre Parkway, Mountain View, CA 94043, USA
You have received this message because someone sent you an email via Gmail confidential mode.
[image: Google logo]

Amxx:
Replaced by #4631

PR #3025, opened 2021-12-15
Replace excodesize assembly with address.code.length 

k06a:
According to https://github.com/ethereum/solidity/releases/tag/v0.8.1:

> #### Compiler Features:
> - Code Generator: Reduce the cost of \<address\>.code.length by using extcodesize directly.

Kudos to this tweet: https://twitter.com/transmissions11/status/1470943473335341059

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -25,15 +25,11 @@ library Address {
      * ====
      */
     function isContract(address account) internal view returns (bool) {
-        // This method relies on extcodesize, which returns 0 for contracts in
-        // construction, since the code is only stored at the end of the
-        // constructor execution.
+        // This method relies on extcodesize/address.code.length, which returns 0
+        // for contracts in construction, since the code is only stored at the end
+        // of the constructor execution.
 
-        uint256 size;
-        assembly {
-            size := extcodesize(account)
-        }
-        return size > 0;
+        return account.code.length > 0;
     }
 
     /**


Amxx (approved):
Thank you @k06a for this PR

axic:
I was considering submitting such a PR back when 0.8.1 was released, but since the version pragma is unchanged from `^0.8.0`, this change can cause grief for users of 0.8.0.

In 0.8.0 the above syntax is supported, but the compiler did not have the shortcut implemented. This means it will first `extcodecopy` the entire target contract and measure size afterwards.

k06a:
@axic thanks for this clarification! But do you think it is still the case after 0.8.10 release? We could specify solidity version in this file as ^0.8.1.

axic:
You can certainly specify a different pragma, but I've found that the maintainers of OZ were reluctant to use different pragma versions in the code base.

Amxx:
@axic You are right, I should have been more careful before merging this!

One thing that might have changed since last time, it the that repo now uses 0.8.3 for other features. Still, this file's pragma needs to be updated to 0.8.1, which is not currently the case.

@frangio should we revert that commit or update the pragma to ^0.8.1 ?

koolamusic:
@Amxx New PR to update 😄 made it easier https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3031

frangio:
My vote in this particular case is to bump this pragma to 0.8.1.

- It's always good to remove assembly.
- 0.8.1 is already almost a year old, I expect most users on 0.8 are using newer versions.
- 0.8.0 to 0.8.1 is unlikely to be very disruptive.

It should be noted as a breaking change in the changelog.

PR #3033, opened 2021-12-16
Update release script to only update version comment for files that changed.

### LINKED ISSUE #3026: Comment in contracts
Release version in each contract cause a very bad compare view https://github.com/OpenZeppelin/openzeppelin-contracts/compare/v4.4.0...v4.4.1.

It seems that 141 files changed but only that comment changed.

I think that it could be better to not to have this in each contract. It adds explicit version in each file but lot of redundancy.

Amxx:
Fixes #3026

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


scripts/release/update-comment.js
```diff
@@ -1,27 +1,49 @@
-#!/usr/bin/env node
-
 const fs = require('fs');
-const glob = require('glob');
-const proc = require('child_process');
+const git = require('gift');
+const semver = require('semver');
 
-const gitStatus = proc.execFileSync('git', ['status', '--porcelain', '-uno', 'contracts/**/*.sol']);
```
frangio, line 7: I'm not very happy about adding a new dependency (`gift`). This line shows how to invoke the `git` command and it should be pretty simple and it doesn't involve callbacks. What do you think about using this instead of gift?

Amxx: I know realize that while it as 16441 download/week (which I think is decent) it hasn't been updated in 4 years. I thought that in some cases it would help with parsing the results. But I guess we can drop that dependency and do the parsing ourselves.

What about `semver` ?

frangio: `semver` is fine!

scripts/release/update-comment.js
```diff
@@ -1,25 +1,34 @@
...
+  .split(/\r?\n/)
+  .filter(Boolean)
+  .filter(file => !file.match(/mock/i));
 
 for (const file of files) {
   const current = fs.readFileSync(file, 'utf8');
   const updated = current.replace(
     /(\/\/ SPDX-License-Identifier:.*)$(\n\/\/ OpenZeppelin Contracts v.*$)?/m,
-    `$1\n// OpenZeppelin Contracts v${version} (${file.replace('contracts/', '')})`,
+    `$1\n// Last updated in OpenZeppelin Contracts v${version} (${file.replace('contracts/', '')})`,
```
frangio, line 30: I'd like "OpenZeppelin Contracts" to come first. Maybe:

```suggestion
    `$1\n// OpenZeppelin Contracts (last updated v${version}) (${file.replace('contracts/', '')})`,
```
?

scripts/release/update-comment.js
```diff
@@ -1,27 +1,35 @@
-#!/usr/bin/env node
-
 const fs = require('fs');
-const glob = require('glob');
 const proc = require('child_process');
+const semver = require('semver');
+const run = (cmd, ...args) => proc.execFileSync(cmd, args, { encoding: 'utf8' }).trim();
```
frangio, line 4: Added this helper. `encoding: 'utf8'` makes `.toString()` unnecessary. `.trim()` makes `.filter(Boolean)` unnecessary.

scripts/release/update-comment.js
```diff
@@ -1,27 +1,35 @@
...
 
-const files = glob.sync('contracts/!(mocks)/**/*.sol');
+// Get latest tag according to semver.
+const [ tag ] = run('git', 'tag')
+  .split(/\r?\n/)
+  .filter(v => semver.valid(v) && semver.lte(v, version))
+  .sort(semver.rcompare);
+
+// Ordering tag → HEAD is important here.
+// Is it right to use HEAD ?
```
frangio, line 21: Yes I think it's right. We can remove this comment.

scripts/release/update-comment.js
```diff
@@ -1,27 +1,35 @@
...
+const files = run('git', 'diff', tag, 'HEAD', '--name-only', 'contracts/**/*.sol')
+  .split(/\r?\n/)
+  .filter(file => !file.match(/mock/i));
```
frangio, line 24: Not all contracts in there are named `*Mock`. I think this is a better filter:

```suggestion
  .filter(file => !file.includes('/mocks/'));
```

scripts/release/update-comment.js
```diff
@@ -1,27 +1,35 @@
-#!/usr/bin/env node
```
frangio, line 1: `version.sh` executes this file without specifying `node`, so this and the executable bit are necessary unless we update that script. `npm version` is currently failing because of that.

scripts/release/update-comment.js
```diff
@@ -1,27 +1,35 @@
...
   process.exit(1);
 }
 
 const { version } = require('../../package.json');
 
-const files = glob.sync('contracts/!(mocks)/**/*.sol');
+// Get latest tag according to semver.
+const [ tag ] = run('git', 'tag')
+  .split(/\r?\n/)
+  .filter(v => semver.valid(v) && semver.lte(v, version))
```
frangio, line 17: What is the effect for prereleases? I think we should filter them out here, and run git diff against the latest non-prerelease.

scripts/release/update-comment.js
Amxx: nice !

scripts/release/update-comment.js
Amxx: This does both, it looks for "mock" in an case-incensitive way. It will match ".../**mock**s/..." and also ".../Something**Mock**.sol.

I think this is more powerful.

scripts/release/update-comment.js
Amxx: good catch !

Amxx: should be addressed now

frangio:
Found some issues through testing but I think it should be good now.

PR #3035, opened 2021-12-17
Use invalid opcode to consume all gas in MinimalForwarder

### LINKED ISSUE #2864: Use invalid opcode in MinimalForwarder
Since Solidity 0.8.0 `assert` no longer triggers an invalid opcode, but uses revert. `MinimalForwarder` should use invalid opcode to consume all gas as a protection for "insufficient gas griefing attack" as described in [this article](https://ipfs.io/ipfs/QmbbYTGTeot9ic4hVrsvnvVuHw4b5P7F5SeMSNX9TYPGjY/blog/ethereum-gas-dangers/).

Amxx:
Fixes #2864 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/metatx/MinimalForwarder.sol
+++ b/contracts/metatx/MinimalForwarder.sol
@@ -54,12 +54,12 @@ contract MinimalForwarder is EIP712 {
         // Validate that the relayer has sent enough gas for the call.
         // See https://ronan.eth.link/blog/ethereum-gas-dangers/
         if (gasleft() <= req.gas / 63) {
-          // We explicitly trigger invalid opcode to consume all gas and bubble-up the effects, since
-          // Panic error do not consume all gas since Solidity 0.8.0
-          // https://docs.soliditylang.org/en/v0.8.0/control-structures.html#panic-via-assert-and-error-via-require
-          assembly {
-            invalid()
-          }
+            // We explicitly trigger invalid opcode to consume all gas and bubble-up the effects, since
+            // Panic error do not consume all gas since Solidity 0.8.0
+            // https://docs.soliditylang.org/en/v0.8.0/control-structures.html#panic-via-assert-and-error-via-require
+            assembly {
+                invalid()
+            }
         }
 
         return (success, returndata);
--- a/contracts/metatx/MinimalForwarder.sol
+++ b/contracts/metatx/MinimalForwarder.sol
@@ -55,7 +55,7 @@ contract MinimalForwarder is EIP712 {
         // See https://ronan.eth.link/blog/ethereum-gas-dangers/
         if (gasleft() <= req.gas / 63) {
             // We explicitly trigger invalid opcode to consume all gas and bubble-up the effects, since
-            // Panic error do not consume all gas since Solidity 0.8.0
+            // neither revert or assert consume all gas since Solidity 0.8.0
             // https://docs.soliditylang.org/en/v0.8.0/control-structures.html#panic-via-assert-and-error-via-require
             assembly {
                 invalid()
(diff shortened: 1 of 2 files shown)


test/metatx/MinimalForwarder.test.js
```diff
@@ -124,43 +125,60 @@ contract('MinimalForwarder', function (accounts) {
...
+        this.req.to = receiver.address;
+        this.req.data = receiver.contract.methods.mockFunctionOutOfGas().encodeABI();
+        this.req.gas = 1000000;
+
+        await expectRevert.assertion(
+          this.forwarder.execute(this.req, this.sign(), { gas: gasAvailable }),
+        );
+
+        const { gasUsed } = await web3.eth.getBlock('latest')
+          .then(({ transactions }) => web3.eth.getTransactionReceipt(transactions.find(Boolean)));
```
frangio, line 178: Let's please avoid `.then`:

```suggestion
        const { transactions } = await web3.eth.getBlock('latest');
        const { gasUsed } = await web3.eth.getTransactionReceipt(transactions.find(Boolean));
```

Additionally I think we should avoid `.find(Boolean)` and `.filter(Boolean)`. I don't think these expressions are intuitive. In this case, I don't understand what we're doing. Isn't there a single transaction in the block?

contracts/metatx/MinimalForwarder.sol
```diff
@@ -50,9 +50,17 @@ contract MinimalForwarder is EIP712 {
         (bool success, bytes memory returndata) = req.to.call{gas: req.gas, value: req.value}(
             abi.encodePacked(req.data, req.from)
         );
+
         // Validate that the relayer has sent enough gas for the call.
         // See https://ronan.eth.link/blog/ethereum-gas-dangers/
-        assert(gasleft() > req.gas / 63);
+        if (gasleft() <= req.gas / 63) {
+            // We explicitly trigger invalid opcode to consume all gas and bubble-up the effects, since
+            // Panic error do not consume all gas since Solidity 0.8.0
```
frangio, line 58:
```suggestion
            // neither revert or assert consume all gas since Solidity 0.8.0
```

frangio (approved):
Thanks!

frangio:
Explained in https://ipfs.io/ipfs/QmbbYTGTeot9ic4hVrsvnvVuHw4b5P7F5SeMSNX9TYPGjY/blog/ethereum-gas-dangers/

lastperson:
@frangio the mentioned fix is only helpful for Forwarder calling another Forwarder. Though it still doesn't guarantee that the target contract gets specified amount of gas.

frangio:
Can you elaborate?

lastperson:
Sure. Consider the target contract
```
contract Target {
    function win(address who) external {
        require(gasleft() > 900000);
        IERC20(0x1234...1234).transfer(who, 100e18);
    }
}
```
User creates a meta tx request with req.gas = 1000000. Relayer picks it up and sends it with 500000 gas.
`Target.win()` reverts but `gasleft() > req.gas / 63` passes.

Amxx:
Its a bit subtle, but here is what happens:

- The forwarder is designed to check at the call in it, *in case it run out of gas*, was given the amount specified.
- You target contract, does not revert because it ran out of gas (and was unable to execute because of that)

You contract check if there is a given amount of gas, but if that is not the case, it doesn't not trigger an out of gas exception.
I'm pretty sure your `Target` contract is an example for the sake of discussion, but IMO it showcases possible bad design by def.

IMO, contract should not check gas that way. 
- They should try to perform operation, and fail with out of gas if that was not enough.
- If they want to guarantee that a subcall has enough gas, they should replicate the Forwarder design (send gas, and if faillure and not enough left, do an invalid call)
- If you really want to ensure `gasleft() > 900000`, don't do it in a `require`, do `if (gasleft() < 900000) assembly { invalid() }`

PR #3037, opened 2021-12-18
[Discussion] ERC20 accounting library (increase library flexibility)

k06a:
Demo only, to demonstrate idea I came up to after this: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3003 :)

This approach would allow having multiple ERC20 accountings in a single contract if needed. Moreover, this ERC20Accounting library allows to patch `totalSupply`, `balances` and `allowances` on the flight (before and after storage) to store additional bits in the existing slots.


frangio:
Moving discussion to a dedicated issue:
- https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3038

PR #3039, opened 2021-12-21
Use effective hash to reduce memory leakage of merkle verification

### LINKED ISSUE #2989: Optimize memory allocation in MerkleProof
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7d17acfb2f12be0a2ad4c08ad4a3d823704b68d6/contracts/utils/cryptography/MerkleProof.sol#L42-L48

`abi.encodePacked` allocates a new memory location on every iteration. This is incredibly wasteful. For this code in particular we don't even need to allocate new memory, since we're hashing exactly two words, we can use the scratch space in memory.

Amxx:
Fixes #2989

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio (approved):
Thank you!

PR #3047, opened 2021-12-22
Add Slither reentrancy check in CI

JulissaDantes:
Adds a new check for the github workflows.

#### PR Checklist (None apply)

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,5 +80,5 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip3 install slither-analyzer
-      - name: Summary of static ananlysis
+      - name: Summary of static analysis
         run: npm run slither
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,10 +71,8 @@ jobs:
           path: '**/node_modules'
           key: npm-v2-${{ hashFiles('**/package-lock.json') }}
           restore-keys: npm-v2-
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python
         uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
 
       - name: Install dependencies
         run: pip3 install slither-analyzer
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,15 +62,14 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 12.x
-      - run: npm install
-      - run: npm compile
-
       - uses: actions/cache@v2
         id: cache
         with:
           path: '**/node_modules'
           key: npm-v2-${{ hashFiles('**/package-lock.json') }}
           restore-keys: npm-v2-
+      - run: npm ci
+        if: steps.cache.outputs.cache-hit != 'true'
       - name: Set up Python
         uses: actions/setup-python@v2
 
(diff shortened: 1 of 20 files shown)


.github/workflows/test.yml
```diff
@@ -54,3 +54,31 @@ jobs:
...
+          restore-keys: npm-v2-
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip3 install slither-analyzer
```
frangio, line 82: Curious to know more about this, was it necessary to upgrade pip?

.github/workflows/test.yml
```diff
@@ -54,3 +54,31 @@ jobs:
...
+      - name: Summary of static ananlysis
```
frangio, line 83: There's a typo in this line.

.github/workflows/test.yml
JulissaDantes: No, I leave it from a local run where I needed to upgrade pip, but after checking its not needed so I removed it

Amxx:
This PR contains a lot of changes that I understand are unrelated (certora in particular).
Are these related to enabling slither?

.github/workflows/test.yml
```diff
@@ -54,3 +54,29 @@ jobs:
...
+
+  slither:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 12.x
+      - run: yarn install
+      - run: yarn compile
```
frangio, line 66: This project doesn't use Yarn. Installing dependencies should be done with npm like the other jobs in this file.

.github/workflows/test.yml
```diff
@@ -54,3 +54,29 @@ jobs:
...
+      - uses: actions/cache@v2
+        id: cache
+        with:
+          path: '**/node_modules'
+          key: npm-v2-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: npm-v2-
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
```
frangio, line 77: We're not using the [matrix](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) and we don't have a `python-version` parameter so this is not necessary.
```suggestion
      - name: Set up Python
        uses: actions/setup-python@v2
```

package.json
```diff
@@ -29,7 +29,8 @@
     "version": "scripts/release/version.sh",
     "test": "hardhat test",
     "test:inheritance": "node scripts/inheritanceOrdering artifacts/build-info/*",
-    "gas-report": "env ENABLE_GAS_REPORT=true npm run test"
+    "gas-report": "env ENABLE_GAS_REPORT=true npm run test",
+    "slither": "npm run clean && slither . --detect reentrancy-eth,reentrancy-no-eth,reentrancy-unlimited-gas"
```
frangio, line 33: Do you know why `clean` is necessary here?

JulissaDantes: If somebody makes a change to fix the reentrancy or just add the diable comment and doesnt compile before pushing the code, slither will check the last compiled contract and show the warning again, unless you clean it before so slither compiles it itself

frangio (approved):

BigBlueHundids (commented):
>.

PR #3056, opened 2021-12-27
Reduce ERC20 allowance before triggering transfer 

frangio:
This has been pointed out many times and even though it is not an issue, I think it's better to change the order to follow the general rule of failing early.

### DIFF
--- a/test/token/ERC20/ERC20.behavior.js
+++ b/test/token/ERC20/ERC20.behavior.js
@@ -63,23 +63,19 @@ function shouldBehaveLikeERC20 (errorPrefix, initialSupply, initialHolder, recip
             });
 
             it('emits a transfer event', async function () {
-              const { logs } = await this.token.transferFrom(tokenOwner, to, amount, { from: spender });
-
-              expectEvent.inLogs(logs, 'Transfer', {
-                from: tokenOwner,
-                to: to,
-                value: amount,
-              });
+              expectEvent(
+                await this.token.transferFrom(tokenOwner, to, amount, { from: spender }),
+                'Transfer',
+                { from: tokenOwner, to: to, value: amount },
+              );
             });
 
             it('emits an approval event', async function () {
-              const { logs } = await this.token.transferFrom(tokenOwner, to, amount, { from: spender });
-
-              expectEvent.inLogs(logs, 'Approval', {
-                owner: tokenOwner,
-                spender: spender,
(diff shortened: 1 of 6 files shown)


test/token/ERC20/ERC20.behavior.js
```diff
@@ -40,7 +40,7 @@ function shouldBehaveLikeERC20 (errorPrefix, initialSupply, initialHolder, recip
       describe('when the recipient is not the zero address', function () {
         const to = anotherAccount;
 
-        describe('when the spender has enough approved balance', function () {
+        describe('when the spender has enough approved allowance', function () {
```
Amxx, line 43: "approved allowance" doesn't sound right to me. What about "when the spender has enough allowance"?

test/token/ERC20/ERC20.behavior.js
```diff
@@ -84,37 +84,50 @@ function shouldBehaveLikeERC20 (errorPrefix, initialSupply, initialHolder, recip
...
+              await expectRevert(
+                this.token.transferFrom(tokenOwner, to, amount, { from: spender }),
+                `${errorPrefix}: transfer amount exceeds balance`,
               );
             });
           });
         });
 
-        describe('when the spender does not have enough approved balance', function () {
+        describe('when the spender does not have enough approved allowance', function () {
```
Amxx, line 102: same a above.

test/token/ERC20/ERC20.behavior.js
```diff
@@ -84,37 +84,50 @@ function shouldBehaveLikeERC20 (errorPrefix, initialSupply, initialHolder, recip
...
               );
             });
           });
 
           describe('when the token owner does not have enough balance', function () {
-            const amount = initialSupply.addn(1);
+            const amount = allowance;
+
+            beforeEach('reducing balance', async function () {
+              await this.token.transfer(to, initialSupply, { from: tokenOwner });
```
Amxx, line 124: The other "reducing balance" `beforeEach` does a transfer of 1. This one does initialSupply. Can we make them more consistent ?

frangio: Can't make it 1 because the test passes, but I can make it 2.

frangio:
@Amxx Your commits look good to me. Should we merge?

Amxx (approved):

PR #3060, opened 2021-12-29
Avoid assembly in signature V and S decomposition

k06a:
Is there any specific reason why assembly was used here?

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Not really. It beeing used extensively in the same file might explain how we were drawn to it initially, but its true we could use native solidity here. I don't expect it to be much more expensive.

k06a:
I find assembly useful for converting `bytes` to `uint256/bytes32`. I see most of the cases in this file are for this purpose.

frangio (approved):
Thank you!

PR #3061, opened 2021-12-29
Update ERC20.sol

raviteja18:
In 'transferFrom' function, 'currentAllowance' should be validated with the 'amount' before transferring the tokens.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Hello @raviteja18 

This has already been discussed many times. The summary is:
- Both orders are valid and secure.
- #3056 is planning to change that in the next version.

PR #3072, opened 2022-01-03
docs: add instructions to install via forge

alephao:
Adding instructions to install openzeppelin contracts via forge due to its increasing popularity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,16 @@
 
 ### Installation
 
+**Using Hardhat**
+
 ```console
 $ npm install @openzeppelin/contracts
+```
+
+**Using Forge**
+
+```console
+$ forge install openzeppelin/openzeppelin-contracts
 ```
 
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @alephao and thank you for this PR.

Being not familiar with forge, I'll need to discuss that with other members of the team, but I think it would be a great addition !

alephao:
@Amxx I should have added this to the comment, this is the tool, it has almost 1k starts and it's getting crazy popular due to twitter solidity "devfluencers": https://github.com/gakonst/foundry

frangio:
I have some concerns about the way `forge install` currently works: https://github.com/gakonst/foundry/issues/401

That said, I'm not sure I'd agree with mentioning particular tools such as Hardhat or Forge in the installation instructions. The current instructions are not for Hardhat, they also apply to Truffle, because they describe our main distribution channel which is npm. We can definitely add a mention of the repository as another distribution channel, and mention the question about the master branch versus the tags that I explained in the Foundry issue.

PR #3073, opened 2022-01-03
EnumerableMap.AddressToUintMap

Tynukua:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/structs/EnumerableMap.sol
+++ b/contracts/utils/structs/EnumerableMap.sol
@@ -252,7 +252,7 @@ library EnumerableMap {
      * already present.
      */
     function set(
-        AddressToUint storage map,
+        AddressToUintMap storage map,
         address key,
         uint256 value
     ) internal returns (bool) {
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @Tynukua 

Do you have a specific need for this flavor of EnumerableMap? Are you planning to use it for storing balances?

I think it would be a great addition to this library, but I would go one step further and `addressToBytes32` and `bytes32ToAddress` functions to cleanup all the casting. They could be file-level function, or we could add them to a library (either safecast or something new)

fdietze:
I am also interested in this addition.

My use case would indeed be to store balances.

gaspardip:
Same use case here, surprised it has't been added yet.

Amxx:
I don't think there is any reason not to include that.

The next steps would be to add testing and a Changelog entry.

gaspardip:
> I don't think there is any reason not to include that.
> 
> The next steps would be to add testing and a Changelog entry.

I can do that, should I continue this in another CR?

frangio:
Closing in favor of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3150.

PR #3079, opened 2022-01-05
Normalize return variable names in IERC721Enumerable

martriay:
For consistency across all functions. This PR does not add a variable name for `totalSupply()` nor to the return variable of the implemented contracts.


Amxx:
FYI, the reason we don't already have this is that we try to have our interfaces match the [corresponding EIP](https://eips.ethereum.org/EIPS/eip-721) as much as possible.

I'm curious about the usecase (I'm guessing the interface will be processed by some library and you want the output named). Also, why not do the same to `totalSupply` ?

martriay:
Just for consistency and I wasn't sure about `totalSupply` so I mentioned just in case. If that's the rationale, I agree we should remove them all then!

Amxx:
@frangio any idea ?

frangio:
I think we should remove the existing name rather than add the missing one.

martriay:
removed

Amxx (approved):

Amxx:
Thank you @martriay

PR #3080, opened 2022-01-05
Enable timestamp-based governor with choice of timer type

### LINKED ISSUE #3081: Timestamp-based Governor
**🧐 Motivation**

I'm using your Governance implementation, and I want to check proposal deadlines with Timestamps instead of Block Numbers. 

Currently, you have a hardcoded Timer.BlockNumber in the ProposalCore struct inside the Governor contract, and I made some research through your implementation checking how to change this behavior. In this search I found that you have already a Timer for timestamps with the same interface as block numbers in the Timers library, so changing the ProposalCore struct to Timers.Timestamp achieves what I want. 

The issue is that in order to change this struct, right now I would need to define another Governor contract and fix the inheritance tree (to redefine the struct), or override a lot of functions to fix the usage of ProposalCore with a new one (struct), I think that is too complicated for the purpose of the library. 

So I propose to use a Generic Timer inside the Timer library that allows us to define the behavior depending on some configurations, specifically in a virtual method in Governor called timerType (added to the interface to get the current Timer used on the Governance), which can be easily overridden to select what Timer we want to use, and we could add also a config in the GovernorSettings module to allow swapping between this types in production. I made a PR for this and is linked to this issue.

ericnordelo:
Fixes #3081 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -25,13 +25,13 @@ import "./IGovernor.sol";
  */
 abstract contract Governor is Context, ERC165, EIP712, IGovernor {
     using SafeCast for uint256;
-    using Timers for Timers.BlockNumber;
+    using Timers for Timers.Generic;
 
     bytes32 public constant BALLOT_TYPEHASH = keccak256("Ballot(uint256 proposalId,uint8 support)");
 
     struct ProposalCore {
-        Timers.BlockNumber voteStart;
-        Timers.BlockNumber voteEnd;
+        Timers.Generic voteStart;
+        Timers.Generic voteEnd;
         bool executed;
         bool canceled;
     }
@@ -82,6 +82,20 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
      */
     function version() public view virtual override returns (string memory) {
         return "1";
+    }
+
+    /**
+     * @dev See {IGovernor-timerType}.
+     */
+    function timerType() public view override returns (string memory) {
+        return _timerType() == Timers.Type.BLOCKNUMBER ? "BlockNumber" : "Timestamp";
+    }
+
+    /**
+     * @dev Override this function to modify Governor Timer type.
+     */
+    function _timerType() internal view virtual returns (Timers.Type) {
+        return Timers.Type.BLOCKNUMBER;
     }
 
     /**
@@ -212,8 +226,8 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor {
         uint64 snapshot = block.number.toUint64() + votingDelay().toUint64();
         uint64 deadline = snapshot + votingPeriod().toUint64();
 
-        proposal.voteStart.setDeadline(snapshot);
-        proposal.voteEnd.setDeadline(deadline);
+        proposal.voteStart.setDeadline(snapshot, _timerType());
+        proposal.voteEnd.setDeadline(deadline, _timerType());
 
         emit ProposalCreated(
             proposalId,
--- a/contracts/governance/IGovernor.sol
+++ b/contracts/governance/IGovernor.sol
@@ -67,6 +67,12 @@ abstract contract IGovernor is IERC165 {
     function version() public view virtual returns (string memory);
 
     /**
+     * @notice module:core
+     * @dev The type of timer being used by this implementation (BlockNumber or Timestamp). Default: "BlockNumber"
+     */
+    function timerType() public view virtual returns (string memory);
+
+    /**
      * @notice module:voting
      * @dev A description of the possible `support` values for {castVote} and the way these votes are counted, meant to
      * be consumed by UIs to show correct vote options and interpret the results. The string is a URL-encoded sequence of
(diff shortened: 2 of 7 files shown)


ericnordelo:
@Amxx

I'm using your Governance implementation, and I want to check proposal deadlines with Timestamps instead of Block Numbers.

Currently, you have a hardcoded Timer.BlockNumber in the ProposalCore struct inside the Governor contract, and I made some research through your implementation checking how to change this behavior. In this search I found that you have already a Timer for timestamps with the same interface as block numbers in the Timers library, so changing the ProposalCore struct to Timers.Timestamp achieves what I want.

The issue is that in order to change this struct, right now I would need to define another Governor contract and fix the inheritance tree (to redefine the struct), or override a lot of functions to fix the usage of ProposalCore with a new one (struct), I think that is too complicated for the purpose of the library.

So I propose to use a Generic Timer inside the Timer library that allows us to define the behavior depending on some configurations, specifically in a virtual method in Governor called timerType (added to the interface to get the current Timer used on the Governance), which can be easily overridden to select what Timer we want to use, and we could add also a config in the GovernorSettings module to allow swapping between this types in production. I made a PR for this and is linked to this issue.

Amxx:
Hello @ericnordelo, and thank you for this very interesting proposal.

Block number is something we inherited from governor bravo compatibility, and IMO timestamp would be better. Your approach is really interesting, and I'd definitely want to explore it deeper.

One thing however, it that we realized that using Timers structure was a bad idea because it forces memory alignment. In this case, even though 

```
Timers.Generic voteStart;
Timers.Generic voteEnd;
bool executed;
bool canceled;
```

use a total of 8 + 8 + 1 + 1 = 18 bytes ... they are not packed into a single storage slot (of 32 bytes) because the structure forces a new slot to be used. [I'd like to fix that using UDVT at some point](https://github.com/Amxx/openzeppelin-contracts/blob/breaking/Timers/udvt/contracts/utils/Timers.sol). 

On one hand, it would make sense to do both changes at once, but since the transition to UDVT would be breaking, we would want to wait for 5.0. Maybe your proposal should make its way into the code before we do this breaking change.

@frangio, any thought ?

ericnordelo:
Got it, makes total sense, just would be nice to get this Idea of Generic as soon as possible because now the workaround that we have is copying the Governor contract to modify the ProposalCore struct, and then copying the rest of the modules to modify the Governor inheritance (or add a lot of overrides in the Governance contract for each one of the module's methods using the Governor).

frangio:
I think this is an elegant solution that makes sense, and doesn't add significant overhead.

I don't think we should offer a timestamp-based version out of the box until it's clearer how it interacts with the Governor ecosystem. But we can include this mechanism for switching to timestamps, and inform people of these potential compatibility issues, but give them the choice.

frangio:
@Amxx I think this is unrelated to moving to UDVT. The storage layout in this PR is compatible, right?

Amxx:
It is compatible with the current storage layout ... that we may break when we move to UDVT (which I should optimize that whenever we get to 5.0).

Really this PR is mostly good. I don't like the fact that it is not backward compatible from an ERC165 point of view, but if we fix that, then it should be good to merge!

tarrencev:
Very nice! This would enable govornor to work on blockchains with irregular blocktimes, like optimism. This current implementation will break ERC20Votes though. We will still need to store a block number for the snapshot block.

TripleD1337:
Hey guys, I would like to contribute and wanted things to clarify.  Next steps would be to find a fitting solution for deadline creation and snapshot generation?

Amxx:
AFAIK, next steps are to figure out how we want the token <> governor interaction to include info about the timestamps vs blocknumbers ... and standardize that interface.

I started working on an ERC for that ... then moved on to other priorities. I should resume working on that soon.

z0r0z:
I completely support this initiative and hope to see an OZ governor templates that can support tokens that use timestamp checkpoints for better multichain compatibility. This would allow OZ to better integrate with [Kali](https://github.com/kalidao) and [Baal](https://github.com/Moloch-Mystics/Baal/blob/feat/baalZodiac/contracts/utils/BaalVotes.sol) DAO voting token implementations.

frangio:
@z0r0z I see that Baal has implemented this by defining `getPriorVotes(address account, uint256 timeStamp)`. In our case we have concerns with this simple approach because it leads to a possible error where a timestamp-driven Governor is connected to a block number-driven ERC20Votes contract. So the key challenge for us is to make sure that the Governor, ERC20Votes, and any other off- or on-chain infrastructure is properly coordinated in terms of what measure of time they're using.

TtheBC01:
Hey everyone, just wondering if there was currently any timeline regarding this improvement making it to the OZ main repository? Would really like to use it in an upcoming project.

frangio:
Closing in favor of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3934. This is our approach to this problem while maintaining backwards compatibility and minimizing the chance of setting up a system with mixed timer types.

PR #3085, opened 2022-01-06
Do not reduce approval on transferFrom if current allowance is type(uint256).max

### LINKED ISSUE #3084: Use WETH gas optimization for ERC-20 transferFrom
**🧐 Motivation**
The Wrapped Ether (WETH) ERC-20 contract has a gas optimization that does not update the allowance if it is the max uint. I'd like to incorporate this into OpenZeppelin to save 5000 gas per `transferFrom` call when a max approve is used.

**📝 Details**
The `transferFrom` code for WETH is:
```
    function transferFrom(address src, address dst, uint wad)
        public
        returns (bool)
    {
        require(balanceOf[src] >= wad);

        if (src != msg.sender && allowance[src][msg.sender] != uint(-1)) {
            require(allowance[src][msg.sender] >= wad);
            allowance[src][msg.sender] -= wad;
        }

        balanceOf[src] -= wad;
        balanceOf[dst] += wad;

        Transfer(src, dst, wad);

        return true;
    }
```

The key line is ` if (src != msg.sender && allowance[src][msg.sender] != uint(-1)) {`. If the allowance is the max uint (`0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`), then the WETH contract does not update the `allowance` mapping. This saves 5000 gas for the SSTORE. The only downside is that this changes the abstraction for approvals; if you set the approval to the maximum amount, then the contract can spend more than the max uint of your token. However, this doesn't seem like a risk, as users wanting to actually specify that amount can simply do the max uint minus 1.

My proposed change is to incorporate this into . This could save 5000 gas (or 10-20%) per ERC-20 transferFrom, which is used all across Unsiwap, Compound, and other big protocols.

0xclaudeshannon:
Fixes #3084

The Wrapped Ether (WETH) ERC-20 contract has a gas optimization that does not update the allowance if it is the max uint. I'd like to incorporate this into OpenZeppelin to save 5000 gas per `transferFrom` call when a max approve is used.

The `transferFrom` code for WETH is:
```
    function transferFrom(address src, address dst, uint wad)
        public
        returns (bool)
    {
        require(balanceOf[src] >= wad);

        if (src != msg.sender && allowance[src][msg.sender] != uint(-1)) {
            require(allowance[src][msg.sender] >= wad);
            allowance[src][msg.sender] -= wad;
        }

        balanceOf[src] -= wad;
        balanceOf[dst] += wad;

        Transfer(src, dst, wad);

        return true;
    }
```

The key line is ` if (src != msg.sender && allowance[src][msg.sender] != uint(-1)) {`. If the allowance is the max uint (`0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`), then the WETH contract does not update the `allowance` mapping. This saves 5000 gas for the SSTORE. The only downside is that this changes the abstraction for approvals; if you set the approval to the maximum amount, then the contract can spend more than the max uint of your token. However, this doesn't seem like a risk, as users wanting to actually specify that amount can simply do the max uint minus 1.

My proposed change is to incorporate this into OpenZeppelin. This could save 5000 gas (or 10-20%) per ERC-20 transferFrom, which is used all across Unsiwap, Compound, and other big protocols.

I need feedback from the team for how to add this to the documentation. It passes all the tests.

The idea for this PR came up in the [GAS DAO](https://www.gasdao.org/) community discord.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -125,8 +125,8 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
     /**
      * @dev See {IERC20-approve}.
      *
-     * Note If `amount` is the maximum `uint256`, the allowance is not updated.
-     * on `transferFrom`. This is semantically equivalent to an infinite approval.
+     * Note If `amount` is the maximum `uint256`, the allowance is not updated on
+     * `transferFrom`. This is semantically equivalent to an infinite approval.
      *
      * Requirements:
      *
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -125,6 +125,9 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
     /**
      * @dev See {IERC20-approve}.
      *
+     * Note If `amount` is the maximum `uint256`, the allowance is not updated.
+     * on `transferFrom`. This is semantically equivalent to an infinite approval.
+     *
      * Requirements:
      *
      * - `spender` cannot be the zero address.
@@ -139,6 +142,9 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
      *
      * Emits an {Approval} event indicating the updated allowance. This is not
      * required by the EIP. See the note at the beginning of {ERC20}.
+     *
+     * Note Does not update the allowance if the current allowance
+     * is the maximum `uint256`.
      *
      * Requirements:
      *
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -125,7 +125,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
     /**
      * @dev See {IERC20-approve}.
      *
-     * Note If `amount` is the maximum `uint256`, the allowance is not updated on
+     * NOTE: If `amount` is the maximum `uint256`, the allowance is not updated on
      * `transferFrom`. This is semantically equivalent to an infinite approval.
      *
      * Requirements:
@@ -143,7 +143,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
      * Emits an {Approval} event indicating the updated allowance. This is not
      * required by the EIP. See the note at the beginning of {ERC20}.
      *
-     * Note Does not update the allowance if the current allowance
+     * NOTE: Does not update the allowance if the current allowance
      * is the maximum `uint256`.
      *
      * Requirements:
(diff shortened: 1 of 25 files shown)


earlit:
This could save users Millions on Gas

contracts/token/ERC20/ERC20.sol
```diff
@@ -153,9 +153,11 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
         uint256 amount
     ) public virtual override returns (bool) {
         uint256 currentAllowance = _allowances[sender][_msgSender()];
-        require(currentAllowance >= amount, "ERC20: transfer amount exceeds allowance");
-        unchecked {
-            _approve(sender, _msgSender(), currentAllowance - amount);
+        if (currentAllowance != 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff) {
```
frangio, line 156:
```suggestion
        if (currentAllowance != type(uint256).max) {
```

Amxx: Why not 
`if (currentAllowance !=type(uint256).max) {`

IMO that would be more explicit

frangio: See my comment above, `.max` requires 0.8.8.

Amxx: 0.8.8 introduced max for enums. For value types like uint256 is available in 0.8.0 ... and we are already using it in many places.

frangio: Oh nice, I read the readme too quickly.

frangio:
Missing changelog entry and documentation before merging.

0xclaudeshannon:
> Missing changelog entry and documentation before merging.

Changelog in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3085/commits/b34920b2e400bd74732f11d70f615cf3c0f0a9c9

What should I add for documentation? And which file is that in

0xclaudeshannon:
Thanks for changing it to `type(uint256).max` that's much more aesthetically pleasing.

frangio:
For documentation I think we should add a mention of the behavior in the NatSpec for transferFrom.

0xclaudeshannon:
> For documentation I think we should add a mention of the behavior in the NatSpec for transferFrom.

This good? https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3085/commits/04d88ddb6cf21e8adb25007dc4a637d270aa0ae4

Amxx:
I think this documentation should go in ERC20.sol ... because it is a particularity of the implementation that is not part of the interface/ERC design.

Also, I'm always personally prefer using `uint256` over `uint`, particularly when talking about the maximum value.

Amxx:
- added tests.
- added equivalent mechanism to ERC777 transferFrom for consistency

0xclaudeshannon:
> I think this documentation should go in ERC20.sol ... because it is a particularity of the implementation that is not part of the interface/ERC design.
> 
> Also, I'm always personally prefer using `uint256` over `uint`, particularly when talking about the maximum value.

fixed

frangio (approved):
LGTM!

PR #3103, opened 2022-01-11
Erc721 enumerable, add: tokensByOwner

Evankhell561:
Erc721 enumerable: adds a function to make it easy to view all ERC721 tokens held in a address.

#### PR Checklist

- [ x] Tests


Amxx:
Hello @Evankhell561 

This function might indeed be interesting to some user, even though we discourage using the enumerability extension. I would however like to point out that `ERC721Enumerable` is standardized in the [corresponding ERC](https://eips.ethereum.org/EIPS/eip-721) and should NOT be modified. If we were to add this function, we should modify the ERC165 interfaceId. Changing that would break things!

The result is that if we add this function, discoverability might be hard :/

Evankhell561:
> Hello @Evankhell561
> 
> This function might indeed be interesting to some user, even though we discourage using the enumerability extension. I would however like to point out that `ERC721Enumerable` is standardized in the [corresponding ERC](https://eips.ethereum.org/EIPS/eip-721) and should NOT be modified. If we were to add this function, we should modify the ERC165 interfaceId. Changing that would break things!
> 
> The result is that if we add this function, discoverability might be hard :/

Thanks for the review, I will close this PR then.

PR #3104, opened 2022-01-12
add Hooks _afterTokenTransfer for ERC1155

CryptoV8:
## Added **Hooks of _afterTokenTransfer for ERC1155**

Just like ERC721, add the hooks of `_afterTokenTransfer` for ERC1155.

Currently ERC1155 only have the `before` Hook, no `after` hook, hope to add this.

## Example

Source Code: https://polygonscan.com/address/0x2048c74851f2ed5c45db70ccb7898a73cfede11e#code#L1343

Merge2048 is ERC-1155, but any address can't hold the same token id.

When the `to` address receives a token, which `to` address already had the same id, it'll be merged to the next.

In this case, we need the `_afterTokenTransfer` to handle it.

### DIFF
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -180,6 +180,8 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
         emit TransferSingle(operator, from, to, id, amount);
 
         _doSafeTransferAcceptanceCheck(operator, from, to, id, amount, data);
+
+        _afterTokenTransfer(operator, from, to, _asSingletonArray(id), _asSingletonArray(amount), data);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Amxx:
Thank you @CryptoV8 for this request!

We'll work on that for the next release !

frangio:
I guess this does make sense for simmetry with the other token standards, would you agree @Amxx?

As it is now there is unnecesary memory overhead by invoking `_asSingletonArray` twice though.

Amxx:
I'm not able to build on top of this branch, I've added commits and pushed that to my repo:
https://github.com/Amxx/openzeppelin-contracts/tree/feature/ERC1155/afterTokenTransfer

frangio:
@Amxx It's because @CryptoV8 created this PR from their `master` branch. You can open a new PR to supercede this one.

Amxx:
Replaced by #3166

PR #3105, opened 2022-01-12
prevent ERC721 from transfer token to owner self

gzliudan:
In current ERC721 contract when both of caller, from, to are tokenId owner,  the functions of transferFrom and safeTransferFrom will runs without error. But logically, it makes no sense. If developers do not notice this, it may lead to business problems.


Amxx:
Hello @gzliudan 

Having `from` = `to` is indeed a no-op, but that doesn't mean it should be an issue. If someone ask for this, I don't see why we should prevent it.  Similarly, transfers of 0 ERC20 are allowed, and I don't think we should prevent them.

Note that the ERC721 specification doesn't forbid such transfers, and we should stick with the specification as much as possible.

PR #3110, opened 2022-01-14
Move abs(int256) from Math to SafeMath

Amxx:
To cherrypick on 4.5.0-rc.0

Improve #2984 considering #2686

#### PR Checklist

- [x] Tests
- [x] Changelog entry


frangio (approved):
There are some linter errors but is good otherwise.

frangio (approved):

PR #3111, opened 2022-01-15
Fix broken pull request links in change log

codler:
Fix broken pull request links in change log


Amxx:
Hello @codler and thank you for finding this.

Amxx:
Note: this was applied in a different PR, because our process require this change to be done on master and then merge into the release branch (and not the other way around)

PR #3115, opened 2022-01-16
Fix ERC721 usage example

### LINKED ISSUE #3068: Fix ERC721 usage example in docs
An example of how to use ERC721 contract on page https://docs.openzeppelin.com/contracts/4.x/erc721#constructing_an_erc721_token_contract seems not working.

It gives me the following errors:
`Derived contract must override function "_burn". Two or more base classes define function with same name and parameter types.`
`Derived contract must override function "tokenURI". Two or more base classes define function with same name and parameter types.`

Looks like it's because `GameItem` contract derives from both ERC721 and ERC721URIStorage, but the ERC721URIStorage also derives from the ERC721, so maybe this example should be edited.

v0xat:
Fixes #3068 

Suggest simply removing the ERC721 import and leaving only ERC721URIStorage.

Another way is to add the overrides for `_burn()` and `tokenURI()` required by Solidity, but I think that would be too much code for a basic tutorial.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/docs/modules/ROOT/pages/erc721.adoc
+++ b/docs/modules/ROOT/pages/erc721.adoc
@@ -16,11 +16,10 @@ Here's what a contract for tokenized items might look like:
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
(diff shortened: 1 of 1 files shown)


frangio (approved):
Thanks @v0xat!

PR #3121, opened 2022-01-19
Improve clarity of error messages in functions checking token ownership

### LINKED ISSUE #3120: Improve clarity of error messages in functions checking token ownership
**🧐 Motivation**
Good morning folks. I feel this somewhat walks the line between feature and bug.

While extending some `.*transfer()` functionality in the OpenZeppelin contracts (thank you for providing them!) I found some of my tests failing due to an error reading:

```
ERC721: transfer caller is not owner nor approved
```

As a newcomer, most of the terminology of ownership in documentation and in the "higher level" contracts refers to the contract owner - with frequently used modifiers like `onlyOwner` and the `Ownable` contract reinforcing this. I spent quite some verifying that the test was indeed making calls with the owner's signature, and that signature matches the `.owner()` in the instance of the contract under test. 

After delving in a little more I found that the term "owner", in this context, refers to the _token_ owner. 

(A more dramatic reading of this experience as at the bottom for those interested in such things)

**📝 Details**
A small amount of clarification may go a long way to saving others the journey I went on myself, for example extending the error message to

```
ERC721: transfer caller is not *token* owner nor approved
```

I have a PR ready with this, and for all similar instances of such messages. Thank you for your consideration.

**Dramatic Reading**
>"But I am the owner", I whinge at the test suite, exhausted. In red, emblazoned in the flickering terminal: "'ERC721: transfer caller is not owner nor approved'" stands its ground - repeatedly unconvinced of my ownership. I checksum, and I verify. The messages are signed with poise, grace, and cryptographic correctness. The deployer is requesting the transfer. The very deployer that is also the current owner of this fine establishment.
So I dig. I fear I'll dig too greedily and too deep, and awaken something in the darkness.
"Ah." I sigh. Owner can mean owner of the contract, but this "owner" refers to a check of owner of the token.
So let this fable guide you all - don't reuse curt terminology for something that is already overloaded. All this could have been avoided with a simple
"ERC721: transfer caller is not *token* owner nor approved".

MarcelCutts:
## Issue

Fixes #3120  

As described in the issue, I have expanded the (hopefully complete set) of error messages in which the user is informed they have insufficient ownership or approval for an action on a token. The addition is essentially `not owner` -> `not token owner` to distinguish from contract ownership.

One additional change was made in an instance where an error message for transferring tokens did not begin with the term `transfer caller`, instead just `caller`. I added `transfer` for uniformity.

Tests have been updated to check against updated error messages.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

are-you-the-owner-cam-gigandet

### DIFF
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -123,7 +123,7 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     ) public virtual override {
         require(
             from == _msgSender() || isApprovedForAll(from, _msgSender()),
-            "ERC1155: caller is not owner nor approved"
+            "ERC1155: transfer caller is not token owner nor approved"
         );
         _safeTransferFrom(from, to, id, amount, data);
     }
@@ -140,7 +140,7 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     ) public virtual override {
         require(
             from == _msgSender() || isApprovedForAll(from, _msgSender()),
-            "ERC1155: transfer caller is not owner nor approved"
+            "ERC1155: transfer caller is not token owner nor approved"
         );
         _safeBatchTransferFrom(from, to, ids, amounts, data);
     }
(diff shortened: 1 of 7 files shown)


Amxx:
Replaced with #3254 (see issue for rational)

PR #3128, opened 2022-01-23
Simplification of ERC777's transfer & transferFrom by using _send

kevupton:
Reuse of the internal `_send` method for the `transfer` and `transferFrom` methods, which allows one method for all, meaning it can be nicely override by a parent class.

There was issues with adding custom logic to the _move function. However this may allow for the cleanest approach to doing this.

Please let me know if this looks ok, if so, will continue with tests, documentation and a changelog entry

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC777/ERC777.sol
+++ b/contracts/token/ERC777/ERC777.sol
@@ -162,9 +162,9 @@ contract ERC777 is Context, IERC777, IERC20 {
      */
     function isOperatorFor(address operator, address tokenHolder) public view virtual override returns (bool) {
         return
-        operator == tokenHolder ||
-        (_defaultOperators[operator] && !_revokedDefaultOperators[tokenHolder][operator]) ||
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @kevupton 

This has been on my roadmap lately, particularly considering the extension of #3104 to ERC777.
I'll have a deeper look this week!

kevupton:
Hey @Amxx ,
Awesome, glad to hear! Appreciate the prompt reply. 
The only other thought aligned with this one as well is making the `_move` method `internal virtual` instead of private. 
Which would also be great to do, if it was also being used by each function! 🥳

Amxx (approved):

Amxx:
I'm don't think we should make _move internal ... because I think tokens should not be transferred without the proper callbacks being made.

PR #3134, opened 2022-01-26
feat(enumerablemap): add UintToUint and UintToBytes32 Map

wiasliaw:
# Map Extension

The file `contracts/utils/structs/EnumerableMap.sol` has the base implementation and extends the `UintToAddress` map.
This PR is to extends two types of map, `UintToUint` and `UintToBytes32`.

BTW, I cannot extract out the testing into a behavior file due to that expect cannot compare bignumber in object.

#### PR Checklist

- [X] Tests
- [X] Documentation
- [ ] Changelog entry


wiasliaw:
Closed. Reason see #3073.

PR #3144, opened 2022-01-28
Add revert message in case of ERC1155TotalSupply:_burn() overflow

### LINKED ISSUE #3141: ERC1155Supply hook into _beforeTokenTransfer causing overflows when attempting to burn more than totalSupply
Hello,

I have recently updated from 4.2 to 4.4, and noticed a change in ERC1155Supply moving supply tracking to the _beforeTokenTransfer hook, and although I enjoy not having to mention ERC1155Supply in the override chain of all the methods it used to override, that change has brought another issue:

The ERC1155Supply extension is now doing arithmetic on transferred tokens before any check on the transferred amount has been done, causing a panic code for underflowing outside an unchecked block in the case of an attempt to burn more tokens than the current total supply.

I am not sure this qualifies as a bug, as it may be an intended behavior, but since I'd selfishly like to rely on this library to revert with neat messages in such cases, I'm still reporting it!

Thank you for your time.

**💻 Environment**

OpenZeppelin 4.4, Hardhat

Amxx:
Fixes #3141

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC1155/extensions/ERC1155Supply.sol
+++ b/contracts/token/ERC1155/extensions/ERC1155Supply.sol
@@ -51,7 +51,13 @@ abstract contract ERC1155Supply is ERC1155 {
 
         if (to == address(0)) {
             for (uint256 i = 0; i < ids.length; ++i) {
-                _totalSupply[ids[i]] -= amounts[i];
+                uint256 id = ids[i];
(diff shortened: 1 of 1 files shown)


frangio (approved):
LGTM

PR #3153, opened 2022-01-31
Add a double ended queue

Amxx:
`DoubleEndedQueue` is an array with O(1) read access and with O(1) push to front and back as well as O(1) pop from front and back. It can also be cleared (reset) in O(1).

This structure can be used to create FIFO, LIFO or other similar structures.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio:
If FIFO and LIFO are just wrappers that restrict the available functions, we should just have Vector.

But FIFO doesn't need negative indices. So I don't know if this is the best implementation.

What is the extra cost of using a mapping vs using a contiguous array?

Amxx:
FIFO doesn't need negative indices, but AFAIK they do not add any cost.

I'd be ok with removing FIFO and LIFO and only having Vector.

Amxx:
```
contract TestArray {
    uint256[] public _array;

    function push() external { // cost: 65410 gas
        _array.push(1);
    }
}

contract TestMapping {
    mapping(uint256 => uint256) public _mapping;
    uint256 public _mappingLength;

    function push() external { // cost: 65566 gas
        _mapping[_mappingLength++] = 1;
    }
}
```

Amxx:
On the other hand, if we use uint128 for pointers, and `bytes[]` instead of the `mapping(int256 → bytes32)`, we would be more easily compatible with array operations (slices ?) if that ever gets into solidity.

Amxx:
Note that, if we want to keep `begin` and `end` in the same slot, we cannot use the array.length that solidity would provide.

AFAIL, this means that we cannot use `.push` and that we have to do all read/write in assembly :/

frangio:
I think the array would win against the mapping if you do multiple pushes in a row, because there is a single hash you need to do and then you just incement the slot number.

Amxx:
```
contract TestArray {
    uint256[] internal _array;

    function push() public { // cost: 65410 gas
        unchecked {
            _array.push(1);
        }
    }
    function pushMultiple() public { // cost: 110188 gas
        push();
        push();
        push();
    }
}

contract TestMapping {
    mapping(uint256 => uint256) internal _mapping;
    uint256 internal _mappingLength;

    function push() public { // cost: 65464 gas
        unchecked {
            _mapping[_mappingLength++] = 1;
        }
    }
    function pushMultiple() public { // cost: 110302 gas
        push();
        push();
        push();
    }
}
```

frangio:
I was wrong in my previous comment, in the case of the array you don't even need to hash on-chain, because you know the position statically, so the compiler just precomputes the hash and embeds it in the bytecode.

Anyway, the difference between these two methods is less than 0.1% of the cost of a push so I think either option is fine in terms of gas cost.

Amxx:
I really think that having an array is not realistic if you are not using the built-in length. And if we are using the built-in, that is one slot that we cannot share. IMO, it makes a lot of sense to use the same slot for begin+end (2x uint128).

And we need begin + end for FIFO operations.

test/utils/structs/Vector.test.js
```diff
@@ -0,0 +1,91 @@
...
+        await this.vector.pushBack(bytesB);
+        await this.vector.pushFront(bytesA);
+        await this.vector.pushBack(bytesC);
+      });
+
+      it('content is valid', async function () {
+        expect(await this.vector.empty()).to.be.equal(false);
+        expect(await this.vector.front()).to.be.equal(bytesA);
+        expect(await this.vector.back()).to.be.equal(bytesC);
+        expect(await getContent(this.vector)).to.have.all.members([ bytesA, bytesB, bytesC ]);
```
frangio, line 40: `members` is not an ordered check. I think we should be checking order, so we should use `deep.equal`.

test/utils/structs/Vector.test.js
```diff
@@ -0,0 +1,91 @@
...
+      });
+
+      it('out of bound access', async function () {
+        await expectRevert.unspecified(this.vector.at(10));
```
frangio, line 44: It's bad that we don't have a way of testing custom errors... What if we introduce Ethers.js in this file? I think it allows testing custom errors, right?

test/utils/structs/Vector.test.js
```diff
@@ -0,0 +1,91 @@
...
+      });
+
+      describe('push', function () {
+        it('front', async function () {
+          await this.vector.pushFront(bytesD);
+
+          expect(await this.vector.front()).to.be.equal(bytesD);
+          expect(await this.vector.back()).to.be.equal(bytesC);
+          expect(await getContent(this.vector)).to.have.all.members([ bytesD, bytesA, bytesB, bytesC ]);
```
frangio, line 53: Would be nice to write the tests in terms of JavaScript array operations. So we could check that the array is now `initialArray.unshift(bytesD)`, and so on.

test/utils/structs/Vector.test.js
```diff
@@ -0,0 +1,91 @@
...
+        it('back', async function () {
+          expectEvent(await this.vector.popBack(), 'OperationResult', { value: bytesC });
+
+          expect(await this.vector.front()).to.be.equal(bytesA);
+          expect(await this.vector.back()).to.be.equal(bytesB);
+          expect(await getContent(this.vector)).to.have.all.members([ bytesA, bytesB ]);
+        });
+      });
+
+      it('reset', async function () {
```
frangio, line 83:
```suggestion
      it('clear', async function () {
```

test/utils/structs/Vector.test.js
Amxx: Would you be using pure ethers for this file, or have a mix of ethers and truffle+web3js. Using ethers + waffle would be clearer, but it would add dependencies, and would make this file look very different from the other ones.

Maybe we can do a first PR with just the vector thing (and no custom error testing) and then do a second PR with the test migration to ethers (so we don't discuss two mostly orthogonal things in the same PR).

contracts/utils/structs/Vector.sol
```diff
@@ -0,0 +1,79 @@
...
+        return vector.data[idx];
+    }
+
+    function clear(Bytes32Vector storage vector) internal {
+        vector.begin = 0;
+        vector.end = 0;
+    }
+
+    function length(Bytes32Vector storage vector) internal view returns (uint256) {
+        unchecked {
```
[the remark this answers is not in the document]
Amxx: `begin` and `end` are only updated by ++ or --. They are not expected to overflow (it would take 2**127 calls for it to overflow).

a consequence is that any checks (like the safemath that solidity includes by default and safecast) would just be wasted gas.

For example, on a `pushX`/`popX` operation, unchecked saves ~100 gas per operation.

contracts/utils/structs/Vector.sol
```diff
@@ -0,0 +1,79 @@
...
+
+    struct Bytes32Vector {
+        int128 begin;
+        int128 end;
+        mapping(int128 => bytes32) data;
+    }
+
+    function pushBack(Bytes32Vector storage vector, bytes32 value) internal {
+        unchecked {
+            vector.data[vector.end++] = value;
```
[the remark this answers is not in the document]
Amxx: Unchecked saves 100gas, removing it would basically double the difference between mapping and array.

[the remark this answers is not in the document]
Amxx: AFAIK, unchecked does affect the ++ operation, and the mapping access is not affected whatsoever. That is why I thought this "one line" would be cleaner and in the send simpler then doing

```
int128 index = vector.end;
unchecked {
    vector.end = index + 1;
}
vector.data[index] = value;
```

contracts/utils/structs/DoubleEndedQueue.sol
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    function length(Bytes32Deque storage deque) internal view returns (uint256) {
+        unchecked {
+            return SafeCast.toUint256(int256(deque.end) - int256(deque.begin));
+        }
+    }
+
+    function empty(Bytes32Deque storage deque) internal view returns (bool) {
+        return deque.end <= deque.begin;
```
Amxx, line 88: == should be good enough, and more gas effective. (there is no <= opcode, so it might be compiled as `!(x > y)`

contracts/utils/structs/DoubleEndedQueue.sol
```diff
@@ -0,0 +1,90 @@
...
+        int128 backIndex;
+        unchecked {
+            backIndex = deque.end - 1;
+        }
+        return deque.data[backIndex];
+    }
+
+    function at(Bytes32Deque storage deque, uint256 i) internal view returns (bytes32 value) {
+        // int256(deque.begin) is a safe upcast
+        int128 idx = SafeCast.toInt128(int256(deque.begin) + SafeCast.toInt256(i));
```
Amxx, line 71: It will never happen in practice, but if the queue is longer then `type(int256).max`, this code will fail.

IMO, we will never really get any deque of length > 2**64, so I would argue that using SafeCast makes all operations more expensive, with no real security gains

contracts/utils/structs/DoubleEndedQueue.sol
```diff
@@ -0,0 +1,90 @@
...
+        int128 end; // exclusive: the last item is at data[end-1]
+        mapping(int128 => bytes32) data;
+    }
+
+    function pushBack(Bytes32Deque storage deque, bytes32 value) internal {
+        int128 backIndex = deque.end;
+        deque.data[backIndex] = value;
+        unchecked {
+            deque.end = backIndex + 1;
+        }
```
Amxx, line 21: IMO, the previous implementation
```
unchecked {
     vector.data[vector.end++] = value;
}
```
was way more easy to read/understand (same argument apply to all push and pop functions)

contracts/utils/structs/DoubleEndedQueue.sol
frangio: The gas difference is negligible in this case. (Unchecked on the other hand easily saves 100 gas.)

The reason for using `<=` was as a "defense in depth" measure in case storage ever gets corrupted... I think it makes sense to do these things.

contracts/utils/structs/DoubleEndedQueue.sol
Amxx: I definitely think we want the `+1` unchecked.

Putting everything in the unchecked block might make it cleaner to read (all the logic is indented at the same level)

frangio (approved):

frangio:
@Amxx Please review and merge if everything looks good.

PR #3162, opened 2022-02-03
Index `awardItem` at zero

### LINKED ISSUE #3123: docs v4.x: ERC721 and Counters example starts index at 1
I'm referring to the following example: https://docs.openzeppelin.com/contracts/4.x/erc721#constructing_an_erc721_token_contract

```sol
// contracts/GameItem.sol
pragma solidity ^0.8.0;

import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
import "@openzeppelin/contracts/token/ERC721/extensions/ERC721URIStorage.sol";
import "@openzeppelin/contracts/utils/Counters.sol";

contract GameItem is ERC721, ERC721URIStorage {
    Counters.Counter private _tokenIds;

    constructor() ERC721("GameItem", "ITM") {}

    function awardItem(address player, string memory tokenURI)
        public
        returns (uint256)
    {
        _tokenIds.increment();

        uint256 newItemId = _tokenIds.current();
        _mint(player, newItemId);
        _setTokenURI(newItemId, tokenURI);

        return newItemId;
    }
}
```

Here, unnecessarily, when initially calling `awardItem`, `increment` is called directly indexing the first game item at `1` and not zero. I mean there's nothing inherently wrong with that, but given that virtually any other default counter usually starts at zero, I found this behavior rather surprising.

In the case of `awardItem` a fix would be to delay `increment` until just before returning.

TimDaub:
- Fixes #3123

as discussed with @[frangio](https://github.com/frangio) here https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3123#issuecomment-1028037626

### DIFF
--- a/docs/modules/ROOT/pages/erc721.adoc
+++ b/docs/modules/ROOT/pages/erc721.adoc
@@ -30,12 +30,11 @@ contract GameItem is ERC721URIStorage {
         public
         returns (uint256)
     {
-        _tokenIds.increment();
-
         uint256 newItemId = _tokenIds.current();
         _mint(player, newItemId);
         _setTokenURI(newItemId, tokenURI);
 
+        _tokenIds.increment();
         return newItemId;
     }
 }


Amxx (approved):

frangio:
Thanks1

PR #3165, opened 2022-02-03
Add a canceller role to the TimelockController

### LINKED ISSUE #2980: Add CANCEL_ROLE to TimelockController
**🧐 Motivation**
In the forthcoming Threshold Network DAO, there's a council with veto power over proposals. The only way to allow this with current implementation of `TimelockController` would be to grant the council the `PROPOSER_ROLE`, but this council shouldn't be able to make proposals, according to our governance design. We need something like a `CANCEL_ROLE` that's different from `PROPOSER_ROLE` and `EXECUTION_ROLE`. This feature could also be a solution for the ["Proposer fight" problem](https://docs.openzeppelin.com/contracts/4.x/api/governance#timelock-proposer) that's discussed in the documentation.

**Discussion**
I wanted to do this by creating a contract that inherits from `TimelockController`, but since its variables and functions are marked as private it's being difficult. In particular, we need to further restrict the cancel operation to a different role than proposers, but the modifier in `cancel()` forces us to override it and re-implement it, and we encounter the problem of `_timestamps` being private.

Anyway, if you find this interesting, I'd love to help, since I'm working on that anyway.

Amxx:
Fixes #2980  

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry


contracts/governance/TimelockController.sol
```diff
@@ -68,14 +69,16 @@ contract TimelockController is AccessControl {
...
         _setRoleAdmin(PROPOSER_ROLE, TIMELOCK_ADMIN_ROLE);
         _setRoleAdmin(EXECUTOR_ROLE, TIMELOCK_ADMIN_ROLE);
+        _setRoleAdmin(CANCELLOR_ROLE, TIMELOCK_ADMIN_ROLE);
 
         // deployer + self administration
         _setupRole(TIMELOCK_ADMIN_ROLE, _msgSender());
         _setupRole(TIMELOCK_ADMIN_ROLE, address(this));
 
-        // register proposers
+        // register proposers (and give them cancellor role)
```
frangio, line 78:
```suggestion
        // register proposers and cancellers
```

We also need to explain in the constructor docs that the same array is used for both roles. And possibly rename the argument.

frangio (commented):
"Cancellor" is a typo, it should be "canceller".

In the changelog, we should explain that if an instance is upgraded the role needs to be set up (admin and members).

Amxx:
AFAIK, both cancellor and canceller are valid. One is American english, the other is British english.

frangio:
No, you may be confusing it with "chancellor"? "Cancellor" doesn't seem to exist at all.

Amxx:
I did put "Cancellor" [in my translator](https://dictionnaire.reverso.net/anglais-francais/cancellor), and got what I expected.

I'm going to change it just to be safe.

(and no, I'm not confusing chancellor Palpatine with cancellor Palpatine :P)

frangio:
I would expect the tests to set up _different_ proposer and canceller accounts, so we can separately test them. Note: it also makes sense to test that in the default constructor the proposer array is assigned canceller roles.

frangio:
This is good but needs a changelog entry.

frangio (approved):

PR #3167, opened 2022-02-04
Rename some ERC20 parameters to match the standard document.

MicahZoltu:
The variable naming was incorrect.  The source of the funds is *not* necessarily (and in most cases isn't) the sender of the transaction.  Also, this code has a `msgSender` which further adds confusion.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -160,12 +160,12 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
         address to,
         uint256 amount
     ) public virtual override returns (bool) {
-        address owner = _msgSender();
-        uint256 currentAllowance = _allowances[from][owner];
+        address spender = _msgSender();
+        uint256 currentAllowance = allowance(from, spender);
         if (currentAllowance != type(uint256).max) {
             require(currentAllowance >= amount, "ERC20: transfer amount exceeds allowance");
             unchecked {
-                _approve(from, owner, currentAllowance - amount);
+                _approve(from, spender, currentAllowance - amount);
             }
         }
 
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -148,25 +148,25 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
      *
      * Requirements:
      *
-     * - `sender` and `recipient` cannot be the zero address.
+     * - `source` and `recipient` cannot be the zero address.
      * - `sender` must have a balance of at least `amount`.
-     * - the caller must have allowance for ``sender``'s tokens of at least
+     * - the caller must have allowance for ``source``'s tokens of at least
      * `amount`.
      */
     function transferFrom(
-        address sender,
+        address source,
         address recipient,
         uint256 amount
     ) public virtual override returns (bool) {
-        uint256 currentAllowance = _allowances[sender][_msgSender()];
+        uint256 currentAllowance = _allowances[source][_msgSender()];
         if (currentAllowance != type(uint256).max) {
             require(currentAllowance >= amount, "ERC20: transfer amount exceeds allowance");
             unchecked {
-                _approve(sender, _msgSender(), currentAllowance - amount);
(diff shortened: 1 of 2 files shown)


Amxx:
Hello @MicahZoltu 

I'm not sure that has any impact at all, but if we should rename them then I'd suggest we use the `from`, `to`, `value` naming as used in the [reference document](https://eips.ethereum.org/EIPS/eip-20).

`_msgSender()` is our binding into the `Context` pattern (for relayed meta-tx support).

MicahZoltu:
Updated PR to use `to`/`from`, I agree that is a better naming scheme.  I'm familiar with the `_msgSender` pattern the OZ contracts use, but that doesn't change the confusion introduced in this function by having two senders, one of which is not actually the "sender" of anything.

DanielVF:
I love the `from`/`to`. Looks like the mini-docs above the function need to be updated. I've sent you a PR Micah. :D

Amxx:
For consistency, it would also make sense to do the same changes to transfer, approve, and all the standard functions.

MicahZoltu:
Updated the PR to improve consistency across methods.  This includes both renaming sender/recipient to from/to and *ALSO* renaming `_msgSender()` in some places to `owner` (via local variable).  The latter brings consistency across the approval methods and helps make the code easier to digest.

⚠️ Someone who understands `_msgSender()` better than I should validate that caching the result of `_msgSender()` in a local for the duration of these functions is not problematic!  I do not know if perhaps it is intentional that `_msgSender()` can change within a function.  If not, then I think the local is appropriate because it not only saves gas and improves readability, but it also reduces the chance that some derived class breaks something by having `_msgSender()` be mutable within a given call context.  If it *is* intentional that it is mutable I will undo those changes.

Amxx (approved):

Amxx:
@MicahZoltu I 100% approve caching _msgSender. This function is not expected to write/change state, and in case it would return different result when called from the same context, that would be breaking.

Caching not only fixes that, but it also might save some gas!

@frangio Can you double-check before we merge?

Amxx (approved):

contracts/token/ERC20/ERC20.sol
```diff
@@ -148,25 +150,26 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
...
     function transferFrom(
-        address sender,
-        address recipient,
+        address from,
+        address to,
         uint256 amount
     ) public virtual override returns (bool) {
-        uint256 currentAllowance = _allowances[sender][_msgSender()];
+        address spender = _msgSender();
+        uint256 currentAllowance = allowance(from, spender);
```
frangio, line 164: Note: This line contains an unrelated change, the allowance is now read through an (overrideable) function instead of directly from storage. We were planning to make this change and it's small enough that I think it's ok to include here.

frangio (approved):
LGTM. Thank you @MicahZoltu for the suggestion.

I'd love to clarify one thing though. I don't believe that argument names in EIPs are normative, even in events, so this was more about improving code clarity than fixing something that was incorrect. Do we disagree on this?

contracts/token/ERC20/ERC20.sol
Amxx: Yes, I included it for that reason

DanielVF:
@frangio, Yes, code clarity is the reason. The original kick off for this was a project (not using OZ) getting rekt by switching the order of their allowance lookups in one function. 

The OZ function for allowances was:

`uint256 currentAllowance = _allowances[sender][_msgSender()];` 

Which is a bit confusing, and could easily lead to errors when someone was copying and modifying it.

The rest of the PR just grew from there. :)

Amxx:
The name of event's arguments are important in terms of backward compatibility. Changing that could break things like the openzeppelin subgraphs

MicahZoltu:
> I'd love to clarify one thing though. I don't believe that argument names in EIPs are normative, even in events, so this was more about improving code clarity than fixing something that was incorrect. Do we disagree on this?

Daniel was correct, this PR was about improving clarity only.  The choice of to/from wasn't even my original suggestion.  😄

PR #3168, opened 2022-02-05
Allow Address library to handle "returnbomb"

k06a:
See https://github.com/nomad-xyz/ExcessivelySafeCall for the details

Implementation progress:
- [x] Call "returnbomb"
- [x] Static call "returnbomb"
- [ ] Delegate call "returnbomb" (not sure we need it because delegate calls usually happen to highly trustworthy destinations)

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -90,6 +90,21 @@ library Address {
 
     /**
      * @dev Same as {xref-Address-functionCall-address-bytes-}[`functionCall`], but with
+     * `gasLimit` and  `returnLimit` as max allowed amount to be returned from the call.
+     *
+     * _Available since v4.5._
+     */
+    function safeFunctionCall(
+        address target,
+        bytes memory data,
+        uint256 gasLimit,
+        uint256 returnLimit
+    ) internal returns (uint256, bytes memory) {
+        return safeFunctionCall(target, data, "Address: low-level call failed", gasLimit, returnLimit);
+    }
+
+    /**
+     * @dev Same as {xref-Address-functionCall-address-bytes-}[`functionCall`], but with
      * `errorMessage` as a fallback revert reason when `target` reverts.
      *
      * _Available since v3.1._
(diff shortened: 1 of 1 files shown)


k06a:
Seems like we should add both `gasLimit` and `returnLimit` to have really safe version of call/staticcall

frangio:
It's still not clear to me in what situations people should use this.

So far everything we have that has the word "safe" in it are things that users should prefer. (Well, ERC721.safeTransfer may be an exception...) But I don't think users should generally use this "excessively" safe version of calls.

Even in the case of Nomad's bridge I still haven't convinced myself that this mechanism is necessary. Have you seen the part of their bridge code where they use this? I understand that a contract can theoretically do a return bomb, that the end-user would pay for it, but... there are other ways to cause an out of gas error, so what is different about this one?

So far my thinking is that the course of action for OpenZeppelin Contracts regarding returnbombs, is that we may want to add a version of `Address.functionCall` (_not_ "safeFunctionCall") with an upper bound for returndata size, plus _maybe_ a warning in the "unbounded" version of functionCall.

frangio:
>what is different about this one?

Answering myself. One thing that's different about this attack from other ways to burn gas is that if you do `(bool ok, bytes res) = target.call{gas: G}(data)`, `G` is not really a limit to the amount of gas that the target may cause to consume. It should be correlated in some way though, but will not be a hard limit as intended.

I don't know if there are other things that make this attack special, but if this is the only thing it's an indication that we may not need the added protection to our `Address.functionCall` family of functions: none of them accepts a gas limit, so the contract would be vulnerable to other kinds of gas burning attacks. Or in other words, the contract doesn't consider them attacks.

k06a:
@frangio that’s correct, that is why we need both arguments: `gasLimit` and `returnLimit`.

frangio:
My point is rather that:
1. Protecting against returnbombs only makes sense if you care about limiting gas by the callee.
2. Our functionCall helpers don't have a gasLimit argument, so they should only be used when the threat model doesn't consider it a problem if the callee purposely burns gas.
3. In my opinion, most users do not work on applications where callee gas burning is a problem. In fact, limiting gas is a likely source of problems because of unstable gas costs. This leads me to think that we should not include any utilities that allow limiting gas for a call, and returnbomb protection is simply a more effective way of limiting gas.

Happy to hear arguments against my thinking. So far I believe this PR should be closed.

Amxx:
> I think limiting gas to some big values but not too huge is still reasonable, like 200k or 500k. For sure limiting to 5k-10k is not sustainable for the future hard forks.

I fear any limit that you could possibly think of will:
- not be enough for some users
- be just enough for others, and then break with some opcode repricing.

Hardcoding gas values is possibly one of the biggest no to me.

frangio:
Closing for now as protection against returnbombs seems very niche and not aligned with the current offering for the reasons explained above.

PR #3171, opened 2022-02-07
ERC4626

### LINKED ISSUE #3393: EIP-4626: Tokenized Vault Standard
**Motivation**

[EIP-4626](https://eips.ethereum.org/EIPS/eip-4626) introduces a standard API for tokenized Vaults representing shares of a single underlying [ERC-20](https://eips.ethereum.org/EIPS/eip-20) token. This standard is an extension of the ERC-20 token that provides basic functionality for depositing and withdrawing tokens and reading balances.

**Reference Implementations**

- See [Solmate ERC-4626](https://github.com/Rari-Capital/solmate/blob/main/src/mixins/ERC4626.sol): a minimal and opinionated implementation of the standard with hooks for developers to easily insert custom logic into deposits and withdrawals.

- See [Vyper ERC-4626](https://github.com/fubuloubu/ERC4626): a demo implementation of the standard in Vyper, with hooks for share price manipulation and other testing needs.

@Amxx from OpenZeppelin has already opened a draft PR currently [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3171). I essentially opened this issue in order to track the development for anyone searching for this new ERC-4626.

Amxx:
Fixes #3393 

[Reference EIP](https://eips.ethereum.org/EIPS/eip-4626#methods).

We still have not decided the scope of customization we want to include in this implementation (fees?)
Should ERC165 be included? The ERC does not mention it.

Feel free to discuss/ask for features.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry
- [ ] Check ERC compliance


contracts/token/ERC20/extensions/draft-ERC4646.sol
```diff
@@ -0,0 +1,127 @@
...
+    function _sharesToAssets(uint256 shares) internal view virtual returns (uint256) {
+        return totalAssets() == 0 && totalSupply() == 0
+            ? shares * (10 ** _asset.decimals()) / (10 ** decimals())
+            : shares * totalAssets() / totalSupply();
+    }
+
+    function _assetsToShares(uint256 assets) internal view virtual returns (uint256) {
+        return totalAssets() == 0 && totalSupply() == 0
+            ? assets * (10 ** decimals()) / (10 ** _asset.decimals())
+            : assets * totalSupply() / totalAssets();
```
frangio, line 125: The contract feels almost too simple.

Can we imagine interesting customizations people will want to make?

Ideally there would be a simple way to override these conversions. What would you think of an `_exchangeRate()` function that returns a fraction expressed in as a pair `(uint256 numerator, uint256 denominator)`? The default would be `(totalSupply(), totalAssets())` (or the inverse... unclear) or something like that, properly handling 0.

DanielVF:
We have a rebasing currency, one thing we've done to checkup on exchange's support for it is to create a pair of accounts with matching amounts of our currency, one on-chain, and one on the exchange, then watch the balance changes between them and totals over time. 

I've taken a similar approach to testing the both the solmate and this OZ ERC4626 implementation as a wrapper to our rebasing currency. Using a mainnet fork, I've fired up several pairs of accounts - one account of each pair never touches the wrapper, and one use the wrapper. Then at each step, the designated one of random twin pair does a random deposit or withdraw to the wrapper. Then after each step, the underlying currency rebases up (sometimes by up to 5x). 

Looks like both solmate and OZ  ERC4626 handle this well. There's a small drift between twins from the rounding, but it seems to be the correct size considering the number of transactions, and the overall 10,000,000x'ing of the base supply over the course of the testing. This rounding dust flows nicely to the control account that is just holding the wrapped currency without transacting.

This doesn't really test anything beyond the deposit/withdraw math, but it's comforting to see

contracts/utils/math/Math.sol
```diff
@@ -3,10 +3,18 @@
 
 pragma solidity ^0.8.0;
 
+import "hardhat/console.sol";
```
DanielVF, line 6: Did you mean to leave this in, @Amxx ?

contracts/token/ERC20/extensions/draft-ERC4646.sol
poplexity: On the topic of interesting customizations, I added and am currently testing support for the native currency via additional deposit/withdrawl functions.  This assumes/requires that _asset be an implementation of WETH and uses it's deposit/withdrawl functions to automatically wrap/unwrap for the user so that _asset can still be ERC20.

Also am planning to add some unstaking duration so there is a configurable lockup period.

contracts/utils/math/Math.sol
Amxx: Definitelly not :)

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /** @dev See {IERC4262-previewRedeem} */
+    function previewRedeem(uint256 shares) public view virtual override returns (uint256) {
+        return _convertToAssets(shares, Math.Rounding.Down);
+    }
+
+    /** @dev See {IERC4262-deposit} */
+    function deposit(uint256 assets, address receiver) public virtual override returns (uint256) {
+        require(assets <= maxDeposit(receiver), "ERC4626: deposit more then max");
```
Oli8, line 81:
```suggestion
        require(assets <= maxDeposit(receiver), "ERC4626: deposit more than max");
```

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -0,0 +1,187 @@
...
+        _mint(receiver, shares);
+
+        emit Deposit(caller, receiver, assets, shares);
+
+        return shares;
+    }
+
+    /** @dev See {IERC4262-mint} */
+    function mint(uint256 shares, address receiver) public virtual override returns (uint256) {
+        require(shares <= maxMint(receiver), "ERC4626: mint more then max");
```
Oli8, line 98:
```suggestion
        require(shares <= maxMint(receiver), "ERC4626: mint more than max");
```

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -0,0 +1,187 @@
...
+        return shares;
+    }
+
+    /** @dev See {IERC4262-redeem} */
+    function redeem(
+        uint256 shares,
+        address receiver,
+        address owner
+    ) public virtual override returns (uint256) {
+        require(shares <= maxRedeem(owner), "ERC4626: redeem more then max");
```
Oli8, line 144:
```suggestion
        require(shares <= maxRedeem(owner), "ERC4626: redeem more than max");
```

contracts/token/ERC20/extensions/ERC4626.sol
DanielVF: The proposed change has the correct grammar. 👍

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -0,0 +1,187 @@
...
+        return assets;
+    }
+
+    /**
+     * @dev Internal convertion function (from assets to shares) with support for rounding direction
+     *
+     * Will revert if assets > 0, totalSupply > 0 and totalAssets = 0. That corresponds to a case where any asset
+     * would represent an infinite amout of shares.
+     */
+    function _convertToShares(uint256 assets, Math.Rounding direction) internal view virtual returns (uint256 shares) {
```
frangio, line 169: From the EIP:

> The preview methods return values that are as close as possible to exact as possible. For that reason, they are manipulable by altering the on-chain conditions and are not always safe to be used as price oracles. **This specification includes convert methods that are allowed to be inexact and therefore can be implemented as robust price oracles. For example, it would be correct to implement the convert methods as using a time-weighted average price in converting between assets and shares.**

Does this preclude bonding curves?

test/utils/math/Math.test.js
```diff
@@ -85,4 +87,94 @@ contract('Math', function (accounts) {
       expect(await this.math.ceilDiv(MAX_UINT256, b)).to.be.bignumber.equal(MAX_UINT256);
     });
   });
+
+  describe('muldiv', function () {
```
frangio, line 91: Please refactor these tests so that there's a single describe block for "rounds down", and the same for "rounds up". There's currently two separate ones.

frangio (commented):
Missing tests for ERC4626 and documentation for the contract.

contracts/token/ERC20/extensions/ERC4626.sol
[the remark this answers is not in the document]
Amxx: I don't see the conflict here. Can you explain how this part of the EIP would be incompatible with curves?

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -0,0 +1,187 @@
...
+        uint256 supply = totalSupply();
+        return
+            (assets == 0 || supply == 0)
```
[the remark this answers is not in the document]
Amxx: This is here for the case where `totalAssets() == 0`.

In that case, the exchange rate from assets to shares is not properly defined:
- if the vault is empty (supply == 0) then we fall back to the decimal rate → ok
- if the vault is not empty (supply > 0), every asset is work an infinite amount of shares
  - in that case, I do `0 * infinity = 0` (using the fallback to decimals as an escape) to avoid the division by 0.
  
  The whole thing would be rewritten
  
  ```
        return
            (supply == 0)
                ? assets.mulDiv(10**decimals(), 10**_asset.decimals(), direction)
                : (assets == 0)
                    ? 0
                    : assets.mulDiv(supply, totalAssets(), direction);
  ```

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -0,0 +1,187 @@
...
+
+        emit Withdraw(caller, receiver, owner, assets, shares);
+
+        return assets;
+    }
+
+    /**
+     * @dev Internal convertion function (from assets to shares) with support for rounding direction
+     *
+     * Will revert if assets > 0, totalSupply > 0 and totalAssets = 0. That corresponds to a case where any asset
```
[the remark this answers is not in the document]
Amxx: I also thought of type(uint).max ... but that felt wrong.

In my opinion, saying `I want to buy 1 Dai worth of &Dai`, when each &Dai is worth 0 (meaning you want to buy an infinite amount) does qualify as an "unreasonably large input" in the context. 

Also, I really think this is a corner case that should be avoided at all costs. For it to happen, you'd need to drain the vault. I'd expect the devs to be super careful about any vault draining mechanism and its consequences on share values.

contracts/token/ERC20/extensions/ERC4626.sol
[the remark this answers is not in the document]
frangio: My reading is that the `convert` methods should be usable as "robust price oracles" and even suggests implementing it as a TWAP. It's unclear to me if these are _shoulds_ or _coulds_.

I would not say that bonding curve prices are robust price oracles.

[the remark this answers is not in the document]
frangio: I posted a question the discussion thread about this to ask for opinions: https://ethereum-magicians.org/t/eip-4626-yield-bearing-vault-standard/7900/83

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -0,0 +1,187 @@
...
+     * would represent an infinite amout of shares.
+     */
+    function _convertToShares(uint256 assets, Math.Rounding direction) internal view virtual returns (uint256 shares) {
+        uint256 supply = totalSupply();
+        return
+            (assets == 0 || supply == 0)
+                ? assets.mulDiv(10**decimals(), 10**_asset.decimals(), direction)
+                : assets.mulDiv(supply, totalAssets(), direction);
```
robercano, line 174: I see that when the vault is empty of assets or shares the assets/shares ratio is set to 1:1. After that `totalAssets()` is used to calculate the ratio. I was considering the case where some amount of asset is sent to the contract (direct transfer) before the first `_mint()`. The first deposit would get 1:1, but subsequent deposits would get significantly less because of the existing initial balance. I understand that this functions will be more likely overridden by the integrator with the specific minting strategy, so I'm not sure if it should be fixed here, but I thought it was worth mentioning it

Amxx: I think this is something that we document in the tests

If a vault has some assets but no shares, then the value of shares is infinite, and the "logic" would mean that every share as an infinite value, and any assets put in by the user would give them 0 shares ... that is obviously not good.

So yes, the first depositor gets all the previous assets for free ... but it would also have got them for free if they had been transferred to the vault just after the deposit.

Also, I'm not sure how integrators are going to override this. I'm expecting that quite a few implementations don't override it at all.

robercano: About integrators, I just imagined a Vault taking into account special addresses, or bonus periods, or something along those lines. Thanks for the answer!

frangio: We shouldn't expect this to be overriden, the contract should have reasonable behavior out of the box.

>So yes, the first depositor gets all the previous assets for free ... but it would also have got them for free if they had been transferred to the vault just after the deposit.

The fact that these things commute makes me think the current behavior is reasonable.

frangio:
I'm thinking we should add fees. Isn't the contract otherwise vulnerable to sandwich attacks of the transaction where interest is accrued? I feel like I must be missing something there. Maybe the interest would just be small enough that a sandwich attack wouldn't really be profitable.

Amxx:
It's true that in case the interest is released at once and not using a flash bot, then a sandwich attack can drain the interests. We should document best practices, in particular, if the interest is per block (like vesting) then this attack is not possible.

If we have a clear way of formulating fees, then I would agree we can add it, but it not clear to me that the fees are that simple to set:
- fees on buy? fees on sell? both?
- percentage fees? flat fees?

AFAIK, the fees are as simple as "override the previewXxx" function.

frangio:
It's not so simple because you need to override the two pairs of functions.

robercano:
It seems to me that it would be enough to have two internal calculateBuyFee
and calculateSellFee that receive the amount being bought or sold. Then
integrators could override this to provide the required fee scheme.

frangio:
On the question of fees we've decided to release initially without fees. They may be added later.

test/token/ERC20/extensions/ERC20TokenizedVault.test.js
```diff
@@ -0,0 +1,593 @@
+const { BN, constants, expectEvent, expectRevert } = require('@openzeppelin/test-helpers');
+const { expect } = require('chai');
+
+const ERC20DecimalsMock = artifacts.require('ERC20DecimalsMock');
+const ERC20TokenizedVaultMock = artifacts.require('ERC20TokenizedVaultMock');
+
+const parseToken = (token) => (new BN(token)).mul(new BN('1000000000000'));
+const parseShare = (share) => (new BN(share)).mul(new BN('1000000000000000000'));
```
frangio, line 8: It's a little confusing that this hardcodes 12 and 18 decimals, given that there's also a test where the token has 18 decimals. I think these helpers should be defined inside a `describe` block where we use these decimals.

frangio:
I wanted to analyze the conversion functions to make sure I understood exactly the way we're pricing shares in the degenerate cases.

Our conversion functions are not symmetrical, they differ depending on whether the input is a number of assets or shares (e.g. mint vs deposit). This is our currently implemented behavior:




| <sub>A</sub> ↗ <sup>S</sup> | Z | NZ
| - | - | - |
| Z | 1 | ! |
| NZ | 1 | % |

| <sub>A</sub> ↙ <sup>S</sup> | Z | NZ
| - | - | - |
| Z | 1 | 0 |
| NZ | 1 | % |

(Sorry for the symbols but I liked the succinct presentation. A: totalAssets, S: totalSupply, A→S: convert to shares, A←S: convert to assets, Z: zero, NZ: non-zero, 1: decimals-adjusted 1-to-1 price, %: price based on vault ratio, !: revert)

Note that they differ in the case that there are zero assets and non-zero shares. I don't think this is right, I think we should aim for symmetry. I would suggest reverting independently of the conversion direction.

contracts/utils/math/Math.sol
```diff
@@ -38,6 +44,109 @@ library Math {
...
+            uint256 prod1; // Most significant 256 bits of the product
+            assembly {
+                let mm := mulmod(x, y, not(0))
+                prod0 := mul(x, y)
+                prod1 := sub(sub(mm, prod0), lt(mm, prod0))
+            }
+
+            // Handle non-overflow cases, 256 by 256 division.
+            if (prod1 == 0) {
+                require(denominator > 0);
```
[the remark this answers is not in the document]
Amxx: Remember we are inside an unchecked block. I'll have to check it does revert.

test/token/ERC20/extensions/ERC20TokenizedVault.test.js
Amxx: they are used everywhere expect for the `multiple mint, deposit, redeem & withdrawal` test.

Amxx:
This last commit adds quite a lot of sloads:
- `totalSupply()` can be expected to be dirty, so its only 100 gas each time
- `totalAssets()` can also be expected to be dirty, but includes a staticcall to another contract, which is not free.

It also checks `totalAssets()` in cases where `totalSupply()` is 0, which used to not be necessary (in that case the sload is cold).
All that to "fix" the behavior in a case where the vault is broken.

This case should never happen unless the vault is overridden in a way to enables slashing or loss of assets. I'm not sure we should include that by default. IMO this is something that should be added only if the vault mechanics are changed in a way that can cause it to be broken. **We** should not include it by default. **Other devs** should include it if their instance requires these additional checks

frangio (approved):
Ok, we're good.

PR #3194, opened 2022-02-16
Refactor governor testing

Amxx:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


test/governance/Governor.test.js
```diff
@@ -18,24 +16,47 @@ const Governor = artifacts.require('GovernorMock');
 const CallReceiver = artifacts.require('CallReceiverMock');
 
 contract('Governor', function (accounts) {
+  const helper = new GovernorHelper();
```
frangio, line 19: By declaring the instance here it's getting shared by all test runs even though they use different Governor instances (set in `beforeEach`).

But `GovernorHelper` could be stateful and be mixing state from different runs and that would lead to issues. So I think we should either:
- define `let helper;` and then assign `helper = new GovernorHelper()` in `beforeEach`
- or assign `this.helper = new GovernorHelper()`, or even `this.g = new GovernorHelper()` to make it nice and short?

test/governance/Governor.test.js
```diff
@@ -18,24 +16,47 @@ const Governor = artifacts.require('GovernorMock');
...
+
     await this.token.mint(owner, tokenSupply);
     await this.token.delegate(voter1, { from: voter1 });
     await this.token.delegate(voter2, { from: voter2 });
     await this.token.delegate(voter3, { from: voter3 });
     await this.token.delegate(voter4, { from: voter4 });
+    await this.token.transfer(voter1, web3.utils.toWei('10'), { from: owner });
+    await this.token.transfer(voter2, web3.utils.toWei('7'), { from: owner });
+    await this.token.transfer(voter3, web3.utils.toWei('5'), { from: owner });
+    await this.token.transfer(voter4, web3.utils.toWei('2'), { from: owner });
```
frangio, line 49: Did you consider putting this in the helper?

Amxx: the helper is not token aware ... 

I'm not sure it would make things clearer, because even if the helper knew about the token, you'd still need one line per allocation, right ?
Also, not sure how NFT would be handeled

test/governance/Governor.test.js
```diff
@@ -447,42 +448,34 @@ contract('Governor', function (accounts) {
...
     });
 
     it('missmatch #1', async function () {
-      this.helper.setProposal([
-        [ ],
-        [ web3.utils.toWei('0') ],
-        [ this.receiver.contract.methods.mockFunction().encodeABI() ],
-        '<proposal description>',
-      ]);
+      this.helper.setProposal({
```
Amxx, line 456: sometimes you do 
```this.helper.setProposal([{ ... }], 'descr');```
but here its
```this.helper.setProposal({ ... }, 'descr');```

how so ?

File: test/helpers/governance.js [hunk 4]
```diff
@@ -31,149 +31,168 @@ class GovernorHelper {
...
+   */
+  setProposal (actions, description) {
+    let targets, values, signatures, data, useCompatibilityInterface;
+
+    if (Array.isArray(actions)) {
+      useCompatibilityInterface = actions.some(a => 'signature' in a);
+      targets = actions.map(a => a.target);
+      values = actions.map(a => a.value || '0');
+      signatures = actions.map(a => a.signature || '');
+      data = actions.map(a => a.data);
```
Amxx, line 157: add `|| '0x'` ?

test/helpers/governance.js
```diff
@@ -31,149 +31,168 @@ class GovernorHelper {
...
+    } else {
+      useCompatibilityInterface = Array.isArray(actions.signatures);
+      ({ targets, values, signatures = [], data } = actions);
```
Amxx, line 160: there is no default (fallback) for value and data. IMO we should have it

frangio (approved):

Issue #2901, opened 2021-10-05
Initializer must be reusable for upgrade Initializers


a-d-j-i:
**🧐 Motivation**
All upgradeables are taking some storage space to limit the execution if the Initializer twice. 
When you upgrade a contract you need to initialize the new variables and is a good idea to reuse the storage used by Initializer.

**📝 Details**
I created an example of the main idea in:  https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/pull/125/files

a-d-j-i:
Ok, this was just an example, I will not insist with this PR.
I just want the functionality. 
Finally maybe patch or patch Level is a possible name.
Thanks!!!

a-d-j-i:
Ok, this was just an example, I will not insist with this PR.
I just want the functionality. 
Finally maybe patch or patch Level is a possible name.
Thanks!!!

PR #3195, opened 2022-02-16
Add `event Initialized()` to Initializable

kevupton:
Copied over from https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/pull/148

Added an event for an initializable contract. To allow for easy identification of when a contract has been initialized.

Suggestion for tracking when a contract has been initialized. 

This is helpful in cases like when using `TheGraph` when we want to do some action on initialize. Such as reading the data and saving it. As some of the data may not be available on construct for example.

Could potentially use `event Initialized(address sender);` with the sender as well. However, thought it might be better to just keep it simple. 

Not sure if there are any negative side effects of this.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/proxy/utils/Initializable.sol
+++ b/contracts/proxy/utils/Initializable.sol
@@ -33,6 +33,12 @@ import "../../utils/Address.sol";
  * ====
  */
 abstract contract Initializable {
+
+    /**
+     * @dev Triggered when the contract has been initialized.
+     */
+    event Initialized();
+
     /**
      * @dev Indicates that the contract has been initialized.
      */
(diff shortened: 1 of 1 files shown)


frangio:
Thanks for the suggestion @kevupton. This may make sense, I'm just not fully convinced yet.

> This is helpful in cases like when using `TheGraph` when we want to do some action on initialize. Such as reading the data and saving it. As some of the data may not be available on construct for example.

Generally the initializer is called during construction of the proxy, or at least in the same transaction, so I'm not fully convinced by this motivation. @Amxx What are your thoughts on this?

One concern that I have is how this will interact with #2901: we will likely have a `reinitializer(n)` modifier, so we may want this event to reflect that with an extra uint argument.

PR #3200, opened 2022-02-18
Support more efficient merkle proofs through calldata

### LINKED ISSUE #3198: Merkle Proof gas cost may be reduced in some cases.
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/eae238417862b4797e36fbff928d906bb9242261/contracts/utils/cryptography/MerkleProof.sol#L28

In the normal mode of operation a Merkle proof is passed into a contract by a external party. As such there is little benefit to copying the Merkle proof to memory before invoking `_efficientHash`. Therefore gas may be saved if in this case calldata is used instead of memory for the proof variable. Recommend either modifying the contract or offering a variant that utilizes calldata.

Cr0wn-Gh0ul:
Fixes #3198  

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
This PR as proposed is a breaking change. I was proposing that we should have a new set of functions where the arguments are `calldata`. Though I see now that this would result in a lot of code duplication. It's not a lot of code so it may be worth doing anyway.

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -55,6 +73,31 @@ library MerkleProof {
...
 
+    /**
+     * @dev Returns the rebuilt hash obtained by traversing a Merkle tree up
+     * from `leaf` using `proof`. A `proof` is valid if and only if the rebuilt
+     * hash matches the root of the tree. When processing the proof, the pairs
+     * of leafs & pre-images are assumed to be sorted.
+     *
+     * Updated to use calldata
+     *
+     * _Available since v4.4._
```
Amxx, line 84: THis should be 4.7, and should also be added to the `verifyCall`

frangio:
This is close to done but we want to rename `verifyCall` to `verifyCalldata`, `processProofCall` to `processProofCalldata`.

frangio (approved):

PR #3209, opened 2022-02-21
Implementation suggestion!

guilhermelim:
This is an implementation suggestion that allows other users to be able to access the functions marked with the nonReentrant modifier, while the reentrancy remains locked to whoever is executing a transaction.

### DIFF
--- a/contracts/security/ReentrancyGuard.sol
+++ b/contracts/security/ReentrancyGuard.sol
@@ -1,63 +1,26 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts v4.4.1 (security/ReentrancyGuard.sol)
+// https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/ReentrancyGuard.sol
 
 pragma solidity ^0.8.0;
 
-/**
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @guilhermelim 

The proposed change is a major breaking change. Also, there are many cases where the proposed implementation would not be an effective protection.

As a general rule of thumb, before making a breaking PR like that, you should create an issue to discuss your use case and identify the best possible solution (which might include creating a new contract).

frangio:
The main issue here is that this is not at all an effective protection, it can be very easily bypassed.

guilhermelim:
> The main issue here is that this is not at all an effective protection, it can be very easily bypassed.

I have a situation where I am forced to make a transfer immediately. I have no way to reset my bankroll, because in fact I don't even register the payment amount.

I'm looking for a solution. And I noticed that ReentrancyGuard has the big problem of preventing everyone from calling the function while a user doesn't finish his transaction.

I would like to find a smart and efficient solution to the problem. It is possible to create a ReentrancyGuard.sol without running the risk of other users receiving the message "ReentrancyGuard: reentrant call"

frangio:
Sorry I don't understand the problem, can you share more context?

PR #3210, opened 2022-02-21
Add ERC1155URIStorage

### LINKED ISSUE #3206: ERC721MetadataMintable missing from v3.0.0
**💻 Environment**
- ✅ OpenZeppelin v2.5.0
- ❌ OpenZeppelin v3.0.0

**📝 Details**

In OpenZeppelin v2 there used to be a [ERC721MetadataMintable extension containing a mintWithTokenURI method](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5fe8f4e93bd1d4f5cc9a6899d7f24f5ffe4c14aa/contracts/token/ERC721/ERC721MetadataMintable.sol#L19) that allowed you to pass the tokenUri when minting a new token.
In v3 and v4 this ERC721 extension seems to be missing. Unfortunately I was not able to figure out why it was removed. The [v3.0.0 release notes](https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v3.0.0) fail to mention this change as well. I didn't find any answer in the [forum](https://forum.openzeppelin.com/) neither.
Is there any particular reason for its removal?

takahser:
Fixes #3206

- add a new extension for `ERC1155` called `ERC1155URIStorage` that implements a `ERC721`-style `_setTokenURI` and `tokenURI` behaviors.

#### PR Checklist

- [x] Tests
- [x] Documentation (inline)
- [x] Changelog entry


Amxx:
Hello @takahser 

You are depending on `ERC1155Supply` instead of `ERC1155`, yet it doesn't look like you are using the supply features. Why is that ?

contracts/token/ERC1155/extensions/ERC1155URIStorage.sol
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @dev Sets `_tokenURI` as the tokenURI of `tokenId`.
+     *
+     * Requirements:
+     *
+     * - `tokenId` must exist.
+     */
+    function _setTokenURI(uint256 tokenId, string memory _tokenURI) internal virtual {
+        require(exists(tokenId), "ERC1155URIStorage: URI set of nonexistent token");
+        _tokenURIs[tokenId] = _tokenURI;
```
Amxx, line 51: You need to emit `event URI(string _value, uint256 indexed _id);` as described in the [ERC](https://eips.ethereum.org/EIPS/eip-1155)

takahser:
> You are depending on `ERC1155Supply` instead of `ERC1155`, yet it doesn't look like you are using the supply features. Why is that ?

This is because this extension depends on the ['`exists`' method](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.5.0/contracts/token/ERC1155/extensions/ERC1155Supply.sol#L29) that's located on the `ERC1155Supply` contract.

contracts/token/ERC1155/extensions/ERC1155URIStorage.sol
takahser: Good catch! Thanks for that, I added a new [commit](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3210/commits/0cbffd5be0ef7aa88b4380d80e081b0c6545f2e3).

Amxx:
I'd personnally remove the `exist` check, and drop the dependency on ERC1155Supply.

@frangio any opinion on that ?

frangio:
This has come up before. Perhaps `exists` should be in ERC1155 itself... (with a default implementation returning true)

For now I'm ok with dropping the exists check in this PR.

contracts/token/ERC1155/extensions/ERC1155URIStorage.sol
```diff
@@ -0,0 +1,50 @@
...
+abstract contract ERC1155URIStorage is ERC1155 {
+    using Strings for uint256;
+
+    // Optional mapping for token URIs
+    mapping(uint256 => string) private _tokenURIs;
+
+    /**
+     * @dev {IERC721Metadata-tokenURI} token URIs for ERC1155 tokens.
+     */
+    function tokenURI(uint256 tokenId) public view virtual returns (string memory) {
```
Amxx, line 22: `tokenURI` is not part of ERC1155. 

IMO you should not define a new function but override the existing `uri` function.

this means using `string memory base = super.uri(tokenId);`

takahser: If I did that, how would I get the base uri (that's returned by the current `uri` method) though?

With the current implementation we can support URIs of the type

1. `https://token.com/1` as well as
2. `https://token.com/one` as well as
3. `ipfs://someIpfsHash`

After overwriting the `uri` method with the new functionality we could only properly support 1 & 3, but not 2, right?
I mean, technically one could still set the "base" uri but not separately request it which seems to me quite inconvenient... What do you think?

Amxx: First things first. The ERC1155 interface documents the `uri(uint256)` method, but not `tokenURI` ... it the the first that is going to be used/called by all frontend/indexing services ... so that is what should be used. `tokenURI` is not part of this standard, so we just should expect anyone to use it.

That being said, my proposal is to do 

```
    function uri(uitn256 tokenId) public view virtual override returns (string memory) {
        string memory _tokenURI = _tokenURIs[tokenId];
        string memory base = super.uri(tokenId);

        // If there is no base URI, return the token URI.
        if (bytes(base).length == 0) {
            return _tokenURI;
        }
        // If both are set, concatenate the base URI and tokenURI (via abi.encodePacked).
        if (bytes(_tokenURI).length > 0) {
            return string(abi.encodePacked(base, _tokenURI));
        }

        // If there is a base URI set but no tokenURI, return the base URI.
        return base;
    }
```

This leaves multiple usecases:

- both `ERC1155._uri` and `ERC1155URIStorage._tokenURIs[tokenId]` are empty:
  → the result is empty
- `ERC1155._uri` is not empty and `ERC1155URIStorage._tokenURIs[tokenId]` is empty:
  → the result is `ERC1155._uri` (transparent)
- `ERC1155._uri` is empty and `ERC1155URIStorage._tokenURIs[tokenId]` is not empty:
  → the result is `ERC1155URIStorage._tokenURIs[tokenId]` (we set a token specific uri)
- both `ERC1155._uri` and `ERC1155URIStorage._tokenURIs[tokenId]` are not empty:
  → the result is the concatenation of both
  
As for your examples:

- `https://token.com/1`, `https://token.com/2`, `https://token.com/3`
→ you should not use `ERC1155URIStorage`, you should set the base uri to `https://token.com/` and override `uri()` to do 
```string.concat(super.uri, String.toString(tokenId))```
  
- `https://token.com/one`, `https://token.com/two`, `https://token.com/tree`
→ you should use  `ERC1155URIStorage`, set the base uri to `https://token.com/` and use the URIStorage feature to store "one", "two", "tree"

- `ipfs://someIpfsHash`, `ipfs://someotherIpfsHash`, ...
→ you could use `ERC1155URIStorage`, set the base uri to `ipfs://` and use URIStorage to manage the hashes

Comment (takahser) [reply contracts/token/ERC1155/extensions/ERC1155URIStorage.sol:22]: @Amxx Fair enough, thanks for elaborating in such detail and adequacy. 👍

I guess the only case we didn't yet cover is the usage of a combo of different protocols, such as `https://` and `ipfs://`, e.g. `https://token.com/abc123` and `ipfs://someIpfsHash`.
Here we could just leave the base `ERC1155._uri` empty but if anybody were to change it at some point it would invalidate all the existing token uris. That being said, that's a limitation I can live with.

I'm going to update the code according to your proposal.

**Comment by frangio** [comment]
I don't think it's right to use `super.uri` as a base URI and use concatenation. `super.uri` should be seen like a "global" URI and `_setTokenURI` should act like an override.

So, someone could set global URI as `ipfs://abcd/{id}.json` and then set an override `ipfs://efgh` for a particular id.

I think the base URI mechanism we have in ERC721 makes less sense here because there is already a built in base URI-like mechanism built into the ERC. So just the global URI plus a storage-based override sounds like enough.

File: contracts/token/ERC1155/extensions/ERC1155URIStorage.sol (hunk 2 above)
Comment (takahser) [reply contracts/token/ERC1155/extensions/ERC1155URIStorage.sol]: @Amxx Do you know how to resolve [this error](https://github.com/OpenZeppelin/openzeppelin-contracts/runs/5295895376?check_suite_focus=true#step:7:13)? What I tried (not committed) was:

```solidity
// file: ERC1155URIStorageMock.sol
contract ERC1155URIStorageMock is ERC1155Mock, ERC1155URIStorage {
    constructor(string memory uri) ERC1155Mock(uri) {}

    function uri(uint256 tokenId) public view virtual override returns (string memory) {
        ERC1155URIStorage.uri(tokenId);
    }

    function setTokenURI(uint256 tokenId, string memory _tokenURI) public {
        _setTokenURI(tokenId, _tokenURI);
    }
}
```

However, that'd fail at `override` with:

> Function needs to specify overridden contracts "ERC1155" and "ERC1155URIStorage".

takahser:
> I think the base URI mechanism we have in ERC721 makes less sense here because there is already a built in base URI-like mechanism built into the ERC. So just the global URI plus a storage-based override sounds like enough.

In that case, the following scenario wouldn't be supported:

- `ERC1155._uri` set to `https://token.com/`
- token uri set to `one`
- concatenated uri results in `https://token.com/one`

Hence, I think @Amxx [proposal](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3210#discussion_r812382989) provides more flexibility.

frangio:
If we want that behavior, I think we need a separate `_baseURI` variable. But again, I don't think it's necessary because the ERC has a built in mechanism for this.

You may as well just publish the metadata in `https://token.com/0...01` instead of `https://token.com/one`.

takahser:
> If we want that behavior, I think we need a separate _baseURI variable. But again, I don't think it's necessary because the ERC has a built in mechanism for this.

Well, don't we already have a `_baseURI` variable, although it's actually called [`_uri`]( https://github.com/OpenZeppelin/openzeppelin-contracts/blob/f6b614a7c5d72fdbb42e3ea0a1a886389e9c265d/contracts/token/ERC1155/ERC1155.sol#L30):

```solidity
// Used as the URI for all token types by relying on ID substitution, e.g. https://token-cdn-domain/{id}.json
string private _uri;
```

> You may as well just set uri to `https://token.com/{id}` and publish the metadata in `https://token.com/0...01` instead of `https://token.com/one`.

While I agree that this should be sufficient in most cases, there might be some edge cases where you'd want to have this extra flexibility, e.g. because you have to support a legacy system. Or maybe you want to have different locations on the same base uri for different categories of NFTs. I'm sure there will be more edge cases.

Amxx:
ERC1155 includes a "base" variable (`_uri`) which can be set using the internal `_setURI(string)` method.

IMO, for naming consistency, the setter in ERC155URIStorage should probably be `_setURI(uint256, string);`

takahser:
> ERC1155 includes a "base" variable (`_uri`) which can be set using the internal `_setURI(string)` method, and read using `super.uri()`
> 
> IMO, for naming consistency, the setter in ERC155URIStorage should probably be `_setURI(uint256, string);`

The problem I see with that approach is that it'd make it impossible to change the base `IERC1155._uri`.
Please note that the constructor [expects a `string memory uri_` argument](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/f6b614a7c5d72fdbb42e3ea0a1a886389e9c265d/contracts/token/ERC1155/ERC1155.sol#L32-L37):

```solidity
    /**
     * @dev See {_setURI}.
     */
    constructor(string memory uri_) {
        _setURI(uri_);
    }
```

Amxx:
Why would it make impossible to change the base URI ?

```
_setURI(string) // this would changes the base URI
_setURI(tokenId, string) // this would changes the "extra" URI for tokenId
```

takahser:
@Amxx Apologies, for some reason, I ignored the fact that the two methods have different signatures. Ofc you're right, I updated the PR. 👍

Amxx:
> If we want that behavior, I think we need a separate `_baseURI` variable. But again, I don't think it's necessary because the ERC has a built in mechanism for this.
> 
> You may as well just set uri to `https://token.com/{id}` and publish the metadata in `https://token.com/0...01` instead of `https://token.com/one`.

We could possibly write this override mechanism as

```
function uri(uint256 tokenId) public view virtual override returns (string memory) {
    string memory _tokenURI = _tokenURIs[tokenId];
    
    return bytes(_tokenURI).length > 0
        ? _tokenURI
        : super.uri(tokenId);
}
```

IMO that would fit most usecases I see. @takahser what do you think?

takahser:
@Amxx tbh I'd prefer the current implementation since it supports combining the `ERC1155. uri` with the `ERC1155URIStorage._tokenURIs[tokenId]`. That's not the case with your suggestion (fee free to correct me if I'm wrong). Is there a particular reason you don't want to support this use case?

Amxx:
> @Amxx tbh I'd prefer the current implementation since it supports combining the `ERC1155. uri` with the `ERC1155URIStorage._tokenURIs[tokenId]`. That's not the case with your suggestion (fee free to correct me if I'm wrong). Is there a particular reason you don't want to support this use case?

Fell, technically, if you wanted to concatenate both you could do that.

```
Contract X is ERC1155URIStorage {
    
    function uri(uint256 tokenId) public view virtual override returns (string memory) {
        string memory base = ERC1155.uri(tokenId);
        string memory extra = ERC1155URIStorage.uri(tokenId);
        
        return keccak256(base) == keccak256(extra) 
            ? base
            : string.concat(base, extra);
    }
}
```

@frangio might say this is terrible, and I would agree, but its technically possible.


This point is that, if the "override" is concatenated to the base, that has strong restriction on the base, and it basically kills the possibility of having a useful base, that would have a real meaning even when its not overridden with token specific info.

Amxx:
Not that ERC721 is designed differently, because the super call (if `_tokenURI` is not set) will append the tokenId string.

In ERC1155, this is achieved through the use of `{id}` in the base

takahser:
> Fell, technically, if you wanted to concatenate both you could do that.
> This point is that, if the "override" is concatenated to the base, that has a strong restriction on the base, and it basically kills the possibility of having a usefull base, that would have a real meaning even when it is not overridden with token specific info.

Sorry, you lost me there. Would you mind elaborating on that using an example?

frangio:
`super.uri()` should not be taken as a base URI. It should be taken as a default global URI, and it may include `{id}` in it to result in a different URI per token. `_setTokenURI` should override that default global URI entirely, it should not concatenate.

If we want concatenation, we need a separate `_baseURI` variable and `_setBaseURI` setter.

@takahser Feel free to implement either of these options but please do not use `super.uri` for concatenation. It should still be the fallback if a custom token URI is not set, though.

takahser:
@frangio that actually sounds logical to me. I'm going to implement the 2nd suggestion you made (separate base URI). 👍

takahser:
@frangio PR updated 👍

Amxx:
The super call on line 42 should be done on line 50, in the return. We don't want to have to pay for it if it turns out it's not needed.

@frangio do you like this baseUri mechanism, considering it is not used by the core erc1155 ?
I'm not really confortable...

takahser:
@Amxx @frangio PR updated, conflict resolved 👍
I'm not sure why _Test / slither_ is failing, it used to work before fixing the conflict. The conflict was just in the CHANGELOG.md so I doubt that this is the reason (correct me if I'm wrong).

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
  * `ERC1155`: Add a `_afterTokenTransfer` hook for improved extensibility. ([#3166](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3166))
  * `DoubleEndedQueue`: a new data structure that supports efficient push and pop to both front and back, useful for FIFO and LIFO queues. ([#3153](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3153))
  * `Governor`: improved security of `onlyGovernance` modifier when using an external executor contract (e.g. a timelock) that can operate without necessarily going through the governance protocol. ([#3147](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3147))
+ * `ERC1155URIStorage`: add a new extension for `ERC1155` that implements a `ERC721`-style `_setTokenURI` and `tokenURI` behavior.
```
takahser, line 10:
```suggestion
 * `ERC1155URIStorage`: add a new extension for `ERC1155` that implements a `ERC721`-style `_setTokenURI` and `tokenURI` behavior called `ERC1155URIStorage.setURI` and `ERC1155URIStorage.uri`, respectively.
```

takahser:
@Amxx @frangio

- [x] PR updated
- [x] new conflicts resolved
- [x] all checks passed

Any feedback would be appreciated:)

takahser:
@Amxx @frangio any updates on this one?

Amxx (approved):

Amxx (approved):

takahser:
Thx guys!

Amxx:
> Thx guys!

Thanks for your work and your patience

PR #3212, opened 2022-02-22
Access _allowances directly instead of via public virtual allowance

### LINKED ISSUE #3211: transferFrom calls public virtual allowance function breaking overridden behaviour
PR #3170 introduced a breaking change which is also a bug in that `transferFrom()` now calls (via `_spendAllowance()`) the overridable virtual function `allowance()` instead of accessing `_allowances` directly.

This is problematic where concrete implementations of ERC20 or ERC777 override virtual functions to adjust amounts so that the amounts exposed in the derived contract are a function of the amounts stored in the base contracts, and vice versa.

**💻 Environment**

@openzeppelin/contracts-upgradeable 4.5.1
Truffle

**📝 Details**

The general design of the ERC20 and ERC777 base contracts follows the pattern that public functions are virtual, with the default implementation that they call the corresponding internal virtual function. For example, `approve()` calls `_approve()`, and `burn()` calls `_burn()`. Other functions in the base contract use the internal implementations of these functions so that the behaviour is deterministic. Concrete implementations may choose to override the internal function to change the internal behaviour. But they may also choose to override the public function so that internal behaviour is unaffected.

This design is now broken by calling the public function `allowance()` in the `transformFrom()` function.

Given the thin nature of the `allowance()` function, the proper approach would be to follow a similar pattern to the public virtual `balance()` function and its underlying `_balances` variable. All reads of `_balances` are done directly. No function calls `balance()`. Similarly, no function should call `allowance()`. All reads of `_allowances` should be done directly.

**🔢 Code to reproduce bug**

```solidity
    function transfer(address recipient, uint256 amount) public virtual override returns (bool) {
        return super.transfer(recipient, f_(amount));
    }

    function allowance(address holder, address spender) public view virtual override returns (uint256) {
        return f(super.allowance(holder, spender));
    }

    function approve(address spender, uint256 value) public virtual override returns (bool) {
        return super.approve(spender, f_(value));
    }

    function transferFrom(address sender, address recipient, uint256 amount) public virtual override returns (bool) {
        return super.transferFrom(sender, recipient, f_(amount));
    }
```

`f()` is a function that modifies the amount passed to it and returns the result. `f_()` is the complement of that function.

NeoXtreem:
Fixes #3211

- Replace call to `allowance` with direct access of `_allowances` variable in ERC20
- Replace call to `allowance` with direct access of `_allowances` variable in ERC777

#### PR Checklist
- [x]  Tests
- [ ]  Documentation
- [x]  Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
  * `ERC1155`: Add a `_afterTokenTransfer` hook for improved extensibility. ([#3166](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3166))
  * `DoubleEndedQueue`: a new data structure that supports efficient push and pop to both front and back, useful for FIFO and LIFO queues. ([#3153](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3153))
  * `Governor`: improved security of `onlyGovernance` modifier when using an external executor contract (e.g. a timelock) that can operate without necessarily going through the governance protocol. ([#3147](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3147))
+ * `ERC20`: Replace call `allowance` function during call to `transferFrom` with direct access of `_allowances` variable to prevent undeterministic behaviour caused by concrete overrides. ([#3212](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3212))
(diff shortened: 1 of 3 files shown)


Amxx:
Hello @NeoXtreem 

I've answered this dirrectly in the related issue

frangio:
Closing as per the discussion in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3211.

PR #3219, opened 2022-02-24
Update ECDSA.sol

jayceftg:
To make the function more intuitive to some readers

Fixes #???? 

 To make the function more intuitive to some readers

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Thanks for the suggestion @jayceftg. The current code looks just fine to me.

PR #3223, opened 2022-02-26
Reduce contract size and deployment gas for onlyOwner modifier

### LINKED ISSUE #3222: Reduce contract size and deployment gas for onlyOwner modifier
**🧐 Motivation**
For inheriting contracts that use the onlyOwner modifier significant gas savings for deployment can be made by not inlining the require statement in every modified function.

**📝 Details**
When you add a function modifier, the code of that function is picked up and put in the function modifier in place of the _ symbol. If the same code is inlined multiple times, it adds up in size. Internal functions, on the other hand, are not inlined but called as separate functions. This means they are very slightly more expensive in run time but save a lot of redundant bytecode in deployment.

We could apply this efficiency in many other modifiers in OpenZeppelin as well.

uglyrobot:
Fixes #3222

For inheriting contracts that use the onlyOwner modifier significant gas savings for deployment can be made by not inlining the require statement in every modified function.

When you add a function modifier, the code of that function is picked up and put in the function modifier in place of the _ symbol. If the same code is inlined multiple times, it adds up in size. Internal functions, on the other hand, are not inlined but called as separate functions. This means they are very slightly more expensive in run time but save a lot of redundant bytecode in deployment.

From my testing with a sample contract this simple PR saves substantial gas per modifier call in contract size. This really adds up when for example in my contract I have 18 calls to onlyOwner!:

Savings per call at 10 Runs Optimization:
1 modifier call: 3883700-3857256 = 26,444 Gwei
2 modifier calls: 3892096-3857268 = 17,414 Gwei
5 modifier calls: 3917411-3857268 = 12,028 Gwei

Savings per call at 1000 Runs Optimization:
1 modifier call: 4311775-4258622 = 53,153 Gwei
2 modifier calls: 4329572-4258622 = 35,475 Gwei
5 modifier calls: 4382740-4258622 = 24,823 Gwei

#### PR Checklist

- [ ] Tests
- [ N/A] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -45,7 +45,8 @@ abstract contract Ownable is Context {
     }
 
     /**
-     * @dev Putting the requre in an internal function decreases contract size when onlyOwner modifier is used multiple times.
+     * @dev Putting the require in an internal function decreases contract size
+     * when onlyOwner modifier is used multiple times.
      */
     function _onlyOwner() internal view {
         require(owner() == _msgSender(), "Ownable: caller is not the owner");


Amxx:
Thank you @uglyrobot for this PR.

Could you please fix the lintting? Running `npm run lint:fix` should do the trick.

Amxx:
@frangio do we need a changelog entry for that ?

contracts/access/Ownable.sol
```diff
@@ -40,10 +40,18 @@ abstract contract Ownable is Context {
...
      */
     modifier onlyOwner() {
-        require(owner() == _msgSender(), "Ownable: caller is not the owner");
+        _onlyOwner();
         _;
     }
 
+    /**
+     * @dev Putting the require in an internal function decreases contract size
+     * when onlyOwner modifier is used multiple times.
```
frangio, line 49:
```suggestion
     * @dev Throws if the sender is not the owner.
```

contracts/access/Ownable.sol
```diff
@@ -40,10 +40,18 @@ abstract contract Ownable is Context {
...
+     */
+    function _onlyOwner() internal view {
```
frangio, line 51: The function should be virtual (we make all our functions virtual).

frangio:
Yes I'd say we should add a changelog entry.

contracts/access/Ownable.sol
frangio: `internal virtual` IMO.

Amxx:
Closed by #3347

PR #3224, opened 2022-02-27
Change time parameters to uint64 for consistency.

### LINKED ISSUE #3208: Governor: Use smaller integers for timing parameters
Some voting timing parameters are expressed as `uint256`, like `votingDelay` and `votingPeriod` in`GovernorSettings`, while the newly introduce `GovernorPreventLateQuorum` has a `votingExtension` parameter with type `uint64`. These should be consistent.

anantasty:
Governor: Inconsistent types for voting timing parameters #3208
Changed time parameters to uint64 for consistency

Fixes #3208 

Set the time parameters in GovrenorSettings.sol to match GovernorPreventLateQuorum

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+ * `GovernorSettings`: Changed the type of time variables to uint64 for consistency. Fixes #3208 . ([#3224](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3224))
  * `AccessControl`: add a virtual `_checkRole(bytes32)` function that can be overridden to alter the `onlyRole` modifier behavior. ([#3137](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3137))
  * `EnumerableMap`: add new `AddressToUintMap` map type. ([#3150](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3150))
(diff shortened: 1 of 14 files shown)


Amxx:
Hello @anantasty 

First: thank you for this PR.

I'm personally ok with this change. 

Still it is a breaking change: your PR needed to modify the `mocks/wizard/*.sol`. These are produced by the OpenZeppelin wizard, and are here to check the compatibility of changes to the governor contracts/modules. Basically, this change would require changes to the wizard, and would make existing code, produced by the wizard, not compatible with the updated version.

The fix is simple, and the error is not silent, so I don't think it's major. 

We will discuss that with the team

anantasty:
@Amxx Thanks for you review. What generates the mocks/wizard/*.sol files? Curious to know where the wizard code lives. i would be happy to add the changes there as well.

Amxx:
> Are you talking about this
> https://wizard.openzeppelin.com/#governor

Yes.

Dont worry, we'll update that if/when we merge your proposed changes.

anantasty:
Awesome. Thanks @Amxx

frangio:
These getters return `uint256` for compatibility with GovernorBravo. We should not do this breaking change.

PR #3225, opened 2022-02-27
Make Votes._getVotingUnits view

Amxx:
Make the `_getVotingUnits` function view, as it should have been.

This could be a breaking change, if anyone starts implementing a voting contract that uses `Votes` and that has a payable or non-payable `_getVotingUnits`. Do we need a breaking change notification for that?

#### PR Checklist

- [x] Tests
- [ ] Changelog


frangio:
This is a breaking change that we need to document in the changelog.

frangio (approved):

PR #3230, opened 2022-03-01
Add ERC721 and ERC1155 receiver support in Governor, Timelock

### LINKED ISSUE #3174: Governor, TimelockController, and similar, should be ERC721Holder
Contracts that can send arbitrary transactions through some protocol, like Governor and TimelockController, should contain the required hook to receive ERC721 tokens via `safeTransfer` (which we implement in `ERC721Holder`).

I think there is a similar hook for ERC1155.

ashwinYardi:
Fixes #3174 

The main purpose of the PR is to add support for receiving ERC721 and ERC1155 tokens in Governor, Timelock and MinimalForwarder contract. These contracts are responsible for sending transactions to and hence interacting with a lot of protocols.

I have added all the hooks in the main implementation contracts. I was thinking of creating a dedicated contracts and inherting them in the main implementation. However as per @Amxx 's comment, this might break the storage compatibility with respect to upgradable contracts. Please let me know if there is any better way to handle this. :)

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Hello @ashwinYardi

I've done some modification, particularly adding the interface inheritance and adding ERC165 support for the ERC1155Receiver (as requested by the ERC)

Notes for @frangio: the code would be simple if we added inheritance from the ERC721Holder & ERC1155Holder, but that would break the storage layout for the upgradeable version. We should do that properly in 5.0.

Amxx (dismissed):

Amxx (approved):

PR #3232, opened 2022-03-01
Allow the re-initialization of contracts

### LINKED ISSUE #2901: Initializer must be reusable for upgrade Initializers

Amxx:
Fixes #2901  

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


contracts/proxy/utils/Initializable.sol
```diff
@@ -91,7 +77,15 @@ abstract contract Initializable {
         _;
     }
 
-    function _isConstructor() private view returns (bool) {
-        return !Address.isContract(address(this));
+    function _setInitializeStep(uint8 version) private returns (bool) {
+        if (_initializing) {
+            require(version == 1 && !Address.isContract(address(this)), "Initializable: contract is already initialized");
```
frangio, line 82: Isn't this what we want? We should also add a comment explaining it. In fact we already have a comment but it's being deleted above.

```suggestion
            require(version > 1 || !Address.isContract(address(this)), "Initializable: contract is already initialized");
```

Amxx: IMO, this should not work outside a constructor.

The current version is 
```
        require(_initializing ? _isConstructor() : !_initialized, "Initializable: contract is already initialized");

        bool isTopLevelCall = !_initializing;
        if (isTopLevelCall) {
            _initializing = true;
            _initialized = true;
        }

        _;

        if (isTopLevelCall) {
            _initializing = false;
        }
```

So right now, an `initializer()` function will set both `_initializing` and `_initialized`
This means that is another such call is nested, we will require that we are in the constructor.

Today, initializer nesting is only allowed in constructor, which is what `&&` does. Using `||` would be a breaking change

frangio: Ok, sorry, I understand now. Can you add back the comment documenting this behavior?

frangio:
This is a sensitive piece of code. It needs comments explaining in depth what behavior it implements and why.

We also need to develop and document the alternative to `constructor() initializer {}`.

Amxx:
> We also need to develop and document the alternative to `constructor() initializer {}`.

Should we add a `lockImplementation` modifier that could be used for constructors?

frangio:
`lockImplementation` sounds too specific for upgrades. It also doesn't need to be a modifier, does it? What about something like this:

```
constructor() {
  _preventInitialize();
}
```

Amxx:
How should this function behave? Should it check that the contract is not yet initialized? Should it be allowed to run within the scope of an initializer?

Here are possible implementation
```
    function _preventInitialize() internal virtual {
        require(!_initializing && _initialized == 0, "Initializable: contract initializing of already initialized");
        _initialized = type(uint8).max;
        _initializing = false;
    }

    function _preventInitialize() internal virtual {
         if (_setInitializeStep(type(uint8).max)) {
            _initializing = false;
        }
    }
```

contracts/proxy/utils/Initializable.sol
```diff
@@ -74,7 +77,18 @@ abstract contract Initializable {
         _;
     }
 
-    function _isConstructor() private view returns (bool) {
-        return !Address.isContract(address(this));
+    function _setInitializeStep(uint8 version) private returns (bool) {
```
frangio, line 80: `_setInitializedVersion` sounds better to me.

frangio:
I think we want the second option:

```solidity
    function _preventInitialize() internal virtual {
        if (_setInitializeStep(type(uint8).max)) {
            _initializing = false;
        }
    }
```

The behavior that results is that after this function is called no more `initializer` or `reinitializer` functions will be callable. As a result, `onlyInitializing` functions will not be callable either.

Amxx:
not, it will be incorrect to write 
```
constructor() initializer {
    __preventInitialize();
}
```

contracts/proxy/utils/Initializable.sol
```diff
@@ -74,7 +99,29 @@ abstract contract Initializable {
...
-        return !Address.isContract(address(this));
+    /**
+     * @dev Locks the contract, preventing any futur re-initialization. This cannot be part of an initializer call.
+     * Calling this in the constructor of an contract will prevent any direct initialization & re-initialization.
+     * It is recommended to use that to lock "implementation" contract that are designed to be called through proxies.
+     */
+    function _preventInitialize() internal virtual {
+        if (_setInitializedVersion(type(uint8).max)) {
+            _initializing = false;
+        }
```
frangio, line 110: This code is simpler to understand the implications off, even though it's slightly more inefficient:

```suggestion
        _setInitializedVersion(type(uint8).max);
        _initializing = false;
```

contracts/proxy/utils/Initializable.sol
```diff
@@ -47,19 +70,21 @@ abstract contract Initializable {
      * @dev Modifier to protect an initializer function from being invoked twice.
```
frangio, line 70:
```suggestion
     * @dev Modifier to protect an initializer function from being invoked twice.
     * Equivalent to `reinitializer(1)`.
```

contracts/proxy/utils/Initializable.sol
```diff
@@ -47,19 +70,21 @@ abstract contract Initializable {
...
      */
     modifier initializer() {
-        // If the contract is initializing we ignore whether _initialized is set in order to support multiple
-        // inheritance patterns, but we only do this in the context of a constructor, because in other contexts the
-        // contract may have been reentered.
-        require(_initializing ? _isConstructor() : !_initialized, "Initializable: contract is already initialized");
-
-        bool isTopLevelCall = !_initializing;
+        bool isTopLevelCall = _setInitializedVersion(1);
```
frangio, line 73: After re-reviewing the code I don't think it makes sense to set `_initializing` as part of `_setInitializedVersion`. I know we did this to optimize bytecode size but I think for now we should move `_initializing = true` outside of the function, and we can consider optimizing it later and do benchmarks.

frangio (approved):

PR #3236, opened 2022-03-02
fix race condition in approve function

noobdreamers:
#fix   ---- >  
line number 136 :
function approve(address spender, uint256 amount) public virtual override returns (bool) {
        require(_allowances[_msgSender()][spender] <= 0, "use increase/decrease allowance to update");  // make condition here
        _approve(_msgSender(), spender, amount);
        return true;
    }

// The standard ERC20 implementation contains a widely-known racing condition in its approve function, wherein a spender is able to witness the token owner broadcast a transaction altering their approval and quickly sign and broadcast a transaction using transferFrom to move the current approved amount from the owner’s balance to the spender. If the spender’s transaction is validated before the owner’s, the spender will be able to get both approval amounts of both transactions.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -134,8 +134,8 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
      * - `spender` cannot be the zero address.
      */
     function approve(address spender, uint256 amount) public virtual override returns (bool) {
-        address owner = _msgSender();
-        _approve(owner, spender, amount);
+        require(_allowances[_msgSender()][spender] <= 0, "use increase/decrease allowance to update");
+        _approve(_msgSender(), spender, amount);
         return true;
     }
 


Amxx:
Hello @noobdreamers 

Your proposal causes issues more then it fix them.

You are doing `_allowances[_msgSender()][spender] <= 0`

First, we are dealing with unsigned integers, so `<= 0` means `== 0`. If we do that check, then we would prevent someone updating the allowance once it set ... unless is has been consumed. In particular, it would not be possible to remove an approval before its used ... so infinite approval would not be revocable. This is bad design, and it breaks the ERC20 standard.

We are aware of the possible approval issues, and that is why we have the [increaseAllowance and decreaseAllowance](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L169-L210) functions.

PR #3245, opened 2022-03-07
Procedural SafeCast.sol generation

### LINKED ISSUE #3159: Add SafeCast for all possible uintX types
**🧐 Motivation**
These methods are required when optimizing and tightly packing storage variables. We often use `uint40` or `uint48` for time and `uint216`/`uint208` for the rest slot value, `uint112` or `uint96` (*presented*) for storing token amounts etc

Amxx:
Fixes #3159

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry


scripts/generate/templates/toIntDownCast.js
```diff
@@ -0,0 +1,20 @@
+module.exports = length => [
```
frangio, line 1: Instead of `[...].join('\n')` we can just a big template string:


```javascript
module.exports = length => `\
    ...
`;
```

scripts/generate/run.js
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+
+const {
+  toInt,
+  toUint,
+  toIntDownCast,
+  toUintDownCast,
+} = require('./templates');
```
frangio, line 10: I would just define these functions inline, possibly at the bottom of the file.

scripts/generate/run.js
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env node
```
frangio, line 1: I'd rename this file to `scripts/generate/safecast.js`.

scripts/generate/run.js
```diff
@@ -0,0 +1,57 @@
...
+const fs = require('fs');
+
+const {
+  toInt,
+  toUint,
+  toIntDownCast,
+  toUintDownCast,
+} = require('./templates');
+
+function genSafeCast (path) {
```
frangio, line 12: For formatting, this script will come in handy: https://github.com/OpenZeppelin/contracts-wizard/blob/master/packages/core/src/utils/format-lines.ts

frangio:
Is this ready for review?

frangio:
Pending:
- [ ] Add a CI script to verify that the file committed to the repo is up to date with the generated file

scripts/generate/templates/SafeCast.js
```diff
@@ -0,0 +1,106 @@
...
+
+module.exports = format(
+  header.trimEnd(),
+  'library SafeCast {',
+  [
+    ...LENGTHS.map(size => toUintDownCast(size)),
+    toUint(256),
+    ...LENGTHS.map(size => toIntDownCast(size)),
+    toInt(256).trimEnd(),
+  ].flatMap(fn => fn.split('\n')),
```
frangio, line 104: The goal of `formatLines` was to not have to do this. I thought `formatLines` was going to take care of the newlines but it doesn't. Instead of using this line here, I'd suggest to adjust the helper function so that it does what we want. Maybe we need some other helper instead. The goal is to have a relatively generic helper that we can use here and in SafeCastMock.js and other places.

scripts/generate/format-lines.js
```diff
@@ -0,0 +1,17 @@
+function formatLines (...lines) {
+  return [...indentEach(0, lines)].join('\n') + '\n';
+}
+
+function *indentEach (indent, lines) {
+  for (const line of lines) {
+    if (line === '') {
+      yield '';
+    } else if (Array.isArray(line)) {
+      yield * indentEach(indent + 1, line);
```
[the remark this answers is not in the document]
Amxx: it doesn't pass linting if I remove the space

scripts/generate/run.js
```diff
@@ -1,7 +1,30 @@
...
+const format = require('./format-lines');
 
-// SafeCast.sol
-fs.writeFileSync('./contracts/utils/math/SafeCast.sol', require('./templates/SafeCast'));
-fs.writeFileSync('./contracts/mocks/SafeCastMock.sol', require('./templates/SafeCastMock'));
+function getVersion (path) {
+  try {
+    return fs
+      .readFileSync(path)
+      .toString()
```
frangio, line 10: The second argument specifies the encoding to get back a string, otherwise it will return a Buffer:

```suggestion
      .readFileSync(path, 'utf8')
```

scripts/generate/templates/SafeCast.js
```diff
@@ -1,8 +1,63 @@
+const { assert } = require('chai');
 const format = require('../format-lines');
 
-const header = `\
-// SPDX-License-Identifier: MIT
+const LENGTHS = Array(31).fill().map((_, i) => (i + 1) * 8).reverse(); // 248 → 8 (in steps of 8)
 
+const version = (selector, length) => {
```
frangio, line 6:
```suggestion
// Returns the version of OpenZeppelin Contracts in which a particular function was introduced.
// This is used in the docs for each function.
const version = (selector, length) => {
```

scripts/generate/templates/SafeCast.js
```diff
@@ -1,8 +1,63 @@
+const { assert } = require('chai');
```
frangio, line 1: It doesn't make sense to use Chai here because there's a native assert module. Chai is a testing library. This change should do:

```suggestion
const assert = require('assert');
```

test/utils/math/SafeCast.test.js
```diff
@@ -41,7 +41,7 @@ contract('SafeCast', async (accounts) => {
     });
   }
 
-  [8, 16, 32, 64, 96, 128, 224].forEach(bits => testToUint(bits));
+  Array(31).fill().map((_, i) => (i + 1) * 8).forEach(bits => testToUint(bits));
```
frangio, line 44: This pattern shows up often. Let's add a helper function `range(start, end, step)` or something like it.

frangio (dismissed):
Looks good!

frangio (approved):

frangio (approved):

PR #3251, opened 2022-03-09
Update ERC20.sol

jayceftg:
this kinda suits solidity's programming more

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Hello @jayceftg 

If we make it pure, the only pure function will be able to override it. We want this function to be overridable by view function (that could read the value from storage, or copy the value from another contract).

PR #3255, opened 2022-03-10
Add Datetime utilities (#1531)

### LINKED ISSUE #1531: DateTimeUtils - Helper logic to handle time
**🧐 Motivation**
Time is important and off-chain there are often cases where time related checks need to be performed. To handle this on chain it would be nice to have some utils that would encapsulate these checks.

There are many use cases for this:
- Assuming we have a multisignature wallet that should only require the approval of all owners if more than a specified amount should be transferred (see https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWalletWithDailyLimit.sol). In the linked contract this limit is always the "in the last 24h" but it would be interesting to map this to a calendar day.
- Assuming we want to perform recurring transfer (see https://groundhog.network/) it makes sense to limit these to calendar dates (On the 1st of November I pay my monthly zeppelin subscription, or on the 13th of November between 1pm and 2pm the company pays me my salary).

**📝 Details**
There is a library that is already implementing this logic (https://github.com/pipermerriam/ethereum-datetime), but this is not really up to date, nor audited.

Also it would be interesting to consider edge cases like leap seconds (at least mention why they are not implemented).

An api should include method to check if what year/ month/ day/ hour/ minute a certain timestamp is.

lesserhatch:
Add date and datetime utility functions to convert from civil date
time to Unix time.

Fixes #1531 

This PR adds datetime utility functions. The Datetime library provides utility functions convert human readable dates / times to Unix time.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/mocks/DatetimeMock.sol
+++ b/contracts/mocks/DatetimeMock.sol
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "../utils/Datetime.sol";
+
+contract DatetimeMock {
+    function date(
+        int32 _year,
+        uint8 _month,
+        uint8 _day
+    ) external pure returns (int256) {
+        return Datetime.date(_year, _month, _day);
+    }
+
+    function datetime(
+        int32 _year,
+        uint8 _month,
+        uint8 _day,
+        uint8 _hour,
+        uint8 _minute,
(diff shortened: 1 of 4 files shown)


ernestognw:
Closing since it's pretty old. If pursued, it'll likely require a different approach

PR #3257, opened 2022-03-10
feat: short circuiting optimisation ERC721

pcaversaccio:
This change assumes that direct transfers are much less than proxy transfers. Afaik there is no gas optimisation involved since the first check does not involve an additional `SLOAD` but in terms of logic it would make sense IMHO.

#### PR Checklist
- [ ] Tests
- [ ] Documentation
- [x] Changelog entry


Vectorized:
You need direct transfers to be way less common than proxy transfers for this to actually reduce expected gas costs. Like orders of magnitude less. 

Looking at the Etherscan for popular collections, I think direct transfers (e.g. giveaways, direct trading, distributing amongst own wallets) are still very common.

Amxx:
Hello @pcaversaccio.

I understand that there is a trend of looking for every possible gas optimization... but this goes too far.

- doing `spender == owner` is a simple cooperation that cost 3 gas.
- doing `isApprovedForAll(owner, spender)` provaly cost 2100 gas (sload from a cold slot).

What is more reasonable.
- in some cases, you pay ~3, if that is not successful fallback and pay a total of 2103+3
- in some cases, you pay 2103, if that is not successful fallback and pay a total of 2103+3

Even the transfer are most often for an operator ... I really don't think this optimization is reasonable in the general case.
If you think in your case it's better that way, you can override the function and do the switch.

PR #3259, opened 2022-03-10
Inherit ERC20Wrapper decimals from the underlying

### LINKED ISSUE #3258: ERC20Wrapper does not pass through underlying token's decimals()
The `ERC20Wrapper` contract is designed to create "wrapped" tokens based on some **underlying** ERC20 token. This should expose the basic properties of the underlying token wrapped on a one-for-one basis.

However, the `ERC20Wrapper` contract does not expose the result of the underlying token's `decimals()` method in its own `decimals()` implementation. Instead, it inherits the standard implementation from `ERC20.sol` which is hard-coded to simply output `18`. Hence an underlying token can have a different `decimals()` result than its wrapped counterpart.

In practice it is, of course, very unusual for tokens to have `decimals() != 18`, but in practice it can pose a problem for front-ends that respect the `decimals()` field of a token and expect its wrapped counterpart to use the same value. Somebody could create an ERC20 with weird `decimals()` and then the UI for wrapping would receive a different decimals value from the wrapped and underlying tokens. This could cause strange math errors that would confuse the user, and convince them that the wrapped token was malfunctioning. 

**💻 Environment**

Latest (in Git)

**📝 Details**

If this is worth fixing (and I don't know), the main issue here is with tokens that don't implement a decimals() method. This edge-case would need to be handled, but doesn't seem like a big deal.

**🔢 Code to reproduce bug**

1. Make a token that has decimals() != 18
2. Create an instance of an ERC20Wrapper based on this token.
3. Call decimals() on the original and underlying, they won't be the same.

Amxx:
Fixes #3258

#### PR Checklist

- [x] Tests
- [ ] Changelog entry

### DIFF
--- a/contracts/mocks/ERC20DecimalsMock.sol
+++ b/contracts/mocks/ERC20DecimalsMock.sol
@@ -18,4 +18,12 @@ contract ERC20DecimalsMock is ERC20 {
     function decimals() public view virtual override returns (uint8) {
         return _decimals;
     }
+
+    function mint(address account, uint256 amount) public {
+        _mint(account, amount);
+    }
+
+    function burn(address account, uint256 amount) public {
+        _burn(account, amount);
+    }
 }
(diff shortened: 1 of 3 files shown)


frangio:
Yes this makes sense to me.

PR #3268, opened 2022-03-16
Changed ERC721Holder.sol visibility to external.

### LINKED ISSUE #3266: ERC721Holder.sol -- onERC721Received function incorrectly has "public" visibility
Per the [ERC721 Spec](https://eips.ethereum.org/EIPS/eip-721), the `ERC721TokenReceiver` interface should include function `onERC721Received` with `external` visibility, e.g. `function onERC721Received(address _operator, address _from, uint256 _tokenId, bytes _data) external returns(bytes4);`

This function's implementation in the [ERC721Holder.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/utils/ERC721Holder.sol#L25) contract incorrectly uses `public` visibility.

[IERC721Receiver.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721Receiver.sol#L26) correctly uses the `external` visibility for this function.

mw2000:
Fixes #3266
Changes the visibility of the function to external as specified in the standard.

### DIFF
--- a/contracts/token/ERC721/utils/ERC721Holder.sol
+++ b/contracts/token/ERC721/utils/ERC721Holder.sol
@@ -22,7 +22,7 @@ contract ERC721Holder is IERC721Receiver {
         address,
         uint256,
         bytes memory
-    ) public virtual override returns (bytes4) {
+    ) external virtual override returns (bytes4) {
         return this.onERC721Received.selector;
     }
 }


Amxx:
Hello @mw2000 

When you make a function external, it is no longer possible to call `super` when overriding the function. Consequently, making it external would possibly break some usages.

What are the pros of making it public?

Issue #3155, opened 2022-02-01
ERC20.sol "decimals" variable description left in comments from prev versions


ivan-kyuchukov:
The decimals variable comment is left from previous versions:
![image](https://user-images.githubusercontent.com/10574951/152009599-9e527df3-e485-4766-b839-28cf3815db02.png)

**💻 Environment**

**📝 Details**

**🔢 Code to reproduce bug**

Amxx:
Hello @ivan-kyuchukov 

That is not a mistake.

This contract include a decimal function.
```
    function decimals() public view virtual override returns (uint8) {
        return 18;
    }
```

If someone wants to change the decimal, the process is to override the `decimals()` function. We left a comment here, because devs coming from a 3.x version would be looking to customize that in the constructor → this message tells them that the upgrade is done elsewhere.

frangio:
Maybe the documentation is confusing because it's in the constructor. We could move the part about `decimals` to the contract-level docs.

PR #3272, opened 2022-03-17
Erc20.sol "decimals" variable description left in comments from prev versions #3155

mw2000:
Fixes ##3155
Removed the default decimals comment from the constructor to avoid confusion.

### DIFF
--- a/contracts/token/ERC721/utils/ERC721Holder.sol
+++ b/contracts/token/ERC721/utils/ERC721Holder.sol
@@ -22,7 +22,7 @@ contract ERC721Holder is IERC721Receiver {
         address,
         uint256,
         bytes memory
-    ) public virtual override returns (bytes4) {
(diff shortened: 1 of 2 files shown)


contracts/token/ERC721/utils/ERC721Holder.sol
```diff
@@ -22,7 +22,7 @@ contract ERC721Holder is IERC721Receiver {
         address,
         uint256,
         bytes memory
-    ) public virtual override returns (bytes4) {
```
Amxx, line 25: This is unrelated with ERC20, and would break the overridability of the function.

Amxx:
Hello @mw2000 

I'm failling to see how 

> The default value of {decimals} is 18. To select a different value for {decimals} you should overload it.

is not relevant anymore. IMO it accurately describes the current behavior of the contract.

PR #3279, opened 2022-03-22
Enable the YUL IR pipeline when building with optimisations

Amxx:
Since [0.8.13](https://github.com/ethereum/solidity/releases/tag/v0.8.13), the YUL IR pipeline is no longer experimental.

This greatly impacts the compilation time when optimization is enabled (from ~20 to ~80sec on my machine), so most tests should continue to run on the non-optimized version.

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -33,7 +33,7 @@ const argv = require('yargs/yargs')()
     compiler: {
       alias: 'compileVersion',
       type: 'string',
-      default: '0.8.9',
+      default: '0.8.13',
     },
     coinmarketcap: {
       alias: 'coinmarketcapApiKey',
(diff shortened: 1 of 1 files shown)


frangio:
Is `viaIR: true` the recommended setting now?

Amxx:
From the 0.8.13 release note:

> Furthermore, compiling via the new Yul IR pipeline is now considered production ready.

That doesn't really answer te question, but that is all I could find

Amxx:
I've also noticed that turning the IR pipeline cause some contract to not compile (some unspecified error showed up)

I don't think we should enable by default, but we should check our contract do compile with it ... and then do some gas comparaison

frangio (approved):

PR #3280, opened 2022-03-22
Add a SafeERC20:safePermit function

### LINKED ISSUE #3145: Add a safePermit function for ensuring reverting calls to ERC20.permit()
**🧐 Motivation**
The Multichain/AnySwap exploits were caused by smart contracts incorrectly assuming that a failed `permit` call would always revert. This is similar to how a smart contract might assume that calls to `transfer` or `approve` will revert if they fail. For this, OpenZeppelin's `SafeERC20` library has the `safeTransfer`, `safeTransferFrom`, `safeApprove`, etc.
 
**📝 Details**
[ERC2612](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2612.md) `permit` functions do not return any value to indicate success, they only revert. However, a "phantom function"  call behaves exactly like a successful call, unless state is inspected.

ERC2612 tokens maintain and expose a mapping of nonces, and a successful call to `permit` MUST increment the nonce of the owner address. This state update can be observed to ensure that a permit function has successfully been called.

The suggested file for this would be `contracts/token/ERC20/utils/SafeERC20Permit.sol`.

```solidity
import "../extensions/draft-IERC20Permit.sol";

library SafeERC20Permit {

    function safePermit(
        IERC20Permit token,
        address owner,
        address spender,
        uint256 value,
        uint256 deadline,
        uint8 v,
        bytes32 r,
        bytes32 s
    ) internal {
        uint256 nonce = token.nonces(owner);
        token.permit(owner, spender, value, deadline, v, r, s);
        uint256 newNonce = token.nonces(owner);
        require (newNonce == nonce + 1, "SafeERC20Permit: permit did not succeed");
    }
}
```

Amxx:
Fixes #3145

Note: IMO it would be slightly better to use `abi.encodeCall` instead of `abi.encodeWithSelector`, but that would require a bump to 0.8.11

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


contracts/token/ERC20/extensions/draft-IERC20Permit.sol
```diff
@@ -11,7 +13,7 @@ pragma solidity ^0.8.0;
  * presenting a message signed by the account. By not relying on {IERC20-approve}, the token holder account doesn't
  * need to send a transaction, and thus is not required to hold Ether at all.
  */
-interface IERC20Permit {
+interface IERC20Permit is IERC20 {
```
frangio, line 16: I don't think this is necessary, can we remove it? I'm scared it might produce unforeseen breaking changes.

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -79,6 +80,25 @@ library SafeERC20 {
...
+        address owner,
+        address spender,
+        uint256 value,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) internal {
+        uint256 nonceBefore = token.nonces(owner);
+        _callOptionalReturn(
```
frangio, line 94: This doesn't make sense to me. `permit` doesn't have a return value, so there is no optional return. We should simply call `permit` directly.

test/token/ERC20/utils/SafeERC20.test.js
```diff
@@ -39,6 +46,120 @@ contract('SafeERC20', function (accounts) {
...
+          this.data.message.deadline,
+          27,
+          '0x71753dc5ecb5b4bfc0e3bc530d79ce5988760ed3f3a234c86a5546491f540775',
+          '0x0049cedee5aed990aabed5ad6a9f6e3c565b63379894b5fa8b512eb2b79e485d',
+        ),
+        'SafeERC20: permit did not succeed',
+      );
+    });
+
+    it('underlying token does not revert on reused or invalid signature', async function () {
```
frangio, line 133: I don't think this test is necessary. It's just testing the mock.

contracts/token/ERC20/extensions/draft-IERC20Permit.sol
Amxx: It enables passing IERC20Permit object to functions where an IERC20 is expected.

Also, ERC2612 states
> **Specification**
> Three new functions are added to the ERC-20 ABI:

So I read that as, ERC2612 extends ERC20, [just like ERC1363](https://eips.ethereum.org/EIPS/eip-1363)

> **Specification**
> Implementing contracts MUST implement the [ERC-1363](https://eips.ethereum.org/EIPS/eip-1363) interface as well as the [ERC-20](https://eips.ethereum.org/EIPS/eip-20) and [ERC-165](https://eips.ethereum.org/EIPS/eip-165) interfaces.
>
>> interface ERC1363 /* is ERC20, ERC165 */ {

And [ERC4626](https://eips.ethereum.org/EIPS/eip-4626)

> **Specification**
> All ERC-4626 tokenized Vaults MUST implement ERC-20 to represent shares.

test/token/ERC20/utils/SafeERC20.test.js
Amxx: Its proving that the revert is the result of the wrapper and that the tests above are indeed testing the right mechanism

frangio (dismissed):

frangio (approved):

PR #3298, opened 2022-03-27
Add `Math.modExp` and a `Panic` library

### LINKED ISSUE #1985: Modular exponentiation precompile wrapper
[EIP 198](https://eips.ethereum.org/EIPS/eip-198) introduced the modular exponentiation precompile, also known as "modexp".

It's a useful primitive for cryptographic operations, which we would like to wrap in a library so that developers can use it without having to write assembly.

One of the challenges is the fact that the precompile receives and returns bignums, i.e. numbers of arbitrary size. In order to provide a good interface, we might need to define a struct to represent bignum types.

mw2000:
Fixes #1985
Introduces the wrapper library for EIP 198 that @frangio mentioned in issue #1985. Code was taken from https://ethereum.stackexchange.com/questions/71565/verifying-modular-exponentiation-operation-in-etherum

Looking forward to code review and comments to add docs and Changelog entry.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/utils/math/ModularExponentiation.sol
```diff
@@ -0,0 +1,40 @@
...
+    function modExp(
+        uint256 b,
+        uint256 e,
+        uint256 m
+    ) internal returns (uint256) {
+        require(m != 0, "ModularExponentiation: Can't calculate for modulus equal to zero");
+
+        uint256 result;
+
+        assembly {
```
frangio, line 13: We want to use as little assembly as possible. I don't think we need any assembly at all. I think we can use `address(5).staticcall(abi.encode(32, 32, 32, b, e, m))`.

mw2000: @frangio noted, will be adding these changes

mw2000: Added the needed changes @frangio

mw2000:
> ### ⚠️ No Changeset found
> Latest commit: [0d78d29](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/0d78d29e3e6e21101b408118da2bee15c85e8dbd)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> 
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/mw2000/openzeppelin-contracts/new/modular-exponentiation-precompile-wrapper-#1985?filename=.changeset/flat-cooks-turn.md&value=---%0A%22%40fake-scope%2Ffake-pkg%22%3A%20patch%0A---%0A%0AAdd%20Wrapper%20Library%20for%20ModularExponentiation%20along%20with%20tests.%0A)

Added the changeset in next commit

mw2000:
@frangio had been a while since I worked on this repo, but here's some updated code if it's worth merging in the next update

test/utils/math/Math.t.sol
```diff
@@ -170,6 +170,29 @@ contract MathTest is Test {
...
+        // Next, let's check the correctness of the result
+        // The result of b^e mod m should be less than m
+        assertTrue(result < m);
+
+        // The result should also satisfy the equation: result = b^e mod m
+        // Due to the large numbers involved in this calculation, we can't directly calculate b^e
+        // However, we can repeatedly multiply and reduce modulo m
+
+        uint256 expected = 1;
+        for (uint i = 0; i < e; i++) {
```
Amxx, line 189: For decently big values of e, this loop won't realistically run.

contracts/utils/math/Math.sol
```diff
@@ -402,4 +402,23 @@ library Math {
...
+     *
+     * - modulus can't be zero
+     * - result should be obtained successfully
+     */
+    function modExp(uint256 b, uint256 e, uint256 m) internal view returns (uint256) {
+        require(m != 0, "ModularExponentiation: Can't calculate for modulus equal to zero");
+        bool success;
+        bytes memory result;
+
+        (success, result) = (address(5).staticcall(abi.encode(32, 32, 32, b, e, m)));
```
Amxx, line 420:
```suggestion
        (bool success, bytes memory result) = (address(5).staticcall(abi.encode(32, 32, 32, b, e, m)));
```

test/utils/math/Math.t.sol
Amxx: This causes #4337

test/utils/math/Math.t.sol
```diff
@@ -170,6 +170,32 @@ contract MathTest is Test {
         } catch {}
     }
 
+    // MOD EXP
+    function testModExp(uint256 b, uint256 e, uint256 m) public {
+        // First, let's assume m is not 0, as that will cause a division by zero error
+        vm.assume(m != 0);
+        vm.assume(e < 10);
+        vm.assume(b != 0);
+        vm.assume(e != 0);
```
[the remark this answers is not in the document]
mw2000: Resolved this by making a hard coded case when m=1

contracts/utils/math/Math.sol
```diff
@@ -406,6 +406,22 @@ library Math {
...
+     * @dev Returns the modular exponentiation of the specified base,
+     * exponent and modulus (b^e % m)
+     *
+     * Requirements:
+     *
+     * - modulus can't be zero
+     * - result should be obtained successfully
+     */
+    function modExp(uint256 b, uint256 e, uint256 m) internal view returns (uint256) {
+        require(m != 0, "ModularExponentiation: Can't calculate for modulus equal to zero");
```
ernestognw, line 419: We need to replace these with a custom error following [our guidelines](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/630844ef50c7581f8a91d15f535e9c67823951a5/GUIDELINES.md).

contracts/utils/math/Math.sol
```diff
@@ -406,6 +406,22 @@ library Math {
...
+        (bool success, bytes memory result) = (address(5).staticcall(abi.encode(32, 32, 32, b, e, m)));
+        require(success, "ModularExponentiation: Failed at calculating the result");
```
ernestognw, line 421: Same about custom errors

mw2000: Added custom error

contracts/utils/math/Math.sol
mw2000: Added custom error

mw2000:
@ernestognw not sure how to get the codecov/patch to pass as not sure how to test for the MathModExpCannotBeCalculated error.

.changeset/shiny-poets-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Added modular exponentiation function to Math.sol
```
Amxx, line 5:
```suggestion
`Math`: Add `modExp` function that exposes the `EIP-198` precompile.
```

Amxx:
I improved the fuzzing test to cover all the space (particularly for the exponent part).

contracts/utils/math/Math.sol
```diff
@@ -275,6 +283,24 @@ library Math {
...
+     * @dev Returns the modular exponentiation of the specified base, exponent and modulus (b ** e % m)
+     *
+     * Requirements:
+     * - modulus can't be zero
+     * - result should be obtained successfully
+     */
+    function modExp(uint256 b, uint256 e, uint256 m) internal view returns (uint256) {
+        if (m == 0) {
+            revert MathModulusEqualsZero();
+        }
```
ernestognw, line 296: The default behavior of the precompile is to return 0 when `m == 0`. Wouldn't make more sense to respect the precompile behavior and treat it as a noop? @Amxx

contracts/utils/math/Math.sol
```diff
@@ -275,6 +283,24 @@ library Math {
...
+        (bool success, bytes memory result) = (address(5).staticcall(abi.encode(32, 32, 32, b, e, m)));
+        if (!success) {
+            revert MathModExpCannotBeCalculated();
+        }
```
ernestognw, line 300: Couldn't find a way to make the precompile fail, not even with `m == 0`. I looked at:

- [EIP 198](https://eips.ethereum.org/EIPS/eip-198)
- [EVM Codes](https://www.evm.codes/precompiled?fork=shanghai)

I'll try sending less gas than required to test this error.

test/utils/math/Math.t.sol
```diff
@@ -206,23 +206,43 @@ contract MathTest is Test {
...
+            assertTrue(result < m);
+            assertEq(result, _nativeModExp(b, e, m));
+        } catch {
+            assertEq(m, 0);
+        }
+    }
+
+    function _nativeModExp(uint256 b, uint256 e, uint256 m) private pure returns (uint256) {
+        uint256 r = 1 % m;
+        while (e > 0) {
```
ernestognw, line 221:
```suggestion
        uint256 r = 1 % m;
        if(r == 0) return 0;
        while (e > 0) {
```

test/utils/math/Math.t.sol
```diff
@@ -206,23 +206,43 @@ contract MathTest is Test {
         } catch {}
     }
 
+    // MOD EXP
+    function testModExp(uint256 b, uint256 e, uint256 m) public {
+        try this.modexp(b, e, m) returns (uint256 result) {
```
ernestognw, line 211: Why does this have to be a external call (`this.modex`) and not just `Math.modExp`?
The revert can be expected conditionally if `m == 0` with:

```
if(m == 0) vm.expectRevert(...)
```

contracts/utils/math/Math.sol
ernestognw: Didn't work, I tried different gasLimit values in the following test;

```
 it.only('reverts with not enough gas', async function () {
  await expect(
    this.mock.$modExp(
      0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffn,
      0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffn,
      0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffn,
      {
        gasLimit: 25_000,
      },
    ),
  ).to.be.revertedWithCustomError(this.mock, 'MathModExpCannotBeCalculated');
});
```

It reverts as out of gas but doesn't trigger the `MathModExpCannotBeCalculated`.

ernestognw:
This looks good to me but I think we have to make an agreement in regards to the case where `m == 0`.

Considering that [other implementations](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/dnssec-oracle/algorithms/ModexpPrecompile.sol) don't care about it and that some sources like [EVM Codes](https://www.evm.codes/precompiled?fork=shanghai) explicitly states that if `m == 0` then it should return 0, then I think we should also treat the case as a noop instead.

contracts/utils/math/Math.sol
Amxx: yes, I guess that is because if the precompile has not enough gas, than the 1/64 that remains after that are not enough to check success and revert with a custom error

test/utils/math/Math.t.sol
Amxx: I'm not sure I understand that change.

The only way for `r` to be 0 is for m to be `1`.
In that case, then everything is 0, all operations done mod m return 0.
This means the while loop will try to do stuff, but in the end, we will have `r=0`.

I'm not sure what we are trying to optimise. Its not production code, its a reference for fuzzing.

test/utils/math/Math.t.sol
Amxx: Can you expect revert internal library operation? I was under the impression that you can only expectRevert actual calls/static calls ... which is why we have this wrapper.

Note we do the same in mulDiv

Amxx:
> This looks good to me but I think we have to make an agreement in regards to the case where `m == 0`.

m = 0 really makes no sens. Trying to do modulo 0 is like trying to divide by 0. There is not correct result for a division by 0 but an error.

contracts/utils/math/Math.sol
Amxx: Here, returning 0 if m == 0 would be like returning 0 when dividing something by 0.

Maybe at the precompile level its ok to do so, because they expect callers of the precompile will check they are not doing anything stupid, and because they don't want to handle reverts in precompile (just like ecrecover will return address(0) with invalid signatures).

But we are at the library level, and I think we should revert if someone tries to divide by 0 (or do modulo 0)

test/utils/math/Math.t.sol
ernestognw: Not trying to optimize, just thought it was better to avoid the full computation and return right ahead.

I'm fine with the changes you made fixing `r` to 1

Amxx (approved):

ernestognw:
> m = 0 really makes no sens. Trying to do modulo 0 is like trying to divide by 0. There is not correct result for a division by 0 but an error.

Yeah, it's just undefined. I just feel safer keeping the original behavior instead of arbitrarily rejecting 0, we can't remove that restriction once released but we can add it if needed.

I agree the operation doesn't make sense, but I feel completely clueless about why (or how) would someone rely on `m == 0` not reverting.

I checked the case for [ENS's DNSSEC RSA verification](https://github.com/ensdomains/ens-contracts/blob/02c9e6b3fcf643b13a35b8c4fcffd2b1f43bf4dc/contracts/dnssec-oracle/algorithms/RSASHA256Algorithm.sol#L40) and the effect is allowing `signatures == 0`, which would be only useful if a preimage for `sha256(data)` is known. That doesn't seem intentional but it's harmful.

contracts/utils/math/Math.sol
ernestognw: Yeah but why is it stupid to allow `% 0`? If there's a concrete reason why this is dangerous, then let's keep it.

The thing is that I haven't come up with any good reason for either other than that it's difficult to remove the check once released 😅 

I agree with restricting the precompile at the library level, but the main difference with ECDSA is that it's dangerous to use it without the wrapper because there's a way of replaying a signature.

Amxx: > Yeah but why is it stupid to allow `% 0`?

`x % 0` is the remainder of the **division of x by 0**. What more can I say ?

ernestognw: And? How is someone stupid for returning 0?

I am really sorry if this still doesn't make sense to you but I can't see a single reason why that is **dangerous** so that someone using it is stupid.

Amxx: Would you seriously consider that `mulDiv(1,1,0)` should return 0 as well?

ernestognw: But 0 div panics by default. modexp doesn't.

As I said, I don't have reasons for saying it's okay or not. Doesn't seem harmful and restricting is irreversible.

Amxx: But 0 div panics by default. modexp doesn't.

- div is an opcode
- modexp is a precompile

Note that there is a mod opcode. If you try to use it to do `1 % 0`, this opcode will revert.

Amxx: interrestingly,

```
uint256 i = 1%0
``` 
does revert
```
unchecked { uint256 i = 1%0; }
```
does not

Amxx: Continuing the discussion on slack.

Amxx:
I noticed that the implementation used `abi.encode`. This allocates a memory buffer (which move the free memory ptr at 0x40). Solidity does not free that memory.

This means that if the function is called multiple time, memory "leaks", which increases the gas cost.

I took the liberty of rewritting the function in assembly, similar to what we do for `_hasPair`.
I also took this opportunity to remove the custom error, and replace it with a `DIVIDE_BY_ZERO` panic code.

This last part is open to discussion, but I think very strongly that modulo 0 should not be allowed, and that triggering a DIVIDE_BY_ZERO panic code is the right thing to do.

frangio:
> some sources like [EVM Codes](https://www.evm.codes/precompiled?fork=shanghai) explicitly states that if `m == 0` then it should return 0

For what it's worth, EVM Codes says the same thing about the DIV and MOD opcodes, but the high level operators in Solidity do panic. I think the MODEXP precompile returns 0 in this case to have a behavior close to those opcodes, and that for a higher level wrapper of the precompile we are free to make a different decision. My 2 cents are that we should revert by default, following the footsteps of `%` in SafeMath/Solidity. But I'll also note that we have the `try*` pattern that we could reuse here, i.e. offer `tryModExp` that returns false without reverting, and `modExp` that does revert on error.

Amxx:
Note: functions in Math.sol are missing a `tryXxx` variant:
- `ceilDiv(uint256,uint256)`: panic 0x12 if denominator is 0
- `mulDiv(uint256,uint256,uint256)`: panic 0x12 if denominator is 0 + Custom error if result overflow
- `invMod`: this one should never revert, it returns 0 is error (its a try that "hides" it)

test/utils/Panic.test.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+describe('Panic', function () {
+  beforeEach(async function () {
+    Object.assign(this, await loadFixture(fixture));
+  });
+
+  for (const [ name, code ] of Object.entries(PANIC_CODES)) {
+    describe(name, function () {
+      it(`exposes panic code as constant`, async function () {
```
Amxx, line 17:
```suggestion
      it('exposes panic code as constant', async function () {
```

contracts/utils/Panic.sol
```diff
@@ -0,0 +1,28 @@
...
+library Panic {
+    uint256 internal constant ASSERTION_ERROR = 0x1;
+    uint256 internal constant ARITHMETIC_UNDER_OR_OVERFLOW = 0x11;
+    uint256 internal constant DIVISION_BY_ZERO = 0x12;
+    uint256 internal constant ENUM_CONVERSION_OUT_OF_BOUNDS = 0x21;
+    uint256 internal constant INCORRECTLY_ENCODED_STORAGE_BYTE_ARRAY = 0x22;
+    uint256 internal constant POP_ON_EMPTY_ARRAY = 0x31;
+    uint256 internal constant ARRAY_ACCESS_OUT_OF_BOUNDS = 0x32;
+    uint256 internal constant TOO_MUCH_MEMORY_ALLOCATED = 0x41;
+    uint256 internal constant ZERO_INITIALIZED_VARIABLE = 0x51;
```
ernestognw, line 18: Let's document these, as their meaning might not be obvious to users using the library without knowing these are already defined by Solidity.

Also, let's consider the original names given in the [Solidity repo](https://github.com/ethereum/solidity/blob/4577aebfd24a8dec47f7ce085d4e8887f5b8b3af/libsolutil/ErrorCodes.h#L36).

```suggestion
    /// @dev Called `assert` with an argument that evaluates to `false`.
    uint256 internal constant ASSERT = 0x1;
    /// @dev An arithmetic operation resulted in underflow or overflow.
    uint256 internal constant UNDER_OVERFLOW = 0x11;
    /// @dev Division or modulo operation by 0 (i.e. `x / 0` or `x % 0`).
    uint256 internal constant DIVISION_BY_ZERO = 0x12;
    /// @dev A number greater than 255 or negative was wrapped in an `enum`.
    uint256 internal constant ENUM_CONVERSION_ERROR = 0x21;
    /// ???
    uint256 internal constant STORAGE_ENCODING_ERROR = 0x22;
    /// @dev Attempted to `pop` an element out of an empty array.
    uint256 internal constant EMPTY_ARRAY_POP = 0x31;
    /// @dev Attempted to access an out-of-bound index at an empty array
    ///  (i.e. `x[i]` where `i >= x.length` or `i < 0`).
    uint256 internal constant ARRAY_OUT_OF_BOUNDS = 0x32;
    /// @dev If you allocate too much memory or create an array that is too large.
    uint256 internal constant RESOURCE_ERROR = 0x41;
    /// @dev Called an uninitialized internal function pointer.
    uint256 internal constant INVALID_INTERNAL_FUNCTION = 0x51;
```

Note I don't know how to trigger the `0x22` code. Also, I couldn't find what "too large" means in code `0x41`

contracts/utils/Panic.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Helper library for emitting standardized panic codes.
+ */
+// slither-disable-next-line unused-state
+library Panic {
```
ernestognw, line 9: Given Solidity lists the `0x0` as "Generic", would it make sense to include an?:

```solidity
uint256 internal constant GENERIC = 0x0;
```

contracts/utils/math/Math.sol
```diff
@@ -149,7 +146,7 @@ library Math {
 
             // Make sure the result is less than 2^256. Also prevents denominator == 0.
             if (denominator <= prod1) {
-                revert MathOverflowedMulDiv();
+                Panic.panic(denominator == 0 ? Panic.DIVISION_BY_ZERO : Panic.ARITHMETIC_UNDER_OR_OVERFLOW);
```
ernestognw, line 149: We shall reflect this in the changesets. It's not a breaking change because we're not guaranteeing backwards compatibility for error interfaces but it'd be cool if we add a changelog entry

contracts/utils/math/Math.sol
```diff
@@ -226,6 +223,9 @@ library Math {
      * If n is not a prime, then Z/nZ is not a field, and some elements might not be inversible.
      *
      * If the input value is not inversible, 0 is returned.
+     *
+     * Note: If you know for sure that n is (big) a prime, it may be cheaper to use Ferma's little theorem and get the
```
ernestognw, line 227:
```suggestion
     * TIP: If you know for sure that n is (big) a prime, it may be cheaper to use Ferma's little theorem and get the
```

test/utils/math/Math.t.sol
```diff
@@ -201,14 +201,31 @@ contract MathTest is Test {
...
+        Math.mulDiv(x, y, d);
     }
 
-    // External call
-    function muldiv(uint256 x, uint256 y, uint256 d) external pure returns (uint256) {
-        return Math.mulDiv(x, y, d);
+    // MOD EXP
+    function testModExp(uint256 b, uint256 e, uint256 m) public {
+        if (m == 0) {
+            vm.expectRevert(); // panic 0x12
```
ernestognw, line 211: I think we can do:

```suggestion
            vm.expectRevert(abi.encode(0x12));
```

test/utils/math/Math.t.sol
```diff
@@ -201,14 +201,31 @@ contract MathTest is Test {
         vm.assume(xyHi >= d);
 
         // we are outside the scope of {testMulDiv}, we expect muldiv to revert
-        try this.muldiv(x, y, d) returns (uint256) {
-            fail();
-        } catch {}
+        vm.expectRevert();
```
ernestognw, line 204: Haven't tested but I think we can do:

```suggestion
        uint8 code = 0x11;
        if(d == 0) {
          code++;
        }
        vm.expectRevert(abi.encode(code));
```

ernestognw:
> For what it's worth, EVM Codes says the same thing about the DIV and MOD opcodes, but the high level operators in Solidity do panic. 

Yeah, that's right! I'm fine defaulting to Solidity's behavior if it's a "generic" error. Note that we agreed to add a `Panic` library, although this still needs some work.

> My 2 cents are that we should revert by default, following the footsteps of % in SafeMath/Solidity. But I'll also note that we have the try* pattern that we could reuse here, i.e. offer tryModExp that returns false without reverting, and modExp that does revert on error

Right. I'll try to implement it in that way

contracts/utils/math/Math.sol
```diff
@@ -275,6 +275,33 @@ library Math {
...
+        /// @solidity memory-safe-assembly
+        assembly {
+            let ptr := mload(0x40)
+            mstore(ptr, 0x20)
+            mstore(add(ptr, 0x20), 0x20)
+            mstore(add(ptr, 0x40), 0x20)
+            mstore(add(ptr, 0x60), b)
+            mstore(add(ptr, 0x80), e)
+            mstore(add(ptr, 0xa0), m)
+            if iszero(staticcall(gas(), 0x05, ptr, 0xc0, ptr, 0x20)) {
```
ernestognw, line 298: I'm pretty sure it's `0xb0`
```suggestion
            if iszero(staticcall(gas(), 0x05, ptr, 0xb0, ptr, 0x20)) {
```

contracts/utils/math/Math.sol
```diff
@@ -275,6 +275,33 @@ library Math {
         }
     }
 
+    /**
+     * @dev Returns the modular exponentiation of the specified base, exponent and modulus (b ** e % m)
+     *
+     * Requirements:
+     * - modulus can't be zero
+     * - result should be obtained successfully
+     */
```
ernestognw, line 284: We should mention this function depends on the availability of the `0x05` precompile on the network where it's used as part of the assumptions of using this function.

contracts/utils/math/Math.sol
```diff
@@ -275,6 +310,30 @@ library Math {
...
+     * sure the chain you're using it on supports the precompiled contract for modular exponentiation
+     * at address 0x05 as specified in https://eips.ethereum.org/EIPS/eip-198[EIP-198]. Otherwise,
+     * the underlying function will succeed given the lack of a revert, but the result may be incorrectly
+     * interpreted as 0.
+     */
+    function modExp(uint256 b, uint256 e, uint256 m) internal view returns (uint256) {
+        if (m == 0) {
+            Panic.panic(Panic.DIVISION_BY_ZERO);
+        }
+        (bool success, uint256 result) = tryModExp(b, e, m);
```
ernestognw, line 330: A potential threat here is that in chains where this precompile is not available, then the result length will be `0`.

I guess that's the most reliable way to protect against but I don't think it's worth the check.

I added a not specifying that developers need to make sure if the precompile is available.

ernestognw:
I couldn't find a way in which calling the precompile returns false. I wonder if that's even possible. In such case the success flag is not even needed.

test/utils/math/Math.t.sol
ernestognw: Hmm, seems like Foundry's `expectRevert` doesn't catch panics.

```shell
[FAIL. Reason: Error != expected error:  != ; counterexample: calldata=0x8a59e100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 args=[0, 0, 0]] testModExp(uint256,uint256,uint256) (runs: 17, μ: 16029, ~: 3918)
Traces:
  [3721] MathTest::testModExp(0, 0, 0)
    ├─ [0] VM::expectRevert()
    │   └─ ← ()
    └─ ← panic: division or modulo by zero (0x12)
```

test/utils/math/Math.t.sol
ernestognw: This shouldn't work according: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3298#discussion_r1475476542

Amxx:
> In such case the success flag is not even needed.

Success MUST return `false` if the modulo is 0. Just like `tryDiv` returns `false` if the denominator is `0`. There shouldn't even be a discussion about that. Dividing by 0 is an error, and taking the remainder of this division is also an error.

test/utils/math/Math.t.sol
Amxx: It doesn but you must encode them properly.

You can do 
`abi.encodeWithSignature("Panic(uint256)", 0x12)` or use the stdErrors from foundry.

contracts/utils/Panic.sol
```diff
@@ -0,0 +1,28 @@
...
+    uint256 internal constant ARITHMETIC_UNDER_OR_OVERFLOW = 0x11;
+    uint256 internal constant DIVISION_BY_ZERO = 0x12;
+    uint256 internal constant ENUM_CONVERSION_OUT_OF_BOUNDS = 0x21;
+    uint256 internal constant INCORRECTLY_ENCODED_STORAGE_BYTE_ARRAY = 0x22;
+    uint256 internal constant POP_ON_EMPTY_ARRAY = 0x31;
+    uint256 internal constant ARRAY_ACCESS_OUT_OF_BOUNDS = 0x32;
+    uint256 internal constant TOO_MUCH_MEMORY_ALLOCATED = 0x41;
+    uint256 internal constant ZERO_INITIALIZED_VARIABLE = 0x51;
+
+    function panic(uint256 code) internal pure {
```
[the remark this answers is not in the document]
Amxx: I'm not too worried about users emitting random panic code ... but I'm not fundamentally opposed to making `panic(uint256)` private and having functions like
```
function TooMuchMemoryAllocated() internal pure { _panic(0x41); }
```

We would have to make them CamelCase though.

contracts/utils/Panic.sol
Amxx: I used the name for hardhat helpers, which is handy for testing. I'm open to using Solidity's names, but then testing will be a bit more confusing.

contracts/utils/Panic.sol
Amxx: If we follow solidity list (with naming, then yes)

contracts/utils/Panic.sol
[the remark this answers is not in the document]
Amxx: The thing, if new panic code are ever added, the current version would allow users to use them without us having to do a release. 
I'm also thinking, what if some L2 whant to use specific panic codes for some crosschain stuff. We won't want to add them to the library, but users will want to emit them.

contracts/utils/math/Math.sol
```diff
@@ -275,6 +276,66 @@ library Math {
...
+                revert Address.FailedInnerCall();
+            }
+        }
+        return result;
+    }
+
+    /**
+     * @dev Returns the modular exponentiation of the specified base, exponent and modulus (b ** e % m).
+     * It includes a success flag indicating if the operation succeeded. Operation will be marked has failed if trying
+     * to operate modulo 0 or if the underlying precompile reverted.
```
ernestognw, line 307:
```suggestion
     * to operate modulo 0 or if the underlying precompile reverted or if `m == 0`.
```

Amxx: > if trying to operate modulo 0
This is already documenting m == 0

contracts/utils/Panic.sol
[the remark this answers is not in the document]
Amxx: The linter rules are not ok with that, and I think we should follow them:
- constant must be SNAKE_CASE
- internal function in libraries are not _prefixed.

> The developers will have defined functions out of the box without the need to check the internal constants.

I'm not sure that is a good point. When testing panic code with hardhat you do 
```
expect(...).to.be.revertedWithPanic(PANIC_CODE.ARITHMETIC_UNDER_OR_OVERFLOW)
```
you don't to
```
expect(...).to.be.revertedWithPanicArithmeticUnderOrOverflow();
```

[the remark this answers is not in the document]
Amxx: I think the cureent version is simple and effective. It's targetting advanced users anyway.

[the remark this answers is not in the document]
ernestognw: > I'm not sure that is a good point. When testing panic code with hardhat you do
> 
> ```
> expect(...).to.be.revertedWithPanic(PANIC_CODE.ARITHMETIC_UNDER_OR_OVERFLOW)
> ```

Fair

> I think the cureent version is simple and effective. It's targetting advanced users anyway.

We should document the function then. Currently it allows panicking with anything it I think it needs some usage instructions in both the contracts NatSpec and the `panic` funciton itself

test/utils/math/Math.t.sol
ernestognw: Look at you! Such an experience Foundry user 🥂

Amxx (approved):
I think this is good.

@ernestognw wdyt ?

Note that the coverage issue comes from the 
```
} else {
    revert Address.FailedInnerCall();
}
```
which we don't know how to trigger.

I think its ok for use admin override for that.

ernestognw:
> I think this is good.
> 
> @ernestognw wdyt ?
> 
> Note that the coverage issue comes from the
> 
> ```
> } else {
>     revert Address.FailedInnerCall();
> }
> ```
> 
> which we don't know how to trigger.
> 
> I think its ok for use admin override for that.

Yeah it's good, just added an usage example and natspec to the panic function.

Also agree with the coverage drop for this, but I wouldn't remove it in case a weird chain has a contract in that address that does revert.

ernestognw (approved):
LGTM

Amxx (approved):

PR #3305, opened 2022-03-31
make ERC2981:royaltyInfo public

### LINKED ISSUE #3304: royaltyInfo() should/could be public
I want to inherit ERC2981.sol, then I  want to override the royaltyInfo() function in order to add a 'require' statement before calling the the inherited ERC2981 royaltyInfo() function (using the super. syntax.)
This creates the following compiler error

TypeError: Member "royaltyInfo" not found or not visible after argument-dependent lookup in type(contract super ...

and I believe this is caused because in ERC2981.sol, royaltyInfo() is declared as external rather than public.

ethereum testing environments

openzeppeling 4.5
Remix IDE
solidity 0.8.4

I would like to add a bit of functionality to the openzeppelin implementation of ERC2981. Specifically, I want to check that a token exists before asking for its royalty information, but the way the code is currently written I don't this can be done.
I think this can be solved by making the function in ERC2981.sol public instead of virtual.
Perhaps there was deliberate reasoning for making this function external, but if not, then it would be useful to have the ability to override royaltyInfo()

// Trying to compile the code below triggers the error.

pragma solidity ^0.8.4;

import "@openzeppelin/contracts@4.5.0/token/ERC721/ERC721.sol";
import "@openzeppelin/contracts@4.5.0/access/Ownable.sol";
import "@openzeppelin/contracts@4.5.0/utils/Counters.sol";
import "@openzeppelin/contracts@4.5.0/token/common/ERC2981.sol";

contract MyToken is ERC721, ERC2981, Ownable {

    Counters.Counter private _tokenIdCounter;

    constructor() ERC721("MyToken", "MTK") {}

    function _baseURI() internal pure override returns (string memory) {
        return "http://some.url/";
    }

    function safeMint(address to) public onlyOwner {
        uint256 tokenId = _tokenIdCounter.current();
        _tokenIdCounter.increment();
        _safeMint(to, tokenId);
    }

    function royaltyInfo(uint256 tokenId, uint256 salePrice)
        external
        view
        override
        returns (address, uint256)
    {
        // require(_exists(tokenId), "Requested royalty for non-existing token.");    //  <-- This line is why I want to override the function.
        return super.royaltyInfo(tokenId, salePrice);
    }

    // The following functions are overrides required by Solidity.

    function supportsInterface(bytes4 interfaceId)
        public
        view
        override(ERC721, ERC2981)
        returns (bool)
    {
        return super.supportsInterface(interfaceId);
    }
}

Amxx:
Fixes #3304

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## Unreleased
+ * `ERC2981`: make `royaltiInfo` public to allow super call in overrides.
+
+## Unreleased
 
  * `AccessControl`: add a virtual `_checkRole(bytes32)` function that can be overridden to alter the `onlyRole` modifier behavior. ([#3137](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3137))
  * `EnumerableMap`: add new `AddressToUintMap` map type. ([#3150](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3150))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
- * `ERC2981`: make `royaltiInfo` public to allow super call in overrides.
+ * `ERC2981`: make `royaltiInfo` public to allow super call in overrides. ([#3305](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3305))
 
 ## Unreleased
 
--- a/contracts/token/common/ERC2981.sol
+++ b/contracts/token/common/ERC2981.sol
@@ -40,13 +40,7 @@ abstract contract ERC2981 is IERC2981, ERC165 {
     /**
      * @inheritdoc IERC2981
      */
-    function royaltyInfo(uint256 _tokenId, uint256 _salePrice)
-        public
-        view
-        virtual
-        override
-        returns (address, uint256)
-    {
+    function royaltyInfo(uint256 _tokenId, uint256 _salePrice) public view virtual override returns (address, uint256) {
         RoyaltyInfo memory royalty = _tokenRoyaltyInfo[_tokenId];
 
         if (royalty.receiver == address(0)) {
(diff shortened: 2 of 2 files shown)


JulissaDantes (approved):
Looks good!

ATI00:
@Amxx and @JulissaDantes  - one thought around this (I was the person who submitted this issue...)
If this change gets implemented, the royaltyInfo() function will always be compiled as 'public' (even contracts that inherit ERC2981.sol and override royaltyInfo() will need to declare it as public, rather than 'external.'
This is really useful in order to allow super., however, I don't know if this means the EIP will no longer be strictly adhered to. Unfortunately the EIP (for 2981) has explicit code in it, which uses 'external' rather than 'public'.
Does the fact that the EIP itself uses the word 'external' rather than 'public' matter? I know that the interface Id is calculated from the functions names and arguments, so the _visibility_ of the function won't affect the actual bytes4 of the interface identifier, but how important is it to adhere to the exact coding/wording of the EIP?

Amxx:
ERC define interfaces which have the function marked external. Making it public doesn't change the interfaceId or the function selector.

You can see that other ERC (20, 721, ...) also define interfaces with external function that end up being implement in a public way.

ATI00:
@Amxx - I was hoping that was the case; makes sense.
Thanks again!

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+ * `ERC2981`: make `royaltiInfo` public to allow super call in overrides. ([#3305](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3305))
```
brooklynbagel, line 4: Nothing major, but `royaltiInfo` should be corrected to `royaltyInfo`

JulissaDantes: Thanks for noticing! Well update it.

PR #3308, opened 2022-03-31
Update ERC20Votes.sol

jayceftg:
The number returned would not exceed 2^32, but letting it be an uint256 is fine.

Fixes #???? 

Not converting uint256 to uint32.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Joeysantoro:
I like this change

PR #3311, opened 2022-04-01
ERC721 extension for efficient batch minting

### LINKED ISSUE #2355: Support EIP-2309: ERC-721 Consecutive Transfer Extension
**🧐 Motivation**
Support EIP-2309: ERC-721 Consecutive Transfer Extension
Enables minting any number of tokens in a single transaction

**📝 Details**

EIP is an extension to ERC-721, for a standardized event emitted when creating/transferring one, or many non-fungible tokens using consecutive token identifiers.
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2309.md

EIP is final status.

Creating issue for discussion on how EIP could be supported

Amxx:
Fixes #2355 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/token/ERC721/ERC721.sol
```diff
@@ -452,4 +459,28 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         address to,
         uint256 tokenId
     ) internal virtual {}
+
+    /**
+     * TODO
```
Amxx, line 464: This is still missing

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -0,0 +1,101 @@
...
+ * @title ERC721 Cheap sequential minting
+ */
+abstract contract ERC721Consecutive is ERC721 {
+    using BitMaps for BitMaps.BitMap;
+    using Checkpoints for Checkpoints.Checkpoint160[];
+
+    Checkpoints.Checkpoint160[] private _sequentialOwnership;
+    BitMaps.BitMap private _sequentialBurn;
+
+    event ConsecutiveTransfer(
```
Amxx, line 21: Do we want to put that in an IERC2309 ? Its final

Amxx:
Waiting for #3589

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -0,0 +1,128 @@
...
+    { receiver: user3, amount: 0 },
+    { receiver: user1, amount: 7 },
+  ];
+
+  beforeEach(async function () {
+    this.token = await ERC721ConsecutiveMock.new(
+      name,
+      symbol,
+      batches.map(({ receiver }) => receiver),
+      batches.map(({ amount }) => amount),
```
frangio, line 24: This is a mock so it doesn't matter much, but for the record I believe it's possible to pass in `batches` if the constructor expects an array of structs, maybe an array of tuples too.

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -0,0 +1,128 @@
...
+    );
+  });
+
+  describe('minting during construction', function () {
+    it('events are emitted at construction', async function () {
+      let first = 0;
+
+      for (const batch of batches) {
+        if (batch.amount == 0) continue;
```
frangio, line 33: We should check that the event was not emitted.

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -0,0 +1,128 @@
...
+    it('ownership is set', async function () {
+      const owners = batches.flatMap(({ receiver, amount }) => Array(amount).fill(receiver));
+
+      for (const tokenId in owners) {
+        expect(await this.token.ownerOf(tokenId))
+          .to.be.equal(owners[tokenId]);
+      }
+    });
+
+    it('balance & voting poser are set', async function () {
```
frangio, line 55:
```suggestion
    it('balance & voting power are set', async function () {
```

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -0,0 +1,116 @@
...
+    function _burn(uint256 tokenId) internal virtual override {
+        super._burn(tokenId);
+        if (tokenId <= _totalConsecutiveSupply()) {
+            _sequentialBurn.set(tokenId);
+        }
+    }
+
+    function _totalConsecutiveSupply() private view returns (uint96) {
+        uint256 length = _sequentialOwnership._checkpoints.length;
+        return length == 0 ? 0 : _sequentialOwnership._checkpoints[length - 1]._key + 1;
```
frangio, line 114: Instead of accessing the struct member, can't we use `latest()`?

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -0,0 +1,128 @@
...
+      for (const account of accounts) {
+        const balance = batches
+          .filter(({ receiver }) => receiver === account)
+          .map(({ amount }) => amount)
+          .reduce((a, b) => a + b, 0);
+
+        expect(await this.token.balanceOf(account))
+          .to.be.bignumber.equal(web3.utils.toBN(balance));
+
+        expect(await this.token.getVotes(account))
```
frangio, line 65: I think it would be better and more explicit if we removed self-delegation from the mock and did it as part of the test, after testing that initially voting power is zero.

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -87,6 +87,48 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
...
+     * - When `from` and `to` are both non-zero, ``from``'s `tokenId` will be
+     * transferred to `to`.
+     * - When `from` is zero, `tokenId` will be minted for `to`.
+     * - When `to` is zero, ``from``'s `tokenId` will be burned.
+     * - `from` cannot be the zero address.
+     * - `to` cannot be the zero address.
+     *
+     * To learn more about hooks, head to xref:ROOT:extending-contracts.adoc#using-hooks[Using Hooks].
+     */
+    function _beforeConsecutiveTokenTransfer(
```
frangio, line 105: This really defeats the purpose of using consecutive transfers to begin with... I think this should just revert.

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -0,0 +1,128 @@
...
+
+  describe('minting after construction', function () {
+    it('consecutive minting is not possible after construction', async function () {
+      await expectRevert(
+        this.token.mintConsecutive(user1, 10),
+        'ERC721Consecutive: batch minting restricted to constructor',
+      );
+    });
+
+    it('simple minting is possible after construction', async function () {
```
frangio, line 102: Let's add a test that it's not possible to mint a token id that was batch-minted.

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -0,0 +1,128 @@
...
+  beforeEach(async function () {
+    this.token = await ERC721ConsecutiveMock.new(
+      name,
+      symbol,
+      batches.map(({ receiver }) => receiver),
+      batches.map(({ amount }) => amount),
+    );
+  });
+
+  describe('minting during construction', function () {
```
frangio, line 28: We should test that it's not possible to mint in the constructor. Requires a different mock I think?

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -0,0 +1,116 @@
...
+            _sequentialBurn.unset(tokenId);
+        }
+    }
+
+    /**
+     * @dev See {ERC721-_mint}. Override needed to avoid looking up in the sequential ownership structure for burnt
+     * token.
+     */
+    function _burn(uint256 tokenId) internal virtual override {
+        super._burn(tokenId);
```
frangio, line 106:
```suggestion
        // Invoke the core burn functionality to adjust balances, invoke hooks, etc.
        super._burn(tokenId);
```

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -0,0 +1,116 @@
...
+        if (tokenId <= _totalConsecutiveSupply()) {
+            _sequentialBurn.set(tokenId);
```
frangio, line 108: As usual we have a problem with the hooks here. `_afterTokenTransfer` will be invoked in a state where the owner is still set. An alternative would be to add this code as an override to `_afterTokenTransfer` but that doesn't fully fix things because some other overrides that execute prior may still get invoked in that state.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * Warning: Using {ERC721Consecutive} prevents using {_mint} during construction in favor of {_mintConsecutive}.
+     * After construction, {_mintConsecutive} is no longer available and {_mint} becomes available.
+     */
+    function _mint(address to, uint256 tokenId) internal virtual override {
+        require(Address.isContract(address(this)), "ERC721Consecutive: can't mint during construction");
+
+        super._mint(to, tokenId);
+        if (_sequentialBurn.get(tokenId)) {
+            _sequentialBurn.unset(tokenId);
```
frangio, line 97: Same concern as for `_burn`.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * Emits a {ConsecutiveTransfer} event.
+     */
+    function _mintConsecutive(address to, uint96 batchSize) internal virtual {
+        // minting a batch of size 0 is a no-op
+        if (batchSize == 0) return;
+
+        require(!Address.isContract(address(this)), "ERC721Consecutive: batch minting restricted to constructor");
+        require(to != address(0), "ERC721Consecutive: mint to the zero address");
+        require(batchSize < 5000, "ERC721Consecutive: batches too large for indexing");
```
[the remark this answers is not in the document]
Amxx: that should be ok yes

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * WARNING: Consecutive mint is only available during construction. ERC721 requires that any minting done after
+     * construction emits a Transfer event, which is not the case of mints performed using this function.
+     *
+     * WARNING: Consecutive mint is limited to batches of 5000 tokens. Further minting is possible from a contract's
+     * point of view but would cause indexing issues for off-chain services.
+     *
+     * Emits a {ConsecutiveTransfer} event.
+     */
+    function _mintConsecutive(address to, uint96 batchSize) internal virtual {
```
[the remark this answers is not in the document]
Amxx: sure

contracts/token/ERC721/extensions/ERC721Consecutive.sol
Amxx: Two things:
- Latest returns the latest value. Here we want the latest key
- We could make a `latestKey` function (its in one of the commits for this PR) and use it, but then we have another issue:

The +1 is part of the right branch, its `0` or `_sequentialOwnership._checkpoints[length - 1]._key + 1`.
If `latestKey` returns `length == 0 ? 0 : _sequentialOwnership._checkpoints[length - 1]._key`, we still need to know if 0 is the key of a checkpoint or if it comes from an empty array.

If the first batch is a batch of length 0, then there is a key 0 in the array for that token. Its very different from the array being empty

test/token/ERC721/extensions/ERC721Consecutive.test.js
Amxx: We should do both. I'll keep self delegation in the constructor, but only for a subset of the recipient. The other will delegate later.

Self delegation in the constructor is the only way I found to test correct delegation during mint (because the mint is only during construction)

test/token/ERC721/extensions/ERC721Consecutive.test.js
Amxx: I'd love to use `expectEvent.notEmitted.inConstruction`, but it doesn't try to match params. It just checks if an event with this name was emitted.

We would either have to improve `expectEvent.notEmitted.inConstruction` or do the decoding here. The `decodeLog` helper is not exported ...

frangio: Is there no built in way to get the decoded events from the constructor?

contracts/token/ERC721/extensions/ERC721Consecutive.sol
Amxx: A batch of 1 will create `tokenId = 0` ... and since that is the last token of the batch, that is what we put in the checkpoint ... so there will be `0` in the history.

If we want `latest()` to return `(key,value)` ... that is a bit harder for dev to work with because you need to allocate the variable before using one. Also we can't change `History.latest()` for backward compatibility ... and I'd rather not have an inconsistency between `TraceXXX.latest()` and `History.latest()`.

contracts/token/ERC721/extensions/ERC721Enumerable.sol
Amxx: To me we have to options:
- revert, and just prevent anyone from using the two extensions together
- (strongly) discourage the two being used ... but make sure that things work as much as we can if devs go against our recommandations.

frangio: Personally I'd go with revert because the revert would be triggered at constructor time. Otherwise it's going to be unwarranted "OpenZeppelin makes everything expensive" all over again.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
frangio: Proposal:

- Add `trace.latestCheckpoint()` returning `(key, value)`
- Add `trace.length()`
- Implement this using the above functions

The reason I insist is that if we have to reach for the struct's internal variables it's because the interface is not enough. So I think it's worth adding these functions for users that may need to do similar things.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
frangio: Yes I'd say so. We should have that interface.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
Amxx: I think we already have `trace.length()`

Amxx: about `trace.latestCheckpoint()`, should it return `(key, value)` or `(pos, key, value)` ?
What if the Trace is empty, should it revert ?

test/token/ERC721/extensions/ERC721Consecutive.test.js
Amxx: we have the txHash, so I guess we can get the tx receipt, and decode the events manually. It would bypass the helpers though

frangio (approved):
LGTM!

xinbenlv:
Hi @Amxx and @frangio sorry for being late.

I discover this PR only when researching for EIP-6047. And I have concern about this PR going into next OZ release because backward incompatibility.

EIP-2309 is incompatible with ERC721 and adopting it, including this PR if release to prod of OZ, will break all indexers who assumes `Transfer` is the only event name they are looking

Example of breakage: 
- The blockscout is one of indexer expecting `Transfer` event, specifically the assumption in this particular line
https://github.com/blockscout/blockscout/blob/708a6b00362689aab50db8f0cd996c2de66013ed/apps/indexer/lib/indexer/transform/token_transfers.ex#L23

```
    erc20_and_erc721_token_transfers =
      logs
      |> Enum.filter(&(&1.first_topic == unquote(TokenTransfer.constant())))
      |> Enum.reduce(initial_acc, &do_parse/2)
```

will be broken without the following update:

```
    erc20_and_erc721_token_transfers =
      logs
      |> Enum.filter(&(&1.first_topic == unquote(TokenTransfer.constant()) ||&1.first_topic == TokenTransfer.erc2309_consecutive_transfer_signature()))
```

Amxx:
Hello @xinbenlv 

If you check out the ERC712 specifications, it is clearly stated that

```
/// @dev This emits when ownership of any NFT changes by any mechanism. 
/// This event emits when NFTs are created (`from` == 0) and destroyed 
/// (`to` == 0). Exception: during contract creation, any number of NFTs
/// may be created and assigned without emitting Transfer. At the time of 
/// any transfer, the approved address for that NFT (if any) is reset to none. 
event Transfer(address indexed _from, address indexed _to, uint256 indexed _tokenId);
```

Link: https://eips.ethereum.org/EIPS/eip-721

As a consequence, this IS compatible with ERC712, and indexers that don't support this exception (which is part of the standard) are the ones that need fixing.

frangio:
> Is ERC721Consecutive.sol only used in contract creation?

Yes. Batch minting reverts when used outside of the constructor. Furthermore, this has already been release. Please see the code or [the documentation](https://docs.openzeppelin.com/contracts/4.x/api/token/erc721#ERC721Consecutive) before commenting.

PR #3317, opened 2022-04-04
Use a customizable _execute function in TimelockController

Joeysantoro:
Fixes #N/A

Set the `_call` method of the TimelockController to internal virtual so it can be overridden.

For context, I am looking to write a TimelockController which exclusively executes through a Gnosis Safe as a module https://github.com/gnosis/safe-contracts/blob/main/contracts/base/ModuleManager.sol#L61-L73

So the "call" would get overridden to:
```
    function _call(
        bytes32 id,
        uint256 index,
        address target,
        uint256 value,
        bytes calldata data
    ) internal override {
        safe.executeTransactionFromModule(target, value, data, Operation.CALL);

        emit CallExecuted(id, index, target, value, data);
```

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/governance/TimelockController.sol
+++ b/contracts/governance/TimelockController.sol
@@ -6,6 +6,7 @@ pragma solidity ^0.8.0;
 import "../access/AccessControl.sol";
 import "../token/ERC721/IERC721Receiver.sol";
 import "../token/ERC1155/IERC1155Receiver.sol";
+import "../utils/Address.sol";
 
 /**
  * @dev Contract module which acts as a timelocked controller. When set as the
@@ -288,13 +289,15 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver 
     function execute(
         address target,
         uint256 value,
-        bytes calldata data,
+        bytes calldata payload,
         bytes32 predecessor,
         bytes32 salt
     ) public payable virtual onlyRoleOrOpenRole(EXECUTOR_ROLE) {
-        bytes32 id = hashOperation(target, value, data, predecessor, salt);
+        bytes32 id = hashOperation(target, value, payload, predecessor, salt);
+
         _beforeCall(id, predecessor);
-        _call(id, 0, target, value, data);
+        _execute(target, value, payload);
+        emit CallExecuted(id, 0, target, value, payload);
         _afterCall(id);
     }
 
@@ -318,11 +321,28 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver 
         require(targets.length == payloads.length, "TimelockController: length mismatch");
 
         bytes32 id = hashOperationBatch(targets, values, payloads, predecessor, salt);
+
         _beforeCall(id, predecessor);
         for (uint256 i = 0; i < targets.length; ++i) {
-            _call(id, i, targets[i], values[i], payloads[i]);
+            address target = targets[i];
+            uint256 value = values[i];
+            bytes calldata payload = payloads[i];
+            _execute(target, value, payload);
+            emit CallExecuted(id, i, target, value, payload);
         }
         _afterCall(id);
+    }
+
+    /**
+     * @dev Execute an operation's call.
+     */
+    function _execute(
+        address target,
+        uint256 value,
+        bytes calldata data
+    ) internal virtual {
+        (bool success, bytes memory  returndata) = target.call{value: value}(data);
+        Address.verifyCallResult(success, returndata, "TimelockController: underlying transaction reverted");
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Amxx:
If these changes follow your idea, I think we can go ahead and add a changelog before we can merge

contracts/governance/TimelockController.sol
```diff
@@ -318,13 +321,30 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
...
 
+    /**
+     * @dev Execute an operation's call.
+     */
+    function _execute(
+        address target,
+        uint256 value,
+        bytes calldata data
+    ) internal virtual {
+        (bool success, bytes memory  returndata) = target.call{value: value}(data);
```
Joeysantoro, line 344: ws nit
```suggestion
        (bool success, bytes memory returndata) = target.call{value: value}(data);
```

frangio (dismissed):
This makes sense to me.

The tests are failing but I'm not sure why. @Amxx Can you check?

Joeysantoro:
> This makes sense to me.
> 
> The tests are failing but I'm not sure why. @Amxx Can you check?

Tests are failing because Addresses.verifyCallResult will bubble up errors by default which is a different behavior to always returning the TimelockController error.

Changing it back to `require(success, "TimelockController: underlying transaction reverted");` would fix this but the newer implementation is slightly better so maybe its best to just fix the test

Amxx:
I already fixed part of the tests. I'll have a second pass

Joeysantoro:
Should I edit this to use the old require logic instead so tests won't be broken?

contracts/governance/TimelockController.sol
```diff
@@ -318,13 +321,30 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
...
+    /**
+     * @dev Execute an operation's call.
+     */
+    function _execute(
+        address target,
+        uint256 value,
+        bytes calldata data
+    ) internal virtual {
+        (bool success, bytes memory returndata) = target.call{value: value}(data);
+        Address.verifyCallResult(success, returndata, "TimelockController: underlying transaction reverted");
```
Joeysantoro, line 345:
```suggestion
require(success, "TimelockController: underlying transaction reverted");
```

test/governance/TimelockController.test.js
```diff
@@ -886,6 +886,38 @@ contract('TimelockController', function (accounts) {
...
+          operation.value,
+          operation.data,
+          operation.predecessor,
+          operation.salt,
+          { from: executor },
+        ),
+        'CallReceiverMock: reverting',
+      );
+    });
+
```
Joeysantoro, line 920:
```suggestion
```

test/governance/TimelockController.test.js
```diff
@@ -905,7 +937,7 @@ contract('TimelockController', function (accounts) {
         { from: proposer },
       );
       await time.increase(MINDELAY);
-      await expectRevert(
+      await expectRevert.unspecified(
```
Joeysantoro, line 940:
```suggestion
      await expectRevert(
```

test/governance/TimelockController.test.js
```diff
@@ -914,7 +946,6 @@ contract('TimelockController', function (accounts) {
           operation.salt,
           { from: executor },
         ),
```
Joeysantoro, line 948:
```suggestion
        ),
        'TimelockController: underlying transaction reverted',
```

test/governance/TimelockController.test.js
```diff
@@ -937,7 +968,7 @@ contract('TimelockController', function (accounts) {
         { from: proposer },
       );
       await time.increase(MINDELAY);
-      await expectRevert(
+      await expectRevert.outOfGas(
```
Joeysantoro, line 971:
```suggestion
      await expectRevert(
```

test/governance/TimelockController.test.js
```diff
@@ -946,7 +977,6 @@ contract('TimelockController', function (accounts) {
           operation.salt,
           { from: executor, gas: '70000' },
         ),
```
Joeysantoro, line 979:
```suggestion
        ),
        'TimelockController: underlying transaction reverted',
```

test/governance/TimelockController.test.js
```diff
@@ -634,7 +634,7 @@ contract('TimelockController', function (accounts) {
             ],
             [
               this.callreceivermock.contract.methods.mockFunction().encodeABI(),
-              this.callreceivermock.contract.methods.mockFunctionThrows().encodeABI(),
+              this.callreceivermock.contract.methods.mockFunctionRevertsNoReason().encodeABI(),
```
Joeysantoro, line 637:
```suggestion
              this.callreceivermock.contract.methods.mockFunctionThrows().encodeABI(),
```

contracts/governance/TimelockController.sol
```diff
@@ -318,13 +321,30 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
...
 
+    /**
+     * @dev Execute an operation's call.
+     */
+    function _execute(
+        address target,
+        uint256 value,
+        bytes calldata data
+    ) internal virtual {
+        (bool success, bytes memory returndata) = target.call{value: value}(data);
```
Joeysantoro, line 344:
```suggestion
        (bool success, ) = target.call{value: value}(data);
```

frangio (approved):
Agree to merge.

frangio (approved):

PR #3324, opened 2022-04-06
Improving comments for the _burn() function in ERC721URIStorage.

mcIovin:
Currently the comments for this function are really documenting what the
inherited function does in the parent contract (in other words, they are
describing what happens when 'super' is called). This commit changes the
comments to be more in line with other comments of other inherited
functions, whereby reference is made to the parent contract, but the
comments describe the local version of the function. An example of this
can be seen in ERC721Royalty.sol, where the comments reference the
parent contract, but describe the functionality of what the override is
doing. This seems like a better approach, because if the parent contract
were to every be changed, the comments would not necessarily need
amending in the descendants.

Fixes #3324

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC721/extensions/ERC721URIStorage.sol
+++ b/contracts/token/ERC721/extensions/ERC721URIStorage.sol
@@ -48,14 +48,9 @@ abstract contract ERC721URIStorage is ERC721 {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


frangio (approved):
Thank you @mcIovin!

PR #3327, opened 2022-04-07
Add customizable fee receiver to ERC20FlashMint

### LINKED ISSUE #3316: ERC3156 Flash loans extension improvement
**🧐 Motivation**
The current flash loans implementation suggests to burn the fee of the flash loan with no easy way for implementers to override this default behavior. e.g. If the token is capped, this leads to shrink the total supply over time [considering the fee is higher than 0].

This behavior is not well documented and the Wizard doesn't consider this.

**📝 Details**
I would suggest to implement two strategies for the flash loan fee processing. Where implementers explicitly set the right strategy for their use case.
1. BURN: The loaned amount and the fee amount get burnt (current implementation).
2. HOLD: The loaned amount get burnt and the fee amount get transferred to the contract itself (address(this)).
3. CUSTOM: The fee amount to be transferred to different address (Breaks the trust?)

mazenkhalil:
Fixes #3316 
Allow implementers to set the address for which the flash loan fee amount is transferred to. Instead of burning the loan fee.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


contracts/token/ERC20/extensions/ERC20FlashMint.sol
```diff
@@ -43,6 +43,19 @@ abstract contract ERC20FlashMint is ERC20, IERC3156FlashLender {
...
     }
 
+    /**
+     * @dev Returns the reciever address of the flash fee. By default this
+     * implementation returns the address(0) which means the fee amount will be burnt.
+     * This function can be overloaded to change the fee reciever.
+     * @param token The token to be flash loaned.
+     * @return The address for which the flash fee will be sent to.
+     */
+    function _flashFeeReceiver(address token) internal view virtual returns (address) {
```
frangio, line 53: I realized we don't need the `address token` argument, because the flash loaned token is always the token itself (`this`).

contracts/token/ERC20/extensions/ERC20FlashMint.sol
```diff
@@ -43,6 +43,19 @@ abstract contract ERC20FlashMint is ERC20, IERC3156FlashLender {
         return 0;
     }
 
+    /**
+     * @dev Returns the reciever address of the flash fee. By default this
+     * implementation returns the address(0) which means the fee amount will be burnt.
+     * This function can be overloaded to change the fee reciever.
```
frangio, line 49:
```suggestion
     * @dev Returns the receiver address of the flash fee. By default this
     * implementation returns the address(0) which means the fee amount will be burnt.
     * This function can be overloaded to change the fee receiver.
```

frangio (commented):
This is looking pretty good, thank you. Some minor comments.

contracts/mocks/ERC20FlashMintMock.sol
```diff
@@ -13,4 +13,8 @@ contract ERC20FlashMintMock is ERC20FlashMint {
     ) ERC20(name, symbol) {
         _mint(initialAccount, initialBalance);
     }
+
+    function mockFlashFeeReceiver(address token) public view returns (address) {
```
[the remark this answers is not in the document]
mazenkhalil: Nope. Just to differentiate it as wrapper for an internal function.

test/token/ERC20/extensions/ERC20FlashMint.test.js
```diff
@@ -92,5 +92,61 @@ contract('ERC20FlashMint', function (accounts) {
       // _mint overflow reverts using a panic code. No reason string.
       await expectRevert.unspecified(this.token.flashLoan(receiver.address, this.token.address, MAX_UINT256, data));
     });
+
+    it('custom flash fee', async function () {
```
frangio, line 96: There is a lot of duplicated code between this test and the next. Can you try to simplify it? Consider using a `describe` block with a `beforeEach`, and/or defining a helper function to do some of the shared checks.

test/token/ERC20/extensions/ERC20FlashMint.test.js
```diff
@@ -93,60 +93,52 @@ contract('ERC20FlashMint', function (accounts) {
       await expectRevert.unspecified(this.token.flashLoan(receiver.address, this.token.address, MAX_UINT256, data));
     });
 
-    it('custom flash fee', async function () {
+    describe('custom flash fee & receiver', async function () {
       const receiver = await ERC3156FlashBorrowerMock.new(true, true);
```
frangio, line 97: Hm, this doesn't look good. The describe function can't be async, and any contract deployments need to go inside in `beforeEach` blocks.

test/token/ERC20/extensions/ERC20FlashMint.test.js
```diff
@@ -86,5 +92,53 @@ contract('ERC20FlashMint', function (accounts) {
...
+        await expectEvent.inTransaction(tx, this.receiver, 'TotalSupply', { token: this.token.address, value: initialSupply.add  (receiverInitialBalance).add(loanAmount) });
+
+        expect(await this.token.totalSupply()).to.be.bignumber.equal(initialSupply.add(receiverInitialBalance).sub(flashFee));
+        expect(await this.token.balanceOf(this.receiver.address)).to.be.bignumber.equal(receiverInitialBalance.sub(flashFee));
+        expect(await this.token.balanceOf(await this.token.flashFeeReceiver())).to.be.bignumber.equal('0');
+        expect(await this.token.allowance(this.receiver.address, this.token.address)).to.be.bignumber.equal('0');
+      });
+
+      it('custom flash fee receiver', async function () {
+        const flashFeeReceiverAddress = this.token.address;
```
Amxx, line 124: Everything looks great, but I think it could be confusing to some people reading the test the fee is sent to the token itself ... I think it would be better to send it an otherwise empty EOA.

contracts/mocks/ERC20FlashMintMock.sol
```diff
@@ -13,4 +16,29 @@ contract ERC20FlashMintMock is ERC20FlashMint {
...
+    function mint(address account, uint256 amount) public {
+        _mint(account, amount);
+    }
+
+    function setFlashFee(uint256 amount) public {
+        _flashFeeAmount = amount;
+    }
+
+    function flashFee(address token, uint256 amount) public view virtual override returns (uint256) {
+        super.flashFee(token, amount);
```
Amxx, line 29: This super call is not needed and can be confusing, even if it is just a mock.

Amxx (changes requested):
Great work overall. There are a few things regarding testing that can be improved.

We also need a Changelog entry before we merge that PR.

contracts/mocks/ERC20FlashMintMock.sol
mazenkhalil: The purpose of this super call is to cover the validation defined in the super implementation. I have opened another ticket to fix the flashFee function. #3331

frangio (approved):
Thank you @mazenkhalil. This looks good.

PR #3329, opened 2022-04-08
Optimize ERC1167 proxy creation code by 1 opcode

Philogy:
#### Description

The existing proxy creation bytecode was a total of 55 bytes:
* `3d602d80600a3d3981f3` 10 bytes, responsible for returning the remaining 45 byte [ERC1167](https://eips.ethereum.org/EIPS/eip-1167) proxy code, equivalent to the constructor in solidity
* `363d3d373d3d3d363d73????????????????????????????????????????5af43d82803e903pd91602b57fd5bf3` 45 bytes, standard ERC1167 proxy bytecode

Looking at the execution of the constructor code one notices that one value always remains, unused on the stack at the end of execution (stack after op depicted in the square brackets; left -> right, top -> bottom):
1. (0) RETURNDATASIZE [ 0, ]
2. (2) PUSH1 0x2d [ 45, 0, ]
3. (3) DUP1 [ 45, 45, 0, ]
4. (5) PUSH1 0x0a [ 10, 45, 45, 0, ]
5. (6) RETURNDATASIZE [ 0, 10, 45, 45, 0, ]
6. (7) CODECOPY [ 45, 0, ]
7. (8) DUP2 [ 0, 45, 0, ]
8. (9) RETURN [ 0, ]

By rearranging the opcodes slightly, one opcode can be removed, resulting in a smaller, 9-byte constructor `602d8060093d393df3`:
1.  (0) PUSH1 0x2d [ 45, ]
2.  (2) DUP1 [ 45, 45, ]
3.  (4) PUSH1 0x09 [ 9, 45, 45, ]
4.  (5) RETURNDATASIZE [0, 9, 45, 45, ]
5.  (6) CODECOPY [ 45, ]
6.  (7) RETURNDATASIZE [ 0, 45, ]
7.  (8) RETURN [ ]

The `proxy/Clones.sol` library's `clone`, `cloneDeterministic` and `predictDeterministicAddress` methods are then adjusted for the smaller code. The total creation bytecode is then 54 bytes instead of the current 55 bytes.

#### PR Checklist

- [X] Tests (functionality should remain unchanged, therefore tests should remain unchanged)
- [X] Documentation (functionality should remain unchanged, therefore docs should remain unchanged)
- [x] Changelog entry (unsure how to change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
  * `ERC2981`: make `royaltiInfo` public to allow super call in overrides. ([#3305](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3305))
+ * `Clones`: optimize clone creation ([#3329](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3329))
 
 ## Unreleased
 
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -25,9 +25,9 @@ library Clones {
     function clone(address implementation) internal returns (address instance) {
         assembly {
             let ptr := mload(0x40)
-            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
-            mstore(add(ptr, 0x14), shl(0x60, implementation))
-            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
+            mstore(ptr, 0x602d8060093d393df3363d3d373d3d3d363d7300000000000000000000000000)
+            mstore(add(ptr, 0x13), shl(0x60, implementation))
+            mstore(add(ptr, 0x27), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
             instance := create(0, ptr, 0x37)
         }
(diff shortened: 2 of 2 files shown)


frangio:
This is a tiny optimization so it has low priority for us, but I don't see a reason not to merge it.

Can you confirm if minimal proxies deployed with this new code are still detected as minimal proxies by Etherscan?

Philogy:
It's completely understandable that a 3 gas optimization is not your priority.

I've verified that etherscan recognizes proxies deployed with the adjusted bytecode as minimal proxies: https://ropsten.etherscan.io/address/0x6d91057e495b2fe6ba421ed7117bde6453c61516#code (creator address is a contract which has the adjusted internal `clone` method)

frangio:
Nice. Thanks.

> * Changelog entry (unsure how to change)

You can add an entry in the topmost "Unreleased" section in CHANGELOG.md. Just a short description, follow a similar format than the rest.

Philogy:
Changelog entry added and merged

frangio (approved):
Thanks! Looks good to me.

frangio:
@Philogy We ended up reverting this change, we realized it was a breaking change of the way deterministic addresses are computed and this has the risk of breaking upgradeable contracts. The risk is not worth taking for such a small improvement.

PR #3330, opened 2022-04-09
Update IERC721.sol

fps8k:
I'm sorry, I just found a requirement in the approve() function that was necessary to be added via comment, according to the ERC721 implementation. 

IERC721.sol comment on Line 111, " `to` must not be the current owner of the `tokenId` ".


fps8k:
> Thanks @fps8k. I'm now wondering why we have this requirement, which seems unnecessary. It doesn't seem to be part of the EIP spec. I'm wondering if we can remove it?

Alright.

frangio:
Following up in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3346.

PR #3337, opened 2022-04-12
Update ERC20.sol

jayceftg:
Checking it is fine. On the other hand, I searched through the repository, there is no way the owner would be 0x?

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -312,7 +312,6 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
         address spender,
         uint256 amount
     ) internal virtual {
-        require(owner != address(0), "ERC20: approve from the zero address");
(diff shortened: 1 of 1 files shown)


Amxx:
Please open an issue to discuss that before submitting a PR.

jayceftg:
@Amxx Why do you discuss other pr here but not this? And you waited so long. If the blatant double standards I will just not submit anything for you. Good luck with your own code. Bye.

Also, the test failed because you included a test to test the case where the sender is '0x', but the situation could never happen in your actual code as I looked through it after searching the code base. Again, enough with the double standards, good luck with your codes.

jayceftg:
@Amxx  Also, this is not some fake pr, this saves real gas, you want the function to cost extra gas doing nothing, fine with me!

PR #3343, opened 2022-04-15
Slight gas optimization of loops

v0xat:
I recently found a way to slightly optimize loops in Solidity and decided to check if OZ contracts use these approaches:

1) Using `++i` instead of `i++`
2) Caching array length

I understand that these are very insignificant changes, but still think they can be implemented because they won't affect overall code style.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry (will add later if this changes are approved)


frangio:
> They won't affect overall code style.

We don't think this is true. Having to read the array length separately manually affects the code style.

We think these optimizations should be done by the compiler and in the latest Solidity versions there has been a lot of work towards this. If you see an opportunity to optimize something that the compiler isn't doing, I encourage you to open an issue in the Solidity repository rather than doing the optimization manually in your code or in this library.

PR #3344, opened 2022-04-16
Restore previous behavior of initializer during construction

Amxx:
The following code is broken with 4.x:

```
contract A is Initializable {
    constructor() initializer { /* ... */ }
    /* ... */
}

contract B is A {
    constructor() initializer { /* ... */ }
    /* ... */
}
```

When B's constructor enters the `initializer` modifier, A's construction is already over. We are:
- not initializing
- version already set to 1.

This falls into the `else` part of `_setInitializedVersion`, which requires the next version to be greater than the previous one.

---

This would work, but requires the dev to affect version numbers that match the linearisation ...
```
contract B is A {
    constructor() reinitializer(2) { /* ... */ }
    /* ... */
}
```

---

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/test/proxy/utils/Initializable.test.js
+++ b/test/proxy/utils/Initializable.test.js
@@ -34,7 +34,7 @@ contract('Initializable', function (accounts) {
     });
 
     describe('nested under an initializer', function () {
-      it('initializer modifier can be nested', async function () {
+      it('initializer modifier reverts', async function () {
         await expectRevert(this.contract.initializerNested(), 'Initializable: contract is already initialized');
       });
 
--- a/test/proxy/utils/Initializable.test.js
+++ b/test/proxy/utils/Initializable.test.js
@@ -85,13 +85,9 @@ contract('Initializable', function (accounts) {
       expect(await this.contract.counter()).to.be.bignumber.equal('2');
     });
 
-    it('can nest reinitializers with same version', async function () {
+    it('cannot nest reinitializers', async function () {
       expect(await this.contract.counter()).to.be.bignumber.equal('0');
       await expectRevert(this.contract.nestedReinitialize(2, 2), 'Initializable: contract is already initialized');
-    });
-
-    it('cannot nest reinitializers with different versions', async function () {
-      expect(await this.contract.counter()).to.be.bignumber.equal('0');
       await expectRevert(this.contract.nestedReinitialize(2, 3), 'Initializable: contract is already initialized');
       await expectRevert(this.contract.nestedReinitialize(3, 2), 'Initializable: contract is already initialized');
     });
--- a/test/proxy/utils/Initializable.test.js
+++ b/test/proxy/utils/Initializable.test.js
@@ -34,8 +34,8 @@ contract('Initializable', function (accounts) {
     });
 
     describe('nested under an initializer', function () {
-      it('initializer modifier reverts', async function () {
-        await expectRevert(this.contract.initializerNested(), 'Initializable: contract is already initialized');
+      it('initializer modifier can be nested', async function () {
+        await this.contract.initializerNested();
       });
 
       it('onlyInitializing modifier succeeds', async function () {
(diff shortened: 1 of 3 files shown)


frangio:
I thought this PR was restoring previous behavior, but based on testing I don't think the example that this intends to fix ever worked. In fact none of the two examples below were working before #3232.

```solidity
contract A is Initializable {
    constructor() initializer {}
}

contract B is Initializable {
    constructor() initializer {}
}

contract DoesItInitialize1 is A, B {
    constructor() {}
}

contract DoesItInitialize2 is A, B {
    constructor() initializer {}
}
```

Amxx:
> I think this PR would be restoring behavior that we intentionally (!) broke when we introduced onlyInitializing.

I'm not sur it was intentional, and if it was, then we possibly didn't understand the full extend on consequences.

Do you want to delay that for 4.7 ?

frangio:
After further review this indeed seems unrelated to #3232, and not related either to #3006 (introduction of `onlyInitializing`). This use case never worked, so I would put this in queue for 4.7.

Amxx:
Indeed, this behavior has been broken since 4.0. It was possible to do so in 3.4.

frangio:
I see. For reference, this was the 3.x code:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/8e0296096449d9b1cd7c5631e917330635244c37/contracts/proxy/Initializable.sol#L36

The `isConstructor` check was removed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2531, for no particularly compelling reason.

Amxx:
> The `isConstructor` check was removed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2531, for no particularly compelling reason.


I think the reason was that we didn't see this modifier being used in constructor of the implementation. We discussed it covering a proxy under construction that would delegating call into the initialiser. That doesn't require special treatment, because you can expect all the initialiser code to be nested under the external initialiser.

Since then, we realised implementations should be locked at construction, and we encourage a usage that is different, that doesn't allow nesting, and this is problematic.

test/proxy/utils/Initializable.test.js
```diff
@@ -33,8 +34,8 @@ contract('Initializable', function (accounts) {
     });
 
     describe('nested under an initializer', function () {
-      it('initializer modifier reverts', async function () {
-        await expectRevert(this.contract.initializerNested(), 'Initializable: contract is already initialized');
+      it('initializer modifier can be nested', async function () {
+        await this.contract.initializerNested();
```
frangio, line 38: This feels like it enables reentrancy, like in https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories/GHSA-9c22-pwxw-p6hx. No?

frangio:
> Since then, we realised implementations should be locked at construction, and we encourage a usage that is different

This is a great point to restore the behavior for constructors.

test/proxy/utils/Initializable.test.js
Amxx: This was just fixed !

frangio:
So apparently this isn't working with `_disableInitializers` in multiple parents. This is now the replacement for `constructor() initializer` so we should probably support it.

Could we do:

```solidity
require((isTopLevelCall || !Address.isContract(address(this))) && version > currentVersion);
```

Amxx:
This will only partially work.

If we have 
```
contract A {
    constructor() initializer() {}
}
contract B {
    constructor() { _disableInitializers(); }
}
contract C is A, B {}
contract D is B, A {}
```

C is constructible, but D is not:
- `B`'s constructor will set the currentVersion to 255
- `A`'s constructor will try to run
  - we are in a `isTopLevelCall` but `version` < `currentVersion` (1 < 255)
  - there is no code at `address(this)` but `version` < `currentVersion` (1 < 255)

D would be constructible with 
```(isTopLevelCall && version > currentVersion) || (!Address.isContract(address(this)) && version == 1),```
And it look like C would also be ...

Amxx:
Would 

``` (isTopLevelCall && version > currentVersion) || (!Address.isContract(address(this))),```

Possibly be enough?

frangio:
That seems a bit reckless to me.

Alternatively we can do this:
`(isTopLevelCall && version > currentVersion) || (!Address.isContract(address(this)) && (version == 1 || version == 255))`
So as to support this only for the old and new patterns to disable initialization.

frangio (approved):

PR #3345, opened 2022-04-17
FlashLoanGuard modifier

Silur:
With the [recent BEAN flash loan attack](https://cryptobriefing.com/beanstalk-hacker-steals-76m-flash-loan-exploit/) it became clear that this year-old vulnerability is still present in complex contracts, and while the prevention is trivial, widespread usage of "flash loan guards" is lacking. This PR serves to allow complex contracts to protect their critical functionality from being called within the same block multiple times, in the same syntactic way as ReentrancyGuard.

### DIFF
--- a/contracts/mocks/FlashLoanAttack.sol
+++ b/contracts/mocks/FlashLoanAttack.sol
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "./FlashLoanMock.sol";
+
+contract FlashLoanAttack is FlashLoanGuard {
+    function increaseSafe(address addr) public {
+        FlashLoanMock(addr).increaseSafe();
+        FlashLoanMock(addr).increaseSafe();
+        FlashLoanMock(addr).increaseSafe();
+    }
+
+    function increaseUnsafe(address addr) public {
+        FlashLoanMock(addr).increaseUnsafe();
+        FlashLoanMock(addr).increaseUnsafe();
+        FlashLoanMock(addr).increaseUnsafe();
+    }
+}
--- a/contracts/mocks/FlashLoanMock.sol
+++ b/contracts/mocks/FlashLoanMock.sol
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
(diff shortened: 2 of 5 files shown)


Amxx:
Hello @Silur 

I'm worried that any function that would include this modifier is only executable once per block, not per transaction. This is IMO a terrible choice that will prevent legitimate user from calling that is anyone else does it in the same block.

Also, I'm not sure how this is supposed to protect against flash loans. Wouldn't the "nonReentrant" modifier be enough, less restrictive to the users, and already available?

Silur:
Once per block is the intented use, as flash loans require to be repaid at by the end of the current block latest. 
Non-Reentrancy only protects against your contract calling back into itself somewhere in the call stack, while this PR is more general, it prevents multiple calls from ANY caller (not just yourself).

Silur:
I do agree that using FlashLoanGuard is a heavy design **decision** one has to make (just like ERC20Permit btw), but restricting the caller to a `mapping` with a caller will only result in the hypothetical flashloan getting transferred to another caller.
I proposed this solution as an easy option for those who are willing to make this decision. The "once-per-block" UX issue is also less severe on high TPS layer2 and sometimes (see the BEAN attack above) it's worth to make the tradeoff.

Amxx:
I think you are mixing flash loans and sandwich attacks.

I don't clearly understand the use case for that modifier. For us it's not just about providing code, it's about providing documentation, guides, so devs know when and how to use it. 

We usually start with an issue where the use case needs are discussed, and we try to co-design the best solution possible. Only then do we move on to create a PR. 

I would juggest your open an issue to discuss that, and we put the PR on hold in the meantime.

Amxx:
I think you are mixing flash loans and sandwich attacks.

I don't clearly understand the use case for that modifier. For us it's not just about providing code, it's about providing documentation, guides, so devs know when and how to use it. 

We usually start with an issue where the use case needs are discussed, and we try to co-design the best solution possible. Only then do we move on to create a PR. 

I would juggest your open an issue to discuss that, and we put the PR on hold in the meantime.

Amxx:
> as flash loans require to be repaid by the end of the block they have been issued

Flash loans are repaid by the end of the transaction, not the block. There is no reason why a flash loan in one transaction should affect the other transactions (by other users?) in the same block

PR #3347, opened 2022-04-21
Optimize Ownable and Pausable modifiers' size impact

### LINKED ISSUE #3222: Reduce contract size and deployment gas for onlyOwner modifier

jnishiyama:
Fixes #3222

#### Affects:
- `contracts/access/Ownable.sol`
- `contracts/security/Pausable.sol`

#### Changes:
- In both contracts, the modifiers require statements are moved into an internal virtual function. This reduces the size of compiled contracts that use the modifiers.
- In both contracts, the modifier definition is moved above the definitions of functions to better adhere to the solidity style guide.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -47,7 +47,7 @@ abstract contract Ownable is Context {
     /**
      * @dev Throws if the sender is not the owner.
      */
-    function _checkOwner() internal virtual {
+    function _checkOwner() internal view virtual {
         require(owner() == _msgSender(), "Ownable: caller is not the owner");
     }
 
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -37,10 +37,17 @@ abstract contract Ownable is Context {
     }
 
     /**
+     * @dev Throws if the sender is not the owner.
+     */
+    function _onlyOwner() internal virtual {
+        require(owner() == _msgSender(), "Ownable: caller is not the owner");
+    }
+
+    /**
      * @dev Throws if called by any account other than the owner.
      */
     modifier onlyOwner() {
-        require(owner() == _msgSender(), "Ownable: caller is not the owner");
+        _onlyOwner();
         _;
     }
 
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -30,6 +30,14 @@ abstract contract Ownable is Context {
     }
 
     /**
+     * @dev Throws if called by any account other than the owner.
+     */
+    modifier onlyOwner() {
+        _onlyOwner();
+        _;
+    }
+
+    /**
      * @dev Returns the address of the current owner.
      */
     function owner() public view virtual returns (address) {
@@ -41,14 +49,6 @@ abstract contract Ownable is Context {
      */
     function _onlyOwner() internal virtual {
         require(owner() == _msgSender(), "Ownable: caller is not the owner");
-    }
-
-    /**
-     * @dev Throws if called by any account other than the owner.
-     */
-    modifier onlyOwner() {
-        _onlyOwner();
-        _;
     }
 
     /**
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -33,7 +33,7 @@ abstract contract Ownable is Context {
      * @dev Throws if called by any account other than the owner.
      */
     modifier onlyOwner() {
-        _onlyOwner();
+        _checkOwner();
         _;
     }
 
@@ -47,7 +47,7 @@ abstract contract Ownable is Context {
     /**
      * @dev Throws if the sender is not the owner.
      */
-    function _onlyOwner() internal virtual {
+    function _checkOwner() internal virtual {
         require(owner() == _msgSender(), "Ownable: caller is not the owner");
     }
 
--- a/contracts/security/Pausable.sol
+++ b/contracts/security/Pausable.sol
@@ -68,14 +68,14 @@ abstract contract Pausable is Context {
     /**
      * @dev Throws if the contract is paused.
      */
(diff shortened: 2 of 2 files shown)


Amxx:
For the `Ownable`, I created two contracts that uses the old and the new versions. Both compiled with 0.8.7 and optimizations on (200)

```
contract Test1 is Ownable {
    function f1() public onlyOwner() {}
    function f2() public onlyOwner() {}
    function f3() public onlyOwner() {}
    function f4() public onlyOwner() {}
    function f5() public onlyOwner() {}
    function f6() public onlyOwner() {}
}

contract Test2 is OwnableV2 {
    function f1() public onlyOwner() {}
    function f2() public onlyOwner() {}
    function f3() public onlyOwner() {}
    function f4() public onlyOwner() {}
    function f5() public onlyOwner() {}
    function f6() public onlyOwner() {}
}
```

Here are the deployment gas cost:

| Number of functions | 0 | 1 | 2 | 3 | 4 | 5 | 6 |
|-|-|-|-|-|-|-|-|
| Test1 | 244762 | 250123 (+5361) | 252869 (+2746) | 255588 (+2719) | 262302 (+6714) | 265034 (+2732) | 267816 (+2782) |
| Test2 | 226366 | 233032 (+6666) | 235757 (+2725) | 238476 (+2719) | 245148 (+6672) | 247867 (+2719) | 250579 (+2712) |

Oddly enough, each additional function is not really cheaper with the new version, but the base cost is cheaper :/

Amxx:
With solidity 0.8.14 and optimisations on (200)

```
contract Ownable is Context {
    address public owner = _msgSender();

    modifier onlyOwner() {
        require(owner == _msgSender(), "Ownable: caller is not the owner");
        _;
    }
}

contract Ownable2 is Context {
    address public owner = _msgSender();

    modifier onlyOwner() {
        _checkOwner();
        _;
    }

    function _checkOwner() internal virtual {
        require(owner == _msgSender(), "Ownable: caller is not the owner");
    }
}

// 0: 107172
// 1: 131600
// 2: 133760
// 3: 135920
// 4: 138080
// 5: 140246
contract T1 is Ownable {
    function f1() external onlyOwner() {}
    function f2() external onlyOwner() {}
    function f3() external onlyOwner() {}
    function f4() external onlyOwner() {}
    function f5() external onlyOwner() {}
}

// 0: 107172
// 1: 132896
// 2: 135056
// 3: 137216
// 4: 139382
// 5: 141542
contract T2 is Ownable2 {
    function f1() external onlyOwner() {}
    function f2() external onlyOwner() {}
    function f3() external onlyOwner() {}
    function f4() external onlyOwner() {}
    function f5() external onlyOwner() {}
}
```

Its actually costing more !

jnishiyama:
I'm afk all day, so I'll take another look when I am at a computer. That being said, regarding the first set of tests, objectively, Ownable has less code than OwnableV2, so it doesn't make much sense to me that the base contract would be larger for Ownable than it is for OwnableV2. Definitely surprising to me that the behavior we are seeing in those tests is literally the exact opposite of what we should expect from the code. Is it possible the tests are not representing the correct contracts? 

If not, maybe there is a nuance between gas cost and contract size I'm not understanding.

Amxx:
There is a nuance between gas cost and contract size. IMO what the devs really care about is deployment gas costs.

> objectively, Ownable has less code than OwnableV2
In terms of number of lines yes, but not in terms of functionality. If functionalities are equal, and if the gas cost is the same (or very close) I would go for the more readable version, which is the current one.

Adding code, that reduces readability, hoping to save gas, not saving gas, and arguing it's normal because "there is more code" is not the way we want to go.

jnishiyama:
To be clear, the argument is not "there is more code" the argument is that the new version can reduce contract size significantly (depending on usage of modifier). If deployment gas cost is what we are optimizing for then it would seem from your tests that it would make sense to not introduce the change.

jnishiyama:
To verify the claim I would just look at the size of a compiled contract that uses the modifier, we should expect the compiled size to increase by more for each function in a contract that uses the original modifier and the increase should be less for a contract that uses the new modifier.

The idea here being that because the modifier literally has less code, when it is inlined in compilation, it will increase the contract size by less than it did originally. 

Regarding the gas cost at run time, we should expect the new modifier to be slightly more expensive.

I'm sorry I'm afk, but will be happy to help when I'm at a computer.

Amxx:
> To verify the claim I would just look at the size of a compiled contract that uses the modifier, we should expect the compiled size to increase by more for each function in a contract that uses the original modifier and the increase should be less for a contract that uses the new modifier.

That is also what I would expect ... but if you take my T1 and T2 example from above ([here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3347#issuecomment-1148337768)), T1 object is 348 bytes, and T2 is 354 bytes.

Amxx:
And if I add more onlyOwner function, I don't get more savings 431 vs 439 when adding more functions with the modifier

Amxx:
My results might be biased by the function being empty, and the compiler possibly removing some of them. If I put some code in the function's body, then I observe the expected behavior.

Amxx:
Capture d’écran du 2022-06-07 13-44-18

Amxx (approved):
LGTM, 

@frangio?

PR #3380, opened 2022-04-29
Add custom error to `CrossChainEnabledPolygonChild` 

pcaversaccio:
In order to preserve custom error consistency within the `crosschain` contracts, I introduce a new custom error `UnauthorizedCrossChainRelayer` used within `CrossChainEnabledPolygonChild`. The reason why I introduce a new custom error is that I don't want to misuse the existing custom error `InvalidCrossChainSender` and make it more specific to the exact error message.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
  * `Clones`: optimize clone creation ([#3329](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3329))
  * `TimelockController`: Migrate `_call` to `_execute` and allow inheritance and overriding similar to `Governor`. ([#3317](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3317))
- * `CrossChainEnabledPolygonChild`: replace the `require` statement with a new custom error `UnauthorizedCrossChainRelayer`. ([#TBD]())
+ * `CrossChainEnabledPolygonChild`: replace the `require` statement with a new custom error `UnauthorizedCrossChainRelayer`. ([#3380](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3380))
 
 ## 4.6.0 (2022-04-26)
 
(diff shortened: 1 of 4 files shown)


Amxx:
Hello @pcaversaccio and thank you for raising that interesting point.

I realize that the 
```require(msg.sender == _fxChild, "unauthorized cross-chain relay");```
Is in fact equivalent to the other implementation's
```if (!isCrossChain(bridge)) revert NotCrossChainCall();```

This makes me think we should probably not introduce a new error and do
```
    function processMessageFromRoot(
        uint256, /* stateId */
        address rootMessageSender,
        bytes calldata data
    ) external override nonReentrant {
        if (!_isCrossChain()) revert NotCrossChainCall();

        _sender = rootMessageSender;
        Address.functionDelegateCall(address(this), data, "crosschain execution failled");
        _sender = DEFAULT_SENDER;
    }
```

What do you think ?

**Comment by pcaversaccio** [comment]
@Amxx good point - fully agreed; will fix this now! Quick suggestion: why not include the address of the emitter in all custom errors. This helps end-users identify which contract was reverted with a failed transaction, which is especially useful for complex transactions involving multiple contracts? What do you think?

```solidity
error SomeError(address emitter, ..., ..., ...);
emit SomeError(address(this), ..., ..., ...);
```

PS: Just fixed a typo in `Address.functionDelegateCall(address(this), data, "cross-chain execution failed");`.

Amxx:
> Ok - so let's first get this PR merged (if you agree) and thereafter I will open another PR with the custom errors that include the emitter address. 
100% agree

> How would you think about creating a dedicated error interface `ICrossChainErrors` (instead of the current `errors.sol`), where we define all the custom errors used in `crosschain`. This design could be later extended for all the other OZ contracts (e.g. ERC20 etc.) once solc `0.8.4` is supported.

I'm not sure, we'll have to discuss that internally.

PR #3382, opened 2022-04-29
fix Memory-Safe Assembly#3336 in ECDSA.sol file

### LINKED ISSUE #3336: Memory-Safe Assembly
**🧐 Motivation**
Solidity 0.8.13 marked the production readiness of the Yul IR pipeline. This, amongst other things, helps to alleviate stack too deep errors by allowing the compiler to move items to memory as and when needed. 

**📝 Details**
One requirement of the pipeline is that any assembly sections be memory-safe, such that they only use memory that has been previously allocated either by high-level Solidity code or by reading from the free memory pointer at 0x40.

Any memory-safe assembly can be marked as such by annotating it `assembly ("memory-safe") { ... }`. However, this is only possible for code that uses 0.8.13 and above. For code that needs to be backward-compatible, the assembly block can be annotated:

```
assembly {
    ...
}
```

Details about what constitutes memory-safe code is available at https://docs.soliditylang.org/en/v0.8.13/assembly.html#memory-safety.

Would it be possible to review and annotate any uses of assembly within the OpenZeppelin libraries where it is determined to be memory-safe, and to modify then annotate when not?

Kartik0099:
Fixes #3336 

I have add annotation for memory-safe in ECDSA.sol one file currently. due to i just start contribute i want to first test out this file if everything seem to right i will change more files

Thank you 

and please also give me advice and correct me if i make some mistake.
issue No **Memory-Safe Assembly #3336** 
add  /// @solidity memory-safe-assembly this line before
            assembly {

#### PR Checklist

- [x] Tests done
- [ ] Documentation 
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -64,6 +64,7 @@ library ECDSA {
             uint8 v;
             // ecrecover takes the signature parameters, and the only way to get them
             // currently is to use assembly.
+            /// @solidity memory-safe-assembly
             assembly {
                 r := mload(add(signature, 0x20))
                 s := mload(add(signature, 0x40))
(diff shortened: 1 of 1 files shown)


Kartik0099:
hello there,

can anyone tell me did i implement right way? so that i get some motivation and i will do complete this task.

frangio:
Should this be reopened?

Kartik0099:
i am reopening

PR #3384, opened 2022-04-30
Add memory safe assembly annotations 

### LINKED ISSUE #3336: Memory-Safe Assembly

nathan-lapinski:
Part of #3336

Adds memory safe annotations for blocks of inline assembly in the following files:
- Address.sol
- Base64.sol
- Clones.sol
- EnumerableSet.sol
- ERC721.sol
- ERC2771Context.sol
- MinimalForwarder.sol
- StorageSlot.sol
- ~SafeMathMock.sol~

As discussed [here](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3336#issuecomment-1101568318), I am happy to add more files into this PR before merging, but I wanted to open this now to gather feedback.

This is also my first contribution to this repository, so please let me know if I have made any errors.🙏 

#### PR Checklist

- [X] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -23,6 +23,7 @@ library Clones {
      * This function uses the create opcode, which should never revert.
      */
     function clone(address implementation) internal returns (address instance) {
+        /// @solidity memory-safe-assembly
         assembly {
             let ptr := mload(0x40)
             mstore(ptr, 0x602d8060093d393df3363d3d373d3d3d363d7300000000000000000000000000)
@@ -41,6 +42,7 @@ library Clones {
      * the clones cannot be deployed twice at the same address.
      */
     function cloneDeterministic(address implementation, bytes32 salt) internal returns (address instance) {
+        /// @solidity memory-safe-assembly
         assembly {
             let ptr := mload(0x40)
             mstore(ptr, 0x602d8060093d393df3363d3d373d3d3d363d7300000000000000000000000000)
@@ -59,6 +61,7 @@ library Clones {
         bytes32 salt,
         address deployer
     ) internal pure returns (address predicted) {
+        /// @solidity memory-safe-assembly
         assembly {
             let ptr := mload(0x40)
             mstore(ptr, 0x602d8060093d393df3363d3d373d3d3d363d7300000000000000000000000000)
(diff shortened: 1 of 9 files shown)


frangio:
Thank you @nathan-lapinski!

A few comments:
- PR titles should be written like commit messages, e.g. "Add memory safe annotation", and should not include the issue number
- These annotations aren't really needed in the mock, because it's just for testing purposes I think we should remove it to keep the files as clean as possible.

nathan-lapinski:
Sounds good. Thank you for the review, @frangio !

frangio (approved):
Thanks! This looks good. Do you want to add the remaining files from https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3336#issuecomment-1101033850 ?

nathan-lapinski:
Thanks for the review @frangio ! Yes, I’ll go ahead and add the remaining files to this PR.

contracts/proxy/Clones.sol
```diff
@@ -23,6 +23,7 @@ library Clones {
      * This function uses the create opcode, which should never revert.
      */
     function clone(address implementation) internal returns (address instance) {
+        /// @solidity memory-safe-assembly
         assembly {
             let ptr := mload(0x40)
```
nathan-lapinski, line 28: since this is using memory allocated at the free pointer, I think it [should be safe](https://docs.soliditylang.org/en/v0.8.13/assembly.html#memory-safety).

frangio: Exactly!

frangio (approved):
Thank you! The Proxy assembly block shouldn't make a difference.

PR #3398, opened 2022-05-05
Add explanation about _tokensOfOwner

### LINKED ISSUE #1806: Document scalability issues with _tokensOfOwner
`_tokensOfOwner` should not be exposed as an external function because it has cost proportional to the length of the array and can stop working when the array is too large. This should be stated in the function docs.

[Discussion in the community forum.](https://forum.zeppelin.solutions/t/expose-tokensofowner-method-on-erc721enumerable/888)

onlineSoftwareDevOK:
Fixes #1806 

#### PR Checklist

- [ ] Tests
- [ x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -15,6 +15,7 @@ import "../../utils/introspection/ERC165.sol";
  * @dev Implementation of https://eips.ethereum.org/EIPS/eip-721[ERC721] Non-Fungible Token Standard, including
  * the Metadata extension, but not including the Enumerable extension, which is available separately as
  * {ERC721Enumerable}.
+ * _tokensOfOwner should not be exposed as an external function because it has cost proportional to the length of the array and can stop working when the array is too large.
  */
 contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
     using Address for address;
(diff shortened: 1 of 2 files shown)


Amxx (commented):
Hello @onlineSoftwareDevOK 

You are adding implementation-specific details in the interface file, which is supposed to be an implementation-agnostic standard document.

The interface does not include any reference to `_tokensOfOwner` and it shouldn't, as other implementations might do things differently.

I suggest we include this information in the `ERC721.sol`

onlineSoftwareDevOK:
> Hello @onlineSoftwareDevOK
> 
> You are adding implementation-specific details in the interface file, which is supposed to be an implementation-agnostic standard document.
> 
> The interface does not include any reference to `_tokensOfOwner` and it shouldn't, as other implementations might do things differently.
> 
> I suggest we include this information in the `ERC721.sol`

Thanks for your feedback @Amxx I have just changed that. Can you review this PR again, please?

Amxx:
Hello @onlineSoftwareDevOK 

We discussed this PR internally, and we feel like, while this is valuable information, we do not have a place in the doc for this kind of knowledge base. IMO, it should go in some FAQ about contract design and interaction.

We plan to redesign our documentation website (this is long overdue), and we will definitely add this to our requirements when doing this redesign.

PR #3403, opened 2022-05-09
Add `address` to `string` conversion

pcaversaccio:
This PR adds a new function to the [`Strings`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Strings.sol) library that converts an `address` to its ASCII `string` hexadecimal representation. The main motivation is the reason that I've encountered already multiple projects that use a sophisticated `tokenURI` strategy, that concatenates multiple attributes - one of which is an `address` - into a string. An example:

```solidity
return string(abi.encodePacked(_baseTokenURI, toString(creator), "-", tokenId.toString()));
```

#### PR Checklist
- [x] Tests
- [x] Documentation => should be automatically added to [Strings](https://docs.openzeppelin.com/contracts/4.x/api/utils#Strings)
- [x] Changelog entry

### DIFF
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -64,4 +64,11 @@ library Strings {
         require(value == 0, "Strings: hex length insufficient");
         return string(buffer);
     }
+
+    /**
+     * @dev Converts an `address` to its ASCII `string` hexadecimal representation.
+     */
+    function addressToHexString(address addr) internal pure returns (string memory) {
+        return toHexString(uint256(uint160(addr)));
+    }
 }
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -67,7 +67,7 @@ library Strings {
     }
 
     /**
-     * @dev Converts an `address` with fixed length of 20 bytes to its ASCII `string` hexadecimal representation.
+     * @dev Converts an `address` with fixed length of 20 bytes to its not checksummed ASCII `string` hexadecimal representation.
      */
     function addressToHexString(address addr) internal pure returns (string memory) {
         return toHexString(uint256(uint160(addr)), _ADDRESS_LENGTH);
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -69,7 +69,7 @@ library Strings {
     /**
      * @dev Converts an `address` with fixed length of 20 bytes to its not checksummed ASCII `string` hexadecimal representation.
      */
-    function addressToHexString(address addr) internal pure returns (string memory) {
+    function toHexString(address addr) internal pure returns (string memory) {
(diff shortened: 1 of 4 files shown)


contracts/utils/Strings.sol
```diff
@@ -64,4 +64,11 @@ library Strings {
         require(value == 0, "Strings: hex length insufficient");
         return string(buffer);
     }
+
+    /**
+     * @dev Converts an `address` to its ASCII `string` hexadecimal representation.
+     */
+    function addressToHexString(address addr) internal pure returns (string memory) {
+        return toHexString(uint256(uint160(addr)));
```
frangio, line 72: We have to specify a fixed length in case the address starts with zeroes.

frangio (changes requested):
Thank you @pcaversaccio! I agree I can imagine this will be very handy for on-chain NFT metadata.

contracts/utils/Strings.sol
pcaversaccio: You are absolutely right! Fixed it here (including test case): https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3403/commits/ac902373e13ad6118c20a905a70e87a174e43111

Amxx (dismissed):

pcaversaccio:
@Amxx good point - I added the comment! Somehow the CI actions don't work on this PR?

frangio:
@pcaversaccio It may have been a GitHub problem.

contracts/utils/Strings.sol
```diff
@@ -64,4 +65,11 @@ library Strings {
         require(value == 0, "Strings: hex length insufficient");
         return string(buffer);
     }
+
+    /**
+     * @dev Converts an `address` with fixed length of 20 bytes to its not checksummed ASCII `string` hexadecimal representation.
+     */
+    function addressToHexString(address addr) internal pure returns (string memory) {
```
frangio, line 72: Any reason not to call this `toHexString`? It shouldn't conflict with the other function.

pcaversaccio: Before I actually opened this PR, I made the change with function overloading (as you are proposing now) but after thinking through it I decided to deliberately call it `addressToHexString` since all other functions of this library `toString(uint256)`, `toHexString(uint256)`, and `toHexString(uint256,uint256)` assume a `uint256` value type as first parameter. Thus, when I read a function starting with `to` I assumed a `uint256`. I'm not really opinionated about this and if you want I can change it back, but my reasoning was really to make a distinction between `uint256` and `address` input.

Amxx: IMO `toString(address)` would be nice. For library functions like this, I think its reasonable to rely on overloading

pcaversaccio: Fine, fixed :-D

frangio (approved):
Thanks @pcaversaccio! I made some small changes to the tests.

PR #3407, opened 2022-05-10
Update governance.adoc

GiraeffleAeffle:
For 3) we will use GovernorCountingSimple, a module that offers 3 options to voters: For, Against, and Abstain, and where only For and A̶b̶s̶t̶a̶i̶n̶ Against votes are counted towards quorum.


frangio:
The current text is correct:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e633ee9ed3fe45290fb02cdc61eaac3c82b6ba76/contracts/governance/extensions/GovernorCountingSimple.sol#L70

PR #3412, opened 2022-05-15
Add Warning about use of delete

### LINKED ISSUE #3400: Document limitation of delete for EnumerableSet and Map
Solidity currently allows executing the `delete` operator on structs like `EnumerableSet` and `EnumerableMap` even though they contain a mapping that will not be cleared, and the operation will in fact corrupt the struct's data.

According to https://github.com/ethereum/solidity/pull/11843 this operation will no longer be allowed in Solidity 0.9, but in the meantime we should document that `delete` should not be used with these types.

lucasAlonso:
Fixes #3400 

Added Warning about delete operator on structs like EnumerableSet and EnumerableMap even though they contain a mapping that will not be cleared, and the operation will in fact corrupt the struct's data.

According to https://github.com/ethereum/solidity/pull/11843 this operation will no longer be allowed in Solidity 0.9, but in the meantime we should document that delete should not be used with these types.

as user frangio stated, i checked the info and link

#### PR Checklist

- [x] Documentation
check if this Warning is useful and apply.
Need any further warning or issue?

### DIFF
--- a/contracts/utils/structs/EnumerableMap.sol
+++ b/contracts/utils/structs/EnumerableMap.sol
@@ -35,7 +35,7 @@ import "./EnumerableSet.sol";
  * [WARNING]
  * ====
  *  Trying to delete such a structure from storage will likely result in data corruption, rendering the structure unusable.
- *  See ethereum/solidity#11843 for more info
+ *  See https://github.com/ethereum/solidity/pull/11843[ethereum/solidity#11843] for more info.
  *
  *  In order to clean an EnumerableMap, you can either remove all elements one by one or create a fresh instance using an array of EnumerableMap.
  * ====
--- a/contracts/utils/structs/EnumerableMap.sol
+++ b/contracts/utils/structs/EnumerableMap.sol
@@ -72,6 +72,10 @@ library EnumerableMap {
      * Returns true if the key was removed from the map, that is if it was present.
      */
     function remove(Bytes32ToBytes32Map storage map, bytes32 key) internal returns (bool) {
+        // WARNING!
+        // delete function should not be used. It will corrupt the data structure,
+        // clear function should be used instead.
+        // ref https://github.com/ethereum/solidity/pull/11843
         delete map._values[key];
         return map._keys.remove(key);
     }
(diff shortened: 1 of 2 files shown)


lucasAlonso:
Hi Amxx,
at the issue opened (#3400), @SamuelOsewa said 
"I believe that it should be when the delete is used so if someone wants to see the use case they would see the comment as people tend to skip over the introduction."
 Maybe two warnings? or add a more general issue that involves change delete for clear
 This is my first PR, so sorry if I'm overthinking :)

Amxx:
It is important for this warning to show up in the documentation. For that to happen, it must be part of the natspec comments.

lucasAlonso:
[Amxx](https://github.com/Amxx) changed Warning placement, now at top part of the file.
Thxs!

contracts/utils/structs/EnumerableMap.sol
```diff
@@ -31,6 +31,11 @@ import "./EnumerableSet.sol";
  * - `uint256 -> address` (`UintToAddressMap`) since v3.0.0
  * - `address -> uint256` (`AddressToUintMap`) since v4.6.0
  * - `bytes32 -> bytes32` (`Bytes32ToBytes32`) since v4.6.0
+ *
+ *    // WARNING!
+ *       // delete function should not be used. It will corrupt the data structure,
+ *       // clear function should be used instead.
+ *       // ref https://github.com/ethereum/solidity/pull/11843
```
Amxx, line 38: Format is not correct.

See `utils/Address.sol` for how to properly format a warning.

Amxx: Also, delete in not a "function", and there is no `clear` function in this structure.

I would suggest something like:

> Trying to delete such a structure from storage will likely result in data corruption, rendering the structure unusable.
> See https://github.com/ethereum/solidity/pull/11843 for more info
>
> In order to clean an EnumerableMap, you can either remove all elements one by one or create a fresh instance using an array of EnumerableMap.

lucasAlonso: Changed wording and format.
maybe add an issue at contracts/utils/cryptography/MerkleProof.sol to make Warning's Format consistent ?

Amxx (approved):

PR #3413, opened 2022-05-16
Fix return parameter

martriay:
The return variable name of the `TimelockController`'s `isOperation` function is misnamed. I propose changing it for `registered`.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/57725120581e27ec469e1c7e497a4008aafff818/contracts/governance/TimelockController.sol#L134


Amxx:
LGTM, and change to return variable names are not breaking changes IMO.

@frangio any opinion?

frangio:
Definitely not a breaking change. Shouldn't we remove the return variable name completely though? We don't use them throughout the codebase.

Amxx (approved):

PR #3414, opened 2022-05-16
Update: the @dev comments for add functions

Rassska:
Fixes #???? 
Dynamically allocated arrays need to increase their size whenever there is no free space, therefore it takes additionally O(n) to handle that and the overall time complexity for adding elements becomes O(1) amortized.

Thanks, Rasul!

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -46,7 +46,7 @@ library EnumerableSet {
     }
 
     /**
-     * @dev Add a value to a set. O(1).
+     * @dev Add a value to a set. O(1) amortized.
      *
      * Returns true if the value was added to the set, that is if it was not
      * already present.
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @Rassska 

I'm not sure you understand how solidity arrays work. There is no "chunk allocation" like with C++ `std::vector`. These functions are O(1) in the worst case.

Amxx:
For the record, lets see the worst case

```
function _add(Set storage set, bytes32 value) private returns (bool) {
        if (!_contains(set, value)) { // ssload (constant cost)
            set._values.push(value); // 2 * sstore (constant cost). In the worst case both slots are empty (length + data)
            set._indexes[value] = set._values.length; // sload (hot) + sstore (constant cost)
            return true; // constant cost
        } else {
            return false; // constant cost
        }
    }
```

PR #3338, opened 2022-04-12
Extend enumerable maps by uint to uint map

ashhanai:
Create `UintToUintMap` enumerable map type. This type is useful for cases where you want to map e.g. id to balance.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

PR #3416, opened 2022-05-17
Add Utils Struct EnumerableMap  Bytes32ToUintMap

arturictus:
Adding **Bytes32ToUintMap** in struct helpers.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/structs/EnumerableMap.sol
+++ b/contracts/utils/structs/EnumerableMap.sol
@@ -31,6 +31,7 @@ import "./EnumerableSet.sol";
  * - `uint256 -> address` (`UintToAddressMap`) since v3.0.0
  * - `address -> uint256` (`AddressToUintMap`) since v4.6.0
  * - `bytes32 -> bytes32` (`Bytes32ToBytes32`) since v4.6.0
+ * - `bytes32 -> uint256` (`Bytes32ToUintMap`) since v4.x.x
  */
 library EnumerableMap {
     using EnumerableSet for EnumerableSet.Bytes32Set;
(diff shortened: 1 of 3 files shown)


Amxx:
Hello @arturictus 

Can you tell us more about your usecase for a bytes32 → uint256 map?

arturictus:
hello @Amxx,
Of course.
I'm building a booking contract which allow to book one night per year.
The approach I'm taking is:
- the frontend will supply (uint16 year, uint16 dayOfTheYear [1-366])
- The bookings and cancellations can not be made to passed dates.
- Each wallet can only book once per year-dayOfTheYear

I use the map as
- key: ```bytes32(abi.encode(year, dayOftheYear))```
- value: timestamp (calculated timestamp to compare to block.timestamp)

This allows me:
 - validate uniqueness of the reservation
 - Validate time constraints with the timestamp
 - create, update and cancel reservations with `O(1)` complexity
- Be able to iterate over reservations for calculations (done in the client)

I can imagine other use cases:
- key: productID, value: price
- key: saleID, value: time until applicable
- etc, etc

Happy to hear your thoughts

Amxx:
Thanks for the details. In your case, I would argue that the key could be a uint corresponding to the day's "timestamp". I would consider that safer then `bytes32(abi.encode(year, dayOftheYear))` (you are sure you're not doing a hash here ?)

We also have #3338. We'll discuss new mapping types soon

arturictus:
Hi @Amxx,
Thanks for your response.
Answering your comments:

>  I would argue that the key could be a uint corresponding to the day's "timestamp"

The problem with that is that a day has `86400` seconds which I can not use timestamp for unique key. (there are some workarounds, but I'm not happy with them)

> you are sure you're not doing a hash here ?

Still considering to hash it unless I need to get the original key.
The inputs are validated so they are quite limited in range.

Amxx:
Could you please add a changelog entry?

contracts/utils/structs/EnumerableMap.sol
```diff
@@ -342,4 +343,98 @@ library EnumerableMap {
...
+    function at(Bytes32ToUintMap storage map, uint256 index) internal view returns (bytes32, uint256) {
+        (bytes32 key, bytes32 value) = at(map._inner, index);
+        return (key, uint256(value));
+    }
+
+    /**
+     * @dev Tries to returns the value associated with `key`.  O(1).
+     * Does not revert if `key` is not in the map.
+     *
+     * _Available since v3.4._
```
Amxx, line 409: That is not correct, all there new function fill be part of the next 4.7 release

arturictus:
> Could you please add a changelog entry?

Yes!, done :)

Amxx (approved):

PR #3429, opened 2022-05-23
Procedurally generate EnumerableSet and EnumerableMap

Amxx:
#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Now that we have gas reporting, with this PR it should be easy to compare the cost of our current approach (cast everything to a Bytes32Set) versus duplicating the code for every type.

Amxx:
We discussed that, and casting, while not being a breaking change on the solidity side would not be supported by the upgrades plugin.

I would do this PR in 4.X, and then do the change for casting (depending on the gas outcome) in 5.0

frangio (approved):
LGTM

nventuro:
I love everything about this

PR #3431, opened 2022-05-24
For gas optimization

rotcivegaf:
## Gas optimization for `for` loop
- Save the length in a local var to save the `SLOAD`
- Make unchecked the cycle increment/decrement
- Use `++i` instead of `i++` to save gas
- Save storage vars in a local var to save the `SLOAD` when they haves more than one use

#### PR Checklist

- [X] Tests
- [X] Documentation
- [X] Changelog entry

### DIFF
--- a/contracts/finance/PaymentSplitter.sol
+++ b/contracts/finance/PaymentSplitter.sol
@@ -48,11 +48,16 @@ contract PaymentSplitter is Context {
      * duplicates in `payees`.
      */
     constructor(address[] memory payees, uint256[] memory shares_) payable {
-        require(payees.length == shares_.length, "PaymentSplitter: payees and shares length mismatch");
-        require(payees.length > 0, "PaymentSplitter: no payees");
+        uint256 payeesLength = payees.length;
+        require(payeesLength == shares_.length, "PaymentSplitter: payees and shares length mismatch");
+        require(payeesLength > 0, "PaymentSplitter: no payees");
 
-        for (uint256 i = 0; i < payees.length; i++) {
+        for (uint256 i = 0; i < payeesLength;) {
             _addPayee(payees[i], shares_[i]);
+
+            unchecked {
+                ++i;
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


eugenioclrc:
Lanice hack

frangio:
I don't think you noticed but the for loops you're modifying are not in storage but in memory. (I haven't looked through all of them because this is a large PR but I couldn't find a storage one.)

This means there is no SLOAD to save, rather an MLOAD which only costs 3 gas. Note that 3 gas is also the cost of stack operations like PUSH and DUP. In order to execute MLOAD you first need to PUSH or DUP the memory position on the stack so the total cost of loading something from memory is closer to 6 rather than 3 gas (at least as a lower bound, assuming the memory position is not at the top of the stack to begin with). 

Thus, at least in theory, it makes sense to store the length on the stack instead of reading it from memory every time, because it reduces the cost of the operation by half. In absolute terms, we are still talking about 3 gas, which may be neutralized by other things the compiler does, so this is the kind of optimization that we reject.

I ran some tests to understand the numbers.

```solidity
pragma solidity 0.8.13;

contract test0 {
    function foo1(uint[] memory a) external returns (uint x) {
        for (uint i = 0; i < a.length; i++) {
            x += a[i];
        }
    }
}
```

I compiled with optimizations. I gradually made the changes suggested in this PR: a) cache the length in a variable, b) use pre-increment, c) use unchecked. Each time I invoked the function with an array of 100 elements (the numbers from 0 to 99).

Each change resulted in the following reduction in gas used _per iteration_ (i.e. 1/100th of the total):

a) 2.92
b) 5
c) 63

Note how caching the length resulted in exactly the same improvement that I theorized earlier (3 gas difference).

I also tried unchecked with post-increment and found it cost exactly the same as unchecked with pre-increment.

The most significant improvement comes from unchecked, though unfortunately it has too much of an effect on code readability. For now, we prefer to wait until Solidity introduces unchecked expressions. (https://github.com/ethereum/solidity/issues/10698) 

---

For future suggestions of gas optimizations, please take this comment as a model of how we expect it to be researched and explained. "Use ++i instead of i++ to save gas" is not acceptable.

PR #3432, opened 2022-05-24
!= 0 to > 0

rotcivegaf:
## Gas optimization for require() statements
- != 0 to > 0 for uint256 to save gas

Gas save: 6 for each require

#### PR Checklist

- [X] Tests
- [X] Documentation
- [X] Changelog entry

### DIFF
--- a/contracts/finance/PaymentSplitter.sol
+++ b/contracts/finance/PaymentSplitter.sol
@@ -130,7 +130,7 @@ contract PaymentSplitter is Context {
         uint256 totalReceived = address(this).balance + totalReleased();
         uint256 payment = _pendingPayment(account, totalReceived, released(account));
 
-        require(payment != 0, "PaymentSplitter: account is not due payment");
+        require(payment > 0, "PaymentSplitter: account is not due payment");
 
         _released[account] += payment;
         _totalReleased += payment;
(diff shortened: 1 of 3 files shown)


frangio:
For unsigned integers, `x > 0` and `x != 0` are provably equivalent, so this is an optimization the compiler can automatically make.

Compiling the following code with optimizations on, both implementations have the same gas cost. So indeed this is an optimization the compiler is already doing.

```solidity
pragma solidity 0.8.7;

contract Test1 {
    function test(uint x) pure external returns (uint) {
        if (x > 0) {
            return 5;
        } else {
            return 7;
        }
    }
}

contract Test2 {
    function test(uint x) pure external returns (uint) {
        if (x != 0) {
            return 5;
        } else {
            return 7;
        }
    }
}
```

rotcivegaf:
You example are correct but the improve is in the `require()` statements
When the optimizer is enabled, gas is wasted by doing a greater-than operation, rather than a not-equals operation inside `require()` statements
When Using `!=`, the optimizer is able to avoid the EQ, ISZERO, and associated operations, by relying on the JUMPI that comes afterwards, which itself checks for zero

Example:
```solidity
pragma solidity 0.8.11;

contract test0 {
    function foo(uint256 x) external {
        require(x != 0);
    }
}

contract test1 {
    function foo(uint256 x) external {
        require(x > 0);
    }
}
```
```
Solc version: 0.8.11
Optimizer enabled: true
Runs: 200
Block limit: 30000000 gas

test0, foo: 21422 gas
test1, foo: 21428 gas
```

frangio:
Compiling this latest snippet with 0.8.13 instead of 0.8.7 results in the same gas cost for both implementations.

Unfortunately it seems to normalize them both to the more expensive version? But this is something that should be reported to the compiler.

rotcivegaf:
Well it seems that from version 0.8.13 it is fixed
I would prefer to use the cheap version for those who use a pragma < 0.8.13

frangio:
Users concerned about this level of optimization should usually be on the latest compiler version to benefit from this as well as other improvements.

Sorry, we don't want to introduce obscure code changes for small compiler-dependent optimizations.

PR #3434, opened 2022-05-25
Remove redundant require in ERC721

rotcivegaf:
Remove redundant require in `ERC721` contract, `_isApprovedOrOwner` function

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio:
@rotcivegaf @Amxx What do you think about changing all revert reasons that are "__ query for nonexistent token" to instead say "token id not minted"?

frangio:
Realizing we can treat my previous suggestion independently of this PR.

frangio (approved):

Amxx:
I don't have a strong opinion.

My only concern is that à token could be minted then burned ... and saying that it was not minted could be confusing.

PR #3438, opened 2022-05-27
Simplify ERC721 revert reasons

frangio:
Also adds a helper to check that a token ID is minted / exists.


Amxx:
I'm not sure "token ID not minted" is the right message here. A token that has been minted and then burned doesn't exist anymore. IMO "token ID not minted" can be interpreted as "the ID was not minted" which in that case would be wrong.

Amxx:
I still prefer "nonexistent token ID" or "invalid token ID"

Amxx (approved):

Issue #3441, opened 2022-05-30
Merkle multi proofs improvements


frangio:
We can reorder arguments to be more consistent with the single proof version:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/82a63f63890d8b136478158ca12346d396154855/contracts/utils/cryptography/MerkleProof.sol#L59-L64

We should consider making it `calldata` only to encourage the more efficient use by default.

PR #3447, opened 2022-06-01
Reorder arguments for multiProof functions

frangio:
Resolves part of https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3441.

Reorders arguments for consistency with simple proofs.

### DIFF
--- a/contracts/mocks/MerkleProofWrapper.sol
+++ b/contracts/mocks/MerkleProofWrapper.sol
@@ -31,18 +31,18 @@ contract MerkleProofWrapper {
 
     function multiProofVerify(
         bytes32[] calldata proofs,
+        bool[] calldata proofFlag,
         bytes32 root,
-        bytes32[] calldata leafs,
-        bool[] calldata proofFlag
+        bytes32[] calldata leafs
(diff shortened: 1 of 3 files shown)


Amxx:
While I understand the logic here, I must say I'm not a fan of this new format. 

In particular because it separates the proof and the proofFlag arguments. The thing is, in multiProofVerify, the `proof` array should possibly be called "proofSiblings", and it is useless without the "proofFlags" array.

Afaik, the idea for having the proof being the first argument of the single version was to be able to do:
```
using Merkle for bytes32[];
...
proof.verify(root, leaf)
```

IMO this was a bad design from the start. Rater then proving a property on the proof, using the root, I would rather prove a property on the tree itself (namely that it contains a leaf) using the proof as an argument.

So I would personally have preferred 
- `root.verifyInclusion(leaf, proof)` 
- `leaf.rebuildRoot(proof)`
- `root.verifyIncusion(leaves, proofSiblings, proofFlags)` (overloaded)
- `leaves.rebuildRoot(proofSiblings, proofFlags)` (overloaded)

Amxx (approved):

PR #3450, opened 2022-06-02
Simplify Initializable

frangio:
In #3344 we introduced changes to Initializable in order to enable the use of `_disableInitializers()` in the constructor of multiple parents:

```solidity
contract A {
    constructor() { _disableInitializers(); }
}
contract B {
    constructor() { _disableInitializers(); }
}
contract C is A, B {
    constructor() { _disableInitializers(); }
}
```

However, the resulting code was difficult to reason about, and it introduced a bug when mixing `_disableInitializers` and the old pattern `constructor() initializer {}`.

We sat down to rethink it and trim it down to more straightforward code.

This PR slightly changes how functions with the `initializer` and `reinitializer` modifiers can be nested in the context of the constructor.
- `initializer` functions work the same as 4.6: they can be nested in this context. This is present for backwards compatibility with versions prior to 4.4.1 and we may remove it in 5.0.
- `reinitializer` functions can't be nested. If one is invoked in the context of another, it will revert. This is a small breaking change for `reinitializer(1)`. We don't think anyone will be affected by this change given that reinitializers were released only 1 month ago and the broken pattern was not a recommended one. It allows us to make the code much simpler.


test/proxy/utils/Initializable.test.js
```diff
@@ -184,4 +186,11 @@ contract('Initializable', function (accounts) {
       expect(await this.contract.child()).to.be.bignumber.equal(child);
     });
   });
+
+  describe('disabling initialization', function () {
+    it('old and new patterns dont interfere', async function () {
```
Amxx, line 191: I'm not sure this is the right name for this `describe` & `it`. AFAIK, these tests show that they do interfere

contracts/proxy/utils/Initializable.sol
```diff
@@ -127,27 +129,8 @@ abstract contract Initializable {
...
-        );
-
-        if (isTopLevelCall) {
-            _initialized = version;
-        }
-
-        return isTopLevelCall;
+        require(!_initializing, "Initializable: contract is initializing");
+        _initialized = type(uint8).max;
+        emit Initialized(type(uint8).max);
```
Amxx, line 134:
```suggestion
        require(!_initializing, "Initializable: contract is initializing");
        if (_initialized < type(uint8).max) {
            _initialized = type(uint8).max;
            emit Initialized(type(uint8).max);
        }
```

Amxx: this could avoid writing the same value multiple times.

Amxx (approved):

PR #3452, opened 2022-06-03
Update SafeMath.sol, added percentage function.

fps8k:
### **OPENZEPELLIN CONTRIBUTION**

Added a simple percentage function to the [SafeMath.sol](https://github.com/fps8k/openzeppelin-contracts/blob/master/contracts/utils/math/SafeMath.sol).

🎉

### DIFF
--- a/contracts/utils/math/SafeMath.sol
+++ b/contracts/utils/math/SafeMath.sol
@@ -224,4 +224,34 @@ library SafeMath {
             return a % b;
         }
     }
+
+
+
+    /*
+    * @dev
+    *
+    * Returns a number's percentage of another number to a specified decimal place.
+    * The result is to the decimal place of the third parameter passed to the function.
+    * 
+    *
+    * @param:
+    *
+    * uint256 a, uint256 b, uint8 c, string memory errorMessage.
+    * It returns tha `a`% of `b` to `c` decimal places.
+    */
+    function perc(uint256 a, uint256 b, uint8 c, string memory errorMessage) internal pure returns(uint256)
+    {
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @fps8k 

If you are afraid of overflows, you should use `Math.mulDiv`. For simpler cases, where overflow is not an issue, I really don't think a dedicated function is needed.

contracts/utils/math/SafeMath.sol
```diff
@@ -224,4 +224,34 @@ library SafeMath {
...
+    *
+    * uint256 a, uint256 b, uint8 c, string memory errorMessage.
+    * It returns tha `a`% of `b` to `c` decimal places.
+    */
+    function perc(uint256 a, uint256 b, uint8 c, string memory errorMessage) internal pure returns(uint256)
+    {
+        // Require that b is greater than 0 to avoid zero division.
+        require(b > 0, errorMessage);
+        // Check for overflows.
+        require((a * b * (10 ** c)) < type(uint256).max, errorMessage);
```
Amxx, line 247: this check doesn't really make sens. 

If `(a * b * (10 ** c)) > type(uint256).max`, the multiplication will revert because of solidity 0.8.x overflow protection... So the only case where this would revert is not equality ... why would you have a specific revert is equality?

fps8k: Thank you for your reply.

Amxx: well, if there is no overflow protection at the solidity level, and overflow happens, this test will not catch it.

PR #3455, opened 2022-06-06
Remove runtime check of ERC1967 storage slots

### LINKED ISSUE #3125: Remove runtime validation of ERC1967 slots in proxies
Following up on the forum [discussion](https://forum.openzeppelin.com/t/release-candidate-for-contracts-4-5-open-review-period/22700/3) - thx for the answer and the note about abstract contracts and constructors Frangio!

The tl;dr is that I wanted to start a constructive discussion around the current implementation of ERC1967.

Particularly, I feel that not validating the storage [slots](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/783ac759a902a7b4a218c2d026a77e6a26b6c42d/contracts/proxy/ERC1967/ERC1967Upgrade.sol#L28) at the `ERC1967Upgrade` contract level and simply delegating the validation on the contracts implementing it can lead to people forgetting about the validation or to people deliberately "forgetting" - to, perhaps, change one of the slots to create a storage clash.

I understand this last scenario might be out of the scope of the intention of the library but I also like to think about the library as a place that people can reference to "validate" code they find in the wild so I think the validation could serve as a way to raise awareness about the perils of using unstructured storage.

Looking forward to hearing your thoughts!

Amxx:
Fixes #3125

#### PR Checklist

- [x] Changelog entry


frangio (approved):
LGTM. Thanks!

PR #3458, opened 2022-06-09
Add a suggestion to disallow transfers to the contract address

### LINKED ISSUE #3157: ERC20 contract loses tokens sent to the contract's own address
**🧐 Motivation**
Many people lose their ERC20 tokens by mistakenly sending their tokens back to the token contract address (presumably to redeem their tokens). However, there is no way to recover those tokens.
https://np.reddit.com/r/ethereum/comments/sfz4kw/did_i_just_lose_half_a_million_dollars_by_sending/

**📝 Details**
ERC20 implementation should reject such transfers.
For example, we may add a `require` statement to the `_transfer` function in [ERC20.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4a9cc8b4918ef3736229a5cc5a310bdc17bf759f/contracts/token/ERC20/ERC20.sol).

```solidity
    function _transfer(
        address sender,
        address recipient,
        uint256 amount
    ) internal virtual {
        ...
        require(recipient != address(this), "ERC20: transfer to the token contract address");
        ...
    }
```

qbzzt:
Fixes #3157 (or at least reduces the risk)

#### PR Checklist

- [ ] Tests
- [X] Documentation
- [ ] Changelog entry

### DIFF
--- a/docs/modules/ROOT/pages/erc20.adoc
+++ b/docs/modules/ROOT/pages/erc20.adoc
@@ -81,3 +81,22 @@ transfer(recipient, 5 * (10 ** 18));
 A preset ERC20 is available, xref:api:presets#ERC20PresetMinterPauser[`ERC20PresetMinterPauser`]. It is preset to allow for token minting (create), stop all token transfers (pause) and allow holders to burn (destroy) their tokens. The contract uses xref:access-control.adoc[Access Control] to control access to the minting and pausing functionality.  The account that deploys the contract will be granted the minter and pauser roles, as well as the default admin role.
 
 This contract is ready to deploy without having to write any Solidity code.  It can be used as-is for quick prototyping and testing, but is also suitable for production environments.
+
+[[Recommendations]]
+== Recommendations
+
+* If your contract does not need to hold any tokens, consider adding to the `_transfer` function a check that prevents token from being sent to the contract address. It is extremely common for users to send tokens to the contract by mistake, and then they cannot be retrieved (see, for example, https://optimistic.etherscan.io/address/0x4200000000000000000000000000000000000042#tokentxns).
+
+  ```solidity
+      function _transfer(
+        address from,
+        address to,
+        uint256 amount
+    ) internal virtual {
+        require(from != address(0), "ERC20: transfer from the zero address");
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @qbzzt 

I'm not sure if we want to include such a recommendation in our docs, but if we want, then the piece of code in the example is really not how people should do it. 

In your example, you modify the ERC20.sol contract, which come from some package manager and should not be tampered with.
If you want to disable transfers to some address, you should probably do it by overriding the "beforeTokenTransfer" hook

```
      function _beforeTokenTransfer(address from, address to, uint256 amount) internal virtual {
          require(to != address(this));
      }
```

This is simpler to do than changing the core code, and it will also prevent minting directly to the contract.

Amxx:
I'm not sure an extensions is needed. 

IMO developers can just add this `_beforeTokenTransfer` override directly in their contract. The wizard could possibly also have a toogle for that.

qbzzt:
The reason I'm trying to get this added *somewhere* is that in many cases developers don't add this feature, even though they don't need the contract to hold tokens and it is a common usage mistake. Mass adoption requires being clueless user friendly.

Where is the source code for the wizard? Adding it there could also solve the problem.

frangio:
I'm really thinking we should just add `ERC20SafetyRails` (naming TBD) into the library. If we think this should be standard, that is one of the strongest ways to make the recommendation.

@Amxx Do you see issues with adding this as a separate contract that can be opted in? It can have a Wizard toggle, and we can consider enabling it by default.

By the way the source for for Wizard is at [OpenZeppelin/contracts-wizard](https://github.com/OpenZeppelin/contracts-wizard).

MarkP42:
Hi Guys,
I'm a programmer but new to crypto.
I'm wondering why you wouldn't put the contract address check in the before..Transfer function in the main (super) ERC20 contract? I've probably missed something haven't I?

frangio:
@MarkP42 No, you're right. This PR needs to be edited so that the suggested check is added in a `_beforeTokenTransfer` hook as opposed to `_transfer`.

qbzzt:
I decided to write this [as an ethereum.org article](https://ethereum.org/en/developers/tutorials/erc20-with-safety-rails/) instead.

PR #3459, opened 2022-06-09
Add slippage protection for EOAs interactions with ERC4626 vaults

Amxx:
ERC4626 states that

> If implementors intend to support EOA account access directly, they should consider adding an additional function call for deposit/mint/withdraw/redeem with the means to accommodate slippage loss or unexpected deposit/withdrawal limits, since they have no other means to revert the transaction if the exact output amount is not achieved.

This should possibly be standardized in an ERC.

#### PR Checklist

- [X] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
I'm not convinced we should add non-standard functions like these. We should discuss other alternatives to this issue.

PR #3465, opened 2022-06-12
added function which returns the index of array from given array

### LINKED ISSUE #3464: to add feature that can find index of array
In Array.sol there is nothing that we can find index of an element from a given element. I think it should be integrated

haidaralimasu:
Fixes #3464

I have added a function which returns the index of an element from a given array

the function loop through the array and find the number exists in the array and return two values one is boolean and other is the uint

### DIFF
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -45,4 +45,18 @@ library Arrays {
             return low;
         }
     }
+
+    /**
+     * @dev Returns the index of element from an array
+     * it's search from the whole array and returns the index of the element in the given array
+     */
+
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @haidaralimasu 

This function has linear complexity, and can therefore become really expensive to run if the array length is too large. Using that in a smart contract can lead to critical issues, where some functions are no longer possible to run due to gas costs being too high.

**We strongly recommend NOT to use that in any contract**

If you need to do lookups for elements in arrays, and if you are ok with using a Set (array with no duplicate) then you should probably use `https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol`.

PR #3468, opened 2022-06-13
Address.functionCall internal call stack simplification

ZumZoom:
It is a bit cheaper to call `functionCallWithValue` directly than via extra intermediate internal call. It also allows to save a bit on contract size in case of `functionCall(address,bytes,string)` not being used anywhere in the contract.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -83,7 +83,7 @@ library Address {
      * _Available since v3.1._
      */
     function functionCall(address target, bytes memory data) internal returns (bytes memory) {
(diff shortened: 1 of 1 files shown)


Amxx:
Thank you @ZumZoom 

Do have you measured the gas savings? If so, can you share the numbers?

Amxx:
@ZumZoom Thank for these number. That is great !

Amxx (approved):

PR #3470, opened 2022-06-13
[next] Remove presets

Amxx:
#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


Amxx:
Tests fail because of the contract migration script.

The migration script rewrite imports between version. From version 3 to version 4, the presets path are rewritten. If we remove the presets, we need to consider how to deal with that. Error ? No rewrite ?

Issue #3410, opened 2022-05-15
Add priority queue as a data structure to be able to implement max/min heaps


Rassska:
Hi, there! Thank you so much for such great source!

**🧐 Motivation**

I'm operating with huge amount of data in my smart contracts and sometimes it's pretty useful to handle queries by simply implementing common data structures in terms of saving more gas. We already do have a dequeue/bitmaps impls and now I want to add a priority queue if it's possible.  

Let me know, what do you think about it <3

**📝 Details**

frangio:
A heap data structure would be a pretty cool addition. O(log n) queue and dequeue operations are acceptable.

Can you please give example use cases?

Rassska:
> A heap data structure would be a pretty cool addition. O(log n) queue and dequeue operations are acceptable.
> 
> Can you please give example use cases?

Okay, I'll provide it a bit later.
Also, I do have an extra question about sorting algorithms. Why they are not presented in openzeppelin?

frangio:
Sorting algorithms are O(n log n) so they may run out of gas with large arrays, which may be manipulable by users and become an attack vector. Specific contracts may have reasonable bounds on array length but we err on the conservative side to help people avoid creating an attack vector unknowingly.

A min/max heap will be a good alternative for use cases that require sorting.

Rassska:
> A heap data structure would be a pretty cool addition. O(log n) queue and dequeue operations are acceptable.
> 
> Can you please give example use cases?

Of course it's possible to implement binary heap, but I wanted to consider a fibonacci heap.

frangio:
Hm, I'm not sure Fibonacci heaps are acceptable for this context. I have to admit I'm not super familiar with them but they apparently can have high overhead which will probably translate to expensive storage operations, and worst case complexity is linear.

Bijan-Massoumi:
This would be fantastic addition

frangio:
@Bijan-Massoumi Please share your use case so we can consider it for prioritization!

rocketmanh:
I just implemented this w/associated tests. Will open a PR / can share code if y'all want. 

Use case is an open submission box that anyone can submit to, and a specific set of accounts can vote on submissions, and gas-efficiently pull off the top voted submission.

rocketmanh:
https://github.com/collectivexyz/revolution-protocol/blob/main/packages/revolution-contracts/src/MaxHeap.sol cc @Rassska

rocketman-21:
[...]

PR #3475, opened 2022-06-15
Addition of Array Sorting functions

fps8k:
## Open-Zeppelin Feature.

Addition of Array Sorting functions.

## Environment used.

- Compiler version: 0.8.7
- Target EVM version (as per compiler settings): default
- Framework/IDE (e.g. Truffle or Remix): Remix
- Operating system: Windows

## Array Sorting functions.

After going through the [Array.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Arrays.sol), it felt quite necessary that a function that would sort arrays in ascending or descending order would be necessary.

That is what this pull request seeks to add.

I would be glad, if the Open-Zeppelin team reviewed, corrected and accepted my code.

# How can this feature help?.

Sorting arrays directly from the smart contract when necessary, especially when an array (which might seem unordered) is wanted to be passed to the findUpperBound function.

###

Thank you.

Anthony (fps).

### DIFF
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -45,4 +45,245 @@ library Arrays {
             return low;
         }
     }
+    
+    
+    /*
+    * @title: Array Sorter.
+    * @author: Anthony (fps) https://github.com/fps8k.
+    * @dev:
+    * This library has two main functions that take an array [unordered, mostly] as an argument.
+    * It sorts the array in ascending or descending order, depending on the user's choice.
+    * Then returns the new array created from the old array.
+    *
+    * [GAS USAGE]
+    * This gas test was done with a constant unordered array of 8 elements.
+    * Ascending: ~ 152696 gas.
+    * Descending: ~ 125730 gas.
+    *
+    * Function sortAsc (Sort Ascending).
+    * It takes in an array of numbers, runs some checks and passes it as an argumant to the function _sortAsc().
+    * The called function [_sortAsc()] arranges the array elements from lowest to highest.
+    * This returns the ordered array returned by the _sortAsc() function.
+    * This uses ~ 152696 gas.
+    *
+    * @param:
+    * uint256[] _array -> Array to be sorted.
+    *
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @fps8k 

There are many things to say here:
- your implementation is a selection sort, which is known to be 0(n²) ... this is bad. If we were to ever implement that, we should go with an in-place quicksort. [I wrote a solidity implementation of quicksort a while back](https://gist.github.com/Amxx/d3a99fcb79abbe3c76a2f2a5773b3815).
- you read from storage, which makes everything super expensive, particularly in the context of a selection sort where there are 0(n²) reads.
- empty arrays are already sorted, reverting if the array is empty makes no sense.

More importantly, I don't think we should provide a sorting function at all. Sorting is expensive, and I would argue devs should design their contract so that sorting is not needed.

If sorting was to be implemented on-chain, it should be done in a way that minimizes the storage change when inserting/extracting elements. So probably some tree (red-black would be my first guess). Arrays are a terrible choice. Lastly, if you really want to sort an array, do it using an O(n log(n)) algorithm, and in memory (not storage !!!)

We always ask contributors to open issues before creating a PR. Doing so would have saved you time and effort.

Amxx:
From your own metrics, sorting an array of 8 elements cost >150k gas. That is 18k gas per element, and its going to grow quadratically. 

That is:
16 elements: ~600k gas
32 elements: ~2.4m gas
64 elements: ~9.6m gas

I hope you realize this is not something we can release. People will build contracts on top of that, reach 120~150 entries onchain ... and then running the sorting algorithm would fit an entire block out of gas ... which will block all operation that rely on this sorting.

fps8k:
Thank you for your reviews. I will try to be better.

Amxx:
Before submitting PR, please open issues so we can discuss the overall design first.

Amxx:
Related issue: #3410

fps8k:
I will, next time.

PR #3480, opened 2022-06-15
Certora's 3rd Round of Formal Verification

teryanarmen:
Certora's 3rd round of verification for Open Zeppelin's contracts. In this round we verified:
Initializable.sol
GovernorPreventLateQuorum.sol
ERC1155Burnable.sol
ERC1155Pausable.sol
ERC1155Supply.sol

We also determined no formal verification needed for:
ERC1155Holder.sol
ERC1155Receiver.sol


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
Amxx, line 1: Is that file needed for running the certora workflow ? I'd guess not.

certora/helpers/DummyERC20B.sol
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: agpl-3.0
+pragma solidity ^0.8.0;
+import "./DummyERC20Impl.sol";
+
+contract DummyERC20B is DummyERC20Impl {}
```
Amxx, line 5: Is that needed? is `DummyERC20A` not enough?

certora/helpers/DummyERC20Impl.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: agpl-3.0
+pragma solidity ^0.8.0;
+
+// with mint
+contract DummyERC20Impl {
```
Amxx, line 5: I'm missing the point of this contract. We have an ERC20 impl, why do we need that ?

certora/helpers/DummyERC20Impl.sol
```diff
@@ -0,0 +1,57 @@
...
+    uint256 t;
+    mapping (address => uint256) b;
+    mapping (address => mapping (address => uint256)) a;
+
+    string public name;
+    string public symbol;
+    uint public decimals;
```
Amxx, line 12: the ERC says decimals should be `uint8`, not `uint`

certora/scripts/Round1/verifyAll.sh
```diff
@@ -16,7 +18,7 @@ do
                 certoraRun certora/harnesses/ERC20VotesHarness.sol certora/harnesses/$contractFile \
                 --link ${contractFile%.*}:token=ERC20VotesHarness \
                 --verify ${contractFile%.*}:certora/specs/$specFile "$@" \
-                --solc solc8.2 \
+                --solc solc \
```
Amxx, line 21: Other scripts, such as `certora/scripts/Round1/sanity.sh` and `certora/scripts/Round1/sanityTokens.sh` uses `solc8.2`.
We should unify that.

certora/harnesses/AccessControlHarness.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.5.0) (access/AccessControl.sol)
+
+pragma solidity ^0.8.0;
+
+import "../munged/access/AccessControl.sol";
```
Amxx, line 6: munged contracts are not part of this PR. What are they? How are they generated? Can they be replaced with the `contracts-exposed` ?

.github/workflows/verify.yml
```diff
@@ -0,0 +1,65 @@
+name: Certora
+
+on:
+  push:
+    branches:
+      - master
+      - main
+      - certora/erc20
+      - certora/erc1155ext
```
Amxx, line 9: do we want to run that on every PR ?

.github/workflows/verify.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Install java
+        uses: actions/setup-java@v1
+        with: { java-version: "11", java-package: jre }
+
+      - name: Install certora
+        run: pip install certora-cli
+
+      - name: Install solc
+        run: |
+          wget https://github.com/ethereum/solidity/releases/download/v0.8.4/solc-static-linux
```
Amxx, line 31: solidity version ?

.vscode/settings.json
teryanarmen: It is not. I'll edit.

certora/helpers/DummyERC20B.sol
teryanarmen: Sometimes we need to have 2 separate token contracts as they have their own state. I will check if its used and can remove it if its unnecessary.

certora/helpers/DummyERC20Impl.sol
teryanarmen: This was created before as a basic implementation. We use this on most projects but are working on having a more diverse set up ERC20 implementations. i can replace it with OZ implementation if you'd like.

certora/helpers/DummyERC20Impl.sol
teryanarmen: Will update or just use OZ implementation.

certora/scripts/Round1/verifyAll.sh
teryanarmen: scripts ran by the ci downloads one version of solc and so uses `solc` executable. Some members on the team have a set of executables to make switching between versions easier. I can remove this script as its not ran by ci or leave it in case others from Certora, OZ, or the community try to add rules or run them.

.github/workflows/verify.yml
teryanarmen: version 0.8.4 is specified in the github link

.github/workflows/verify.yml
teryanarmen: Maybe master branch push and PRs? This way direct pushes to master are checked along with all PRs.

certora/harnesses/AccessControlHarness.sol
teryanarmen: munged contracts are created on the fly with a using a `make` file. commands are explained in the file. Im not sure you mean by `contracts-exposed`. We can have a fixed folder of the changed contracts but that would make it difficult to update when changes are made to the original contracts.

.github/workflows/verify.yml
Amxx: Sorry my message was not clear (it was mostly an internal post-it of what we have to discuss). The real question in the line of:
- do we want to have a solc version hardcoded ? 
- was processes do we put in place to maintain that ?
- does this version have to match the scripts ?

Amxx:
There are some merge conflict that I can't solve since I don't have write access to Certora:master

PR #3482, opened 2022-06-16
Simplified SignatureChecker

k06a:
Removed redundant condition


Amxx:
Hello @k06a 

If there are any error (such as an invalid signature length) the `ECDSA.tryRecover(hash, signature)` function will return
```
return (address(0), RecoverError.InvalidSignatureLength)
```

With your changes, this would be recognized as a valid `address(0)` signature ... which it is not.

k06a:
@Amxx then I would suggest making test for this case :)

PR #3483, opened 2022-06-16
ERC721 transfer should not emit Approval event

AnAllergyToAnalogy:
As per the ERC721 standard, the Approval event should not be emitted on token transfer. It is implicit in the emitted Transfer event.

```
    /// @dev This emits when the approved address for an NFT is changed or
    ///  reaffirmed. The zero address indicates there is no approved address.
    ///  When a Transfer event emits, this also indicates that the approved
    ///  address for that NFT (if any) is reset to none.
    event Approval(address indexed _owner, address indexed _approved, uint256 indexed _tokenId);
 ```

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -338,9 +338,10 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         _beforeTokenTransfer(from, to, tokenId);
 
         // Clear approvals from the previous owner
-        _approve(address(0), tokenId);
+        delete _tokenApprovals[tokenId];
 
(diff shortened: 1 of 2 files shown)


k06a:
I think this discussion could be also related to the topic: https://github.com/makerdao/dss/issues/76

AnAllergyToAnalogy:
Those discussions refer to ERC20, whereas this refers to ERC721. The [ERC721 Standard](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md) is unambiguous about this, the Approval event shouldn't be emitting on a transfer.

Amxx:
You oppened this PR twice.

Please lets only keep one instance opened #3483

PR #3493, opened 2022-06-20
Support memory arrays in MerkleTree multiproof

### LINKED ISSUE #3492: MerkleProof.processMultiProof should be callable with parameters stored in memory
**🧐 Motivation**
`MerkleProof.processMultiProof` only accepts parameters stored in `calldata`.
Leaves are often constructed by the calling code (in `memory`), and sometimes even the proof.
Functions need to be created for each use case

**📝 Details**
#### processMultiProof
```solidity
function processMultiProof(
        bytes32[] calldata proof,
        bool[] calldata proofFlags,
        bytes32[] calldata leaves
) ...
```

## Proposed
In line with the approach taken to solve this problem by `MerkleProof.processProof` & `MerkleProof.processProofCalldata` I suggest that there be three variants of the function.

#### processMultiProofCalldata
```solidity
function processMultiProofCalldata(
        bytes32[] calldata proof,
        bool[] calldata proofFlags,
        bytes32[] calldata leaves
) ...
```
Most gas efficient, least flexible
Simply a rename `MerkleProof.processMultiProof` -> `MerkleProof.processMultiProofCalldata`

#### processMultiProofCalldataProof
```solidity
function processMultiProofCalldata(
        bytes32[] calldata proof,
        bool[] calldata proofFlags,
        bytes32[] memory leaves
) ...
```
This function has the correct gas/flexibility tradeoff for many applications
_Note. Probably not the best name_

#### processMultiProof
```solidity
function processMultiProof(
        bytes32[] memory proof,
        bool[] memory proofFlags,
        bytes32[] memory leaves
) ...
```
Most flexible, but least gas efficient

Amxx:
Fixes #3492

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/mocks/MerkleProofWrapper.sol
+++ b/contracts/mocks/MerkleProofWrapper.sol
@@ -30,19 +30,36 @@ contract MerkleProofWrapper {
     }
 
     function multiProofVerify(
-        bytes32[] calldata proofs,
-        bool[] calldata proofFlag,
+        bytes32[] memory proofs,
+        bool[] memory proofFlag,
         bytes32 root,
-        bytes32[] calldata leaves
+        bytes32[] memory leaves
     ) public pure returns (bool) {
         return MerkleProof.multiProofVerify(proofs, proofFlag, root, leaves);
     }
 
-    function processMultiProof(
+    function multiProofVerifyCalldata(
         bytes32[] calldata proofs,
         bool[] calldata proofFlag,
-        bytes32[] calldata leaves
+        bytes32 root,
+        bytes32[] memory leaves
+    ) public pure returns (bool) {
+        return MerkleProof.multiProofVerifyCalldata(proofs, proofFlag, root, leaves);
+    }
+
+    function processMultiProof(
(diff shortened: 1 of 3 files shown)


frangio:
Why do we think the leaves are likely to be in memory?

Amxx:
Most of the time, the leaves are a hash produced in by the contract. For example, you want to claim 100 tokens to for user X, you hash that (sometimes with a nonce) 

[Like shown in this contract of the merkledrop workshop](https://github.com/OpenZeppelin/workshops/blob/master/06-nft-merkle-drop/contracts/ERC721MerkleDrop.sol).

Now if you want to claim multiple allocations at once, you would get an array of descriptions (number of tokens, recipient, ...), you would hash each element of the array, and that would be your leaves that need to be verified. It would be an array in memory.
The array in calldata would be the preimages.

You could imagine having the preimage AND the leaves in calldata, hashing all the preimages, and then checking that the hashes match the calldata array, but that doesn't sound right to me.

PR #3497, opened 2022-06-22
ERC4626: Add donation test case

jaglinux:
Signed-off-by: Jagadish Krishnamoorthy <jagdish.krishna@gmail.com>

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/test/token/ERC20/extensions/ERC4626.test.js
+++ b/test/token/ERC20/extensions/ERC4626.test.js
@@ -609,4 +609,30 @@ contract('ERC4626', function (accounts) {
       expect(await this.vault.totalAssets()).to.be.bignumber.equal('0');
     }
   });
+
+  //Inspired by https://forum.openzeppelin.com/t/erc4626-vault-implementation-for-totalassets-in-base-contract/29474
+  describe('deposit, donation and redeem', function () {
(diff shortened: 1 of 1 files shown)


jaglinux:
Inspired by discussion at https://forum.openzeppelin.com/t/erc4626-vault-implementation-for-totalassets-in-base-contract/29474
If there is interest, planning to add mock contract which tracks deposits and withdrawals. In case of donation, withdrawal by users will not include donated amount.
cc @Amxx

Amxx:
Hello @jaglinux 

This is already tested as part of the `multiple mint, deposit, redeem & withdrawal` test:
- alice mint share
- bob mints share
- minting token into the vault (minting directly into the vault is equivalent to the donation
→ at that point we check balances, and their equivalent assets amounts
... then we do more stuff, including withdrawals

jaglinux:
Yes @Amxx , "multiple mint, deposit, redeem & withdrawal" test covers it.
Before I close this PR, how about my other comment "If there is interest, planning to add mock contract which tracks deposits and withdrawals. In case of donation, withdrawal by users will not include donated amount."

PR #3504, opened 2022-06-24
Make ERC4626 _deposit and _withdraw internal virtual

### LINKED ISSUE #3495: Support taxed tokens transfers in the ERC4626 contract
**Scope**
ERC4626

**Motivation**
Make the standard more inclusive and usable across the hugely-diversified DeFi ecosystem by supporting taxed tokens transfers.

**Description**
Implement a token transfer library that takes into account tokens with transfer taxes or non-standard behaviours (example: a user calls `deposit(...amount = 10)` but the actual deposit is 9 _TKN_ because of a 10% transfer tax.

**Example**
```solidity
function transferTokens(
        IERC20 token,
        address from,
        address to,
        uint256 amount
    ) internal returns (uint256 originalBalance, uint256 received) {
        if (token.balanceOf(from) < amount) revert TransferHelper__Insufficient_Token_Balance(from, address(token));

        if (token.allowance(from, address(this)) < amount)
            revert TransferHelper__Insufficient_Token_Allowance(from, address(this), address(token));

        // computes transferred balance for tokens with tax on transfers
        originalBalance = token.balanceOf(to);
        token.safeTransferFrom(from, to, amount);

        received = token.balanceOf(to) - originalBalance;
    }
```
    
Now instead of `SafeERC20.transferFrom(...)` one could use the function above using the `received` value instead of the user-inputted `amount`.

Source: [TransferHelper.sol](https://github.com/Ithil-protocol/v1-core/blob/master/contracts/libraries/TransferHelper.sol)

Amxx:
Fixes #3495

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/extensions/ERC4626.sol
+++ b/contracts/token/ERC20/extensions/ERC4626.sol
@@ -175,7 +175,7 @@ abstract contract ERC4626 is ERC20, IERC4626 {
         address receiver,
         uint256 assets,
         uint256 shares
-    ) private {
+    ) internal virtual {
         // If _asset is ERC777, `transferFrom` can trigger a reenterancy BEFORE the transfer happens through the
         // `tokensToSend` hook. On the other hand, the `tokenReceived` hook, that is triggered after the transfer,
         // calls the vault, which is assumed not malicious.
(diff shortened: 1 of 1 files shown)


Amxx:
If we put that in 4.7, we don't need a changelog entry

frangio (approved):

PR #3505, opened 2022-06-24
Use @inheritdoc instead of See

### LINKED ISSUE #3502: [natspec] Use @inheritdoc instead of "See {***}"
When compiling a contract, the devdoc doesn't extends from the parent.
For example when compiling the ERC20 contract, the devdoc of method `balanceOf` is 
```
See {IERC20-balanceOf}.
```
Instead of 
```
Returns the amount of tokens owned by `account`.
```

**💻 Environment**
Hardhat

**📝 Details**
```
npx hardhat compile
```
Check the build-info with the compile output : `JSON.parse(output.contracts[sourceName][contractName].metadata).output.devdoc`.
The methods don't inherit the parent class.

**🛠️ Solution**
Use `@inheritdoc <contractName>` instead of `See {<contractName>-<method>}`.
Here are two example from ERC20, with full inherited documentation & extended documentation
```solidity
  /**
   * @inheritdoc IERC20
   */
   function balanceOf(...) ... {...)
   
  /**
   * NOTE: If `amount` is the maximum `uint256`, the allowance is not updated on
   * `transferFrom`. This is semantically equivalent to an infinite approval.
   *
   * Requirements:
   *
   * @inheritdoc IERC20
   */
    function approve(...) ... { ... }
```
 
 I can make a draft PR to discuss the implementation if you want.
 Thanks 🙂

GrandSchtroumpf:
Fixes #3502

Remove comments of type `@dev See {CONTRACT-METHOD}` that do not add more context.
Solc will inherit the comment by default.

#### PR Checklist
- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -71,9 +71,6 @@ abstract contract AccessControl is Context, IAccessControl, ERC165 {
         _;
     }
 
-    /**
-     * @dev See {IERC165-supportsInterface}.
-     */
     function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
         return interfaceId == type(IAccessControl).interfaceId || super.supportsInterface(interfaceId);
     }
--- a/contracts/access/AccessControlCrossChain.sol
+++ b/contracts/access/AccessControlCrossChain.sol
@@ -25,9 +25,6 @@ import "../crosschain/CrossChainEnabled.sol";
 abstract contract AccessControlCrossChain is AccessControl, CrossChainEnabled {
     bytes32 public constant CROSSCHAIN_ALIAS = keccak256("CROSSCHAIN_ALIAS");
 
-    /**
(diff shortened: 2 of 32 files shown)


frangio:
Thank you @GrandSchtroumpf. I'm noticing that the API docs are no longer present in the deploy preview that was generated for this PR and I'm not sure what's causing it yet.

frangio:
Following up, I still don't understand why the API docs are no longer present so we can't merge this. I don't have a lot of time to look into this for the moment. Sorry! For the record, here are the build logs though I couldn't find anything to indicate the cause: https://app.netlify.com/sites/openzeppelin-contracts-docs/deploys/62cd77f4ab4af80009167c4b

ernestognw:
This issue has evolved and there are too many conflicts. Closing, afaik `@inheritdoc` works now so a newer alternative to this PR would be better if anyone is interested

PR #3508, opened 2022-06-27
Document Address.isContract for zero-byte contracts

### LINKED ISSUE #3507: Address.isContract returns false for zero size bytecode contracts
Even though a zero size bytecode contract doesn't behave differently from an EOA, it's technically still a contract, so `isContract` should probably return `true`.

**📝 Details**

I've been trying to make a quick fix and change `account.code.length > 0` to `account.codehash != 0`, but this change breaks 66 different tests, so there will be quite a lot of cleaning up. This is also a hint that this change is a breaking one and should be published no sooner than with a major version of OZ.

**🔢 Code to reproduce bug**

CodeSandwich:
Fixes #3507 (kind of)

Adds an explicitly warning for `Address.isContract` about returning `false` for zero-byte contracts.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -22,6 +22,7 @@ library Address {
      *  - a contract in construction
      *  - an address where a contract will be created
      *  - an address where a contract lived, but was destroyed
+     *  - an address of a contract with zero-length bytecode
      * ====
      *
      * [IMPORTANT]


Amxx:
Are "zero-byte contracts" a thing ???

CodeSandwich:
They're mentioned multiple times in [EIP-1052](https://eips.ethereum.org/EIPS/eip-1052) as a valid edge case for account analysis, so maybe?

Amxx:
I see references to accounts without code, but not to contracts without code.

PR #3510, opened 2022-06-27
Minor wording fixes `ERC4626` contract

pcaversaccio:
When I was reading through the `ERC4626` contract I realised that it should be conversion (instead of convertion) and also some periods were missing for certain NatSpec comments (in order to keep consistency with the other contract comments). I assume this change does not involve any changelog entry.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx (approved):
Thank you for this PR

Issue #3514, opened 2022-06-29
Modify `ReentrancyGuard.sol` to reduce the contract size by `0.028 KB` per call


abhi3700:
**🧐 Motivation**
In my codebase, I was trying to reduce the contract size.

**📝 Details**
The inheritance of `ReentrancyGuard.sol` inside a contract is leading to reduction in contract size of `0.028 KB` per call of `nonReentrant` modifier.

The verification is shown in the image below:
<img width="447" alt="image" src="https://user-images.githubusercontent.com/16472948/176391113-876f9785-db97-4655-a994-b4a318cb1211.png">

PR #3515, opened 2022-06-29
Modify ReentrancyGuard to reduce contract size

abhi3700:
Refer issue #3514 

I wrapped the code present in function `modifier` with 2 functions (private) outside & just call the function(s) inside the modifier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+ * `ReentrancyGuard`: Reduce code size impact of the modifier by using internal functions. ([#3515](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3515))
+
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


Amxx:
Hello @abhi3700 and thank you for this PR.

I think we would want better naming. Maybe something like `_nonReentrantBefore`, `_nonReentrantAfter` ?

abhi3700:
Yeah great!

frangio (changes requested):
@abhi3700 Can you implement the changes suggested by @Amxx?

frangio:
Did you measure the additional runtime cost?

abhi3700:
> @abhi3700 Can you implement the changes suggested by @Amxx?

Done! ✅

abhi3700:
> Did you measure the additional runtime cost?

In my repo, 
the compile result with standard "ReentrancyGuard" is `9.72s`
image

And the compile result with standard "ReentrancyGuard" is `9.29s`
image

The compile time differs because of the machine running concurrent applications at this time.

So, it can be seen that the contract size is highly reduced by `0.308 KB`

frangio (approved):
Thanks!

PR #3517, opened 2022-06-29
Correcting comments, forwards all but 1/64th gas

rodrigoherrerai:
Fixes #???? 

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Thanks for the suggestion. I understand that the comment may be somewhat misleading, but the truth is that our code forwards all available gas. The fact that the EVM caps that at 63/64ths does not IMO deserve a mention here.

PR #3518, opened 2022-06-30
Fixing comment typo for ERC4626

micahriggan:
I believe this was the wrong ERC version in the comment


Amxx (approved):
Thank you for this PR!

PR #3520, opened 2022-06-30
Add Array.sort

### LINKED ISSUE #3490: Add sorting function for memory arrays
**🧐 Motivation**
On-chain sorting is a risky design choice, but for small data sizes it could be useful for some algorithms. I'm proposing sorting of `memory` arrays, NOT `storage` arrays, that would be prohibitively expensive.

**📝 Details**
This issue is basically a temperature check. The implementation of an in-memory binary heap [already exists here](https://github.com/radicle-dev/drips-contracts/blob/igor/heapsort/src/Drips.sol#L679), it's polished, optimized, benchmarked and tested. It only needs to be simplified and generalized for sorting `uint256`s instead of being a single-use heap for specialized user-defined types. If this issue gets a positive reception, I can do that work.

CodeSandwich:
Fixes #3490

Add sorting function for memory arrays

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/mocks/ArraysImpl.sol
+++ b/contracts/mocks/ArraysImpl.sol
@@ -17,10 +17,8 @@ contract ArraysImpl {
         return _array.findUpperBound(element);
     }
 
-    function sort(uint256[] memory array) external view returns (uint256[] memory sorted, uint256 gasUsed) {
-        gasUsed = gasleft();
+    function sort(uint256[] memory array) external pure returns (uint256[] memory sorted) {
         array.sort();
-        gasUsed -= gasleft();
-        return (array, gasUsed);
+        return (array);
     }
 }
--- a/contracts/mocks/ArraysImpl.sol
+++ b/contracts/mocks/ArraysImpl.sol
@@ -16,4 +16,11 @@ contract ArraysImpl {
     function findUpperBound(uint256 element) external view returns (uint256) {
         return _array.findUpperBound(element);
     }
+
+    function sort(uint256[] memory array) external view returns (uint256[] memory sorted, uint256 gasUsed) {
+        gasUsed = gasleft();
+        array.sort();
+        gasUsed -= gasleft();
+        return (array, gasUsed);
+    }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
  * `Address`: optimize `functionCall` by calling `functionCallWithValue` directly. ([#3468](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3468))
  * `Address`: optimize `functionCall` functions by checking contract size only if there is no returned data. ([#3469](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3469))
  * `GovernorCompatibilityBravo`: remove unused `using` statements ([#3506](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3506))
+ * `Arrays`: add `sort` function
 
 ## 4.7.0 (2022-06-29)
 
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -45,4 +45,66 @@ library Arrays {
             return low;
         }
     }
+
+    /**
+     * @dev Sorts `array` of integers in an ascending order.
+     *
+     * Sorting is done in-place using the heap sort algorithm.
+     * Examples of gas cost with optimizer enabled for 200 runs:
+     * - 10 random items: ~11K gas
+     * - 100 random items: ~224K gas
+     */
+    function sort(uint256[] memory array) internal pure {
+        unchecked {
+            uint256 length = array.length;
+            if (length < 2) return;
+            // Heapify the array
+            for (uint256 i = length / 2; i-- > 0; ) {
+                _siftDown(array, length, i, array[i]);
+            }
+            // Drain all elements from highest to lowest and put them at the end of the array
+            while (--length != 0) {
+                uint256 val = array[0];
+                _siftDown(array, length, 0, array[length]);
+                array[length] = val;
+            }
+        }
+    }
+
+    /**
+     * @dev Insert a `inserted` value into an empty space in a binary heap.
+     * Makes sure that the space and all items below it still form a valid heap.
+     * Index `empty` is considered empty and will be overwritten.
+     */
+    function _siftDown(
+        uint256[] memory array,
+        uint256 length,
+        uint256 empty,
+        uint256 inserted
+    ) private pure {
+        unchecked {
+            while (true) {
+                // The first child of empty, one level deeper in the heap
+                uint256 child = empty * 2 + 1;
+                // Empty has no children
+                if (child >= length) break;
+                uint256 childVal = array[child];
+                uint256 otherChild = child + 1;
+                // Pick the larger child
+                if (otherChild < length) {
+                    uint256 otherChildVal = array[otherChild];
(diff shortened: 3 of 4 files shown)


CodeSandwich:
The second commit cuts gas by ~1/3 by switching memory access from Solidity to Yul, there's no way around index checking overhead in pure Solidity.

The memory location is calculated in Yul using the formula: `add(32, add(array, shl(5, idx)))`. It seems to be the sweet spot when it comes to gas. What I don't understand is why even slight changes to it cause a sharp gas increase, in the ballpark of 10% for the whole sorting algorithm. Formula like `add(array, add(32, shl(5, idx)))` should yield almost identical opcodes, but it somehow manages to be MUCH more expensive. Does anybody know why it is like this? I've tried compiling that snippet in the vacuum and looking at the opcodes, but they both look similarly efficient.

contracts/mocks/ArraysImpl.sol
```diff
@@ -16,4 +16,11 @@ contract ArraysImpl {
     function findUpperBound(uint256 element) external view returns (uint256) {
         return _array.findUpperBound(element);
     }
+
+    function sort(uint256[] memory array) external view returns (uint256[] memory sorted, uint256 gasUsed) {
+        gasUsed = gasleft();
```
Amxx, line 21: Please don't use that gas estimation trick in the repo. That is ok for benchmarking in remix, but not here.

Amxx:
Indeed, I was considering adding a function that skips the length check when reading memory (or storage) arrays.

For the gas optimization part, I believe that the arguments should be consumed as early as possible to avoid reordering the stack. This means using them in the most nested operation (that is going to be performed first).

Have you tried `add(32, add(shl(5, idx), array))`, and if yes, was it better/worst?

contracts/mocks/ArraysImpl.sol
Amxx: I would recommend doing a transaction (we can force doing a transaction to a view function), and check the gas usage. Of the entire function (the offset should always be the same). Using the `REPORT=true` will also show up some values.

Amxx: If there is better tooling that works with hardhat, we would love to try it out.

CodeSandwich:
> Have you tried add(32, add(shl(5, idx), array)), and if yes, was it better/worst?

I just did, the results are marginally worse, around 2.5% for an entire sort of 100 elements. The shalowness on stack would make a lot of sense, especially when inlined.

contracts/mocks/ArraysImpl.sol
CodeSandwich: I've enforced sending transactions in tests and here are the result of `enableGasReport=true REPORT=true npm test`:
```
|  Contract    ·  Method  ·  Min        ·  Max        ·  Avg        ·  # calls      ·  usd (avg)  │
···············|··········|·············|·············|·············|···············|··············
|  ArraysImpl  ·  sort    ·      22226  ·     216532  ·      75205  ·           13  ·          -  │
···············|··········|·············|·············|·············|···············|··············
```
It's not very useful, it still creates vague statistics from all the test runs. It also adds all kinds of overheads from calldata handling. For the largest case it's 52K gas or 31% and it's probably rather consistent, but for the cheapest case it's 22K gas or a 322-fold increase, which makes any benchmarks obscured by an elephant. If we move the test data from calldata to storage, it'll be even more dramatic.

What's wrong with in-EVM gas measurements for this particular function? I understand that things like storage warmness or memory allocations have hard to predict costs, but here we aren't doing any such things.

frangio:
Using unchecked arithmetic and array accesses on @Amxx's quicksort implementation seems to be more efficient than the heapsort implementation in this PR, at least in the array I'm testing with [here](https://github.com/frangio/gas-bench/tree/e76022b9fa8e13290b2200014d7bc58a6b7d0263). More testing and benchmarking is needed to conclude anything. We should look at the specific case of sorting an already sorted array, which is known to be bad for quicksort with the pivot selection we're using (always the leftmost item). We could consider selecting the middle element in the array to mitigate for that case.

After a deep dive into search algorithms and the implementations in https://g.solidity.cc/challenges/sort, I've started to think that our goal shouldn't be to maximize efficiency. It's hard to optimize for all cases, and in any case sorting in a smart contract should be discouraged. Instead, I'm thinking we should aim for simple auditable code that is reasonably efficient. In this regard, quicksort is quite attractive because of its simplicity.

CodeSandwich:
@frangio Thank you for benchmarking it! What are the exact results?

Auditability is a good reason to choose quicksort. OTOH heapsort has a better worst case complexity of `O(N log N)` while quicksort has `O(n^2)`. No matter how we tune it, the worst case scenario will probably be arrangeable, which may lead to attacks and gas usage surprises. Heapsort is safer in this regard and more predictable, which are nice properties if we're looking for a general-purpose algorithm which doesn't require diving deep into the problem.

frangio:
Away from the computer right now so I can't share the exact results.

Definitely agree that the better worst case complexity of heap sort is a factor to consider. That said, I still wouldn't be comfortable with n log n either, it could still lead to vulnerabilities if n can be controlled by an attacker. On chain sorting should be used sparingly and with arrays known to be small, or in a path where the impact of a large array is contained.

contracts/mocks/ArraysImpl.sol
Amxx: This enableGasReport is indeed very rought, and its mostly used to check the delta when modifying code.
For new code, and comparing designs,  @frangio is working on another worklow.

frangio:
What is needed to proceed here is to gain some more confidence that Quicksort would generally be more efficient, by running some more [benchmarks](https://github.com/frangio/gas-bench/tree/e76022b9fa8e13290b2200014d7bc58a6b7d0263):
- An already sorted array, or one that is neraly sorted
- Arrays of different sizes (what is the largest that can be sorted under the gas limit?)

Note that my gas-bench repo is benchmarking sort of an array that is defined in Solidity and this isn't very usable for running multiple tests with different arrays so it would really help to change it to load the array from calldata.

We also need to try different methods for selecting the pivot. Currently using the first in the array, this will be bad for sorted or near-sorted arrays. We can try using the item at the middle of the array. There are better methods such as the median of three where you take the median of the first, middle, and last items; this should be tried but may add too much overhead for small arrays which is our main focus here.

There is also the possibiliy to consider other partitioning algorithms, I've seen discussion about [Hoare vs Lomuto partitioning](https://www.geeksforgeeks.org/hoares-vs-lomuto-partition-scheme-quicksort/) that we should look into. One is said to be more efficient than the other, and I don't think we're using it. I don't know if the differences are relevant for the specifics of the EVM but we should try.

Proposals for other algorithms are also welcome, considering auditability as a main factor to optimize, with reasonable efficiency that isn't too far behind the most optimized implementations.

contracts/mocks/ArraysImpl.sol
CodeSandwich: Removed, if anybody needs them back, they can revert the removal commit of just copy the code like frangio did.

burke-md:
> so it would really help to change it to load the array from calldata

@frangio Would you be open to PRs for your gas-bench repo?
Something along the lines of setting up unit tests to run the bench marking w/ various inputs (and having the function load from call data as you suggest from above)

frangio:
@burke-md Yes that sounds good!

Amxx:
This is based on an old commit. Replaced by #4846.

Issue #3512, opened 2022-06-28
Why not wrap the ` _balances[to] += amount` instruction in an `unchecked` block?


paulrberg:
This issue is related to https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2875, but my question comes from a different angle:

Why not wrap *all* calculations in the `_transfer` function in an `unchecked` block? I'm specifically referring to line 241 in the `_transfer` function:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e734b42fc2245b520372bca0099870f40f1e6f38/contracts/token/ERC20/ERC20.sol#L241

It seems to me that the balance of the `to` account can never overflow (it's an invariant), due to the fact that the `_mint` function uses checked arithmetic (and rightly so).

I looked at the tests for the `transfer` function and couldn't find a test that checks for an overflow on line 241:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e734b42fc2245b520372bca0099870f40f1e6f38/test/token/ERC20/ERC20.behavior.js

The only argument that I can think of in favor of keeping the current behavior is the precautionary principle: many people inherit `ERC20` and modify it according to their project idea, which in some cases might involve changing the `_balances` mapping outside of `_mint`.

Thoughts?

Amxx:
Indeed, the `_balances[to] += amount;`  cannot overflow, because we already ensure that the total supply doesn't overflow.
Also, the balance increase on line 263 could be unchecked. 

I think we haven't done that previously because we were not fully comfortable with using unchecked with a low granularity, but this is an obvious gas improvement that we should implement!

Would you like to open a PR for that?

paulrberg:
> Also, the balance increase on line 263 could be unchecked.

Good catch!

> Would you like to open a PR for that?

Thanks for letting me do it. Here it is!

https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3513

frangio:
So for the record I want to explain why we did not have this optimization before and why we're going to merge it now.

The reasoning until now was that we use `unchecked` when lack of overflow can be proven through _local_ reasoning. For example:

```solidity
require(amount <= balances[to]);
unchecked {
    balances[to] -= amount;
}
```

We know `b - a` will not overflow because `a <= b` is asserted immediately before.

[...]

PR #3524, opened 2022-07-01
Use unchecked for ERC721 balance updates

Amxx:
Following the logic in described in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3512#issuecomment-1172614573.

- `_balances` are private
- `_balances` are only increased/decreased when tokens are minted, burned, or transferred

A balance cannot decrease below 0 because that would require transferring/burning more tokens out of a wallet than were transferred/minted to this wallet in the first place.

The balance cannot overflow, because that would require minting all the tokenIds, and putting them all in the same wallet (there are 2**256 tokenIds, so that would overflow by one). While this is mathematically possible, it is impossible in practice (with token being transferred/minted one at a time)

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
Given that we want to explore batch minting in ERC721 it seems we should not merge this, as the assumption that minting happens one at a time will be invalidated. Also please include benchmarks, what kind of improvement do you see with this PR?

Amxx:
I understand that we want to experiment new minting mechanism. However, I believe that these minting mechanisms should remain ERC721 compliant, which means emitting a Transfer event for all the tokens ids. This means that, in practice, it's impossible to mint all the tokens, and the balance increment overflow is not an issue.

Amxx:
For reference: https://twitter.com/optimizoor/status/1544626818996613120?t=8tV84gD3blTSsxk14EotiQ

TLDR: batch minting should be limited to "not to big batches" ... which means that the limit should never be reached, even minting 10**18 token each time

contracts/token/ERC721/ERC721.sol
```diff
@@ -340,8 +350,15 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
...
 
-        _balances[from] -= 1;
-        _balances[to] += 1;
+        unchecked {
+            // `_balances[from]` cannot overflow for the same reason as described in `_burn`:
+            // `from`'s balance is the number of token held, which is at least one before the current
+            // transfer.
+            // `_balances[to]` could overflow in the conditions described in `_mint`. That would require
+            // all 2**256 token ids to be minted, which in practice is impossible.
+            _balances[from] -= 1;
```
frangio, line 359: `_beforeTokenTransfer` could in theory move the token away and this would result in overflow if we use unchecked.

```suggestion
        _balances[from] -= 1;
        unchecked {
            // `_balances[to]` could overflow in the conditions described in `_mint`. That would require
            // all 2**256 token ids to be minted, which in practice is impossible.
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -308,7 +314,11 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         // Clear approvals
         _approve(address(0), tokenId);
 
-        _balances[owner] -= 1;
+        unchecked {
+            // Cannot overflow, as that would require more tokens to be burned/transfered
+            // out then the owner initialy received through minting and transfering in.
+            _balances[owner] -= 1;
+        }
```
frangio, line 321: For the same reason, because of `_beforeTokenTransfer` we can't use unchecked here.

```suggestion
        _balances[owner] -= 1;
```

frangio (changes requested):

contracts/token/ERC721/ERC721.sol
Amxx: If `_beforeTokenTransfer` transfers the token, then the balances would be invalid regardless of the `unchecked`. This is a wider issue, that we should address independently, and that IMO should not prevent the use of unchecked here.

Amxx:
I believe once #3611 is merged, the requested changes will no longer be needed.

frangio (approved):
LGTM

PR #3527, opened 2022-07-05
Move adds on total earlier to enable the use of unchecked

songlh:
For two consecutive additions, we can put the addition on the total variable before the addition on the individual variable. After that, we can use unchecked for the second addition, because if the addition on total does not overflow, then the addition on individual variable does not overflow either. 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/finance/PaymentSplitter.sol
+++ b/contracts/finance/PaymentSplitter.sol
@@ -149,6 +149,8 @@ contract PaymentSplitter is Context {
 
         require(payment != 0, "PaymentSplitter: account is not due payment");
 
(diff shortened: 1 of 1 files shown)


frangio (changes requested):
Thanks, this is a good idea. Can you add comments inline explaining why unchecked is safe?

songlh:
@frangio I just commit some comments.

frangio (approved):

Amxx (approved):
Looks good!

PR #3528, opened 2022-07-06
update: visibility of the delegate function

Rassska:
Since it's not used inside of the contract, it's better to mark it as external

Fixes #???? 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


Amxx:
Hello @Rassska 

This has already been discussed. If we make it external it will no longer be possible to call "super" when overriding it.

PR #3537, opened 2022-07-08
Fix broken forum URL

tonynoce:
The website address was going nowhere, it had to go to the openzeppelin forum. Also the title was reversed.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
The "[How" is the AsciiDoc syntax necessary for the site... I don't know if it will work if we put it in another line. This is meant to work in the site mainly, not when looking at the raw code.

frangio (approved):

PR #3539, opened 2022-07-10
erc721: add uncheck for _balances decrement operation

jaglinux:
#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


jaglinux:
_balances can never underflow since burn or transfer will fail if not minted.
We can also use unchecked for _balances increment, but in theory a user can have tokens more than uint256 max ?

PR #3540, opened 2022-07-11
Use calldata to reduce gas costs in ERC1155

cygaar:
We can use `calldata` instead of `memory` in several places in ERC1155 (as well as the mocks/extensions). This will both save gas and also keep the implementation function signatures the same as the interface function signatures (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/IERC1155.sol). The EIP itself has the input parameters as calldata: https://eips.ethereum.org/EIPS/eip-1155.

I ran the full set of ERC1155 tests (`env ENABLE_GAS_REPORT=true npx hardhat test --grep ERC1155`) and got these gas savings: https://www.diffchecker.com/0Mac8iiV.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
  * `ERC20`: optimize `_transfer`, `_mint` and `_burn` by using `unchecked` arithmetic when possible. ([#3513](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3513))
  * `ERC721`: optimize transfers by making approval clearing implicit instead of emitting an event. ([#3481](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3481))
  * `ERC721`: optimize burn by making approval clearing implicit instead of emitting an event. ([#3538](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3538))
+ * `ERC1155`: optimize mint, transfer, and burn by using calldata. ([#3540](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3540))
 
 ### Compatibility Note
 
--- a/contracts/mocks/ERC1155BurnableMock.sol
+++ b/contracts/mocks/ERC1155BurnableMock.sol
@@ -11,7 +11,7 @@ contract ERC1155BurnableMock is ERC1155Burnable {
         address to,
         uint256 id,
         uint256 value,
-        bytes memory data
+        bytes calldata data
     ) public {
         _mint(to, id, value, data);
     }
--- a/contracts/mocks/ERC1155Mock.sol
+++ b/contracts/mocks/ERC1155Mock.sol
@@ -11,7 +11,7 @@ import "../token/ERC1155/ERC1155.sol";
 contract ERC1155Mock is ERC1155 {
     constructor(string memory uri) ERC1155(uri) {}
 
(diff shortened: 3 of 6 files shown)


Amxx:
Thank you @cygaar for the PR and the gas saving analysis. Note that since last week we have a workflow that generates these reports automatically, just check the test summary for your commit, and its all there

cygaar:
@Amxx are you referring to one of the github workflows?

cygaar:
Super convenient, looks like the gas went down for all the 1155 tests. For some reason ERC20VotesMock gas went up, but I'm guessing this might be some sort of transient result since I didn't touch anything outside of 1155.

Amxx:
The gas usage is good.

The thing is that is also limit some usages, such as an additional function (added by a dev) calling `batchBalanceOf`. Basically this change might be breaking some usecases.

We'll discuss that internally

cygaar:
I hear you, but my arguments for adding in this change are:
1) I don't think anything should be breaking per se - calldata is allowed in internal function calls. At worst it'll increase gas costs if you call any of these functions internally. If you think about how often someone will be implementing custom functionality vs the default usecase of this contract, the savings seem worth it.
2) The EIP itself has these arguments as `calldata`. Devs should come in expecting these functions to already be calldata because at the end of the day, this library is implementing EIP1155.

Amxx:
About point 2:

The EIP writes calldata because the EIP describes an interface, with external functions ... and at the time external function only supported calldata. If you have a look at our repo, you'll see that [the interface](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC1155/IERC1155.sol#L63) uses calldata, just like the EIP, and that its the implementation that "overrides" that to be memory.

Now about point 1:

It is not possible for a contract to "build" an array (or a struct) in calldata. Only in memory. That means that if you have a function that wants to create a list of tokensIds (and not get it as an argument) it will be memory, and you can't simply pass it to a calldata function. You can do  `this.functionName(myArrayInMemory)` but that is more expensive and it changes the execution context.

I'm not too worried about `balanceOfBatch` because it's easy to call `balanceOf` in a loop, so creating an in memory array to then pass it to balanceOfBatch doesn't sound right anyway.

The `_safeBatchTransferFrom` is more concerning though. Not being to call it with internally built memory arrays is an issue IMO.

cygaar:
Can you elaborate more on `it changes the execution context.`?

Amxx:
> Can you elaborate more on `it changes the execution context.`?

Lets say I have this contracts:

```
contract Test {
    event SomeEvent(address caller);
    
    function something() public {
        emit SomeEvent(msg.sender);
    }
    function endpoint1() public {
        something();
    }
    function endpoint2() public {
        this.something();
    }
}
```

If you call `endpoint1`, msg.sender used when emitting the event is your address. If you call `endpoint2`, the msg.sender is the address of the contract itself.

When calling `safeTransferFrom` this distinction is important.

frangio:
I think the gas savings are too small (<1%) to justify the breaking change.

ernestognw:
This PR has gotten old and is a breaking change because it limits the current functionality. We just released 5.0 and I agree the gas savings are too small.

Closing.

PR #3542, opened 2022-07-12
Improve grammar in transfer error messages

lexmag:


frangio:
Thanks @lexmag. A bit reluctant on this one because it increases the size of the string and thus increases deployment costs. What would you say if we change "nor" to "or"?

frangio:
The tests need to be updated to match the new revert reason, is this something you can do?

lexmag:
The tests are updated. Also, fixed more occurrences of that wording.

frangio (approved):
Thank you!

Issue #3555, opened 2022-07-17
URIStorage for ERC721 with Royalty


namelessperson0:
The [ERC721URIStorage](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721URIStorage.sol) extension contract does not work for [ERC721Royalty](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Royalty.sol) contract. 

We need an URIStorage extension contract for ERC721 with Royalty.

Amxx:
Can you clarify "does not work" ?

Amxx:
I don't see any issues on my end. Here is an example contract that is ERC721URIStorage & ERC721Royalty:

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.4;

import "@openzeppelin/contracts/token/ERC721/ERC721.sol";
import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Royalty.sol";
import "@openzeppelin/contracts/token/ERC721/extensions/ERC721URIStorage.sol";
import "@openzeppelin/contracts/access/Ownable.sol";

contract MyToken is ERC721, ERC721Royalty, ERC721URIStorage, Ownable {
    constructor() ERC721("MyToken", "MTK") {}

    function safeMint(address to, uint256 tokenId, string memory uri)
        public
        onlyOwner
    {
        _safeMint(to, tokenId);
        _setTokenURI(tokenId, uri);
    }

    // The following functions are overrides required by Solidity.

    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC721, ERC721Royalty) returns (bool) {
          return super.supportsInterface(interfaceId);
      }

    function _burn(uint256 tokenId) internal override(ERC721, ERC721Royalty, ERC721URIStorage) {
        super._burn(tokenId);
    }

    function tokenURI(uint256 tokenId)
        public
        view
        override(ERC721, ERC721URIStorage)
        returns (string memory)
    {
        return super.tokenURI(tokenId);
    }
}
```

PR #3557, opened 2022-07-17
URIStorage for ERC721 with Royalty

namelessperson0:
For fixing https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3555


Amxx:
Creating a new contract for that really doesn't look like the right solution. Let's discuss the issue before jumping to PR

PR #3561, opened 2022-07-18
Keep track of historical quorum values

Amxx:
Checkpointing quorum is necessary to avoid quorum changes making old, failed because of quorum, proposals suddenly successful.

#### PR Checklist

- [x] Tests
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/extensions/GovernorVotesQuorumFraction.sol
+++ b/contracts/governance/extensions/GovernorVotesQuorumFraction.sol
@@ -97,7 +97,7 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
         // make sure we keep track of the old numerator (applicable to upgrade only)
         if (_quorumNumeratorHistory._checkpoints.length == 0) {
             _quorumNumeratorHistory._checkpoints.push(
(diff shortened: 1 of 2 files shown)


frangio (dismissed):

frangio:
Waiting until next week to discuss an alternative fix.

frangio:
The issue with this PR is that it increases the cost of voting when the PreventLateQuorum module is used. It seems like this can be fixed by first checking if the latest checkpoint is the one we're looking for (in `quorumNumerator(uint blockNumber)`) and in that case avoid the binary search.

Amxx:
I've implemented an optimistic lookup, that skips the binary search if the checkpoint we are looking for is the last one. However that only concerns `quorumNumerator(uint256)`. This function is mostly called through 

```
function quorum(uint256 blockNumber) public view virtual override returns (uint256) {
    return (token.getPastTotalSupply(blockNumber) * quorumNumerator(blockNumber)) / quorumDenominator();
}
```

The `token.getPastTotalSupply(blockNumber)` call will not be optimized the same way

Amxx:
Not, the parts not covered are mechanism used in cases of migration.

frangio (approved):

PR #3565, opened 2022-07-20
Optimize safe casting of signed numbers

CodeSandwich:
Fixes #nothing

This PR reduces gas needed to perform safe downcasting of signed integers by 23 gas ('gasleft' microbenchmark went down from 110 to 87 gas). Both underflows and overflows are still caught, but instead of two comparisons there's only one.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio (approved):
Very simple, I like it. Thanks!

Keep them coming!

PR #3567, opened 2022-07-21
Fix error in documentation and typo

ptrcarta:
Hello, just minor nitpicks. RLC encoding (RLP maybe?) doesn't seem to be a thing, and abi.encode doesn't do RLP anyway


frangio (approved):
Agree. Thanks.

PR #3573, opened 2022-07-23
Optimize toString

CodeSandwich:
Fixes #???? 

This PR reduces gas usage of `Strings.toString` for decimal numbers. Here are some results:

```
Input                                                                          | Current | New  | Gas saved
-------------------------------------------------------------------------------|---------|------|-----------------
0                                                                              | 131     | 458  | -327  (-249,62%)
1                                                                              | 833     | 458  | 375   (45,02%)
12                                                                             | 1383    | 534  | 849   (61,39%)
123                                                                            | 1933    | 611  | 1322  (68,39%)
1234                                                                           | 2483    | 687  | 1796  (72,33%)
12345                                                                          | 3033    | 735  | 2298  (75,77%)
1234567                                                                        | 4133    | 888  | 3245  (78,51%)
1234567890                                                                     | 5783    | 1059 | 4724  (81,69%)
123456789012345                                                                | 8533    | 1413 | 7120  (83,44%)
12345678901234567890                                                           | 11283   | 1708 | 9575  (84,86%)
123456789012345678901234567890                                                 | 16783   | 2357 | 14426 (85,96%)
1234567890123456789012345678901234567890                                       | 22289   | 2983 | 19306 (86,62%)
12345678901234567890123456789012345678901234567890                             | 27789   | 3574 | 24215 (87,14%)
123456789012345678901234567890123456789012345678901234567890                   | 33289   | 4252 | 29037 (87,23%)
1234567890123456789012345678901234567890123456789012345678901234567890         | 38795   | 4820 | 33975 (87,58%)
115792089237316195423570985008687907853269984665640564039457584007913129639935 | 43195   | 5345 | 37850 (87,63%)
```
(Edit: results updated for the newest commits) Each measurement was done in a separate dapptools test to remove memory usage bias. Solc was in version 0.8.13 with optimizer set to 1 million runs.

A little more gas can be pinched, e.g. manual string allocation with Yul saves ~119 gas, but that both complicates the codebase and makes the optimizer behavior harder to predict.

Tests are cleaned up and extended.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


Amxx:
Thanks @CodeSandwich for this PR

Note: We could use the same length-search pattern for the hex version.

CodeSandwich:
Yes, if this PR goes through, we can add these optimizations there too. One thing at a time :smiley:

contracts/mocks/StringsMock.sol
```diff
@@ -5,19 +5,19 @@ pragma solidity ^0.8.0;
 import "../utils/Strings.sol";
 
 contract StringsMock {
-    function fromUint256(uint256 value) public pure returns (string memory) {
+    function toStringDecimal(uint256 value) public pure returns (string memory) {
```
Amxx, line 8: why not name it "toString" ... just like the underlying library function ?

CodeSandwich: Because when calling `await strings.toString(input)` JS gets confused and instead of calling the contract it just returns a string `"[Object object]"`.

contracts/utils/Strings.sol
```diff
@@ -13,26 +13,55 @@ library Strings {
...
      * @dev Converts a `uint256` to its ASCII `string` decimal representation.
      */
-    function toString(uint256 value) internal pure returns (string memory) {
-        // Inspired by OraclizeAPI's implementation - MIT licence
-        // https://github.com/oraclize/ethereum-api/blob/b42146b063c7d6ee1358846c198246239e9360e8/oraclizeAPI_0.4.25.sol
+    function toString(uint256 value) internal pure returns (string memory result) {
+        unchecked {
+            if (value < 10) {
+                return string(abi.encodePacked(uint8(value + 48)));
+            }
```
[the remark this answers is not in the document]
CodeSandwich: We can't remove it entirely, we still need to handle the special case of `0`. That change decreases gas for the case of `0` from 170 to 131, but for other single-digit numbers increases from 170 to 525.

Why do you want to remove it, is it to clean up the logic or for a different reason? If it's only to clean up, IMO the gas savings justify the tiny additional complexity, especially since single-digit conversions seem to be a common use case, but that's just that: an opinion.

contracts/utils/Strings.sol
```diff
@@ -13,26 +13,55 @@ library Strings {
...
+                valueCopy /= 10000;
+                length += 4;
+            }
+            if (valueCopy >= 10) {
+                valueCopy /= 100;
+                length += 2;
+            }
+            if (valueCopy >= 1) {
+                length += 1;
+            }
```
Amxx, line 49:
```suggestion
uint256 length = 1;
            uint256 x = value;

            if (x >= 10**64) {
                x /= 10**64;
                length += 64;
            }
            if (x >= 10**32) {
                x /= 10**32;
                length += 32;
            }
            if (x >= 10**16) {
                x /= 10**16;
                length += 16;
            }
            if (x >= 10**8) {
                x /= 10**8;
                length += 8;
            }
            if (x >= 10**4) {
                x /= 10**4;
                length += 4;
            }
            if (x >= 10**2) {
                x /= 10**2;
                length += 2;
            }
            if (x >= 10**1) {
                length += 1;
            }
```

CodeSandwich: Very clever and no gas penalty :+1: Applied.

contracts/utils/Strings.sol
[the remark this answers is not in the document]
Amxx: we can, see my commit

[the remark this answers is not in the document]
CodeSandwich: Which commit?

contracts/mocks/StringsMock.sol
Amxx: I'm fixing that

contracts/utils/Strings.sol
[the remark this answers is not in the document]
CodeSandwich: This is the commit: https://github.com/CodeSandwich/openzeppelin-contracts/commit/8597e3dcc34e3ae5f48a0ad8a19bf3d0f504703c

The problem is that it creates bad output, in all cases except `0` it creates a string with length 1 too big. Plus now **all** single-digit numbers have cost increased 3-fold to >500, not only `1` to `9`.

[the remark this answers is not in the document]
CodeSandwich: Ok, I see that it'll work just fine, the change in powers of 10 counter the initial length of 1, very clever! But the gas usage for single-digit numbers is still much higher. What are the advantages?

contracts/utils/Strings.sol
```diff
@@ -70,16 +71,35 @@ library Strings {
...
-        uint256 length = 0;
-        while (temp != 0) {
-            length++;
-            temp >>= 8;
+        unchecked {
+            uint256 length = 1;
+
+            // compute log256(value), and add it to length
+            uint256 valueCopy = value;
+            if (valueCopy >> 128 > 0) {
```
CodeSandwich, line 79: `if (valueCopy > 1 << 128) {` will be equivalent, but should be cheaper, the shifting will be evaluated at compilation time.

Amxx: I'm not sure the shift is done at compile time. Can you confirm ?

CodeSandwich: As of 0.8.13, yes, I've changed some constants to `1 << X` and the gas usage hasn't changed at all.

contracts/utils/Strings.sol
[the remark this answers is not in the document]
CodeSandwich: Does removal of the special case reduce a lot of bytecode? Lower gas usage is always a good thing, because we don't know how this code will be used. Otherwise this entire PR is pointless, it complicates the logic in exchange for lower gas. IMO we can keep all your changes including `length = 1`, but still restore the single-digit shortcut.

contracts/utils/Strings.sol
```diff
@@ -77,23 +77,23 @@ library Strings {
...
                 valueCopy >>= 32;
                 length += 4;
             }
-            if (valueCopy >> 16 > 0) {
+            if (valueCopy > 1 << 16 ) {
                 valueCopy >>= 16;
                 length += 2;
             }
-            if (valueCopy >> 8 > 0) {
+            if (valueCopy > 1 << 8 ) {
```
CodeSandwich, line 96: I'm quite certain that it should be `>=` everywhere, let's look at the example of `0x100`, it SHOULD trigger that `if`, but it won't, because it's `== 1 << 8`

CodeSandwich:
I've pushed an optimization of the memory writing loop. I don't exactly understand why, streamlining the loop and putting the `iszero` command as the last thing of the loop just clicks with the optimizer. The new gas usage results:
```
0: 458
1: 458
12: 534
123: 611
1234: 687
12345: 735
1234567: 888
1234567890: 1059
123456789012345: 1413
12345678901234567890: 1708
123456789012345678901234567890: 2357
1234567890123456789012345678901234567890: 2983
12345678901234567890123456789012345678901234567890: 3574
123456789012345678901234567890123456789012345678901234567890: 4252
1234567890123456789012345678901234567890123456789012345678901234567890: 4820
115792089237316195423570985008687907853269984665640564039457584007913129639935: 5345 (8 fold decrease from master 😆)
```
The usage of the symbols table has zero impact on gas, but IMO it's much clearer than the addition of a magic number.

contracts/utils/Strings.sol
```diff
@@ -7,7 +7,7 @@ pragma solidity ^0.8.0;
  * @dev String operations.
  */
 library Strings {
-    bytes16 private constant _HEX_SYMBOLS = "0123456789abcdef";
+    bytes16 private constant _SYMBOLS = "0123456789abcdef";
```
Amxx, line 10: Does that change anything ?

If no, I would suggest to change as little as possible.

contracts/utils/Strings.sol
```diff
@@ -51,17 +51,18 @@ library Strings {
...
+                let ptr := add(buffer, add(32, length))
                 for {
 
-                } gt(ptr, pos) {
+                } 1 {
 
                 } {
                     ptr := sub(ptr, 1)
-                    mstore8(ptr, add(48, mod(value, 10)))
+                    mstore8(ptr, byte(mod(value, 10), _SYMBOLS))
```
Amxx, line 61: how is that compiled? Is the result cheaper than `add 48` I doubt it

CodeSandwich: It is, I've measured it. Plus it's the same thing, have `value` and 2 constants on stack, perform one `mod` and one `Gverylow`operation.

contracts/utils/Strings.sol
```diff
@@ -51,17 +51,18 @@ library Strings {
...
                     value := div(value, 10)
+                    if iszero(value) {
```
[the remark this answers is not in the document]
CodeSandwich: No, it's equivalent, because it's performed once per iteration anyway. We need to cover the `0` case, so we needed to turn a "while" loop into a "do-while". If we special-case it, we can move `iszero` to the for loop exit clause.

contracts/utils/Strings.sol
CodeSandwich: It's private, so it can't break anything. I'm proposing unifying logic for getting decimal and hex characters, so the lookup table can be used in both cases making it no longer tied to the hex logic. It's purely cosmetic though.

contracts/utils/Strings.sol
Amxx: Fair enough.

Ideally we would need tooling that checks that for us.

contracts/utils/Strings.sol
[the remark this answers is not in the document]
Amxx: I agree that the isZero must be at the end so that `toString(0)` is correctly processed.

contracts/utils/Strings.sol
frangio: >but should be cheaper

Why? I'm seeing the compiler emit code sequences that cost exactly the same for the two expressions.

`PUSH1 0x1 PUSH1 0x80 SHL DUP2 LT`
vs
`PUSH1 0x80 DUP2 SWAP1 SHR ISZERO`

frangio:
Sorry I'm late to this PR.

My thoughts are:

- For loops in Yul have very unfamiliar syntax, at the moment we don't have any in the entire codebase, and I would prefer if that remained the case. The main gains in this PR are due to an optimized log10 implementation, and probably some more from removing bounds checks on array accesses. Some assembly is necessary for the second part but I would strongly prefer if the for loop was outside of that.
- I would really appreciate if the log10 implementation was added as `Math.log10` so it could be used on its own. That also implies some more testing though, so I would be ok with doing that in a later PR.
- As I commented above, `>= 1 << 128` vs `>> 127 > 0` seems to not make a difference at all so I'd like us to revert the unrelated changes in Math.sol.

Amxx:
> I would really appreciate if the log10 implementation was added as `Math.log10` so it could be used on its own. That also implies some more testing though, so I would be ok with doing that in a later PR.

Making a PR that includes Math.log10 and Math.log2 would be great. 
- Math.log2 could be used in `Math.sqrt(uint256)` and `Strings.toHexString(uint256)`
- Math.log10 would be used in `Strings.toString(uint256)`

@CodeSandwich, would you be able to do that ?

contracts/utils/Strings.sol
```diff
@@ -7,48 +7,101 @@ pragma solidity ^0.8.0;
...
+
+                } {
+                    ptr := sub(ptr, 1)
+                    mstore8(ptr, byte(mod(value, 10), _SYMBOLS))
+                    value := div(value, 10)
+                    if iszero(value) {
+                        break
+                    }
+                }
+            }
```
frangio, line 67: If we declare `symbols` as a local variable before use, the compiler emits better code, loading the constant on the stack before the loop and then reusing that value, whereas using the state variable directly the constant is loaded on the stack on each iteration.

Using a while loop instead of an assembly for loop the gas costs are exactly the same and we get better readability. I didn't check the bytecode but I wouldn't be surprised if they're the same.

```suggestion
            uint ptr;
            assembly {
                ptr := add(buffer, add(32, length))
            }

            bytes16 symbols = _SYMBOLS;

            while (true) {
                ptr -= 1;
                assembly {
                    mstore8(ptr, byte(mod(value, 10), symbols))
                }
                value /= 10;
                if (value == 0) {
                    break;
                }
            }
```

contracts/utils/Strings.sol
```diff
@@ -59,7 +112,7 @@ library Strings {
         buffer[0] = "0";
         buffer[1] = "x";
         for (uint256 i = 2 * length + 1; i > 1; --i) {
-            buffer[i] = _HEX_SYMBOLS[value & 0xf];
+            buffer[i] = _SYMBOLS[value & 0xf];
             value >>= 4;
         }
```
frangio, line 117: This could be written in the same or similar way that the decimal `toString`.

contracts/utils/Strings.sol
Amxx: `ptr -= 1;` should be unchecked

frangio: All of the function is in an `unchecked` block!

CodeSandwich: Indeed the amount of assembly can be reduced with an exactly zero impact on gas. Good catch!

Copying of `_SYMBOLS` to a local variable increases the gas usage by over 17% for larger numbers. Why making a copy is supposed to reduce gas? Unlike managing yet another local variable on stack pushing a constant is a single opcode.

frangio: I saw it reduce gas but didn't test large numbers. In retrospect I agree with the reasoning that managing a value on the stack will overall just add overhead.

contracts/utils/Strings.sol
frangio: >did you enable the optimizer?

Yes. For constants Solidity will often prioritize code size before runtime cost.

frangio (dismissed):
LGTM

PR #3579, opened 2022-07-25
Report changes to contract construction cost in CI

Amxx:
Report changes to contract construction cost in CI

#### PR Checklist
- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/scripts/checks/compareGasReports.js
+++ b/scripts/checks/compareGasReports.js
@@ -47,14 +47,16 @@ class Report {
 
     const deployments = update.info.deployments
       .filter(contract => contract.gasData?.length)
-      .map(contract => Object.assign(contract, { previousGasData: ref.info.deployments.find(({ name }) => name === contract.name).gasData }))
-      .map(contract => ({
+      .map(contract => Object.assign(contract, { previousVersion: ref.info.deployments.find(({ name }) => name === contract.name) }))
+      .flatMap(contract => [{
         contract: contract.name,
-        method: '[constructor]',
-        min: variation(...[contract.gasData, contract.previousGasData].map(x => ~~Math.min(...x))),
-        max: variation(...[contract.gasData, contract.previousGasData].map(x => ~~Math.max(...x))),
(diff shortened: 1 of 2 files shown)


scripts/checks/compareGasReports.js
```diff
@@ -44,7 +44,25 @@ class Report {
...
+    const deployments = update.info.deployments
+      .map(contract => Object.assign(
+        contract,
+        { previousVersion: ref.info.deployments.find(({ name }) => name === contract.name) },
+      ))
+      .filter(contract => contract.gasData?.length && contract.previousVersion?.gasData?.length)
+      .flatMap(contract => [{
+        contract: contract.name,
+        method: '[bytecode length]',
+        avg: variation(contract.bytecode.length - 2, contract.previousVersion.bytecode.length - 2),
```
frangio, line 57: I think we should divide by 2 given that it is a hex string.

scripts/checks/compareGasReports.js
```diff
@@ -44,7 +44,25 @@ class Report {
...
+      }, {
+        contract: contract.name,
+        method: '[construction cost]',
+        avg: variation(...[contract.gasData, contract.previousVersion.gasData].map(x => ~~average(...x))),
```
frangio, line 61:
```suggestion
        avg: variation(...[contract.gasData, contract.previousVersion.gasData].map(x => Math.floor(average(...x)))),
```

scripts/checks/compareGasReports.js
```diff
@@ -44,7 +44,25 @@ class Report {
...
+      }])
+      .sort((a, b) => `${a.contract}:${a.method}` < `${b.contract}:${b.method}` ? -1 : 1);
```
frangio, line 63: Minor issue for the purposes of this script, but the comparator function is expected to be symmetric, which this one isn't. On the other hand, maybe in this case we know that the array doesn't have duplicates? In any case a simple alternative is to use `str1.localeCompare(str2)`.

frangio (approved):

PR #3586, opened 2022-07-27
update comment for _beforeTokenTransfer

hamiltoncyber:
This commit makes the comment clearer for ERC721Enumerable._beforeTokenTransfer() by combing its calling condition items 4) and 5) into one condition: `from` and 'to' cannot be the zero address at the same time.


frangio (approved):
Thanks!

PR #3588, opened 2022-07-27
Add a GH action that checks changelog changed

Amxx:


.github/workflows/checks.yml
```diff
@@ -13,6 +13,23 @@ concurrency:
...
+        uses: ./.github/actions/setup
+      - name: Check diff
+        id: diff
+        run: |
+          git fetch origin ${{ github.base_ref }} --depth=1
+          git diff --exit-code origin/${{ github.base_ref }} -- CHANGELOG.md
+        continue-on-error: true
+      - name: Produce output
+        if: steps.diff.outcome == 'success'
+        run: echo 'Missing changelog entry'; exit 1
```
frangio, line 31:
```suggestion
          if git diff --exit-code origin/${{ github.base_ref }} -- CHANGELOG.md ; then
            echo 'Missing changelog entry'
            exit 1
          fi
```

.github/workflows/checks.yml
```diff
@@ -13,6 +13,23 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  changelog:
+    if: github.repository != 'OpenZeppelin/openzeppelin-contracts-upgradeable'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up environment
+        uses: ./.github/actions/setup
```
frangio, line 22: This isn't needed is it?

```suggestion
```

.github/workflows/checks.yml
```diff
@@ -13,6 +13,23 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  changelog:
+    if: github.repository != 'OpenZeppelin/openzeppelin-contracts-upgradeable'
```
frangio, line 17: I think we just want to run this in PRs, so we're missing a check here.

frangio:
I wonder if we're going to find this annoying for PRs that don't need a changelog update. But let's merge and find out!

Amxx:
> I wonder if we're going to find this annoying for PRs that don't need a changelog update. But let's merge and find out!

We won't be able to enable auto merge.

PR #3589, opened 2022-07-28
Extend Checkpoints with new sizes and lookup mechanisms

### LINKED ISSUE #3585: Add a Checkpoints lookup that optimizes for recency
First look at the most recent checkpoint, then move towards previous checkpoints in exponentially increasing intervals. Once past the target, do a normal binary search in the range.

Amxx:
This PR includes quite a few things:

- Procedural generation of the Checkpoint library, automating the support for various key/value lengths
  - `Checkpoints.Checkpoint224`: key is uint32 and value is uint224. Usecase: history checkpoints for voting
  - `Checkpoints.Checkpoint160`: key is uint96 and value is uint160. Usecase: Address sequence (for NFT batch minting)
- For each Checkpoint type, functions availables are:
  - `latest()`
  - `push(key, value)`
  - `lowerLookup(key)`: returns the `c.value` of the first checkpoint for which `key <= c.key` (or 0 if no such checkpoint exists)
  - `upperLookup(key)`: returns the `c.value` of the last checkpoint for which `key >= c.key` (or 0 if no such checkpoint exists)
  - `upperLookupRecent(key)`: same as `upperLookup` but optimised to look in recent checkpoints.
- `Checkpoint.History` now uses the `Checkpoints.Checkpoint224` structure, with block.number for keys.
  - ~~`getAtBlock` now uses `upperLookupRecent`, optimizing for recent value lookup~~ → delayed to further PR

Fixes #3585

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


scripts/generate/templates/Checkpoints.js
```diff
@@ -0,0 +1,167 @@
...
+}
+
+function upperLookupExpEnd(Checkpoint${length}[] storage self, uint${256 - length} key) internal view returns (uint224) {
+    uint256 length = self.length;
+    uint256 offset = 1;
+
+    while (offset <= length && self[length - offset]._key > key) {
+        offset <<= 1;
+    }
+
```
Amxx, line 114: Here, we could possibly do

```
if (offset > length && self[0]._key > key) {
    return 0;
}
```

I'm not sure how effective this optimization would be

Amxx:
Uses https://github.com/ethereum/solidity/issues/9117#issuecomment-1198003153 to skip checks

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
  * `SafeCast`: optimize downcasting of signed integers. ([#3565](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3565))
  * `ERC20FlashMint`: add an internal `_flashFee` function for overriding. ([#3551](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3551))
  * `VestingWallet`: remove unused library `Math.sol`. ([#3605](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3605))
+ * `Checkpoints`: Use provedural generation to support multiple key/value lengths. ([#3589](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3589))
```
frangio, line 16:
```suggestion
 * `Checkpoints`: Use procedural generation to support multiple key/value lengths. ([#3589](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3589))
```

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
...
+ * `Checkpoints`: Add new lookup mechanisms. ([#3589](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3589))
+ * `Array`: Add `unsafeAccess` functions that allow reading and writting to a element in a storage array while bypassing solidity's "out-of-array" check. ([#3589](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3589))
```
frangio, line 18:
```suggestion
 * `Array`: Add `unsafeAccess` functions that allow reading and writting to an element in a storage array bypassing Solidity's "out-of-bounds" check. ([#3589](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3589))
```

contracts/utils/Checkpoints.sol
```diff
@@ -15,21 +16,233 @@ import "./math/SafeCast.sol";
  * _Available since v4.5._
  */
 library Checkpoints {
-    struct Checkpoint {
```
frangio, line 18: Are we removing this one? Technically it should be kept for backwards compatibility...

contracts/utils/Checkpoints.sol
```diff
@@ -15,21 +16,233 @@ import "./math/SafeCast.sol";
...
+        uint32 _key;
         uint224 _value;
     }
 
+    function latest(Checkpoint224[] storage self) internal view returns (uint224) {
+        uint256 pos = self.length;
+        return pos == 0 ? 0 : _unsafeAccess(self, pos - 1)._value;
+    }
+
+    function push(
```
frangio, line 29: I don't think it's correct to implement all these methods for `CheckpointN[]` because they rely on invariants that are not guaranteed to be valid for an arbitrary array. The `History` struct encapsulates the array and guarantees the invariants.

I imagine this was done this way to avoid generating a `HistoryN` struct, but I think we need to do that and define these methods for that struct. The name could be something other than "History".

contracts/utils/Checkpoints.sol
```diff
@@ -15,21 +16,233 @@ import "./math/SafeCast.sol";
...
+        }
+
+        uint256 low = offset < length ? length - offset : 0;
+        uint256 high = length - (offset >> 1);
+        uint256 pos = _upperDichotomicLookup(self, key, low, high);
+
+        return pos == 0 ? 0 : _unsafeAccess(self, pos - 1)._value;
+    }
+
+    function _upperDichotomicLookup(
```
frangio, line 83: I wasn't familiar with the term "dichotomic search/lookup", looking at the Wikipedia page it says binary search is a kind of dichotomic search, so this is a dichotomic search but also a binary search (right?). I would prefer to use the term "binary search" or "binary lookup" which is more widely known.

contracts/utils/Checkpoints.sol
```diff
@@ -15,21 +16,233 @@ import "./math/SafeCast.sol";
  * _Available since v4.5._
  */
 library Checkpoints {
```
frangio, line 18: I think we should include a comment in the source code saying "the following is procedurally generated by the script X". Not in the natspec docs, just as a comment for anyone reading this file directly. This applies to SafeCast as well.

scripts/generate/templates/Checkpoints.js
```diff
@@ -0,0 +1,190 @@
+const format = require('../format-lines');
+
+const LENGTHS = [ 224, 160 ];
```
frangio, line 3: More like `VALUE_SIZES` in my opinion.

contracts/utils/Arrays.sol
```diff
@@ -31,18 +34,60 @@ library Arrays {
...
+        if (low > 0 && unsafeAccess(array, low - 1).value == element) {
             return low - 1;
         } else {
             return low;
         }
     }
+
+    /**
+     * @dev Access an array in an "unsafe" way. Skips solidity "index-out-of-range" check.
+     * @notice WARNING: only use if you are certain pos is lower than the array length.
```
frangio, line 54: This shouldn't be a `@notice`, it should just be part of `@dev`. `@notice` is what's shown to the end-user as they send a transaction. We're actually using it all backwards by putting everything in `@dev`, this is something we should fix at some other point.

contracts/utils/Checkpoints.sol
Amxx: This one is replaced by `Checkpoint224`. 

I asked that question a while back, and we said that since this is an internal structure, that devs are not expected to interact with directly, it was ok to rename it.

Ideally I would to something like `type Checkpoint is Checkpoint224;` ... but that is only authorized for value-types.

contracts/utils/Checkpoints.sol
Amxx: We need to come up with a name then.

contracts/utils/Checkpoints.sol
Amxx: AFAIK, dichotomic search and binary search are mostly equivalent. I believe that dichotomic is technically more precise, but if devs in our space are more familiar with the "binary search", I'm ok with using that instead

contracts/utils/Checkpoints.sol
Amxx: Also, If we have a `HistoryN` structure, then I guess the legacy `History` would include a `History224` instead of a `Checkpoint224[]` array ... which is breaking

contracts/utils/Checkpoints.sol
Amxx: Resolved in the generation file

contracts/utils/Checkpoints.sol
```diff
@@ -83,4 +68,314 @@ library Checkpoints {
...
+                self.push(Checkpoint({_blockNumber: key, _value: value}));
+            }
+            return (last._value, value);
+        } else {
+            self.push(Checkpoint({_blockNumber: key, _value: value}));
+            return (0, value);
+        }
+    }
+
+    function _upperBinaryLookup(
```
frangio, line 99: I think we should have at least minimal documentation for these functions, in this case for example we can say that the array is assumed to be sorted, and that high is assumed to be at most the length.

contracts/utils/Checkpoints.sol
```diff
@@ -83,4 +68,314 @@ library Checkpoints {
...
+
+    function _insert(
+        Checkpoint[] storage self,
+        uint32 key,
+        uint224 value
+    ) private returns (uint224, uint224) {
+        uint256 pos = self.length;
+
+        if (pos > 0) {
+            // Use of memory is important here.
```
frangio, line 80:
```suggestion
            // Copying to memory is important here
```

contracts/utils/Checkpoints.sol
```diff
@@ -83,4 +68,314 @@ library Checkpoints {
...
+            Checkpoint memory last = _unsafeAccess(self, pos - 1);
+
+            // Checkpoints keys must be increassing.
```
frangio, line 83:
```suggestion
            // Checkpoints keys must be increasing.
```

scripts/generate/templates/Checkpoints.js
```diff
@@ -0,0 +1,232 @@
...
+  valueFieldName: '_value',
+}, override);
+
+const OPTS = VALUE_SIZES.map(defaultOpts);
+
+const LEGACY_OPTS = defaultOpts(224, {
+  historyTypeName: 'History',
+  checkpointTypeName: 'Checkpoint',
+  keyFieldName: '_blockNumber',
+});
```
frangio, line 213: Renaming length to size and using spread operator for overrides which I think is a little simpler:

```suggestion
const defaultOpts = (size) => ({
  historyTypeName: `Trace${size}`,
  checkpointTypeName: `Checkpoint${size}`,
  checkpointFieldName: '_checkpoints',
  keyTypeName: `uint${256 - size}`,
  keyFieldName: '_key',
  valueTypeName: `uint${size}`,
  valueFieldName: '_value',
});

const OPTS = VALUE_SIZES.map(defaultOpts);

const LEGACY_OPTS = {
  ...defaultOpts(224),
  historyTypeName: 'History',
  checkpointTypeName: 'Checkpoint',
  keyFieldName: '_blockNumber',
};
```

scripts/generate/templates/Checkpoints.js
```diff
@@ -0,0 +1,232 @@
...
+  historyTypeName: `Trace${length}`,
+  checkpointTypeName: `Checkpoint${length}`,
+  checkpointFieldName: '_checkpoints',
+  keyTypeName: `uint${256 - length}`,
+  keyFieldName: '_key',
+  valueTypeName: `uint${length}`,
+  valueFieldName: '_value',
+}, override);
+
+const OPTS = VALUE_SIZES.map(defaultOpts);
```
frangio, line 207: As a heads up, `.map` will invoke the function with multiple arguments: the value, the index, and also the input array. So here it would invoke `defaultOpts(x, i)`, with `override = i`!

This is something to keep in mind when using `map`. It's preferable to always use the syntax `.map(x => fn(x))`.

contracts/utils/Checkpoints.sol
```diff
@@ -83,4 +68,314 @@ library Checkpoints {
...
+    struct Trace160 {
+        Checkpoint160[] _checkpoints;
+    }
+
+    struct Checkpoint160 {
+        uint96 _key;
+        uint160 _value;
+    }
+
+    function latest(Trace160 storage self) internal view returns (uint160) {
```
frangio, line 270: Need docs for all these functions.

contracts/utils/Checkpoints.sol
```diff
@@ -83,4 +68,314 @@ library Checkpoints {
...
+        uint256 pos = _upperBinaryLookup(self._checkpoints, key, 0, length);
+        return pos == 0 ? 0 : _unsafeAccess(self._checkpoints, pos - 1)._value;
+    }
+
+    function upperLookupRecent(Trace160 storage self, uint96 key) internal view returns (uint160) {
+        uint256 length = self._checkpoints.length;
+        uint256 offset = 1;
+
+        while (offset <= length && _unsafeAccess(self._checkpoints, length - offset)._key > key) {
+            offset <<= 1;
```
frangio, line 300: Should we be concerned about overflow? The shift operator will truncate whereas the below would revert.

```suggestion
            offset = offset ** 2;
```

contracts/utils/Arrays.sol
```diff
@@ -31,18 +34,63 @@ library Arrays {
...
+     *
+     * WARNING: only use if you are certain pos is lower than the array length.
+     */
+    function unsafeAccess(address[] storage arr, uint256 pos) internal pure returns (StorageSlot.AddressSlot storage) {
+        bytes32 slot;
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0, arr.slot)
+            slot := add(keccak256(0, 0x20), pos)
+        }
```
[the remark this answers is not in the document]
Amxx: I don't think so. I'm expecting other libraries to use that. We could make it private, but I think we will end up making it internal soon.

contracts/utils/Checkpoints.sol
Amxx: Offset is a power of 2. It starts at 1, and gets shifted.

For it to overflow, we would need `offset = 2**255` and the shift to make it go over. That would require `length >= 2**255`, which is unrealistic.

The overflow protection has a cost, and I think we can agree it will never be triggered by any legitimate usage of the library. Also, you code would revert while the current code will consume all the gas. In both cases you will not get a result.

contracts/utils/Checkpoints.sol
```diff
@@ -267,11 +314,19 @@ library Checkpoints {
...
+    /**
+     * @dev Returns the value in the latest checkpoint, or zero if there are no checkpoints.
+     */
     function latest(Trace160 storage self) internal view returns (uint160) {
         uint256 pos = self._checkpoints.length;
         return pos == 0 ? 0 : _unsafeAccess(self._checkpoints, pos - 1)._value;
     }
 
+    /**
+     * @dev Pushes a (`key`, `value`) pair into a History so that it is stored as the checkpoint.
```
frangio, line 326: We're saying History even though it's a Trace here.

scripts/generate/templates/Checkpoints.js
```diff
@@ -146,6 +166,12 @@ function _insert(
     }
 }
 
+/**
+ * @dev Return the index of the last checkpoint for which \`key\` >= checkpoint.key, i.e. the most recent checkpoint which key is lower or equal than the search key.
```
frangio, line 170: This is not right! It is not the most recent with lower or equal key. It is the next one after that one, so the _least_ recent with _greater_ key.

```suggestion
 * @dev Return the index of the oldest checkpoint whose key is greater than the search key, or `high` if there is none.
```

scripts/generate/templates/Checkpoints.js
```diff
@@ -163,6 +189,12 @@ function _upperBinaryLookup(
     return high;
 }
 
+/**
+ * @dev Return the index of the first checkpoint for which \`key\` <= checkpoint.key, i.e. the oldest checkpoint which key is higher or equal than the search key.
```
frangio, line 193: The first part is confusing due to ambiguity of the first occurence of the word "key".

```suggestion
 * @dev Return the index of the oldest checkpoint whose key is greater or equal than the search key, or `high` if there is none.
```

scripts/generate/templates/Checkpoints.js
```diff
@@ -32,11 +32,19 @@ struct ${opts.checkpointTypeName} {
 
 /* eslint-disable max-len */
 const operations = opts => `\
+/**
+ * @dev Returns the value in the latest checkpoint, or zero if there are no checkpoints.
```
frangio, line 36:
```suggestion
 * @dev Returns the value in the most recent checkpoint, or zero if there are no checkpoints.
```

scripts/generate/templates/Checkpoints.js
```diff
@@ -45,18 +53,27 @@ function push(
...
+ * @dev Return the value most recent checkpoint which key is lower or equal than the search key.
+ */
 function lowerLookup(${opts.historyTypeName} storage self, ${opts.keyTypeName} key) internal view returns (${opts.valueTypeName}) {
     uint256 length = self.${opts.checkpointFieldName}.length;
     uint256 pos = _lowerBinaryLookup(self.${opts.checkpointFieldName}, key, 0, length);
     return pos == length ? 0 : _unsafeAccess(self.${opts.checkpointFieldName}, pos).${opts.valueFieldName};
 }
 
+/**
+ * @dev Return the value oldest checkpoint which key is higher or equal than the search key.
```
frangio, line 66: I think you got the docs for upper and lower reversed.

```suggestion
 * @dev Returns the value in the most recent checkpoint with key lower or equal than the search key.
```

scripts/generate/templates/Checkpoints.js
```diff
@@ -45,18 +53,27 @@ function push(
...
+ */
 function upperLookup(${opts.historyTypeName} storage self, ${opts.keyTypeName} key) internal view returns (${opts.valueTypeName}) {
     uint256 length = self.${opts.checkpointFieldName}.length;
     uint256 pos = _upperBinaryLookup(self.${opts.checkpointFieldName}, key, 0, length);
     return pos == 0 ? 0 : _unsafeAccess(self.${opts.checkpointFieldName}, pos - 1).${opts.valueFieldName};
 }
 
+/**
+ * @dev Return the value oldest checkpoint which key is higher or equal than the search key (similarly to {upperLookupRecent}) doing an exponential lookup first to optimize for recent checkpoints.
```
frangio, line 75:
```suggestion
 * @dev Returns the value in the most recent checkpoint with key lower or equal than the search key (similarly to {upperLookup}), optimized for the case when the search key is known to be recent.
```

scripts/generate/templates/Checkpoints.js
```diff
@@ -45,18 +53,27 @@ function push(
     return _insert(self.${opts.checkpointFieldName}, key, value);
 }
 
+/**
+ * @dev Return the value most recent checkpoint which key is lower or equal than the search key.
```
frangio, line 57:
```suggestion
 * @dev Returns the value in the oldest checkpoint with key greater or equal than the search key, or zero if there is none.
```

test/utils/Checkpoints.test.js
```diff
@@ -4,71 +4,144 @@ const { expect } = require('chai');
...
-    beforeEach('pushing checkpoints', async function () {
-      this.tx1 = await this.checkpoint.push(1);
-      this.tx2 = await this.checkpoint.push(2);
-      await time.advanceBlock();
-      this.tx3 = await this.checkpoint.push(3);
-      await time.advanceBlock();
-      await time.advanceBlock();
-    });
+  for (const length of [160, 224]) {
+    describe(`Checkpoint${length}[]`, function () {
```
frangio, line 81:
```suggestion
    describe(`Trace${length}`, function () {
```

test/utils/Checkpoints.test.js
```diff
@@ -4,71 +4,144 @@ const { expect } = require('chai');
...
-        () => this.checkpoint.push(8, { gas: 100000 }),
-        () => this.checkpoint.push(9, { gas: 100000 }),
-        () => this.checkpoint.push(10, { gas: 100000 }),
-      ]);
-      const lengthAfter = await this.checkpoint.length();
-
-      expect(lengthAfter.toNumber()).to.be.equal(lengthBefore.toNumber() + 1);
-      expect(await this.checkpoint.latest()).to.be.bignumber.equal('10');
+          await this.contract.push(last(...this.checkpoints).key, newValue);
+          expect(await this.contract.latest()).to.be.bignumber.equal(newValue);
```
frangio, line 125: Can we also check that this was not due to adding a new checkpoint? I.e. the number of checkpoints remained the same. Currently the test wouldn't detect if a new checkpoint was pushed.

test/utils/Checkpoints.test.js
```diff
@@ -4,71 +4,144 @@ const { expect } = require('chai');
...
+      });
 
-    it('reverts if block number >= current block', async function () {
-      await expectRevert(
-        this.checkpoint.getAtBlock(await web3.eth.getBlockNumber()),
-        'Checkpoints: block not yet mined',
-      );
+      describe('with checkpoints', function () {
+        beforeEach('pushing checkpoints', async function () {
+          this.checkpoints = [
```
frangio, line 100: This is scarily simple for "coverage". I'd change it so the values are mostly not equal to the keys, and not in increasing order.

scripts/generate/templates/Checkpoints.js
```diff
@@ -0,0 +1,265 @@
...
+
+// GENERATE
+module.exports = format(
+  header.trimEnd(),
+  'library Checkpoints {',
+  [
+    // Legacy types & functions
+    types(LEGACY_OPTS),
+    legacyOperations(LEGACY_OPTS),
+    helpers(LEGACY_OPTS),
```
frangio, line 256: We're missing `upperLookupRecent` for the `History` struct. It should also be tested, currently isn't.

frangio (commented):
Mostly minor comments. We can merge after this.

scripts/generate/templates/Checkpoints.js
Amxx: should it be called `getAtBlockRecent`, just like `getAtBlock` is the name of `upperLookup` for History (with the extra check) ?

PR #3608, opened 2022-08-08
Preserve camelCase in GovernorCountingSimple

goncaloMagalhaes:
Renamed `proposalvote` to `proposalVote` in `GovernorCountingSimple` to preserve camelCase.


Amxx:
Thank you @goncaloMagalhaes 

For the record, this is a small change that doesn't need a changelog entry. Thank for adding it to have the tests pass, but I removed it and will bypass this test when merging.

PR #3609, opened 2022-08-10
fix _verify function in utilities.adoc

seokwoo-ahn:
Fixes #???? 

#### PR Checklist

- [y] Tests
- [y] Documentation
- [ ] Changelog entry

According to utilities.adoc provided by openzeppelin, the _verify function is implemented just as address is required to validate ECDSA signatures.
However, the Ethereum engine does not contain the from address in tx, and generate address from ECDSA verification process.
and ECDSA verification is possible only with signature and hash values without comparing the address information.

Therefore, I changed the _verify function of utilities.doc without address information.
and I changed it based on when executing ECDSA.recover function with invalid hash value and signatures, ECDSA.recover function emits an error message or return the address(0) value.

### DIFF
--- a/docs/modules/ROOT/pages/utilities.adoc
+++ b/docs/modules/ROOT/pages/utilities.adoc
@@ -15,10 +15,13 @@ The data signer can be recovered with xref:api:cryptography.adoc#ECDSA-recover-b
 ----
 using ECDSA for bytes32;
 
-function _verify(bytes32 data, bytes memory signature, address account) internal pure returns (bool) {
-    return data
-        .toEthSignedMessageHash()
-        .recover(signature) == account;
+function _verify(bytes32 data, bytes memory signature) internal pure returns (bool) {
+    address account = data.toEthSignedMessageHash().recover(signature);
+    if(account == address(0)) {
+        return false;
+    } else {
+        return true;
+    }
 }
 ----
 


Amxx:
Hello @seokwoo-ahn 

I'm not really sure what your issue with the current do is, but the propose changes do not look right to me for many reason:

- you remove the `account` variable, which means that you are not checking the recovered address.
- you check if the recovered address is 0, but that check is already part of `ECDSA.recover` ... so I don't see the point.
- in the end, any account could produce a valid signature, which defeats the entire point of having a signature.

Amxx:
Also,

```
    if(account == address(0)) {
        return false;
    } else {
        return true;
    }
```
can be rewritten as
```
return account != address(0);
``` 

So your entire proposal can be rewritten as

```
    return data.toEthSignedMessageHash().recover(signature) != address(0);
```
which is obviously wrong.

seokwoo-ahn:
hello @Amxx
to verify ECDSA signature, **address are not required for ECDSA verification**. just by check with hash value and signature you can verify ECDSA, and during verification process you can get the address of the person who signed it.
if we can get address by using ECDSA.recover function without error or getting address(0), that means verification process was completed and we can be sure that address get from by using ECDSA.recover is signatory address

so i want to recommend to delete address argument in _verify function. because you can verify signature without address.

seokwoo-ahn:
@Amxx 
additionally, I agree with all accounts can generate valid transactions and signatures.
but in the process of signing, since sign with private key, all accounts generate different signatures with same transactions.
Therefore in the process of verification, different address can get from ECDSA.recover(signature)

Amxx:
I'm sorry, I don't really understand what you are trying to say. It looks like your usecase (which I don't understand) is not the usecase that we are trying to document.

The normal usecase is:
- We expect address A to sign a message
- We get some operation request, with a signature
- We rebuild the message, and recover the signer
- We check that the signer is A
- If the signer is A, then we do the operation, knowing that A approved it.

This is what this library is usually used for, and this is what we document

You may have a different usage of this library (which again, I don't understand, and might be unsafe) ... but that is not what we are documenting.

seokwoo-ahn:
thank you for your answer.

my purpose of this pull request is remove the confusion by changing _verify function.

Many people think they **must** receive the signer's address In order to verify the ECDSA signature in the field.
because _verify function needs signer's address as written in openzeppelin docs.

However, for only ECDSA verification, it is not necessary to receive signer's address.
and if the verification process is completed without any problems, the address generated by ECDSA verification is sure to be the signers address.
so I created pull requester because I want to tell it is not necessary to receive signer's address and compare the address with generated address for ECDSA verification.

I will organize this contents so that it is easy to understand and rewrite them on the issue.

Amxx:
For ECDSA verification it is necessary to check the sender (I don't know any usecase where the signer is not somehow verified).

This documents a possible internal function. It shows how ECDSA is most frequently used. That's it. This documentation is not the place to explain uncommon non-standard usecases.

PR #3610, opened 2022-08-10
Fix ECDSA signature malleability

frangio:
Fixes a kind of signature malleability that is present in the `ECDSA.recover` and `tryRecover` variants that take a `bytes signature` argument, which can be both 65 bytes long and the compact 64 bytes encoded version.


Amxx (approved):
All good !

y0wl:
For reference, this pull request is motivated by this security advisory - https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories/GHSA-4h98-2769-gh6h

PR #3611, opened 2022-08-11
Change execution order to avoid reentry through the _beforeTokenTransfer hook

Amxx:
If a developer was to transfer/burn/mint the token during the `_beforeTokenTransfer`, that would currently not be catched by checks, resulting in invalid event being emitted, and corrupted balances.

This could in particular affect a voting instance.

This risk could become serious if the `_beforeTokenTransfer` hook was to perform an external call that could reenter!

We eliminate this risk by applying strict "check→effect" logic to the `_mint`, `_transfer` and `_burn` function, at the risk of being more gas expensive. For `_mint` and `_transfer`, the increase in gas cost only affect reverting transaction (the revert latter). For `_burn`, an addition hot-sload (100gas) is needed, making all burns more exepensive. Hopefully burn is not as widelly and frequently used as mint and transfer.

IMO this is one more reason to drop these hooks when possible (5.0)

#### PR Checklist

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
  * `SafeCast`: optimize downcasting of signed integers. ([#3565](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3565))
  * `ERC20FlashMint`: add an internal `_flashFee` function for overriding. ([#3551](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3551))
  * `VestingWallet`: remove unused library `Math.sol`. ([#3605](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3605))
- 
+
 ### Compatibility Note
 
 ERC-721 integrators that interpret contract state from events should make sure that they implement the clearing of approval that is implicit in every transfer according to the EIP. Previous versions of OpenZeppellin Contracts emitted an explicit `Approval` event even though it was not required by the specification, and this is no longer the case.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,16 @@
  * `ERC165Checker`: add `supportsERC165InterfaceUnchecked` for consulting individual interfaces without the full ERC165 protocol. ([#3339](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3339))
  * `Address`: optimize `functionCall` by calling `functionCallWithValue` directly. ([#3468](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3468))
  * `Address`: optimize `functionCall` functions by checking contract size only if there is no returned data. ([#3469](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3469))
- * `GovernorCompatibilityBravo`: remove unused `using` statements ([#3506](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3506))
+ * `GovernorCompatibilityBravo`: remove unused `using` statements. ([#3506](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3506))
  * `ERC20`: optimize `_transfer`, `_mint` and `_burn` by using `unchecked` arithmetic when possible. ([#3513](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3513))
  * `ERC721`: optimize transfers by making approval clearing implicit instead of emitting an event. ([#3481](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3481))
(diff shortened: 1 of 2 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -302,19 +302,21 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
      * Emits a {Transfer} event.
      */
     function _burn(uint256 tokenId) internal virtual {
-        address owner = ERC721.ownerOf(tokenId);
+        address from = ERC721.ownerOf(tokenId);
+
+        _beforeTokenTransfer(from, address(0), tokenId);
 
-        _beforeTokenTransfer(owner, address(0), tokenId);
+        require(from == ERC721.ownerOf(tokenId), "ERC721: burn from incorrect owner");
```
frangio, line 309: Instead of this require, why don't we just burn from whoever the owner is after `_beforeTokenTransfer`? After all, the function signature isn't requesting that we burn this from a specific owner.

We should add a comment why we're reading the owner again after the hook.

CHANGELOG.md
```diff
@@ -5,10 +5,11 @@
...
  * `Address`: optimize `functionCall` functions by checking contract size only if there is no returned data. ([#3469](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3469))
- * `GovernorCompatibilityBravo`: remove unused `using` statements ([#3506](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3506))
+ * `GovernorCompatibilityBravo`: remove unused `using` statements. ([#3506](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3506))
  * `ERC20`: optimize `_transfer`, `_mint` and `_burn` by using `unchecked` arithmetic when possible. ([#3513](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3513))
  * `ERC721`: optimize transfers by making approval clearing implicit instead of emitting an event. ([#3481](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3481))
  * `ERC721`: optimize burn by making approval clearing implicit instead of emitting an event. ([#3538](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3538))
+ * `ERC721`: Change execution order to avoid reentry through the `_beforeTokenTransfer` hook. ([#3611](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3611))
```
frangio, line 12: This is not about _avoiding_ reentrancy is it? It's more about preserving correct balances, and not just about reentrancy.

```suggestion
 * `ERC721`: Change execution order to preserve correct balances with custom `_beforeTokenTransfer` hooks. ([#3611](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3611))
```

contracts/token/ERC721/ERC721.sol
Amxx: If we do that, and if the `_beforeTokenTransfer` does a transfer, we end up with inconsistent events:

```
Transfer(<original>, <receiver>, tokenId);
Transfer(<original>, address(0), tokenId);
```

When we should have

```
Transfer(<original>, <receiver>, tokenId);
Transfer(<receiver>, address(0), tokenId);
```

Amxx: Also, the `_balances[from] -= 1;` is no longer correct.

Amxx: We could however replace this require with a refresh of the owner.

```
address from = ERC721.ownerOf(tokenId);
```

frangio: That's what I meant. To refresh the owner.

Amxx: done

contracts/token/ERC721/ERC721.sol
```diff
@@ -333,11 +336,11 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         address to,
         uint256 tokenId
     ) internal virtual {
-        require(ERC721.ownerOf(tokenId) == from, "ERC721: transfer from incorrect owner");
```
frangio, line 336: It is wrong to remove this from here... We need to check that `from` is the owner at the time of the transfer. We should do the same we did in `_burn`, and re-read the owner after the hook...

Amxx: My impression was that it would be checked after the hook, and that would revert in the owner is not correct.

The risk is that the hook would be executed with an invalid `from`, and the revert would come after. 

There is a world in which the from is not valid, the before hook move the token from its current owner to the from, and then the test passes and we do what we have to do... The code would have to be really f****d up, but its possible.

My concern is that doing the test twice increase transfer cost by >100 gas (on probably the most common and scrutinized operation)

frangio (approved):
LGTM

PR #3617, opened 2022-08-14
Add a MerkleTree builder

Amxx:
Structure inspired by Tornardo Cash's commitment tree.

- Initialize a complete merkle tree of given depth, with all leaves having the same "zero" value.
- Insert leaves one at a time, recomputing the merkle root in the process
- Keep a "limited" (rotating) history of past merkle roots.

Fixes: #4758

#### PR Checklist
- [x] Use #3589 when merged
- [x] Tests
- [x] Documentation
- [x] Changelog entry


Amxx:
Depends on 3589

Amxx:
Just realized that if someone could be creative and commit merle roots of subtrees that use the same hashings function. That would allow verification of the leafs deep inside the subtrees

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,189 @@
...
+        bytes32[] sides;
+        bytes32[] zeros;
+        bytes32[] roots;
+        function(bytes32, bytes32) view returns (bytes32) fnHash;
+    }
+
+    /**
+     * @dev Initialize using the default hash
+     */
+    function initialize(TreeWithHistory storage self, uint256 depth, uint256 length, bytes32 zero) internal {
```
ernestognw, line 63: What do you think of renaming this function to `setUp`? I feel like `initializable` conflicts with the concept of initializer in upgradeable contracts.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,189 @@
...
+ * - length of the root history
+ * - zero values (for "empty" leaves)
+ * - hash function
+ *
+ * WARNING:
+ *
+ * By design, the tree include zero leaves. Customizing the "zero value" might be necessary to ensure that empty leaves
+ * being provably part of the tree is not a security issue.
+ *
+ * _Available since v4.x._
```
ernestognw, line 28:
```suggestion
 * _Available since v5.1._
```

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,189 @@
...
+ * Each tree can be customized to use specific
+ * - depth
+ * - length of the root history
+ * - zero values (for "empty" leaves)
+ * - hash function
+ *
+ * WARNING:
+ *
+ * By design, the tree include zero leaves. Customizing the "zero value" might be necessary to ensure that empty leaves
+ * being provably part of the tree is not a security issue.
```
ernestognw, line 26:
```suggestion
 * WARNING: By design, the tree include zero leaves. Customizing the "zero value" might be necessary to ensure that empty leaves
 * being provably part of the tree is not a security issue.
```

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,189 @@
...
+    function initialize(
+        TreeWithHistory storage self,
+        uint256 depth,
+        uint256 length,
+        bytes32 zero,
+        function(bytes32, bytes32) view returns (bytes32) fnHash
+    ) internal {
+        if (depth > MAX_DEPTH) {
+            revert MerkleTreeInvalidDepth(depth, MAX_DEPTH);
+        }
```
ernestognw, line 83: Is it realistic to make the check? Not sure how costly will be to write to storage and perform a hash operation 255 times, but I wonder if it actually fits in a block.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,189 @@
...
+     * The hashing function used during initialization to compute the `zeros` values (value of a node at a given depth
+     * for which the subtree is full of zero leaves). This function is kept in the structure for handling insertions.
+     */
+    struct TreeWithHistory {
+        uint256 currentRootIndex;
+        uint256 nextLeafIndex;
+        bytes32[] sides;
+        bytes32[] zeros;
+        bytes32[] roots;
+        function(bytes32, bytes32) view returns (bytes32) fnHash;
```
ernestognw, line 57: If we receive a `function(bytes)` instead, we can add built-in leaf assymetry, I think we should strive to have assymetry before releasing this tree.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,189 @@
...
+    function insert(TreeWithHistory storage self, bytes32 leaf) internal returns (uint256) {
+        // Cache read
+        uint256 depth = self.zeros.length;
+        function(bytes32, bytes32) view returns (bytes32) fnHash = self.fnHash;
+
+        // Get leaf index
+        uint256 leafIndex = self.nextLeafIndex++;
+
+        // Check if tree is full.
+        if (leafIndex == 1 << depth) Panic.panic(Panic.RESOURCE_ERROR);
```
ernestognw, line 114: Shouldn't this ideally be?

```suggestion
        if (leafIndex >= 1 << depth) Panic.panic(Panic.RESOURCE_ERROR);
```

contracts/mocks/MerkleTreeMock.sol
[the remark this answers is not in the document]
Amxx: The exposed version does not give access to the internals of the structure:
- _tree.currentRootIndex;
- _tree.nextLeafIndex;
- _tree.sides[i];
- _tree.zeros[i];
- _tree.roots[i];

IMO, it doesn make sens to expose them through internal functions in the library. This is purelly internal and devs should not touch that. However, in unit testing I think its nice to check that it "moves" correctly. That requires accessor that are not exposed ... hence the custom mock.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,189 @@
...
+     * - All leaves are initialize to `zero`
+     * - Hashing function for a pair of leaves is fnHash
+     * and keep a root history of length `length` when leaves are inserted.
+     */
+    function initialize(
+        TreeWithHistory storage self,
+        uint256 depth,
+        uint256 length,
+        bytes32 zero,
+        function(bytes32, bytes32) view returns (bytes32) fnHash
```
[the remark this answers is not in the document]
Amxx: > How many fnHash variants people might end up? I personally think probably just sha256 and keccak256 so perhaps it's worth just an enum to avoid having function pointers.

They may want to use zk friendly hashes that we don't have implementation for.

contracts/utils/structs/MerkleTree.sol
Amxx: `setUp` is good.

contracts/utils/structs/MerkleTree.sol
Amxx: I'm not sure what bytes would bring here ... appart from the cost of encoding/decoding.

ernestognw: We would be able to do `fnHash(fnHash(...))` for the leafs 🤔

contracts/utils/structs/MerkleTree.sol
Amxx: unless someone starts manually tampering with leafIndex, it will only be increment one-by-one ... so it would have to be equal (and revert) before it is greater. 

But the cost of that change is not much, so if you feel like its safer ... I don't see an issue with this change.

contracts/utils/structs/MerkleTree.sol
Amxx: I don't think this structure should take care of hashing the leafs. The dev using that should hash the leafs, then inject them into that structure, and we only take care of maintaining internal nodes. We don't know which hashes are going to be used in the future, both for the leaf part and the internal part. Particularly in the case of ZK structure. 
I don't want to make the (strong) assumption that leafs are generated by applying twice the internal hash.

Amxx: I don't want to get into trouble, so I'm not going to post T*rnad* code here.
I can share it privatly on slack.

ernestognw: Right, sorry for bringing the assymetry discussion back again haha

> The dev using that should hash the leafs, then inject them into that structure, and we only take care of maintaining internal nodes. 

My feeling is that we need an usage example in the docs and also make clear this is the intention. Otherwise the convo will come back again.

contracts/utils/structs/MerkleTree.sol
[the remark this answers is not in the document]
Amxx: (example: tørnadø's MiMC)

contracts/utils/structs/MerkleTree.sol
ernestognw: Solving and adding this to the docs backlog

contracts/mocks/MerkleTreeMock.sol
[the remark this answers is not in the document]
ernestognw: Ok, I agree this is one of the cases where a custom mock is better suited.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * @dev Return the current root of the tree.
+     */
+    function getLastRoot(TreeWithHistory storage self) internal view returns (bytes32) {
+        return Arrays.unsafeAccess(self.roots, self.currentRootIndex).value;
+    }
+
+    /**
+     * @dev Look in root history,
```
ernestognw, line 167:
```suggestion
     * @dev Look in root history. Limited to {getLength} elements.
```

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,187 @@
...
+     */
+    function isKnownRoot(TreeWithHistory storage self, bytes32 root) internal view returns (bool) {
```
ernestognw, line 169: What are we trying to achieve with this function?

I think the length of the tree reduces the utility of this function since it will start reporting `false` after the length has been reached and the values are started to get overridden. If the function is on-chain, I think it's still useful if the tree keeps values for a certain period, but doesn't sound like that's going the be everyone's use case.

In my opinion, a function like this should be callable in `O(1)` and keep knowing all the roots that have been registered. So it'll be better if we have a getter to `self.roots[i]` and we let developers implement their `isKnownRoot` (perhaps we do it through an extension and just use a Bitmap for it).

However, if we add a getter for `self.roots[i]` we should be careful of allowing arbitrary length, as we may be adding a getter for the whole. I'd suggest restricting root's length to `uint8`

contracts/utils/structs/MerkleTree.sol
Amxx: removed

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,187 @@
...
+        if (root == 0) {
+            return false;
+        }
+
+        // Cache read
+        uint256 currentRootIndex = self.currentRootIndex;
+        uint256 length = self.roots.length;
```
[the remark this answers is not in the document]
Amxx: removed

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,187 @@
...
+                isLeft ? currentLevelHash : Arrays.unsafeAccess(self.sides, i).value,
+                isLeft ? Arrays.unsafeAccess(self.zeros, i).value : currentLevelHash
+            );
+
+            // Update node index
+            currentIndex >>= 1;
+        }
+
+        // Record new root
+        self.currentRootIndex = (self.currentRootIndex + 1) % self.roots.length;
```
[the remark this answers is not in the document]
Amxx: no longer relevant

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,187 @@
...
+        TreeWithHistory storage self,
+        uint256 depth,
+        uint256 length,
+        bytes32 zero,
+        function(bytes32, bytes32) view returns (bytes32) fnHash
+    ) internal {
+        if (depth > MAX_DEPTH) {
+            revert MerkleTreeInvalidDepth(depth, MAX_DEPTH);
+        }
+
```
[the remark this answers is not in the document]
Amxx: no longer relevant

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,187 @@
...
+        // Store depth & length in the dynamic array
+        Arrays.unsafeSetLength(self.sides, depth);
+        Arrays.unsafeSetLength(self.zeros, depth);
+        Arrays.unsafeSetLength(self.roots, length);
```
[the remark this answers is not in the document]
Amxx: no longer relevant

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,187 @@
...
+        bytes32[] zeros;
+        bytes32[] roots;
+        function(bytes32, bytes32) view returns (bytes32) fnHash;
+    }
+
+    /**
+     * @dev Initialize using the default hash
+     */
+    function setUp(TreeWithHistory storage self, uint256 depth, uint256 length, bytes32 zero) internal {
+        return setUp(self, depth, length, zero, MerkleProof.hashPair);
```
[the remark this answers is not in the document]
Amxx: Moved (and renamed) `MerkleProof.hashPair` to `Hashes.stdPairHash`.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * @dev The `sides` and `zero` arrays are set, at initialization, to have a length equal to the depth of the tree.
+     * No push/pop operations should be performed of these array, and their lengths should not be updated.
+     *
+     * The hashing function used during initialization to compute the `zeros` values (value of a node at a given depth
+     * for which the subtree is full of zero leaves). This function is kept in the structure for handling insertions.
+     *
+     * Contracts using this structure may want to use a secondary structure to store a (partial) list of historical
+     * roots. This could be done using a cicular buffer (to keep the last N roots) or the {Checkpoints} library to
```
Amxx, line 48:
```suggestion
     * roots. This could be done using a circular buffer (to keep the last N roots) or the {Checkpoints} library to
```

contracts/utils/structs/MerkleTree.sol
Amxx: I think we need to put an arbitrary limit. I used 255, but really we would use anything "bug enough". 32 may be on the border. 64 would definitelly be enough for any real usecase.

Do you want to re-evaluate the limit ?

contracts/utils/structs/MerkleTree.sol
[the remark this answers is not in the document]
Amxx: If tree is already initialized, we could do everything in this function + set the nextLeafIndex to 0 ... that would "cleanup" everything. You'd start with a "brand new" tree.

contracts/utils/structs/MerkleTree.sol
ernestognw: No, I changed my main for two reasons:

1. 255 already allows 2**255 leaves, which is just before the `type(uint256).max`
2. 255 is `type(uint8).max`

I think we should treat `depth` the same way ERC20 treats `decimals` and restrict it to a `uint8`. I'll update

contracts/utils/cryptography/Hashes.sol
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * @dev Library of standard hash functions.
+ */
+library Hashes {
+    /**
+     * @dev Keccak256 hash of a sorted pair of bytes32. Frequently used when working with merkle proofs.
+     */
+    function stdPairHash(bytes32 a, bytes32 b) internal pure returns (bytes32) {
+        return a < b ? _efficientHash(a, b) : _efficientHash(b, a);
```
ernestognw, line 13: Just realized this function is commutative. Should the tree work with non-commutative functions?

I guess it should, so I'd feel better if we add tests with a custom non-commutative `keccak256` function. It may require updating the javascript library for testing.

If the tree expects only commutative functions we should be explicit about it.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,160 @@
...
+    function getDepth(Bytes32MerkleTree storage self) internal view returns (uint256) {
+        return self.zeros.length;
+    }
+
+    /**
+     * @dev Return the current root of the tree.
+     */
+    function getRoot(Bytes32MerkleTree storage self) internal view returns (bytes32) {
+        return self.root;
+    }
```
ernestognw: This function is not needed since anyone can just call `tree.root`. Let's remove it

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,160 @@
...
+        // Record new root
+        self.root = currentLevelHash;
+
+        return leafIndex;
+    }
+
+    /**
+     * @dev Tree's depth (set at initialization)
+     */
+    function getDepth(Bytes32MerkleTree storage self) internal view returns (uint256) {
```
ernestognw, line 140: I'd rename this as just `depth`:

```suggestion
    function depth(Bytes32MerkleTree storage self) internal view returns (uint256) {
```

What do you think?

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,160 @@
...
+     * for which the subtree is full of zero leaves). This function is kept in the structure for handling insertions.
+     *
+     * Contracts using this structure may want to use a secondary structure to store a (partial) list of historical
+     * roots. This could be done using a circular buffer (to keep the last N roots) or the {Checkpoints} library to
+     * keep a more complete history. Note that if using the Checkpoints.Trace224 structure for storing roots, you will
+     * be limited to keeping "only" 26 bytes out of the root's 32. This should not be a security issue.
+     */
+    struct Bytes32MerkleTree {
+        bytes32 root;
+        uint256 nextLeafIndex;
```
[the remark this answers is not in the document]
ernestognw: What do you think of renaming the `Bytes32MerkleTree` to `Bytes32InsertableTree` or similar? The more I think about this the more I realize this structure is not only a merkle tree but one that's designed for sequential insertion and I think its name should reflect that.

We can still remove leaves from a `Bytes32InsertableTree` if we don't allow setting it again.

ernestognw (approved):

contracts/utils/structs/MerkleTree.sol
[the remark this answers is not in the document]
ernestognw: Right, this sounds about right.

test/utils/structs/Merkletree.test.js
```diff
@@ -0,0 +1,61 @@
...
+async function fixture() {
+  return { mock: await ethers.deployContract('MerkleTreeMock', [DEPTH, ZERO]) };
+}
+
+describe('MerkleTree', function () {
+  beforeEach(async function () {
+    Object.assign(this, await loadFixture(fixture));
+  });
+
+  it('sets initial values at setup', async function () {
```
ernestognw: We're missing a test case with a second setup after using the tree. We should have that test case since we're documenting it can be done.

contracts/utils/cryptography/Hashes.sol
```diff
@@ -0,0 +1,27 @@
...
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Library of standard hash functions.
+ */
+library Hashes {
+    /**
+     * @dev Keccak256 hash of a sorted pair of bytes32. Frequently used when working with merkle proofs.
+     */
```
ernestognw, line 11: I think this property is important, he

```suggestion
     * @dev Commutative Keccak256 hash of a sorted pair of bytes32. Frequently used when working with merkle proofs.
     */
```

ernestognw:
Took the freedom to update the documentation. I left some comments and I think this still needs some work but it's looking really good in my opinion.

contracts/utils/cryptography/Hashes.sol
```diff
@@ -0,0 +1,27 @@
...
+     * @dev Implementation of keccak256(abi.encode(a, b)) that doesn't allocate or expand memory.
+     */
+    function _efficientHash(bytes32 a, bytes32 b) private pure returns (bytes32 value) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0x00, a)
+            mstore(0x20, b)
+            value := keccak256(0x00, 0x40)
+        }
+    }
```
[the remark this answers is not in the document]
Amxx: Going further, what about `sortedPairKeccak256` instead `stdPairKeccak256` ?

contracts/utils/structs/MerkleTree.sol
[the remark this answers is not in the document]
Amxx: > I see the tree should be strictly sequential to allow insertion with the sides variable. However, we have a removal functionality in our roadmap. How's that going to interact with this index?

Quick answer is: this structure cannot do removals.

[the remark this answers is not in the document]
Amxx: > The more I think about this the more I realize this structure is not only a merkle tree but one that's designed for sequential insertion and I think its name should reflect that.

That is fair. I'm not a huge fan of `Bytes32InsertableTree`. Lets brainstorm the name.

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,160 @@
...
+        bytes32[] sides;
+        bytes32[] zeros;
```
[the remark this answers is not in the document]
Amxx: > I think it would be better if there was a way to precompute the zeros in constants or immutable variables so each access is cheaper.

immutable onyl support value type. Here we are dealing with an array which length is only known at runtime (after compilation). Immutable (and constant) don't support that.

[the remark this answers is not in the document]
Amxx: > I think the "zero" value is only used to avoid the actual zero being a provable part of the tree.

I've seen other zero used. In particular ones that are produced by doing `hash("")` with hash being any hashing function (keccak256, MiMc, Pedersen, ...)

contracts/utils/cryptography/Hashes.sol
Amxx: We should support non commutative hashes, so yes, we should test that.

contracts/utils/structs/MerkleTree.sol
Amxx: that creates naming conflicts, doesn't it?

contracts/utils/structs/MerkleTree.sol
Amxx: Ideally, we would like the users to never touch the internal variable. The risk of them touching `zeros` or `nextLeafIndex` are too big. 

Imagine someone forgets a `=` and does
```
require(tree.root = expected, "invalid tree");
```

Ideally we would mark the variables as private, and provide getters for everything.
I'd rather rename the variable to add a prefix, and keep that function.

contracts/utils/cryptography/Hashes.sol
```diff
@@ -0,0 +1,27 @@
...
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Library of standard hash functions.
+ */
+library Hashes {
+    /**
+     * @dev Keccak256 hash of a sorted pair of bytes32. Frequently used when working with merkle proofs.
+     */
+    function stdPairHash(bytes32 a, bytes32 b) internal pure returns (bytes32) {
```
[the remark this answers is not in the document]
Amxx: see https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3617#discussion_r1497081320

test/utils/structs/Merkletree.test.js
Amxx: added

ernestognw:
I gave it a thought and I started to see why you wanted to call the function `stdPairKeccak256` like that in the `Hashes` library considering the context of the Javascript library (assuming it's for `standard`). I'd like to discuss the naming again.

Right now I'm not convinced of `sortedPairKeccak256` either. The reason is that I think `Pair` is redundant. I feel we should agree on a naming convention for future additions to the Hashes library, and the distinction between sorting a pair or a single value should come from the function signature.

For example, for a `sortedKeccak256`, it'd be:

```solidity
function sortedKeccak256(bytes32 a, bytes32 b) internal pure returns (bytes32) {
    return a < b ? _efficientKeccak256(a, b) : _efficientKeccak256(b, a);
}

function sortedKeccak256(bytes32[] memory ptr) {
  ...
}
```

I'd say users would need to decide three things: the algorithm, the sorting (for multiproofs), and some particularity of the algorithm. In this case, it's just `keccak256` with sorting, so it's fine if we call it `sortedKeccak256` but I have a couple of questions:

- Would it make sense an inverse sorted hashing function at all? For example, we could have `sortedKeccak256(bytes32 a, bytes32 b, bool asc)`
- Furthermore, would it make sense to have a function with signature `keccak256(bytes32, bytes32, Sorting)` assuming `keccak256` is not creating a conflict? In this case, `Sorting` would be either `None`, `Asc`, or `Desc`.

So far I'm feeling the best naming convention we can use would be using the name of the algorithm and that's it, but then I'd reconsider calling it `stdHash` to avoid the naming conflict.

I know the naming convention and variants would make it difficult for a user to decide what to use but we can wrap the common use cases:

```solidity
function stdHash(bytes32 a, bytes32 b) {} // Default to sorted for multiproofs
function stdHash(bytes32 a, bytes32 b, Sorting sorting) {}
function poseidon(bytes32 a) {}
function poseidon(bytes32 a, bytes32 b) {} // Default to sorted for multiproofs
functon posiedon(bytes32 a, bytes32 b, Sorting sorting) {}
```

The main issue I see with this is the gas overhead of each hashing operation since it'd be evaluating the `sorting` each time, perhaps unnecessarily, but I think this is cleaner.

Sorry for bringing it up again, just want to make sure we release an API that behaves well over time, he. What do you think @Amxx?

contracts/utils/cryptography/Hashes.sol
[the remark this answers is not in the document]
ernestognw: I shared some extra thoughts [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3617#issuecomment-1977731019). I wasn't convinced of either so I'm open to chat about this again if we can agree on a better API for future additions.

contracts/utils/structs/MerkleTree.sol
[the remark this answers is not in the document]
ernestognw: Agree that the bottom line is that we shouldn't worry about removals now. I was trying to come up with the simplest representation of the tree but I think this is good enough as an abstraction of a `Bytes32PushTree`.

contracts/utils/structs/MerkleTree.sol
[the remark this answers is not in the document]
ernestognw: > Immutable (and constant) don't support that.

Yeah, and the possible combinations are too much to procedurally generate them. I think it's not worth pursuing this, but hey it was almost a good idea lol

> I've seen other zero used. In particular ones that are produced by doing hash("")

Would you mind elaborating on the use cases of this? Sounds interesting

contracts/utils/structs/MerkleTree.sol
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * @dev Insert a new leaf in the tree, and compute the new root. Returns the position of the inserted leaf in the
+     * tree, and the resulting root.
+     *
+     * Hashing the leaf before calling this function is recommended as a protection against
+     * second pre-image attacks.
+     */
+    function push(Bytes32PushTree storage self, bytes32 leaf) internal returns (uint256, bytes32) {
```
ernestognw, line 99:
```suggestion
    function push(Bytes32PushTree storage self, bytes32 leaf) internal returns (uint256 leafIndex, bytes32 root) {
```

contracts/utils/cryptography/Hashes.sol
ernestognw: Right, we should do this after the javascript library is released.

test/utils/structs/MerkleTree.test.js
```diff
@@ -0,0 +1,102 @@
...
+    // empty tree
+    const zeroLeafs = Array.from({ length: 2 ** Number(DEPTH) }, () => ethers.ZeroHash);
+    const zeroTree = makeTree(zeroLeafs);
+
+    // tree with one element
+    const leafs = Array.from({ length: 2 ** Number(DEPTH) }, () => ethers.ZeroHash);
+    const hashedLeaf = hashLeaf((leafs[0] = generators.bytes32())); // fill first leaf and hash it
+    const tree = makeTree(leafs);
+
+    // root should that of zero tree
```
ernestognw, line 80:
```suggestion
    // root should be that of a zero tree
```

ernestognw (commented):
Almost there!

contracts/utils/cryptography/Hashes.sol
```diff
@@ -0,0 +1,27 @@
...
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Library of standard hash functions.
+ */
+library Hashes {
+    /**
+     * @dev Commutative Keccak256 hash of a sorted pair of bytes32. Frequently used when working with merkle proofs.
+     */
+    function sortedPairKeccak256(bytes32 a, bytes32 b) internal pure returns (bytes32) {
```
Amxx, line 12: In the merkle library, this is called `standardNodeHash`.

Issue #3535, opened 2022-07-07
Consider alternative design for token transfers and hooks


frangio:
The way to customize token transfers today is by overriding the hooks `_beforeTokenTransfer` and `_afterTokenTransfer`. This was designed this way because there are multiple related operations (mint, transfer, burn) that would generally be customized in a uniform way (e.g. "pause transfers" should also pause mints), and they are different functions all of which invoke the same hooks.

An alternative that we can consider is instead to have a single entry point for all of these operations, a `_transfer` function that mints if the "from" argument is 0, etc., as suggested in #3109 for ERC20 (the same applies to other tokens). Hooks can then be removed because customizing transfers can be done by overriding this single function instead of overriding the hooks.

songhobby:
In our `ERC1155` contract, we have a single `TokenTransfer` function which is called by `safeTransferFrom`, `safeBatchTransferFrom`, `mintBatch` and `burnBatch`. For batch operations, the `TokenTransfer` is called inside of the loop for `tokenIds`. For minting and burning, we set the `from` and `to` to `address(0)`

frangio:
This is implemented in `next-v5.0` (#3838, #3921, #3876). There was an attempt to implement it for ERC721 (https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3893) but we ran into issues with the design summarized by @Amxx in [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3893#issuecomment-1382051919):

> Here is a summary of the identified issues that would need to be fixed:
> 
> - `_update` does not check the `from`. This means that a dev could call the `_update` function with an invalid from, and cause:
> 	-  an invalid event to be emitted
> 	- corruption of the `balances` mapping.
> - `_update` does different operation depending on the batchSize:
> 	- if the batchSize is 1, it will update the ownership, reset the approval, update the balances, and emit an event.
> 	- if the batchSize is not 0, it will just update the balance, but not do the transfer.
> This is error prone, because the dev have to understand that this function will do two very different thinks depending on this "small" parameter. It is possible that devs call this with a batchsize > 1 hoping that it will transfer a batch of token. This argues against having a single `_update` for both single and batched operations.
> - Cleanly support batchSize = 0 in extensions. 
[...]

PR #3636, opened 2022-08-19
[next-v5.0] Refactor ERC721 _transfer function to support minting and burning

Amxx:
Part of https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3535

Also remove before and after transfer hooks

Replace #3635

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -190,34 +190,6 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
     }
 
     /**
-     * @dev Safely transfers `tokenId` token from `from` to `to`, checking first that contract recipients
-     * are aware of the ERC721 protocol to prevent tokens from being forever locked.
-     *
-     * `data` is additional data, it has no specified format and it is sent in call to `to`.
-     *
-     * This internal function is equivalent to {safeTransferFrom}, and can be used to e.g.
-     * implement alternative mechanisms to perform token transfer, such as signature-based.
-     *
-     * Requirements:
-     *
-     * - `from` cannot be the zero address.
-     * - `to` cannot be the zero address.
-     * - `tokenId` token must exist and be owned by `from`.
-     * - If `to` refers to a smart contract, it must implement {IERC721Receiver-onERC721Received}, which is called upon a safe transfer.
-     *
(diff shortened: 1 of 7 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -234,103 +241,11 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
...
-     * WARNING: Usage of this method is discouraged, use {_safeMint} whenever possible
-     *
-     * Requirements:
-     *
-     * - `tokenId` must not exist.
-     * - `to` cannot be the zero address.
-     *
-     * Emits a {Transfer} event.
-     */
-    function _mint(address to, uint256 tokenId) internal virtual {
```
frangio, line 279: I don't think we should remove `_mint`, `_burn`, etc. It's a lot nicer and logical to write `_mint(...)` than `_transfer(address(0), ...)`.

I realize this introduces serious concerns with overrides so these shorthand functions should not be virtual and people should be directed to override `_transfer` instead.

frangio:
We should do the same with ERC20.

contracts/token/ERC721/ERC721.sol
Amxx: +1 for the "keep them in a non virtual way"

Amxx:
> We should do the same with ERC20 and 1155.

yes, but in other PRs :)

contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -48,12 +48,16 @@ abstract contract ERC721URIStorage is ERC721 {
...
-    function _burn(uint256 tokenId) internal virtual override {
-        super._burn(tokenId);
+    function _transfer(
+        address from,
+        address to,
+        uint256 tokenId
+    ) internal virtual override {
+        super._transfer(from, to, tokenId);
 
         if (bytes(_tokenURIs[tokenId]).length != 0) {
```
frangio, line 62: This function should only do this if `to` is 0!

frangio:
Superceded by https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3893.

PR #3638, opened 2022-08-19
[next-v5.0] Refactor voting, adding a Nonce contract

Amxx:
Replaces #3471

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


JulissaDantes:
This work will continue in #3816

k06a:
What do you think about implementing Nonces as `library` and using different instances for permits and votes and anything else?

PR #3639, opened 2022-08-19
Inherit asset decimals in ERC4626

### LINKED ISSUE #3549: Reduce reliance on underlying decimals in ERC4626
4626 uses `_asset.decimals()` to calculate initial share/asset conversions

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol#L167

EIP4626 states:

> All ERC-4626 tokenized Vaults MUST implement ERC-20’s optional metadata extensions.

But EIP20 states for `decimals`:

> OPTIONAL - This method can be used to improve usability, but interfaces and other contracts MUST NOT expect these values to be present.

So even if the vault exposes `decimals` for shares, the 20 spec says not to expect assets to expose `decimals`.

Even if both tokens expose `decimals` the method is for usability, not for onchain calculations.

Maybe a virtual `_initialShareRatio()` function could be exposed that implementing contracts can override without needing to override the entire conversion process

Amxx:
Inherit the decimal value from the asset (if any) while still supporting customization on top of ERC4626

Fixes #3549

#### PR Checklist

- [x] Tests
- [x] Changelog entry


Amxx:
The current tests already check decimal overriding.

We don't have a test where the fallback is done. Will work on that.

Amxx:
Testing should be good now

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -153,7 +162,7 @@ abstract contract ERC4626 is ERC20, IERC4626 {
         uint256 supply = totalSupply();
         return
             (assets == 0 || supply == 0)
-                ? assets.mulDiv(10**decimals(), 10**_asset.decimals(), rounding)
+                ? assets.mulDiv(10**decimals(), 10**ERC4626.decimals(), rounding)
```
frangio, line 165: To avoid confusion around inheritance, I would define a private function `_assetDecimals()` then use that here and as the default implementation of `decimals()`.

contracts/mocks/ERC4626Mock.sol
```diff
@@ -20,3 +19,20 @@ contract ERC4626Mock is ERC4626 {
...
+contract ERC4626DecimalMock is ERC4626Mock {
+    uint8 private immutable _decimals;
+
+    constructor(
+        IERC20Metadata asset,
+        string memory name,
+        string memory symbol,
+        uint8 decimaloverride
+    ) ERC4626Mock(asset, name, symbol) {
+        _decimals = decimaloverride;
```
frangio, line 32:
```suggestion
        uint8 decimalsOverride
    ) ERC4626Mock(asset, name, symbol) {
        _decimals = decimalsOverride;
```

frangio (changes requested):

frangio:
This is an improvement over the status quo because it will work for tokens with no decimals. However, one of the points raised in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3549 was:

> Even if both tokens expose `decimals` the method is for usability, not for onchain calculations.

I find this pretty compelling, but to stop using decimals in the conversion functions would be a breaking change now...

From https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3549#issuecomment-1186125815:
>What are the _initialConvertToXxx for. They make the code more complexe, and I don't see a reason for that.

The reason for those functions was to define conversion without reference to the decimals values.

Amxx:
Initial conversion function added. They are tests trough our current test suite (that checks decimals mismatch).

Need a breaking change description

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -153,19 +162,43 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
+    /**
+     * @dev Internal conversion function (from assets to shares) to apply when the vault is empty.
+     */
+    function _initialConvertToShares(uint256 assets, Math.Rounding rounding)
+        internal
+        view
+        virtual
+        returns (uint256 shares)
+    {
+        rounding;
```
frangio, line 178: Is there a compiler error if we remove the name from the argument list?

contracts/mocks/ERC4626Mock.sol
```diff
@@ -20,3 +19,42 @@ contract ERC4626Mock is ERC4626 {
...
+    }
+
+    function _initialConvertToShares(uint256 assets, Math.Rounding rounding)
+        internal
+        view
+        virtual
+        override
+        returns (uint256 shares)
+    {
+        return assets.mulDiv(10**decimals(), 10**super.decimals(), rounding);
```
frangio, line 48: Can we simplify it like this just to keep the test setup simpler

```suggestion
        return assets.mulDiv(10**_decimals, 1e18, rounding);
```

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: no, but then people might miss what the argument is about.

What do you feel about 
```
function _initialConvertToShares(uint256 assets, Math.Rounding /* rounding */)
```
?

contracts/mocks/ERC4626Mock.sol
Amxx: I like the fact that it is generic. Maybe people will ask us how to override to keep the old behavior and we can point to that ?

Also, the `1e18` is wrong ... it should be the tokens decimals, which in our test is not 18

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -153,19 +162,43 @@ abstract contract ERC4626 is ERC20, IERC4626 {
         uint256 supply = totalSupply();
         return
             (assets == 0 || supply == 0)
-                ? assets.mulDiv(10**decimals(), 10**_asset.decimals(), rounding)
+                ? _initialConvertToShares(assets, rounding)
                 : assets.mulDiv(supply, totalAssets(), rounding);
     }
 
+    /**
+     * @dev Internal conversion function (from assets to shares) to apply when the vault is empty.
```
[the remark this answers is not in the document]
Amxx: addressed

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -26,15 +26,24 @@ import "../../../utils/math/Math.sol";
...
      * @dev Set the underlying asset contract. This must be an ERC20-compatible contract (ERC20 or ERC777).
      */
-    constructor(IERC20Metadata asset_) {
+    constructor(IERC20 asset_) {
         _asset = asset_;
     }
 
+    /** @dev See {IERC4626-asset}. */
+    function decimals() public view virtual override(IERC20Metadata, ERC20) returns (uint8) {
+        try IERC20Metadata(address(_asset)).decimals() returns (uint8 value) {
```
frangio, line 40: I have some serious doubts about making this dynamic. I feel it introduces attack vectors. What do you think about reading the asset decimals in the constructor and storing it as an immutable?

frangio (changes requested):

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: That would be better for the vanilla version, but worst for the the upgradable version :/

Amxx: Attack vector seems limited to the underlying asset, which can already attack the vault anyway. I don't see any context where the vault would not trust the underlying token contract.

frangio (approved):

frangio:
I changed the variable name in the constructor which was `assetDecimals`, because in the `catch` branch those are expressly not the asset's decimals. If you agree with this let's merge.

frangio:
@alexflorence This is a good point we didn't notice. It's solved by calling the initializer though in a function like this:

```solidity
function migrateToV48() public reinitializer(2) {
    __ERC4626_init(asset);
}
```

We should add a note in the changelog for this.

We can consider defaulting to `super.decimals` if the stored decimals are 0 too.

@alexflorence Let us know your thoughts.

PR #3640, opened 2022-08-19
Optimize Clones assembly

Vectorized:
Changes:

- Deployments use the scratch space.
- Optimize `predictDeterministicAddress` to use lesser operations, while maintaining the same memory expansion costs.

Gas Before:
```
·------------------------------------------------------------|---------------------------|-------------|-----------------------------·
|                    Solc version: 0.8.13                    ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 10000000 gas  │
·····························································|···························|·············|······························
|  Methods                                                                                                                           │
···············|·············································|·············|·············|·············|···············|··············
|  Contract    ·  Method                                     ·  Min        ·  Max        ·  Avg        ·  # calls      ·  usd (avg)  │
···············|·············································|·············|·············|·············|···············|··············
|  ClonesMock  ·  clone(address,bytes)                       ·      90724  ·      97783  ·      93136  ·           12  ·          -  │
···············|·············································|·············|·············|·············|···············|··············
|  ClonesMock  ·  cloneDeterministic(address,bytes32,bytes)  ·      65269  ·      98378  ·      89662  ·           14  ·          -  │
···············|·············································|·············|·············|·············|···············|··············
|  Deployments                                               ·                                         ·  % of limit   ·             │
·····························································|·············|·············|·············|···············|··············
|  ClonesMock                                                ·          -  ·          -  ·     435385  ·        4.4 %  ·          -  │
·····························································|·············|·············|·············|···············|··············
|  DummyImplementation                                       ·          -  ·          -  ·     398946  ·          4 %  ·          -  │
·------------------------------------------------------------|-------------|-------------|-------------|---------------|-------------·
```

Gas After:
```
·------------------------------------------------------------|---------------------------|-------------|-----------------------------·
|                    Solc version: 0.8.13                    ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 10000000 gas  │
·····························································|···························|·············|······························
|  Methods                                                                                                                           │
···············|·············································|·············|·············|·············|···············|··············
|  Contract    ·  Method                                     ·  Min        ·  Max        ·  Avg        ·  # calls      ·  usd (avg)  │
···············|·············································|·············|·············|·············|···············|··············
|  ClonesMock  ·  clone(address,bytes)                       ·      90707  ·      97766  ·      93119  ·           12  ·          -  │
···············|·············································|·············|·············|·············|···············|··············
|  ClonesMock  ·  cloneDeterministic(address,bytes32,bytes)  ·      65237  ·      98361  ·      89645  ·           14  ·          -  │
···············|·············································|·············|·············|·············|···············|··············
|  Deployments                                               ·                                         ·  % of limit   ·             │
·····························································|·············|·············|·············|···············|··············
|  ClonesMock                                                ·          -  ·          -  ·     432690  ·        4.3 %  ·          -  │
·····························································|·············|·············|·············|···············|··············
|  DummyImplementation                                       ·          -  ·          -  ·     398946  ·          4 %  ·          -  │
·------------------------------------------------------------|-------------|-------------|-------------|---------------|-------------·
```

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -27,9 +27,9 @@ library Clones {
         assembly {
             // Cleans the upper 96 bits of the `implementation` word, then packs the first 3 bytes
             // of the `implementation` address with the bytecode before the address.
-            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(232, shl(96, implementation))))
+            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(0xe8, shl(0x60, implementation))))
             // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
-            mstore(0x20, or(shl(120, implementation), 0x5af43d82803e903d91602b57fd5bf3))
+            mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3))
             instance := create(0, 0x09, 0x37)
         }
         require(instance != address(0), "ERC1167: create failed");
@@ -47,9 +47,9 @@ library Clones {
         assembly {
             // Cleans the upper 96 bits of the `implementation` word, then packs the first 3 bytes
             // of the `implementation` address with the bytecode before the address.
-            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(232, shl(96, implementation))))
+            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(0xe8, shl(0x60, implementation))))
             // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
-            mstore(0x20, or(shl(120, implementation), 0x5af43d82803e903d91602b57fd5bf3))
+            mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3))
             instance := create2(0, 0x09, 0x37, salt)
         }
         require(instance != address(0), "ERC1167: create2 failed");
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -25,11 +25,12 @@ library Clones {
     function clone(address implementation) internal returns (address instance) {
         /// @solidity memory-safe-assembly
         assembly {
-             // Packs the size 3 bytes of `implementation` with the bytecode before the address. 
-            mstore(0x00, or(shr(0x88, implementation), 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000))
-             // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.             
-             mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3)) 
-            instance := create(0, 0x00, 0x37)
+            // Cleans the upper 96 bits of the `implementation` word, then packs the first 3 bytes
+            // of the `implementation` address with the bytecode before the address.
+            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(0xe8, shl(0x60, implementation))))
+            // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
+            mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3))
+            instance := create(0, 0x09, 0x37)
         }
         require(instance != address(0), "ERC1167: create failed");
     }
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -25,11 +25,9 @@ library Clones {
     function clone(address implementation) internal returns (address instance) {
         /// @solidity memory-safe-assembly
         assembly {
-            let ptr := mload(0x40)
-            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
-            mstore(add(ptr, 0x14), shl(0x60, implementation))
-            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
-            instance := create(0, ptr, 0x37)
+            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(232, shl(96, implementation))))
+            mstore(0x20, or(0x5af43d82803e903d91602b57fd5bf3, shl(120, implementation)))
+            instance := create(0, 0x09, 0x37)
         }
         require(instance != address(0), "ERC1167: create failed");
     }
@@ -44,11 +42,9 @@ library Clones {
     function cloneDeterministic(address implementation, bytes32 salt) internal returns (address instance) {
         /// @solidity memory-safe-assembly
         assembly {
-            let ptr := mload(0x40)
-            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
-            mstore(add(ptr, 0x14), shl(0x60, implementation))
-            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
-            instance := create2(0, ptr, 0x37, salt)
+            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(232, shl(96, implementation))))
+            mstore(0x20, or(0x5af43d82803e903d91602b57fd5bf3, shl(120, implementation)))
+            instance := create2(0, 0x09, 0x37, salt)
         }
         require(instance != address(0), "ERC1167: create2 failed");
     }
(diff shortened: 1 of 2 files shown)


Vectorized:
@frangio Used back the original implementation for the `predictDeterministicAddress`.

Saves a tiny bit of gas during deployment makes the deployer bytecode a little smaller.

contracts/proxy/Clones.sol
```diff
@@ -64,13 +66,13 @@ library Clones {
...
-            mstore(add(ptr, 0x4c), salt)
-            mstore(add(ptr, 0x6c), keccak256(ptr, 0x37))
-            predicted := keccak256(add(ptr, 0x37), 0x55)
+            mstore(add(ptr, 0x38), deployer)
+            mstore(add(ptr, 0x24), 0x5af43d82803e903d91602b57fd5bf3ff)
+            mstore(add(ptr, 0x14), implementation)
+            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73)
+            mstore(add(ptr, 0x58), salt)
+            mstore(add(ptr, 0x78), keccak256(add(ptr, 0x0c), 0x37))
+            predicted := keccak256(add(ptr, 0x43), 0x55)
```
frangio, line 75: What do you think of using the scratch space to hash the init code and then free memory for the rest of the calculation? Should result in less memory expansion.

frangio:
For the record, I looked at defining a helper like below that we could reuse in all places. Unfortunately Solidity doesn't inline the call so I don't think it's worth it.

```solidity
    function _getInitCode(address implementation) internal pure returns (bytes32 w0, bytes32 w1) {
        assembly {
            w0 := or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(0xe8, shl(0x60, implementation)))
            // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
            w1 := or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3)
        }
    }
```

contracts/proxy/Clones.sol
Vectorized: Sounds like a good idea, i’ll try it later in the day.

Vectorized: Not worth it. Runtime gas `65237` -> `65249`. Deployment gas `432690` -> `435253`.

Much of the magic of the original code is due to using `0x5af43d82803e903d91602b57fd5bf3ff` to store both the init code and the prefix in a single operation.

Amxx:
The reusability of such a function is great though

frangio (approved):
LGTM

Vectorized:
Resolved the `CHANGELOG.md` conflict.

frangio:
For reference here's the memory layout in `predictDeterministicAddress`. Rows are written onto memory from top to bottom.

```
0                               1                               2                               3                               4                               5                               6                               7                               8                               9
0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 8 9 a b c d e f 0 1 2 3 4 5 6 7 
                                                                                                                ~~~~~~~~~~~~~~~~~~~~~~~~dddddddddddddddddddddddddddddddddddddddd
                                                                        000000000000000000000000000000005af43d82803e903d91602b57fd5bf3ff
                                        ~~~~~~~~~~~~~~~~~~~~~~~~iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii
0000000000000000000000003d602d80600a3d3981f3363d3d373d3d3d363d73
                                                                                                                                                                                ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss
                                                                                                                                                                                                                                                hhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

                        [                                                                                                            ][                                                                                                                                                                        ]
```

frangio (approved):

contracts/proxy/Clones.sol
```diff
@@ -25,11 +25,12 @@ library Clones {
...
         /// @solidity memory-safe-assembly
         assembly {
-            let ptr := mload(0x40)
-            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
-            mstore(add(ptr, 0x14), shl(0x60, implementation))
-            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
-            instance := create(0, ptr, 0x37)
+            // Cleans the upper 96 bits of the `implementation` word, then packs the first 3 bytes
+            // of the `implementation` address with the bytecode before the address.
+            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(0xe8, shl(0x60, implementation))))
```
Amxx, line 30: I think this can be optimized:

```suggestion
            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(0x88, implementation)))
```

contracts/proxy/Clones.sol
```diff
@@ -25,11 +25,12 @@ library Clones {
...
+            // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
+            mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3))
+            instance := create(0, 0x09, 0x37)
```
Amxx, line 33: I'm not really sure why you are aligning to the end, and then starting the instance at 0x09. I would rather start at 0x00 like thid.
```suggestion
            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000, shr(0x40, input)))
            mstore(0x20, or(0x00000000000000005af43d82803e903d91602b57fd5bf3000000000000000000, shl(0xc0, input)))
            instance := create(0, 0x00, 0x37)
```

contracts/proxy/Clones.sol
```diff
@@ -44,11 +45,12 @@ library Clones {
...
-            mstore(ptr, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
-            mstore(add(ptr, 0x14), shl(0x60, implementation))
-            mstore(add(ptr, 0x28), 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
-            instance := create2(0, ptr, 0x37, salt)
+            // Cleans the upper 96 bits of the `implementation` word, then packs the first 3 bytes
+            // of the `implementation` address with the bytecode before the address.
+            mstore(0x00, or(0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000, shr(0xe8, shl(0x60, implementation))))
+            // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
+            mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3))
+            instance := create2(0, 0x09, 0x37, salt)
```
Amxx, line 53: same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3640/files#r953506189

contracts/proxy/Clones.sol
Vectorized: It’s to reduce the number of trailing zeroes in the constants. Smaller bytecode size.

Vectorized: If the optimizer runs is low, the compiler may sometimes choose to break up constants with many trailing zeroes into multiple operations.

Amxx: ok, then we can do a different change optimization. (just updated the code above)

Vectorized: The `shl` and `shr` is for cleaning the upper 96 bits. 

https://twitter.com/optimizoor/status/1560590413768994816?s=21&t=gSQglMPV2vg0zdLFwmzedA

Vectorized:
Brb, i’m outside right now.

contracts/proxy/Clones.sol
Amxx: Hooo, right. I was assuming the address in clean

Amxx (approved):

LuKks:
@Vectorized Any hints on how to create the equivalent of `predictDeterministicAddress` but on JavaScript? For example, having all the data needed pre-loaded locally (deployer addr, implementation addr, contract bytecode, etc) to later being able to calculate addresses on pure JS?

I tried a lot but nothing seems to result in the correct address, maybe the optimizations and using the proxy prefix + suffix bytes makes it not straightforward to replicate it on JavaScript?

References:
https://stackoverflow.com/questions/71573083/im-trying-to-implement-openzeppelins-minimal-proxy-clone-contract-on-tron-bloc/71581522#71581522
https://github.com/thegostep/solidity-create2-deployer

Thank you in advance! :)

Amxx:
> @Vectorized Any hints on how to create the equivalent of `predictDeterministicAddress` but on JavaScript? For example, having all the data needed pre-loaded locally (deployer addr, implementation addr, contract bytecode, etc) to later being able to calculate addresses on pure JS?
> 
> I tried a lot but nothing seems to result in the correct address, maybe the optimizations and using the proxy prefix + suffix bytes makes it not straightforward to replicate it on JavaScript?
> 
> References: https://stackoverflow.com/questions/71573083/im-trying-to-implement-openzeppelins-minimal-proxy-clone-contract-on-tron-bloc/71581522#71581522 https://github.com/thegostep/solidity-create2-deployer
> 
> Thank you in advance! :)

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/helpers/create2.js

LuKks:
@Amxx Thanks, I ended up doing a tiny JS library to solve the issue so I don't have to deal with it again! Haha
https://github.com/LuKks/predict-deterministic-address

So that allowed me to finish this other one:
https://github.com/LuKks/wallet-accounts (wallet with sub-accounts as a smart contract)

PR #3646, opened 2022-08-25
Add Buyable.sol

aglawson:
This change introduces a new library contract: Buyable.sol. 
Buyable inherits from Ownable, and allows a contract owner to sell ownership of their contract. 

#### PR Checklist

- [X ] Tests
- [X ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/access/Buyable.sol
+++ b/contracts/access/Buyable.sol
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "./Ownable.sol";
+
+abstract contract Buyable is Ownable {
+    address private _originalOwner;
+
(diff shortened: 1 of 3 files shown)


Amxx:
Sorry @aglawson, but I don't think this feature is something we want to merge.

It is very opinionated (for example through the minimum price), and doesn't correspond match any standard. Also, what if you want to pay with something that is not eth, such as an ERC20.
IMO it would be way more effective to make the ownability an NFT, just like is done [here](https://github.com/Amxx/NonFungibleFinance/blob/master/solidity/contracts/VestingFactory.sol), so that you can use the extensive NFT trading tools.

aglawson:
Thank you for your input, I will consider it and take a look at the linked file. 

Have a great day.

PR #3647, opened 2022-08-25
Improve recommendation on ERC721 transferFrom vs safeTransferFrom

### LINKED ISSUE #3477: Provide more nuance about ERC721's transferFrom vs safeTransferFrom
We have to mention the risk of reentrancy introduced by `safeTransferFrom`, and give more nuance to whatever we encourage or discourage.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/109778c17c7020618ea4e035efb9f0f9b82d43ca/contracts/token/ERC721/IERC721.sol#L84

goncaloMagalhaes:
Provide more nuance about ERC721's transferFrom vs safeTransferFrom
Fixes #3477

### DIFF
--- a/contracts/token/ERC721/IERC721.sol
+++ b/contracts/token/ERC721/IERC721.sol
@@ -82,8 +82,8 @@ interface IERC721 is IERC165 {
      * @dev Transfers `tokenId` token from `from` to `to`.
      *
      * WARNING: Note that the caller is responsible to confirm that the recipient is capable of receiving ERC721
-     * or else they may be permanently lost. Usage of {safeTransferFrom} is encouraged, though the caller must understand
-     * this adds an external call to {transferFrom}, potentially creating a reentrancy vulnerability.
+     * or else they may be permanently lost. Usage of {safeTransferFrom} prevents loss, though the caller must 
+     * understand this adds an external call which potentially creates a reentrancy vulnerability.
      *
      * Requirements:
      *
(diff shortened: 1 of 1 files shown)


contracts/token/ERC721/IERC721.sol
```diff
@@ -81,7 +81,9 @@ interface IERC721 is IERC165 {
     /**
      * @dev Transfers `tokenId` token from `from` to `to`.
      *
-     * WARNING: Usage of this method is discouraged, use {safeTransferFrom} whenever possible.
+     * WARNING: Note that the caller is responsible to confirm that the recipient is capable of receiving ERC721
+     * or else they may be permanently lost. Usage of {safeTransferFrom} is encouraged, though the caller must understand
```
frangio, line 85: Let's avoid "encouraged".

```suggestion
     * or else they may be permanently lost. Usage of {safeTransferFrom} prevents loss, though the caller must understand
```

contracts/token/ERC721/IERC721.sol
```diff
@@ -81,7 +81,9 @@ interface IERC721 is IERC165 {
...
+     * this adds an external call to {transferFrom}, potentially creating a reentrancy vulnerability.
```
frangio, line 86: Current wording seems to imply that there is a call to `transferFrom`.

```suggestion
     * this adds an external call which potentially creating a reentrancy vulnerability.
```

frangio (commented):
Thank you @goncaloMagalhaes!

contracts/token/ERC721/IERC721.sol
goncaloMagalhaes: You're right! I'll commit the changes addressing your feedback :)

frangio (dismissed):
Thank you!

frangio (approved):

PR #3650, opened 2022-08-26
Remove unnecessary virtual

fulldecent:
This PR removes unnecessary `virtual` in all mock contracts where those mocks are not intended to be subclassed.

This reduces unnecessary surface area of those mocks and also reduces the risk of somebody seeing those mocks and thinking they are intended to be subclassed.


frangio:
Thank you @fulldecent. Our automated changelog check is a little simple and this was a false positive since it's not a user facing change. Can you remove the changelog entry?

fulldecent:
@frangio cool, all done, other checks passed

frangio (approved):
Thanks!

PR #3660, opened 2022-08-30
Add Infix Notation Expression Evaluation Utility

njrapidinnovation:
Add Infix Notation Expression Evaluation Utility.

Provides method that can be used to evalute any Infix Notation Expression which 
 *     consist of 1 and 0 as operands .
 *     consist of a (&&) , o (||) , ( , ) as operators.
 *     precedence is also considered for LOGICAL AND (&&) and LOGICAL OR (||)

This PR adds infix expression evaluation utility function. The EvalInfix library provides utility function to evaluate an Infix Expression. It is really usefull when we have to check multiple boolean conditions in the form of an infix expression
like (TRUE or (FALSE or TRUE)).

### DIFF
--- a/contracts/mocks/EvalInfixMock.sol
+++ b/contracts/mocks/EvalInfixMock.sol
@@ -5,22 +5,27 @@ pragma solidity ^0.8.0;
 import "../utils/EvalInfix.sol";
 
 contract EvalInfixMock {
-
-    function getResult() external pure returns(bool){
-        return (true || false && false);  //straight -> false , precedence -> true
+    function getResult() external pure returns (bool) {
+        return (true || (false && false)); //straight -> false , precedence -> true
     }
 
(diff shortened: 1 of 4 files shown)


contracts/mocks/EvalInfixMock.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "../utils/EvalInfix.sol";
+
+contract EvalInfixMock {
+    function getResult() external pure returns (bool) {
+        return (true || (false && false)); //straight -> false , precedence -> true
```
Amxx, line 9: I'm genuinely confused 😕 

What is that function supposed to test ?

Amxx:
We need to first discuss that in the issues, to see if that is even something we need. 

Then, and only then, we can discuss implementation. but for sure `string[]` is not the right approach!

PR #3665, opened 2022-08-31
Make VestingWallet payable to allow funding at construction

Amxx:
If `PaymentSplitter`'s constructor is payable, then so should `VestingWallet`'s

#### PR Checklist

- [ ] Tests
- [x] Changelog entry


frangio (approved):
LGTM

PR #3666, opened 2022-08-31
Use hardhat-exposed to reduce the need for mocks

Amxx:
#### PR Checklist

- [x] Tests


frangio:
Inspected the above postinstall scripts and they're safe.

frangio:
I would like us to review the "mock" name and switch to a new one. Maybe we can just rename everything to `Test` instead of `Mock`.

See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1320

pcaversaccio:
FYI, just don't underestimate this breaking change - it will break many of my unit tests since I pull some of your mocks via submodules. So once you merge this PR to `master` and I update the submodules it will break (i.e. before you make a new release). Just wanted to quickly make you aware of this.

frangio (dismissed):
Ready to merge this.

contracts/utils/Base64.sol
```diff
@@ -12,7 +12,7 @@ library Base64 {
     /**
      * @dev Base64 Encoding/Decoding Table
      */
-    string internal constant _TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+    string private constant _TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
```
frangio, line 15: Actually, I realized I had this pending comment. We should do this in a separate PR if we want to do it.

```suggestion
    string internal constant _TABLE = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
```

contracts/mocks/ERC20Mock.sol
```diff
@@ -1,33 +1,16 @@
 // SPDX-License-Identifier: MIT
-
 pragma solidity ^0.8.0;
 
-import "../token/ERC20/ERC20.sol";
+import {ERC20} from "../token/ERC20/ERC20.sol";
```
Amxx, line 4: I believe that is the first time we use this syntax. Is that something we want to be doing more ?

contracts/utils/Base64.sol
Amxx: ok, lets make another PR

Amxx:
LGTM

contracts/mocks/ERC20Mock.sol
frangio: Not sure. It wasn't a conscious decision, I think I went along with the syntax from `a16z/erc4626-tests`.

frangio (approved):

PR #3670, opened 2022-09-01
Add log2, log10 and log256 functions

Amxx:
Logs are already used in different places:

- in `Math.sqrt`, we do a log in base 4 (log2 divided by 2)
- in `String.toString(uint256)`, we do a log base 10
- in `String.toHexString(uint256)`, we do a log base 256

In #3573, it was mentioned that these operation should all be in math, rather then scattered all over the code base. This rationalizes the code, and allow reuse by devs.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -190,8 +190,10 @@ library Math {
      */
     function sqrt(uint256 a, Rounding rounding) internal pure returns (uint256) {
         uint256 result = sqrt(a);
-        if (rounding == Rounding.Up && result * result < a) {
-            result += 1;
+        unchecked {
+            if (rounding == Rounding.Up && result * result < a) {
+                result += 1;
+            }
         }
         return result;
     }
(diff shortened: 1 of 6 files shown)


contracts/utils/math/Math.sol
```diff
@@ -223,4 +195,146 @@ library Math {
...
+        return result;
+    }
+
+    /**
+     * @dev Return the log in base 10, rounded down, of a positive value.
+     * Returns 0 if given 0.
+     */
+    function log10(uint256 value) internal pure returns (uint256) {
+        uint256 result = 0;
+        if (value >= 10**64) {
```
Atarpara, line 257: @Amxx you can replace `>=` by `>` and save more gas.

Example: 
```
value >= 10**64 == value > 10**64-1
```

contracts/utils/math/Math.sol
```diff
@@ -167,35 +167,7 @@ library Math {
...
-            result <<= 4;
-        }
-        if (x >> 4 > 0) {
-            x >>= 4;
-            result <<= 2;
-        }
-        if (x >> 2 > 0) {
-            result <<= 1;
-        }
+        uint256 result = 1 << (log2(a) / 2);
```
Atarpara, line 170: We can replace `1 << (log2(a) / 2)` by `1 << (log2(a) >> 2)`.

Amxx: That is interresting, but the correct formula would be `1 << (log2(a) >> 1)`

Atarpara: Ohh, sorry. It >>1.

frangio: Before change that I would check if Solidity does not make that optimization on its own.

contracts/utils/math/Math.sol
Atarpara: @Amxx what about this one?

contracts/utils/math/Math.sol
Amxx: tested both. using `>> 1` saves 54 gas compared to ```/ 2```
(0.8.16 with optimizations)

contracts/utils/math/Math.sol
Amxx: I tested both, and your version is more expensive :/ it looks like the compiler does the `-1` operation at runtime

frangio:
Shouldn't `log256` be `log16`? :thinking:

Amxx:
> ~Shouldn't `log256` be `log16`? thinking~
> 
> Edit: Just realized that `toHexString` indeed uses `log256`.

Yes, hex chars go by pairs.

contracts/utils/math/Math.sol
Atarpara: @Amxx Because It happen due to solidity checking underflow that. But When we are use shift operator it bypass that check.

Amxx: Overflow check make sens for +, - and * ... not for /
The only reasonable check is that we don't divide by 0 ... but there is no reason to check that here.

frangio (approved):
LGTM! Looking forward to fuzzing too.

PR #3672, opened 2022-09-01
fix: detailed explanation of what a function does

Chiadikaobixo:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -30,7 +30,7 @@ abstract contract Ownable is Context {
     }
 
     /**
-     * @dev Throws if called by any account other than the owner.
+     * @dev Throws error if called by any account other than the owner.
      */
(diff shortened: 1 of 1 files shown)


frangio:
@Chiadikaobixo Are you a human? Otherwise we will block the account from further contributions due to spam.

Chiadikaobixo:
Yes I’m human. I’m finding it difficult for my changelog to pass while
creating a pull request.

frangio:
The changelog won't pass unless you modify the changelog, but you don't need to do that for every PR. This change you're trying to make would not need a changelog. That said, I am not sure this change would be accepted.

Chiadikaobixo:
Thanks very much, this is my first attempt on opensource and I am glad to know this. I look forward to contributing soon.

Issue #3460, opened 2022-06-10
Request for the add forge testing for the openzeppelin contracts


Atarpara:
**🧐 Motivation**

In today's Foundry is one of the fast and most powerfull tools for testing smart contracts. 

**📝 Details**

Benefit of the use Foundry 
1. Foundry provides In-Built Fuzz Testing Facility. 
2. No Need to learn the Java Scripts or Chai Library.
3. Simultaneously Running the Multiple Test Cases we can save a lot of time in testing (Long test case file).
4. Gas Reporting Function Wise
5. If something goes wrong we can trace the entire call

So My Question is Openzeppelin-Team Should ready to adopt the foundry? If Yes, So many Devs are happy to contribute to that.

frangio:
This project has a huge existing test suite in JavaScript and migrating it to something else is a huge effort and risk. There are many incremental improvements we can make to the existing tests for whatever issues we see with them. At the moment I see no strong reasons to migrate.

Fuzzing is definitely interesting, but we should start experimenting with it before attempting to migrate the entire project.

Amxx:
Hello @Atarpara 

We know about Foundry and its features. We would definitely love to benefit from them...
... however, the amount of work required to rewrite our tests in solidity is just to much for us to do this transition in the near future.

We also believe that testing in solidity has real drawbacks, that the foundry team might understate. In particular, I don't think "No Need to learn the Java Scripts" is an improvement. Most devs will design contract front-ends using javascript/typescript as their base language. Writing the test in this language allows us to check contracts against the library that will, in the real world, be used to interact with them. Javascript also provides math functions that we can used as a reference for testing math written in solidity.

PR #3676, opened 2022-09-03
Add fuzz tests for Math.sqrt & Math.logX using Foundry

nventuro:
I wrote this while doing my own research on Foundry: feel free to ignore it (though I do think that the results are interesting in a use case like this one).

I had to rename one test function since it started with the `test` prefix, which is what forge looks for when filtering out test functions.

Other than that, everything seems fairly reasonable. This is the first Foundry test I've written, so there's likely much more idiomatic ways to achieve what I'm trying to do.


Ellajoke (commented):
Get me actived

Atarpara:
I opened one issue some time regarding foundry test add into openzeppelin contract. Issue #3460.

Amxx:
We are currently considering removing a lot of mocks in favour of the hardhat-exposed plugin. I'm not sure how we can make the two work together.

We would love to have fuzzing though

nventuro:
@Amxx these are not quite the same as the mocks that expose internal functions, since the tests themselves are written in Solidity. So I'm not sure those two efforts are related at all.

frangio:
I think the proposal here is to add fuzzing in addition to our current test suite. The math functions are probably a good target.

I don't think it conflicts with the mocks removal.

contracts/mocks/tests/MathTest.sol
```diff
@@ -0,0 +1,36 @@
...
+
+        // `sqrt` should return a rounded-down value, so `square` is either less or equal to `input`.
+        if (square < input) {
+            // We know `result` is less than the true value for the square root, but we want to check that the error is
+            // minimal, in this case, less than 1.
+            // This should only happen if `input` is not a perfect square. We can then check that any result larger than
+            // the one we got would yield a square larger than `input`, meaning we got the smallest value for which the
+            // square is less than `input`.
+
+            (bool noOverflow, uint256 nextSquare) = SafeMath.tryMul(result + 1, result + 1);
```
Amxx, line 23: if `input` is `type(uint256).max`, and if `result` is correct, 

Then we have `result * result <= input` and `SafeMath.tryMul(result + 1, result + 1)` that overflow. The overflow here is correct.

IMO the check should possibly be

```
uint256 result = Math.sqrt(input);
uint256 resultSquare = result * result;
(bool noOverflow, uint256 resultPlusOneSquare) = SafeMath.tryMul(result + 1, result + 1);
asserTrue(resultSquare <= input && (!noOverflow || input < resultPlusOneSquare));
```

What do you think ?

nventuro: Yup, that seems fully equivalent. I didn't try very hard to have this be idiomatic or even not very verbose. I don't even have strong thoughts on what a good Solidity test might look like - if you think this reads better I can change it to look like that. I do find it slightly odd to run both assertions in the perfect square scenario though.

contracts/mocks/tests/MathTest.sol
```diff
@@ -0,0 +1,36 @@
...
+            // does fit in 256 bits, this still means that `nextSquare` is larger than `input`.
+            vm.assume(noOverflow);
+
+            assertTrue(input < nextSquare);
+        } else {
+            // If `square` is not less than input, then it must be a perfect match (because `input` is a perfect
+            // square).
+            assertEq(square, input);
+        }
+    }
```
Amxx, line 35:
```suggestion
    function testSquareRootUp(uint256 input) public {
        _testSquareRoot(input, Math.Rounding.Up);
    }

    function testSquareRootDown(uint256 input) public {
        _testSquareRoot(input, Math.Rounding.Down);
    }

    function testSquareRootZero(uint256 input) public {
        _testSquareRoot(input, Math.Rounding.Zero);
    }

    function _testSquareRoot(uint256 input, Math.Rounding rounding) internal {
        uint256 result = Math.sqrt(input, rounding);
        (bool resultSquareNoOverflow, uint256 resultSquare) = SafeMath.tryMul(result, result);

        // square of result is bigger than input
        if (input < resultSquare || !resultSquareNoOverflow)
        {
            assertTrue(rounding == Math.Rounding.Up);
            assertTrue((result - 1) * (result - 1) < input);
        }
        // square of result is smaller than input
        else if (resultSquare < input)
        {
            // we must be rounding down, or toward zero
            assertFalse(rounding == Math.Rounding.Up);
            (bool resultPlusOneSquareNoOverflow, uint256 resultPlusOneSquare) = SafeMath.tryMul(result + 1, result + 1);
            assertTrue(input < resultPlusOneSquare || !resultPlusOneSquareNoOverflow);
        }
    }
```

Amxx:
This doesn't need a Changelog entry IMO. 
Ready for review.

holic:
Has anyone had issues using this repo as a forge dependency since forge is now used in this repo? Having a hard time updating my forge deps due to inner submodule changes of this repo.

frangio:
@holic I'm not sure what you mean by "inner submodule changes". Do you think there is something we need to fix?

holic:
> @holic I'm not sure what you mean by "inner submodule changes". Do you think there is something we need to fix?

I think that running `forge update` does a recursive submodule update. If I have a project with openzeppelin-contracts forge dependency (submodule) and I want to update it, running `forge update` not only updates openzeppelin-contracts, but also updates the inner forge dep/submodule inside this repo (lib/forge-std), which causes git to sort of freak out in terms of committing those changes.

Feels like we need some way to opt out of these deep updates? Or pin to a specific commit? Unclear if this is a forge issue or something to be fixed in this repo.

frangio:
@holic It looks like a forge issue to me. I wasn't able to consistently reproduce it either. Please report it to forge.

PR #3679, opened 2022-09-05
Optimize Math.max and SignedMath.max

gzliudan:
In Solidity, there is no single op-code for `<=` or `>=` expressions. Solidity compiler executes the LT/GT (less than/greater than) op-code and afterwards it executes an ISZERO op-code to check if the result of the previous comparison (LT/ GT) is zero and validate it or not. So the use of `<` and `>` are cheaper than `<=` and `>=`.  Below is current version of max function:

```solidity
    function max(uint256 a, uint256 b) internal pure returns (uint256) {
        return a >= b ? a : b;
    }
```

The above code works as below steps:

```solidity
    function max(uint256 a, uint256 b) internal pure returns (uint256) {
        if (a > b) {
            return a;
        } else if (a == b) {
            return a;
        } else {
            return b;
        }
    }
```

If we use the new version:

```solidity
    function max(uint256 a, uint256 b) internal pure returns (uint256) {
        return a > b ? a : b;
    }
```

Then the function will works as below steps:
```solidity
    function max(uint256 a, uint256 b) internal pure returns (uint256) {
        if (a > b) {
            return a;
        } else {
            return b;
        }
    }
```

So the step of check `a == b` is omitted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
  * `Arrays`: Add `unsafeAccess` functions that allow reading and writing to an element in a storage array bypassing Solidity's "out-of-bounds" check. ([#3589](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3589))
  * `Strings`: optimize `toString`. ([#3573](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3573))
  * `Ownable2Step`: extension of `Ownable` that makes the ownership transfers a two step process. ([#3620](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3620))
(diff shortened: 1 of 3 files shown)


Amxx:
Nice find @gzliudan 

I was hopping the compiler would translate `a >= b` as `!(a<b)` and you "just" pay for the negation part (with can be done using isZero).

Still, that is a nice optimization!

Can you add a small changelog entry ?

gzliudan:
Ok, I will add a entry to changelog.

Amxx (approved):

frangio (approved):
It's not even necessary to do the negation. The compiler should be able to convert `a >= b ? a : b` into `a < b ? b : a`... Indeed, compiling with `viaIR: true` this optimization is done automatically.

Reluctantly approving this PR because these peephole optimizations are the job of the compiler.

PR #3682, opened 2022-09-06
Remove Address.isContract

### LINKED ISSUE #3417: Remove isContract due to potential misuse
In spite of the warnings provided in the comments of the function itself, the name `isContract` is a misnomer and creates a potential security risk for anyone who doesn't bother to read the notes or someone who is reviewing a 3rd party contract that uses this fn.

There is a misconception that calling this function will return false if the address is an eoa.  This can lead to the inadvertant introduction of an exploit and other [risks already clearly identified in the comments](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/57725120581e27ec469e1c7e497a4008aafff818/contracts/utils/Address.sol#L15).  But comments inside the function are not enough in this case where the name of the fn is so blatantly misleading.

Propose changing the name of the function to `hasCode` which is much more descriptive of what the function does.  This should be a breaking change and may end up being a wake up call to anyone who has been misusing the fn to date.

ashwinYardi:
Fixes #3417 

- This PR removes isContract utility function from Address.sol library. 
- Wherever isContract was used in the other contracts, is replaced by inlining `address(...).code.length > 0`  
#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 ### Deprecations
 
  * `EIP712`: Added the file `EIP712.sol` and deprecated `draft-EIP712.sol` since the EIP is no longer a Draft. Developers are encouraged to update their imports. ([#3621](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3621))
+  * `Address.isContract`: `isContract` utilitiy function is deprecated from v5.0.0 because of its ambiguous nature. Reasons and discussions here: [#3417](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3417).
 
 ```diff
 -import "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
--- a/contracts/mocks/AddressImpl.sol
+++ b/contracts/mocks/AddressImpl.sol
@@ -9,10 +9,6 @@ contract AddressImpl {
 
     event CallReturnValue(string data);
 
-    function isContract(address account) external view returns (bool) {
-        return Address.isContract(account);
-    }
-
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 ### Deprecations
 
  * `EIP712`: Added the file `EIP712.sol` and deprecated `draft-EIP712.sol` since the EIP is no longer a Draft. Developers are encouraged to update their imports. ([#3621](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3621))
+  * `Address.isContract`: `isContract` utilitiy function is deprecated from v5.0.0 because of its ambiguous nature. Reasons and discussions here: [#3417](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3417).
```
frangio, line 24: Please put this under a new "Unreleased" section at the top of the file.

frangio, line 24:
```suggestion
  * `Address`: Removed `isContract` because of its ambiguous nature and potential for misuse. ([#3682](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3682))
```

frangio (commented):
Thanks!

The pragma in each of these files should be bumped to at least `^0.8.1` because prior to that Solidity would not use `extcodesize`.

Additionally, this is the removal that we want to do in 5.0, but we should also simply mark it as deprecated in the documentation prior to 5.0.

JulissaDantes:
This will be closed due to inactivity. This effort will continue on #3945 .

PR #3683, opened 2022-09-06
Run tests with viaIR

frangio:
Just testing, I don't intend for us to merge this.


frangio:
There seems to be a Hardhat bug.

Amxx:
I've seen that in the past. Its some king of timeout

PR #3690, opened 2022-09-07
init erc20 token name and symbol in constructor

### LINKED ISSUE #3689: ERC4626 should initialize ERC20 token in the constructor
The ERC4626 contract does initialize the inherited ERC20 contract in the constructor.

**💻 Environment**
MacOS 12.5.1 M1

v4.7.3
**📝 Details**
The ERC4626 contract inherited an ERC20 token. But the ERC4626 contract does not initialize this token with name and symbol.
It would be more intuitive and easy to understand if the token name and symbol are initialized together with the ERC4626 directly.

zoey-t:
Fixes #3689  

Added the ERC20 initialization with name and symbol into ERC4626's constructor.

See issue #3689 
It would make more sense and easy to understand if the token name and symbol are initialized together with the ERC4626 directly.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio:
See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3689#issuecomment-1239876564.

PR #3695, opened 2022-09-09
Ignore specific warnings and make the rest into errors

### LINKED ISSUE #3686: Emit compiler warnings as errors
We have had compiler warnings that we should pay attention to but we missed them due to a bunch of other warnings that not so relevant. We should configure Hardhat to emit warnings as errors. We should make it ignore code size warnings unless optimizations are enabled (in which case it requires attention).

Additionally we need a way to ignore warnings selectively in some cases. This is a longstanding missing feature in solc but we can make a Hardhat plugin for it.

frangio:
Fixes #3686 

- Ignores code size warnings if we're compiling without optimizations.
- Ignores unreachable code warnings caused by files in `contracts/mocks/unreachable`.
- Ignores unused parameter warnings that are emitted during coverage analysis (I guess because of coverage transforms).
- Promote the rest of the warnings into errors.
- Makes the revert in ERC721Enumerable._beforeConsecutiveTransfer conditional on the batch size in order to remove the warning about unreachable code that otherwise users will see when they combine with other modules like ERC721Pausable.


hardhat.config.js
```diff
@@ -75,6 +76,13 @@ module.exports = {
       viaIR: withOptimizations && argv.ir,
     },
   },
+  warnings: {
+    ignore: {
+      '**/*': {
```
Amxx, line 81: isn't `*` enough?

frangio: `*` matches `A.sol` but it doesn't match `c/d/A.sol`.

frangio: I updated the plugin to allow `*`.

hardhat/ignore-unreachable-warnings.js
```diff
@@ -0,0 +1,47 @@
...
+} = require('hardhat/builtin-tasks/task-names');
+
+const marker = Symbol('unreachable');
+const markedCache = new WeakMap();
+
+task(TASK_COMPILE_SOLIDITY_GET_COMPILATION_JOB_FOR_FILE, async (params, _, runSuper) => {
+  const job = await runSuper(params);
+  // If the file is in the unreachable directory, we make a copy of the config and mark it, which will cause it to get
+  // compiled separately (along with the other marked files).
+  if (params.file.sourceName.startsWith('contracts/mocks/unreachable')) {
```
Amxx, line 21: Putting all the mocks with unreachable code in the same folder feels strange.

Why not do something like `.include('unreachable')`, and have the possibility to place them in subfolders (or even put unreachable in the name/extension)

hardhat/ignore-unreachable-warnings.js
```diff
@@ -0,0 +1,47 @@
...
+// Thus, we need to handle these warnings separately. We force Hardhat to compile them in a separate compilation job and
+// then ignore the warnings about unreachable code that come from that compilation job.
+
+const { task } = require('hardhat/config');
+const {
+  TASK_COMPILE_SOLIDITY_GET_COMPILATION_JOB_FOR_FILE,
+  TASK_COMPILE_SOLIDITY_COMPILE,
+} = require('hardhat/builtin-tasks/task-names');
+
+const marker = Symbol('unreachable');
```
Amxx, line 14: Are we sure we want just one marker to be supported ? what about a list of `{ symbol: Symbol('unreachable'), code: '5740' }` like objects ?

hardhat/ignore-unreachable-warnings.js
```diff
@@ -0,0 +1,47 @@
...
+const W_UNREACHABLE_CODE = '5740';
+
+task(TASK_COMPILE_SOLIDITY_COMPILE, async (params, _, runSuper) => {
+  const marked = params.compilationJob.solidityConfig[marker];
+  const result = await runSuper(params);
+  if (marked) {
+    result.output = {
+      ...result.output,
+      errors: result.output.errors?.filter(
+        e => e.severity !== 'warning' || e.errorCode !== W_UNREACHABLE_CODE,
```
Amxx, line 42: is the `e.severity !== 'warning'` needed ?

hardhat/ignore-unreachable-warnings.js
```diff
@@ -0,0 +1,47 @@
...
+  // If the file is in the unreachable directory, we make a copy of the config and mark it, which will cause it to get
+  // compiled separately (along with the other marked files).
+  if (params.file.sourceName.startsWith('contracts/mocks/unreachable')) {
+    const originalConfig = job.solidityConfig;
+    let markedConfig = markedCache.get(originalConfig);
+    if (markedConfig === undefined) {
+      markedConfig = { ...originalConfig, [marker]: true };
+      markedCache.set(originalConfig, markedConfig);
+    }
+    job.solidityConfig = markedConfig;
```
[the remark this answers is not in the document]
frangio: It would work but it would be suboptimal. Hardhat groups compilations together based on reference equality of the config objects. Here, every invocation of `Object.assign` will return a new object and that file will be compiled separately.

hardhat/ignore-unreachable-warnings.js
frangio: It's a safety fallback so that we don't accidentally filter out errors.

hardhat/ignore-unreachable-warnings.js
frangio: It feels like premature optimization. Every other warning is properly handled by `hardhat-ignore-warnings`.

frangio:
Changed the file naming convention as suggested. Still only applies to contracts under `contracts/mocks/` because we want to see the warning for the production contracts.

hardhat/ignore-unreachable-warnings.js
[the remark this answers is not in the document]
Amxx: Ok, so the caching is for making sure an instance is reused accross all jobs that need the override. ok

contracts/mocks/unreachable/ERC721ConsecutiveEnumerableMock.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: MIT
```
Amxx, line 1: I would move this back to something like `contracts/mocks/ERC721ConsecutiveEnumerableMock.unreachable.sol` so that all ERC721 mocks are together

Amxx (approved):
LGTM

PR #3699, opened 2022-09-14
Make ERC721Votes no longer a Draft

frangio:
The contract was marked as a draft because it has a dependency on EIP712, but EIP712 was made Final recently.

I've also made some missing necessary changes regarding EIP712.

This should go in 4.8.


Amxx (approved):
LGTM

PR #3702, opened 2022-09-15
made the roles also enumerable for accounts on access control

kevupton:
Added functionality to also enumerate the roles for accounts on the `AccessControlEnumerable`.

Naming is just inverted, as I thought that was the most intuitive.

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [x] Changelog entry


frangio:
Appreciate the desire to contribute but please open an issue for discussion first.

This has been considered before in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2139, although at the time we did not have the distinction between `AccessControl` and `AccessControlEnumerable`.

In your issue, please list the context in which on-chain enumerability is useful, and why it is not enough to use the events.

kevupton:
ahhh very true of event enumeration would be just as good. 🙏

PR #3703, opened 2022-09-15
Update the informational link

emretepedev:
https://ronan.eth.link is currently not available, but we can use https://ronan.eth.limo.


frangio (approved):
Thanks!

PR #3707, opened 2022-09-16
Update docs pipeline to solidity-docgen@0.6

frangio:
Simply upgrading the setup and templates to work with solidity-docgen 0.6. The rendered site shouldn't change.


frangio:
Found some things I still need to fix for parity with the current site.

frangio:
Ok this seems to be at parity now, will go ahead and merge.

PR #3714, opened 2022-09-21
Add ReentrancyGuard status getter

zhiqiangxu:
This PR adds a entered modifier to ReentrancyGuard contract, it's useful for callbacks.


Amxx:
Hello @zhiqiangxu 

I'm not sure about this one. If multiple functions in the contract are nonReentrant, you don't know which one was you are entering from. You are also not sure what path you are entering trough, and if you are not entering twice (either one after the other or one within the other).

Can you give us more details about your usecase?

zhiqiangxu:
> Hello @zhiqiangxu
> 
> I'm not sure about this one. If multiple functions in the contract are nonReentrant, you don't know which one was you are entering from. You are also not sure what path you are entering trough, and if you are not entering twice (either one after the other or one within the other).
> 
> Can you give us more details about your usecase?

Your concern makes sense, I now prefer just exposing the guard status and let users decide his own logic?

frangio:
I'm ok with exposing the guard status but we shouldn't make this function virtual.

zhiqiangxu:
> I'm ok with exposing the guard status but we shouldn't make this function virtual.

Yeah internal is enough.

frangio:
Coverage wants you to add a test for the new getter.

contracts/security/ReentrancyGuard.sol
```diff
@@ -66,4 +66,11 @@ abstract contract ReentrancyGuard {
         // https://eips.ethereum.org/EIPS/eip-2200)
         _status = _NOT_ENTERED;
     }
+
+    /**
+     * expose _status as internal function, the caller decides what to do with it
+     */
+    function _guardStatus() internal view returns (uint256) {
```
frangio, line 73: Let's call this `_reentrancyGuardStatus`?

zhiqiangxu: Done.

zhiqiangxu:
> Coverage wants you to add a test for the new getter.

Added a check in an existing nonReentrant function, let's see if Coverage will be happy this time.

frangio (dismissed):
Looks good to me! Awaiting review by @Amxx.

contracts/mocks/ReentrancyMock.sol
```diff
@@ -35,6 +35,8 @@ contract ReentrancyMock is ReentrancyGuard {
         _count();
         bytes4 func = bytes4(keccak256("callback()"));
         attacker.callSender(func);
+
+        require(_reentrancyGuardStatus() == 2, "ReentrancyMock: _reentrancyGuardStatus failed");
```
frangio, line 39: I don't think this is getting executed, the attacker line above probably reverts.

Please add a separate test specifically for the guard status, and for the various values it can have.

frangio (commented):
Realized tests still need work. The coverage report is clearly showing that the getter is not tested.

contracts/mocks/ReentrancyMock.sol
Amxx: +1 for specific testing

Amxx (changes requested):
Looks mostly good. Need better testing.

I'd do an function that emit the value of the guard ... and this function would be called through 2 public function (one with the modifier, one without)

contracts/mocks/ReentrancyMock.sol
zhiqiangxu: Done. 

Sorry for late response, busy doing a bunch of other things lol:)

frangio:
Sorry for the back and forth. I realized the getter was leaking the exact values we use for entered and not entered, and these are values we might want to change in the future, given that they were chosen to optimize gas usage and the optimal values may be different once again in the future.

So in order to avoid leaking the exact values I've changed the function to return a boolean and renamed the function accordingly.

@zhiqiangxu @Amxx Let me know what you think.

zhiqiangxu:
> Sorry for the back and forth. I realized the getter was leaking the exact values we use for entered and not entered, and these are values we might want to change in the future, given that they were chosen to optimize gas usage and the optimal values may be different once again in the future.
> 
> So in order to avoid leaking the exact values I've changed the function to return a boolean and renamed the function accordingly.
> 
> @zhiqiangxu @Amxx Let me know what you think.

Cool, I think that's better !

zhiqiangxu:
Any update?

Amxx (approved):

PR #3716, opened 2022-09-22
Fix fuzzing coverage

frangio:
We're missing some assertions.


Amxx:
I think they are redundant (tautology of the if / elseif) but if you think we should add that its fine by me.

Amxx (approved):

frangio:
I don't think it's redundant because the "bigger" and "smaler" helpers use strict inequalities.

Amxx:
What I meant is that if its neither bigger not smaller, then we know its equal. We can check it, but we known we wouldn't be there is it was not true.

PR #3717, opened 2022-09-22
Add fuzz testing of mulDiv

frangio:
We need a way to state in `vm.assume` that `x * y / d` doesn't overflow uint256. I'm not sure how to express that condition. I put in a stronger condition that is sufficient but not necessary, I think...


test/utils/math/Math.t.sol
```diff
@@ -115,4 +115,59 @@ contract MathTest is Test {
...
     }
+
+    function testMulDiv(uint256 x, uint256 y, uint256 d) public {
+        vm.assume(d != 0);
+
+        // assume no overflow
+        // TODO: make this condition weaker
+        uint256 M = type(uint256).max;
+        vm.assume(y == 0 || x / d <= M / y);
+        vm.assume(x == 0 || y / d <= M / x);
```
Amxx, line 126:
```suggestion
uint256 p0, p1;
assembly {
    p0 := mul(x, y)
    p1 := mulmod(x, y, not(0))
}
vm.assume(p0 == p1 || (p0 == 1 && p1 == 0));
```

Amxx:
IDEA:

Given `x`, `y` and `d` ... muldiv gives use `q` such that

`x * y = q * d + r` with `0 <= r < d`.

Its pretty easy to get `r` by doing `mulmod(x, y, d)`.

The formula above can be rewritten as
`q * d = x * (y-k1) + (k1*y -r)`
`q * d = (x-k2) * y + (k2*x -r)`

with k1 = ceil(r/y)
and k2 = ceil(r/x)

If we manage to compute k1 and k2, which should be easy (no overflow), then the test should "just" be

```
uint256 r;
assembly { r := mulmod(x, y, d) }
uint256 k1 = _ceilDiv(r, y);
uint256 k2 = _ceilDiv(r, x);

uint256 a, b;
assembly { a := mulmod(q, d, x) }
assembly { b := mulmod(q, d, y) }

vm.assume(a == k1*y-r);
vm.assume(b == k1*y-r);
```

with

```
function _ceilDiv(uint256 a, uint256 b) {
    uint256 c = a / b;
    return c + (c * b ==a ? 0 : 1);
}
```

test/utils/math/Math.t.sol
```diff
@@ -115,4 +115,59 @@ contract MathTest is Test {
...
+        uint256 a_x_b_lo =  a_lo * b_lo;
+
+        uint256 carry_bit = (uint256(uint128(a_x_b_mid)) + uint256(uint128(b_x_a_mid)) + (a_x_b_lo >> 128)) >> 128;
+
+        high = a_x_b_hi + (a_x_b_mid >> 128) + (b_x_a_mid >> 128) + carry_bit;
+
+        unchecked {
+            low = a * b;
+        }
+    }
```
Amxx, line 172:
```suggestion

    // MULDIV
    function testMulDiv(uint256 x, uint256 y, uint256 d) public {
        vm.assume(d != 0);

        // assume no overflow
        // TODO: make this condition weaker
        uint256 M = type(uint256).max;
        vm.assume(y == 0 || x / d <= M / y);
        vm.assume(x == 0 || y / d <= M / x);

        uint256 q = Math.mulDiv(x, y, d);

        if (x == 0 || y == 0) {
            assertEq(q, 0);
        } else {
            uint256 r = _mulmod(x, y, d);
            assertEq(_mulmod(q, d, x), _ceildiv(r, x) * x - r);
            assertEq(_mulmod(q, d, y), _ceildiv(r, y) * y - r);
        }
    }

    function _mulmod(uint256 a, uint256 b, uint256 c) private pure returns (uint256 r) {
        assembly { r := mulmod(a, b, c) }
    }

    function _ceildiv(uint256 a, uint256 b) private pure returns (uint256) {
        return a / b + (a / b * b == a ? 0 : 1);
    }
```

Amxx:
counter example for the domain

```
x = 115792089237316195423570985008687907853269984665640564039457584007913129639935
y = 115792089237316195423570985008687907853269984665640564039457584007913129639933
d = 57896044618658097711785492504343953926634992332820282019728792003956564819968

x / d == UINT256_MAX / y == 1
y / d == UINT256_MAX / x == 1
```

frangio:
Great thing to also use fuzzing to test the domain of the function.

Our conditions to define the domain are wrong though. The inequalities are only equivalent to lack (or presence) of overflow using real division, and not using integer division.

@Amxx What is the benefit of the new test that you propose as an alternative to using `mulHighLow`? I think that was more straightforward than the one using ceilDiv, and easier to understand what the test implied. I'm not sure that the new tests are actually showing the right thing.

Amxx:
I find it easier to reason by then your version... but maybe your version is more straight forward. I'll re-read both with a fresh head.

Maybe we can have both checks ?

Amxx:
My math is not good, its only checking that the result is good modulo `x* y`. That is not enough. reverting back your you check

test/utils/math/Math.t.sol
```diff
@@ -110,9 +126,101 @@ contract MathTest is Test {
...
+        (uint256 y0, uint256 y1) = (y & type(uint128).max, y >> 128);
+
+        // Karatsuba algorithm
+        // https://en.wikipedia.org/wiki/Karatsuba_algorithm
+        uint256 z2 = x1 * y1;
+        uint256 z1a = x1 * y0;
+        uint256 z1b = x0 * y1;
+        uint256 z0 = x0 * y0;
+
+        uint256 carry = (uint256(uint128(z1a)) + uint256(uint128(z1b)) + (z0 >> 128)) >> 128;
```
frangio, line 211:
```suggestion
        uint256 carry = ((z1a & type(uint128).max) + (z1b & type(uint128).max) + (z0 >> 128)) >> 128;
```

test/utils/math/Math.t.sol
```diff
@@ -110,9 +126,101 @@ contract MathTest is Test {
...
+        uint256 x,
+        uint256 y,
+        uint256 z
+    ) private pure returns (uint256 r) {
+        assembly {
+            r := mulmod(x, y, z)
+        }
+    }
+
+    // https://stackoverflow.com/a/28904636/667959
```
frangio, line 199: We've deviated enough that this doesn't help.

```suggestion
```

frangio:
Left two comments but looks good to me!

test/utils/math/Math.t.sol
```diff
@@ -110,9 +126,101 @@ contract MathTest is Test {
...
+        (uint256 xyHi, ) = _mulHighLow(x, y);
+
+        // Violate {testMulDiv} assumption (covers d is 0 and result overflow)
+        vm.assume(xyHi >= d);
+
+        // we are outside the scope of {testMulDiv}, we expect muldiv to revert
+        try this.muldiv(x, y, d) returns (uint256) {
+            assertTrue(false);
+        } catch {
+            assertTrue(true);
```
frangio, line 170: The reason you saw `fail()` fail before was that the test was wrong.

```suggestion
            fail();
```

PR #3719, opened 2022-09-22
Remove inaccurate comment from `_flashFee` function

tinchoabbate:
Docstrings for the `_flashFee` function say that the function calls `flashFee`, but that's not the case. I'm removing that.

### DIFF
--- a/contracts/token/ERC20/extensions/ERC20FlashMint.sol
+++ b/contracts/token/ERC20/extensions/ERC20FlashMint.sol
@@ -42,7 +42,7 @@ abstract contract ERC20FlashMint is ERC20, IERC3156FlashLender {
     }
 
     /**
-     * @dev Returns the fee applied when doing flash loans. This function calls the {flashFee} function which returns the fee applied when doing flash loans.
+     * @dev Returns the fee applied when doing flash loans.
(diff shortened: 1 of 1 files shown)


Amxx:
Thank you @tinchoabbate for this PR

There is indeed an inconsistency in the documentation, but I'm not sure this is the right way of fixing it. Right now we have

> @dev Returns the fee applied when doing flash loans. This function calls the {flashFee} function which returns the fee applied when doing flash loans.

for `function _flashFee(address token, uint256 amount) internal view virtual returns (uint256)` and

> @dev Returns the fee applied when doing flash loans. By default this implementation has 0 fees. This function can be overloaded to make the flash loan mechanism deflationary.

for `function flashFee(address token, uint256 amount) public view virtual override returns (uint256)`

The truth is that its `flashFee` that calls `_flashFee`, and `_flashFee` that by default is set to 0 and that must be overriden (not overloaded)

Amxx:
I believe this is a more appropriate fix. Can you check it matches what you would expect from the doc?

tinchoabbate:
yesss, that makes more sense now! 👍 thanks for pushing those changes

Amxx (approved):

PR #3722, opened 2022-09-22
Change admin role allocation in TimelockControler constructor

### LINKED ISSUE #3720: Add explicit admin address to TimelockControl constructor
I've found a dozen or active projects on mainnet using multi-sigs or voting that missed the note about revoking the admin permissions from the deployer of TimelockController. 

The timelock is often deployed unmodified, and often a developer's only interaction with the timelock is a line or two in a deploy file. This means that the only thing they see is the constructor and its method names. Seeing these names, it's logical to conclude that the timelock only has two permissions - and to not know that super admin permissions are granted in the constructor to the deployer key.

I think a way of clearing up this confusion is to have the super admin powers, outside the timelock controlling itself, be explicitly granted as a list in the constructor, just like the other permissions.

(Tweets: https://twitter.com/danielvf/status/1572963306725318657)

Amxx:
Fixes #3720

- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/governance/TimelockController.sol
```diff
@@ -67,14 +67,15 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
...
-     * operations to perform future maintenance.
+     * NOTE: At construction, an optional admin can be set (in addition to the
+     * timelock itself). Setting such an admin can help further configuration
+     * of the timelock. After configuration is done, it is recommended to have
+     * this admin renounces its role and to rely on timelocked operations to
+     * perform future maintenance.
      */
     constructor(
         uint256 minDelay,
+        address admin,
```
frangio, line 78: I would add the argument at the end, it makes the breaking change somewhat more gentle, I think...

contracts/governance/TimelockController.sol
```diff
@@ -67,14 +67,15 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
...
      *
-     * NOTE: At construction, both the deployer and the timelock itself are
-     * administrators. This helps further configuration of the timelock by the
-     * deployer. After configuration is done, it is recommended that the
-     * deployer renounces its admin position and relies on timelocked
-     * operations to perform future maintenance.
+     * NOTE: At construction, an optional admin can be set (in addition to the
+     * timelock itself). Setting such an admin can help further configuration
+     * of the timelock. After configuration is done, it is recommended to have
+     * this admin renounces its role and to rely on timelocked operations to
```
frangio, line 73:
```suggestion
     * of the timelock. After configuration is done, this admin should
     * renounce its role, and only timelocked operations should be used to
```

contracts/governance/TimelockController.sol
```diff
@@ -67,14 +67,15 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
      * proposer and the canceller role (for backward compatibility). The
      * executors receive the executor role.
      *
-     * NOTE: At construction, both the deployer and the timelock itself are
-     * administrators. This helps further configuration of the timelock by the
-     * deployer. After configuration is done, it is recommended that the
-     * deployer renounces its admin position and relies on timelocked
-     * operations to perform future maintenance.
+     * NOTE: At construction, an optional admin can be set (in addition to the
```
frangio, line 70:
```suggestion
     * CAUTION: At construction, an optional admin can be set (in addition to the
```

contracts/governance/README.adoc
```diff
@@ -153,7 +153,7 @@ Operations status can be queried using the functions:
 [[timelock-admin]]
 ===== Admin
 
-The admins are in charge of managing proposers and executors. For the timelock to be self-governed, this role should only be given to the timelock itself. Upon deployment, both the timelock and the deployer have this role. After further configuration and testing, the deployer can renounce this role such that all further maintenance operations have to go through the timelock process.
+The admins are in charge of managing proposers and executors. For the timelock to be self-governed, this role should only be given to the timelock itself. Upon deployment, the admin role can be granted to any address (in addition to the timelock itself). After further configuration and testing, this optional admin should be renounce its role such that all further maintenance operations have to go through the timelock process.
```
frangio, line 156:
```suggestion
The admins are in charge of managing proposers and executors. For the timelock to be self-governed, this role should only be given to the timelock itself. Upon deployment, the admin role can be granted to any address (in addition to the timelock itself). After further configuration and testing, this optional admin should renounce its role such that all further maintenance operations have to go through the timelock process.
```

CHANGELOG.md
```diff
@@ -37,6 +38,8 @@
...
 
+ * `TimelockController`: During deployment, the TimelockController used to grant the `TIMELOCK_ADMIN_ROLE` to the deployer and to the timelock itself. The deployer was then expected to renounce this role once configuration of the timelock is over. Failing to renounce that role allows the deployer to change the timelock permissions (but not to bypass the delay for any time-locked actions). The role is no longer given to the deployer by default. A new parameter `admin` can be set to a non-zero address to grant the admin role during construction (to the deployer or any other address). Just like previously, this admin role should be renounced after configuration. If this param is given `address(0)`, the role is not allocated and doesn't need to be revoked. In any case, the timelock itself continues to have this role
```
frangio, line 41:
```suggestion
 * `TimelockController`: During deployment, the TimelockController used to grant the `TIMELOCK_ADMIN_ROLE` to the deployer and to the timelock itself. The deployer was then expected to renounce this role once configuration of the timelock is over. Failing to renounce that role allows the deployer to change the timelock permissions (but not to bypass the delay for any time-locked actions). The role is no longer given to the deployer by default. A new parameter `admin` can be set to a non-zero address to grant the admin role during construction (to the deployer or any other address). Just like previously, this admin role should be renounced after configuration. If this param is given `address(0)`, the role is not allocated and doesn't need to be revoked. In any case, the timelock itself continues to have this role.
```

contracts/governance/TimelockController.sol
DanielVF: Agree. End is better.

Amxx: done

frangio:
This looks good, my only remaining question is if we should include a warning note in the documentation about _past_ versions of the timelock and their treatment of the deployer account.

Amxx:
Its true that we don't keep past version of the doc for every minor... I think this is something we should do in the future (each version is build from the corresponding release branch)

Fell free to add any additional warning.

frangio (approved):

Issue #3721, opened 2022-09-22
Deploying an ERC4626 vault with an asset without code reverts in v4.8.0-rc.0


tinchoabbate:
In v4.7.0, deploying an `ERC4626` vault for an asset with no code worked. Either for an asset that still hasn't been deployed, or an asset under construction.

In v4.8.0-rc.0, the same fails. As far as I can tell, this is due to the changes introduced in #3639.

Here's a Foundry test:

```solidity
pragma solidity ^0.8.0;

import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
import "@openzeppelin/contracts/token/ERC20/extensions/ERC4626.sol";
import "forge-std/Test.sol";

contract Vault is ERC4626 {
    constructor(IERC20Metadata asset) ERC20("vToken", "vToken") ERC4626(asset) {}
}

contract ExampleToken is ERC20 {
    Vault public vault;

    constructor() ERC20("Token", "Token") {
        vault = new Vault(
            IERC20Metadata(address(this))
        );
    }
}

contract VaultTest is Test {
    ExampleToken token;

    function test_deploy() public {
        // Deploy token that deploys vault
        // In 4.7.0 this works
        // In 4.8.0-rc.0 this fails
        token = new ExampleToken();
    }
}
```

The root cause seems to be that after #3639, the constructor of `ERC4626` executes a `try/catch` to call the `decimals` function of the passed asset.
```solidity
constructor(IERC20 asset_) {
        uint8 decimals_;
        try IERC20Metadata(address(asset_)).decimals() returns (uint8 value) { ... }
```
However, when the asset doesn't have code, the call to `decimals` returns no data. Solidity attempts to decode it as a `uint8`, but fails. This triggers a revert _in the constructor_ of the `ERC4626` contract, instead of going into the `catch` clause. This is explained by the Solidity docs:

> If an error happens during the decoding of the return data inside a try/catch-statement, this causes an exception in the currently executing contract and because of that, it is not caught in the catch clause.

To make the error nicer, we could include a `require` in the constructor of the `ERC4626` to ensure the asset has code. In any case, it seems this new behavior is worth including as a breaking change.

PR #3733, opened 2022-09-27
Improve handling and docs for ERC4626.decimals

frangio:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3721 

Handles more errors when reading decimals from the asset: the asset may not be created yet, or the return value may not decode properly.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/extensions/ERC4626.sol
+++ b/contracts/token/ERC20/extensions/ERC4626.sol
@@ -33,18 +33,19 @@ abstract contract ERC4626 is ERC20, IERC4626 {
      * @dev Set the underlying asset contract. This must be an ERC20-compatible contract (ERC20 or ERC777).
      */
     constructor(IERC20 asset_) {
-        uint8 decimals_;
-        try IERC20Metadata(address(asset_)).decimals() returns (uint8 value) {
-            decimals_ = value;
-        } catch {
-            decimals_ = super.decimals();
-        }
-
+        (bool ok, bytes memory encodedDecimals) = address(asset_).call(
+            abi.encodeWithSelector(IERC20Metadata.decimals.selector)
+        );
+        _decimals = ok && encodedDecimals.length >= 32 ? abi.decode(encodedDecimals, (uint8)) : super.decimals();
         _asset = asset_;
-        _decimals = decimals_;
     }
 
-    /** @dev See {IERC20Metadata-decimals}. */
+    /**
+     * @dev Decimals are read from the underlying asset in the constructor and cached. If this fails (e.g., the asset
+     * has not been created yet), the cached value is set to a default obtained by `super.decimals()` (which depends on
+     * inheritance but is most likely 18). Override this function in order to set a guaranteed hardcoded value.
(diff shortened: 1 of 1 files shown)


contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -33,18 +33,21 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
      */
     constructor(IERC20 asset_) {
-        uint8 decimals_;
-        try IERC20Metadata(address(asset_)).decimals() returns (uint8 value) {
-            decimals_ = value;
-        } catch {
-            decimals_ = super.decimals();
-        }
-
+        (bool ok, bytes memory encodedDecimals) = address(asset_).call(
```
Amxx, line 36: We use `success` everywhere. For consistency I'll use it here as well.

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -33,18 +33,21 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
+            abi.encodeWithSelector(IERC20Metadata.decimals.selector)
+        );
+        _decimals = ok && encodedDecimals.length >= 32
+            ? uint8(abi.decode(encodedDecimals, (uint256)))
```
Amxx, line 40: Should we SafeCast ?

frangio: We want to avoid reverts here. But I realize now that if the decimals is > 255 we should consider it an error and fallback to super.

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -33,18 +33,33 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
+    /**
+     * @dev Attempts to fetch the asset decimals. A return value of false indicates that the attempt failed in some way.
+     */
+    function _tryGetAssetDecimals(IERC20 asset_) private returns (bool, uint8) {
+        (bool success, bytes memory encodedDecimals) = address(asset_).call(
+            abi.encodeWithSelector(IERC20Metadata.decimals.selector)
+        );
+        if (success && encodedDecimals.length >= 32) {
+            uint256 returnedDecimals = abi.decode(encodedDecimals, (uint256));
+            if (returnedDecimals <= type(uint8).max) {
```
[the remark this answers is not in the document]
frangio: Apps may struggle with 240 decimals, but there is nothing in the ERC20 spec that prevents that!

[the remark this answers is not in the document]
Amxx: keep in mint that one token is 10**decimals

So the totalSupply of your token is `type(uint256).max` (the maximum possible value) and the decimal number is 240 ... then to total supply is 1.157e-163 tokens.

That is technically working, but also REALLY stupid.

JulissaDantes (approved):
Looks good!

PR #3734, opened 2022-09-27
Add combined update and get functions to the Counters library

### LINKED ISSUE #2786: Return the current `_value` when incrementing or decrementing a `Counter`

elawbek:
Fixes #2786

Functions have been added to the `Counters` library that return the previous/next counter value and increment/decrement it in the storage by one.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,57 +1,58 @@
 # Changelog
 
 ## Unreleased
+* `Counters`: Added new functions that return the previous/next counter value and increase/decrease it by one in the storage. ([#3734](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3734))
+## Unreleased
 
-- `Counters`: Added new functions that return the previous/next counter value and increase/decrease it by one in the storage. ([#3734](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3734))
-- `TimelockController`: Added a new `admin` constructor parameter that is assigned the admin role instead of the deployer account. ([#3722](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3722))
-- `Initializable`: add internal functions `_getInitializedVersion` and `_isInitializing` ([#3598](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3598))
-- `ERC165Checker`: add `supportsERC165InterfaceUnchecked` for consulting individual interfaces without the full ERC165 protocol. ([#3339](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3339))
-- `Address`: optimize `functionCall` by calling `functionCallWithValue` directly. ([#3468](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3468))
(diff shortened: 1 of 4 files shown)


contracts/mocks/CountersImpl.sol
```diff
@@ -21,6 +22,22 @@ contract CountersImpl {
         _counter.decrement();
     }
 
+    function getAndIncrement() public {
+        helper = _counter.getAndIncrement();
```
frangio, line 26: We normally use events to inspect the return values in this case. If you could do that it would be appreciated.

CHANGELOG.md
```diff
@@ -2,55 +2,56 @@
 
 ## Unreleased
```
frangio, line 3: There a lot of changes in this file, it looks like a linter processed it.

Also, please add a new separate `Unreleased` section at the top, since this section corresponds to the 4.8 release which is already feature-frozen.

contracts/utils/Counters.sol
```diff
@@ -37,6 +37,44 @@ library Counters {
         }
     }
 
+    /**
+     * @dev Returns the current value and increments the value in the storage.
+     */
+    function getAndIncrement(Counter storage counter) internal returns (uint256 counter_) {
```
frangio, line 43: Can we use the variable name `value` for the return value?

elawbek: done

contracts/mocks/CountersImpl.sol
elawbek: done

CHANGELOG.md
elawbek: may be done, please check

elawbek: The reason for these big changes was because Prettier rewrote some previously written data when saving the file.
Maybe you should add .prettierignore for CHANGELOG.md?

elawbek:
May I ask what else is required of me?

frangio (approved):

frangio:
@elawbek Hopefully nothing else. We will let you know otherwise. Thanks!

frangio:
Sorry this never got merged. We ended up removing Counters in 5.0 so this needs to be closed now.

PR #3737, opened 2022-09-28
Erc4907 support#3735

### LINKED ISSUE #3735: EIP-4907 support
**🧐 Motivation**

There is a number of games with gene sequences mechanism like CryptoKitties, DeRace, and Axie. Imagine I have a Kitty with rare genes and I want to breed it with others for profit. There is no easy way of doing so because all games have different interfaces and thus It is difficult for exchanges to implement lend/borrow mechanics. ERC4907 will standardize the way how it is done without implementing a flash-borrower and magic tricks

**📝 Details**

Here is ready to use implementation
https://thirdweb.com/doubledev.eth/ERC4907

LanceSnow:
Fixes #3735

support [ERC4907](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4907.md)

ERC4907 is an extension of [EIP-721](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md). It proposes an additional role (user) which can be granted to addresses, and a time where the role is automatically revoked (expires). The user role represents permission to "use" the NFT, but not the ability to transfer it or set users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Unreleased
-
+ * `ERC4907`: add implementation of the ERC4907 standard. ([#3735](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3737))
  * `TimelockController`: Added a new `admin` constructor parameter that is assigned the admin role instead of the deployer account. ([#3722](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3722))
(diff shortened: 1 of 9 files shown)


frangio:
Please allow for discussion first so we can understand if this is something we want in the library.

frangio:
Closing as explained in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3735#issuecomment-1261019778.

PR #3744, opened 2022-09-30
[IDEA] Don't rely on _beforeConsecutiveTokenTransfer for balance tracting

Amxx:
This PR
- uses a new `ERC721Concutive._batchBalances` to keep track of the tokens minted sequentially.
- don't write to `ERC721._balances` when minting batches.
- rely on the unchecked sum of both values to get the actual result.

Note: this uses the fact that `ERC721._balances` will overflow when transferring out a token that was minted in batch.

**The upside** is that we don't need to manipulate the `ERC721._balances` in the `_beforeConsecutiveTokenTransfer`. Unfortunatelly the hooks still have to be defined in `ERC721` so that ERC721Consecutive and ERC721Pausable can be binded together.

**The downside** is that we now have two memory slots. The cost of transferring should be similar in the long run, but a bit higher at first since the two slot needs to be initialized. Also, any account that got token minted and batch and then transfer them will have both slots set to non zero value (when the sum is actually zero). In addition, `balanceOf` now need 2 sload, increassing the cost by 2100 gas (plus the mapping lookup).

Is that worth it?

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/mocks/ERC721ConsecutiveEnumerableMock.unreachable.sol
+++ b/contracts/mocks/ERC721ConsecutiveEnumerableMock.unreachable.sol
@@ -27,6 +27,10 @@ contract ERC721ConsecutiveEnumerableMock is ERC721Consecutive, ERC721Enumerable 
         return super.supportsInterface(interfaceId);
     }
 
+    function balanceOf(address account) public view virtual override(IERC721, ERC721, ERC721Consecutive) returns (uint256) {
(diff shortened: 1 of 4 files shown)


frangio:
We're not going with this idea for now. Might revisit for 5.0 given if hooks are removed.

PR #3745, opened 2022-09-30
Reduce gas in `log256`

TomAFrench:
This PR replaces a multiplication by 8 inside `log256(uint256, Rounding)` with shifting the bits by 3. There's other usages of multiplication/division through bitshifts in the same file so I don't think this is an example of being overly "optimizooor".

#### PR Checklist

The change is covered by the existing tests and this changes doesn't warrant docs.

- [x] Tests
- [x] Documentation
- [x] Changelog entry


TomAFrench:
Ah, looks like the changelog entry is enforced. I'll add something.

frangio:
Do you know how much gas this saves? Waiting to see what our gas reports say.

Regardless, it looks good to me. I would put this on 4.8 (currently in release candidate, feature frozen). @Amxx would you agree?

TomAFrench:
A `MUL` is 5 gas whereas `SHL` is 3 so we should get a saving of 2 gas. If I run `forge snapshot` I get exactly that saving in practice as well.

Amxx:
I'm usually against adding non essential thing in the middle of a rc ... and saving 2 gas really doesn't feel essential.

However, we already have a rc update planned, so if you feel like adding it, I wouldn't oppose it.

Issue #3739, opened 2022-09-29
move _beforeConsecutiveTokenTransfer and _afterConsecutiveTokenTransfer to ERC721Consecutive.sol


TrejGun:
ERC721Consecutive introduces backward incompatibility between 4.7 and 4.8

My contracts are not compiling after the upgrade because I have to override _beforeConsecutiveTokenTransfer _afterConsecutiveTokenTransfer that I'm never ever going to use. These methods belong to extension and are used only by extension so why were they added to the core contract?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L495-L519

**💻 Environment**

**📝 Details**

**🔢 Code to reproduce bug**

Amxx:
Hello @TrejGun 

These hooks are in the core ERC721 contract because we need them to operate on the balances, which are private and can only be modified with the core ERC721 contract itself.

Having this new hooks should be backward compatible. Since any code written for 4.7 does not include them, they should not be defined or overridden anywhere. Consequently, just adding them to the core should not require any manual override.

Could you please provide an example of code that compile when matched to 4.7 but that triggers errors when matched with 4.8 ?

frangio:
Are you using ERC721Enumerable or ERC721Pausable?

frangio:
There is no way to make ERC721Pausable compatible with ERC721Consecutive without introducing some form of breaking change around overrides.

TrejGun:
> Are you using ERC721Enumerable or ERC721Pausable? Please do share your code.

yes I do

```pragma solidity ^0.8.9;

import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Burnable.sol";
import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
import "@openzeppelin/contracts/access/AccessControl.sol";
import "@openzeppelin/contracts/utils/Counters.sol";

contract ERC721ABE is AccessControl, ERC721Burnable, ERC721Enumerable {
  using Counters for Counters.Counter;

  bytes32 public constant MINTER_ROLE = keccak256("MINTER_ROLE");

  Counters.Counter internal _tokenIdTracker;

  constructor(
    string memory name,
    string memory symbol
  ) ERC721(name, symbol) {
    _setupRole(DEFAULT_ADMIN_ROLE, _msgSender());
    _setupRole(MINTER_ROLE, _msgSender());
  }

[...]

PR #3753, opened 2022-10-06
Refactor consecutive transfer hooks

frangio:
In #3739 we realized that the introduction of ERC721Consecutive results in a breaking change, not because the behavior of a contract changes, but because it may stop compiling due to new required disambiguating overrides. An example is the combination of ERC721+ERC721Pausable, or ERC721+ERC721Enumerable, even though these combinations don't involve ERC721Consecutive the user still has to write a disambiguating override for `_beforeConsecutiveTokenTransfer`.

This understandably results in confusion. This is a proposal for an alternative way to design the hooks, which gives in to the breaking change but with an arguably less disruptive result. The idea is to reuse `_beforeTokenTransfer` but add a new argument, so that the signature becomes `_beforeTokenTransfer(address from, address to, uint firstTokenId, uint lastTokenId)`. Thus, the same hook is reused for simple and consecutive token transfers. A user that updates the dependency to 4.8 "simply" has to add the new `lastTokenId` argument to get their contract to compile, and as long as they don't use ERC721Consecutive they can ignore that argument.

This PR is incomplete but I wanted to get the idea out since this is blocking the release of 4.8. I personally think this is an important thing to consider. The effect of having consecutive-related overrides in a contract that doesn't use ERC721Consecutive is not nice at all, and I think it just shows that the design needs to be improved.

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -287,7 +287,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
         require(to != address(0), "ERC721: mint to the zero address");
         require(!_exists(tokenId), "ERC721: token already minted");
 
-        _beforeTokenTransfer(address(0), to, tokenId);
+        _beforeTokenTransfer(address(0), to, tokenId, tokenId);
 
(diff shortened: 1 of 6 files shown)


frangio:
This needs better documentation.

PR #3755, opened 2022-10-07
ERC721VotesForced - give tokens different voting weight

toledoroy:
👋

As a follow-up on a forum thread regarding the assignment of different voting power for different ERC721 tokens.
https://forum.openzeppelin.com/t/erc721-voting-power-based-on-some-property/24550/22

## Problems
- The current version of Votes has an opt-out mechanism that requires all tokens to be of equal weight (so the getBalance() function would return a current voting power)
- Since ERC1155 doesn't support a function for getting the total number of tokens per account it's incompatible with the current Votes architecture.

## Solution
Remove the opt-out  option and track all voting power movements for all accounts at all time. 
- Allowing to assign different power to different tokens
- Paves the way to ERC1155 compatibility

Drawbacks:
- Using this contract will make all token transfers more expensive.

However, in some cases, post merge, and on L2s this increase in transaction fees might be worth it.

## Description

This PR includes a new version of Votes, dubbed VotesForced, with a new integration for ERC721 that uses an arbitrary token power function to determine the voting power for each token.

#### PR Checklist

- [X] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/utils/VotesForced.sol
+++ b/contracts/governance/utils/VotesForced.sol
@@ -1,0 +1,211 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.6.0) (governance/utils/Votes.sol)
+pragma solidity ^0.8.0;
+
+import "../../utils/Context.sol";
+import "../../utils/Counters.sol";
+import "../../utils/Checkpoints.sol";
+import "../../utils/cryptography/EIP712.sol";
+import "./IVotes.sol";
+
+/**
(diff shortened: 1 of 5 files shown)


toledoroy:
Removed the forced opt-in to make things more compatible with the existing Votes extension.

Everything's much simpler and more elegant now. 

Still wondering what to do about the behaviour tests. It passes only if the voting power for each token is worth 1 votes.
What would be the right way to address this? 
Maybe make changes to the behaviour tests / write another behaviour test / use 1 as token power on the Mock contract?

PR #3760, opened 2022-10-13
equality check for empty queue

0xbok:
`_begin<=_end` is an invariant for DoubleEndedQueue, so `empty()` can just check `_end == _begin` as `_end` is never less than `_begin`.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [x] Changelog entry


Amxx:
> By memory corruption, do you mean `dequeue` might be changed outside of this library by the user, and hence `<=` check saves from this corruption?

exactly

PR #3769, opened 2022-10-19
Rename code to creationCode for upcoming Solidity change

nishant-sachdeva:
code is becoming a reserved keyword in upcoming breaking change in solidity

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
  * `ReentrancyGuard`: Add a `_reentrancyGuardEntered` function to expose the guard status. ([#3714](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3714))
  * `ERC20Votes`: optimize by using unchecked arithmetic. ([#3748](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3748))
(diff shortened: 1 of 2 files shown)


cameel:
I think we should provide more context here.

We're currently implementing `immutable` for dynamic types in the compiler. With things like arrays and structs, however, it's much more tedious to enforce the immutability at contract construction time (and not everyone even thinks that this immutability is desirable) so we decided to make them freely writable at that stage instead. To make this clear we're introducing a new keyword: `code`. This is a breaking change so it will only really become a keyword in 0.9.0.

Turns out that OppenZeppelin has one instance of `code` being used as an identifier. This PR changes its name for future compatibility and also will help us test the compiler against OZ without having to maintain a fork just for this small tweak.

Amxx:
Thanks for the notification.

We would really value the ability to have immutable strings, for example for ERC20 name and symbols.

Considering this is expected to ship in 0.9.0, it won't affect OpenZeppelin contract's 4.x.

cameel:
They're coming :)

And to clarify - dynamic immutables will very likely land before 0.9.0 and be non-breaking. It's only the introduction of a new keyword that's breaking. On 0.8.x it will simply not be a keyword and we'll use a parser hack to get it working instead.

frangio:
`Create2Impl` is no longer in the repo so this can be closed.

PR #3771, opened 2022-10-19
func name implies array return but returns address

robriks:
'delegates()' implies a plural return value, ie an array of delegates (addresses) changing the function name to 'delegatee()' more closely represents its return value, and it matches the comments in lines 93 and 101  'delegation()' would also be acceptable

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/governance/utils/Votes.sol
+++ b/contracts/governance/utils/Votes.sol
@@ -85,7 +85,7 @@ abstract contract Votes is IVotes, Context, EIP712 {
     /**
      * @dev Returns the delegate that `account` has chosen.
      */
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @robriks 

We use this name in order to maintain compatibility with [Comp.sol](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/Comp.sol#L24), which pre-dates our contracts.

Changing it would be a breaking change that would negatively affect the ecosystem. As much as I agree that this function's name could been better, changing it right now would do more harm than good.

PR #3773, opened 2022-10-19
Add Strings.toString for signed integers

galadd:
- [x] Tests
- [ ] Documentation
- [x] Changelog entry


Amxx:
Hello, @GbolahanAnon 

Considering that we already have `SignedMath:abs(int256)` and `Strings.toString(uint256)`, I think the implementation could be greatly simplified.

```
function toString(int256 value) internal pure returns (string memory) {
    return string.packed(value < 0 ? "-" : "", toString(abs(value));
}
```

GbolahanAnon:
That makes sense. I just implemented that. You can check the latest update

contracts/utils/Strings.sol
```diff
@@ -37,6 +38,13 @@ library Strings {
         }
     }
 
+    /**
+     * @dev Converts a `int256` to its ASCII `string` decimal representation.
+     */
+    function toString(int256 value) internal pure returns (string memory) {
+        return string(abi.encodePacked(value < 0 ? "-" : "",toString(SignedMath.abs(value))));
```
Amxx, line 45: any reason to do `string(abi.encodePacked(...))` and not `string.concat(...)`. I think the second one is clearer and cleaner

GbolahanAnon: I used `abi.encodePacked` because it works with previous version of solidity and if a beginner wants to use the library, they aren’t faced with much difficulty. But on second thought, it's better a beginner gets used to the new implementation. 

I have made a new commit

Amxx: That would require bumping the pragma from `pragma solidity ^0.8.0;` to `pragma solidity ^0.8.12;` ...

It not just a matter of beginner devs, but also of good support with big projects that are fixing a solidity version that is before 0.8.12.

Considering we can do without the pragma bump, you might be right that we should probable use `string(abi.encodePacked(...))`.

GbolahanAnon:
Alright then, I think it's all good to go 👍🏽

GbolahanAnon:
This PR has not been merged. Is there any change I need to effect?

Amxx:
We are still missing a Changelog entry for this PR.

Amxx:
Note: I'd love #3666 to be merge first, so that the tests can use the new syntax (instead of having to update #3666 yet again)

frangio (approved):
LGTM

test/utils/Strings.test.js
```diff
@@ -29,12 +29,24 @@ contract('Strings', function (accounts) {
       '12345678901234567890123456789012345678901234567890',
       '123456789012345678901234567890123456789012345678901234567890',
       '1234567890123456789012345678901234567890123456789012345678901234567890',
-    ].reduce((acc, value) => Object.assign(acc, { [value]: new BN(value) }), {
-      MAX_UINT256: constants.MAX_UINT256.toString(),
+    ].reduce((acc, value) => Object.assign(acc, { [value]: new BN(value), [-value]: new BN(-value) }), {
```
frangio, line 32: This is unwieldy but I want to tackle it separately by adding a linter rule to disallow `reduce`.

ernestognw (commented):
For the record, let's group tests by type so they don't mix with each other

```js
describe('toString', function () {
  ...
  describe('int256', function () { ... });
  describe('uint256', function () { ... });
  ...
})
```

ernestognw (approved):

Issue #1795, opened 2019-06-13
String utilities


Arachnid:
Does Open Zeppelin have any interest in integrating a library for string manipulation? I created the most widely used one [here](https://github.com/Arachnid/solidity-stringutils) nearly three years ago, but lack the time to maintain it as a separate library. If there's interest, I'm happy to license and contribute it in any way that works for the OZ team.

Skyge:
Emmm, could you please give me an example of such an operation, or why do not I do this offline?

nventuro:
Hi there @Arachnid, thank you very much for this! We've actually discussed your library a couple days ago in https://github.com/OpenZeppelin/openzeppelin-solidity/pull/1746, and started work on a [`Strings` library](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/fbbff535285fbff1a55a18af110b46eafb15f4bf/contracts/drafts/Strings.sol) (rather barebones so far). I think it'd be great to have your work here!

We'd have to migrate the test cases from Solidity to JavaScript (to bring them in line with the rest of our test suite), and update the code so that we can target solc v0.5.0, but I don't see any major blockers in this regard.

Also, note that `openzeppelin-solidity` is licensed under MIT, whereas `stringutils` uses Apache v2: would you be okay with re-licensing it under MIT for this?

We'll take a look at the API to analyse if we want to bring the whole library over, or just a portion of it to start with, and figure out the best way to make this happen while bothering you as little as possible (e.g. we won't ask you to rewrite the tests in JavaScript for us).

Once again, thanks a lot!

crazyrabbitLTC:
That would be fantastic! I've used your string library before @Arachnid and I think it's great. (Maybe I messaged you about it once). I think it would also be a great EVM package! Let see if they can incorporate it!

Arachnid:
@nventuro Sounds good! I'm happy to relicense my contributions under any license you see fit - MIT is absolutely fine. Please let me know if you need anything else from me.

abcoathup:
That is so awesome @Arachnid.  

@Skyge for your background, in two contracts this year, to reduce gas costs when minting ERC721 tokens with metadata, I used string concatenation and uint to string from a different implementation (I only found Nick's library later).  I proposed this as an addition in the [forum](https://forum.zeppelin.solutions/t/add-support-for-erc721-token-uri-concatenation-of-base-token-uri-and-tokenid/566)

Skyge:
[...]

PR #3774, opened 2022-10-21
Add Strings.equal

0xCaso:
Hi! I'd like to propose the addition of the new method ```equal``` for Strings, which would make things more elegant and make the code less repetitive. Also, it should be really useful for testing purposes with Foundry.

I'm referring to this (old) issue: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1795#issuecomment-1286701350

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/utils/Strings.sol
```diff
@@ -67,4 +67,12 @@ library Strings {
     function toHexString(address addr) internal pure returns (string memory) {
         return toHexString(uint256(uint160(addr)), _ADDRESS_LENGTH);
     }
+
+    /**
+     * @dev Returns true if the two strings are equal.
+     */
+    function compare(string memory a, string memory b) internal pure returns (bool) {
+        if (bytes(a).length != bytes(b).length) return false;
+        return keccak256(abi.encodePacked(a)) == keccak256(abi.encodePacked(b));
```
Amxx, line 76: Lets make use of lazy evaluation (also, the packing is not needed, and expensive)
```suggestion
        return bytes(a).length == bytes(b).length && keccak256(a) == keccak256(b);
```

Amxx: I'm curious how much the length check cost compared to what it saves. Is it really worth it ?

0xCaso: Actually, the compiler gives error if I try with `keccak256(a) == keccak256(b)`, am I doing something wrong? Also, [here](https://github.com/0xCaso/test-strings-compare-function) I compared four different functions, and the below one seems the winner.
```suggestion
        return keccak256(bytes(a)) == keccak256(bytes(b));
```

Amxx:
Note: I'd love #3666 to be merge first, so that the tests can use the new syntax (instead of having to update #3666 yet again)

0xCaso:
What's the state of #3666? @Amxx

contracts/utils/Strings.sol
```diff
@@ -67,4 +67,11 @@ library Strings {
     function toHexString(address addr) internal pure returns (string memory) {
         return toHexString(uint256(uint160(addr)), _ADDRESS_LENGTH);
     }
+
+    /**
+     * @dev Returns true if the two strings are equal.
+     */
+    function compare(string memory a, string memory b) internal pure returns (bool) {
```
frangio, line 74: Can we rename this to `equal`?

0xCaso: Sure!

frangio (dismissed):
LGTM

frangio (approved):

ernestognw (approved):
LGTM.

I'd like to explorer further if there's a cheaper way of doing this since keccak is an expensive operation.

I don't think it's worth it at the moment but mentioning for the record.

PR #3778, opened 2022-10-24
Fix comment in ERC721Consecutive

matthiasgeihs:
Fixes a comment in `ERC721Consecutive.sol`.

The comment states that ERC721 does not require a Transfer event when minting outside the constructor. However, it is the other way around. ERC721 does not require it *inside* the constructor.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio (approved):
Thanks!

PR #3781, opened 2022-10-25
Improve tests for ERC1155 and ERC721 error handling

### LINKED ISSUE #3775: Support for Custom Errors in ERC721 and 1155 onReceived hooks try catch
**🧐 Motivation**

Right now ERC1155 and 721 checks on safeTransfers do a try {} catch {} with only a catch on string and then a global catch that returns the default `ERCXXXX: transfer to non-ERCXXXXReceiver implementer`

This leads to always getting the default revert string when reverting in an onERCXXXReceived hook with a Custom Error, which does not allow the caller to get accurate revert reasons. 

It would be very nice for testing purpose and also to show the right reason of reverts on the front-end, to bubble the revert code when it's a custom error

**📝 Details**

It is possible to easily bubble revert codes using this simple code before or after the catch on string:

```
            assembly {
                revert(add(32, reason), mload(reason))
            }
        }
```

Amxx:
Fixes #3775

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry


EminenceAbhi (commented):
hello @Amxx

frangio:
Superceded by https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4314?

Amxx:
Yes, but that is a shame we did not merge that sooner. Not only would it have made #4314 simpler, but this PR is more complete:
- it checks panic, which #4314 doesn't
- it also does the checks on 721 (in addition to 1155).

I think we should revive this one during the "quality control" part of 5.0, for the sake of test completeness.

frangio (approved):

PR #3786, opened 2022-10-27
Support custom error parameters object

Amxx:
Prepare for wider use of custom errors with parameters

#### PR Checklist

- [x] Tests


frangio:
We're already handling parameters on `master`.

Amxx:
> We're already handling parameters on `master`.

Not using named (key/value). Only using lists.

This proposed to use objects instead of arrays to represent the params.

PR #3787, opened 2022-10-27
Optimize condition in _disableInitializers

MathisGD:
Minor optimization (2 gas) of `_disableInitializers` in `Initializable`.

A `JUMP` should happen if the condition is false. With `!=` it requires only `EQ` (3 gas) while with `<` it requires `LT` + `IZERO` (3 + 2 gas).

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio (approved):
Thanks @MathisGD. I'm really not happy that Solidity doesn't do this minor optimization on its own.

PR #3799, opened 2022-11-03
Use default admin role in TimelockController

### LINKED ISSUE #3389: TimelockController: Use default admin role
TimelockController currently uses a custom non-zero `TIMELOCK_ADMIN` role that has complicated the introduction of the new `CANCELLER_ROLE`  in 4.6 and necessitated a custom migration to set the role admin that would not have been necessary had `DEFAULT_ADMIN_ROLE` been used.

Moving forward we should always use the default admin role rather than custom roles.

JulissaDantes:
Fixes #3389  

The TimelockController contract currently uses a custom non-zero TIMELOCK_ADMIN role, in order to make easier the introduction of new roles without custom migrations the DEFAULT_ADMIN_ROLE needs to be used.

#### PR Checklist

- [X] Tests
- [X] Documentation
- [X] Changelog entry


Amxx:
That PR looks way bigger than it needs to be.

frangio:
Seems like the PR includes everything from `master`. Those are things we want to merge into `next-v5.0` because we need to keep it up to date but we should do it separately.

JulissaDantes:
Now it's ready!

frangio:
@JulissaDantes Can you rebase on `next-v5.0`? Goal would be to remove these commits from the commit list:

image

contracts/governance/TimelockController.sol
```diff
@@ -80,17 +79,16 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
         address[] memory executors,
         address admin
     ) {
-        _setRoleAdmin(TIMELOCK_ADMIN_ROLE, TIMELOCK_ADMIN_ROLE);
-        _setRoleAdmin(PROPOSER_ROLE, TIMELOCK_ADMIN_ROLE);
-        _setRoleAdmin(EXECUTOR_ROLE, TIMELOCK_ADMIN_ROLE);
-        _setRoleAdmin(CANCELLER_ROLE, TIMELOCK_ADMIN_ROLE);
+        _setRoleAdmin(PROPOSER_ROLE, DEFAULT_ADMIN_ROLE);
+        _setRoleAdmin(EXECUTOR_ROLE, DEFAULT_ADMIN_ROLE);
+        _setRoleAdmin(CANCELLER_ROLE, DEFAULT_ADMIN_ROLE);
```
Amxx, line 84:
```suggestion
```

There are not needed (already set by default)

contracts/governance/TimelockController.sol
```diff
@@ -80,17 +79,16 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
...
 
         // self administration
-        _setupRole(TIMELOCK_ADMIN_ROLE, address(this));
+        _setupRole(DEFAULT_ADMIN_ROLE, address(this));
```
Amxx, line 87: Lets use this opportunity to stop using the deprecated functions

```suggestion
        _grantRole(DEFAULT_ADMIN_ROLE, address(this));
```

contracts/governance/TimelockController.sol
```diff
@@ -80,17 +79,16 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
...
 
         // optional admin
         if (admin != address(0)) {
-            _setupRole(TIMELOCK_ADMIN_ROLE, admin);
+            _setupRole(DEFAULT_ADMIN_ROLE, admin);
```
Amxx, line 91: Lets use this opportunity to stop using the deprecated functions

```suggestion
            _grantRole(DEFAULT_ADMIN_ROLE, admin);
```

package-lock.json
```diff
@@ -28547,4 +28547,4 @@
       "dev": true
     }
   }
-}
+}
```
Amxx, line 28550: do we want that ?

contracts/governance/TimelockController.sol
JulissaDantes: Good call!

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+ * `TimelockController`: During deployment, the TimelockController used to grant the `TIMELOCK_ADMIN_ROLE` to the `admin` parameter, it will now assign the `DEFAULT_ADMIN_ROLE` to the `admin` parameter to be consistent with the rest of the repository.([#3799](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3799))
```
frangio, line 5: This needs to be under yet another (:sweat_smile:) "Unreleased" section at the top... Let's name it "Unreleased (Breaking)" so it's less ambiguous for now.

Also, there's a missing space after the period.

frangio, line 5: This changelog entry is not very accurate. It's not just about the `admin` parameter, the timelock itself also has the admin role.

We should say something about how the contract used to use a special admin role and it will now use the default like all other contracts.

frangio (approved):
Thank you!

PR #3802, opened 2022-11-05
Remove outdated blurb

robriks:
In GovernorCompatibilityBravo, which is the version being used in this tutorial, the queue () and execute() function do in fact only accept a uint proposalId parameter (contrary to the deletd paragraph). This proposalId is used to pull a ProposalDetails from storage which in turn is used to populate the rest of the required parameters for a call to queue/execute.

The deleted blurb directly contradicts the Bravo code, presumably because it was updated at some point but this tutorial was not.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/docs/modules/ROOT/pages/governance.adoc
+++ b/docs/modules/ROOT/pages/governance.adoc
@@ -308,7 +308,7 @@ image::tally-admin.png[Administration Panel in Tally]
 
 We will see now how to do this manually using Ethers.js.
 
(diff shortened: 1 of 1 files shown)


JulissaDantes:
Can you share the tutorial you were using?

Amxx:
This blurb accurately describes OZ's core governor. It indeed does not apply to the GovernorBravo compatibility layer, but the page you are modifying is a the governor help page (does not qualify as a tutorial) and is **not** specific to GovernorBravo.

frangio:
The page modified in this PR is [How to set up on-chain governrnance](https://docs.openzeppelin.com/contracts/4.x/governance). I consider this a tutorial, and it is true that the code snippets build a Governor with Bravo compatibility.

I actually think it's weird we're adding Bravo compatibility in this tutorial and that we should change that instead.

cairoeth:
[Docs tutorial](https://docs.openzeppelin.com/contracts/5.x/governance#governor) is now not using bravo, so the blurb is aligned with the contract. Closing PR.

PR #3811, opened 2022-11-11
change SLOAD to MLOAD

MollyHe5:
### Gas optimization
If a state variable is read immediately after assignment, then replacing this read directly with the previously assigned local variable will change one SLOAD to one MLOAD to save some gases. In the following example, function test0 costs 188 more gases than function test1 and function test2 costs 203 more gases than function test3.

```
contract Demo {
    uint public sa;
    address public add;

    event opUpdate(uint b);
    event addUpdate(address c);

    function test0(uint va) public {
        sa = va;
        emit opUpdate(sa);
    }

    function test1(uint va) public {
        sa = va;
        emit opUpdate(va);
    }

    function test2() public {
        add = msg.sender;
        emit addUpdate(add);
    }

    function test3() public {
        add = msg.sender;
        emit addUpdate(msg.sender);
    }
}
```

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/mocks/compound/CompTimelock.sol
+++ b/contracts/mocks/compound/CompTimelock.sol
@@ -81,7 +81,7 @@ contract CompTimelock {
         require(delay_ <= MAXIMUM_DELAY, "Timelock::setDelay: Delay must not exceed maximum delay.");
         delay = delay_;
 
-        emit NewDelay(delay);
+        emit NewDelay(delay_);
     }
 
     function acceptAdmin() public {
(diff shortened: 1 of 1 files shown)


frangio:
This contract is code [from Compound](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Timelock.sol) which we have for testing purposes only. Feel free to report this to them.

PR #3815, opened 2022-11-16
Gas optimization

MollyHe5:
Bools use 8 bit but they only need 1 bit. And it takes 20,000 gas to set a slot from zero to non-zero with SSTORE. So changing bool to uint can save a lot of gas.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/mocks/ERC1155ReceiverMock.sol
+++ b/contracts/mocks/ERC1155ReceiverMock.sol
@@ -7,9 +7,13 @@ import "../utils/introspection/ERC165.sol";
 
 contract ERC1155ReceiverMock is ERC165, IERC1155Receiver {
     bytes4 private _recRetval;
-    bool private _recReverts;
+    uint private constant _RECREVERTS_FALSE = 1;
+    uint private constant _RECREVERTS_TRUE = 2;
+    uint private _recReverts;
     bytes4 private _batRetval;
-    bool private _batReverts;
+    uint private constant _BATREVERTS_FALSE = 1;
+    uint private constant _BATREVERTS_TRUE = 2;
+    uint private _batReverts;
 
     event Received(address operator, address from, uint256 id, uint256 value, bytes data, uint256 gas);
     event BatchReceived(address operator, address from, uint256[] ids, uint256[] values, bytes data, uint256 gas);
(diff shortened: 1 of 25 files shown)


contracts/mocks/ERC1155ReceiverMock.sol
```diff
@@ -7,9 +7,13 @@ import "../utils/introspection/ERC165.sol";
...
     bytes4 private _recRetval;
-    bool private _recReverts;
+    uint private constant _RECREVERTS_FALSE = 1;
+    uint private constant _RECREVERTS_TRUE = 2;
+    uint private _recReverts;
     bytes4 private _batRetval;
-    bool private _batReverts;
+    uint private constant _BATREVERTS_FALSE = 1;
+    uint private constant _BATREVERTS_TRUE = 2;
+    uint private _batReverts;
```
Rassska, line 16: After that the number of slots required is 4, since: 
- 0: `bytes4 private _recRetval;`
- 1: `uint private _recReverts;`
- 2: `bytes4 private _batRetval;`
- 3: `uint private _batReverts;`

But in the initial version we're using only a single storage slot. 

Changing `bool` to `uint`  does make sense, since stack operations with bools require casting 32 bytes slots and that occurs overhead, but it's not a decent number in comparison.

Rassska (commented):
I don't see the value behind it.

contracts/mocks/ERC1155ReceiverMock.sol
MollyHe5: Changing the value of a bool variable from false to true costs 20,000 units of gas, which is much more than changing the value of a non-zero integer variable.

Rassska (commented):
As I said, using 4 slots is not ideal

contracts/mocks/ERC1155ReceiverMock.sol
Rassska: changing the value of a bool variable from false to true doesn't cost 20,000 units of gas, since the same slot contains another variables (packed variables with the current bool) i.e. the slot != 0. 

Pay attention on the fact, that in the same slot we have 4 variables and each of them have their own value.

MollyHe5: I reverted these changes, only leaving cases where a bool field occupies a whole slot.

frangio:
The files in the `mocks` directory are for testing purposes only. It doesn't make sense to optimize them.

The change suggested for `Pausable` is a breaking change to the storage layout so it would have to be done for the next major. Please open an issue with the suggestion.

PR #3816, opened 2022-11-16
Make ERC20Votes independent from ERC20Permit

### LINKED ISSUE #3160: Make ERC20Votes independent of ERC20Permit
**🧐 Motivation**

We need a more clear and independent structure of `ERC20` extensions.

**📝 Details**

Now `ERC20Votes` inherits `ERC20Permit` to use `EIP712` inheritance and moreover to abuse its permit `nonces`. I believe we can remove the `ERC20Permit` constructor to keep it abstract and make `ERC20Votes` have its own nonces for delegation and direct inheritance from `EIP712`.

Moreover, I see you have `Votes` implementation since 4.5, so hopefully, you are going to use this implementation in `ERC20Votes`.

JulissaDantes:
Fixes #3160 

This PR introduces a new contract, `Nonces`, to be used where keeping track of nonces is required. Also, ERC20Votes will now be inherited directly from EIP712 instead of using ERC20Permit.

This is a breaking change due to the changes in the storage layout.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


CHANGELOG.md
```diff
@@ -3,12 +3,14 @@
...
+ * `ERC20Votes`: This contract will not inherit from `ERC20Permit`, it have its own nonces for delegation and direct inheritance from EIP712. [#3816](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3816)
 
 ## Unreleased
 
  * `ReentrancyGuard`: Add a `_reentrancyGuardEntered` function to expose the guard status. ([#3714](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3714))
  * `ERC20Votes`: optimize by using unchecked arithmetic. ([#3748](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3748))
  * `Initializable`: optimize `_disableInitializers` by using `!=` instead of `<`. ([#3787](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3787))
+ * `Nonces`: This is a new contract to keep track of nonces and substitute the `_nonces` storage variable inside `Votes` contract. ([#3816](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3816))
```
frangio, line 13: Please move this to the right section.

contracts/token/ERC20/extensions/ERC20Permit.sol
```diff
@@ -70,8 +66,8 @@ abstract contract ERC20Permit is ERC20, IERC20Permit, EIP712 {
     /**
      * @dev See {IERC20Permit-nonces}.
      */
-    function nonces(address owner) public view virtual override returns (uint256) {
-        return _nonces[owner].current();
+    function nonces(address owner) public view virtual override(IERC20Permit, Nonces) returns (uint256) {
+        return super.nonces(owner);
     }
```
frangio, line 71: Does it compile if we remove this function definition? Please try with the latest Solidity version. This is a requirement that was relaxed at some point.

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -24,7 +26,7 @@ import "../../../utils/cryptography/ECDSA.sol";
  *
  * _Available since v4.2._
  */
-abstract contract ERC20Votes is IVotes, ERC20Permit {
+abstract contract ERC20Votes is ERC20, IVotes, Nonces, EIP712 {
```
frangio, line 29: We need this to use `Votes`. Currently it defines its own vote tracking and delegation logic, it should reuse that instead. See `ERC721Votes` for reference.

frangio, line 29: `Nonces` should be a part of `Votes`.

contracts/token/ERC20/extensions/ERC20Permit.sol
JulissaDantes: it does compile, but doesn't work properly always returning zero, tried with the latest version.

frangio: That doesn't make sense to me... There is something wrong there.

test/utils/Nonces.test.js
```diff
@@ -0,0 +1,20 @@
...
+
+  beforeEach(async function () {
+    this.nonces = await NoncesImpl.new();
+  });
+
+  it('gets a nonce', async function () {
+    expect(await this.nonces.nonces(sender)).to.be.bignumber.equal('0');
+  });
+
+  it('increment a nonce', async function () {
```
frangio, line 16: Let's add an assertion that the nonce for an "other" account is unaffected by `useNonce`. So we can check that the nonce is specific to the address argument.

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -24,174 +22,14 @@ import "../../../utils/cryptography/ECDSA.sol";
...
     function _maxSupply() internal view virtual returns (uint224) {
         return type(uint224).max;
     }
 
-    /**
-     * @dev Snapshots the totalSupply after it has been increased.
-     */
-    function _mint(address account, uint256 amount) internal virtual override {
-        super._mint(account, amount);
-        require(totalSupply() <= _maxSupply(), "ERC20Votes: total supply risks overflowing votes");
```
frangio, line 181: This check at `_mint` is necessary for `ERC20VotesComp`. I think it might be a good idea to remove that contract though (see https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3833).

If we're going to remove this check now, we should also remove `ERC20VotesComp` and resolve #3833 with this PR. We can then also entirely remove the `_maxSupply` function.

Alternatively, we should keep `_mint` with the max supply check, and remove it along with `ERC20VotesComp` in a future PR.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## Unreleased (Breaking)
 
  * `TimelockController`: Changed the role architecture to use `DEFAULT_ADMIN_ROLE` as the admin for all roles, instead of the bespoke `TIMELOCK_ADMIN_ROLE` that was used previously. This aligns with the general recommendation for `AccessControl` and makes the addition of new roles easier. Accordingly, the `admin` parameter and timelock will now be granted `DEFAULT_ADMIN_ROLE` instead of `TIMELOCK_ADMIN_ROLE`. ([#3799](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3799))
+ * `ERC20Votes`: This contract will not inherit from `ERC20Permit`, it have its own nonces for delegation and direct inheritance from EIP712. [#3816](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3816)
+
+ * `Nonces`: This is a new contract to keep track of nonces and substitute the `_nonces` storage variable inside `Votes` contract. ([#3816](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3816))
```
frangio, line 8: Reworded to lead with the verb and be more direct.

```suggestion
 * `ERC20Votes`: Changed internal vote accounting to reusable `Votes` module previously used by `ERC721Votes`. Removed implicit `ERC20Permit` inheritance. [#3816](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3816)
 * `Nonces`: Added a new contract to keep track of user nonces. Used for signatures in `ERC20Permit`, `ERC20Votes`, and `ERC721Votes`. ([#3816](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3816))
```

test/governance/utils/Votes.behavior.js
```diff
@@ -255,28 +256,29 @@ function shouldBehaveLikeVotes () {
...
       it('returns zero if < first checkpoint block', async function () {
         await time.advanceBlock();
-        const t2 = await this.votes.mint(this.account1, this.NFT1);
+        const t2 = await this.votes.mint(this.account1, this.token1);
         await time.advanceBlock();
         await time.advanceBlock();
+        const block = await time.latestBlock();
 
         expect(await this.votes.getPastTotalSupply(t2.receipt.blockNumber - 1)).to.be.bignumber.equal('0');
-        expect(await this.votes.getPastTotalSupply(t2.receipt.blockNumber + 1)).to.be.bignumber.equal('1');
```
frangio, line 263: Why was it necessary to change this? The previous code looks to me like it should've worked.

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -62,209 +41,12 @@ contract('ERC20Votes', function (accounts) {
...
     }
     const block = await web3.eth.getBlockNumber();
-    expect(await this.token.numCheckpoints(holder)).to.be.bignumber.equal('6');
     // recent
     expect(await this.token.getPastVotes(holder, block - 1)).to.be.bignumber.equal('5');
     // non-recent
     expect(await this.token.getPastVotes(holder, block - 6)).to.be.bignumber.equal('0');
   });
 
-  describe('set delegation', function () {
```
frangio, line 72: Let's add back all of the tests for ERC20Votes, so we can make sure that this PR is purely a refactor and doesn't change behavior in unexpected ways!

We should remove the redundant tests in a subsequent PR.

frangio (changes requested):

contracts/token/ERC20/extensions/ERC20Votes.sol
JulissaDantes: I think this PR can handle #3833

test/governance/utils/Votes.behavior.js
JulissaDantes: You are right it works as it was before, I must have been testing something and then drop it, forgetting to recall my update.

contracts/token/ERC20/extensions/ERC20Permit.sol
JulissaDantes: My bad, it doesn't compile

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -347,60 +347,11 @@ contract('ERC20Votes', function (accounts) {
       });
     });
 
-    describe('numCheckpoints', function () {
```
frangio, line 350: We need to keep all tests. This shows why: this PR should be a refactor, and it turns out we're accidentally (I hadn't realized) removing the functions `numCheckpoints` and `checkpoints`. If we want to remove these functions, it's a separate discussion that we haven't had, so we need to add them back.

frangio (changes requested):
Need to add back `numCheckpoints` and `checkpoints`.

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -202,79 +40,36 @@ abstract contract ERC20Votes is IVotes, ERC20Permit {
...
-            if (pos > 0 && oldCkpt.fromBlock == block.number) {
-                _unsafeAccess(ckpts, pos - 1).votes = SafeCast.toUint224(newWeight);
-            } else {
-                ckpts.push(
-                    Checkpoint({fromBlock: SafeCast.toUint32(block.number), votes: SafeCast.toUint224(newWeight)})
-                );
-            }
-        }
+    function numCheckpoints(address account) public view virtual returns (uint32) {
+        return super._numCheckpoints(account);
```
frangio, line 51: We shouldn't use super here. Only when we need to invoke the same function we're overriding.

```suggestion
        return _numCheckpoints(account);
```

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -202,79 +40,36 @@ abstract contract ERC20Votes is IVotes, ERC20Permit {
...
     }
 
-    function _add(uint256 a, uint256 b) private pure returns (uint256) {
-        return a + b;
+    /**
+     * @dev Get the `pos`-th checkpoint for `account`.
+     */
+    function checkpoints(address account, uint32 pos) public view virtual returns (Checkpoints.Checkpoint memory) {
+        return super._checkpoints(account, pos);
```
frangio, line 58:
```suggestion
        return _checkpoints(account, pos);
```

contracts/utils/Checkpoints.sol
```diff
@@ -68,6 +68,14 @@ library Checkpoints {
         return pos == 0 ? 0 : _unsafeAccess(self._checkpoints, pos - 1)._value;
     }
 
+    /**
+     * @dev Returns checkpoint at given position.
+     */
+    function getAtPosition(History storage self, uint32 pos) internal view returns (Checkpoint memory) {
+        Checkpoint storage last = _unsafeAccess(self._checkpoints, pos);
+        return last;
```
frangio, line 76: I don't think we should use `_unsafeAccess` here, we should be checking that the position is within bounds.

```suggestion
        return self._checkpoints[pos];
```

frangio (approved):
Thanks!

k06a:
Isn't it an issue that inheriting from both ERC20Permit and ERC20Votes would reuse same nonces for 2 different actions? I would suggest to refactor `Nonces` to `NoncesLib` to be able to have separate instances on it in these 2 abstract contracts. Same way as we are proposing to have `ReentrancyGuardLib` here: #3846

k06a:
Proposed these changes here: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3848

CHANGELOG.md
mrnadyy: 3

PR #3817, opened 2022-11-16
Introduce ERC20Pods and ERC1155Pods extensions

k06a:
Hi, everyone!

We are trying to submit Pods concept to `openzeppelin-contracts` library: https://github.com/1inch/erc20-pods

This concept allows token holders pemissionlessly and risk-free participate in incentivisation programs (aka farmings) and moreover participate in other activities as delegations or anything new.

#### PR Checklist

!!! Will add the following after the discussion.

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/interfaces/IERC1155Pods.sol
+++ b/contracts/interfaces/IERC1155Pods.sol
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts v5.0.0 (interfaces/IERC20Pods.sol)
+
+pragma solidity ^0.8.0;
+
+import "./IERC1155.sol";
+
+interface IERC1155Pods is IERC1155 {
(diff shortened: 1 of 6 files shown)


frangio:
What is the role of OpenZeppelin Contracts in this? Can users do this by installing a separate package?

k06a:
@frangio this ERC20Pods and ERC1155Pods looks like one of the most promising token extensions in the nearest future. We could setup a separate EIP for them if you think it makes sense. I believe having these extensions in OZ Contracts library is a good idea.

frangio:
This should be offered in a separate repository/package for now. For inclusion in OpenZeppelin Contracts a mature EIP and adoption (or significant interest) would be a prerequisite.

In the meantime I think a more thorough explanation in the readme for 1inch/erc20-pods would really help understand exactly what a pod is and how a token interacts with it.

PR #3820, opened 2022-11-17
Remove admin and implementation getters from TransparentUpgradeableProxy

### LINKED ISSUE #3599: Remove ifAdmin getters from TransparentUpgradeableProxy
TransparentUpgradeableProxy has getters `admin()` and `implementation()` but these functions are guarded by `ifAdmin`, as a result calling the function by anyone except the admin will not return the values. This is specially visible on Etherscan where admin will display as `0x`:

image

This essentially makes the getters useless so we might as well remove them.

JulissaDantes:
Fixes #3599 

The `TransparentUpgradeableProxy` contract had getters only accessible if the caller was the admin, but returned `0x` if anybody else calls them, this was more visible on Etherscan, making the getters useless. As part of the removal, the `ProxyAdmin` was updated as well to remove the 2 methods calling the `admin` and `implementation` getters.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased (Breaking)
 
  * `TimelockController`: Changed the role architecture to use `DEFAULT_ADMIN_ROLE` as the admin for all roles, instead of the bespoke `TIMELOCK_ADMIN_ROLE` that was used previously. This aligns with the general recommendation for `AccessControl` and makes the addition of new roles easier. Accordingly, the `admin` parameter and timelock will now be granted `DEFAULT_ADMIN_ROLE` instead of `TIMELOCK_ADMIN_ROLE`. ([#3799](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3799))
+ * `TransparentUpgradeableProxy`: The `admin()` and `implementation()` getters were removed, this impacts the `ProxyAdmin` contract as well, removing the `getProxyAdmin` and `getProxyImplementation` getters. ([#3820](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3820))
```
frangio, line 6:
```suggestion
 * `TransparentUpgradeableProxy`: Removed `admin` and `implementation` getters, which were only callable by the proxy owner and thus not very useful. ([#3820](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3820))
 * `ProxyAdmin`: Removed `getProxyAdmin` and `getProxyImplementation` getters. ([#3820](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3820))
```

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -50,32 +50,6 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
...
     }
 
-    /**
-     * @dev Returns the current admin.
-     *
-     * NOTE: Only the admin can call this function. See {ProxyAdmin-getProxyAdmin}.
-     *
-     * TIP: To get this value clients can read directly from the storage slot shown below (specified by EIP1967) using the
-     * https://eth.wiki/json-rpc/API#eth_getstorageat[`eth_getStorageAt`] RPC call.
-     * `0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103`
```
frangio, line 60: We should make sure we are keeping this documentation somewhere. We may have to move it to the top-level contract docs.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -50,32 +50,6 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
...
-    }
-
-    /**
-     * @dev Returns the current implementation.
-     *
-     * NOTE: Only the admin can call this function. See {ProxyAdmin-getProxyImplementation}.
-     *
-     * TIP: To get this value clients can read directly from the storage slot shown below (specified by EIP1967) using the
-     * https://eth.wiki/json-rpc/API#eth_getstorageat[`eth_getStorageAt`] RPC call.
-     * `0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc`
```
frangio, line 73: Same as above.

test/proxy/transparent/ProxyAdmin.test.js
```diff
@@ -51,18 +40,8 @@ contract('ProxyAdmin', function (accounts) {
...
-  describe('#getProxyImplementation', function () {
-    it('returns proxy implementation address', async function () {
-      const implementationAddress = await this.proxyAdmin.getProxyImplementation(this.proxy.address);
-      expect(implementationAddress).to.be.equal(this.implementationV1.address);
-    });
-
-    it('call to invalid proxy', async function () {
-      await expectRevert.unspecified(this.proxyAdmin.getProxyImplementation(this.implementationV1.address));
+      // Allows new admin to call ifAdmin functions
+      await this.proxy.changeAdmin(newAdmin, { from: newAdmin });// expect to not revert
```
frangio, line 44: Instead of this, to verify that the admin changed we should read the EIP-1967 storage slot.

test/proxy/transparent/ProxyAdmin.test.js
```diff
@@ -79,8 +58,8 @@ contract('ProxyAdmin', function (accounts) {
     context('with authorized account', function () {
       it('upgrades implementation', async function () {
         await this.proxyAdmin.upgrade(this.proxy.address, this.implementationV2.address, { from: proxyAdminOwner });
-        const implementationAddress = await this.proxyAdmin.getProxyImplementation(this.proxy.address);
-        expect(implementationAddress).to.be.equal(this.implementationV2.address);
+        const dummy = await new ImplV2(this.proxy.address);
+        expect(await dummy.version()).to.be.equals('V2');
```
frangio, line 62: This is good, but we should also read from the EIP-1967 storage slot to check the implementation was set.

test/proxy/transparent/ProxyAdmin.test.js
```diff
@@ -116,8 +95,8 @@ contract('ProxyAdmin', function (accounts) {
           await this.proxyAdmin.upgradeAndCall(this.proxy.address, this.implementationV2.address, callData,
             { from: proxyAdminOwner },
           );
-          const implementationAddress = await this.proxyAdmin.getProxyImplementation(this.proxy.address);
-          expect(implementationAddress).to.be.equal(this.implementationV2.address);
+          const dummy = await new ImplV2(this.proxy.address);
+          expect(await dummy.version()).to.be.equals('V2');
```
frangio, line 99: Same as above.

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
```diff
@@ -107,11 +94,6 @@ module.exports = function shouldBehaveLikeTransparentUpgradeableProxy (createPro
             this.receipt = await this.proxy.upgradeToAndCall(this.behavior.address, initializeData, { from, value });
           });
 
-          it('upgrades to the requested implementation', async function () {
```
frangio, line 110: We need to keep this test. We just have to use the new recommended method to retrieve the implementation address of a proxy: reading the EIP-1967 storage slot directly.

I think the same applies to most of the following tests that were removed, can you review them?

JulissaDantes: You are right, as a side note: I'll remove the `storage` section since I use storage to get implementation and admin everywhere, having a specific test for it is redundant.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
JulissaDantes: I moved both to the functions they were calling, `_getAdmin` and `_implementation`

test/proxy/transparent/ProxyAdmin.test.js
```diff
@@ -51,18 +39,10 @@ contract('ProxyAdmin', function (accounts) {
 
     it('changes proxy admin', async function () {
       await this.proxyAdmin.changeProxyAdmin(this.proxy.address, newAdmin, { from: proxyAdminOwner });
-      expect(await this.proxy.admin.call({ from: newAdmin })).to.eq(newAdmin);
-    });
-  });
+      const proxyAdminSlot = await getSlot(this.proxy, AdminSlot);
+      const proxyAdminAddress = web3.utils.toChecksumAddress(proxyAdminSlot.substr(-40));
```
frangio, line 43: Let's define a helper in `erc1967.js` called `getAddressInSlot` that does what these two lines do, so that we don't have to repeat them every time we need to read an address.

frangio (changes requested):

frangio (approved):
Thanks!

PR #3822, opened 2022-11-18
Fixed GPLQ spec, all rules passing

teryanarmen:
Updated GovernorPreventLateQuorum.spec by proving assumptions, removing unnecessary rules, and reorganizing.

[x] specs pass*
* failing on ci in this PR because of security measures that prevent usage of a premium certora key. All rules are passing in Certora's fork: https://github.com/Certora/openzeppelin-contracts/actions/runs/3505767253


teryanarmen:
Adding for clarification
-----------------------

The flag
```--solc solc```
was removed from all scripts as it is redundant. The prover will by default look for solc as the executable.

frangio (approved):
Thanks!

PR #3824, opened 2022-11-19
feat: implementation and tests for toString(int256)

### LINKED ISSUE #3758: Add string operation for int256 in Strings.sol
**🧐 Motivation**
There is no implementation of the conversion of int256 to String

**📝 Details**
Using [Strings.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Strings.sol) , We can convert uint256 to String,HexString and address to HexString. But int256 to string is required in many cases.

KolevDarko:
Fixes #3758

Added the toString(int256) function. It takes care of the sign and then uses toString(uint256) to perform the conversion to string.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


Amxx:
Duplicate of #3773

PR #3827, opened 2022-11-19
feat: make proposalId indexed in ProposalCreated event #3826

### LINKED ISSUE #3826: IGovernor: ProposalId indexed in ProposalCreated event
**🧐 Motivation**
Faster querying of ProposalCreated Events.

**📝 Details**
Make the ProposalId indexed in the ProposalCreated event in IGovernor.

paknecht:
Fixes #3826 

make proposalId indexed in ProposalCreated event

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
  * `ERC20Votes`: optimize by using unchecked arithmetic. ([#3748](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3748))
  * `Initializable`: optimize `_disableInitializers` by using `!=` instead of `<`. ([#3787](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3787))
  * `Math`: optimize `log256` rounding check. ([#3745](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3745))
+ * `IGovernor`: make `proposalId` indexed in `ProposalCreated` event. ([#3826](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3826))
 
 ### Deprecations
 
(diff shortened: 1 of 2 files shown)


paknecht:
Question 1: Should I also make a PR in this repo: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable or is that transferred somehow automatically? 
Question 2: Do I need to change anything in the documentation? If so, where exactly? Have found nothing in a quick search

Amxx:
Hello @paknecht 

For the record, the event is that way for compatibility with [GovernorBravo](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/GovernorBravoInterfaces.sol#L7) (which predates our implementation).

Changing that, while it would not change the event's "topic0", would break a lot of indexing systems. So yes, it would be nice to index that, but it would also be a breaking change :/

paknecht:
Hi @Amxx 
Thanks for looking at this PR.
I'm new in solidity but I would assume that `uint` to `uint256` is already a breaking change and therefore the GovernorBravo Event isn't compatible to Governor?

Who can decide if this breaking change here is acceptable? Or can't we make breaking changes in general?

Amxx:
`uint` is just an alias to `uint256`. They are the same thing, and using one over the other has no impact on event topic or function selector.

We are currently fully compatible with GovernorBravo events.

Amxx:
Whenever we consider making a breaking change, we try to weight in the advantages and the drawback. We sometimes do them, but we need a very strong reason.

For example, fixing a severe bug fully justify a breaking change.

In your case, the drawback is to break some indexing services. The governor subgraphs are one thing, but other services like Tally would also be affected. The upside of having this parameter indexed exists but its not that clear to me.

PR #3831, opened 2022-11-23
perf(proxy): add memory safe

wiasliaw:
Add `@solidity memory-safe-assembly` to `_delegate`.

I also found some assembly blocks aren't labeled `memory-safe`.
Should I add memory-safe to all assembly blocks?

### DIFF
--- a/contracts/proxy/Proxy.sol
+++ b/contracts/proxy/Proxy.sol
@@ -20,6 +20,7 @@ abstract contract Proxy {
      * This function does not return to its internal call site, it will return directly to the external caller.
      */
     function _delegate(address implementation) internal virtual {
+        /// @solidity memory-safe-assembly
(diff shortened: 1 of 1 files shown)


frangio:
This assembly lock is not memory safe according to the definition [in the Solidity docs](https://docs.soliditylang.org/en/v0.8.17/assembly.html#memory-safety) because of:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b9d504c09b7a4bae90a86145006bc7ab52136807/contracts/proxy/Proxy.sol#L25-L28

There is even a note in the documentation to clear up potential doubts about this assembly block:
>these restrictions still need to be followed, even if the assembly block reverts or terminates.

wiasliaw:
Thanks, got it

PR #3834, opened 2022-11-27
Update ERC1155Supply.sol

haolong:
//Adds isALIVE parameter to avoid misunderstanding and improve the usability

Fixes #???? 

Adds isALIVE parameter to avoid misunderstanding and improve the usability.

When we are tracking supplies, it is important to know whether we want to include tokens that are dead or alive. Just a simple totalSupply is misleading when it is not explicitly explained as totalSupply of only the existing tokens. This new parameter provides a quick fix that allows smart contract developers to know exactly what they will get when they code.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC1155/extensions/ERC1155Supply.sol
+++ b/contracts/token/ERC1155/extensions/ERC1155Supply.sol
@@ -14,13 +14,18 @@ import "../ERC1155.sol";
  * same id are not going to be minted.
  */
 abstract contract ERC1155Supply is ERC1155 {
-    mapping(uint256 => uint256) private _totalSupply;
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @haolong 

Please open an issue to discuss the need/usecase for this before submitting a PR. 

Your proposed change significantly increase the gas cost, with unclear upside. IMO we should not propose that by default, as most user wont need this, and the few that do can easily be added to after the fact.

haolong:
Yes. Your point is reasonable. We should discuss more. Thank you for your review, Amxx!

PR #3838, opened 2022-11-28
Remove hooks from ERC20

### LINKED ISSUE #3535: Consider alternative design for token transfers and hooks

JulissaDantes:
Fixes #3535  

The current way to customize token transfers(mint, burn, transfer) is by overriding the hooks `_beforeTokenTransfer` and `_afterTokenTransfer`. This refactor removes both hooks from the `ERC20` contract, and implements the logic inside the `transfer` function, so future customizations only rely on overriding a single function.

#### PR Checklist

- [X] Tests
- [X] Documentation
- [x] Changelog entry


contracts/token/ERC20/extensions/ERC20Capped.sol
```diff
@@ -28,10 +28,14 @@ abstract contract ERC20Capped is ERC20 {
...
-     * @dev See {ERC20-_mint}.
+     * @dev See {ERC20-_transfer}.
      */
-    function _mint(address account, uint256 amount) internal virtual override {
+    function _transfer(
+        address,
+        address account,
+        uint256 amount
+    ) internal virtual override {
         require(ERC20.totalSupply() + amount <= cap(), "ERC20Capped: cap exceeded");
```
frangio, line 38: We only need to do this on mint.

contracts/token/ERC20/extensions/ERC20Capped.sol
```diff
@@ -28,10 +28,14 @@ abstract contract ERC20Capped is ERC20 {
...
-        super._mint(account, amount);
+        super._transfer(address(0), account, amount);
```
frangio, line 39: This is not right, we have to invoke super with the exact same arguments.

test/token/ERC20/ERC20.test.js
```diff
@@ -203,7 +203,14 @@ contract('ERC20', function (accounts) {
...
     it('rejects a null account', async function () {
       await expectRevert(
-        this.token.mint(ZERO_ADDRESS, amount), 'ERC20: mint to the zero address',
+        this.token.mint(ZERO_ADDRESS, amount), 'ERC20: invalid transfer operation',
+      );
+    });
+
+    it('rejects overflow', async function () {
+      const maxUint256 = new BN('2').pow(new BN(256)).subn(1);
+      await expectRevert.unspecified(
```
frangio, line 212: This shouldn't be `unspecified`.

frangio (changes requested):

test/token/ERC20/ERC20.behavior.js
```diff
@@ -240,14 +225,6 @@ function shouldBehaveLikeERC20Transfer (errorPrefix, from, to, balance, transfer
       });
     });
   });
-
-  describe('when the recipient is the zero address', function () {
```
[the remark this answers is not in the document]
JulissaDantes: Since everything goes through the transfer, it cannot know if its coming from mint or burn, so `transfer` has a check to ensure that `from` and `to` cannot be the zero address, but if a user sends a mint to address zero it would received both `from` and `to` as the zero address. There is a test for that case on [ln 206](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/27c7090480c2dc26f361def7b4a62960a591f55b/test/token/ERC20/ERC20.test.js#L206)

test/token/ERC20/ERC20.behavior.js
```diff
@@ -148,21 +148,6 @@ function shouldBehaveLikeERC20 (errorPrefix, initialSupply, initialHolder, recip
           });
         });
       });
-
-      describe('when the recipient is the zero address', function () {
```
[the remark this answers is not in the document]
JulissaDantes: Since everything goes through the transfer, it cannot know if its coming from mint or burn, so `transfer` has a check to ensure that `from` and `to` cannot be the zero address, but if a user sends a mint to address zero it would received both `from` and `to` as the zero address. There is a test for that case on [ln 206](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/27c7090480c2dc26f361def7b4a62960a591f55b/test/token/ERC20/ERC20.test.js#L206)

frangio:
There is a separate point we need to decide, which is whether the external `transfer` and `transferFrom` functions should accept `to = address(0)`.

I think this PR should keep the current behavior, and so we need to add in those external functions the necessary `require` statements.

contracts/token/ERC20/ERC20.sol
```diff
@@ -224,7 +227,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
         address to,
         uint256 amount
     ) internal virtual {
-        require(from != address(0) || to != address(0), "ERC20: invalid transfer operation");
+        require(!(from == address(0) && to == address(0)), "ERC20: invalid transfer operation");
 
         uint256 fromBalance = _balances[from];
```
frangio, line 232: We are reading this from storage even during mints when it's not necessary. Please move this inside the branches where it needs to be read.

contracts/token/ERC20/ERC20.sol
```diff
@@ -237,11 +240,12 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
...
             unchecked {
-                _balances[from] -= amount;
+                // Overflow not possible: balance - amount is at most totalSupply - amount, which is checked above.
+                _balances[from] = fromBalance - amount;
             }
         } else {
             require(fromBalance >= amount, "ERC20: transfer amount exceeds balance");
-            _balances[from] -= amount;
+            _balances[from] = fromBalance - amount;
             _balances[to] += amount;
```
frangio, line 249: This can be inside unchecked.

test/token/ERC20/ERC20.behavior.js
```diff
@@ -148,6 +148,21 @@ function shouldBehaveLikeERC20 (errorPrefix, initialSupply, initialHolder, recip
...
+        const to = ZERO_ADDRESS;
+
+        beforeEach(async function () {
+          await this.token.approve(spender, amount, { from: tokenOwner });
+        });
+
+        it('reverts', async function () {
+          await expectRevert(this.token.transferFrom(
+            tokenOwner, to, amount, { from: spender }), `${errorPrefix}: transfer to the zero address`,
+          );
```
frangio, line 163:
```suggestion
          await expectRevert(
            this.token.transferFrom(tokenOwner, to, amount, { from: spender }),
            `${errorPrefix}: transfer to the zero address`,
          );
```

CHANGELOG.md
```diff
@@ -10,6 +10,31 @@
  * `ProxyAdmin`: Removed `getProxyAdmin` and `getProxyImplementation` getters. ([#3820](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3820))
  * `ERC20`: Deleted `_beforeTokenTransfer` and `_afterTokenTransfer` hooks, and refactor all extensions using those hooks for customization, by using `transfer` instead. ([#3838](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3838))
 
+### How to upgrade from 4.x
+These breaking changes will require additional modifications in your implementation of ERC20, ERC721, and ERC1155, since the `afterTokenTransfer` and `beforeTokenTransfer` were removed. Any customization made through those hooks should now use the `transfer` function instead, keeping in mind that `_mint` and `_burn` are no longer virtual, so they are not overridable.
+
+For example, a contract using `ERC20`'s `afterTokenTransfer` hook would have to be changed in the following way.
```
frangio, line 16:
```suggestion
For example, a contract using `ERC20`'s `_afterTokenTransfer` hook would have to be changed in the following way.
```

CHANGELOG.md
```diff
@@ -10,6 +10,31 @@
...
+-         require(!condition(), "ERC20: wrong condition");
+-     }
++function _transfer(
++         address from,
++         address to,
++         uint256 amount
++     ) internal virtual override {
++         require(!condition(), "ERC20: wrong condition");
++         super._transfer(from, to, amount);
++     }
```
frangio, line 35: This should be simpler and clearer:

```diff
- function _beforeTokenTransfer(
+ function _transfer(
      address from,
      address to,
      uint256 amount
  ) internal virtual override {
-     super._beforeTokenTransfer(from, to, amount);
      require(!condition(), "ERC20: wrong condition");
+     super._transfer(from, to, amount);
  }
```

The below suggestion implements this

```suggestion
- function _beforeTokenTransfer(
+ function _transfer(
      address from,
      address to,
      uint256 amount
  ) internal virtual override {
-     super._beforeTokenTransfer(from, to, amount);
      require(!condition(), "ERC20: wrong condition");
+     super._transfer(from, to, amount);
  }
```

frangio (changes requested):

frangio:
I am concerned about risks from accepting address 0 as an argument. For context see [forum thread](https://forum.openzeppelin.com/t/removing-address-0x0-checks-from-openzeppelin-contracts/2222/13?u=frangio).

Basically imagine that someone implements a function that transfers using a signature, with a line such as `address from = ecrecover(...)` without checking that the result is nonzero and later using it in `_transfer(from, to, amount)`. This hypothetical piece of code would be vulnerable to an attack where an invalid signature is submitted that results in `from == address(0)` and as a result mints tokens instead of transferring.

This risk is inherent in having a function that accepts address 0 as a valid argument, but this is the design that we're planning so we can get rid of hooks. What I would suggest is to rename our current function `_transfer` into something else like `_update`, and having a `_transfer` function that calls `_update` after checking that the arguments are nonzero. In this way we preserve the current semantics of `_transfer` in 4.x instead of introducing a new risk unnecessarily.

CHANGELOG.md
```diff
@@ -8,31 +8,29 @@
...
-+     }
+- function _beforeTokenTransfer(
++ function _transfer(
+      address from,
+      address to,
+      uint256 amount
+  ) internal virtual override {
+-     super._beforeTokenTransfer(from, to, amount);
+      require(!condition(), "ERC20: wrong condition");
++     super._transfer(from, to, amount);
```
frangio, line 32:
```suggestion
+ function _update(
      address from,
      address to,
      uint256 amount
  ) internal virtual override {
-     super._beforeTokenTransfer(from, to, amount);
      require(!condition(), "ERC20: wrong condition");
+     super._update(from, to, amount);
```

CHANGELOG.md
```diff
@@ -8,31 +8,29 @@
...
 
 ### How to upgrade from 4.x
-These breaking changes will require additional modifications in your implementation of ERC20, ERC721, and ERC1155, since the `afterTokenTransfer` and `beforeTokenTransfer` were removed. Any customization made through those hooks should now use the `transfer` function instead, keeping in mind that `_mint` and `_burn` are no longer virtual, so they are not overridable.
 
-For example, a contract using `ERC20`'s `afterTokenTransfer` hook would have to be changed in the following way.
+#### ERC20, ERC721, and ERC1155
+
+These breaking changes will require modifications to ERC20, ERC721, and ERC1155 contracts, since the `_afterTokenTransfer` and `_beforeTokenTransfer` functions were removed. Any customization made through those hooks should now be done overriding the `_transfer` function instead.
+
+Minting and burning are implemented by `_transfer` and customizations should be done by overriding this function as well. `_mint` and `_burn` are no longer virtual (meaning they are not overridable) to guard against possible inconsistencies.
```
frangio, line 19:
```suggestion
These breaking changes will require modifications to ERC20, ERC721, and ERC1155 contracts, since the `_afterTokenTransfer` and `_beforeTokenTransfer` functions were removed. Any customization made through those hooks should now be done overriding the new `_update` function instead.

Minting and burning are implemented by `_update` and customizations should be done by overriding this function as well. `_mint` and `_burn` are no longer virtual (meaning they are not overridable) to guard against possible inconsistencies.
```

contracts/token/ERC20/ERC20.sol
```diff
@@ -229,74 +229,73 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
...
+            require(fromBalance >= amount, "ERC20: burn amount exceeds balance");
+            _totalSupply -= amount;
+            unchecked {
+                // Overflow not possible: balance - amount is at most totalSupply - amount, which is checked above.
+                _balances[from] = fromBalance - amount;
+            }
+        } else {
+            require(fromBalance >= amount, "ERC20: transfer amount exceeds balance");
+            unchecked {
+                // Overflow not possible: fromBalance - amount is at most 0, which is checked above.
```
frangio, line 265: Not needed because of the `require` immediately above it. Sorry for the back and forth on this one, I realized our guidelines are actually that if there is the necessary `require` statement right before the unchecked operation we don't need to include a comment.

```suggestion
```

contracts/token/ERC20/ERC20.sol
```diff
@@ -110,6 +110,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
      * - the caller must have a balance of at least `amount`.
      */
     function transfer(address to, uint256 amount) public virtual override returns (bool) {
+        require(to != address(0), "ERC20: transfer to the zero address");
```
frangio, line 113: I believe not needed anymore.

```suggestion
```

contracts/token/ERC20/ERC20.sol
```diff
@@ -159,6 +160,8 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
         address to,
         uint256 amount
     ) public virtual override returns (bool) {
+        require(from != address(0), "ERC20: transfer from the zero address");
+        require(to != address(0), "ERC20: transfer to the zero address");
```
frangio, line 164:
```suggestion
```

contracts/token/ERC20/ERC20.sol
```diff
@@ -256,13 +253,12 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
             require(fromBalance >= amount, "ERC20: burn amount exceeds balance");
             _totalSupply -= amount;
             unchecked {
-                // Overflow not possible: balance - amount is at most totalSupply - amount, which is checked above.
+                // Overflow not possible: amount <= accountBalance <= totalSupply.
```
frangio, line 256:
```suggestion
                // Overflow not possible: amount <= fromBalance <= totalSupply.
```

frangio (approved):
LGTM

contracts/token/ERC20/extensions/ERC20Capped.sol
```diff
@@ -28,10 +28,17 @@ abstract contract ERC20Capped is ERC20 {
...
-    function _mint(address account, uint256 amount) internal virtual override {
-        require(ERC20.totalSupply() + amount <= cap(), "ERC20Capped: cap exceeded");
-        super._mint(account, amount);
+    function _update(
+        address from,
+        address to,
+        uint256 amount
+    ) internal virtual override {
+        if (from == address(0)) {
+            require(ERC20.totalSupply() + amount <= cap(), "ERC20Capped: cap exceeded");
```
Amxx, line 39: What is the rational for using `ERC20.totalSupply()` instead of `totalSupply()` ?

contracts/token/ERC20/extensions/ERC20Snapshot.sol
```diff
@@ -140,6 +138,7 @@ abstract contract ERC20Snapshot is ERC20 {
             _updateAccountSnapshot(from);
             _updateAccountSnapshot(to);
         }
```
Amxx, line 140: this part will not work properly if update from 0 to 0

test/token/ERC20/ERC20.behavior.js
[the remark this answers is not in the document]
Amxx: Line 206 is testing that you cannot mint to address 0. That is a valid test, but its separate from this. This tests that you cannot call transfer to the address 0. 

Tests are checking for behavior correctness. Its ok (and I think its needed) to have multiple tests that are resolved by the same require in different context. 

I think this test should have been kept.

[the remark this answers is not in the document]
Amxx: I wonder if that is the cause of the coverage decrease.

PR #3839, opened 2022-11-29
Add additional conditions when testing ERC4626 roundtrip

frangio:
We have a flaky test in the ERC4626 property tests due to a property that is not valid in some edge cases. Namely, when the vault loses all assets (through negative yield), the round trip `mint-withdraw` is not valid as specified. For now we filter out those edge cases.


JulissaDantes (approved):
Looks good!

PR #3849, opened 2022-12-02
Add ERC-4626 Upgrade Note

frangio:
As reported in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3639#issuecomment-1332253967, vaults that are upgraded to 4.8 need to initialize a new variable and we should include this in the changelog.


JulissaDantes (approved):
LGTM!

PR #3850, opened 2022-12-02
Introduce SequentialOperations abstraction for signature-based ops

k06a:
Beware that this PR is build on top of #3848 and includes its commits

#### PR Checklist

TBD

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


k06a:
Abstraction related to this implementation is also discussed here: https://github.com/ethereum/EIPs/pull/6077

Amxx:
Hello @k06a 

This looks like a very significant change. You are rightfully targeting next-5.0 as this would be breaking. Still, I believe we should have more discussion before jumping to an implementation. 

I appreciate that there is standardization effort making an ERC with that, but it is still in a very early stage, and likely to evolve. Maybe a good option would be to keep the current interface/approach, and "just" change the storage under the hood to be future proof and allow this to be implemented as a non-storage-breaking change during 5.x lifecycle.

There are other nonce related changes that we could implement. I'm thinking of [the timeline system described here](https://github.com/amxx/permit#out-of-order-execution).

Lets discuss features first. Don't worry, the 5.0 train is not leaving anytime soon :)

Amxx:
Note: we might reopen this based on the discussion in #3927

TrejGun:
@k06a please explain where operationHash comes from?

k06a:
@TrejGun it depends on specific operation EIP/definition.

TrejGun:
@k06a can you show any example?

k06a:
@TrejGun my bad, you should use EIP-712 for structured data hash.

TrejGun:
@k06a 
Ok I will ask again)

did you abstract this code as operationHash?
```
       keccak256(
          abi.encode(
            SOME_TYPEHASH,
            same_variable1,
            same_variable2,
            ...
          )
        )
```

and hash remains the same for validation with signature and with VRS

TrejGun:
I ask this because I really like this abstraction and it does not look like you are doing this in solidity so maybe in JS and I want to understand how. For now I use ethers._signTypedData that will return _hashTypedDataV4(operationHash) but not just operationHash. How can I generate it?

k06a:
@TrejGun maybe this would work for you: https://github.com/ethers-io/ethers.js/blob/ce8f1e4015c0f27bf178238770b1325136e3351a/packages/hash/src.ts/typed-data.ts#L371

TrejGun:
thanks

PR #3851, opened 2022-12-02
Add SafeERC20.forceApprove()

k06a:
Fixes LIB-705

#### PR Checklist

TBD

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/utils/SafeERC20.sol
+++ b/contracts/token/ERC20/utils/SafeERC20.sol
@@ -58,6 +58,20 @@ library SafeERC20 {
         _callOptionalReturn(token, abi.encodeWithSelector(token.approve.selector, spender, value));
     }
 
+    function forceApprove(
+        IERC20 token,
+        address spender,
+        uint256 value
+    ) internal {
+        if (!_callOptionalReturnBool(token, abi.encodeWithSelector(token.approve.selector, spender, value))) {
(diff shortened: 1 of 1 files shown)


frangio:
Please provide more detail. If we offer this function, what should we recommend to developers? Should they always use this function when they need to call approve? Are there circumstances where they shouldn't use it? Are there risks they should consider?

k06a:
@frangio if your contract is performing token approve and expecting to spend it later in the same transaction This `forceApprove` is the cheapest way to do so without any allowance pre-checks. In best scenario it will only perform approve(amount), in worst scenario it will make 3 calls: approve(amount), approve(0), approve(amount). So this methods do best effort to set token approve to desired amount.

frangio:
My question is different: should we be recommending that everyone use `forceApprove`?

k06a:
@frangio I think it depends on the situation. If developer is going to spend this approve in the same transaction, then yes, I would recommend to use it. If this approve should be spent by someone else in different transaction then I would suggest to use safe increase/decrease.

frangio:
Great! That recommendation makes a lot of sense to me. We should include this in the docs.

Amxx:
This targets next-5.0. Since we want it in 4.9 we need to target master.

Replacing this PR with #4067

PR #3852, opened 2022-12-03
Remove unused variables in tests

itinance:
Removes some **unused** 
 - variables
 - import statements
 - and constants 
in Tests.

While investigating the unit tests, I came across some things that are not used anywhere and probably result from copy-paste behavior. To keep the codebase clean, I'd like to submit this small PR

#### PR Checklist

- [X] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio (approved):
Thanks!

PR #3855, opened 2022-12-04
Force Slither to use Hardhat

frangio:


frangio:
I can't find a way to disable Foundry installation by slither-action.

PR #3863, opened 2022-12-07
Add ERC721 Wrapper

ernestognw:
Fixes LIB-647

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -0,0 +1,51 @@
...
+    IERC721 public immutable underlying;
+
+    constructor(IERC721 underlyingToken) {
+        underlying = underlyingToken;
+    }
+
+    /**
+     * @dev Allow a user to deposit an underlying token and mint the corresponding tokenId.
+     */
+    function depositFor(address account, uint256 tokenId) public virtual returns (bool) {
```
JulissaDantes, line 27: Since this would only allow for a single wrapped token creation at a time, shouldn't we add a batch mechanism to allow multiple creation given an array of tokenIds?

JulissaDantes: Same for `recover`

test/token/ERC721/ERC721.behavior.js
```diff
@@ -328,7 +329,7 @@ function shouldBehaveLikeERC721 (errorPrefix, owner, newOwner, approved, another
 
         describe('to a contract that does not implement the required function', function () {
           it('reverts', async function () {
-            const nonReceiver = this.token;
```
JulissaDantes, line 331: Why are we not using the same token deployed from `ERC721Wrapper.test.js` here and deploying a new token instead?

ernestognw: The behavioral tests assume that `this.token` is not an ERC721 receiver, however, the wrapper is required to be a receiver so this assumption breaks.

Here I'm creating any nonreceiver contract just to fix it. By default I'm creating an ERC721 to keep the past assumption consistent with previous tests.

contracts/token/ERC721/extensions/ERC721Wrapper.sol
frangio: If someone needs to migrate multiple tokens, it would be cheaper to do it all in one transaction even if there is no native batch transfer in ERC721 (due to warm vs. cold state). Batching could be handled either by `Multicall` or by making the argument to this function an array of `tokenIds`. I feel like the latter is simple enough that I would make that change. Do we see reasons not to?

test/token/ERC721/ERC721.behavior.js
JulissaDantes: Ah, I understand.

contracts/token/ERC721/extensions/ERC721Wrapper.sol
ernestognw: Make sense to me. I added `tokenIds` to both `depositFor` and `withdrawTo` as discussed.
Updated tests accordingly.

TrejGun:
this definitely should be extended to support erc20 and erc1155

ernestognw:
> this definitely should be extended to support erc20 and erc1155

[ERC20Wrapper](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Wrapper.sol) already exists.

The case for ERC1155 might need more discussion imo.

TrejGun:
I mean erc721 wrapper for erc20

Amxx:
I would add a "onERC721Received" hook, so you can wrap by doing a safeTransfer to this.

```
function onERC721Received(address, address from, uint256 tokenId, bytes calldata data) external returns (bytes4) {
    require(msg.sender == address(underlying));
    _safeMint(data.length == 0 ? from : abi.decode(data, (address)), tokenIds);
    return IERC721Receiver.onERC721Received.selector;
}
```

contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -0,0 +1,55 @@
...
+            underlying.safeTransferFrom(_msgSender(), address(this), tokenIds[i]);
+            _safeMint(account, tokenIds[i]);
+        }
+        return true;
+    }
+
+    /**
+     * @dev Allow a user to burn wrapped tokens and withdraw the corresponding tokenIds of the underlying tokens.
+     */
+    function withdrawTo(address account, uint256[] memory tokenIds) public virtual returns (bool) {
```
Amxx, line 38: WOW, this function doesn't have any sort of access restriction!

We need to check that `_msgSender` is `isApprovedOrOwner`

ernestognw: Same for [ERC20Wrapper](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Wrapper.sol#L48). The check is assumed to be performed in `safeTransferFrom`

Amxx: No, 

`underlying.safeTransferFrom` will check that the caller (this contract) is approved of owner of the underlying token ... which is the case because the underlying token is held by the wrapper.

What you need to check is that the caller has authority over the the wrapped token.

In ERC20Wrapper, this is done by burning from the _msgSender() balance. You can only burn your token to unwrap them. Here anyone can burn any token, and unwrap them to any token

Amxx: Just add a test where an account tries to unwrap a token it doesn't own

ernestognw: Ahh got it, thanks for flagging. I assumed _burn was restricted for some reason.
Added also tests accordingly.

Amxx:
> > I would add a "onERC721Received" hook, so you can wrap by doing a safeTransfer to this.
> > ```
> > function onERC721Received(address, address from, uint256 tokenId, bytes calldata data) external returns (bytes4) {
> >     require(msg.sender == address(underlying));
> >     _safeMint(data.length == 0 ? from : abi.decode(data, (address)), tokenIds);
> >     return IERC721Receiver.onERC721Received.selector;
> > }
> > ```
> 
> Not sure about this. It'll require to use `_transfer()` inside `depositFor` instead of `safeTransferFrom()`, otherwise `_safeMint` will be called twice. We'll be exposing our users to the same reason why there's no `transfer()` in the ERC721.
> 
> Also, if we decide to only keep `onERC721Received`, we'll lose the batch minting allowed in `depositFor`.
> 
> What do you think?

If we add this hook, we can keep safeTransfer in the `depositFor` function, with the right "data", and remove the `_mint` from deposit. The mint will be handled by the hook.

Amxx:
```
function depositFor(address account, uint256[] memory tokenIds) public virtual returns (bool) {
    bytes memory data = abi.encode(account);
    for (uint256 i = 0; i < tokenIds.length; ++i) {
        underlying.safeTransferFrom(_msgSender(), address(this), tokenIds[i], data); // This will trigger the hook
    }
    return true;
}

// The hook will mint
function onERC721Received(address, address from, uint256 tokenId, bytes calldata data) external returns (bytes4) {
    require(msg.sender == address(underlying));
    _safeMint(data.length == 0 ? from : abi.decode(data, (address)), tokenIds);
    return IERC721Receiver.onERC721Received.selector;
}
```

Amxx:
OR, we can use "unsafeTransfer" in deposit, which is ok because we know the receiver is able to handle NFTs

ernestognw:
> // The hook will mint
> function onERC721Received(address, address from, uint256 tokenId, bytes calldata data) external returns (bytes4) {
>     require(msg.sender == address(underlying));
>     _safeMint(data.length == 0 ? from : abi.decode(data, (address)), tokenIds);
>     return IERC721Receiver.onERC721Received.selector;
> }

I have to admit I'm a fan of this solution. So I'll go for it.

Before implementing, I wonder, is there a case for keeping `_recover`? EIP-721 requires every application/wallet/holder to implement `onERC721Received` so `_recover` won't be necessary if this is used for a compliant ERC721.

ernestognw:
> I have to admit I'm a fan of this solution. So I'll go for it.
> 
> Before implementing, I wonder, is there a case for keeping `_recover`? EIP-721 requires every application/wallet/holder to implement `onERC721Received` so `_recover` won't be necessary if this is used for a compliant ERC721.

For the record, `transferFrom` is standard behavior and doesn't enforce `onERC721Received` check, so `_recover` should be kept for general unsafe transfers.

Amxx (approved):
LGTM,

Waiting for @frangio second review

docs/modules/ROOT/pages/governance.adoc
```diff
@@ -119,7 +119,7 @@ contract MyToken is ERC20, ERC20Permit, ERC20Votes, ERC20Wrapper {
 }
 ```
 
-NOTE: Voting power could be determined in different ways: multiple ERC20 tokens, ERC721 tokens, sybil resistant identities, etc. All of these options are potentially supported by writing a custom Votes module for your Governor. The only other source of voting power available in OpenZeppelin Contracts currently is xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`].
+NOTE: Voting power could be determined in different ways: multiple ERC20 tokens, ERC721 tokens, sybil resistant identities, etc. All of these options are potentially supported by writing a custom Votes module for your Governor. The other sources of voting power available in OpenZeppelin Contracts currently are xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`] and xref:api:token/ERC721Wrapper.adoc#ERC721Wrapper[`ERC721Wrapper`].
```
Amxx, line 122: No, `ERC721Wrapper` does not provide the voting interface that governance need.

You could use a combo of `ERC721Wrapper` and `ERC721Votes` to wrap non-voting NFTs into voting NFTs ... but the voting part will come from inheriting `ERC721Votes`

ernestognw: Thanks for bringing it up! In fact, I want to find a place in the governance docs to share visibility about it.

I see there's a section for ERC20Wrapper [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/77fcda8f09b5fc7c72d5f91cdb3b38bce52a0c7c/docs/modules/ROOT/pages/governance.adoc#setup) but includes a code example

```markdown
If your project already has a live token that does not include ERC20Votes and is not upgradeable, you can wrap it in a governance token by using ERC20Wrapper. This will allow token holders to participate in governance by wrapping their tokens 1-to-1.

<Code>
```

I think a code example for ERC721Wrapper is way too much, so what about this?:

```suggestion
NOTE: Voting power could be determined in different ways: multiple ERC20 tokens, ERC721 tokens, sybil resistant identities, etc. All of these options are potentially supported by writing a custom Votes module for your Governor. The only other source of voting power available in OpenZeppelin Contracts currently is xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`] (optionally wrapped in xref:api:token/ERC721Wrapper.adoc#ERC721Wrapper[`ERC721Wrapper`]).
```

Amxx: I would say.

The only other source of voting power available in OpenZeppelin Contracts currently is xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`]. NFT tokens that don't provide this functionnality can be wrapped into voting token isn't a combination of xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`] and xref:api:token/ERC721Wrapper.adoc#ERC721Wrapper[`ERC721Wrapper`].

ernestognw: Added with [minor spelling changes](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3863/commits/24ac3dc5f09949532ec843bb899dfd891663be4f)

docs/modules/ROOT/pages/governance.adoc
```diff
@@ -119,7 +119,7 @@ contract MyToken is ERC20, ERC20Permit, ERC20Votes, ERC20Wrapper {
 }
 ```
 
-NOTE: Voting power could be determined in different ways: multiple ERC20 tokens, ERC721 tokens, sybil resistant identities, etc. All of these options are potentially supported by writing a custom Votes module for your Governor. The other sources of voting power available in OpenZeppelin Contracts currently are xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`] and xref:api:token/ERC721Wrapper.adoc#ERC721Wrapper[`ERC721Wrapper`].
+NOTE:The only other source of voting power available in OpenZeppelin Contracts currently is xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`]. NFT tokens that don't provide this functionality can be wrapped into a voting token isn't a combination of xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`] and xref:api:token/ERC721Wrapper.adoc#ERC721Wrapper[`ERC721Wrapper`].
```
Amxx, line 122:
```suggestion
NOTE:The only other source of voting power available in OpenZeppelin Contracts currently is xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`]. ERC721 tokens that don't provide this functionality can be wrapped into a voting tokens using a combination of xref:api:token/ERC721.adoc#ERC721Votes[`ERC721Votes`] and xref:api:token/ERC721Wrapper.adoc#ERC721Wrapper[`ERC721Wrapper`].
```

1
```diff
@@ -0,0 +1,11 @@
+Remove warning
```
frangio, line 1: This file should be removed.

contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -0,0 +1,83 @@
...
+ * @dev Extension of the ERC721 token contract to support token wrapping.
+ *
+ * Users can deposit and withdraw an "underlying token" and receive a "wrapped token" with a matching tokenId. This is useful
+ * in conjunction with other modules. For example, combining this wrapping mechanism with {ERC721Votes} will allow the
+ * wrapping of an existing "basic" ERC721 into a governance token.
+ *
+ * _Available since v4.9.0_
+ */
+abstract contract ERC721Wrapper is ERC721, ERC721Holder {
+    IERC721 public immutable underlying;
```
frangio, line 18: All state variables are private. This seems fine if it's immutable, but in the upgradeable version it will become a storage variable so we would want it to be private so it can't be set by children contracts.

contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -0,0 +1,83 @@
...
+     * for recovering in that scenario.
+     */
+    function onERC721Received(
+        address,
+        address from,
+        uint256 tokenId,
+        bytes memory data
+    ) public override returns (bytes4) {
+        require(msg.sender == address(underlying));
+        _safeMint(data.length == 0 ? from : abi.decode(data, (address)), tokenId);
```
frangio, line 70: This seems risky to me. How do we know the transfer initiator isn't just passing some other data in here for whatever reason?

Instead of assuming any non-empty data encodes a receiver address, I would prefer if we required some magic value to be present, like a 4 byte selector, at the beginning of the data.

frangio (changes requested):

contracts/token/ERC721/extensions/ERC721Wrapper.sol
ernestognw: Moved and [created an issue for ERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4026)

contracts/token/ERC721/extensions/ERC721Wrapper.sol
frangio: Would like to hear @Amxx's thoughts on this.

ernestognw: > How do we know the transfer initiator isn't just passing some other data in here for whatever reason?

The `underlying` is the only one who can call `onERC721Received`, so whoever is deploying this should make sure the underlying pass the correct data and we can't enforce. So I'd say the `underlying` is trusted. [The data is encoded in the `depositFor` function.](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3863/files#diff-0b3d65cc80b6f6aa8b2eec2949c36d3eba9035fa132d90d367cc62291d97dafdR28)

> I would prefer if we required some magic value to be present, like a 4 byte selector, at the beginning of the data.

We can append ERC721Wrapper's `interfaceId` with a value like:

```solidity
    bytes32 private constant WRAPPER_AWARE = type(ERC721Wrapper).interfaceId;
```

and use it in depositFor as:

```solidity
        bytes memory data = abi.encode(WRAPPER_AWARE, account);
```

Then just check the prefix in `onERC721Check`.

Also, I made sure they produce different interfaceIds:

```solidity
pragma solidity ^0.8.0;

interface A {
    function interfaceId() external pure returns(bytes4);
}

abstract contract B is A {
    function interfaceId() external pure override returns(bytes4) {
        return type(A).interfaceId;
    }

    function anotherInterfaceId() public pure returns(bytes4) {
        return type(B).interfaceId;
    }
}

contract C is B {}
```

To be honest, I think it's too much, but will wait for comments from @Amxx

frangio: >so whoever is deploying this should make sure the underlying pass the correct data

This is not something anyone can guarantee. ERC721 allows any token holder to initiate a transfer with arbitrary data attached to it, to be passed through to `onERC721Received`. Just as ERC721Wrapper is encoding the address in `depositFor`, a wallet may decide to include anything in there too. It would be slightly reckless to do so because it could have consequences, but I wouldn't find it too surprising to learn of some wallet that includes data to track wallet use or something.

frangio: > We can append ERC721Wrapper's `interfaceId` with a value like:

This sounds good if we go forward with the magic value. Note that it would be more efficient to use `abi.encodePacked` rather than `abi.encode`.

ernestognw: > I wouldn't find it too surprising to learn of some wallet that includes data to track wallet use or something.

This is a good example of a use case. Thanks!

> This sounds good if we go forward with the magic value.

I realized it may not be backwards compatible if we introduce a new function to the interface, so I decided to use a hardcoded bytes4 value.

4 bytes are arbitrary, though. We can use the last 12 bytes to pack it with the address. Keeping it 4 bytes will keep the risk of a collision in 1/2**32.

[Here's the implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3863/commits/fb22e8dfec224f49bb6730846e3a3dd929332a73)

> abi.encodePacked is even more efficient but in that case I don't think we can use abi.decode, this may be the best choice though.

I used `abi.encodePacked` with a decoding private function. Why would you keep `abi.encodeWithSelector`?

frangio: In your implementation you require the magic value to be present even when the token should be minted to the sender. I don't think this is necessary. It's safe to assume that the sender wants to wrap the token and receive it in the same account if data is empty.

ernestognw: [Updated and simplified.](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3863/commits/24b103306346c8511b8ea38b1047888aa0ae8c51)

contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -0,0 +1,101 @@
...
+ * wrapping of an existing "basic" ERC721 into a governance token.
+ *
+ * _Available since v4.9.0_
+ */
+abstract contract ERC721Wrapper is ERC721, ERC721Holder {
+    IERC721 private immutable _underlying;
+
+    // Kept as bytes12 so it can be packed with an address
+    // Equal to 0xb125e89df18e2ceac5fd2fa8
+    bytes12 private constant _WRAPPER_ACCEPT_MAGIC = bytes12(keccak256("WRAPPER_ACCEPT_MAGIC"));
```
Amxx, line 22: I'd make that public constant.

ernestognw: Not sure. I see multiple `private constant` and `public constant` across the repo with no clear decision criteria.
In this case, the only reason I see to put it is so other contracts can be aware of the value, however, if they're implementing the `_WRAPPER_ACCEPT_MAGIC`, they should already know its value beforehand.

The magic value is not standard afaik so there's no discovery use case in which a contract _figures out_ there's a magic value, is it?

Amxx: IMO making it public contributes to discoverability (by users in etherscan, by frontend). This is particularly when the value is not standard. 

For example, in `AccessControl`, `DEFAULT_ADMIN_ROLE` is a public constant. Same for the roles in `TimelockController`.

You could imagine trying to call this, and detect if the contract is a wrapper based on the result (does it revert?)

contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -17,6 +17,10 @@ import "../utils/ERC721Holder.sol";
 abstract contract ERC721Wrapper is ERC721, ERC721Holder {
     IERC721 private immutable _underlying;
 
+    // Kept as bytes12 so it can be packed with an address
+    // Equal to 0xb125e89df18e2ceac5fd2fa8
+    bytes12 public constant WRAPPER_ACCEPT_MAGIC = bytes12(keccak256("WRAPPER_ACCEPT_MAGIC"));
```
frangio, line 22: Can you check if Solidity is evaluating the constant at compile time or runtime?

ernestognw: It's written at the bytecode but only once:

```solidity
contract ERC721Wrapped is ERC721Wrapper {
    constructor(address underlying) ERC721Wrapper(IERC721(underlying)) ERC721("Wrapped", "WRPD") {}
}
```

It appears only once in the bytecode and it always trims it:

```solidity
JUMPDEST
PUSH2 0x01fe
PUSH32 0xb125e89df18e2ceac5fd2fa8b7887eec71a2a2c0a2378e293dccc6be1aef7449
DUP2
JUMP
JUMPDEST
PUSH1 0x40
MLOAD
PUSH1 0x01
PUSH1 0x01
PUSH1 0xa0
SHL
SUB
NOT
SWAP1
SWAP2
AND
DUP2
```

One place reaching this code is the `onERC721Received` function, which selector is located exactly before the selector of the second case `WRAPPER_ACCEPT_MAGIC`

Every other instance is hardcoded like (idk why Solidity does this lmao):

```solidity
PUSH12 0x1624bd13be31c59d58bfa5f5
PUSH1 0xa3
SHL
```

contracts/token/ERC721/extensions/ERC721Wrapper.sol
frangio: I think what @Amxx meant by "does it revert?" was that in order to check if a contract is a wrapper you can call `WRAPPER_ACCEPT_MAGIC()` and check if it reverts: if it doesn't revert you can assume that it's a wrapper.

I don't think we need to worry about discoverability of magic values as a general problem.

It's good to make the value publicly accessible as a convenience and that's all we need to care about at this point IMO.

frangio (approved):
LGTM

contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -0,0 +1,101 @@
...
+    function onERC721Received(
+        address,
+        address from,
+        uint256 tokenId,
+        bytes memory data
+    ) public override returns (bytes4) {
+        require(address(underlying()) == _msgSender(), "ERC721Wrapper: caller is not underlying");
+        if (data.length == 32 && WRAPPER_ACCEPT_MAGIC == bytes12(data)) {
+            from = address(bytes20(bytes32(data) << 96));
+        }
```
Amxx, line 80: In the current implementation, if the first 12 bytes are not the right ones we disregard it completely, and mint to the original from. I'd revert.

```suggestion
        if (data.length == 32) {
            (bytes12 magic, from) = abi.decode(data, (bytes12, address));
            require(magic == WRAPPER_ACCEPT_MAGIC, "ERC721Wrapper: invalid data format");
        }
```
We may also want to revert if the data.length is > 32

Amxx: I would even add 
```solidity
} else {
    require(data.length == 0, "ERC721Wrapper: invalid data format");
}

Comment (ernestognw) [reply contracts/token/ERC721/extensions/ERC721Wrapper.sol:80]: The case in which the data is malformed is still safe because it mints to the original owner, so I wouldn't worry too much (he can always unwrap), and we allow noops, any reasons for adding this?

Comment (Amxx) [reply contracts/token/ERC721/extensions/ERC721Wrapper.sol:80]: If the sender is a smart contract, that thinks it unwrapped to someone else, and remove the token for its internal tracking, that might be bad.

As a general rule I think that if we ever get extra instruction that we can't process we should refuse to do anything, letting the caller know that the instructions are not clear.

Comment (Amxx) [reply contracts/token/ERC721/extensions/ERC721Wrapper.sol:80]: @frangio ?

Comment (frangio) [reply contracts/token/ERC721/extensions/ERC721Wrapper.sol:80]: For context, the reason we added the magic value in the first place was because we didn't think we could make any assumption about the data that a sender puts in the data field. I proposed a possible scenario where a wallet uses the data field to track transfers done with that wallet software (independently of whether that's a good idea). In that case we wouldn't want to accidentally transfer to a random location because the sender happened to put data in there.

In that scenario, where the data is used for tracking purposes, my thinking was we would want to ignore the data if it doesn't contain the magic value.

However, I find this argument more convincing:
>if we ever get extra instruction that we can't process we should refuse to do anything, letting the caller know that the instructions are not clear.

So I would accept @Amxx's suggestion.

Comment (ernestognw) [reply contracts/token/ERC721/extensions/ERC721Wrapper.sol:80]: I'm moving forward with the suggestion but I just wanted to share a meme and see if everybody gets happier plz

325669940_2200356000148602_3945984664946041884_n

**Review by frangio — APPROVED** [review]

File: test/token/ERC721/extensions/ERC721Wrapper.test.js [hunk 12]
```diff
@@ -0,0 +1,338 @@
...
+      await expectEvent.inTransaction(tx, this.underlying, 'Transfer', {
+        from: this.token.address,
+        to: initialHolder,
+        tokenId: firstTokenId,
+      });
+      await expectEvent.inTransaction(tx, this.token, 'Transfer', {
+        from: initialHolder,
+        to: constants.ZERO_ADDRESS,
+        tokenId: firstTokenId,
+      });
```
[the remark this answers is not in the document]
ernestognw: Tested in another branch and it passes. I'll solve this but open a PR only with this change to preserve approval.

PR #3864, opened 2022-12-08
✍️ IERC20: use `amount` instead of `value`

z0r0z:
✍️  use `amount` instead of `value` for event params to make match underlying functions

- [x] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/token/ERC20/IERC20.sol
+++ b/contracts/token/ERC20/IERC20.sol
@@ -8,18 +8,18 @@ pragma solidity ^0.8.0;
  */
 interface IERC20 {
     /**
-     * @dev Emitted when `value` tokens are moved from one account (`from`) to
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @z0r0z 

[The EIP](https://eips.ethereum.org/EIPS/eip-20) uses value and not amount. I agree that amount would have been a better choice of words, but so far we favored consistency with the EIP wording.

@frangio any opinion?

frangio:
I think this PR is pointing out that the events and the functions use a different word for the same concept. Perhaps we should use "value" instead of "amount", but I would agree that it should be consistent.

The ERC20 implementation uses "amount". Should we change it all to "value"?

ernestognw:
Closed by #4938

PR #3867, opened 2022-12-11
Update Ownable.sol

ShobhitVishnoi30:
Reduces the size of Ownable library.

In the _transferOwnership() function we can replace "address oldOwner = _owner" => "address oldOwner = owner()"

 [ ] Tests
 [ ] Documentation
 [ ] Changelog entry


frangio:
This doesn't seem to have any effect on code size according to [our benchmarks](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/3668217128).

PR #3869, opened 2022-12-11
Use unchecked in MerkleProof.processMultiProof

pcaversaccio:
As title. The array index cannot become negative by design (i.e. cannot underflow) so I would suggest using unchecked arithmetic for the `totalHashes - 1` expression. Also, this PR fixes a small typo. Since it's a refactor PR, I do not include a `CHANGELOG` entry.


frangio (approved):
I reduced the scope of `unchecked` to only what is needed. Did not include a comment to explain `unchecked` because the reason is found in the immediately prior line.

pcaversaccio:
Makes sense, thx @frangio.

ernestognw (approved):
Thanks for the catch @pcaversaccio

frangio:
@pcaversaccio You might be the first to mint this GitPOAP. :eyes:

pcaversaccio:
> @pcaversaccio You might be the first to mint this GitPOAP. 👀

oh nice :-D - actually just found out that am eligible for two others as well :) 

image

PR #3872, opened 2022-12-13
Virtual DOMAIN_SEPARATOR

jaketimothy:
The inability to override `DOMAIN_SEPARATOR` makes `ERC20Permit` and `Votes` incompatible.  When inheriting from both contracts, solidity requires an override to disambiguate.  Making these views virtual allows for tokens to inherit from both.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio:
Thanks @jaketimothy. Where did you run into this? We would expect you would use `ERC20Permit` along with `ERC20Votes` instead of `Votes`.

I've opened an issue to discuss a fix for this. The solution in this PR is not a good use of virtual functions since they wouldn't be chained by `super`.

https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3905

jaketimothy:
> Thanks @jaketimothy. Where did you run into this? We would expect you would use `ERC20Permit` along with `ERC20Votes` instead of `Votes`.
> 
> I've opened issue #3905 to discuss a fix for this. The solution in this PR is not a good use of virtual functions since they wouldn't be chained by `super`.

Agree

I was prototyping my own `ERC20Votes` and `ERC1155Votes` based off `5.0-next` and running into this.

PR #3876, opened 2022-12-13
Remove ERC1155 hooks

### LINKED ISSUE #3535: Consider alternative design for token transfers and hooks

JulissaDantes:
Fixes #3535  

The current way to customize token transfers(mint, burn, transfer) is by overriding the hooks _beforeTokenTransfer and _afterTokenTransfer. This refactor removes both hooks from the ERC1155 contract, and implements the logic inside the new `_update` internal function, so future customizations only rely on overriding a single entry point.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


TrejGun:
what about erc721? will it be different from 20 and 1155?

JulissaDantes:
> what about erc721? will it be different from 20 and 1155?

Thanks for showing interest. It will be worked on after this to use the same `_update` pattern instead of multiple hooks.

frangio:
@TrejGun Keep in mind that this is planned for 5.0 release which is still months away.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -165,25 +218,9 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
         bytes memory data
     ) internal virtual {
```
frangio, line 219: As in ERC20, the transfer/burn/mint functions that call into `_update` shouldn't be virtual.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -165,25 +218,9 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
         require(to != address(0), "ERC1155: transfer to the zero address");
```
frangio, line 220: As in ERC20, we need to check that `from` isn't 0 either.

frangio (commented):
I've merged `master` into `next-v5.0` which included some significant changes due to #3898. Please merge `next-v5.0` into this PR as there may be conflicts.

contracts/token/ERC1155/ERC1155.sol
JulissaDantes: Since the check wasn't there before the change, I wasn't sure if it  was needed here, but it makes sense that after this change we want to make sure that the transfer is called only for transfers. Therefore, either `from` or `to` can be the zero address.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -145,6 +145,59 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
+            uint256 fromBalance = _balances[id][from];
+            if (to == address(0)) {
+                require(fromBalance >= amount, "ERC1155: burn amount exceeds balance");
+            }
+            if (from != address(0)) {
+                require(fromBalance >= amount, "ERC1155: insufficient balance for transfer");
+                unchecked {
+                    _balances[id][from] = fromBalance - amount;
+                }
+            }
```
Amxx, line 182: There is a check duplication, that is also an actual error:
- we only need to check the from balance if from an account (not if its 0x0)
- if from is 0, then we mint. In that case fromBalance is 0, but we don't care that is less then the amount.
- Since `fromBalance` is only necessary if from is not 0x0, we should only do the load inside this block


```suggestion
            if (from != address(0)) {
                uint256 fromBalance = _balances[id][from];
                require(fromBalance >= amount, "ERC1155: insufficient balance for transfer");
                unchecked {
                    _balances[id][from] = fromBalance - amount;
                }
            }
```

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -145,6 +145,59 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
+
+            if (to != address(0)) {
+                _balances[id][to] += amount;
+            }
+        }
+        if (ids.length == 1) {
+            emit TransferSingle(operator, from, to, ids[0], amounts[0]);
+            if (to != address(0)) {
+                _doSafeTransferAcceptanceCheck(operator, from, to, ids[0], amounts[0], data);
+            }
```
Amxx, line 192:
```suggestion
            uint256 id = ids[0];
            uint256 amount = amounts[0];
            emit TransferSingle(operator, from, to, id, amount);
            if (to != address(0)) {
                _doSafeTransferAcceptanceCheck(operator, from, to, id, amount, data);
            }
```

contracts/mocks/ERC1155PausableMock.sol
```diff
@@ -16,14 +16,13 @@ contract ERC1155PausableMock is ERC1155Mock, ERC1155Pausable {
...
-        address operator,
+    function _update(
         address from,
         address to,
         uint256[] memory ids,
         uint256[] memory amounts,
         bytes memory data
-    ) internal override(ERC1155, ERC1155Pausable) {
-        super._beforeTokenTransfer(operator, from, to, ids, amounts, data);
+    ) internal virtual override(ERC1155, ERC1155Pausable) {
```
Amxx, line 25: mocks function don't need to be virtual, do they? we don't plan on overriding them

```suggestion
    ) internal override(ERC1155, ERC1155Pausable) {
```

contracts/mocks/ERC1155SupplyMock.sol
```diff
@@ -8,14 +8,13 @@ import "../token/ERC1155/extensions/ERC1155Supply.sol";
...
-        address operator,
+    function _update(
         address from,
         address to,
         uint256[] memory ids,
         uint256[] memory amounts,
         bytes memory data
-    ) internal override(ERC1155, ERC1155Supply) {
-        super._beforeTokenTransfer(operator, from, to, ids, amounts, data);
+    ) internal virtual override(ERC1155, ERC1155Supply) {
```
Amxx, line 17: same 
```suggestion
    ) internal override(ERC1155, ERC1155Supply) {
```

package-lock.json
```diff
@@ -26458,4 +26458,4 @@
       "dev": true
     }
   }
-}
+}
```
Amxx, line 26461: not sure where this change comes from, but I think its a mistake

```suggestion
}

```

Amxx (changes requested):
Almost good.

package-lock.json
frangio: Not sure if this is the reason but @JulissaDantes can you make sure you have installed [EditorConfig](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d59306bd06a241083841c2e4a39db08e1f3722cc/.editorconfig#L10

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -143,44 +143,75 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
     }
 
     /**
-     * @dev Transfers `amount` tokens of token type `id` from `from` to `to`.
+     * @dev Transfers `amount` tokens of token type `id` from `from` to `to`. Will mint (or burn) if `from` (or `to`) is the zero address.
      *
-     * Emits a {TransferSingle} event.
+     * Emits a {TransferSingle} event if only one transfer is done, and {TransferBatch} event for multiple transfer operations.
```
frangio, line 148:
```suggestion
     * Emits a {TransferSingle} event if the arrays contain one element, and {TransferBatch} otherwise.
```

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -143,44 +143,75 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
      *
      * Requirements:
      *
-     * - `to` cannot be the zero address.
-     * - `from` must have a balance of tokens of type `id` of at least `amount`.
-     * - If `to` refers to a smart contract, it must implement {IERC1155Receiver-onERC1155Received} and return the
+     * - If `to` refers to a smart contract, it must implement {IERC1155Receiver-onERC1155BatchReceived} and return the
      * acceptance magic value.
```
frangio, line 153:
```suggestion
     * - If `to` refers to a smart contract, it must implement either {IERC1155Receiver-onERC1155Received}
     *   or {IERC1155Receiver-onERC1155BatchReceived} and return the acceptance magic value.
```

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -143,44 +143,75 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
+                    _balances[id][from] = fromBalance - amount;
+                }
+            }
 
-        _afterTokenTransfer(operator, from, to, ids, amounts, data);
+            if (to != address(0)) {
+                _balances[id][to] += amount;
+            }
+        }
+        if (ids.length == 1) {
```
frangio, line 182: Whitespace:

```suggestion

        if (ids.length == 1) {
```

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -143,44 +143,75 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
-
-        uint256 fromBalance = _balances[id][from];
-        require(fromBalance >= amount, "ERC1155: insufficient balance for transfer");
-        unchecked {
-            _balances[id][from] = fromBalance - amount;
-        }
-        _balances[id][to] += amount;
+        for (uint256 i = 0; i < ids.length; ++i) {
+            uint256 id = ids[i];
+            uint256 amount = amounts[i];
```
frangio, line 168: We can define and use a function `_unsafeAccess` here since we don't need bounds checking.

We could add this function in the `Arrays` library. Previously we insisted in keeping this function private, but I think as long as it is called "unsafe" we can make it available. What do you think @Amxx?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/d59306bd06a241083841c2e4a39db08e1f3722cc/contracts/utils/Checkpoints.sol#L199-L207

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -143,44 +143,75 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
+     * - `to` cannot be the zero address.
+     * - `from` must have a balance of tokens of type `id` of at least `amount`.
+     * - If `to` refers to a smart contract, it must implement {IERC1155Receiver-onERC1155Received} and return the
+     * acceptance magic value.
+     */
+    function _safeTransferFrom(address from, address to, uint256 id, uint256 amount, bytes memory data) internal {
+        require(to != address(0), "ERC1155: transfer to the zero address");
+        require(from != address(0), "ERC1155: transfer from the zero address");
+        uint256[] memory ids = _asSingletonArray(id);
+        uint256[] memory amounts = _asSingletonArray(amount);
```
frangio, line 213: We can probably gain some gas efficiency by defining a single function `_asSingletonArrays(id, amount) returns (uint256[] ids, uint256[] amounts)` that allocates the two arrays in assembly at the same time. We should use this function throughout the contract.

contracts/token/ERC1155/extensions/ERC1155Supply.sol
```diff
@@ -31,18 +31,15 @@ abstract contract ERC1155Supply is ERC1155 {
...
         address to,
         uint256[] memory ids,
         uint256[] memory amounts,
         bytes memory data
     ) internal virtual override {
-        super._beforeTokenTransfer(operator, from, to, ids, amounts, data);
-
         if (from == address(0)) {
             for (uint256 i = 0; i < ids.length; ++i) {
                 _totalSupply[ids[i]] += amounts[i];
```
frangio, line 45: Here and below we can use `_unsafeAccess` as well.

frangio (changes requested):
A few suggestions for gas efficiency.

contracts/token/ERC1155/ERC1155.sol
Amxx: `unsafeAccess`'s main point is to avoid the `sload` that gets the length. Here, we are in memory, so we would avoid a `mload` (and a comparison). It would save gas, but not as much as when dealing with storage.

Its really a balance of savings vs readability.

I'm ok with implementing it ... maybe in a following PR were we would go through the codebase and finds all relevant accesses.

contracts/token/ERC1155/extensions/ERC1155Supply.sol
frangio: I meant unsafe access for the arrays. I agree that the arithmetic here should not be unchecked.

contracts/token/ERC1155/ERC1155.sol
frangio: That's true, the savings are a lot less significant in memory. But in relative terms doing bounds checking probably doubles or triples the cost of an array access. We should look into this, but agree that it's not critical.

frangio:
We've agreed to push the optimizations for later. Opened https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3940.

Amxx (approved):
LGTM

One note: In this code batch transfer are valid if the batch is empty. The EIP doesn't appear to discuss that corner case.

Rassska:
@Amxx, @frangio, it's not worth including `_doSafeTransferAcceptanceCheck()` into the functions like: _mint()/_burn()/_mintBatch(). 

The only reason is because it costs huge amount of gas and opens a lot of doors towards re-entrancy! 
Please, keep internal _mint()/burn() as simple as they were.

Thanks!

frangio:
> Please, keep internal `_mint()`/`_burn()` as simple as they were.

This PR did not introduce new behavior. The acceptance check has been present since the first release of our ERC1155 implementation.

We are considering building in a reentrancy guard in mint to address the reentrancy risk, as well as in ERC721.

PR #3878, opened 2022-12-15
internal getter for token uri

### LINKED ISSUE #3023: ERC721URIStorage: change _tokenURIs visibility from private to internal
**🧐 Motivation**
This would allow much easier (and cheaper) implementations of an `hasTokenURI` helper.

**📝 Details**
Currently the `ERC721URIStorage._tokenURIs` variable is defined as `private`, which prevents inheriting contracts from directly accessing the data stored in it.  
Additionally, when the `_baseURI` method provides a string but the token itself has no URI set, the `ERC721URIStorage.tokenURI` method ends up returning the `super`/`ERC721` implementation, where the `_baseURI` is concatenated with `tokenId` resulting in a non-empty string. Checking against this return value to know if the URI is already set is unnecessarily costly.  

The change in visibility to `internal` would allow for a significantly simpler implementation of this check, something like:
```
abstract contract ERC721URIStorage is ERC721 {
    using Strings for uint256;

    mapping(uint256 => string) internal _tokenURIs;
...
```
```
function _hasTokenURI(uint256 tokenId) internal view returns (bool) {
    return bytes(_tokenURIs[tokenId]).length > 0;
}
```

Would this raise security concerns?

quantum13:
Fixes #3023

Added internal getter for retrieving token URI without base URI

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -35,6 +35,18 @@ abstract contract ERC721URIStorage is ERC721 {
...
     }
 
+    /**
+     * @dev Gets the tokenURI of `tokenId`.
+     *
+     * Requirements:
+     *
+     * - `tokenId` must exist.
+     */
+    function _getTokenURI(uint256 tokenId) internal view virtual returns (string memory) {
```
frangio, line 45: This function needs another name that makes it clearer that this is different from `tokenURI`.

contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -35,6 +35,18 @@ abstract contract ERC721URIStorage is ERC721 {
...
+        require(_exists(tokenId), "ERC721URIStorage: URI get of nonexistent token");
```
frangio, line 46: I would remove this for efficiency.

frangio (commented):
Sorry for the late response.

Amxx:
This is old and causes conflicts (uses mocks that no longer exists, writes directly into the changelog, ...)

replaced by #6175

PR #3879, opened 2022-12-15
Optimize "_isApprovedOrOwner"

paulrberg:
#### Description

There is no point in calling `ERC721.ownerOf` in the `_isApprovedOrOwner` function, because the requirement that the `tokenId` must exist is already checked in the `getApproved` function (which calls the `_requireMinted` function).

In this PR, I change the call to the internal `ERC721._ownerOf` function (with an underscore) to avoid checking twice if the owner is the zero address, saving a little gas whenever `_isApprovedOrOwner` is called.

I did not touch any of the tests, but they still pass.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
  * `ReentrancyGuard`: Add a `_reentrancyGuardEntered` function to expose the guard status. ([#3714](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3714))
  * `ERC20Votes`: optimize by using unchecked arithmetic. ([#3748](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3748))
+ * `ERC721`: optimize `_isApprovedOrOwner` by not checking that `tokenId` exists twice. ([#3879](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3879))
  * `Initializable`: optimize `_disableInitializers` by using `!=` instead of `<`. ([#3787](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3787))
  * `Math`: optimize `log256` rounding check. ([#3745](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3745))
 
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -237,7 +237,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
      * - `tokenId` must exist.
      */
     function _isApprovedOrOwner(address spender, uint256 tokenId) internal view virtual returns (bool) {
-        address owner = ERC721._ownerOf(tokenId);
+        address owner = _ownerOf(tokenId);
         return (spender == owner || isApprovedForAll(owner, spender) || getApproved(tokenId) == spender);
     }
 
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -237,7 +237,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
      * - `tokenId` must exist.
      */
     function _isApprovedOrOwner(address spender, uint256 tokenId) internal view virtual returns (bool) {
-        address owner = ERC721.ownerOf(tokenId);
+        address owner = ERC721._ownerOf(tokenId);
         return (spender == owner || isApprovedForAll(owner, spender) || getApproved(tokenId) == spender);
     }
 


JulissaDantes:
Please check the failing checks. This test is failing:
```
Contract: ERC721Consecutive
       with valid batches
         ERC721 behavior
           tokens can be burned and re-minted #1:
     Error: VM Exception while processing transaction: reverted with reason string 'ERC721: caller is not token owner or approved'
```

paulrberg:
Thanks for the heads-up, @JulissaDantes.

The issue was that I left the `ERC721_` prefix in front of the call to the internal function `_ownerOf`. I find it pretty odd that Solidity didn't throw any warning at all, and it just returned `address(0)` instead.

Anyhow, the tests should pass now.

paulrberg:
Hi @frangio, sorry to tag you, but do you think you could take a quick look at this?

We are working hard to get our contract size below 24kB - merging this PR would enable us to delete a function in our code base.

Amxx:
> The issue was that I left the `ERC721_` prefix in front of the call to the internal function `_ownerOf`. I find it pretty odd that Solidity didn't throw any warning at all, and it just returned `address(0)` instead.


That is definitely not an error.

It was decided at some point (before we had the `_ownerOf` internal function) that internal calls to `ownerOf` should target `ERC721.ownerOf`. The consequence of that is that if someone overrides `ownerOf` it would not alter the behavior of the function that call it internally. 

This was debated a lot. Its always a balance between letting devs override function to alter the contract behavior (which comes with risk) and "freezing" some functionalities to prevent risks. 

This was before we had  `_ownerOf` which was added with the ERC721Consecutive module.

Considering we today have `_ownerOf`, its clear to me that this behavior is designed to be overridden, and that using `ERC721.ownerOf` internally doesn't feel right. In my personal opinion, all calls to `ERC721.ownerOf` should be replaced with calls to `_ownerOf`, and the requirements that that the owner is not address(0) should be handled locally.

Amxx:
> We are working hard to get our contract size below 24kB - merging this PR would enable us to delete a function in our code base.

That is probably not the cleanest way to achieve that, but have you tried 
```
debug: { revertStrings: "strip" },
```
in your compiler settings?

paulrberg:
> That is definitely not an error.

It might have been - I was not referring to `ERC721.ownerOf`(notice the *lack* of underscore), but rather to `ERC721._ownerOf` (notice the *existence* of underscore), which I added myself. Though I am not sure if this is a bug or just a weird behavior of Solidity.

> It was decided at some point ... that internal calls to `ownerOf` should target `ERC721.ownerOf`

I talked to @frangio about this in this [blog post](https://forum.openzeppelin.com/t/why-do-you-call-certain-public-functions-by-using-the-contract-name/33723?u=paulrberg) on the OZ forum.

> In my personal opinion, all calls to `ERC721.ownerOf` should be replaced with calls to `_ownerOf`.

I agree.

> `debug: { revertStrings: "strip" }`

Thanks for the tip - the issue is not that we can't run our tests. Foundry is able to deploy contracts even if they have more than 24kB. The problem is when deploying to a testnet.

frangio:
> merging this PR would enable us to delete a function in our code base

Would you mind elaborating on how this change helps you with that? On the surface it seems unrelated to code size.

frangio (dismissed):
This change on its own is fine. I would want us to look at the other occurences of `ERC721.ownerOf` but we need to do that more carefully and it is covered by #3883.

paulrberg:
Thanks for accepting the PR, @frangio.

> Would you mind elaborating on how this change helps you with that?

We simply had another function in our code base that was doing what the updated `_isApprovedOrOwner` in this PR does. Now we can remove that other function and just use the one from your implementation.

Amxx:
The proposed changes cause "_isApprovedOrOwner" to return true for none existing tokens with `spender == address(0)`.
Is that ok?

frangio:
Yes, that is concerning, I would want to more carefully consider this change then.

> We simply had another function in our code base that was doing what the updated `_isApprovedOrOwner` in this PR does. Now we can remove that other function and just use the one from your implementation.

In that case, what you could do is override `_isApprovedOrOwner` with the same function definition you're proposing in this PR, and without calling `super._isApprovedOrOwner`. Solidity will include only your code in the final bytecode, and not the original function definition.

paulrberg:
> In that case, what you could do is override _isApprovedOrOwner with the same function definition you're proposing in this PR, and without calling super._isApprovedOrOwner

Oh, great idea. Thanks!

> The proposed changes cause "_isApprovedOrOwner" to return true for none existing tokens with spender == `address(0)`.

Oh, good catch, I didn't consider this edge case when I opened the PR.

> what would happen if the spender was `ecrecovered` from an invalid signature

Just so what I understand the concern better - by this, you mean a situation whereby a user inherits from your implementation, don't you?

Asking because if in our case we are sure that we never pass a zero `spender`, and just want to make sure there's no other place in OZ implementation that could lead to this.

frangio:
> Just so what I understand the concern better - by this, you mean a situation whereby a user inherits from your implementation, don't you?

Yeah, that's what we're thinking about.

paulrberg:
Thanks for confirming.

In this case, given the concern raised by @Amxx, I am happy to let go and close this PR. I will re-write my internal function to override `_isApprovedOrOwner` so that I reduce the bytecode of my contract.

frangio:
Ok! Closing this in favor of #3883 where we will review the use of `ERC721.ownerOf` anyway.

PR #3880, opened 2022-12-15
Remove ERC165Storage

### LINKED ISSUE #3641: Consider removing ERC165Storage
Does it ever make sense to use this contract over ERC165 with overrides and if statements?

The API is a little simpler because it's just a call to `_registerInterface`. The use of storage makes it more expensive to execute `supportsInterface`, and also to deploy the contract. Deployment costs are not necessarily worse though, because when storage is not used it requires more bytecode, but probably not enough to match it.

JulissaDantes:
Fixes #3641  

 The use of storage makes it more expensive to execute supportsInterface, and also to deploy the contract. Deployment costs are not necessarily worse though, because when storage is not used it requires more bytecode, but probably not enough to match it. After verifying that the percentage of people using `ERC165storage` as opposed to the `ERC165` version is considerably lower, we are removing the `ERC165Storage` contract.

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [x] Changelog entry


CHANGELOG.md
```diff
@@ -33,6 +34,15 @@ For example, a contract using `ERC20`'s `_beforeTokenTransfer` hook would have t
   }
 ```
 
+#### ERC165Storage
```
frangio, line 37:
```suggestion
#### ERC165Storage

```

contracts/utils/introspection/ERC165.sol
```diff
@@ -17,7 +17,6 @@ import "./IERC165.sol";
  * }
  * ```
  *
```
frangio, line 19:
```suggestion
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
  * `TransparentUpgradeableProxy`: Removed `admin` and `implementation` getters, which were only callable by the proxy owner and thus not very useful. ([#3820](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3820))
  * `ProxyAdmin`: Removed `getProxyAdmin` and `getProxyImplementation` getters. ([#3820](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3820))
  * `ERC20`: Deleted `_beforeTokenTransfer` and `_afterTokenTransfer` hooks, added a new internal `_update` function for customizations, and refactored all extensions using those hooks to use `_update` instead. ([#3838](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3838))
+ * `ERC165Storage`: contract was removed. ([#](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/))
```
frangio, line 12:
```suggestion
 * `ERC165Storage`: Removed this contract in favor of inheritance based approach. ([#3880](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3880))
```

CHANGELOG.md
```diff
@@ -33,6 +34,15 @@ For example, a contract using `ERC20`'s `_beforeTokenTransfer` hook would have t
   }
 ```
 
+#### ERC165Storage
+As a result of removing the ERC165Storage contracts, users wanting to indicate their support for interfaces should do so by overriding the `supportsInterface` function as seen below:
+
+```
```
frangio, line 40:
```suggestion
```solidity
```

CHANGELOG.md
```diff
@@ -33,6 +34,15 @@ For example, a contract using `ERC20`'s `_beforeTokenTransfer` hook would have t
...
+ function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+    return interfaceId == type(MyInterface).interfaceId || super.supportsInterface(interfaceId);
+ }
```
frangio, line 43:
```suggestion
function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
    return interfaceId == type(MyInterface).interfaceId || super.supportsInterface(interfaceId);
}
```

CHANGELOG.md
```diff
@@ -33,6 +34,15 @@ For example, a contract using `ERC20`'s `_beforeTokenTransfer` hook would have t
   }
 ```
 
+#### ERC165Storage
+As a result of removing the ERC165Storage contracts, users wanting to indicate their support for interfaces should do so by overriding the `supportsInterface` function as seen below:
```
frangio, line 38: This isn't bad but here is what I would write:

```suggestion
Users that were registering EIP-165 interfaces with `_registerInterface` from `ERC165Storage` should instead do so so by overriding the `supportsInterface` function as seen below:
```

frangio (approved):
LGTM!

PR #3887, opened 2022-12-18
🥢Pausable: Don't set pause false via constructor.

z0r0z:
It seems wasteful to set `_paused` to `false` in constructor as this is default value anyways.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/contracts/security/Pausable.sol
+++ b/contracts/security/Pausable.sol
@@ -30,9 +30,7 @@ abstract contract Pausable is Context {
     /**
      * @dev Initializes the contract in unpaused state.
      */
-    constructor() {
-        _paused = false;
-    }
+    constructor() {}
 
     /**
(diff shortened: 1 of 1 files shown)


Rassska (commented):
LGTM!

ernestognw:
Pausable.sol has been updated but doesn't include this change.

Personally I think it needs discussion because with the EIP-7201 addition to upgradeable contracts, a Pausable module can be removed and added back again between upgrades, therefore removing this would transpile into a `PausableUpgradeable` that behaves differently on initialization and could be considered a breaking change.

Unfortunately we didn't review back then and this doesn't have activity anymore. Closing, but feel free to open an issue for discussion first

Issue #3890, opened 2022-12-20
Transfer a specific access


njrapidinnovation:
**🧐 Motivation**
I want to transfer my `OPERATOR` access to another wallet using access control but the thing i can do is only `RENOUNCE` my role. I want a method where my `OPERATOR` role gets `REVOKED` from my current address and `GRANTED` to a new address.

**📝 Details**
* A method to inform the `ADMIN OF MY CURRENT ROLE` that i want to transfer my role instead of `REVOKING` it.
* A method which only `ADMIN OF MY CURRENT ROLE` can call  to `REVOKE` my access from my current address and `GRANT` it to new given address.

Issue #3623, opened 2022-08-16
AccessControl Admin Rules


frangio:
The default admin in AccessControl is a very sensitive role and it requires special treatment. In the docs we recommend "extra precautions" but this refers mostly to off-chain operational security:

![image](https://user-images.githubusercontent.com/481465/184980752-491c725f-1a30-42a7-9522-633db7f32427.png)

Is there something we can do at the contract level to encourage better security practices around the default admin role?

I am thinking we can have an extension of AccessControl that adds special rules for this role.

Some ideas I've gathered from @nchamo and @nventuro:

1) Enforce that just one account holds the role.
2) Enforce a 2-step process to transfer the role to another account: grant then claim.
3) Enforce a configurable delay between the two steps, with the ability to cancel in between.

Related to https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3593

hrik2001:
Hello @frangio, you might remember me from #3598.
Interested in implementing this, but I believe that you might have more things in mind.

frangio:
We're still discussing ideas. If you want to help I think it would be great to put together a proof of concept implementing the 3 points I mentioned. You can share it here in a comment. Please no PRs yet until we pin down the design!

Amxx:
POC:

```

// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

import "@openzeppelin/contracts/access/AccessControl.sol";
import "@openzeppelin/contracts/access/Ownable.sol";

contract AccessControlWithDelayedOwnership is Ownable, AccessControl {
    // Making this updatable causes issues:
    // - who can update it? Only the owner!
    // - if the owner what to instant transfer, can he just set it to 0 and do an instant transfer?
    // - should we enforce a minimum, non updatable delay?
    // - if so, why do we need an updatable delay and not just stick to the minimum value.
    uint32  public immutable delay;
    
    address public pendingOwner;
    uint32  public pendingOwnerTimer;
    
    constructor(uint32 initialDelay) {
        delay = initialDelay;
    }

    // Admin is owner
    function hasRole(bytes32 role, address account) public view override returns (bool) {
        return role == DEFAULT_ADMIN_ROLE ? account == owner() : super.hasRole(role, account);
    }

    function _grantRole(bytes32 role, address account) internal override {
        require(role != DEFAULT_ADMIN_ROLE, "Admin role overriden with ownership");
        super._grantRole(role, account);
    }

[...]

PR #3891, opened 2022-12-20
AccessControl2Step to transfer a specific access

njrapidinnovation:
Refer issue #3890 

I made a state variable to store the pending access related to a user and a role , using this a user can initiate a role transfer transaction to a new user and admin of that particular role can only grant this so that role will be revoked from the initial user and will be granted to the new one.


frangio:
Please discuss in issues before submitting a PR.

The approach we will follow for this is probably the one laid out in #3623.

PR #3893, opened 2022-12-21
Remove ERC721 tokenTransfer hooks 

### LINKED ISSUE #3535: Consider alternative design for token transfers and hooks

JulissaDantes:
Fixes #3535 
Replaces #3636 

The current way to customize token transfers(mint, burn, transfer) is by overriding the hooks _beforeTokenTransfer and _afterTokenTransfer. This refactor removes both hooks from the ERC721 contract, and implements the logic inside the new _update internal function, so future customizations only rely on overriding a single entry point.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


contracts/token/ERC721/ERC721.sol
```diff
@@ -330,32 +296,53 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
...
-        // Check that tokenId was not transferred by `_beforeTokenTransfer` hook
-        require(ERC721.ownerOf(tokenId) == from, "ERC721: transfer from incorrect owner");
+    /**
+     * @dev Transfers `tokenId` from `from` to `to`, or alternatively mints (or burns) if `from` (or `to`) is
+     * the zero address. All customizations to transfers, mints, and burns should be done by overriding this function.
+     *
+     * Emits a {Transfer} event.
+     */
+    function _update(address from, address to, uint256 tokenId, uint256 batchSize) internal virtual {
+        if (batchSize > 1) {
```
Amxx, line 313: I'm not a fan of this switch.

In all cases
```
if (from != address(0)) {
    _balances[from] -= batchSize;
}
if (to != address(0)) {
    _balances[to] += batchSize;
}
```
(the unchecked part might be challenging)

**But more importantly, there is no check that `batchSize > 0`!**

If batchSize is 0, then we go in the else section, which will burn the token (oups)

contracts/token/ERC721/ERC721.sol
```diff
@@ -330,32 +296,53 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
...
+                }
+            }
 
-        emit Transfer(from, to, tokenId);
+            _owners[tokenId] = to;
 
-        _afterTokenTransfer(from, to, tokenId, 1);
+            emit Transfer(from, to, tokenId);
+        }
     }
```
Amxx, line 346: I'd do

```
function _update(address from, address to, uint256 tokenId, uint256 batchSize) internal virtual {
    if (batchSize == 1) {
        if (from != address(0)) {
            // Clear approvals from the previous owner
            delete _tokenApprovals[tokenId];

            unchecked {
                // Cannot overflow, as that would require more tokens to be burned/transferred
                // out than the owner initially received through minting and transferring in.
                _balances[from] -= 1;
            }
        }

        if (to != address(0)) {
            unchecked {
                // Will not overflow unless all 2**256 token ids are minted to the same owner.
                // Given that tokens are minted one by one, it is impossible in practice that
                // this ever happens. Might change if we allow batch minting.
                // The ERC fails to describe this case.
                _balances[to] += 1;
            }
        }

        _owners[tokenId] = to;

    } else if (batchSize > 1) {
        if (from != address(0)) {
            _balances[from] -= batchSize;
        }
        if (to != address(0)) {
            _balances[to] += batchSize;
        }
    }
    // else case (batchSize == 0) is a no-op, so we do nothing.
}
```

Amxx:
Overall I feel uncomfortable with this change. 

Devs that call  `_update(from, to, tokenId, batchSize)` will expect the batch to be transferred, including ownership transfers and events. 

The truth is, only batchsize == 1 does the transfer, and any other value is just a hack to update the balances. I think this is very error prone and is not aligned with our guidelines!

frangio:
@Amxx What alternative do you recommend? We can revisit https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3744, but I would like to consider other options.

Amxx:
My first idea would be to remove the batchSize from the `_update` function, and manage the balances change of batch minting in a different way.

frangio:
But how? And we should keep in mind that the goal is to minimize overrides for the end user.

frangio:
We might have to consider adding `_mintConsecutive` by default in EIP721. It stretches the limits of what's comfortably possible with inheritance.

Amxx:
> We might have to consider adding `_mintConsecutive` by default in ERC721. It stretches the limits of what's comfortably possible with inheritance.

That is one option

There was another option that used overflows, but it used 2 slots per account ... and feels very "hacky"

My concern is that putting `_mintConsecutive` by default in ERC721 will add gas cost to usecases that don't do any concecutive minting.

contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -48,15 +48,16 @@ abstract contract ERC721URIStorage is ERC721 {
...
      * the storage mapping.
      */
-    function _burn(uint256 tokenId) internal virtual override {
-        super._burn(tokenId);
-
-        if (bytes(_tokenURIs[tokenId]).length != 0) {
-            delete _tokenURIs[tokenId];
+    function _update(address from, address to, uint256 firstTokenId, uint256 batchSize) internal virtual override {
+        super._update(from, to, firstTokenId, batchSize);
+        if (to == address(0)) {
```
Amxx, line 57: if `to == address(0)` and `batchSize > 1`, we should probably do something. Revert?

contracts/token/ERC721/extensions/ERC721Royalty.sol
```diff
@@ -29,10 +29,12 @@ abstract contract ERC721Royalty is ERC2981, ERC721 {
...
     /**
-     * @dev See {ERC721-_burn}. This override additionally clears the royalty information for the token.
+     * @dev See {ERC721-_update}. This override additionally clears the royalty information for the token after burning.
      */
-    function _burn(uint256 tokenId) internal virtual override {
-        super._burn(tokenId);
-        _resetTokenRoyalty(tokenId);
+    function _update(address from, address to, uint256 firstTokenId, uint256 batchSize) internal virtual override {
+        super._update(from, to, firstTokenId, batchSize);
+        if (to == address(0)) {
```
Amxx, line 36: if `to == address(0)` and `batchSize > 1`, we should probably do something. Revert?

Amxx (commented):
Maybe not for this PR, but we will have to worry about these extensions.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -90,41 +82,27 @@ abstract contract ERC721Consecutive is IERC2309, ERC721 {
...
-        uint256 firstTokenId,
-        uint256 batchSize
-    ) internal virtual override {
+    function _update(address from, address to, uint256 firstTokenId, uint256 batchSize) internal virtual override {
+        if (from == address(0) && batchSize == 1) {
+            // Using {ERC721Consecutive} prevents using {_mint} during construction in favor of {_mintConsecutive}.
+            // After construction, {_mintConsecutive} is no longer available and {_mint} becomes available.
+            require(Address.isContract(address(this)), "ERC721Consecutive: can't mint during construction");
+        }
+        super._update(from, to, firstTokenId, batchSize);
```
Amxx, line 105: Also realized that here we are not following the rule that overrides should do their own stuff before calling super (and not after). We need to add comments documenting why we are doing so.

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -55,16 +55,9 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
...
-    function _beforeTokenTransfer(
-        address from,
-        address to,
-        uint256 firstTokenId,
-        uint256 batchSize
-    ) internal virtual override {
-        super._beforeTokenTransfer(from, to, firstTokenId, batchSize);
-
+    function _update(address from, address to, uint256 firstTokenId, uint256 batchSize) internal virtual override {
         if (batchSize > 1) {
```
Amxx, line 61: We have to consider `batchSize == 0`

contracts/token/ERC721/ERC721.sol
frangio: I would say let's go with the first option.

Amxx:
Here is a summary of the identified issues that would need to be fixed:

- `_update` does not check the `from`. This means that a dev could call the `_update` function with an invalid from, and cause:
	-  an invalid event to be emitted
	- corruption of the `balances` mapping.
- Cleanly support batchSize = 0 in extensions. 
	- Some of them are not designed to handle batches, and just don't know not to process burns of size > 1. The fact that they "hook" into `_update(address,address,uint256,uint256)` means they could see signals they don't know how to handle. Hopefully they never have to, but this argues against having a single `_update` for both single and batched operations.
	- Affected extensions:
		- ERC721Storage
		- ERC721Royalty
		- ERC721Enumerable
- Document situations where we don't follow our own guidelines.
	- In ERC721Consecutive, we do some operations before, and other after the  `super._uodate(...)` call. They is in conflict with our guidelines

frangio:
Closing this PR in favor of the issue, where I shared the above summary: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3535

PR #3900, opened 2022-12-27
Rename approved parameter in IERC721

paulrberg:
The name of the second argument of the `setApprovedForAll` is different between the interface (`IERC721`) and the implementation contract (`ERC721`).


frangio (approved):
Thanks!

PR #3903, opened 2022-12-29
Simplify ESLint config

frangio:
This PR removes a large part of our ESLint config, including all style rules (this is done by `eslint-config-prettier`). In a follow-up PR, I would like to set up Prettier for all JS files.

Part of the motivation here is to have fewer npm dependencies.


ernestognw (approved):
LGTM, thanks!

PR #3904, opened 2022-12-29
Disable npm audit when installing deps in CI

frangio:
This makes it faster to install dependencies.


frangio:
I don't think this makes a difference actually. `npm ci` seems to automatically detect it's in a CI environment and not run audit.

PR #3913, opened 2023-01-02
Use Prettier for JS files

frangio:
Follow up to #3903. Adds Prettier and reformats all JavaScript according to it.

The main change with respect to our previous formatting rules is removal of the space after the function name in a definition: `function foo (` -> `function foo(`. I'm personally more used to this style and prefer it.

The other thing that Prettier implies is that we lose finer control over where we insert newlines to break long lines. But the line length (100 characters) should still lead to readable code.

Should merge after #2712.


Amxx:
Any reason we use `printWidth: 100`? I'm used to 80 and 120, but not 100.

frangio:
120 seemed like too much when I tried it, you can look in earlier commits in this PR to compare.

Note that Prettier's "print width" [is not a hard upper limit](https://prettier.io/docs/en/options.html#print-width).

Open to changing it.

Amxx:
I've tried changing it to 120 and running the linter. IMO it looks better, but that might be my IDE config. 

I don't think its a big deal as long as we have more consistency.

frangio:
Yeah it's hard to judge globally and in some cases we will like the output in others we won't so much. But 120 seems to look good in general, and it's also what we were already using for Solidity.

frangio:
I reran Prettier on top of `master` and it was able to collapse a few more lines. I think because I went to 100 characters and then 120 it had left a few lines shorter than they could be.

Amxx (approved):

PR #3917, opened 2023-01-03
Rename param to match the interface

bxmmm1:
Function parameters & interface names don't match.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry


frangio (approved):
Thanks! We should have a linter rule for this...

PR #3920, opened 2023-01-03
Add keys() accessor to EnumerableMaps

### LINKED ISSUE #3882: Add method `keys` for `EnumerableMap`
**🧐 Motivation**

Add convenient way for retrieving all keys of `EnumerableMap`. I understand that this method will have limited usage (as `EnumerableSet.values()`)

**📝 Details**
There are possibilities to get all the keys of EnumerableMap, but they are quite inconvenient. Examles for `EnumerableMap.UintToUintMap`:
```solidity
        bytes32[] memory temp = map._inner._keys._inner._values;
        uint256[] memory result;

        assembly {
            result := temp
        }
```
or
```solidity
using EnumerableSet for EnumerableSet.Bytes32Set;
...
        bytes32[] memory temp = map._inner._keys.values();
        uint256[] memory result;

        assembly {
            result := temp
        }
```
or even loop by all keys.

I propose to add method keys with imlementation as in the example 2 above

Amxx:
Fixes #3882

#### PR Checklist

- [x] Tests
- [x] Changelog entry


frangio (approved):
I changed the docs in several places including the new function said "set" instead of "map". Please review that.

PR #3921, opened 2023-01-03
Fix ERC20._update

JulissaDantes:
There is a risk with the current implementation of the `_update` function of altering the tokens total supply when sending the zero address as the `from` and `to` parameters. 

If someone does `_update(address(0), address(0), amount)`, this will go into
`if (from == address(0)) { ... }`
that will increase the total supply and mint tokens to address(0), then because of the `else` it will not go into
`else if (to == address(0)) { ... }`
so the total supply amount would remain unbalanced.

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changelog entry

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -252,7 +252,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
 
         if (to == address(0)) {
             unchecked {
-                // Overflow not possible: amount <= totalSupply.
+                // Overflow not possible: amount <= totalSupply or amount <= fromBalance <= totalSupply.
                 _totalSupply -= amount;
             }
         } else {
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -234,38 +234,33 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
      *
      * Emits a {Transfer} event.
      */
-    function _update(
-        address from,
-        address to,
-        uint256 amount
-    ) internal virtual {
-        if (from == address(0)) {
-            _totalSupply += amount;
-            unchecked {
-                // Overflow not possible: balance + amount is at most totalSupply + amount, which is checked above.
-                _balances[to] += amount;
-            }
-        } else if (to == address(0)) {
-            uint256 fromBalance = _balances[from];
-            require(fromBalance >= amount, "ERC20: burn amount exceeds balance");
-            _totalSupply -= amount;
-            unchecked {
-                // Overflow not possible: amount <= fromBalance <= totalSupply.
-                _balances[from] = fromBalance - amount;
-            }
-        } else {
-            uint256 fromBalance = _balances[from];
-            require(fromBalance >= amount, "ERC20: transfer amount exceeds balance");
-            unchecked {
-                _balances[from] = fromBalance - amount;
-                // Overflow not possible: the sum of all balances is capped by totalSupply, and the sum is preserved by
-                // decrementing then incrementing.
-                _balances[to] += amount;
-            }
+function _update(
+    address from,
+    address to,
+    uint256 amount
+) internal virtual {
+    if (from == address(0)) {
+        _totalSupply += amount;
+    } else {
+        uint256 fromBalance = _balances[from];
+        require(fromBalance >= amount, "ERC20: transfer amount exceeds balance");
+        unchecked {
+            // Overflow not possible: amount <= fromBalance <= totalSupply.
+            _balances[from] = fromBalance - amount;
         }
-
-        emit Transfer(from, to, amount);
     }
+
+    if (to == address(0)) {
+        _totalSupply -= amount;
+    } else {
+        unchecked {
+            // Overflow not possible: balance + amount is at most totalSupply + amount, which is checked above.
+            _balances[to] += amount;
+        }
+    }
+
+    emit Transfer(from, to, amount);
+}
 
     /**
      * @dev Creates `amount` tokens and assigns them to `account`, by transferring it from address(0).
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -251,7 +251,9 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
         }
 
         if (to == address(0)) {
-            _totalSupply -= amount;
+            unchecked {
+                _totalSupply -= amount;
+            }
         } else {
             unchecked {
                 // Overflow not possible: balance + amount is at most totalSupply, which we know fits into a uint256.
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -254,7 +254,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
             _totalSupply -= amount;
         } else {
             unchecked {
-                // Overflow not possible: balance + amount is at most totalSupply + amount, which is checked above.
+                // Overflow not possible: balance + amount is at most totalSupply, which we know fits into a uint256.
                 _balances[to] += amount;
             }
         }
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -252,6 +252,7 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
 
         if (to == address(0)) {
             unchecked {
+                // Overflow not possible: amount <= totalSupply.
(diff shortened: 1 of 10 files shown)


contracts/token/ERC20/ERC20.sol
```diff
@@ -241,25 +241,20 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
...
+        if (to == address(0)) {
+            _totalSupply -= amount;
         } else {
-            uint256 fromBalance = _balances[from];
-            require(fromBalance >= amount, "ERC20: transfer amount exceeds balance");
             unchecked {
-                _balances[from] = fromBalance - amount;
-                // Overflow not possible: the sum of all balances is capped by totalSupply, and the sum is preserved by
-                // decrementing then incrementing.
+                // Overflow not possible: balance + amount is at most totalSupply + amount, which is checked above.
```
Amxx, line 257: That comment is wrong, totalSupply+amount is not checked.

A better comment would be balance+amount is at most totalSupply, which we know fits into a uint256

contracts/token/ERC20/ERC20.sol
```diff
@@ -241,25 +241,20 @@ contract ERC20 is Context, IERC20, IERC20Metadata {
...
-            _totalSupply -= amount;
+            require(fromBalance >= amount, "ERC20: transfer amount exceeds balance");
             unchecked {
                 // Overflow not possible: amount <= fromBalance <= totalSupply.
                 _balances[from] = fromBalance - amount;
             }
+        }
+
+        if (to == address(0)) {
+            _totalSupply -= amount;
```
Amxx, line 254: this should be unchecked

Amxx:
ERC20Snapshot should also be updated. Should that be done in a separate PR?

contracts/token/ERC20/ERC20.sol
JulissaDantes: why?

JulissaDantes:
> ERC20Snapshot should also be updated. Should that be done in a separate PR?

It will be done here.

contracts/token/ERC20/ERC20.sol
Amxx: this cannot overflow:
- either the update is from 0 to 0, in which case we are removing what we just added → this is safe
- or we are updating from X to 0 (burn), in which case the amount burned is constrained by the from balance, which we know cannot be bigger than the totalSupply → this is also safe

test/token/ERC20/ERC20.behavior.js
```diff
@@ -178,6 +178,17 @@ function shouldBehaveLikeERC20 (errorPrefix, initialSupply, initialHolder, recip
         );
       });
     });
+
+    describe('when the token owner and the recipient are the zero address', function () {
```
Amxx, line 182: This is checking that the `transferFrom` function is not correctly preventing the 0 address.

We need a test of the internal function (`$_update`) to check that doing from 0 to 0 internally does not revert, emits the right event, and doesn't change the totalsupply

Amxx:
We should merge master into next-5.0. 

This will enable hardhat-exposed, removing the needs for mocks.

contracts/token/ERC20/ERC20.sol
frangio: Yes, in fact this explanation is included above where it says:

```
// Overflow not possible: amount <= fromBalance <= totalSupply.
```

test/token/ERC20/ERC20.test.js
```diff
@@ -287,6 +287,46 @@ contract('ERC20', function (accounts) {
     });
   });
 
+  describe('_update', function () {
+    const amount = initialSupply + 1;
```
frangio, line 291: This seems simpler. I don't see a reason to involve `initialSupply`.

```suggestion
    const amount = 1;
    
```

test/token/ERC20/ERC20.test.js
```diff
@@ -287,6 +287,46 @@ contract('ERC20', function (accounts) {
...
+    it('to is the zero address', async function () {
+      const totalSupply = await this.token.totalSupply();
+      const burnAmount = await this.token.balanceOf(initialHolder);
+      await this.token.update(initialHolder, ZERO_ADDRESS, burnAmount);
+      
+      expect(await this.token.totalSupply()).to.be.bignumber.equal(totalSupply.sub(burnAmount));
+      expect(await this.token.balanceOf(initialHolder)).to.be.bignumber.equal(new BN(0));
+    }); 
+
+    describe('from and to are the zero address', function () {
```
frangio, line 310: It's odd that this is tested differently to the other two cases.

That said, I'm not sure that from and to zero addresses should be a valid function call. It sounds to me like it should revert.

contracts/token/ERC20/ERC20.sol
frangio: Yes, in fact this explanation is included above where it says:

```
// Overflow not possible: amount <= fromBalance <= totalSupply.
```

test/token/ERC20/ERC20.test.js
Amxx: > That said, I'm not sure that from and to zero addresses should be a valid function call. It sounds to me like it should revert.

IMO this is a no-op, and we always said that no-op should no revert. Also the whole point of this PR is to fix the behavior when this is called.

test/token/ERC20/ERC20.test.js
```diff
@@ -287,6 +287,46 @@ contract('ERC20', function (accounts) {
     });
   });
 
+  describe('_update', function () {
+    const amount = 1;
+    
+    it('from is the zero address', async function () {
```
frangio, line 293: Let's add an assertion for the transfer event in this test.

test/token/ERC20/ERC20.test.js
```diff
@@ -287,6 +287,46 @@ contract('ERC20', function (accounts) {
...
+      const totalSupply = await this.token.totalSupply();
+      const newSupply = totalSupply.add(new BN(amount));
+      await this.token.update(ZERO_ADDRESS, recipient, amount);
+
+      expect(await this.token.totalSupply()).to.be.bignumber.equal(newSupply);
+      expect(await this.token.balanceOf(recipient)).to.be.bignumber.equal(amount);
+    });
+
+    it('to is the zero address', async function () {
```
frangio, line 302: Let's add an assertion for the transfer event in this test.

test/token/ERC20/ERC20.test.js
```diff
@@ -287,6 +287,46 @@ contract('ERC20', function (accounts) {
...
+      const totalSupply = await this.token.totalSupply();
+      const burnAmount = await this.token.balanceOf(initialHolder);
+      await this.token.update(initialHolder, ZERO_ADDRESS, burnAmount);
+
+      expect(await this.token.totalSupply()).to.be.bignumber.equal(totalSupply.sub(burnAmount));
+      expect(await this.token.balanceOf(initialHolder)).to.be.bignumber.equal(new BN(0));
+    });
+
+    describe('from and to are the zero address', function () {
```
frangio, line 311: Please convert this `describe` block into an `it` that looks like the other two cases. I can't see a reason why there should be a separate test for the event.

frangio (changes requested):

test/token/ERC20/ERC20.test.js
JulissaDantes: Fixed in last commit.

test/token/ERC20/ERC20.test.js
JulissaDantes: Fixed in last commit.

test/token/ERC20/ERC20.test.js
JulissaDantes: Fixed in last commit.

contracts/token/ERC20/ERC20.sol
frangio: Still need to add unchecked to this line.

frangio:
Really confused by why the linter is still failing. It doesn't show any errors locally.

Amxx:
> Really confused by why the linter is still failing. It doesn't show any errors locally.

same here

frangio:
I was somehow able to fix linting by running the latest version of Prettier with the new configuration.

Amxx:
> I was somehow able to fix linting by running the latest version of Prettier with the new configuration.

is it possible that github action as a cache with the old version ?

frangio:
I tried deleting the caches and it didn't make a difference.

Amxx:
Note that linting of the mocks is irrelevant as they will be deleted when we merge master.

frangio (approved):

PR #3930, opened 2023-01-05
Prepare tests for hardhat-exposed transition

Amxx:
#3816 included a change to [ERC721VotesMock.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/e2d2ebc8fcb8d5582b262dd94f432a40df43b114#diff-4636cd79383e8fd8ecbcb6e103d5ac9903f6cdd3132f09eb3f97950721bbc0d2), adding a dummy argument to match ERC20Votes burn function.

This causes issues when removing the mocks in favor of the `hardhat-exposed` plugin.

This PR removes this dummy parameter, and gives the `Votes.behavior.js` the ability to detect the burn function signature, and to adapt the arguments accordingly

#### PR Checklist

- [x] Tests


test/governance/utils/Votes.test.js
```diff
@@ -21,41 +27,56 @@ contract('Votes', function (accounts) {
...
-      this.tx1 = await this.votes.mint(account1, 1);
-      this.tx2 = await this.votes.mint(account2, 1);
-      this.tx3 = await this.votes.mint(account3, 1);
+      this.txs = await Promise.all(Object.entries(amounts).map(kv => this.votes.mint(...kv)));
     });
 
     it('reverts if block number >= current block', async function () {
       await expectRevert(
-        this.votes.getPastTotalSupply(this.tx3.receipt.blockNumber + 1),
+        this.votes.getPastTotalSupply(this.txs.at(-1).receipt.blockNumber + 1),
```
frangio, line 35: Tests are failing because `.at` is not available until Node 16. Ideally our tests should run on Node 14 which is not deprecated until May this year. This is why the CI is configured with Node 14.

We can fix this by installing the dependency `array.prototype.at` and adding `require('array.prototype.at/auto')` at the top of this file.

test/governance/utils/Votes.test.js
```diff
@@ -21,41 +27,56 @@ contract('Votes', function (accounts) {
 
   describe('performs voting operations', function () {
     beforeEach(async function () {
-      this.tx1 = await this.votes.mint(account1, 1);
-      this.tx2 = await this.votes.mint(account2, 1);
-      this.tx3 = await this.votes.mint(account3, 1);
+      this.txs = await Promise.all(Object.entries(amounts).map(kv => this.votes.mint(...kv)));
```
frangio, line 30: I feel that we should avoid this pattern because it hurts readability, which is important for tests as made explicit [in the new guidelines](https://github.com/frangio/openzeppelin-contracts/blob/5533d7e9fcba862797231c5961ea452a12ba3083/GUIDELINES.md#testing).

My recommendation here would be to use a for loop, sort of like this:

```suggestion
      this.txs = [];
      for (const [account, amount] of Object.entries(amounts)) {
        this.txs.push(await this.votes.mint(account, amount));
      }
```

It doesn't have to be exactly like this but I'm just hoping we can make it a bit more readable with simpler code.

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -9,8 +9,14 @@ const { shouldBehaveLikeVotes } = require('../../../governance/utils/Votes.behav
...
   const [ account1, account2, account1Delegatee, other1, other2 ] = accounts;
-  this.name = 'My Vote';
+  const name = 'My Vote';
   const symbol = 'MTKN';
+  const tokens = [
+    '10000000000000000000000000',
+    '10',
+    '20',
+    '30',
+  ].map(web3.utils.toBN);
```
frangio, line 19: Because of the other arguments passed from `map` (index, etc.) we should do this to be sure they aren't affecting the result:

```suggestion
  ].map(n => web3.utils.toBN(n));
```

test/governance/utils/Votes.behavior.js
```diff
@@ -245,95 +247,95 @@ function shouldBehaveLikeVotes () {
...
-      it('generally returns the voting balance at the appropriate checkpoint', async function () {
-        const t1 = await this.votes.mint(this.account1, this.token1);
-        await time.advanceBlock();
-        await time.advanceBlock();
-        const t2 = await this.votes.burn(this.account1, this.token1);
-        await time.advanceBlock();
-        await time.advanceBlock();
-        const t3 = await this.votes.mint(this.account1, this.token2);
+      it('generally returns the voting balance at the appropriate checkpoint #2', async function () {
+        const blockNumber = await time.latestBlock().then(Number);
```
frangio, line 269:
```suggestion
        const blockNumber = Number(await time.latestBlock());
```

test/governance/utils/Votes.behavior.js
```diff
@@ -245,95 +247,95 @@ function shouldBehaveLikeVotes () {
         expect(await this.votes.getPastTotalSupply(0)).to.be.bignumber.equal('0');
       });
 
-      it('returns the latest block if >= last checkpoint block', async function () {
-        const t1 = await this.votes.mint(this.account1, this.token0);
-        await time.advanceBlock();
+      it('generally returns the voting balance at the appropriate checkpoint #1', async function () {
```
frangio, line 250: This test title can be improved. Why is there `#1` and `#2`? And note that this is testing for the total supply, not the balance.

frangio (approved):
LGTM

frangio:
- https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3949

PR #3932, opened 2023-01-05
Add `isValidERC1271SignatureNow` to SignatureChecker library

### LINKED ISSUE #3912: Add strict ERC1271 Signature Checking in SignatureChecker library
### 🧐 Motivation
In the current `SignatureChecker` library, there is support to check if a signature is valid with the `isValidSignatureNow(..)` function that checks the following:
- If the `hash` and `signature` recover an address that is the same as the `signer`, it returns true.
- When it's not the case, it calls `isValidSignature(..)` on the `signer` assuming it's a contract. 

What would be a great addition to the library is to have another internal function that checks if a signature is valid with **ERC1271** only. So calling `isValidSignature(..)` directly on the signer. 

This function will be useful for functions that want to check the signature **only via ERC1271**. Then this internal function can be re-used in `isValidSignatureNow(..)`.

### 📝 Details

The only thing that I would argue is that the newly created internal function, e.g `isValidERC1271SignatureNow(..)` should return the standard 4 bytes that will be returned by the function in case it's there, and if the function doesn't exist we can return the `0xffffffff` fail value.

```solidity
function isValidERC1271SignatureNow(address contract, bytes32 dataHash, bytes memory signature) public view returns (bytes4) {

      (bool success, bytes memory result) = contract.staticcall(
            abi.encodeWithSelector(IERC1271.isValidSignature.selector, hash, signature)
        );
  
         if (success) return result.length != 0 ? abi.decode(result, (bytes4)) : bytes4(0xffffffff);
         else return 0xffffffff;
}
```
> Code might not be working, just drafting the concept 

In this way, people can use directly in isValidSignature standard function:

```solidity
function isValidSignature(bytes32 dataHash, bytes memory signature) public view returns (bytes4) {
        address _extensionContract = extension1271();  
        return SignatureChecker.isValidERC1271SignatureNow(_extensionContract, dataHash, signature);
}
```

instead of having the following:

```solidity
function isValidERC1271SignatureNow(address contract, bytes32 dataHash, bytes memory signature) public view returns (bool) {
  
        (bool success, bytes memory result) = signer.staticcall(
            abi.encodeWithSelector(IERC1271.isValidSignature.selector, hash, signature)
        );
        return (success &&
            result.length == 32 &&
            abi.decode(result, (bytes32)) == bytes32(IERC1271.isValidSignature.selector));
}
```

In this way, people are doing the work twice in the standard function:

```solidity
function isValidSignature(bytes32 dataHash, bytes memory signature) public view returns (bytes4) {
        address _extensionContract = extension1271();  
        bool validSig = SignatureChecker.isValidERC1271SignatureNow(_extensionContract, dataHash, signature);
        if(validSig) return 0x1626ba7e; // IERC1271.isValidSignature.selector
        else return 0xffffffff; 
}
```

The code above is doing the work twice, first in `isValidERC1271SignatureNow`, we are comparing the `returnValue` to return a boolean, and then in `isValidSignature(..)` we are checking the boolean again to return the bytes4 value.

This function will be mainly used for signature checking related to ERC1271, if you think this issue is relevant, let's discuss.

YamenMerhi:
Fixes #3912 

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


Amxx (approved):
LGTM,

Waiting on @frangio 's review

Amxx (approved):

frangio (dismissed):

frangio (approved):

PR #3934, opened 2023-01-06
Add timestamp based governor with EIP-6372 and EIP-5805

### LINKED ISSUE #3081: Timestamp-based Governor

Amxx:
Fixes #3081

WIP

This includes a lot of retypes that should be safe, but that must be double-checked, and that the plugin needs to understand.

#### PR Checklist

- [x] Update IVotes to include `clock()`
- [x] Refactor `Votes` & `ERC20Votes` to use the value returned by `clock()`
  - [x] Check upgradeability
- [x] Refactor `Governor` to use a `clock()` function
  - [x] Extract the `clock()` value from the token contract
  - [x] Include a fallback for pre-EIP-5805 tokens
  - [x] Check upgradeability
- [x] Refactor the governance helper to support the two mode (blockNumber/timestamp)
- [x] Test the new governor in both modes
  - [x] Including the comp compatibility layer
  - [x] Including the NFT voting
- [x] Test the new governor with legacy tokens
- [ ] Documentation
- [x] Changelog entry


contracts/governance/Governor.sol
```diff
@@ -248,8 +246,10 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
         bytes[] memory calldatas,
         string memory description
     ) public virtual override returns (uint256) {
+        uint256 timepoint = clock();
+
         require(
-            getVotes(_msgSender(), block.number - 1) >= proposalThreshold(),
+            getVotes(_msgSender(), timepoint - 1) >= proposalThreshold(),
```
frangio, line 252: There seems to be a missing property in EIP-5805 saying that `getPastVotes` should be defined over all integers less than the current clock.

contracts/governance/extensions/GovernorPreventLateQuorum.sol
```diff
@@ -62,16 +61,14 @@ abstract contract GovernorPreventLateQuorum is Governor {
     ) internal virtual override returns (uint256) {
         uint256 result = super._castVote(proposalId, account, support, reason, params);
 
-        Timers.BlockNumber storage extendedDeadline = _extendedDeadlines[proposalId];
+        if (_extendedDeadlines[proposalId] == 0 && _quorumReached(proposalId)) {
+            uint64 extendedDeadline = clock().toUint64() + lateQuorumVoteExtension();
```
frangio, line 65: The clock in EIP-5805 should be uint64 IMO. Do you agree?

contracts/governance/utils/Votes.sol
```diff
@@ -49,30 +57,33 @@ abstract contract Votes is IVotes, Context, EIP712 {
...
      * Requirements:
      *
-     * - `blockNumber` must have been already mined
+     * - `timepoint` must be in the past. If operating using block numbers, the block must be already mined.
      */
-    function getPastVotes(address account, uint256 blockNumber) public view virtual override returns (uint256) {
-        return _delegateCheckpoints[account].getAtProbablyRecentBlock(blockNumber);
+    function getPastVotes(address account, uint256 timepoint) public view virtual override returns (uint256) {
+        require(timepoint < clock(), "Votes: future lookup");
+        return _delegateCheckpoints[account].upperLookup(SafeCast.toUint32(timepoint)); // optimistic ?
```
frangio, line 69: What do you mean by the "optimistic" comment here?

contracts/governance/utils/Votes.sol
```diff
@@ -152,21 +163,37 @@ abstract contract Votes is IVotes, Context, EIP712 {
     function _moveDelegateVotes(address from, address to, uint256 amount) private {
         if (from != to && amount > 0) {
             if (from != address(0)) {
-                (uint256 oldValue, uint256 newValue) = _delegateCheckpoints[from].push(_subtract, amount);
+                (uint256 oldValue, uint256 newValue) = _push(
+                    _delegateCheckpoints[from],
+                    _subtract,
+                    SafeCast.toUint224(amount)
```
frangio, line 169: Since we defined the `_push` function, can we move the cast inside of it?

frangio (commented):
Some initial comments.

test/helpers/time.js
```diff
@@ -0,0 +1,19 @@
+const { time } = require('@openzeppelin/test-helpers');
+
+const clock = () => web3.eth.getBlock('latest').then(block => block.number);
```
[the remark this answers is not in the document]
Amxx: Its not. I liked the idea of having a blockNumber as a default, but we can remove it.

contracts/governance/Governor.sol
Amxx: Right. We have
> For any timepoint that is strictly smaller than clock(), the value returned by getPastVotes should be constant.

but we should add a constraint that it should not revert!

contracts/governance/extensions/GovernorPreventLateQuorum.sol
Amxx: Yes and no.

Yes:
- uint64 is enough for all realistic usecases
- if someone created an 5805 implementation that uses more than 64 bit integers for the clock, our governor would not work with it.

No:
- I don't want to change `getPastVotes(address,uint256)` selector for backward compatibility. 

2 options:
- we change `clock()` return type but don't change `getPastVotes` selector → this cause a "small" inconsistency
- we had a recommendation in the `clock` function (that is not very strict :/)

I think there is an EIP somewhere that wants to have stronger typing on `block.number` and `block.timestamp`, enforcing a specific type (and possibly not uint by default)

contracts/governance/utils/Votes.sol
Amxx: - `getAtProbablyRecentBlock` was optimistic.
- `upperLookup` is not.

The comment is a TODO, should we re-enforce optimistic lookups here (we don't have the right function available in Trace224).

contracts/governance/utils/Votes.sol
Amxx: I thought `_push` would be cleaner using uint224 for both the operator function AND the delta.

contracts/governance/utils/Votes.sol
frangio: Oh you meant the "recenct" function variant. We should definitely use that. Would this be `upperLookupRecent`? Does that name make sense?

contracts/governance/utils/Votes.sol
frangio: I see what you mean. I agree.

frangio:
FYI I removed this PR from the milestone because the corresponding issue is already in it.

contracts/governance/extensions/GovernorVotesComp.sol
```diff
@@ -18,14 +18,25 @@ abstract contract GovernorVotesComp is Governor {
         token = token_;
     }
 
+    /**
+     * See EIP 5805.
+     */
+    function clock() public view virtual override returns (uint256) {
```
frangio, line 24: Do you think this function needs to be public? What we really want to add and make public is `CLOCK_MODE` I think, but `clock` serves no use externally I think.

contracts/governance/utils/Votes.sol
```diff
@@ -29,18 +29,26 @@ import "./IVotes.sol";
...
+    mapping(address => Checkpoints.Trace224) private _delegateCheckpoints;
+    Checkpoints.Trace224 private _totalCheckpoints;
 
     mapping(address => Counters.Counter) private _nonces;
 
+    /**
+     * @dev Clock used for flagging checkpoints. Can be overridden to implement timestamp based checkpoints (and voting).
+     * See EIP-5805.
+     */
+    function clock() public view virtual returns (uint256) {
```
frangio, line 48: We need to add `CLOCK_MODE`. Not mandatory but recommended by the EIP ("SHOULD").

Are we planning to offer a timestamp extension? We could consider a contract `ClockTimestamp` implementing the two clock functions which could be combined with `Votes` and `ERC20Votes`, although the overrides aren't very clean (because it's not chaining via super).

test/helpers/time.js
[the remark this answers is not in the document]
frangio: Yes I would remove what we don't use.

contracts/governance/extensions/GovernorVotesComp.sol
Amxx: I think that the clock used by the governor should be exposed, just like it is exposed by the token. Other contracts might want to connect to a governor, and if we don't expose the `clock()` we are seriously limiting composability.

`CLOCK_MODE` should be a complement on top of `clock()` and not the other way around.

contracts/governance/utils/Votes.sol
Amxx: I'm not sure how the inheritance tree between `ClockTimestamp` and `Votes` would work. Maybe we will need a `Clock` interface ... but interfaces have only external and we need clock to be public

contracts/governance/Governor.sol
```diff
@@ -29,22 +28,22 @@ import "./IGovernor.sol";
...
 
     bytes32 public constant BALLOT_TYPEHASH = keccak256("Ballot(uint256 proposalId,uint8 support)");
     bytes32 public constant EXTENDED_BALLOT_TYPEHASH =
         keccak256("ExtendedBallot(uint256 proposalId,uint8 support,string reason,bytes params)");
 
     struct ProposalCore {
-        Timers.BlockNumber voteStart;
-        Timers.BlockNumber voteEnd;
+        uint256 voteStart; // retyped from Timers.BlockNumber
+        uint256 voteEnd; // retyped from Timers.BlockNumber
```
frangio, line 38: Even though this works currently, I'm not sure it will remain working in future versions of Upgrades Plugins because it doesn't meet our "safety" criteria. We should do something like this:

```suggestion
        uint224 __gap_unused0; // backwards compatibility with <4.9
        uint32 voteStart; // retyped from Timers.BlockNumber
        uint224 __gap_unused1; // backwards compatibility with <4.9
        uint32 voteEnd; // retyped from Timers.BlockNumber
```

contracts/governance/Governor.sol
```diff
@@ -29,22 +28,22 @@ import "./IGovernor.sol";
...
         bool executed;
         bool canceled;
         address proposer;
```
frangio, line 41: Now that I think of it we should be able to put this inside the unused space from the timers!

ernestognw: Shouldn't that break backwards compatibility? We're changing the positions of variables in storage.

ernestognw: Oh but this gets reset after executing. I'm worried about the order of deletion. For example, changing the slot first and then deleting the previous slot.

Where is that order?

contracts/governance/extensions/GovernorVotes.sol
```diff
@@ -4,28 +4,51 @@
...
     constructor(IVotes tokenAddress) {
-        token = tokenAddress;
+        token = IERC5805(address(tokenAddress));
+    }
+
+    /**
+     * See EIP 5805.
+     */
+    function clock() public view virtual override returns (uint48) {
+        try token.clock() returns (uint48 timepoint) {
```
ernestognw, line 25: Aren't these prone to return bombing?

```solidity
contract Token {
    function clock(bytes4 interfaceId) external pure returns (bytes memory) {
        assembly{
            revert(0, 1_000_000)
        }
    }
}
```

Or is it safe because it only copies `uint48` from the `RETURNDATA`?

ernestognw (commented):
Will continue tomorrow.

contracts/governance/Governor.sol
Amxx: I think it should be the other way around

```
        uint64 voteStart; // retyped from Timers.BlockNumber
        uint192 __gap_unused0; // backwards compatibility with <4.9
        uint64 voteEnd; // retyped from Timers.BlockNumber
        uint192 __gap_unused1; // backwards compatibility with <4.9
```

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -103,7 +103,7 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
         ProposalDetails storage details = _proposalDetails[proposalId];
 
         require(
-            _msgSender() == details.proposer || getVotes(details.proposer, block.number - 1) < proposalThreshold(),
+            _msgSender() == details.proposer || getVotes(details.proposer, clock() - 1) < proposalThreshold(),
```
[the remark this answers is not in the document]
Amxx: I don't think so. We should discuss that during a call.

contracts/governance/extensions/GovernorVotes.sol
Amxx: The Token is not an arbitrary contract. Its one contract that the governor heavily relies on. IMO, the governor should assume the token is safe.

That being said, we could write it in assembly to prevent that. I hate that we often have to do it, because it reduces readability. Not sure if he want/need to do it here.

contracts/governance/Governor.sol
Amxx: If I do

```solidity
pragma solidity ^0.8.17;

struct Struct1 {
    uint256 value;
}

struct Struct2 {
    uint8 valueBefore;
    uint240 __gap;
    uint8 valueAfter;
}

function getStruct1(bytes32 slot) pure returns (Struct1 storage r) { assembly { r.slot := slot } }
function getStruct2(bytes32 slot) pure returns (Struct2 storage r) { assembly { r.slot := slot } }

contract T {
    bytes32 SLOT = keccak256("some.storage.slot");

    constructor(uint256 value) {
        getStruct1(SLOT).value = value;
    }
    
    function get1() external view returns (Struct1 memory) {
        return getStruct1(SLOT);
    }  

    function get2() external view returns (Struct2 memory) {
        return getStruct2(SLOT);
    }
}
```

And call the constructor with `0xEEFF`

- `get1()` returns `61183`
- `get2()` returns `255,238,0`

Amxx: Similarly

```solidity
struct M1 {
    mapping(uint256 => uint256) value;
}

struct M2 {
    mapping(uint256 => uint64) value;
}

contract T2 {
    bytes32 constant SLOT = keccak256("some.storage.slot");
    uint256 constant KEY = uint256(keccak256("some.mapping.key"));

    function set(uint256 v) public {
        bytes32 slot = SLOT;
        M1 storage ptr;
        assembly { ptr.slot := slot }
        ptr.value[KEY] = v;
    }

    function get() public view returns (uint64) {
        bytes32 slot = SLOT;
        M2 storage ptr;
        assembly { ptr.slot := slot }
        return ptr.value[KEY];
    }
}
```

works

contracts/interfaces/IERC5805.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.8.0) (interfaces/IERC4626.sol)
```
ernestognw, line 2:
```suggestion
// OpenZeppelin Contracts (interfaces/IERC5805.sol)
```

contracts/interfaces/IERC6372.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.8.0) (interfaces/IERC4626.sol)
```
ernestognw, line 2:
```suggestion
// OpenZeppelin Contracts (interfaces/IERC6372.sol)
```

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -24,7 +24,7 @@ import "../../../utils/cryptography/ECDSA.sol";
  *
  * _Available since v4.2._
  */
-abstract contract ERC20Votes is IVotes, ERC20Permit {
+abstract contract ERC20Votes is IERC5805, ERC20Permit {
     struct Checkpoint {
         uint32 fromBlock;
```
ernestognw, line 29:
```suggestion
        uint32 fromTimepoint;
```

contracts/governance/extensions/GovernorVotes.sol
ernestognw: > That being said, we could write it in assembly to prevent that.

Thanks for the explanation, but I don't think it should be done at the moment. However, this is something to consider in 6372's security considerations, isn't it?

Amxx: > However, this is something to consider in 6372's security considerations, isn't it?

EIP-6372 is about what contract should return (to be compliant) ... not what nasty contract might do wrong. As an example, EIP-20 doesn't (and shouldn't) mention return bombing by the `decimals` or `totalSupply` function.

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
```diff
@@ -107,12 +110,12 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
...
         if (oldQuorumNumerator != 0 && _quorumNumeratorHistory._checkpoints.length == 0) {
             _quorumNumeratorHistory._checkpoints.push(
-                Checkpoints.Checkpoint({_blockNumber: 0, _value: SafeCast.toUint224(oldQuorumNumerator)})
+                Checkpoints.Checkpoint224({_key: 0, _value: oldQuorumNumerator.toUint224()})
             );
         }
 
         // Set new quorum for future proposals
-        _quorumNumeratorHistory.push(newQuorumNumerator);
+        _quorumNumeratorHistory.push(uint256(clock()).toUint32(), newQuorumNumerator.toUint224());
```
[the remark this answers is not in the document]
Amxx: The current `history` / `trace224` uses uint32 for storing the "clock". This is something we inherit from Compound:
https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/Comp.sol#L27-L30

For block numbers is mostly ok. For timestamp (this has been discussed many times) this will overflow in `2106`. We already spent ~50 years since 1970, we have ~75 years left.

In EIP-6372, we decided to go with `uint48` for the clock ... and that is what we will use ... in 5.0
Right now we just cant change the type of the structure in case a governor wants to upgrade from <4.9 to 4.9.

NOTE: Its not just about the `updateQuorumNumerator`, its mostly about the voting token that uses the same clock.

contracts/governance/IGovernor.sol
```diff
@@ -3,14 +3,15 @@
 
 pragma solidity ^0.8.0;
```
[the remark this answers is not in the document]
Amxx: should be good now

contracts/token/ERC20/extensions/ERC20Votes.sol
Amxx: If we rename here, we will require some instruction for the upgrades plugin.

contracts/governance/extensions/GovernorVotes.sol
ernestognw: I'm solving, but I'll still keep pushing for that security consideration in the EIP. EIP-20 didn't have it because there's no way to backwards update an EIP. Following EIP guidelines:

> Security Considerations: ... E.g. include security-relevant design decisions, concerns, important discussions, implementation-specific guidance and pitfalls, an outline of threats and risks and how they are being addressed. EIP submissions missing the “Security Considerations” section will be rejected. An EIP cannot proceed to status “Final” without a Security Considerations discussion deemed sufficient by the reviewers.

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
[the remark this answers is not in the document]
ernestognw: It's fine with both blocks and timestamps. I'm asking if there's a use case for a more granular clock since that might unexpectedly alter the behavior. Strictly talking, there's a gap between `uint32` and `uint48` that's dangerous.

Solving because of the compatibility constraint, but let's keep it in mind if there's another clock we should care about (not just Governor but other implementations).

contracts/token/ERC20/extensions/ERC20Votes.sol
ernestognw: Is this the discussion we've had about the annotation for struct variables?

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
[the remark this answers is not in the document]
frangio: We agree that 32 bits is not enough for some legitimate use cases, but the current code is using 32 bits for backwards compatibility. The new EIPs (5805, 6372) have settled on 48 bits which we believe should be enough for reasonable granularity. Contracts 5.0 will be changed to use 48 bits in accordance with the EIPs.

@ernestognw I'm not sure if you're making a particular suggestion here.

I would say that we should add a warning to the contract in 4.9 that timepoints need to fit in 32 bits.

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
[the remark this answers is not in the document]
ernestognw: Yes. Thanks!

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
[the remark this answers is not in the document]
ernestognw: > We agree that 32 bits is not enough for some legitimate use cases
Amazing, thanks!

> @ernestognw I'm not sure if you're making a particular suggestion here.

I'm asking because this conversation is new for me.
The warning for now sounds good to me.

contracts/governance/IGovernor.sol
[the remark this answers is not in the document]
Amxx: did you refresh the changes. I can't find these in the current version

contracts/token/ERC20/extensions/ERC20Votes.sol
Amxx: Another argument, there is this function that we have for backward compatibility (with Compound)
```solidity 
function checkpoints(address account, uint32 pos) public view virtual returns (Checkpoint memory)
```

Changing the field name would be modifying the ABI/interface and possibly break compatibility with some tooling

contracts/governance/IGovernor.sol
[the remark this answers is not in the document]
ernestognw: No, I think I was indeed reviwing an outdated. Thanks!

contracts/governance/utils/Votes.sol
Amxx: Added

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -145,10 +142,12 @@ abstract contract GovernorTimelockCompound is IGovernorTimelock, Governor {
 
         uint256 eta = proposalEta(proposalId);
         if (eta > 0) {
+            // update state first
+            delete _proposalTimelocks[proposalId];
+            // do external call later
```
frangio, line 147: Why is this?

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
```diff
@@ -14,10 +14,13 @@ import "../../utils/math/SafeCast.sol";
  * _Available since v4.3._
  */
 abstract contract GovernorVotesQuorumFraction is GovernorVotes {
-    using Checkpoints for Checkpoints.History;
+    using SafeCast for uint256;
+    using Checkpoints for Checkpoints.Trace224;
 
     uint256 private _quorumNumerator; // DEPRECATED
```
frangio, line 20: This is off topic but we should add some explanation here of why this is deprecated / what it was replaced with.

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
```diff
@@ -40,23 +43,23 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
...
-        Checkpoints.Checkpoint memory latest = _quorumNumeratorHistory._checkpoints[length - 1];
-        if (latest._blockNumber <= blockNumber) {
+        Checkpoints.Checkpoint224 memory latest = _quorumNumeratorHistory._checkpoints[length - 1];
+        if (latest._key <= timepoint) {
             return latest._value;
         }
 
         // Otherwise, do the binary search
-        return _quorumNumeratorHistory.getAtBlock(blockNumber);
+        return _quorumNumeratorHistory.upperLookup(timepoint.toUint32());
```
frangio, line 62: Is there a reason why this wouldn't use the "recent" variant?

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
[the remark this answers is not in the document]
frangio: Ok let's add a warning. I suppose it should go in `GovernorVotes`?

frangio, line 118: This seems to compile:

```suggestion
        _quorumNumeratorHistory.push(clock().toUint32(), newQuorumNumerator.toUint224());
```

contracts/governance/utils/Votes.sol
```diff
@@ -28,19 +28,40 @@ import "./IVotes.sol";
...
 
+    /**
+     * @dev Clock used for flagging checkpoints. Can be overridden to implement timestamp based checkpoints (and voting).
+     */
+    function clock() public view virtual override returns (uint48) {
+        return uint48(block.number);
+    }
+
+    /**
+     * @dev Description of the clock
```
frangio, line 56:
```suggestion
     * @dev Machine-readable description of the clock as specified in EIP-6372.
```

contracts/governance/utils/Votes.sol
```diff
@@ -28,19 +28,40 @@ import "./IVotes.sol";
...
+    /// @custom:oz-retyped-from mapping(address => Checkpoints.History)
+    mapping(address => Checkpoints.Trace224) private _delegateCheckpoints;
+
+    /// @custom:oz-retyped-from Checkpoints.History
+    Checkpoints.Trace224 private _totalCheckpoints;
 
     mapping(address => Counters.Counter) private _nonces;
 
+    /**
+     * @dev Clock used for flagging checkpoints. Can be overridden to implement timestamp based checkpoints (and voting).
```
frangio, line 49:
```suggestion
     * @dev Clock used for flagging checkpoints. Can be overridden to implement timestamp based
     * checkpoints (and voting), in which case {CLOCK_MODE} should be overriden as well to match.
```

contracts/token/ERC20/extensions/ERC20Votes.sol
frangio: >modifying the ABI/interface and possibly break compatibility with some tooling

Is that something we consider breaking? We've changed event name arguments in the past.

contracts/governance/utils/Votes.sol
```diff
@@ -28,19 +28,40 @@ import "./IVotes.sol";
...
+     */
+    function clock() public view virtual override returns (uint48) {
+        return uint48(block.number);
```
frangio, line 52: Should be using SafeCast

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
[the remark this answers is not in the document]
ernestognw: Hadrien also mentioned the voting tokens, so maybe every `IVotes` implementers as well

contracts/token/ERC20/extensions/ERC20Votes.sol
ernestognw: All of the comparisons against `block.number` is replaced with the clock, so no problem with the naming, it doesn't change the interface.

contracts/governance/Governor.sol
```diff
@@ -162,13 +169,15 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
             revert("Governor: unknown proposal id");
         }
 
-        if (snapshot >= block.number) {
+        uint256 timepoint = clock();
```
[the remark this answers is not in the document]
Amxx: now is shadowing a solidity built in function. That is why we named the function `clock()` (I first wanted to name it `now()`.

`currentTimepoint` is fine!

contracts/governance/Governor.sol
```diff
@@ -29,22 +28,27 @@ import "./IGovernor.sol";
...
-    using Timers for Timers.BlockNumber;
 
     bytes32 public constant BALLOT_TYPEHASH = keccak256("Ballot(uint256 proposalId,uint8 support)");
     bytes32 public constant EXTENDED_BALLOT_TYPEHASH =
         keccak256("ExtendedBallot(uint256 proposalId,uint8 support,string reason,bytes params)");
 
     struct ProposalCore {
-        Timers.BlockNumber voteStart;
-        Timers.BlockNumber voteEnd;
+        // --- start retyped from Timers.BlockNumber at offset 0x00 ---
```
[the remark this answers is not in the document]
Amxx: I planed to moving it to UDVT in 5.0 ... we can also just remove it. Either is fine by me.

If we go for UDVT in uint48, we can use one bit to say if the timer is blocknumber or timestamp and use the remaining 47 bits to store the value, that way the timer is transparent ...

Lets discuss that after 4.9.

.github/actions/storage-layout/action.yml
```diff
@@ -40,7 +40,7 @@ runs:
     - name: Compare layouts
       if: steps.reference.outcome == 'success' && github.event_name == 'pull_request'
       run: |
-        node scripts/checks/compare-layout.js --head ${{ inputs.layout }} --ref ${{ inputs.ref_layout }} >> $GITHUB_STEP_SUMMARY
+        node scripts/checks/compare-layout.js --head ${{ inputs.layout }} --ref ${{ inputs.ref_layout }}
```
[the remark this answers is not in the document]
Amxx: Its confusing to have to go to the summary to get the details.

It made sens to send the gasreport there, because its something we want to have available for review without having to care what step produced it. But here we only check it is there is an error.

- either there are no errors, and we don't care where its send (report is empty)
- nor there are errors, and we expect to see them on the test that failed, not somewhere else.

Also, in GITHUB_STEP_SUMMARY it doesn't look that good (formating)

contracts/governance/Governor.sol
```diff
@@ -29,22 +28,27 @@ import "./IGovernor.sol";
 abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receiver, IERC1155Receiver {
```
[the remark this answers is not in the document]
Amxx: IGovernor is already `IERC6372`

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -100,10 +100,10 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
     }
 
     function cancel(uint256 proposalId) public virtual override(IGovernor, Governor) {
-        ProposalDetails storage details = _proposalDetails[proposalId];
+        address proposer = _proposalDetails[proposalId].proposer;
```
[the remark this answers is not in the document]
Amxx: I plan to remove that in 5.0

Cannot remove it earlier, because in an upgrade from 4.8 to 4.9 we would lose the proposer info, that is not in the core. 

We could do a fallback lookup in the core, and only write/read from here if the core doesn't have the data ... is that worth it for 4.9?

contracts/governance/extensions/GovernorTimelockCompound.sol
Amxx: now that `cancel` is public, slither detects a possible "reentrancy attack" here because `_timelock.cancelTransaction` is an external call that is done before a state change.

In fact I don't think there is any risk, because `super._cancel` does the reentrency protection already.


It just to calm slither

contracts/governance/Governor.sol
```diff
@@ -263,16 +269,14 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
-        uint64 deadline = snapshot + votingPeriod().toUint64();
+        uint256 snapshot = timepoint + votingDelay();
+        uint256 deadline = snapshot + votingPeriod();
 
-        proposal.voteStart.setDeadline(snapshot);
-        proposal.voteEnd.setDeadline(deadline);
-        proposal.proposer = proposer;
+        _proposals[proposalId].proposer = proposer;
+        _proposals[proposalId].voteStart = snapshot.toUint64();
+        _proposals[proposalId].voteEnd = deadline.toUint64();
```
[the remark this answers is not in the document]
Amxx: `_proposals[proposalId] = ProposalCore({ ... })`  require to all the filed, and write the entire structure.

I would look like
```solidity
        _proposals[proposalId] = ProposalCore({
            proposer: proposer,
            voteStart: snapshot.toUint64(),
            voteEnd: deadline.toUint64(),
            executed: false,
            canceled: false,
            __gap_unused0: 0,
            __gap_unused1: 0
        });
```
This result in 3 sstores (the size of the struct):
- 2 from zero to non zero: 22100 x 2 = 44200
- 1 from zero to zero: 2200

The current version does 3 sstores:
- 2 from zero to non zero: 22100 x 2 = 44200
- 1 from non zero to zero: 2200

So both operation should cost the same in storage, there are unclear differences in memory preparation / bit shifting ...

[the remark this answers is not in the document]
Amxx: After experimenting, your approach saves ~23k gas of contract deployment cost and ~2k gas per proposal 

Not exactly sure why, but that is good!

contracts/governance/Governor.sol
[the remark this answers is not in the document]
Amxx: FYI: https://gist.github.com/Amxx/ee86e8b0783604c73cf5112e00339cd5

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
Amxx: `Member "toUint32" not found or not visible after argument-dependent lookup in uint48.`

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
Amxx: We already do an optimistic lookup 
```solidity
// Optimistic search, check the latest checkpoint
Checkpoints.Checkpoint224 memory latest = _quorumNumeratorHistory._checkpoints[length - 1];
if (latest._key <= timepoint) {
    return latest._value;
}
```
Whould we remove that "latest" check and use the recent lookup ? should we do both ?

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
frangio: That's just because we only have `using SafeCast for uint256` in this contract. We can change that to `using SafeCast for *` or simply add `uint48`.

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
frangio: Ah I see. I _think_ it makes sense to do both latest and recent.

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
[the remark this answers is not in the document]
frangio: The code complexity is not too much, what is complicated is managing the differences between vanilla and upgradeable contracts if we don't want the fallback logic to be present in the vanilla contracts.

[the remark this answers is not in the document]
Amxx: So do we want it?

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
```diff
@@ -69,8 +72,8 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
     /**
      * @dev Returns the quorum for a block number, in terms of number of votes: `supply * numerator / denominator`.
```
frangio, line 73:
```suggestion
     * @dev Returns the quorum for a timepoint, in terms of number of votes: `supply * numerator / denominator`.
```

frangio:
The event `ProposalCreated` has parameters `startBlock` and `endBlock`. I think we should rename these. This also raises an interesting question that the event doesn't say whether the timepoints are blocks or timestamps. But I suppose it just means you need to consult the clock mode with the Governor contract.

frangio:
The storage layout errors we are seeing, which say "Layout could have changed...", are because the compiler is not producing storage layout information so the Upgrades plugin conservatively reports that there could be some error. I've added the layout in the config in https://github.com/OpenZeppelin/openzeppelin-contracts/commit/3b591a48acaab78008ed39d60fbcf429a83155ca. After that finishes producing the new layout artifact, we should merge and run it in this PR.

frangio:
The diff test coverage is not looking super good. [See on Codecov](https://app.codecov.io/gh/OpenZeppelin/openzeppelin-contracts/pull/3934). It seems to be in large parte because `CLOCK_MODE` is never tested. What do you think about adding an EIP-6372 behavior test file?

Amxx:
> What do you think about adding an EIP-6372 behavior test file?

done

test/governance/Governor.test.js
```diff
@@ -1,630 +1,648 @@
...
+        expectEvent(txPropose, 'ProposalCreated', {
+          proposalId: this.proposal.id,
+          proposer,
+          targets: this.proposal.targets,
+          // values: this.proposal.values,
+          signatures: this.proposal.signatures,
+          calldatas: this.proposal.data,
+          startBlock: web3.utils.toBN(await clockFromReceipt[mode](txPropose.receipt)).add(votingDelay),
+          endBlock: web3.utils
+            .toBN(await clockFromReceipt[mode](txPropose.receipt))
```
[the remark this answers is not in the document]
Amxx: There are many places (25 occurences) where we do:

```javascript
const timepoint = await clockFromReceipt[mode](receipt);
// [...]
expect(await this.token.getPastVotes(holder, timepoint - 1)).to.be.bignumber.equal('0');
```

we would need to to `timepoint.subn(1)` which affects readability.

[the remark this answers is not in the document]
Amxx: this is to compare to 9 occurences of 
```javascript
.toBN(await clockFromReceipt[mode](...))
```

frangio (approved):
This is looking good to me.

Documentation still needs work but as discussed elsewhere we can do this in a follow up PR before release.

contracts/governance/Governor.sol
```diff
@@ -109,7 +109,6 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
                 this.getVotesWithParams.selector) ||
             // Previous interface for backwards compatibility
             interfaceId == (type(IGovernor).interfaceId ^ type(IERC6372).interfaceId ^ this.cancel.selector) ||
-            interfaceId == type(IGovernor).interfaceId ||
```
frangio, line 104: Why are we removing this?

frangio (approved):
LGTM!

contracts/governance/Governor.sol
Amxx: Yes, also I'm not sure we want to have that many ERC165 identifier that we'll have to maintain forever.

jonwalch:
When is this getting released?

Amxx:
@jonwalch in the next minor (4.9) in a few weeks.

jonwalch:
> @jonwalch in the next minor (4.9) in a few weeks.

Hey @Amxx, any update on when 4.9 comes out? Thanks!

jonwalch:
All good. Thanks for the update!

PR #3939, opened 2023-01-09
🌗 Support for old solidity versions and new solidity versions.

Chainmaven:
There is a new version of solidity that I wish to use with Openzeppelin support.

Fixes #???? 

 - Version change

- Supporting the new 0.9.0 version of solidity

#### PR Checklist

- [x] Tests ✅
- [x] Documentation ✅
- [x] Changelog entry ✅

### DIFF
--- a/contracts/security/ReentrancyGuard.sol
+++ b/contracts/security/ReentrancyGuard.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v4.8.0) (security/ReentrancyGuard.sol)
 
-pragma solidity ^0.8.0;
+pragma solidity >=0.7.0 <0.9.0;
 
 /**
  * @dev Contract module that helps prevent reentrant calls to a function.


frangio:
> * Supporting the new 0.9.0 version of solidity

There is no 0.9 version as far as I can tell. What do you mean?

Chainmaven:
I meant 0.8.9 
0.8.9  means  : < 0.9.0

Amxx:
The `4.X` version of OpenZeppelin contracts are compatible with all version matching `^0.8.9`.

This include `0.8.9` and newer.

Note that we usually recommend using the latest version, which at this time is 0.8.17

Amxx:
Please note that `^0.8.0` means
- 0.8.0
- 0.8.1
- 0.8.2
- 0.8.3
- ...
- 0.8.13194259314613296
- ...

Amxx:
What you propose (`pragma solidity >=0.7.0 <0.9.0;`) includes all `0.7.x` version of solidity, which the latest code doesn't support.

frangio:
> which the latest code doesn't support

This specific contract actually does compile with 0.7.0 so we could in theory adopt the pragma that is suggested in this PR. But we wouldn't be checking that the contract indeed compiles with 0.7.0 so future changes might break that and go unnoticed. Because of that, I would prefer to keep the pragma as it is currently.

@Chainmaven Let us know if you have any other thoughts about this.

PR #3943, opened 2023-01-11
Use a staticcall to fetch ERC20.decimals in ERC4626

Amxx:
Fixes an informational issue raised through immunefi.

Do we need a changelog entry for that?

#### PR Checklist
- [x] Changelog entry


frangio:
> Do we need a changelog entry for that?

Yes definitely.

Masoudahgl (commented):
Like

Issue #2845, opened 2021-09-02
Make `permit` compatible with smart-contract wallets like Gnosis and Argent


juniset:
### Motivation

Make the `permit` method compatible with smart-contract wallets that implement EIP1271, and make smart-contract wallets compatible with dapp flows that leverage the `permit` method (e.g. via WalletConnect).

### Requested change

Replace `ECDSA.recover` by `SignatureChecker. isValidSignatureNow` in https://github.com/OpenZeppelin/openzeppelin-contracts/blob/aefcb3e8aa4ee8da8e2b7022ffe4dcb57fbb0fdf/contracts/token/ERC20/extensions/draft-ERC20Permit.sol#L55 .

Amxx:
Thank you @juniset for raissing this issue.

frangio:
We really want to do this, but [the current text of EIP-2612](https://eips.ethereum.org/EIPS/eip-2612#specification) is very explicit that the permit is an ECDSA signature by the token owner address:

> - `r`, `s` and `v` is a valid secp256k1 signature from `owner` of the message [...]

So we don't feel comfortable making this change unilaterally, and the EIP is still in Draft so there is an opportunity to make this change directly in the spec. The fact that it is backwards compatible makes me hopeful that it should be accepted without much pushback.

We think the best course of action is to propose a pull request in [ethereum/EIPs](https://github.com/ethereum/EIPs) with the required change to EIP-2612. @juniset What do you think?

juniset:
@frangio I agree that the spec of EIP-2612 should be clarified to include EIP-1271 signatures as being valid. I had raised that issue a while ago (https://github.com/ethereum/EIPs/issues/2613#issuecomment-712663371) and mentioned it again today.

However, one could argue that using `SigantureChecker` instead of `ECDSA` in your implementation today is compliant with the specification since a EIP-1721 signature is a valid secp256k1 signature from the owner of the message when the owner is a smart-contract, i.e. it depends of your definition of "valid".

Adding it today would also give a signal to the supporters of EIP-2612 and encourage them to clarify the definition of a valid signature.

frangio:
@juniset Just to clarify I meant opening a pull request directly with the change ready to be accepted by the authors of the EIP. I think this will be a more sure way to get it accepted.

PR #3951, opened 2023-01-12
Use SignatureChecker inside Permit to allow smart contract wallets to use it

k06a:
I see current implementation of EIP-2612 (aka Permit) is not supporting smart contract wallets and they are required to invent own signature-based operations to execute token approve by smart contract wallet owner signature (pseudocode example: https://ethereum.stackexchange.com/a/142814/3032). Trying to fix it without violating EIP-2612 specification.

First ([EIP-2612](https://eips.ethereum.org/EIPS/eip-2612)) defines signature as following:
> `r`, `s` and `v` is a valid `secp256k1` signature from `owner`

But if `owner` is smart contract wallet, can we consider signature to be valid `secp256k1` signature but of wallet `owner()` of ERC20 `owner`?

I think it's possible and doesn't violate EIP that much. Let's discuss.


Amxx:
Hello @k06a 

This was already discussed (can't find the issue though :/) and we decided against doing it for the following reasons:

- smart wallets are likely to support meta transaction (and batch transactions), therefore the approve-transferFrom is not as bad for them as it is for EOA
- some smart wallets want to limit daily/monthly transfers (that was raised by argent). If the transfers are done by the smart wallet, directly calling the `ERC20.transfer` function, then its possible to enforce that onchain. However, if the transfers are the consequence of an permit, then the quota can no longer be enforced.

frangio:
The issue is https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2845.

The main reason is that EIP-2612 specifically says `v,r,s` a signature by the owner.

frangio:
> But if `owner` is smart contract wallet, can we consider signature to be valid `secp256k1` signature but of wallet `owner()` of ERC20 `owner`?
> 
> I think it's possible and doesn't violate EIP that much. Let's discuss.

I see this comment now. It's true that it doesn't violate the EIP that much... EIP-1271 goes beyond these signatures though. For example, Gnosis Safe's implementation wouldn't work with this. What wallets would be enabled by this change?

k06a:
@frangio I think smart contract wallets will be able to let receiver claim their tokens without performing transaction. In case of permit to other wallet (EOA or SC).

k06a:
BTW Uniswap’s permit2 supports EIP1271: https://github.com/Uniswap/permit2/blob/ca6b6ff2b47afc2942f3c67b0d929ca4f0b32631/src/libraries/SignatureVerification.sol#L43

Amxx:
Interresting: they also support both 65 and 64 bytes signatures

k06a:
> Interresting: they also support both 65 and 64 bytes signatures

I think they support any length and effectively support Gnosis Safes

elshan-eth:
I agree with @Amxx's comment.

It seems to me that there is no advantage in using `permit` for `Account Abstraction` because the gas efficiency would either be too negligible, but it introduces additional "variables" that need to be considered when working with tokens. For example, if `Account Abstraction` decides to prohibit the use of `permit`, as in the case of `Argent`, it would add extra complexity on the side of `AA` contracts, but the absence of the `permit` function wouldn't cause any inconvenience when using `AA`. Any `AA` wallet focuses on signature management within its own contracts, and built-in batching eliminates the issues that `permit` was designed to solve.

k06a:
USDC finally supports ERC-1271 signatures, could be done as new ERC, extending ERC-2612…

elshan-eth:
> USDC finally supports ERC-1271 signatures, could be done as new ERC, extending ERC-2612…

I don't understand the technical benefits, but if it's becoming standard then it might be worth it

k06a:
@elshan-eth you can’t place limit order from miltisig without manually executing approve.

frangio:
It's not just about batching, but it is true that _in theory_ AA allows this to be fixed at the account level. That said it won't happen in practice until there's a standard way to do it.




> USDC finally supports ERC-1271 signatures

For reference: https://github.com/circlefin/stablecoin-evm/blob/a4ba37d0495b1c979ff0dcfdbad2e5e076828208/contracts/v2/EIP2612.sol#L77


I'm not convinced this is worth doing... It will only work for the simplest accounts, it won't work for Safes, it won't work for passkey accounts, etc.

elshan-eth:
@k06a, good example :thumbsup:

k06a:
I think we could have separate `permit()` method with `bytes signature` to support ERC-1271. Exactly the same way as Circle team (USDC) did it.

frangio:
Ah I see, this is what you mean: https://github.com/circlefin/stablecoin-evm/blob/a4ba37d0495b1c979ff0dcfdbad2e5e076828208/contracts/v2/FiatTokenV2_2.sol#L95-L101

k06a:
Could be `calldata`, but yes.

PR #3953, opened 2023-01-12
Use actionlint on GitHub Actions workflows

frangio:
[Actionlint](https://github.com/rhysd/actionlint) is a linter for GitHub Actions workflows. It can detect security problems like:
- https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
- https://github.com/nikitastupin/pwnhub

I set this to run only when the workflows change because it won't happen often and we don't want to see the check all the time.

This is what it would look like on a workflow with errors:

image

Our existing workflows don't have any errors. (Great!)

One thing to note is that _actions_ like our [`setup`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/.github/actions/setup/action.yml) action aren't linted. There is a feature request for this (https://github.com/rhysd/actionlint/issues/46).


.github/workflows/actionlint.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Add problem matchers
+        run: |
+          # https://github.com/rhysd/actionlint/blob/3a2f2c7/docs/usage.md#problem-matchers
+          echo "::add-matcher::.github/actionlint-matcher.json"
```
ernestognw, line 16: Not sure about this but it may make sense to [download the matcher](https://github.com/rhysd/actionlint/blob/main/docs/usage.md#problem-matchers:~:text=Copy-,actionlint%2Dmatcher.json,-to%20.github/actionlint), and put it in the cache with the SHA of the file.

Just in case the matches update but doesn't seem to change frequently and it's what the docs recommend, so keeping it like this is fine.

ernestognw (approved):

.github/workflows/actionlint.yml
frangio: I thought about this but I don't think the matcher will be loaded unless we put it in the repo. Let's try.

frangio: Changed it and it seems to work.

ernestognw: Amazing, thanks!

ernestognw (approved):

PR #3955, opened 2023-01-13
Ignore reentrancy in`executeBatch` and update Slither config

0xalpharush:
@frangio [indicated](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3949) this reentrancy isn't a vulnerability since the callbacks are guarded, so I've added a disable comment to `executeBatch` This result was detected in slither 0.9.2 because this [bug](https://github.com/crytic/slither/issues/1019) was fixed and now reentrancy in for loops is considered properly.  

Additionally, I've configured slither to compile with hardhat and upgraded its version in the GitHub action to 0.9.2.

ref https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3949

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -77,8 +77,6 @@ jobs:
       - name: Set up environment
         uses: ./.github/actions/setup
       - uses: crytic/slither-action@v0.2.0
-        with:
-          slither-version: 0.9.2
 
   codespell:
     if: github.repository != 'OpenZeppelin/openzeppelin-contracts-upgradeable'
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -76,10 +76,9 @@ jobs:
       - uses: actions/checkout@v3
       - name: Set up environment
         uses: ./.github/actions/setup
-      - run: rm foundry.toml
       - uses: crytic/slither-action@v0.2.0
         with:
-          slither-version: 0.9.1
+          slither-version: 0.9.2
 
   codespell:
     if: github.repository != 'OpenZeppelin/openzeppelin-contracts-upgradeable'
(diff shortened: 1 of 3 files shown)


0xalpharush:
Slither collects variables read prior to external calls and variables written after. If any one those are in common and there are functions that could be called back into (no call to `nonReentrant`), those functions are analyzed to see if they would be impacted i.e. they read from the variable whose value is non-deterministic. In this case, I believe it's discovering that `_beforeCall` and `_afterCall` read/ write to `_timestamp` with an external call in between.

We have added ways to indicate that a smart contract is non-reentrant (trusted), but that's currently limited to state variables https://github.com/crytic/slither/pull/1089. Do you have suggestions on how you would expect this case to be filtered out?

frangio:
> how you would expect this case to be filtered out?

Just more advanced static analysis. :slightly_smiling_face: It probaly exceeds the scope of Slither.

frangio (approved):

frangio:
@0xalpharush I had to add `rm foundry.toml` again because it was failing installation. The failure is due to https://github.com/foundry-rs/foundry/issues/3942, so it's not exactly a Slither issue, but it does show that `"compile_force_framework": "hardhat"` isn't disabling Foundry installation.

Amxx:
Hello @0xalpharush 

We recently noticed that the `"compile_force_framework": "hardhat"` line in the slither config (that was introduced in this PR) is causing issues in the community repo: https://github.com/OpenZeppelin/openzeppelin-community-contracts

On that repo, slither appears to work just fine without this line. We would like to understand waht the consequences would be if we removed it here as well.

Amxx:
FYI, we get an issue with the hardhat option because [this script](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/hardhat/remappings.js), which our hardhat.config.js is apparently not processed by slither ... and the remappings are unaccessible

PR #3956, opened 2023-01-13
Clean up after merging master into next-v5.0

Amxx:


frangio:
To keep the commit history in `next-v5.0` cleaner, I took your merge commit https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3956/commits/fb1e7838b4945ced9180d6cab5fc857c47a380e6 and manually applied it in `next-v5.0` as 147dc890540995f93add4982973850d69e49eb87. You can see that they are identical [here](https://github.com/OpenZeppelin/openzeppelin-contracts/compare/147dc890540995f93add4982973850d69e49eb87..fb1e7838b4945ced9180d6cab5fc857c47a380e6).

This PR now consists of the cleanup commits only.

Note that you're removing `ERC20Mock` but it's necessary for the Foundry tests which are failing since it was removed.

Amxx:
I forgot about the foundry tests, and though that these mocks were supplanted by hardhat-exposed. I reverted the corresponding commits.

frangio (approved):

PR #311, opened 2017-07-13
Correctly call super.transferableTokens in VestedToken

adamdossa:
Fix #307

PR #3958, opened 2023-01-16
Show if event parameter is indexed in docs

JulissaDantes:
Fixes an issue in the docs repo [#311](https://github.com/OpenZeppelin/docs.openzeppelin.com/issues/311) 

The documentation is rendered without showing if an event parameter was indexed or not. This PR modifies the existing helper to consider if indexed too.

To check how it would show run `nom run docs:watch`

Result:
Screen Shot 2023-01-16 at 10 31 56 AM
#### PR Checklist

- [ ] Tests
- [x] Documentation
- [ ] Changelog entry

### DIFF
--- a/docs/templates/helpers.js
+++ b/docs/templates/helpers.js
@@ -11,7 +11,7 @@ module.exports.names = params => params.map(p => p.name).join(', ');
 module.exports['typed-params'] = params => {
   console.log(params);
   if (params == undefined) return [];
-  return params.map(p => `${p.type}${p.indexed ? ' indexed': ''}${p.name ? ' ' + p.name : ''}`).join(', ');
+  return params.map(p => `${p.type}${p.indexed ? ' indexed' : ''}${p.name ? ' ' + p.name : ''}`).join(', ');
 };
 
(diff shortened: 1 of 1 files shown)


docs/templates/helpers.js
```diff
@@ -9,7 +9,9 @@ module.exports['readme-path'] = opts => {
 module.exports.names = params => params.map(p => p.name).join(', ');
 
 module.exports['typed-params'] = params => {
-  return params.map(p => `${p.type}${p.name ? ' ' + p.name : ''}`).join(', ');
+  console.log(params);
+  if (params == undefined) return [];
+  return params.map(p => `${p.type}${p.indexed ? ' indexed' : ''}${p.name ? ' ' + p.name : ''}`).join(', ');
```
frangio, line 14: With `?.` if params is undefined we will return undefined, which should work more or less the same as the empty array.

```suggestion
  return params?.map(p => `${p.type}${p.indexed ? ' indexed' : ''}${p.name ? ' ' + p.name : ''}`).join(', ');
```

PR #3961, opened 2023-01-16
Annotate Multicall contract

### LINKED ISSUE #3843: Annotate MulticallUpgradeable to allow delegatecall for Upgrades Plugins
**🧐 Motivation**
When MulticallUpgradeable is inherited in an upgradeable contract, using the contract with [Upgrades Plugins](https://docs.openzeppelin.com/upgrades-plugins/1.x/) would cause a validation error with
```
@openzeppelin/contracts-upgradeable/utils/MulticallUpgradeable.sol:41: Use of delegatecall is not allowed
    https://zpl.in/upgrades/error-002
```
However, the delegatecall in this context is not necessarily unsafe.

To get around this, users would currently need to provide an option to allow delegatecalls for the entire validation run, which could hide issues elsewhere.

**📝 Details**
[MulticallUpgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/MulticallUpgradeable.sol) contains a `delegatecall`, but the target of that `delegatecall` is always `address(this)`.  This delegatecall can be considered safe as long as the contract including its dependencies does not have any exposed functions with `selfdestruct`.

Consider annotating MulticallUpgradeable with `@custom:oz-upgrades-unsafe-allow delegatecall` so that Upgrades Plugins will not raise a validation error for the delegatecall.  Note that the Upgrades Plugins will still perform validations to ensure that the contract does not have selfdestructs.

JulissaDantes:
Fixes #3843  

Annotates `MulticallUpgradeable` with `@custom:oz-upgrades-unsafe-allow delegatecall` so that Upgrades Plugins will not raise a validation error for the delegatecall. Note that the Upgrades Plugins will still perform validations to ensure that the contract does not have selfdestructs.

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [ ] Changelog entry


Amxx (dismissed):
Nice!!!

I've done that manually in the upgrade plugin commands so many times ... it pains me to not have though of that earlier!

frangio:
Let's update the changelog for this one.

frangio (dismissed):

PR #3962, opened 2023-01-16
Add ERC1155.totalSupply that returns overall supply count

### LINKED ISSUE #3301: ERC1155Supply: Add totalSupply() function that returns overall supply count
**🧐 Motivation**
[ERC1155Supply](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.5.0/contracts/token/ERC1155/extensions/ERC1155Supply.sol) only counts by token ID. In some cases we may want a general overall count of all tokens that have been minted under the contract. e.g. etherscan and polygonscan tries to call `totalSupply()` when showing information about an NFT, and shows an error on the page if it failed.

image

**📝 Details**
Having only a count for each token ID in `ERC1155Supply` means to get an overall total count, options are:

- loop through all minted token IDs and call `totalSupply(id)`
- or not use `ERC1155Supply` and within the ERC1155 contract:
-- use [Counters](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Counters.sol). But `Counters` cannot increment by more than 1, which we'd want after a batch mint. Here is an issue about this: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3296
-- or just use `uint256` instead of both `ERC1155Supply` and `Counters`

I still want to use `ERC1155Supply` as its purpose is supply tracking. I propose to add overloaded function `totalSupply()` that returns the overall count.

Some code that could be added are:

```uint256 private _totalSupplyAll;```

```solidity
    function totalSupply() public view virtual returns (uint256) {
        return _totalSupplyAll;
    }
```
In `_beforeTokenTransfer`:
```solidity
        if (from == address(0)) {
                _totalSupply[ids[i]] += amounts[i];
                _totalSupplyAll += amounts[i];
            }
        }

        if (to == address(0)) {
                _totalSupply[ids[i]] -= amounts[i];
                _totalSupplyAll -= amounts[i];
            }
        }
```

JulissaDantes:
Fixes #3301  

This PR adds a tracking storage variable inside the `ERC1155Supply` contract, to keep track of the amount of all the circulating tokens, without having to check each tokenId's total supply.

#### PR Checklist

- [X] Tests
- [X] Documentation
- [x] Changelog entry


Amxx (changes requested):
There is something terribly wrong here:

- `totalAmount` starts at 0
- you increase is for every new token
- you add that sum to the `_totalSupplyAll`
- you then increase it again (without resetting it) with every token being burned
- you decrease that from the `_totalSupplyAll`

So if someone transfers from 0 to 0 (which would happen with the new _update transition), you would mess up the _totalSupplyAll :/

Also, on a completely different point, this change will restrict a total of `2**256-1` tokens minted accross ALL ids. This is a significant change, which needs to be documented.

Amxx:
```
        uint256 totalAmount;
        if (from == address(0)) {
                _totalSupply[ids[i]] += amounts[i];
                totalAmount += amounts[i];
            }
            _totalSupplyAll += totalAmount;
        }

        if (to == address(0)) {
                uint256 id = ids[i];
                uint256 amount = amounts[i];
                uint256 supply = _totalSupply[id];
                require(supply >= amount, "ERC1155: burn amount exceeds totalSupply");
                unchecked {
                    _totalSupply[id] = supply - amount;
                    totalAmount += amount;
                }
            }
            _totalSupplyAll -= totalAmount;
        }
```

should be

```
        if (from == address(0)) {
            uint256 totalMintAmount = 0;
                uint256 amount = amounts[i];
                _totalSupply[ids[i]] += amount;
                totalMintAmount += amount;
            }
            _totalSupplyAll += totalMintAmount;
        }

        if (to == address(0)) {
                uint256 totalBurnAmount = 0;
                uint256 id = ids[i];
                uint256 amount = amounts[i];
                uint256 supply = _totalSupply[id];
                require(supply >= amount, "ERC1155: burn amount exceeds totalSupply");
                unchecked {
                    _totalSupply[id] = supply - amount;
                    totalBurnAmount += amount;
                }
            }
            unchecked {
                _totalSupplyAll -= totalBurnAmount;
            }
        }
```

JulissaDantes:
> Also, on a completely different point, this change will restrict a total of `2**256-1` tokens minted accross ALL ids. This is a significant change, which needs to be documented.

Thanks for pointing this out, now it uses `uint`.

Amxx:
> Thanks for pointing this out, now it uses `uint`.

`uint` is an alias to `uint256`. Using one or the other is strictly equivalent and results in the same bytecode. Any limitation that you have with `uint256` you will also have with `uint`.

We should be using `uint256` all the time, just to avoid possible confusion, and also because function signature & EIP-712 type hash are produced using `uint256`

Amxx (commented):
This is a breaking change on 2 fronts:

- this change restrict the ability to mint token, imposing a new limit that was not there before
- upgrading from the previous version to this is ok from a storage layout point of view... but it would lead to inconsistent values (and even mint reverting)

For the mint reverting: 
- you already have 100 token
- you upgrade, the totalSupply() is 0
- you burn 1 token ... decrease is unchecked so is overflows
- totalSupply() is now 2**256-1
- you try to mint 1 token ... increase is checked and reverts because of the overflow

frangio:
Yeah I share the concerns. Let's move this to 5.0. @JulissaDantes Please rebase or rewrite this on top of next-v5.0.

JulissaDantes:
> Yeah I share the concerns. Let's move this to 5.0. @JulissaDantes Please rebase or rewrite this on top of next-v5.0.

Ready.

contracts/token/ERC1155/extensions/ERC1155Supply.sol
```diff
@@ -41,21 +49,30 @@ abstract contract ERC1155Supply is ERC1155 {
...
         if (to == address(0)) {
+            uint256 totalBurnAmount = 0;
             for (uint256 i = 0; i < ids.length; ++i) {
                 uint256 id = ids[i];
                 uint256 amount = amounts[i];
                 uint256 supply = _totalSupply[id];
                 require(supply >= amount, "ERC1155: burn amount exceeds totalSupply");
                 unchecked {
                     _totalSupply[id] = supply - amount;
+                    totalBurnAmount += amount;
```
frangio, line 70: It isn't obvious to me why this variable won't overflow. We need to add the explanation.

Amxx: It comes from two inequalities:
- `amounts[i] <= balanceOf(from, i)` (from the require on line 67)
- `sum(balanceOf(from, i) <= sum of all balances = totalSupply` (contract invariant)

so `totalBurnAmount = sum(amounts[i]) <= sum(balanceOf(from, i) <= totalSupply <= 2*256-1`

Amxx: Technically we could even get rid of the check on line 67 if we were to check `balance >= amount` before. (the check if done after, because this is the "pre" hook)

CHANGELOG.md
```diff
@@ -53,8 +53,9 @@ function supportsInterface(bytes4 interfaceId) public view virtual override retu
  * `Math`: optimize `log256` rounding check. ([#3745](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3745))
  * `Strings`: add `equal` method. ([#3774](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3774))
  * `Strings`: add `toString` method for signed integers. ([#3773](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3773))
- * `MerkleProof`: optimize by using unchecked arithmetic. ([#3745](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3745))
+ * `MerkleProof`: optimize by using unchecked arithmetic. ([#3869](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3869))
```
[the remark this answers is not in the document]
JulissaDantes: yes, apparently its different on master and next-v5.0.

test/token/ERC1155/extensions/ERC1155Supply.test.js
```diff
@@ -99,9 +109,19 @@ contract('ERC1155Supply', function (accounts) {
...
-        expect(await this.token.totalSupply(secondTokenId)).to.be.bignumber.equal('0');
+        expect(await this.token.methods['totalSupply(uint256)'](firstTokenId)).to.be.bignumber.equal('0');
+        expect(await this.token.methods['totalSupply(uint256)'](secondTokenId)).to.be.bignumber.equal('0');
+        expect(await this.token.methods['totalSupply()']()).to.be.bignumber.equal('0');
+      });
+    });
+  });
+
+  context('total supply of token', function () {
+    it('totalSupply unaffected by no-op', async function () {
```
frangio, line 120:
```suggestion
  context('other', function () {
    it('unaffected by no-op', async function () {
```

test/token/ERC1155/extensions/ERC1155Supply.test.js
```diff
@@ -99,9 +109,19 @@ contract('ERC1155Supply', function (accounts) {
...
+      this.token.safeTransferFrom(ZERO_ADDRESS, ZERO_ADDRESS, firstTokenId, firstTokenAmount, '0x', {
+        from: ZERO_ADDRESS,
       });
+      expect(await this.token.methods['totalSupply()']()).to.be.bignumber.equal('0');
```
frangio, line 124: Is there a reason not to check `totalSupply(firstTokenId)` as well?

JulissaDantes: Since Is checked multiple times before this test I did not think we needed it, but do you think we should include it for a no-op test?

frangio: Yes, `totalSupply(uint256)` is not checked for no-op transfers from zero to zero. I added the test.

contracts/token/ERC1155/extensions/ERC1155Supply.sol
frangio: @Amxx I think we didn't do that because we were concerned what would happen if ERC1155Supply was added in an upgrade...

frangio:
I added a missing comment in one of the `unchecked` blocks. Please refer to the item in the [guidelines](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/GUIDELINES.md) about unchecked.

frangio (approved):
LGTM

Amxx (commented):
I'm not sure I understand why we have such a CHANGELOG.md diff. Is that an error ?

JulissaDantes:
> I'm not sure I understand why we have such a CHANGELOG.md diff. Is that an error ?

The last merge of master and nextv5.0 remove all the breaking changes entries in the changeling from nextv5.0. That's why they show here, unless it was intentional and I should remove the entries too.

frangio (approved):
Thanks, this looks good now.

Amxx (approved):

pcaversaccio:
Sorry for being pedantic here, but the PR does not include any user-facing documentation wrt the overflow (except in the `CHANGELOG`). Do you plan to add something [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/next-v5.0/contracts/token/ERC1155/extensions/ERC1155Supply.sol#L28) since it is used in the OZ docs:

```solidity
/**
 * @dev Total amount of tokens.
 * @custom:security Implementing this function will restrict the total tokens minted across all `ids` to 2**256-1.
 */
function totalSupply() public view virtual returns (uint256) {
    return _totalSupplyAll;
}
```

frangio:
@pcaversaccio Thanks for raising that. The note should be at the contract level and it should be a "NOTE:" section.

frangio:
Added note in https://github.com/OpenZeppelin/openzeppelin-contracts/commit/2d05db171a99e93132abef98bda15534129f7805.

PR #3967, opened 2023-01-17
Check storage layout consistency in PRs

Amxx:
Fixes LIB-674

Adds a new step in the CI checks that will compare the storage layout between the PR and the reference branch.


frangio:
The layout comparison script looks very good. But the build info artifacts are about 100MB each, which is a lot of overhead to be downloading/uploading on every run. I believe we should instead store the layout objects as the artifacts, i.e. the result of `extractLayouts`.

Amxx:
Good point, I have not thought about it. I will split the extract/compare into 2 scripts, and we can save the intermediary representation.

Amxx:
Done. From my tests the layout in json format are ~35k

Amxx:
Does this require a changeset ?

.github/actions/storage-layout/action.yml
```diff
@@ -0,0 +1,55 @@
...
+        gh run download ${RUN_ID} --repo ${{ github.repository }} -n layout
+      env:
+        GITHUB_TOKEN: ${{ inputs.token }}
+      shell: bash
+      continue-on-error: true
+      id: reference
+    - name: Compare layouts
+      if: steps.reference.outcome == 'success' && github.event_name == 'pull_request'
+      run: |
+        node scripts/checks/layout-rebuild-and-compare.js --head $${{ inputs.layout }} --ref ${{ inputs.ref_layout }} >> $GITHUB_STEP_SUMMARY
```
frangio, line 43: Is this correct? Shouldn't it be using `compare-layouts.js`? And as far as I can tell `layout-rebuild-and-compare.js` should be removed?

Amxx: yes

frangio (approved):
Looks good!

PR #3969, opened 2023-01-18
Include EIP-5267 discovery in EIP-712

### LINKED ISSUE #3968: Implement EIP-5267 Retrieval of EIP-712 domain
[EIP-5267](https://eips.ethereum.org/EIPS/eip-5267) is in Last Call and should be Final by the time we release 4.9.

I think it should be added by default in `EIP712`, it doesn't add runtime cost and my opinion is that it should always be used as it makes EIP-712 signatures more usable.

Amxx:
Fixes #3968
Fixes LIB-633

Note: the changes to `EIP712.sol` will require a re-initialization to set the `_NAME` and `_VERSION` variables. This will be a sensitive operation, as the re initialization risks changing the domain separator.

#### PR Checklist

- [x] Tests
- [ ] Documentation
- [x] Changelog entry


Amxx:
Note that in 5.0 we should do the following breaking change:

```diff
-bytes32 private immutable _TYPE_HASH = ...
+bytes32 private constant _TYPE_HASH = ...
```

It is breaking because in the upgradeable version this is transpiled to a "live" storage slot (it shouldn't)

contracts/utils/cryptography/EIP712.sol
```diff
@@ -24,17 +25,21 @@ import "./ECDSA.sol";
...
+abstract contract EIP712 is ERC5267 {
     /* solhint-disable var-name-mixedcase */
     // Cache the domain separator as an immutable value, but also store the chain id that it corresponds to, in order to
     // invalidate the cached domain separator if the chain id changes.
     bytes32 private immutable _CACHED_DOMAIN_SEPARATOR;
     uint256 private immutable _CACHED_CHAIN_ID;
     address private immutable _CACHED_THIS;
 
+    string private _NAME;
+    string private _VERSION;
```
frangio, line 37: It's a little frustrating that we have to put this in storage when it's going to be very short strings that could be immutable. There are things we could do to have an immutable and a storage fallback. Are we open to introducing that complexity here?

Ideally we would define a library or structure that would do the optimization and fallback mechanism but I don't know if it's possible.

frangio: I'm thinking something like this (rough sketch):

```solidity
ShortString immutable _name;
ShortStringFallback _nameFallback; // struct { string value; }
...
_name = _setShortWithFallback(name, _nameFallback);
...
string memory name = _getShortWithFallback(_name, _nameFallback);
```

Amxx: I'm open to it, but it it really worth it.

- Is setting an immutable dynamically really less expensive than writing to storage once ?
- This should only be read "off-chain" by the eip712Domain accessor. There is no real "sload" to avoid.

Also, on upgradeable contract this will be storage anyway and the "short" strings will be optimized into a single slot.

frangio: >Also, on upgradeable contract this will be storage anyway and the "short" strings will be optimized into a single slot.

That's still better though! Normal strings require at least two slots.



> Is setting an immutable dynamically really less expensive than writing to storage once ?

Pretty sure it is significantly less expensive but we would need to check.



> This should only be read "off-chain" by the eip712Domain accessor. There is no real "sload" to avoid.

Yeah totally. The concern is just the SSTOREs during deployment. 2 additional string variables, i.e. 4 SSTORE, would result in 40k gas. Looking at [these benchmarks](https://github.com/alephao/solidity-benchmarks/blob/main/benchmarks/0.8.17/ERC20.md#deployment) it seems like this would be an additional 10% cost in deployment for the Permit variant.

frangio: We can see it in the gas report in this PR. This is the increase to deployment cost:
image

Amxx: > That's still better though! Normal strings require at least two slots.

My understanding is that string of length <=31 used a single slot.

Amxx: I can confirm that small strings only use one storage slot:

- see [this contract ](https://goerli.etherscan.io/address/0x18c6b374f8cba732fb36e4b03f10a2320584aea9#code)
  - the `push` function write a string into a mapping, this is the only sstore
- see [this transaction](https://goerli.etherscan.io/tx/0xfbe70e846da08e0efa3b3e7771b43a015bafd85a9f8b4505d207466625cd3df3)
  - only one slot is modified: the string fits into a single slot

Capture d’écran du 2023-01-19 11-17-39

Amxx: so its "just" 2 sstore in almost all cases

frangio: Nice, I did not know that! It's in the [Solidity docs](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#bytes-and-string) too.

So comparing with the benchmarks repo it would be "just" 5% increase. It's still the case that we see a 14% increase in our gas report here.

Amxx:
[The storage error](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/4102070852) is caused by the absence of gaps for EIP712 to grow into. This will not be an issue in  the upgradeable contracts.

We should run that checks after the transpilation

contracts/interfaces/ERC5267.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+interface ERC5267 {
```
frangio, line 5: Should be renamed to `IERC5267`.

.changeset/short-roses-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`EIP721`: add EIP-5267 support for better domain discovery.
```
frangio, line 5:
```suggestion
`EIP712`: add EIP-5267 support for better domain discovery.
```

contracts/utils/cryptography/EIP712.sol
```diff
@@ -71,16 +78,12 @@ abstract contract EIP712 {
...
         } else {
-            return _buildDomainSeparator(_TYPE_HASH, _HASHED_NAME, _HASHED_VERSION);
+            return _buildDomainSeparator();
         }
     }
 
-    function _buildDomainSeparator(
-        bytes32 typeHash,
-        bytes32 nameHash,
-        bytes32 versionHash
```
frangio, line 81: I think this function was defined with arguments initially because immutable variables couldn't be read in the constructor. Is this fixed for Solidity >=0.8.8?

The same immutable constraints where the reason for a few of the other quirks in how this contract was written.

test/governance/extensions/GovernorWithParams.test.js
```diff
@@ -22,7 +22,7 @@ contract('GovernorWithParams', function (accounts) {
   const [owner, proposer, voter1, voter2, voter3, voter4] = accounts;
 
   const name = 'OZ-Governor';
-  const version = '1';
+  // const version = '1';
```
frangio, line 25: Can this be removed?

test/helpers/eip712.js
```diff
@@ -24,13 +25,30 @@ function hexStringToBuffer(hexstr) {
...
+  }
+
+  return domain;
+}
+
+function domainType(domain) {
+  return EIP712Domain.filter(({ name }) => domain[name] !== undefined);
+}
+
+async function domainSeparator(domain) {
```
frangio, line 49: This function doesn't need to be `async`.

test/governance/utils/Votes.behavior.js
```diff
@@ -86,16 +73,13 @@ function shouldBehaveLikeVotes() {
...
-            }),
-          ),
-        );
+        const { v, r, s } = await buildData(this.votes, {
+          delegatee: delegatorAddress,
+          nonce,
+          expiry: MAX_UINT256,
+        })
+          .then(data => ethSigUtil.signTypedMessage(delegator.getPrivateKey(), { data }))
+          .then(fromRpcSig);
```
frangio, line 82: Can we agree to avoid `.then` except for one-liners? :sweat_smile: Let's please come to an agreement so it doesn't show up in all reviews.

test/governance/utils/Votes.behavior.js
```diff
@@ -6,62 +6,49 @@ const { fromRpcSig } = require('ethereumjs-util');
...
     it('domain separator', async function () {
-      expect(await this.votes.DOMAIN_SEPARATOR()).to.equal(
-        await domainSeparator({
-          name: this.name,
-          version,
-          chainId: this.chainId,
-          verifyingContract: this.votes.address,
-        }),
-      );
+      expect(await this.votes.DOMAIN_SEPARATOR()).to.equal(await getDomain(this.votes).then(domainSeparator));
```
frangio, line 24: Why use `.then` here? Do you find it more readable in this case? I want to understand.

```suggestion
      expect(await this.votes.DOMAIN_SEPARATOR()).to.equal(domainSeparator(await getDomain(this.votes)));
```

test/governance/Governor.test.js
```diff
@@ -20,7 +20,7 @@ contract('Governor', function (accounts) {
   const empty = web3.utils.toChecksumAddress(web3.utils.randomHex(20));
 
   const name = 'OZ-Governor';
-  const version = '1';
+  // const version = '1';
```
frangio, line 23: Can this be removed?

test/utils/cryptography/EIP712.test.js
```diff
@@ -21,12 +22,20 @@ contract('EIP712', function (accounts) {
...
+    it('is internally available', async function () {
+      const expected = await domainSeparator(this.domain);
 
-    expect(await this.eip712.$_domainSeparatorV4()).to.equal(expected);
+      expect(await this.eip712.$_domainSeparatorV4()).to.equal(expected);
+    });
+
+    it("can be rebuilt using EIP-5267's eip712Domain", async function () {
+      const rebuildDomain = await getDomain(this.eip712);
+      expect(mapValues(rebuildDomain, String)).to.be.deep.equal(mapValues(this.domain, String));
```
frangio, line 37: I'm not very confident in this `mapValues(.., String)` operation. I wouldn't be surprised if some part of the domain is converted into a `"[object Object]"` string or something similar.

frangio:
I have a slightly bigger issue/question I want to raise in addition to the above comments.

The current upgradeable version of EIP712 is not efficient. The domain separator cache is removed, which is necessary because the implementation contract doesn't know the address of the proxy (needed for `verifyingContract`) at the time of deployment. The bad part is that `_HASHED_NAME` and `_HASHED_VERSION` are storage variables. I believe the main reason for that was that our upgrades tooling didn't support immutable variables at the moment. The change we're introducing in this PR will require reinitializing the name and version (this should be mentioned in the changelog!). If the upgradeable contract stores those in immutable `ShortString`s, which I think it should, the user will need to pass name and version in the constructor. (This is great for us because it's unmissable, unlike a required `_init` function call which is easy to miss!). This all sounds like a good opportunity to properly make EIP712Upgradeable more efficient by keeping everything that is possible in immutable variables, fixing the current problem of `_HASHED_NAME` and `_HASHED_VERSION`.

What do you think?

contracts/utils/cryptography/EIP712.sol
```diff
@@ -34,7 +38,13 @@ abstract contract EIP712 {
...
     bytes32 private immutable _HASHED_NAME;
     bytes32 private immutable _HASHED_VERSION;
-    bytes32 private immutable _TYPE_HASH;
+    bytes32 private immutable _TYPE_HASH =
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+
+    ShortString private immutable _NAME;
+    ShortString private immutable _VERSION;
+    string private _NAME_FALLBACK;
+    string private _VERSION_FALLBACK;
```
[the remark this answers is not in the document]
Amxx: I just followed the other variables already in that contracts. Should we take this opportunity to change all that ?

contracts/utils/cryptography/EIP712.sol
Amxx: Yes it was fixed in 0.8.8, which is the same version where UDVT were introduced.

Since ShortString uses UDVT, we have to bump to 0.8.8, so we can also use the other features provided by that version

test/governance/utils/Votes.behavior.js
Amxx: this pattern (buildData → signTypedMessage → fromRpcSig) is present many times. I fealt that having it as one block, with no extra intermediary declaration would be cleaner.

I'll probably rewrite it as "buildAndSignData"

test/governance/utils/Votes.behavior.js
Amxx: Yes, I find it more readeable to have the await first then a list of chained operations.

I think it mostly comes down to the fact that using "then" makes the successive operations flow from left to right (in the direction i'm reading) ... It feels way more natural then having to keep a pile of operation in my head before getting to the first part.

test/utils/cryptography/EIP712.test.js
Amxx: would you prefer `mapValues(.., x => x.toString())` ???

contracts/utils/cryptography/EIP712.sol
```diff
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v4.8.0) (utils/cryptography/EIP712.sol)
 
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.8;
```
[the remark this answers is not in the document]
Amxx: Done

Can you please check

contracts/utils/cryptography/EIP712.sol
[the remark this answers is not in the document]
frangio: Yeah I would go with lowercase like the rest of the codebase.

test/utils/cryptography/EIP712.test.js
frangio: I checked and `String` seems to be working ok. Let's leave it.

frangio:
>will break the storage layout by removing variables (and expanding the gap) which the upgrades plugin does not currently support

We will fix this with a manual patch with placeholder variables.

---

I've renamed variables to camel case.

test/governance/utils/Votes.behavior.js
frangio: But that's a general problem with the notation `f(g(x))`. It seems to me like we just have to come to terms with the fact that operations are right to left.

frangio (dismissed):
This looks good. Once we merge we should check in the upgradeable repo to confirm we're getting the results we expected.

frangio (approved):
LGTM

songhobby:
> I have a slightly bigger issue/question I want to raise in addition to the above comments.
> 
> The current [upgradeable version of EIP712](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/62aa67491a9b46a21c2b5e3600560150a2710f98/contracts/utils/cryptography/EIP712Upgradeable.sol) is not efficient. The domain separator cache is removed, which is necessary because the implementation contract doesn't know the address of the proxy (needed for `verifyingContract`) at the time of deployment. The bad part is that `_HASHED_NAME` and `_HASHED_VERSION` are storage variables. I believe the main reason for that was that our upgrades tooling didn't support immutable variables at the moment. The change we're introducing in this PR will require reinitializing the name and version (this should be mentioned in the changelog!). If the upgradeable contract stores those in immutable `ShortString`s, which I think it should, the user will need to pass name and version in the constructor. (This is great for us because it's unmissable, unlike a required `_init` function call which is easy to miss!)
> 
> This all sounds like a good opportunity to properly make `EIP712Upgradeable` more efficient by keeping everything that is possible in immutable variables, fixing the current problem of `_HASHED_NAME` and `_HASHED_VERSION`. I think we should do this (maybe not necessarily in this PR). What do you think?
> 
> Note: The key thing we would need to do is to add annotations on all immutable variables:
> 
> ```
>     /// @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
> ```
> 
> (`state-variable-assignment` only needed if actually assigned.)

Will this be adopted in the future? (Passing `name` and `version` or other custom variables to the constructor of proxy contracts to initialize immutable state)

frangio:
> It messes up with the current proxy contract interface.

Can you elaborate on what you mean here?

songhobby:
> > It messes up with the current proxy contract interface.
> 
> Can you elaborate on what you mean here?

Correct me if I'm wrong. Proxies can not utilize the logic in the constructor of implementation contract.
Immutable variables must and can only be assigned in the constructor. To use such immutable variables in the proxy contract, they must be declared in the proxy contract too. This changes the standard proxy contract interface. 
We can abstract away the immutable variables into a separate contract (Immutable) for proxy contract to inherit from, but we still need to invoke the constructor of Immutable in the proxy's constructor.
```
  uint256 immutable private dummy;
  constructor(address _logic, bytes memory _data) payable {
    _upgradeToAndCall(_logic, _data, false);
    dummy = 0;
  }
```

frangio:
> To use such immutable variables in the proxy contract, they must be declared in the proxy contract too.

This is not correct.

A proxy can definitely use immutable variables defined in its implementation contract. However, there is a caveat, in that the values of those immutable variables will be set when the implementation is deployed (in its constructor), and will be shared by all proxies that use that implementation contract.

The reason we disallow them by default is that this behavior may be unexpected.

In the case of EIP712, with this PR the name and version will be set as immutables in the implementation contract. The proxy will then use those immutable values. The domain separator is also cached as an immutable, but in `_domainSeparatorV4` the proxy will detect that it can't use that cached value (because `address(this)` is different) and will recompute it.

If using the OpenZeppelin Upgrades Plugins, check out the [`constructorArgs` option](https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades#common-options).

songhobby:
Thanks for the correction! I agree. I think the point I was trying to make was

> To initialize immutable variables in the proxy contract, they must be declared in the proxy contract.

What I was trying to achieve is that multiple proxy contracts can use the same implementation contract but with different immutable variables initialized when deploying the proxy contracts just like those state varaibles initialized in the proxy contract's `initializer`.

It's a contradiction because immutable variables must be initialized in constructor. To initialize immutable variables when deploying proxy contracts, they must be initialized in the proxy's constructor. But proxy contract itself can not be upgradeaded. 

So It is impossible to have immutable variables initialized when deploying proxy contract when making them upgradeable in the future.

In that regard, it's really no difference in defining a function returning a constant and an immutable variable.

frangio:
> In that regard, it's really no difference in defining a function returning a constant and an immutable variable.

That's correct.



> > 
> 
> What I was trying to achieve is that multiple proxy contracts can use the same implementation contract but with different immutable variables

This will not be possible with this version of EIP712. Is this ability important to you?

PR #3970, opened 2023-01-18
Use allow-reachable delegatecall in Multicall

frangio:
Related to #3961. @ericglau points out that we should be using the new `allow-reachable` annotation because `delegatecall` is indirectly invoked through `functionDelegateCall`. Also there was an existing natspec comment we can use to put the annotation.


Amxx (approved):

JulissaDantes (approved):
LGTM!

PR #3971, opened 2023-01-18
Use oz-upgrades-unsafe-allow-reachable in ERC1967Upgrade

Amxx:
This is needed to work with the latest version of the upgrade plugin.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changelog entry


frangio:
>This is needed to work with the latest version of the upgrade plugin.

A correction, this isn't _needed_, but it's an improvement to this library because the unsafe-allow is used more granularly.

I think this PR is doing the opposite of what we want. These functions by themselves are not safe. I was thinking it should be `_upgradeToAndCallUUPS` that should have `unsafe-allow-reachable delegatecall`, but the `proxiableUUID` check is not what makes it safe. What makes it safe is the `onlyProxy` modifier in `UUPSUpgradeable`, so **the annotation should be placed in `UUPSUpgradeable.upgradeTo[AndCall]`**.

frangio:
Looks good! I would actually update the changelog, it's important to mention that this requires a recent version of Upgrades Plugins (`@openzeppelin/upgrades-core@1.21.0`).

frangio (approved):

PR #3977, opened 2023-01-19
Improve TransparentUpgradeableProxy's transparency

### LINKED ISSUE #3871: TransparentUpgradeableProxy not transparent
TUP is designed to be transparent ie. that the logic contract ABI is the only and exact ABI you see when interacting with the contract other than the proxy admin. This is unfortunately not true.

**💻 Environment**

Any environment

**📝 Details**

The functions used by the proxy admin, namely

```
function changeAdmin(address newAdmin) external virtual ifAdmin {
    _changeAdmin(newAdmin);
}

/**
 * @dev Upgrade the implementation of the proxy.
 *
 * NOTE: Only the admin can call this function. See {ProxyAdmin-upgrade}.
 */
function upgradeTo(address newImplementation) external ifAdmin {
    _upgradeToAndCall(newImplementation, bytes(""), false);
}
```

are not marked payable and thus a logic contract with these functions but payable will fail. Now these functions itself on the logic contract are not likely to be payable, but due to only 4 billion signature hashes it's not inconceivable a hash collision could happen with another signature. Or someone malicious could design a signature that hashes to one of these and make a apparently correct contract that would opaquely break when used as a logic contract with TUP.

**🔢 Code to reproduce bug**

Trivial, solution trivial by making them payable and just require msg.value ==0 manually

Amxx:
Fixes #3871

#### PR Checklist

- [x] Tests

### DIFF
--- a/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
+++ b/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
@@ -56,7 +56,7 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
      * `0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103`
      */
     function admin() external payable ifAdmin returns (address admin_) {
-        _nonPayable();
+        _requireZeroValue();
         admin_ = _getAdmin();
     }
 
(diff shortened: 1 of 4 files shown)


contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -117,4 +121,8 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
         require(msg.sender != _getAdmin(), "TransparentUpgradeableProxy: admin cannot fallback to proxy target");
         super._beforeFallback();
     }
+
+    function _nonPayable() private {
```
frangio, line 125: This issue is pretty esoteric so we should add a good comment here explaining why it's needed.

I would also consider a different name for this function. Something like `_requireZeroValue()`?

Amxx: done

frangio (approved):
LGTM!

PR #3979, opened 2023-01-19
ERC4626 inflation attack mitigation

### LINKED ISSUE #3706: Implement or recommend mitigations for ERC4626 inflation attacks
Empty ERC4626 vaults can be manipulated to inflate the price of a share and cause depositors to lose their deposits due to rounding in favor of the vault. In particular, this can be done as a frontrunning attack if the vault is used without specifying a minimum of shares to be received using an [ERC4626 router](https://github.com/fei-protocol/ERC4626#erc4626router-and-base).

We caution users against slippage in the documentation but we could offer concrete mitigations either in code or as recommendations. 
Morpho Labs [mentions](https://twitter.com/MerlinEgalite/status/1570765265272713216) they have made some changes on top of our implementation as a mitigation, we should look into that (where is the code?). See also https://github.com/transmissions11/solmate/issues/178.

Amxx:
Fixes #3706
Fixes LIB-599
Fixes LIB-653

Use virtual shares and assets to mitigate inflation attacks. The vault adds 1:N virtual asset/shares to enforce an initial exchange rate. This solves the issue of unhealty/broken vaults. It also limits the effectiveness of inflation attacks

The N value is `10**offset`, with `offset` being the decimal (precision) difference between the vault and the underlying token.

Technically, the `offset` correspond the difference (in orders of magnitude) between the funds required by the attackers and the funds being deposited by the user. 
- An offset of 0 doesn't offer protection
- An offset of 9 makes attack very difficult, but changes the decimal value of the vault.

Note that this is a breaking change: the presence of virtual shares means the vault is likely to "capture" some value, and 1 unit (wei) to underlying asset might be unrecoverable. The test file shows that on a typical lifecyle

[Analysis of the offset effect on the inflation attack](https://gist.github.com/Amxx/ec7992a21499b6587979754206a48632)

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -179,44 +179,14 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
-     * @dev Internal conversion function (from assets to shares) to apply when the vault is empty.
-     *
-     * NOTE: Make sure to keep this function consistent with {_initialConvertToAssets} when overriding it.
-     */
-    function _initialConvertToShares(
-        uint256 assets,
-        Math.Rounding /*rounding*/
-    ) internal view virtual returns (uint256 shares) {
-        return assets;
+        return assets.mulDiv(totalSupply() + 10 ** _decimalsOffset(), totalAssets() + 1, rounding);
```
Joeysantoro, line 182: does anything weird happen if an attacker mints 1 wei of assets worth of shares, then redeems 10**decimalsOffset - 1 shares to bring the exchange rate back to 1 wei : 1 wei?

Seems like there could still be some manipulation issues

Amxx: It the vault is empty and someone mints 1 wei of assets, it will get `10**offset` "wei of shares".

At the point the vault contains:
- 1 wei of asset + 1 virtual wei of asset
- `10**offset` shares unit + `10**offset` shares unit.

It the user redeems  `10**offset - 1`, we will get 0 asset for it, and the vault will contain:
- 1 wei of asset + 1 virtual wei of asset
- `1` shares unit + `10**offset` shares unit.
So the exchange rate will be `2:10**offset+1` ... very far from `1:1`

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -17,28 +17,45 @@ import "../../../utils/math/Math.sol";
...
  * https://github.com/fei-protocol/ERC4626#erc4626router-and-base[ERC4626Router].
  *
+ * Since v4.9, this implementation uses virtual assets and shares to mitigate that risk. The `_decimalsOffset()`
+ * corresponds to an offset in the decimal representation between the underlying asset's decimals and the vault
+ * decimals. This offset also determines the rate of virtual shares to virtual assets in the vault, which itself
+ * determines the initial exchange rate. While not fully preventing the attack, analysis shows that the default offset
+ * (0) makes it non-profitable, as a result of the value being captured by the virtual shares (out of the attacker's
+ * donation) matching the attacker's expected gains. With a larger offset, the attack becomes orders of magnitude more
+ * expensive than it is profitable. More details about the underlying math can be found
+ * https://docs.openzeppelin.com/contracts/4.x/erc4626#inflation-attack[here].
```
Amxx, line 35: @frangio is this the right way to do internal doc links, or is there a better way?

docs/modules/ROOT/pages/erc4626.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+| after attacker's deposit
+| $de$
+| $de$
+| $1$
+
+| after attacker's donation
+| $de+do$
+| $de$
+| $\frac{de}{de+do}$
```
Amxx, line 79: how do we add latex support ?

contracts/token/ERC20/extensions/ERC4626.sol
frangio: Better way is `xref:erc4626.adoc#inflation-attack[here]`.

docs/modules/ROOT/pages/erc4626.adoc
frangio: It's been enabled now, we should write `stem:[\frac{de}{de+do}]` and so on.

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -17,28 +17,45 @@ import "../../../utils/math/Math.sol";
  * the ERC20 standard. Any additional extensions included along it would affect the "shares" token represented by this
  * contract and not the "assets" token which is an independent contract.
  *
- * CAUTION: When the vault is empty or nearly empty, deposits are at high risk of being stolen through frontrunning with
- * a "donation" to the vault that inflates the price of a share. This is variously known as a donation or inflation
+ * CAUTION: In empty (or nearly empty) ERC-4626 vaults, deposits are at high risk of being stolen through frontrunning
```
frangio, line 20: The current way it's written, this paragraph will show up in a "Caution" box, while the next two paragraphs will show up outside of it. We can remove "CAUTION:" to remove the box completely. Alternative, we put multiple paragaphs in the box by changing the syntax to:

```
[CAUTION]
====
content here
====
```

```suggestion
 * In empty (or nearly empty) ERC-4626 vaults, deposits are at high risk of being stolen through frontrunning
```

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -17,28 +17,45 @@ import "../../../utils/math/Math.sol";
...
+ * Since v4.9, this implementation uses virtual assets and shares to mitigate that risk. The `_decimalsOffset()`
+ * corresponds to an offset in the decimal representation between the underlying asset's decimals and the vault
+ * decimals. This offset also determines the rate of virtual shares to virtual assets in the vault, which itself
+ * determines the initial exchange rate. While not fully preventing the attack, analysis shows that the default offset
+ * (0) makes it non-profitable, as a result of the value being captured by the virtual shares (out of the attacker's
+ * donation) matching the attacker's expected gains. With a larger offset, the attack becomes orders of magnitude more
+ * expensive than it is profitable. More details about the underlying math can be found
+ * https://docs.openzeppelin.com/contracts/4.x/erc4626#inflation-attack[here].
+ *
+ * The drawback of this approach is that the virtual shares do capture (a very small) part of the value being donated
```
frangio, line 37:
```suggestion
 * The drawback of this approach is that the virtual shares do capture (a very small) part of the value being accrued
```

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -17,28 +17,45 @@ import "../../../utils/math/Math.sol";
...
+ * to the vault. Also, if the vault experiences losses, the users try to exit the vault, the virtual shares and assets
+ * will cause the first user to exit to experience reduced losses in detriment to the last users that will experience
```
frangio, line 39: Are these "reduced losses" significant or also "very small"?

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -38,372 +20,378 @@ contract('ERC4626', function (accounts) {
...
+
+        /**
+         * | offset | deposited assets     | redeemable assets    |
+         * |--------|----------------------|----------------------|
+         * | 0      | 1.000000000000000000 | 0.                   |
+         * | 6      | 1.000000000000000000 | 0.999999000000000000 |
+         * | 18     | 1.000000000000000000 | 0.999999999999999999 |
+         *
+         * Attack is possible, but made difficult by the offset. For the attack to be successful
+         * the attacker needs to frontrun a deposit 10**offset times bigger then what the victim
```
frangio, line 148:
```suggestion
         * the attacker needs to frontrun a deposit 10**offset times bigger than what the victim
```

docs/modules/ROOT/pages/erc4626.adoc
```diff
@@ -0,0 +1,164 @@
...
+$$\iff 10^{off} \times du < 1 + \frac{do}{1+de}$$
+
+$$\iff 10^{off} \times du \le loss$$
+
+- If the offset is 0, the attacker loss is at least equal to the users deposit.
+- If the offset is greater than 0, the attacker will have to suffer losses that are orders of magnitude bigger than the amount of value that can hypothetically be stolen from the user.
+
+This shows that even with an offset of 0, the virtual shares and assets make this attack non profitable for the attacker. Bigger offsets increase the security even further by making any attack on the user extremely wasteful.
+
+The following figure show how the offset impacts the initial rate and limits the hability of an attacker with limited fund to inflate it effectivelly.
```
frangio, line 152:
```suggestion
The following figure show how the offset impacts the initial rate and limits the ability of an attacker with limited fund to inflate it effectivelly.
```

frangio (commented):
The Solidity code looks good to me. I have to finish going through the tests and guide.

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: Made a big caution box.

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: they should be very small

docs/modules/ROOT/pages/erc4626.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+| after attacker's donation
+| stem:[de+do]
+| stem:[de]
+| stem:[\frac{de}{de+do}]
+|===
+
+This means a deposit of stem:[du] will give stem:[\frac{du*de}{de+do}] shares.
+
+For the attacker to dilute that deposit to 0 shares, causing the user to lose all its deposit, it must ensure that
```
frangio, line 85: A question that came up for me here was: diluting to 0 shares is necessary to cause a loss of 100%, but the attacker could also for example dilute to 1 share to cause a loss of 50% of the deposit. How does this affect the reasoning?

docs/modules/ROOT/pages/erc4626.adoc
```diff
@@ -0,0 +1,184 @@
...
+
+Figure 6 shows how an attacker can manipulate the rate of an empty vault. First the attacker must deposit a small amount of tokens (1 token) and follow up with a donation of 1e5 tokens directly to the vault to move the exchange rate "right". This puts the vault in a state where any deposit smaller than 1e5 would be completely lost to the vault. Given that the attacker is the only share holder (from their donation), the attacker would steal all the tokens deposited.
+
+An attacker would typically wait for a user to do the first deposit into the vault, and would frontrun that operation with the attack described above. The risk is low, and the size of the "donation" required to manipulate the vault is equivalent to the size of the deposit that is being attacked.
+
+In math that gives:
+
+- stem:[de] the attacker deposit
+- stem:[do] the attacker donation
+- stem:[du] the user deposit
```
frangio, line 58: An issue with this notation for me is that it looks like differentials. I feel like single-letter variables would look better.

The other thing I have an issue with is $10^{off}$. The 3 letters have spacing between them as if they were separate variables. $10^\mathit{off}$ (`10^\mathit{off}`) looks better but is more cumbersome to write.

frangio (commented):
Minor comments but all looks good to me.

docs/modules/ROOT/pages/erc4626.adoc
Amxx: > An issue with this notation for me is that it looks like differentials. I feel like single-letter variables would look better.

both deposit and donation start with a $d$, so I was not sure which letter to use. If you have ideas for renaming please share them.

frangio: We could just use $a_0$ attacker deposit, $a_1$ attacker donation, $u$ user deposit.

docs/modules/ROOT/pages/erc4626.adoc
frangio: >Should I replace the formula in the docs so that $n$ appears ?

I would just add a comment saying that the argument also extends to partial losses of the user's deposit.

frangio (approved):
LGTM

boringcrypto:
You're welcome 😉

frangio:
@boringcrypto Oh, not sure if this was your point but we should probably credit YieldBox in our docs. It took us a lot of time to be convinced but that was definitely how we got to this solution (https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3706#issuecomment-1297230505). Do you know if you were the first to propose this approach?

test/token/ERC20/extensions/ERC4626.t.sol
```diff
@@ -15,22 +15,4 @@ contract ERC4626StdTest is ERC4626Test {
         _vaultMayBeEmpty = false;
```
pcaversaccio, line 15: shouldn't be `_vaultMayBeEmpty = true;` now? See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3792#issuecomment-1302682873

Amxx: Thanks for the heads up, fixed in #4068

pcaversaccio: awesome

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -38,372 +20,378 @@ contract('ERC4626', function (accounts) {
...
+          expect(await this.vault.totalAssets()).to.be.bignumber.equal('0');
+        });
+
+        it('deposit', async function () {
+          expect(await this.vault.maxDeposit(holder)).to.be.bignumber.equal(constants.MAX_UINT256);
+          expect(await this.vault.previewDeposit(parseToken(1))).to.be.bignumber.equal(parseShare(1));
+
+          const { tx } = await this.vault.deposit(parseToken(1), recipient, { from: holder });
+
+          await expectEvent.inTransaction(tx, this.token, 'Transfer', {
```
pcaversaccio, line 58: Maybe I missed the somehow, but the tests do not include any event emission tests for `Deposit` and `Withdraw`. I would say those are the most important ones to test in this context.

Amxx: 👍🏻

Amxx: Addressed here: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4072

(you should work with us 😄 )

pcaversaccio: highly appreciate your comment and am happy it improves the overall code quality - I can't commit anything full-time currently and will keep monitoring the OZ contracts as a side project for now (but never say never 😄)

PR #3981, opened 2023-01-19
Add `RELEASING.md` docs

ernestognw:
## Description

Added docs about #3915 

#### PR Checklist

- [x] Documentation


frangio:
I've simplified the RELEASING.md to remove implementation details about the release process. For implementation details we should make sure that the release-cycle.yml scripts are well documented.

Personally I find this flowchart to be the main useful thing, it would be nice to include it in a comment in release-cycle.yml but we'd need to draw it in ASCII.

```mermaid
graph LR
  Development -->|D| RC-Unreleased -->|M| RC-Released -->|D| Final-Unreleased -->|M| Final-Released
  RC-Released -->|C| RC-Unreleased
  Final-Released -->|C| Final-Unreleased
```

ernestognw:
> I've simplified the RELEASING.md to remove implementation details about the release process. For implementation details we should make sure that the release-cycle.yml scripts are well documented.
> 
> Personally I find this flowchart to be the main useful thing, it would be nice to include it in a comment in release-cycle.yml but we'd need to draw it in ASCII.

Well, I was hoping it was added if I put it in the comments. I tried.

Anyway, I was wondering what is the problem with putting it in the comments and having a link to mermaid.live? I think it's still better developer experience.

frangio:
>  I was wondering what is the problem with putting it in the comments and having a link to mermaid.live?

It's not a bad idea. ASCII art diagram might be simpler though.

ernestognw:
> > I was wondering what is the problem with putting it in the comments and having a link to mermaid.live?
> 
> It's not a bad idea. ASCII art diagram might be simpler though.

Let's keep it in ASCII

RELEASING.md
```diff
@@ -1,151 +1,47 @@
...
-
 > Visit the documentation for [details about release schedule](https://docs.openzeppelin.com/contracts/releases-stability).
 
-## Changesets
-
-[Changesets](https://github.com/changesets/changesets/) is used as part as our release process for `CHANGELOG.md` management, so each change that's relevant for the codebase is expected to include a changeset entry, so it's eventually added to a release branch.
+OpenZeppelin Contracts uses a fully automated release process that takes care of compiling, packaging, and publishing the library, all of which is carried out in a clean CI environment (GitHub Actions), implemened in the ([`release-cycle`](.github/workflows/release-cycle.yml)) workflow. This helps to reduce the potential for human error and inconsistencies, and ensures that the release process is ongoing and reliable.
```
Amxx:
```suggestion
OpenZeppelin Contracts uses a fully automated release process that takes care of compiling, packaging, and publishing the library, all of which is carried out in a clean CI environment (GitHub Actions), implemented in the ([`release-cycle`](.github/workflows/release-cycle.yml)) workflow. This helps to reduce the potential for human error and inconsistencies, and ensures that the release process is ongoing and reliable.
```

frangio (approved):

PR #3983, opened 2023-01-20
Add a public Governor.cancel function

### LINKED ISSUE #3436: Add external `cancel` to OZ Governor by default
**🧐 Motivation**
It would be great to allow the OZ Governor to have the `cancel` proposal functionality offered by default. It should be something similar like the `cancel` offered in Governor Bravo.

**📝 Details**
Even though currently an internal `_cancel` is offered in the OZ Governor, this cannot be used as one would probably need to store the `proposer` on chain, which is not part of the Governor implementation.
```
  function _cancel(
        address[] memory targets,
        uint256[] memory values,
        bytes[] memory calldatas,
        bytes32 descriptionHash
    ) internal override(Governor, GovernorTimelockControl) returns (uint256) {
        return super._cancel(targets, values, calldatas, descriptionHash);
    }
```
It would be great if the OZ governor provided already with the `proposer` on-chain and the ability to have an external `cancel()` that we can call:

```
  function cancel(
        address[] memory targets,
        uint256[] memory values,
        bytes[] memory calldatas,
        bytes32 descriptionHash
    ) external;
   
```

Default implementation could be based on the Governor Bravo cancel, by default could be something on these lines:

```
   function cancel( address[] memory targets,
        uint256[] memory values,
        bytes[] memory calldatas,
        bytes32 descriptionHash) public virtual {
        
        // 1. - Get proposalID
        // 2. - Get proposer based on proposalID 
       
        // 3 - add checks for proposer
        require(
            _msgSender() == proposer || getVotes(proposer, block.number - 1) < proposalThreshold(),
            "Governor: proposer above threshold"
        );
       // 4 -  Cancel
        _cancel(
            targets,
            values,
            calldatas,
            descriptionHash
        );
    }
```
Also this function signature would have to be added to `Tally` for compatibility. Currently Tally does not allow to add external `cancel` to the OZ Governor as this is not part of the common interface. (As shown here: https://docs.tally.xyz/user-guides-1/supported-dao-frameworks/openzeppelin-style)

Amxx:
Fixes #3436 

Note: in next breaking (5.0) we should remove the proposer `GovernorCompatibilityBravo.ProposalDetails` as its is now in `Govenor.Proposal`

#### PR Checklist

- [x] Tests
- [ ] Documentation


test/helpers/governance.js
```diff
@@ -62,10 +62,10 @@ class GovernorHelper {
         );
   }
 
-  cancel(opts = null) {
+  cancel(usePublic = false, opts = null) {
```
frangio, line 65: Using the internal function instead of the external function can lead to a very wrong test. So I would default to `usePublic = true` so that the test fails instead of silently passing. To improve readability I would also change this from a boolean to a string `'internal' | 'external'` (or `public`).

```suggestion
  cancel(visibility = 'external', opts = null) {
```

frangio (commented):
This looks great! I only have a comment about the test helper.

test/helpers/governance.js
Amxx: done

frangio:
Looking at the gas benchmarks it seems there is a +30% increase in `propose`. I believe this is because even though the `address proposer` field is packed and does not result in a new storage slot in the layout, it is packed with `bool executed` and `bool cancelled`, both of which are false initially.

At the same time, `execute` (and `cancel`) are now 30% cheaper because the slot is already non-zero.

I think this is a reasonable tradeoff. Note that in 5.0 with better packing the cost of `propose` will decrease to current levels.

frangio (approved):
This is good with me if you agree with the tradeoff I explained above.

Amxx:
Agree with your analysis and with this tradeoff.

Two things for future PRs:
- in 5.0 we want to merge that!
- do we want to had an accessor for the proposer ?

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -99,20 +99,15 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
         );
     }
 
-    function cancel(uint256 proposalId) public virtual override {
+    function cancel(uint256 proposalId) public virtual override(IGovernor, Governor) {
```
frangio, line 102: Needs docs explaining the logic, also that it overrides and doesn't call `super.cancel` from the core Governor.

contracts/governance/IGovernor.sol
```diff
@@ -216,6 +216,14 @@ abstract contract IGovernor is IERC165 {
         bytes32 descriptionHash
     ) public payable virtual returns (uint256 proposalId);
 
+    /**
+     * @dev Cancel a proposal. This is restricted to Pending proposal (before the vote starts) and is restricted to
+     * the proposal's proposer.
```
frangio, line 221: This isn't really accurate here. The specific logic depends on the modules, for example it's different between Governor and GovernorCompatibilityBravo, so it should be documented in `Governor`.

Issue #3923, opened 2023-01-03
Make upgradeTo functions public


frangio:
The `upgradeTo` and `upgradeToAndCall` functions are `external virtual`. As a result even though they can be overriden they can't be invoked via `super`, which is the correct way to override a function in this library.

From https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3870.

PepperAndSalt999:
This one is for me :B

PR #3984, opened 2023-01-21
Make upgradeTo functions public #3923

Ankitvnsbihar:
Title: Make function upgradeTo public

Description:
This PR makes the function upgradeTo(address newImplementation) public by adding the public keyword before the function keyword. This allows external contracts or users to call the function and upgrade the implementation of the contract.

Changes:

The public keyword has been added before the function keyword in the function definition of upgradeTo(address newImplementation).

Testing:
I have tested the function by calling it from an external contract and it worked as expected.

Context:
This change is needed to allow external contracts or users to upgrade the implementation of the contract.

Breaking Changes:
None, this change only makes the function public and should not affect the existing functionality of the contract.

### DIFF
--- a/contracts/mocks/proxy/UUPSLegacy.sol
+++ b/contracts/mocks/proxy/UUPSLegacy.sol
@@ -44,11 +44,11 @@ contract UUPSUpgradeableLegacyMock is UUPSUpgradeableMock {
     }
 
     // hooking into the old mechanism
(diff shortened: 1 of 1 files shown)


frangio:
Duplicate of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3959.

PR #3988, opened 2023-01-23
fix: lint should not override files

vittominacori:
The `npm run lint` task use `--write` so it will override files instead of check only. Should not?

This PR only change prettier behaviour for lint using `--check`.


Amxx (approved):
Thank you @vittominacori for catching this one!

PR #3989, opened 2023-01-23
Improve the propose overrides in `GovernorCompatibilityBravo`

### LINKED ISSUE #3881: Redundant storeProposal in GovernorCompatibilityBravo
In the Bravo Compatibility Contract the proposal data is duplicated with different proposalIds
The initial Bravo Contract does not have a proposal method which takes 4 parameters,  why do we require storage for that proposalHash?

`    // ============================================== Proposal lifecycle ==============================================
    /**
     * @dev See {IGovernor-propose}.
     */
    function propose(
        address[] memory targets,
        uint256[] memory values,
        bytes[] memory calldatas,
        string memory description
    ) public virtual override(IGovernor, Governor) returns (uint256) {
        _storeProposal(_msgSender(), targets, values, new string[](calldatas.length), calldatas, description);
        return super.propose(targets, values, calldatas, description);
    }

    /**
     * @dev See {IGovernorCompatibilityBravo-propose}.
     */
    function propose(
        address[] memory targets,
        uint256[] memory values,
        string[] memory signatures,
        bytes[] memory calldatas,
        string memory description
        _storeProposal(_msgSender(), targets, values, signatures, calldatas, description);
        return propose(targets, values, _encodeCalldata(signatures, calldatas), description);
    }
`

Amxx:
Fixes #3881

#### PR Checklist

- [x] Tests

### DIFF
--- a/contracts/governance/compatibility/GovernorCompatibilityBravo.sol
+++ b/contracts/governance/compatibility/GovernorCompatibilityBravo.sol
@@ -70,7 +70,7 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
         string memory description
     ) public virtual override returns (uint256) {
         _storeProposal(_msgSender(), targets, values, signatures, calldatas, description);
-        return propose(targets, values, _encodeCalldata(signatures, calldatas), description);
+        return super.propose(targets, values, _encodeCalldata(signatures, calldatas), description);
     }
 
     /**
--- a/.changeset/moody-foxes-shave.md
+++ b/.changeset/moody-foxes-shave.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`GovernorCompatibilityBravo`: Improve the propose-with-signature workflow.


Amxx:
Put on hold: [see this discussion](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3881#issuecomment-1400474803)

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -70,7 +70,7 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
         string memory description
     ) public virtual override returns (uint256) {
         _storeProposal(_msgSender(), targets, values, signatures, calldatas, description);
-        return propose(targets, values, _encodeCalldata(signatures, calldatas), description);
+        return super.propose(targets, values, _encodeCalldata(signatures, calldatas), description);
```
frangio, line 73: Using `super.propose` here has the problem that we were discussing the other day, and this change is only improving gas by about 1%. So I'm tempted to say we should not do this change, since this contract is already inheritance heavy we need to be really sure there won't be any problems caused by it.

Why do we even need to store the signatures? I can't find a single place where we actually use them. I assumed they would be returned from the `proposals` getter but they aren't. It seems we could just be storing the encoded calldata?

frangio: If we don't need the signatures we can just remove `_storeProposal` in this function and call `propose` without super.

Amxx: The `getActions` function returns the signatures. We could just store the encoded data, but then if someone calls propose (with signature) and then calls `getActions` he would not get the signature that are in the proposal.

I think we discussed that when making this contract, and we wanted the signature to be available for frontends (Tally?) to display.

frangio: Ok thanks I don't know why I didn't see `getActions`. In that case, there is something else we can do:

```solidity
        uint256 proposalId = propose(targets, values, _encodeCalldata(signatures, calldatas), description);
        _proposalSignatures[proposalId] = signatures;
```

We can store signatures separately and in `getActions` we read from two different mappings.

Is there an issue with this?

Amxx: If we do that, the `calldata` stored (and returned by `getActions`) would include the function selector. This is not the case of GovernorBravo (and of the current implementation)

frangio: We could work around that by iterating over signatures, if length is non-zero, we truncate the function selector from calldata. This sounds a little complex but IMO it's fine.

Amxx: Not sure its better then what we have right now

Amxx: (right now in master, not this PR)

frangio: It's a tradeoff between the cost of `propose` and the cost of `getActions`. It makes sense to prioritize `propose`.

To be fair though, this PR only shows a 1% improvement to `propose` with signatures.

We should close this PR and consider if we want to implement the alternative I proposed, either now or in 5.0 (or never).

PR #3990, opened 2023-01-23
Use SafeMath in ERC20 implementation due readability

k06a:
Why not to continue to use SafeMath with extra argument? I would advocate for higher readability of this code.

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx:
Hello @k06a

There are two reasons:

1. [The safe math with custom error allocates the string for every calls, including the ones that don't revert](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeMath.sol#L159-L160). This would make all the non reverting calls more expensive. We don't want that.
2. We are soon going to use custom errors instead of revert strings anyway.


We could use `Math.trySub`

PR #3993, opened 2023-01-24
Improve ERC1363 documentation

vittominacori:
Noticed that the interface id to be returned in IERC1363 comment has a wrong value.

It splits the interface ids in [2 different values](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/interfaces/IERC1363.sol#L10) instead of an unique one as defined in the final [EIP](https://eips.ethereum.org/EIPS/eip-1363#specification) specification.

Then I also added some comments and changed the `value` param to `amount` to be ERC20 like.

I think that all the pending PRs (#3017, #3525) to add ERC1363 implementation are now obsolete so if you need some help to add ERC1363 implementation feel free to ask and I will try adding it opening a fresh new PR following the new OZ guidelines.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx (commented):
The IERC1363 interfaceId is indeed documented incorrectly.

For the value vs amount issue, I'm not sure if we want to deviate from the EIP document. Could it possibly break library that extracts names of arguments?

vittominacori:
> For the value vs amount issue, I'm not sure if we want to deviate from the EIP document. Could it possibly break library that extracts names of arguments?

Also the IERC20 interface differs from the official EIP using `to` and `amount` instead of `_to` and `_value`. 
I would like to keep the same signature of ERC20. 
Let me know if you want to revert back to official EIP1363 signature.

vittominacori:
Any update?

frangio (approved):
Thank you @vittominacori.

Re: amount vs value. There is currently an inconsistency around the use of these terms that we will need to tackle (see https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3864), but I agree with the reasoning for the change in this PR.

PR #4000, opened 2023-01-25
Rename `ERC20ReturnFalseMock copy.sol` to `ERC20ReturnFalseMock.sol`

pcaversaccio:
As title. I couldn't figure out any specific reason why this naming (i.e. `copy`) was chosen so let's see whether the CI fails somewhere :). Since it's a simple refactor I didn't include a `CHANGELOG` entry.


Amxx:
Thank you @pcaversaccio 

I think this was fixed in the `next-v5.0` branch but not in master.

Amxx (approved):

PR #4001, opened 2023-01-25
Create ProposalSalt event for TimelockController

### LINKED ISSUE #3216: TimelockController 'salt' is unrecoverable yet required for execution
When scheduling a transaction using the `TimelockController`, one is required to provide a `salt` to distinguish otherwise identical transactions. The `salt` parameter is not emitted in the `CallScheduled` event which makes it impossible to recover. Yet `salt` is required for executing the scheduled transaction transaction, ~~even though it isn't necessary and more gas efficient if `id` is used instead~~.

**💻 Environment**

Openzeppelin v4.5.0  / hardhat

**📝 Details**

`salt` is missing from the `CallScheduled` event: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/458697be32f41946e1ab66e946c7cf373c79577d/contracts/governance/TimelockController.sol#L35

`salt` is required for the `execute` & `executeBatch` calls: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/458697be32f41946e1ab66e946c7cf373c79577d/contracts/governance/TimelockController.sol#L267

~~Instead of `salt` the `execute` methods should be consuming the hashed operation (`id`) directly, since that id is readily available via event data. This would also save on gas since there is no need to re-run the hashing operation.~~

Happy to submit a PR

JulissaDantes:
Fixes #3216  
Fixes LIB-639

When scheduling a transaction using the TimelockController, one is required to provide a salt to distinguish otherwise identical transactions. The salt parameter is not emitted in the CallScheduled event which makes it impossible to recover. This PR introduces the `ProposalSalt` event to get the salt after scheduling an operation, if salt == 0 it will not emit an event.

#### PR Checklist

- [X] Tests
- [X] Documentation


contracts/governance/TimelockController.sol
```diff
@@ -249,6 +257,9 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
         _schedule(id, delay);
         for (uint256 i = 0; i < targets.length; ++i) {
             emit CallScheduled(id, i, targets[i], values[i], payloads[i], predecessor, delay);
+            if (salt != bytes32(0)) {
```
Amxx, line 260: This should not be inside the loop.

contracts/governance/TimelockController.sol
```diff
@@ -51,6 +51,11 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
      */
     event CallExecuted(bytes32 indexed id, uint256 indexed index, address target, uint256 value, bytes data);
 
+    /**
+     * @dev Emitted when new proposal is scheduled.
+     */
+    event ProposalSalt(bytes32 indexed proposalId, bytes32 salt);
```
Amxx, line 57: I think we should use `id` for naming consistency accross events

.changeset/five-poets-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`TimelockController`: Add the `ProposalSalt` event to emit on operation schedule.([#4001](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4001))
```
frangio, line 5:
```suggestion
`TimelockController`: Add the `ProposalSalt` event to emit on operation schedule.
```

contracts/governance/TimelockController.sol
```diff
@@ -223,6 +228,9 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
         bytes32 id = hashOperation(target, value, data, predecessor, salt);
         _schedule(id, delay);
         emit CallScheduled(id, 0, target, value, data, predecessor, delay);
+        if (salt != bytes32(0)) {
+            emit ProposalSalt(id, salt);
```
frangio, line 232: The functoin documentation says that `CallScheduled` is emitted. We need to add `ProposalSalt` in there as well. Same in `scheduleBatch`.

contracts/governance/TimelockController.sol
```diff
@@ -51,6 +51,11 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
      */
     event CallExecuted(bytes32 indexed id, uint256 indexed index, address target, uint256 value, bytes data);
 
+    /**
+     * @dev Emitted when new proposal is scheduled.
+     */
+    event ProposalSalt(bytes32 indexed id, bytes32 salt);
```
frangio, line 57: The issue suggested `ProposalSalt` but the rest of the events are `CallExecuted`, `CallScheduled`, etc.

This should probably be called `CallSalt`. @Amxx?

contracts/governance/TimelockController.sol
```diff
@@ -51,6 +51,11 @@ contract TimelockController is AccessControl, IERC721Receiver, IERC1155Receiver
      */
     event CallExecuted(bytes32 indexed id, uint256 indexed index, address target, uint256 value, bytes data);
 
+    /**
+     * @dev Emitted when new proposal is scheduled.
```
frangio, line 55: We need to expain it's only emitted for non-zero salt.

frangio (changes requested):

contracts/governance/TimelockController.sol
Amxx: Yes, `CallSalt` would make more sens.

frangio (approved):
Thanks!

PR #4002, opened 2023-01-25
Add guidelines for pull requests

frangio:
These are my suggestions, they can be discussed.


Amxx (approved):

ernestognw:
Late to the party, but, should we add a guideline similar for the changelog entries? Or is it implied?

See [this changeset](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0c89a8b771094828fbeeb84f17b475fe4a2d56f1/.changeset/funny-rockets-compete.md)

Issue #3996, opened 2023-01-25
ERC721 voting with dynamic voting power


RitzyDevBox:
**🧐 Motivation**
I'm working on an NFT based DAO, currently I don't see a way to override the voting power for an individual NFT.  

**📝 Details**
Expose an Interface on in the ERC721Votes.sol

      setVotingPower(uint256 tokenId) {
      }

Update the GetVotes and GetPastVotes to support dynamic voting power 

    function getVotes(address account) public view virtual override returns (uint256) {
        
        uint256[] memory ownedNFTs =_nftOwnerCheckpoints[account];
         //Pseudo code
        return ownedNFTs.sum(n =>_delegateNFTCheckpoints[n].latest())

    }

Amxx:
Hello @RitzyDevBox 

> function getVotes(address account) public view virtual override returns (uint256) {
>     
>     uint256[] memory ownedNFTs =_nftOwnerCheckpoints[account];
>      //Pseudo code
>     return ownedNFTs.sum(n =>_delegateNFTCheckpoints[n].latest())
> 
> }

The `ownedNFTs` is an unbounded array, that might be arbitrarily expensive to fetch
The `sum` part is also arbitrarily expensive to read/compute.

While this approach is technically possible, it would result in very expensive voting, and could cause DOS issues when the cost of voting becomes greater then the size of a block.

frangio:
The right way to do this would be to use the `params` system. The user would call `voteWithParams` where the parameters are their owned token IDs they want to vote with. The array of token ids is iterated to check the ownership, compute the voting power, and validate that it hasn't been used to vote for the particular proposal.

This works because the params array doesn't need to include all owned token IDs, and a token owner can vote multiple times with different batches of tokens if the array is so large that voting would run out of gas.

This Governor extension is also valuable because it allows voting with a standard non vote-enabled ERC721 token.

frangio:
I see. Changing that breaks compatibility. In my opinion `hasVoted` should be left as is. It's not really used to determine if an account can vote again so it's merely informational.

frangio:
> The weight resolution is wrong, since params requires to option of partial votes per NFT

Can you explain this further? I didn't understand.

> Also this will run into issues for: `_getVotingUnits`

`_getVotingUnits` is used for vote-tracking in the token itself. The suggestion here is to make a Governor module, not an ERC721 extension.

Feel free to share links to the code you're talking about.

[...]

PR #4003, opened 2023-01-25
Creating new ERC721Votes for weighted voting

### LINKED ISSUE #3997: ERC721Votes._afterTokenTransfer assumes 1 token = 1 vote
I found this while unit testing my contract.  I wanted to providing more voting power per nft, mostly because I've written a governor which handles partial voting.

I overriden the `_getVotingUnits` like so:

    function _getVotingUnits(address account) internal view virtual override returns (uint256) {
        return balanceOf(account) * 10000;
    }

When minting and delegating I'm getting the wrong result:

    await nft.mint(owner.address, { value: ethers.utils.parseUnits("1", 18)})
    await nft.delegate(owner.address);

    ownerVotes = await nft.getVotes(owner.address);
    console.log(`owner votes: ${ownerVotes.toString()}`);

    await nft.mint(owner.address, { value: ethers.utils.parseUnits("1", 18)})
    await nft.delegate(owner.address);

    ownerVotes = await nft.getVotes(owner.address);
    console.log(`owner votes: ${ownerVotes.toString()}`);

this outputs 

     10000
     10001

It increments sequentially after.  The issue is that `_transferVotingUnits(from, to, batchSize )` does not take the units into account.

    
 2. `_afterTokenTransfer` does not apply the Units to the batch size

`abstract contract ERC721Votes is ERC721, Votes {
     internal uint256 votingPowerBase = 1

    _setVotingPowerUnits(uint256 _votingPowerBase) internal virtual {
        votingPowerBase = _votingPowerBase
    }
    
    getVotingPowerUnits(uint256 _votingPowerBase) public virtual returns (uint 256) {
        return votingPower;
    }

    /**
     * @dev See {ERC721-_afterTokenTransfer}. Adjusts votes when tokens are transferred.
     *
     * Emits a {IVotes-DelegateVotesChanged} event.
     */
    function _afterTokenTransfer(
        address from,
        address to,
        uint256 firstTokenId,
        uint256 batchSize
    ) internal virtual override {
        _transferVotingUnits(from, to, batchSize * VotingPowerBase );
        super._afterTokenTransfer(from, to, firstTokenId, batchSize);
    }

    /**
     * @dev Returns the balance of `account`.
     */
    function _getVotingUnits(address account) internal view virtual override returns (uint256) {
        return balanceOf(account) * VotingPowerBase ;
    }`

I'm not sure how to increment the units otherwise, seeing as even if I override this directly in my contract it will still end up calling the base.

RitzyDevBox:
Fixes #3997 

Fixes issues with modifying vote weights for NFT's

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC721/extensions/ERC721WeightedVotes.sol
+++ b/contracts/token/ERC721/extensions/ERC721WeightedVotes.sol
@@ -19,6 +19,7 @@ import "./IERC721Supply.sol";
  */
 abstract contract ERC721WeightedVotes is ERC721, Votes, IERC721Supply {
 
+    function totalSupply() public view virtual override returns (uint256);
     function _votesPerToken() internal pure virtual returns (uint256) { return 1; }
     /**
      * @dev See {ERC721-_afterTokenTransfer}. Adjusts votes when tokens are transferred.
@@ -31,17 +32,22 @@ abstract contract ERC721WeightedVotes is ERC721, Votes, IERC721Supply {
         uint256 firstTokenId,
         uint256 batchSize
     ) internal virtual override {
-        _transferVotingUnits(from, to, batchSize * _votesPerToken());
+        _transferVotingUnits(from, to, _getVotesForToken(firstTokenId, batchSize));
         super._afterTokenTransfer(from, to, firstTokenId, batchSize);
     }
-
-    function totalSupply() public view virtual override returns (uint256);
 
     /**
      * @dev Returns the balance of `account`.
      */
     function _getVotingUnits(address account) internal view virtual override returns (uint256) {
         return balanceOf(account) * _votesPerToken();
+    }
+
+    /**
+     * @dev Returns the vote amounts sequential token batches enumerating from the first token
+     */
+    function _getVotesForToken(uint256 firstTokenId, uint256 batchSize) internal view virtual returns (uint256) {
+       return batchSize * _votesPerToken();
     }
 
     function _mint(address to, uint256 tokenId) internal override virtual {
--- a/contracts/token/ERC721/extensions/ERC721WeightedVotes.sol
+++ b/contracts/token/ERC721/extensions/ERC721WeightedVotes.sol
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.8.0) (token/ERC721/extensions/ERC721Votes.sol)
+
+pragma solidity ^0.8.0;
+
+import "../ERC721.sol";
+import "../../../governance/utils/Votes.sol";
+import "./IERC721Supply.sol";
+
+/**
+ * @dev Extension of ERC721 to support voting and delegation as implemented by {Votes}, where each individual NFT counts
+ * as 1 vote unit.
+ *
+ * Tokens do not count as votes until they are delegated, because votes must be tracked which incurs an additional cost
+ * on every transfer. Token holders can either delegate to a trusted representative who will decide how to make use of
+ * the votes in governance decisions, or they can delegate to themselves to be their own representative.
+ *
+ * _Available since v4.5._
+ */
+abstract contract ERC721WeightedVotes is ERC721, Votes, IERC721Supply {
+
+    function _votesPerToken() internal pure virtual returns (uint256) { return 1; }
+    /**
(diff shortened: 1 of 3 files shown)


contracts/token/ERC721/extensions/ERC721WeightedVotes.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @dev Returns the vote amounts sequential token batches enumerating from the first token
+     */
+    function _getVotesForToken(uint256 firstTokenId, uint256 batchSize) internal view virtual returns (uint256) {
+       return batchSize * _votesPerToken();
+    }
+
+    function _mint(address to, uint256 tokenId) internal override virtual {
+        require((totalSupply() + 1) * _votesPerToken() <= 2**256 -1, "ERC721Votes: voting power exceeds max uint256");
```
Amxx, line 54: This check will not work as expected. 

- `2**256` can't be represented, so `2**256-1` cannot be computed. If you want to represent that number, you should use `type(uint256).max`
- `x <= type(uint256).max` will always be true, because by definition x is a at most `uint256`, and so it cannot contain anything that is bigger than the max uint256
- if `(totalSupply() + 1) * _votesPerToken()` overflows, then that will trigger a mathematical error (remember we are in solidity 8.0)

contracts/token/ERC721/extensions/IERC721Enumerable.sol
```diff
@@ -3,18 +3,13 @@
...
  * @title ERC-721 Non-Fungible Token Standard, optional enumeration extension
  * @dev See https://eips.ethereum.org/EIPS/eip-721
  */
-interface IERC721Enumerable is IERC721 {
-    /**
-     * @dev Returns the total amount of tokens stored by the contract.
-     */
-    function totalSupply() external view returns (uint256);
-
+interface IERC721Enumerable is IERC721Supply {
```
Amxx, line 12: This is going to change the definition of `IERC721Enumerable` and thus the value returned by `type(IERC721Enumerable).interfaceId`. 

So this change is breaking the ERC165 support that is implemented in ERC721Enumerable.
```solidity
    function supportsInterface(bytes4 interfaceId) public view virtual override(IERC165, ERC721) returns (bool) {
        return interfaceId == type(IERC721Enumerable).interfaceId || super.supportsInterface(interfaceId);
    }
```

contracts/token/ERC721/extensions/ERC721WeightedVotes.sol
```diff
@@ -0,0 +1,57 @@
...
+ * @dev Extension of ERC721 to support voting and delegation as implemented by {Votes}, where each individual NFT counts
+ * as 1 vote unit.
+ *
+ * Tokens do not count as votes until they are delegated, because votes must be tracked which incurs an additional cost
+ * on every transfer. Token holders can either delegate to a trusted representative who will decide how to make use of
+ * the votes in governance decisions, or they can delegate to themselves to be their own representative.
+ *
+ * _Available since v4.5._
+ */
+abstract contract ERC721WeightedVotes is ERC721, Votes, IERC721Supply {
```
Amxx, line 20: Requiring `IERC721Supply` means that this module either needs a `ERC721Enumerability`, or needs a lighter custom module (that we don't provide).

We are very strongly pushing toward NOT using `ERC721Enumerability`, so I don't think its right to require using it in modules like this.

Amxx (changes requested):
Considering all the issues that this change cause (interface issues + need for a non standard `totalSupply()` function) I'm starting to think its a bad idea to have this `_votesPerToken()` function in the ERC721 token.

If its just a matter of scaling the votes by an fix value, I think it would be much better to do that in the governor:

```solidity
    function _getVotes(
        address account,
        uint256 blockNumber,
        bytes memory /*params*/
    ) internal view virtual override returns (uint256) {
        return token.getPastVotes(account, blockNumber) * _votesPerToken();
    }
```

contracts/token/ERC721/extensions/ERC721WeightedVotes.sol
Amxx: > if it throws in the require statement it should return that error message no?

no

frangio:
Closing due to the issues described above and in favor of the alternative approaches described. Also see https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3996.

PR #4008, opened 2023-01-27
Add `string` and `bytes` support to the `StorageSlots` library

Amxx:
String manipulation usefull for an upcoming ShortString library. 

See #3969

#### PR Checklist
- [x] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/rare-kids-punch.md
+++ b/.changeset/rare-kids-punch.md
@@ -1,5 +1,0 @@
----
-'openzeppelin-solidity': minor
----
-
-`StorageSlot`: library is now generated from a template.
--- a/.changeset/rare-kids-punch.md
+++ b/.changeset/rare-kids-punch.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`StorageSlot`: library is now generated from a template.
--- a/contracts/utils/StorageSlot.sol
+++ b/contracts/utils/StorageSlot.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v4.7.0) (utils/StorageSlot.sol)
+// This file was procedurally generated from scripts/generate/templates/StorageSlot.js.
 
 pragma solidity ^0.8.0;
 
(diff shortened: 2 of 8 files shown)


.changeset/rare-kids-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`StorageSlot`: library is now generated from a template.
```
frangio, line 5: I think we can consider this an implementation detail and remove the changeset.

contracts/utils/StorageSlot.sol
```diff
@@ -85,4 +95,44 @@ library StorageSlot {
...
+        /// @solidity memory-safe-assembly
+        assembly {
+            r.slot := slot
+        }
+    }
+
+    /**
+     * @dev Returns an `StringSlot` representation of the string storage pointer `store`.
+     */
+    function getStringSlot(string storage store) internal pure returns (StringSlot storage r) {
```
frangio, line 112: I _think_ we can make the assumption that a string will always have offset zero. But I would consider adding an assertion/require just to make sure. Do you know what I mean?

scripts/generate/templates/StorageSlot.js
```diff
@@ -0,0 +1,88 @@
+// const assert = require('assert');
```
frangio, line 1:
```suggestion
```

contracts/utils/StorageSlot.sol
Amxx: I don't

frangio: Storage pointers have an `.offset` field in addition to the `.slot` field. The offset indicates where in the slot the value begins. It can be non-zero when multiple values are packed in a slot. What I'm saying is that for a string pointer we can probably assume offset 0 but it's something we need to be aware of.

Amxx: Should we do 
```solidity
        assembly {
            r.slot := add(store.slot, store.offset)
        }
```
???

Amxx: I was not able to create a storage string pointer with an offset.

```solidity
contract Test {
    struct MyStruct {
        uint256 a;
        bool b;
        string c;
    }

    uint256 private constant KEY = uint256(keccak256("some.key.for.the.mapping"));
    mapping(uint256 => MyStruct) private mymapping;


    function getMyStruct() public view returns (uint256 slot, uint256 offset) {
        MyStruct storage ptr = mymapping[KEY];
        assembly {
            slot := ptr.slot
            offset := ptr.offset
        }
    }

    function getMyStructString() public view returns (uint256 slot, uint256 offset) {
        string storage ptr = mymapping[KEY].c;
        assembly {
            slot := ptr.slot
            offset := ptr.offset
        }
    }
}
```

Offset is 0 in both cases,
slot for `getMyStructString` is 2 more than `getMyStruct`

PR #4009, opened 2023-01-28
Add AccessControlDefaultAdminRules

### LINKED ISSUE #3623: AccessControl Admin Rules

ernestognw:
Fixes #3623
Fixes LIB-618

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/access/IAccessControlAdminRules.sol
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @dev Cancels a `DEFAULT_ADMIN_ROLE` transfer.
+     *
+     * Requirements:
+     *
+     * - Should allow to call {beginAdminTransfer} again.
+     * - Can be called even after the timer is met.
+     * - Can only be called by the current `DEFAULT_ADMIN_ROLE` owner.
+     */
+    function cancelAdminTransfer() external;
```
Amxx, line 70: Again, I don't see the point of having a `cancelAdminTransfer` in addition of `beginAdminTransfer`

Right now, we can:
- begin a transfer
- cancel
- re-initiate a transfer

So the inability to call begin when the transfer when another one is in progress does not really exist, we are just forcing calling cancel first, which IMO is just an inefficiency.

I would personally remove the `cancelAdminTransfer` function, and allow `beginAdminTransfer` to override a pending transfer.

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,220 @@
...
+    uint48 private _delayedUntil;
+
+    address private _pendingAdmin;
+
+    /**
+     * @dev Sets the initial values the internal delay and {owner}.
+     *
+     * The `delay` value is immutable. It can only be set at the constructor.
+     */
+    constructor(uint48 delay, address initialAdmin) {
```
Amxx, line 57: shouldn't the constructor set `_currentAdmin` ?

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,220 @@
...
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(IAccessControlAdminRules).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @dev See {IAccessControlAdminRules-beginAdminTransfer}
+     */
+    function beginAdminTransfer(address newAdmin) public override onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(delayedUntil() == 0, "AccessControl: pending admin already set");
```
Amxx, line 94:
```suggestion
```

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,220 @@
...
+        _delayedUntil = SafeCast.toUint48(block.timestamp) + _delay;
+        _pendingAdmin = newAdmin;
+        emit AdminRoleChangeStarted(pendingAdmin(), delayedUntil());
```
Amxx, line 97: That would be cheaper (avoid 2 sloads). It raises the question of "how to deal with overrides".
```suggestion
        emit AdminRoleChangeStarted(_pendingAdmin, _delayedUntil);
```

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * @dev See {AccessControl-_revokeRole}.
+     */
+    function _revokeRole(bytes32 role, address account) internal override {
+        if (role == DEFAULT_ADMIN_ROLE) {
+            delete _currentAdmin;
+        }
+        super._revokeRole(role, account);
+    }
```
Amxx, line 203: when doing 

```solidity
        _revokeRole(DEFAULT_ADMIN_ROLE, owner());
        _grantRole(DEFAULT_ADMIN_ROLE, pendingAdminOwner);
```

The current `_currentAdmin` is first set to 0, then written to. That is gas consuming (maybe its ok).
Also, if someone was to change that to

```solidity
        _grantRole(DEFAULT_ADMIN_ROLE, pendingAdminOwner);
        _revokeRole(DEFAULT_ADMIN_ROLE, owner());
```

things would break in a non obvious way. I don't have a better solution, but that is dangerous.

contracts/access/AccessControlAdminRules.sol
Amxx: Also, I would mark this function virtual

contracts/access/AccessControlAdminRules.sol
Amxx: Note, right now the functions are not virtual, so the overriding is not a valid reason to do the call

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,220 @@
...
+     */
+    function _resetAdminTransfer() private {
+        delete _pendingAdmin;
+        delete _delayedUntil;
+    }
+
+    /**
+     * @dev Checks if a {delayedUntil} has been set and met.
+     */
+    function _adminTransferIsUnlocked() private view returns (bool) {
```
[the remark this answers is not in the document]
ernestognw: > I would go for something like `_hasDelayPassed `

Same! I'd define a guideline for boolean private functions, maybe we can define them as `_is* private view returns (bool)` or `_has* private view returns(bool)`

contracts/access/AccessControlAdminRules.sol
ernestognw: This function is also not `virtual`, so doing `_currentAdmin` instead of owner can be the solution to writing `_grantRole` first.

The only other way it can break is by overriding `_grantRole`, but that itself is also dangerous so I wouldn't expect people doing it.

> That is gas consuming (maybe its ok).

Here are some benchmarks:

```bash
## `_revokeRole` -> `_grantRole`
Running 1 test for test/access/AccessControlAdminRules.t.sol:AdminRules
[PASS] testAccept(address) (runs: 10000, μ: 53573, ~: 53558)

## `_grantRole` -> `_revokeRole`
Running 1 test for test/access/AccessControlAdminRules.t.sol:AdminRules
[PASS] testAccept(address) (runs: 10000, μ: 51334, ~: 51318)
```

So, there IS a difference, but I needed to remove the `require(owner() == address(0), "AccessControl: admin already granted");` check in both to make the second one pass.

contracts/access/IAccessControlAdminRules.sol
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @dev Returns the timestamp in which the pending admin can claim the
+     * `DEFAULT_ADMIN_ROLE` role.
+     */
+    function delayedUntil() external view returns (uint48);
+
+    /**
+     * @dev Returns the address of the pending `DEFAULT_ADMIN_ROLE` owner.
+     */
+    function pendingAdmin() external view returns (address);
```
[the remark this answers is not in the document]
ernestognw: I had the same feeling, but I wouldn't like to introduce the concept of `owner` to the whole contract.

We may create an ERC-173 interface in `/common` and implement it in both access control and ownable contracts.
This way we can keep the `admin()` naming here and allow of others to extend without adding significant overhead (or can be implemented by default with `function owner() ... { return admin(); }`)

contracts/access/IAccessControlAdminRules.sol
ernestognw: I'd love to do this but I see no clear path for handling the `address(0)` case which is needed for renouncing, so we can't use it as a way to cancel.

Only way I see is using the `delayUntil()` as a flag

```solidity
    function beginAdminTransfer(address newAdmin) public override onlyRole(DEFAULT_ADMIN_ROLE) {
        if(delayedUntil() == 0) {
            _delayedUntil = SafeCast.toUint48(block.timestamp) + _delay;
            _pendingAdmin = newAdmin;
            emit AdminRoleChangeStarted(_pendingAdmin, _delayedUntil);
        } else {
            _resetAdminTransfer();
        }
    }
```

But I'm not convinced

Amxx: you did not get what I'm saying. 

I would do
```solidity
function beginAdminTransfer(address newAdmin) public override onlyRole(DEFAULT_ADMIN_ROLE) {
        _delayedUntil = SafeCast.toUint48(block.timestamp) + _delay;
        _pendingAdmin = newAdmin;
        emit AdminRoleChangeStarted(_pendingAdmin, _delayedUntil);
    }
```
and remove the cancel function.

If you want to cancel, just call `beginAdminTransfer(0x1)` and it won't be possible to accept.

contracts/access/AccessControlAdminRules.sol
Amxx: I just noticed that `_grantRole` requires that `owner()` is `address(0)` ... so that is good.

contracts/access/IAccessControlAdminRules.sol
ernestognw: I don't get what I'm supposed to get, then. I don't think address(1) (or any other non-zero) for cancelling is the best, I think is misleading and may require analyzing its effect with precompiles (ecrecover it's safe but I don't want to add that complexity) while making recommendations on how to cancel.

Also, `begin*` for canceling is misleading.

I'd keep the cancel.

Amxx:
It feels to me that this contract is safe, but that this safety relies on the function not being virtual, and on the user not doing "tricks".

As a dev, I just need to call `AccessControl._grantRole` instead of `_grantRole` to add a second admin.

ernestognw:
> It feels to me that this contract is safe, but that this safety relies on the functions not being virtual, and on the user not doing "tricks".
> 
> As a dev, I just need to call `AccessControl._grantRole(...)` instead of `_grantRole(...)` to add a second admin.

Yes, we agreed this is by design, is the same as in Ownable's `_transferOwnership`

ernestognw:
> It feels to me that this contract is safe, but that this safety relies on the functions not being virtual, and on the user not doing "tricks".

Sorry, it was not intentional to avoid `virtual`, the contract should still be safe. However, do you have any concerns with some override?

> As a dev, I just need to call `AccessControl._grantRole(...)` instead of `_grantRole(...)` to add a second admin.

Hmm, that's correct, but I wonder why that might happen.

frangio:
> As a dev, I just need to call `AccessControl._grantRole(...)` instead of `_grantRole(...)` to add a second admin.

This same argument applies to all of our extensions. You can mess up ERC20Votes or something by calling `ERC20._transfer` somewhere. I don't think it's something that is within our threat model at all.

contracts/access/AccessControlAdminRules.sol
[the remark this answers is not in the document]
frangio: I'm ok with the "has delay passed" concept, I would just like to specify _what_ delay. But if we can't find a good name that does that, I'm fine with `_hasDelayPassed`.

contracts/access/IAccessControlAdminRules.sol
frangio: > If you want to cancel, just call `beginAdminTransfer(0x1)` and it won't be possible to accept.

This isn't a good option. I don't see why we wouldn't want the cancel function.

contracts/access/AccessControlAdminRules.sol
frangio: | op | gas | refund |
|-|-|-|
SSTORE currentAdmin 0 | 2100 + 2900 | +4800 
SSTORE role[currentAdmin] 0 | 2100 + 2900 | +4800
SSTORE currentAdmin newAdmin | 100 | - 4800
SSTORE role[newAdmin] 1 | 2100 + 20000 | 

The total gas is 32200, the total refund is 4800. The refund is less than half of the total gas so it gets applied, resulting in 27400 final gas.

If we simply did `SSTORE currentAdmin newAdmin` in 1 step, the cost of that operation would be 5000 gas. In 2 steps, the cost is 5100 gas, so we would only be saving 100 gas.

I think the overhead of the additional step is fine.

contracts/access/IAccessControlAdminRules.sol
ernestognw: I got the length from both deployment bytecodes and this is the result:

- With cancel function: 834564
- Without the cancel: 825345 **(-1.104%)**

Using `beginAdminTransfer` would be more costly than just canceling. I don't think the deployment cost savings are worth it.

contracts/access/IAccessControlAdminRules.sol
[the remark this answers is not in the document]
Amxx: ERC-173 is already in `/interfaces`

contracts/access/AccessControlAdminRules.sol
Amxx: > I removed them, but, should `IAccessControlAdminRules` functions be virtual? thinking

The naming suggest its an interface, so all functions should be virtual.

ernestognw: It is an interface. Sorry, I'll rephrase:

> should `IAccessControlAdminRules` functions implemented in `AccessControlAdminRules` be virtual?

This is the discussion I started internally. I made them virtual.

contracts/access/IAccessControlAdminRules.sol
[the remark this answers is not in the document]
ernestognw: > Reference

I saw. Anyhow, we'll support only `owner()` with IERC5313 to avoid introducing a the concept of "owner" and make clear it's from another domain (IERC5313)

contracts/access/AccessControlAdminRules.sol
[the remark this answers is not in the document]
ernestognw: I went for `_hasAdminTransferDelayPassed` considering both criterias (`_has*` and _what_ delay passsed).

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /**
+     * @dev See {IAccessControlAdminRules-acceptAdminTransfer}
+     */
+    function acceptAdminTransfer() public virtual {
+        address pendingAdminOwner = pendingAdmin();
+        require(
+            _hasAdminTransferDelayPassed() && _msgSender() == pendingAdminOwner,
+            "AccessControl: delay must be met and caller must be pending admin"
```
frangio, line 115: String revert reasons should be shorter to avoid taking up too much bytecode.

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,228 @@
...
+        super.renounceRole(role, account);
+    }
+
+    /**
+     * @dev See {AccessControl-grantRole}. Reverts for `DEFAULT_ADMIN_ROLE`.
+     */
+    function grantRole(
+        bytes32 role,
+        address account
+    ) public virtual override(IAccessControl, AccessControl) onlyRole(getRoleAdmin(role)) {
```
frangio, line 159: This check will already be done in `super.grantRole`.

```suggestion
    ) public virtual override(IAccessControl, AccessControl) {
```

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,228 @@
...
+        super.grantRole(role, account);
+    }
+
+    /**
+     * @dev See {AccessControl-revokeRole}. Reverts for `DEFAULT_ADMIN_ROLE`.
+     */
+    function revokeRole(
+        bytes32 role,
+        address account
+    ) public virtual override(IAccessControl, AccessControl) onlyRole(getRoleAdmin(role)) {
```
frangio, line 170:
```suggestion
    ) public virtual override(IAccessControl, AccessControl) {
```

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,228 @@
...
+        require(role != DEFAULT_ADMIN_ROLE, "AccessControl: can't directly revoke admin role");
+        super.revokeRole(role, account);
+    }
+
+    /**
+     * @dev See {AccessControl-_setRoleAdmin}. Reverts for `DEFAULT_ADMIN_ROLE`.
+     */
+    function _setRoleAdmin(bytes32 role, bytes32 adminRole) internal virtual override {
+        require(role != DEFAULT_ADMIN_ROLE, "AccessControl: can't override admin's admin");
```
frangio, line 179: >admin's admin

See here it is very clear that we have a name clash between two different concepts.

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,228 @@
...
+ *}
+ * ```
+ *
+ * NOTE: The `delay` is only configurable in the constructor to avoid issues related with handling
+ * delay management during the transfer is pending to be completed.
+ *
+ * _Available since v4.9._
+ */
+abstract contract AccessControlAdminRules is IAccessControlAdminRules, IERC5313, AccessControl {
+    uint48 private immutable _delay;
```
frangio, line 46: There should be a getter for this value too.

frangio (commented):
Core logic is overall good. I want us to review naming. (See discussion above.

Still have to review docs and tests.

contracts/access/IAccessControlAdminRules.sol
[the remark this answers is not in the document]
frangio: I think @ernestognw's point is that this contract should have its own native terminology, and that it should not be `owner`, but that we can implement `owner()` in addition to the native interface for EIP-5313 compliance.

I'm also feeling unsure about "admin" here, particularly because `AccessControl` already has an "admin" concept and it means something different.

How would you feel about using the term `root` as in "root account"? A more straightforward option that could be okay is `defaultAdmin`.

contracts/access/AccessControlAdminRules.sol
Amxx: I'm ok to keep the `cancelAdminTransfer` for workflow were the cancel is not preceding a `beginAdminTransfer`.

Said otherwize: yes, lets keep the cancel function, but lets also remove the `require(delayedUntil() == 0, "AccessControl: pending admin already set");` check in `beginAdminTransfer`

Amxx: for the revert reason, why not use custom errors ?

contracts/access/IAccessControlAdminRules.sol
frangio: Copying the reply from another thread, not sure why it was there:

>I'm ok to keep the `cancelAdminTransfer` for workflow were the cancel is not preceding a `beginAdminTransfer`.
>
>Said otherwize: yes, lets keep the cancel function, but lets also remove the `require(delayedUntil() == 0, "AccessControl: pending admin already set");` check in `beginAdminTransfer`

contracts/access/AccessControlAdminRules.sol
ernestognw: Added!

contracts/access/AccessControlAdminRules.sol
ernestognw: Tested, it works. Thanks!

contracts/access/AccessControlAdminRules.sol
```diff
@@ -0,0 +1,228 @@
...
+     * @dev See {IAccessControlAdminRules-admin}
+     */
+    function admin() public view virtual returns (address) {
+        return _currentAdmin;
+    }
+
+    /**
+     * @dev See {IAccessControlAdminRules-delayedUntil}
+     */
+    function delayedUntil() public view virtual returns (uint48) {
```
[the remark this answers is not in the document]
ernestognw: Updated and set to `defaultAdminTransferDelayedUntil` as suggested since it also corresponds with what was previously agreed of using the `defaultAdmin` terminology instead.

contracts/access/IAccessControlAdminRules.sol
```diff
@@ -0,0 +1,71 @@
...
+pragma solidity ^0.8.0;
+
+import "./IAccessControl.sol";
+
+/**
+ * @dev External interface of AccessControladminRules declared to support ERC165 detection.
+ *
+ * _Available since v4.9._
+ */
+interface IAccessControlAdminRules is IAccessControl {
```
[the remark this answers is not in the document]
ernestognw: Definitely, I don't like having the docs split across functions and only referenced with `@dev See {...}`.

> which would not work with just AccessControl.

Are you referring to `type(IAccessControlAdminRules)` and `IAccessControlAdminRules` instead? Otherwise, can you clarify? I don't follow _what_ would not work.

contracts/access/AccessControlAdminRules.sol
ernestognw: It also feels odd to me to introduce them right now. I think we can safely wait until 5.0.

> I'm ok to keep the ...

Will follow in the other comments thread.

contracts/access/IAccessControlAdminRules.sol
ernestognw: Keeping both sounds good to me. Although I'd mention this implies that the `defaultAdmin` can front-run an `acceptDefaultAdminTransfer` call from the pending `defaultAdmin`, which I'd say is expected.

contracts/access/AccessControlAdminRules.sol
ernestognw: Removed 👍🏻

frangio:
Sorry from bringing up naming again but I came up with another option and don't want to let it pass. What are your thoughts on "main admin" as opposed to "default admin"?

ernestognw:
> Sorry for bringing up naming again but I came up with another option and don't want to let it pass. What are your thoughts on "main admin" as opposed to "default admin"?

I'd question the reasons. Why do you think it's better?
From my perspective, `mainAdmin` would be shorter but not enough to justify the lack of clarity when it refers to default admin. I think it's better it's clear relation with the `DEFAULT_ADMIN_ROLE`

I'd consider renaming both `DEFAULT_ADMIN_ROLE` to `MAIN_ADMIN_ROLE` and `defaultAdmin` to `mainAdmin` in 5.0  to keep consistency but I don't completely like the idea, `DEFAULT_ADMIN_ROLE` is a well-known terminology.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,255 @@
...
+ *
+ * _Available since v4.9._
+ */
+abstract contract AccessControlDefaultAdminRules is IERC5313, AccessControl {
+    uint48 private immutable _delay;
+
+    address private _currentDefaultAdmin;
+    address private _pendingDefaultAdmin;
+
+    uint48 private _defaultAdminTransferDelayedUntil;
```
Amxx, line 49: I'm not sure we discussed that, but is there a reason to favor `DelayedUntil` over `Deadline` ?

IMO deadline is simpler and works just as well.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,255 @@
...
+     */
+    constructor(uint48 delay_, address initialDefaultAdmin) {
+        _delay = delay_;
+        _grantRole(DEFAULT_ADMIN_ROLE, initialDefaultAdmin);
+    }
+
+    /**
+     * @dev Returns the delay between each `DEFAULT_ADMIN_ROLE` transfer.
+     */
+    function delay() public view returns (uint48) {
```
Amxx, line 71: any reason why this one is not virtual ?

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,255 @@
...
+    function defaultAdminTransferDelayedUntil() public view virtual returns (uint48) {
+        return _defaultAdminTransferDelayedUntil;
+    }
+
+    /**
+     * @dev See {IERC165-supportsInterface}.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return
+            interfaceId == 0x4a02b518 || // type(IAccessControlDefaultAdminRules).interfaceId
```
Amxx, line 108: Now that there is no interface anymore, we should have an comment somewhere that lists all the functions which are part of the interfaceId computation

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,255 @@
...
+     * @dev Starts a `DEFAULT_ADMIN_ROLE` transfer by setting a pending default admin
+     * and a timer to be met.
+     *
+     * Requirements:
+     *
+     * - Only can be called by the current `DEFAULT_ADMIN_ROLE` holder.
+     *
+     * Emits a {DefaultAdminRoleChangeStarted}.
+     */
+    function beginDefaultAdminTransfer(address newAdmin) public virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```
Amxx, line 122: Should we have an internal function `_beginDefaultAdminTransfer` with the unprotected logic, and just call it from this `public onlyRole(DEFAULT_ADMIN_ROLE)` function?

That would allow workflows such as `beginDefaultAdminTransferWithSignature`

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * @dev Completes a `DEFAULT_ADMIN_ROLE` transfer.
+     *
+     * Requirements:
+     *
+     * - Caller should be the pending default admin.
+     * - `DEFAULT_ADMIN_ROLE` should be granted to the caller.
+     * - `DEFAULT_ADMIN_ROLE` should be revoked from the previous holder.
+     */
+    function acceptDefaultAdminTransfer() public virtual {
```
Amxx, line 137: Same here, we should probably have

```solidity
    function acceptDefaultAdminTransfer() public virtual {
        require(_msgSender() == pendingDefaultAdmin(), "AccessControl: can't accept defaultAdmin");
        _acceptDefaultAdminTransfer();
    }

    function _acceptDefaultAdminTransfer() internal {
        require(_hasDefaultAdminTransferDelayPassed(), "TODO");
        _revokeRole(DEFAULT_ADMIN_ROLE, defaultAdmin());
        _grantRole(DEFAULT_ADMIN_ROLE, pendingDefaultAdmin());
        _resetDefaultAdminTransfer();
    }
```

Amxx (commented):
Almost there!

Still have some minor questions/comments, sorry.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,255 @@
...
+ * * Enforce a configurable delay between the two steps, with the ability to cancel in between.
+ *   - Even after the timer has passed to avoid locking it forever.
+ * * The `DEFAULT_ADMIN_ROLE` 's admin can be only itself.
+ *
+ * Example usage:
+ *
+ * ```solidity
+ * contract MyToken is AccessControlDefaultAdminRules {
+ *   constructor() AccessControlDefaultAdminRules(
+ *     60 * 60 * 24, // 3 days
```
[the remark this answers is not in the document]
ernestognw: Totally right, just changed it for `3 days`. Thanks for the catch 👍🏻

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Not in particular, we were using the raw private value (`_delay`) in `beginDefaultAdminTransfer` but I double-checked and it makes sense to have it virtual.

Fixed (and also solving since the recommendation is clear)

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Do you think it's enough with the tests?
Somehow we're explicitly setting them in [these tests](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/files#diff-156d4e821e84d3ce615fe8bce6b2d6c21c5274752de071ab271cf84b4581d780R40-R47), but we don't pack them for release.

No strong opinion, but not sure how we'll put them in the code.

Amxx: IMO tests are not documentation. I'd have a list in the comments here.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Answering this and this [other comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009#discussion_r1112855585), I recall @frangio mentioned he was not a big fan of this pattern.

From my perspective:
1. I don't see a valuable reason to include it **now** (other use cases are definitely valuable but not sure if they're a priority)
2. Makes it more difficult to deal with overrides since now you have 2 (although this may add value)
3. I like `beginDefaultAdminTransferWithSignature` but I don't want to rush it and may add complexity to this PR

If we decide to eventually pursue `beginDefaultAdminTransferWithSignature`, let's move separate them :p

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: We agreed during a call to do the following:

- Add `IAccessControlDefaultAdmin` back to its own file
- Don't show it in the docs yet to avoid docs duplication.
- Implemented [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/f420162c0eafad65187bdc219a67d5a4854beb64).

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: We agreed in our weekly to separate the `public` from the `internal` implementation of both `beginDefaultAdminTransfer` and `acceptDefaultAdminTransfer`.

[Implemented in this commit](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/8747bf4863fe4dd3cdee0904888315d6f5552fed) and solving since the recommendation was taken.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009#discussion_r1113438214

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+
+  describe('accepts transfer admin', async function () {
+    let correctPendingDefaultAdmin;
+    let correctIncreaseTo;
+
+    beforeEach(async function () {
+      // Set as correct so they're explicitly needed in revert tests, since conditions
+      // are not mutually exclusive and accidentally an incorrect value will make it revert
+      // possiblty creating false positives (eg. expect revert for incorrect caller but getting it
+      // from a badly expected delayed until)
```
frangio, line 328: I don't understand what this is saying. Can you rephrase it?

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+      correctPendingDefaultAdmin = newDefaultAdmin;
+      correctIncreaseTo = (await time.latest()).add(delay).addn(1);
+
+      await this.accessControl.beginDefaultAdminTransfer(correctPendingDefaultAdmin, { from: defaultAdmin });
+    });
+
+    describe('caller is pending default admin and delayed until is met', async function () {
```
frangio, line 335:
```suggestion
    describe('caller is pending default admin and delay has passed', async function () {
```

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+          role: DEFAULT_ADMIN_ROLE,
+          account: newDefaultAdmin,
+        });
+      });
+
+      it('accepts a transfer resetting pending default admin and delayed until', async function () {
+        await this.accessControl.acceptDefaultAdminTransfer({ from });
+        expect(await this.accessControl.defaultAdminTransferDelayedUntil()).to.be.bignumber.equal(web3.utils.toBN(0));
+        expect(await this.accessControl.pendingDefaultAdmin()).to.equal(ZERO_ADDRESS);
+      });
```
frangio, line 366: We should put all of these in a single test IMO. @Amxx What do you think?

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+
+      it('should revert if block.timestamp is equal to delayed until', async function () {
+        await time.increaseTo(incorrectIncreaseTo);
+        await expectRevert(
+          this.accessControl.acceptDefaultAdminTransfer({ from: correctPendingDefaultAdmin }),
+          `${errorPrefix}: delayedUntil not met`,
+        );
+      });
+
+      it('should revert if block.timestamp is less to delayed until', async function () {
```
frangio, line 392:
```suggestion
      it('should revert if block.timestamp is less than delayed until', async function () {
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+        super._grantRole(role, account);
+    }
+
+    /**
+     * @dev See {acceptDefaultAdminTransfer}.
+     *
+     * Internal function without access restriction.
+     */
+    function _acceptDefaultAdminTransfer() internal virtual {
+        require(_hasDefaultAdminTransferDelayPassed(), "AccessControl: delayedUntil not met");
```
frangio, line 200: Also not sure why camel case here.

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+      });
+    });
+  });
+
+  describe('cancel transfer default admin', async function () {
+    beforeEach(async function () {
+      await this.accessControl.beginDefaultAdminTransfer(newDefaultAdmin, { from: defaultAdmin });
+    });
+
+    it('resets pending default admin and delayed until', async function () {
```
frangio, line 407: I would add assertions in this test showing that the previously pending admin can't accept after cancellation.

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+        await time.increaseTo(incorrectIncreaseTo.subn(1));
+        await expectRevert(
+          this.accessControl.acceptDefaultAdminTransfer({ from: correctPendingDefaultAdmin }),
+          `${errorPrefix}: delayedUntil not met`,
+        );
+      });
+    });
+  });
+
+  describe('cancel transfer default admin', async function () {
```
frangio, line 402: We should add another test showing that the transfer can be cancelled after the delay.

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+  });
+
+  it('should not grant the default admin role twice', async function () {
+    await expectRevert(
+      this.accessControl.$_grantRole(DEFAULT_ADMIN_ROLE, defaultAdmin),
+      `${errorPrefix}: defaultAdmin already granted`,
+    );
+  });
+
+  describe('begins transfer default admin', async function () {
```
frangio, line 259: This should not be async! Please review all `describe`s. (I don't think it has any effect though.)

```suggestion
  describe('begins transfer default admin', function () {
```

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+  describe('renouncing admin', async function () {
+    let correctIncreaseTo;
+    let from = defaultAdmin;
+
+    beforeEach(async function () {
+      correctIncreaseTo = (await time.latest()).add(delay).addn(1);
+      await this.accessControl.beginDefaultAdminTransfer(ZERO_ADDRESS, { from });
+    });
+
+    describe('caller is default admin and delayed until is met', async function () {
```
frangio, line 430: I find the phrase "delayed until is met" very hard to parse.

```suggestion
    describe('caller is default admin and delay has passed', async function () {
```

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+      it('emits events', async function () {
+        expectEvent(receipt, 'RoleRevoked', {
+          role: DEFAULT_ADMIN_ROLE,
+          account: from,
+        });
+      });
+
+      it('marks the contract as not owned', async function () {
+        expect(await this.accessControl.owner()).to.equal(ZERO_ADDRESS);
+      });
```
frangio, line 453: These also look to me like they should be just one test... Now I'm not sure if what I'm saying goes against what we've done across the test suite in general.

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+
+      it('reverts if block.timestamp is equal to delayed until', async function () {
+        await time.increaseTo(incorrectIncreaseTo);
+        await expectRevert(
+          this.accessControl.renounceRole(DEFAULT_ADMIN_ROLE, defaultAdmin, { from }),
+          `${errorPrefix}: only can renounce in two delayed steps`,
+        );
+      });
+
+      it('reverts if block.timestamp is less to delayed until', async function () {
```
frangio, line 487:
```suggestion
      it('reverts if block.timestamp is less than delayed until', async function () {
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+import "./IAccessControlDefaultAdminRules.sol";
+import "../utils/math/SafeCast.sol";
+import "../interfaces/IERC5313.sol";
+
+/**
+ * @dev Extension of {AccessControl} that allows specifying special rules to manage
+ * the `DEFAULT_ADMIN_ROLE` holder, which is a sensitive role with special permissions
+ * over other roles that may potentially have privileged rights in the system.
+ *
+ * If a specific role doesn't have an `adminRole` assigned, the holder of the
```
frangio, line 16:
```suggestion
 * If a specific role doesn't have an admin role assigned, the holder of the
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+ * `DEFAULT_ADMIN_ROLE` will have the ability to manage it, as determined by the
+ * function {getRoleAdmin}'s default value (`address(0)`).
```
frangio, line 18:
```suggestion
 * `DEFAULT_ADMIN_ROLE` will have the ability to grant it and revoke it.
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * This contract implements the following risk mitigations on top of the {AccessControl} implementation:
```
frangio, line 20:
```suggestion
 * This contract implements the following risk mitigations on top of {AccessControl}:
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+ *
+ * * Only one account holds the `DEFAULT_ADMIN_ROLE` at every time after construction except when it's renounced.
```
frangio, line 22:
```suggestion
 * * Only one account holds the `DEFAULT_ADMIN_ROLE` since deployment until it's potentially renounced.
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+ * contract MyToken is AccessControlDefaultAdminRules {
+ *   constructor() AccessControlDefaultAdminRules(
+ *     3 days,
+ *     _msgSender() // Explicit initial `DEFAULT_ADMIN_ROLE` holder
+ *    ) {}
+ *}
+ * ```
+ *
+ * NOTE: The `delay` is only configurable in the constructor to avoid issues related with handling
+ * delay management during the transfer is pending to be completed.
```
frangio, line 40: No need to mention the reasons IMO.

```suggestion
 * NOTE: The `delay` can only be set in the constructor and is fixed thereafter.
```

frangio (commented):
Mostly comments about the tests.

frangio:
Reviewing the docs made me think that the delay should be configurable... That said I want us to merge this PR and we can discuss it as a potential follow up.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+     */
+    constructor(uint48 delay_, address initialDefaultAdmin) {
+        _delay = delay_;
+        _grantRole(DEFAULT_ADMIN_ROLE, initialDefaultAdmin);
+    }
+
+    /**
+     * @dev Returns the delay between each `DEFAULT_ADMIN_ROLE` transfer.
+     */
+    function delay() public view virtual returns (uint48) {
```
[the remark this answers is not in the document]
ernestognw: Done [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/24ce7fa5850617ee3a396bcf0ed2659e5fac90cb). Although I'd need to double-check the namin.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+    function _resetDefaultAdminTransfer() private {
+        delete _pendingDefaultAdmin;
+        delete _defaultAdminTransferDelayedUntil;
+    }
+
+    /**
+     * @dev Checks if a {defaultAdminTransferDelayedUntil} has been set and met.
+     */
+    function _hasDefaultAdminTransferDelayPassed() private view returns (bool) {
+        uint48 defaultAdminTransferDelayedUntilTimestamp = defaultAdminTransferDelayedUntil();
```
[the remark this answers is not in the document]
ernestognw: Fixed at https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/2c4e15eb6ccbc5832c13ae360defd3bc53c078b1

contracts/access/README.adoc
```diff
@@ -23,3 +23,7 @@ This directory provides ways to restrict who can access the functions of a contr
 {{IAccessControlEnumerable}}
 
 {{AccessControlEnumerable}}
+
+{{IAccessControlDefaultAdminRules}}
```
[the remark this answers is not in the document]
ernestognw: Sorry, this was a mistake from my side, I initially didn't remove it when `IAccessControlDefaultAdminRules` was removed at all.

Just removed it from the [docs here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/281209c1ef0c9df275c6d6e04de5a51ac95ba696).

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -0,0 +1,244 @@
...
+            );
+        }
+        super.renounceRole(role, account);
+    }
+
+    /**
+     * @dev See {AccessControl-grantRole}. Reverts for `DEFAULT_ADMIN_ROLE`.
+     */
+    function grantRole(bytes32 role, address account) public virtual override(AccessControl, IAccessControl) {
+        require(role != DEFAULT_ADMIN_ROLE, "AccessControl: can't directly grant defaultAdmin role");
```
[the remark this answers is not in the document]
ernestognw: I tried to do it for clear referencing to the `defaultAdmin` variable but no strong opinion right now. It made sense when we only had `admin` and I brought that idea here.

Replaced in [this commit](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/2c49a5dd5b58f9d07b9bc5dd470d0b1481ad3e88)

test/access/AccessControl.behavior.js
```diff
@@ -210,8 +211,293 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
...
+    });
+
+    it('should be able to begin a transfer again before delay pass', async function () {
+      // First defaultAdmin transfer started
+      await this.accessControl.beginDefaultAdminTransfer(newDefaultAdmin, { from: defaultAdmin });
+      const firstDefaultAdminTransferDelayedUntil = (await time.latest()).add(delay);
+      expect(await this.accessControl.pendingDefaultAdmin()).to.equal(newDefaultAdmin);
+      expect(await this.accessControl.defaultAdminTransferDelayedUntil()).to.be.bignumber.equal(
+        firstDefaultAdminTransferDelayedUntil,
+      );
```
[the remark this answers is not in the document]
ernestognw: Added [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/39d176ad8cfde6e062711e361bcc997b8887cb44)

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/2c49a5dd5b58f9d07b9bc5dd470d0b1481ad3e88 as well

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Not convinced about this one.

What about?

```suggestion
 * * Since deployment, only one account holds the `DEFAULT_ADMIN_ROLE` until it's renounced.
```

test/access/AccessControl.behavior.js
ernestognw: Any reasons to be strict with this? I was looking for a linter rule but only [found for Typescript](https://typescript-eslint.io/rules/no-misused-promises/) if it helps.

Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/5a31431a59746687d655ff35ab0264e533513380

test/access/AccessControl.behavior.js
ernestognw: Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/14994be0e20da5b1a19046f0f435cdd304583dc1

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/fd64f5d4bb5038a3e037afb3a4a22c7c7bbf1dba since I couldn't accept this suggestion directly.

Solving since it'd have had the same effect

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: For clarity, "potentially" didn't sound right to me but now it does.

Thought about it twice and it now makes sense since I couldn't parse it at first because I was reading code. In the docs should be fine.

Solving with the initial suggestion

test/access/AccessControl.behavior.js
ernestognw: I think I put them separately because the last one that allows recovery via `_grantRole` used the same `before`, but it's cleaner the way you propose.

Added in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/ac093f42f13bbb6b52386c5c16212e3207d51ace

test/access/AccessControl.behavior.js
ernestognw: Added in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/394824665fe7bb4f15bb8b899111cff0b02a6dfe

test/access/AccessControl.behavior.js
ernestognw: Also added in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/394824665fe7bb4f15bb8b899111cff0b02a6dfe

test/access/AccessControl.behavior.js
ernestognw: I double-checked and I think it's not needed anymore so I'm removing it at https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/880ac0b876437d1618b81ff8785f0233901517c2

Still, I have a question, so let me explain my reasoning before asking:

Currently we have 2 revert strings in `acceptDefaultAdminTransfer`:

- `AccessControl: pending admin must accept`
- `AccessControl: transfer delay not passed`

Those two come from different conditions that were previously merged in the same require:

```solidity
require(isPendingDefaultAdmin() && delayHasPassed(), "AccessControl: Only one revert string for both");
```

The reason why I initially set this comment was because I was explicitly renaming `newDefaultAdmin` to `correctDefaultAdmin` to be clear about the fact that inside the `accepts transfer admin` describe block, the `newDefaultAdmin` is **the** `correctDefaultAdmin`. This is relevant because we had tests like this:

```javascript
it('should revert if block.timestamp is less than delayed until', async function () {
  await time.increaseTo(delayNotPassed.subn(1));
  await expectRevert(
    this.accessControl.acceptDefaultAdminTransfer({ from: newDefaultAdmin }),
    `${errorPrefix}: transfer delay not passed`,
  );
});
```

The reasoning was that it's easier to make a mistake if it reads `newDefaultAdmin` and someone overrode it accidentally in the `before` function. The function would start to throwing the same error without nobody noticing it was because of an incorrect `defaultAdmin`, not because of a delay not yet passed.

I initially didn't split the `require(a() && b(), "a && b")` into `require(a(), "a")` and `require(b(), "b")` because I know it's costly to keep two revert strings.

My question is:
What do we usually prefer? More revert strings for better clarity or less for gas savings?

test/access/AccessControl.behavior.js
ernestognw: Kept this one until the end so the commit can be easily reverted, but I unified them at https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/a43abe7e27f82755ba2d3d47546a7bf6286cb4c7

It's consistent with what I did at https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4009/commits/ac093f42f13bbb6b52386c5c16212e3207d51ace

ernestognw:
> Reviewing the docs made me think that the delay should be configurable... That said I want us to merge this PR and we can discuss it as a potential follow up. But let me know your thoughts.

We can discuss it further. The initial reason for not doing it was avoiding the management it comes with.
That said, we can lock the delay if there's no started transfer if that makes sense to you.

If we'll do it. I'd like to do it in a follow-up PR.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: I think it expresses the opposite idea.

While a `delayedUntil` gives a sense of something that can be done _after_, whereas `deadline` seems to communicate that an action can be done _before_.

Any reason to keep it as `Deadline`?

contracts/access/AccessControlDefaultAdminRules.sol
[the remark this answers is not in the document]
frangio: I only meant the function name, but renaming the variable as well doesn't hurt either.

test/access/AccessControl.behavior.js
frangio: > What do we usually prefer? More revert strings for better clarity or less for gas savings?

I don't have a good answer for this. We have to find some good point in the middle.

Regarding the variable name, I sort of see your point, but I'm not as concerned about it.

test/access/AccessControl.behavior.js
frangio: As far as I know, unless we use `await`, there should be no bad consequence from using async functions here.

contracts/access/AccessControlDefaultAdminRules.sol
frangio: I agree that deadline seems like the opposite idea. But I would also like it if we had a shorter simpler option than "delayed until".

frangio (approved):
This is good to merge, but we should probably follow up with modifiable delay in another PR.

We could also review the naming for "delayed until".

frangio (approved):
LGTM. Also @Amxx agreed to merge. We will follow up with configurable delay in another PR, and a final review round.

frangio:
There is a flaky test because of the use of `time.increaseTo`, which mines a block with the argument timestamp, but does not guarantee the timestamp of the following block.

frangio (approved):

PR #4010, opened 2023-01-28
Add `GUIDELINES.md` for marking `abstract` contracts

### LINKED ISSUE #4350: Mark all contract that are not "ready to use" as abstract
**🧐 Motivation**
Some of our contract are not abstract from a solidity point of view, but are useless if deployed as is. I'm thinking of contracts like ERC20 that could be deployed with a name an symbol, but the resulting contract would have no way to mint tokens, so the supply would remain 0 forever (making the contract useless).

It appears that some users are making that mistake, particularly using Remix:

Remix will let me deploy ERC20, 

Capture d’écran du 2023-06-14 11-27-37

even though what I should deploy is MyToken, that inherit ERC20 and adds some minting logic

Capture d’écran du 2023-06-14 11-27-48

if ERC20 was marked as abstract I would have received a warning when trying to deploy ERC20

Capture d’écran du 2023-06-14 11-30-48

**📝 Details**
By default make all our contract "abstract". Only remove that keyword for some handpicked contract (VestingWallet, TimelockController, ...)

ernestognw:
Fixes #4350

## Description

This is one of our guidelines we use across the repository, so it's now formally mentioned in our `GUIDELINES.md` document


Amxx:
AFAIK, this is all our contracts with very few exceptions (proxies / timelockcontroller).
Should we change them all to abstract

ernestognw:
> AFAIK, this is all our contracts with very few exceptions (proxies / timelockcontroller / vestingwallet / paymentsplitter / presets). Should we change them all to abstract

You're right, the wording is not precise, however, I chatted about this with Fran and there's clearly a need for a guideline here

Maybe this?

> Contracts intended not to be used standalone...

instead of 

> Contracts not intended to be used standalone...

Amxx:
> What about...?
> 
> > Contracts not designed for standalone usage are usually marked abstract so they are required to be inherited from other contracts.

Honestly I prefer the guidelines to be strict, and not open to interpretations. I like the way its phrased right now, but it also mean we should probably change how our contracts are defined.

frangio:
> we should probably change how our contracts are defined.

Are you suggesting we should do this now or in 5.0?

ernestognw:
> Honestly I prefer the guidelines to be strict, and not open to interpretations. I like the way its phrased right now, but it also mean we should probably change how our contracts are defined.

I also don't like ambiguity, but I think it's fine since the guidelines will be all correctly implemented long-term, so I agree we shouldn't change it until 5.0.

Same happens right now with changesets (eg. they should not have the PR URL, but they do now because of older PRs. Will get fixed over time).

ernestognw:
We agreed to keep this guideline but applying breaking changes only in 5.0

frangio (approved):

Amxx:
I added the abstract keyword to some contract definitions.

The list of non abstract contract can be obtain by doing
```sh
grep -rl ^contract --exclude-dir=mocks contracts
``` 

I left the following ones non-abstract
- `contracts/finance/VestingWallet.sol`
- `contracts/governance/TimelockController.sol`
- `contracts/metatx/MinimalForwarder.sol`
- `contracts/proxy/ERC1967/ERC1967Proxy.sol`
- `contracts/proxy/beacon/BeaconProxy.sol`
- `contracts/proxy/beacon/UpgradeableBeacon.sol`
- `contracts/proxy/transparent/ProxyAdmin.sol`
- `contracts/proxy/transparent/TransparentUpgradeableProxy.sol`

Amxx (dismissed):

frangio (approved):

PR #4012, opened 2023-01-30
Add IERC4906 interface

Amxx:
#### PR Checklist

### DIFF
--- a/.changeset/lovely-dragons-appear.md
+++ b/.changeset/lovely-dragons-appear.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`IERC4906`: Add an interface for EIP-4906 that is now final.
--- a/contracts/interfaces/IERC4906.sol
+++ b/contracts/interfaces/IERC4906.sol
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "./IERC165.sol";
+import "./IERC721.sol";
+
+/// @title EIP-721 Metadata Update Extension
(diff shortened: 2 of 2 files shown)


TrejGun:
Hi @Amxx

I just tried to use this extension with my contracts and it looks like there is no need to extend IERC165 and IERC721. This only makes me override `supportInterface`

TrejGun:
Hi @Amxx I tried to use this extension today and have a couple of questions

The standard says IERC4906 extends IERC165 and IERC721

This makes me override `supportInterfaces` in a couple of places in my project but gives me nothing useful because 
```solidity
console.logBytes4(type(IERC4906).interfaceId);
```
returns `0x00000000` which corresponds to the empty or broken interface so I can't check my contract for IERC4906. Maybe it makes sense to remove those two?

frangio (approved):
Added a changeset for ERC721URIStorage. LGTM.

pcaversaccio:
Hey, maybe am wrong here but I was just going through these changes - one question: shouldn't you add a test [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/utils/introspection/SupportsInterface.behavior.js) as well? I noted that the interface ID `0x49064906` for `IERC4906` is kinda special since it does not have any functions (not sure how it was calculated, maybe you know?), so `makeInterfaceId` wouldn't work.

Amxx:
That is a good point. In fact in `ERC721Storage` we did not add ERC165 support for `0x49064906`.

pcaversaccio:
Right, btw I quickly [asked](https://github.com/ethereum/EIPs/pull/5542#issuecomment-1434670168) about the `interfaceId` and it was chosen randomly.

PR #4013, opened 2023-01-30
Add IERC5313.sol

Amxx:
EIP is final: https://eips.ethereum.org/EIPS/eip-5313

Should we inherit that in `Ownable`? `AccessControlAdmin`?

#### PR Checklist

### DIFF
--- a/contracts/interfaces/IERC5313.sol
+++ b/contracts/interfaces/IERC5313.sol
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/// @title EIP-5313 Light Contract Ownership Standard
+interface IERC5313 {
+    /// @notice Get the address of the owner
+    /// @return The address of the owner
+    function owner() view external returns(address);
+}
--- a/contracts/interfaces/IERC5313.sol
+++ b/contracts/interfaces/IERC5313.sol
@@ -2,9 +2,16 @@
 
 pragma solidity ^0.8.0;
 
-/// @title EIP-5313 Light Contract Ownership Standard
+/**
+ * @dev Interface for the Light Contract Ownership Standard.
+ *
+ * A standardized minimal interface required to identify an account that controls a contract
+ *
+ * _Available since v4.9._
+ */
 interface IERC5313 {
-    /// @notice Get the address of the owner
-    /// @return The address of the owner
+    /**
+     * @dev Gets the address of the owner.
+     */
     function owner() external view returns (address);
 }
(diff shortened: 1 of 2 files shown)


contracts/interfaces/IERC5313.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/// @title EIP-5313 Light Contract Ownership Standard
+interface IERC5313 {
+    /// @notice Get the address of the owner
+    /// @return The address of the owner
+    function owner() external view returns (address);
+}
```
ernestognw, line 10: This will conflict with #4009. I can merge back but I'd still advocate for docs consistent with ours

```suggestion
/**
 * @dev Interface for the Light Contract Ownership Standard.
 *
 * A standardized minimal interface required to identify an account that controls a contract
 *
 * _Available since v4.9._
 */
interface IERC5313 {
    /**
     * @dev Gets the address of the owner.
     */
    function owner() external view returns (address);
}
```

Both `@notice` and `@title` are not picked up from our docs generation

ernestognw (changes requested):

contracts/interfaces/IERC5313.sol
Amxx: I just copied the interface provided in the (final) EIP. How is that in conflict with #4009? If that is the case, #4009 should either not advertise 5313 compatibility ... of change to be compatible

Amxx:
@frangio Do we want the `interface` folder to contain:

- exact copied of the interfaces defined in the EIPs (and not properly show up in our docs)
- modified versions that are not exact copies of the EIP document, but that show up in our docs.

ernestognw:
> @frangio Do we want the `interface` folder to contain:
> 
> * exact copied of the interfaces defined in the EIPs (and not properly show up in our docs)
> * modified versions that are not exact copies of the EIP document, but that show up in our docs.

It's already modified, we're not keeping the Solidity version and LICENSE. I know it's a minor detail but since we're doing that, why not keep our doc slightly more explicit?

contracts/interfaces/IERC5313.sol
ernestognw: Conflicts because I added it as well there to make it 5313 compatible with different comments, just that.

frangio:
I don't think we need interfaces to be copied exactly from the EIPs. We can definitely improve the documentation.

frangio:
>Should we inherit that in Ownable?

It doesn't really make any difference does it? Either way is fine with me.

>`AccessControlAdmin`?

Yes I think so. By the way, that may be a good name rather than `AccessControlAdminRules`.

frangio (approved):

frangio:
Upstream Foundry/Solidity error in foundry tests.

PR #4017, opened 2023-01-31
Add warning for `supportsERC165InterfaceUnchecked` edge case

### LINKED ISSUE #4011: Document Edge Case for `supportsERC165InterfaceUnchecked(..)`
## What this issue is about ?

The `supportsERC165(..)` function when calling to check if an address supports a  specific interfaceId, according to the ERC165 specs, it will check whether the address queried is supporting the ERC165 interfaceId, the queried interfaceId, also will check that the address queried is not supporting the `0xffffffff` interfaceId.

The `supportsERC165InterfaceUnchecked(..)` function will check whether a specific address supports a single interfaceId, through ERC165. However, some addresses will return always true regardless if they support the interfaceId or not.

These are the precompile addresses `0x000..0000002`, `0x000..0000003`, and `0x000..0000004`. It's unclear to me why this behavior is happening. This issue was pointed out by Runtime verification auditors while doing some fuzzing tests.

**📝 Details**

This happens only to `supportsERC165InterfaceUnchecked(..)`, because since the precompile addresses `0x000..0000002`, `0x000..0000003`, and `0x000..0000004` will return always true for all interfaceIds, it means that these addresses are also supporting the `0xffffffff` interfaceId. 

Since `supportsERC165(..)` checks that the address does not support `0xffffffff` interfaceId, then `supportsERC165(..)` is excluded from this behavior.

This issue is opened to add documentation to `supportsERC165InterfaceUnchecked(..)` function using natspec, to acknowledge this edge case, and also to make it easy for developers doing fuzzing tests using foundry, to identify why the tests are failing when lacking context that these addresses will always return true.   

**🔢 Code to reproduce bug**

You can call `supportsERC165InterfaceUnchecked(..)` on `0x000..0000002` , `0x000..0000003` and `0x000..0000004` addresses with any interfaceId to reproduce the case.

```js

const precompiledAddress = [
      "0x0000000000000000000000000000000000000002",
      "0x0000000000000000000000000000000000000003",
      "0x0000000000000000000000000000000000000004",
    ];

    for (let i = 0; i < precompiledAddress.length; i++) {
      for (let ii = 0; ii < 100; ii++) {
        const result = await contract.supportsERC165Interface(
          precompiledAddress[i],
          ethers.utils.hexlify(ethers.utils.randomBytes(4))
        );
        expect(result).to.be.true;
      }
    }
```

Please let me know if it makes sense to add this to the function natspec.

YamenMerhi:
Fixes #4011 

#### PR Checklist

- [ ] Tests
- [X] Documentation

### DIFF
--- a/contracts/utils/introspection/ERC165Checker.sol
+++ b/contracts/utils/introspection/ERC165Checker.sol
@@ -101,6 +101,10 @@ library ERC165Checker {
      * @dev Assumes that account contains a contract that supports ERC165, otherwise
      * the behavior of this method is undefined. This precondition can be checked
      * with {supportsERC165}.
+     *
+     * Some precompiled contracts will falsely indicate support for a given interface, so caution
+     * should be exercised when using this function.
+     *
      * Interface identification is specified in ERC-165.
      */
     function supportsERC165InterfaceUnchecked(address account, bytes4 interfaceId) internal view returns (bool) {


Amxx (approved):
Thank you @YamenMerhi 

This looks good to me. Waiting on @frangio's review

frangio (approved):

PR #4021, opened 2023-02-01
ERC777 gas optimization

MollyHe5:
For two numbers a and b, if a is always larger than or equal to b, unchecking addition operations on b or subtraction operations on a can save some gases. 
Each item in `_balances` is smaller than `_totalSupply`. So unchecking addition operations on items in `_balances` can save some gases.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC777/ERC777.sol
+++ b/contracts/token/ERC777/ERC777.sol
@@ -326,7 +326,9 @@ contract ERC777 is Context, IERC777, IERC20 {
 
         // Update state variables
         _totalSupply += amount;
-        _balances[account] += amount;
+        unchecked{
+            _balances[account] += amount;
+        }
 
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @MollyHe5 

This is an optimization that we have in ERC20, but that we did not port to ERC777 since we decided to deprecate it.
We will officially mark ERC777 as deprecated in 4.9, and remove it from the library in 5.0, so I'm not sure whether we want to do such change or not.

Note: very similarly, the `totalSupply` update in `_burn` (line 388) could be unchecked.

frangio:
As @Amxx mentioned, we are deprecating ERC777 and for this reason we prefer to not making any changes to it at this moment.

PR #4022, opened 2023-02-01
Gas optimisation ERC721

Deathstore:
Optimising ERC721 check length in `token URI` function
in line 97 
`bytes(baseURI).length == 0` cost less than `bytes(baseURI).length > 0`


Amxx:
Hello @Deathstore 

Both the `gt` and the `eq` opcode cost the same 3 gas ... how do you expect this change to improve gas usage?

The [test suite report](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/4069356069) doesn't show any improvement.

frangio:
Closing due to lack of follow up.

PR #4023, opened 2023-02-01
Add a library for handling short strings in a gas efficient way

Amxx:
Allow string → bytes32 with string storage fallback.

This provides an effective way to store short strings (<32chars) in immutable storage.

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx:
For now compilation gives
```
Error: No node with id 25682 of type StructDefinition,EnumDefinition,ContractDefinition
    at deref (/home/amxx/Work/OZ/openzeppelin-contracts/node_modules/solidity-ast/src/ast-dereferencer.ts:36:11)
    at curried (/home/amxx/Work/OZ/openzeppelin-contracts/node_modules/solidity-ast/src/ast-dereferencer.ts:54:14)
    at isTypeExternalizable (/home/amxx/Work/OZ/openzeppelin-contracts/node_modules/hardhat-exposed/src/core.ts:287:21)
    at /home/amxx/Work/OZ/openzeppelin-contracts/node_modules/hardhat-exposed/src/core.ts:280:53
    at Array.every (<anonymous>)
    at isExternalizable (/home/amxx/Work/OZ/openzeppelin-contracts/node_modules/hardhat-exposed/src/core.ts:280:42)
    at /home/amxx/Work/OZ/openzeppelin-contracts/node_modules/hardhat-exposed/src/core.ts:90:84
    at Array.filter (<anonymous>)
    at /home/amxx/Work/OZ/openzeppelin-contracts/node_modules/hardhat-exposed/src/core.ts:90:72
    at Function.from (<anonymous>)
```
because hardhat-exposed doesn't support UDVT

contracts/utils/ShortStrings.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @dev Return the length of a `ShortString`.
+     */
+    function length(ShortString sstr) internal pure returns (uint256) {
+        return uint256(ShortString.unwrap(sstr)) & 0xFF;
+    }
+
+    /**
+     * @dev Encode a string into a `ShortString`, or write it to storage if it is too long.
+     */
+    function setWithFallback(string memory value, string storage store) internal returns (ShortString) {
```
frangio, line 53: The "fallback" is storage, so I don't like `setWithFallback` as a name because you only "set" when you fallback.

What about `toShortStringWithFallback`?

Similarly we could use `toStringWithFallback`

contracts/utils/ShortStrings.sol
```diff
@@ -0,0 +1,72 @@
...
+        if (bytes(value).length < 32) {
+            return toShortString(value);
+        } else {
+            StorageSlot.getStringSlot(store).value = value;
+            return toShortString("");
```
frangio, line 58: This just seems like it will be obviously more efficient:

```suggestion
            return ShortString.wrap("");
```

.changeset/violet-frogs-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`ShortStrings`: New library for handling shortstring storage in immutable variables.
```
frangio, line 5:
```suggestion
`ShortStrings`: Added a library for handling short strings in a gas efficient way, with fallback to storage for longer strings.
```

contracts/utils/ShortStrings.sol
```diff
@@ -0,0 +1,72 @@
...
+
+pragma solidity ^0.8.8;
+
+import "./StorageSlot.sol";
+
+/**
+ * @dev Short string operations.
+ */
+library ShortStrings {
+    type ShortString is bytes32;
```
frangio, line 11: I'd prefer to define this outside of the library, so we can just write `ShortString` when we use the type. Otherwise we will have to use `ShortStrings.ShortString`.

So far we've defined types inside libraries, but I think we need to chage that in 5.0.

test/utils/ShortStrings.test.js
```diff
@@ -0,0 +1,44 @@
+const { expect } = require('chai');
+const { expectRevertCustomError } = require('../helpers/customError');
+
+const ShortStrings = artifacts.require('$ShortStrings');
+
+function decode(sstr) {
+  return web3.utils.toUtf8(sstr).slice(0, parseInt(sstr.slice(64), 16));
```
frangio, line 7:
```suggestion
  const length = parseInt(sstr.slice(64), 16);
  return web3.utils.toUtf8(sstr).slice(0, length);
```

test/utils/ShortStrings.test.js
```diff
@@ -0,0 +1,44 @@
...
+          expect(decoded).to.be.equal(str);
+        } else {
+          await expectRevertCustomError(this.mock.$toShortString(str), `StringTooLong("${str}")`);
+        }
+      });
+
+      it('set / get with fallback', async function () {
+        const { ret0 } = await this.mock
+          .$setWithFallback(str, 0)
+          .then(({ logs }) => logs.find(({ event }) => event == 'return$setWithFallback').args);
```
frangio, line 35: Let's keep `then` for one-liners only, please.

```suggestion
        const { logs } = await this.mock.$setWithFallback(str, 0);
        const { ret0 } = logs.find(({ event }) => event == 'return$setWithFallback').args);
```

frangio:
Please review the PR title according to the guidelines: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/GUIDELINES.md#pull-requests

contracts/utils/ShortStrings.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.8;
+
+import "./StorageSlot.sol";
+
+/**
+ * @dev Short string operations.
```
frangio, line 8:
```suggestion
 * @dev This library provides functions to convert short memory strings
 * into a `ShortString` type that can be used as an immutable variable.
 * Strings of arbitrary length can be optimized if they are short enough by
 * the addition of a storage variable used as fallback.
```

contracts/utils/ShortStrings.sol
Amxx: I was thinking

```solidity
contract NamedContract {
    ShortStrings.ShortString private _name;
    string private _nameFallback;
    
    constructor(string memory name_) {
        _name = ShortStrings.setWithFallback(name_, _nameFallback);
    }
    
    function name() public view returns (string memory) {
        return ShortStrings.getWithFallback(_name, _nameFallback);
    }
}
```

How would you do it?

contracts/utils/ShortStrings.sol
Amxx: what about the custom error ?

contracts/utils/ShortStrings.sol
Amxx: would you prefer

```
constructor(string memory name_) {
    _name = name_.toShortStringWithFallback(_nameFallback);
}
function name() public view returns (string memory) {
    return _name.toStringWithFallback(, _nameFallback);
}
```
?

frangio: I was thinking the same. Those are good.

We should add that as an example snippet in the docs.

contracts/utils/ShortStrings.sol
frangio: I would say the custom error should stay inside the library.

frangio (approved):
LGTM!

PR #4028, opened 2023-02-04
Update ERC1155.sol

ersanyakit:
No need call functions and address zero checks...

Fixes #???? 

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx:
This is needed in case `balanceOf` is overriden.

PR #4038, opened 2023-02-09
Allow return data length >= 32 in SignatureChecker

### LINKED ISSUE #4035: Use `>=32` instead of `==32` in `SignatureChecker` for return data length check
In the `ERC4626` function [`_tryGetAssetDecimals`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol) you use the following pattern with respect to the return data length check:

```solidity
if (success && encodedDecimals.length >= 32) {
    uint256 returnedDecimals = abi.decode(encodedDecimals, (uint256));
    if (returnedDecimals <= type(uint8).max) {
        return (true, uint8(returnedDecimals));
    }
}
```

In the function [`isValidSignatureNow`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol#L34) you use `result.length == 32` instead for the return data length check. I think it would make sense due to consistency to also use `result.length >= 32` in that case. FYI, I already discussed this issue with @frangio over Discord.

HarshitSharma007:
Used  >=32 instead of ==32 in SignatureChecker for return data length check

Fixes #4035 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/warm-masks-obey.md
+++ b/.changeset/warm-masks-obey.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-Allow return data length >= 32 in SignatureChecker
+`SignatureChecker`: Allow return data length greater than 32 from EIP-1271 signers.
--- a/.changeset/warm-masks-obey.md
+++ b/.changeset/warm-masks-obey.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Allow return data length >= 32 in SignatureChecker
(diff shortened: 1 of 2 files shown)


Amxx (dismissed):
LGTM

need @frangio review

frangio (approved):
LGTM

frangio (approved):

Amxx (dismissed):

Amxx (approved):

PR #4040, opened 2023-02-10
Add StaticDelegateCall contract

vsergeev:
This primitive allows the creation of view-only getters to contracts that call dynamic external libraries with `delegatecall`. It is inspired by Gnosis' `StorageSimulation` ( https://github.com/gnosis/util-contracts/blob/main/contracts/storage/StorageSimulation.sol ), but this implementation prohibits disallows callers and is meant to be used internally for building view functions.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/mocks/StaticDelegateCallMock.sol
+++ b/contracts/mocks/StaticDelegateCallMock.sol
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "../utils/StaticDelegateCall.sol";
+
+contract StaticDelegateCallImplementationMock {
+    uint256 private _x;
+
+    function process(uint256 c) external view returns (uint256) {
+        require(c < 5);
+        return _x + c;
+    }
+}
+
+contract StaticDelegateCallMock is StaticDelegateCall {
+    uint256 private _x;
+
+    address private _implementation;
+
+    constructor(uint256 x, address implementation) {
+        _x = x;
+        _implementation = implementation;
+    }
+
+    function process(uint256 c) external view returns (uint256) {
+        (bool success, bytes memory result) = _staticDelegateCall(
+            _implementation,
+            abi.encodeCall(StaticDelegateCallImplementationMock.process, (c))
+        );
+        if (!success) revert("Implementation reverted");
+        return abi.decode(result, (uint256));
+    }
+}
--- a/contracts/utils/StaticDelegateCall.sol
+++ b/contracts/utils/StaticDelegateCall.sol
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Provides emulation of static delegatecalls for implementing view-only
+ * functions that execute another contract's code with the calling contract's
+ * context.
+ */
+abstract contract StaticDelegateCall {
+    /**
+     * @dev Perform a static delegate call to the specified `target` with
+     * calldata `data`. Returns success boolean and return data.
+     */
+    function _staticDelegateCall(address target, bytes memory data) internal view returns (bool, bytes memory) {
+        /* This staticcall always reverts, so we can ignore the success boolean */
+        (, bytes memory returndata) = address(this).staticcall(
+            abi.encodeCall(this.delegateCallAndRevert, (target, data))
+        );
+        return abi.decode(returndata, (bool, bytes));
+    }
+
+    /**
+     * @dev Simulate a delegatecall to the specified `target` with calldata
+     * `data` by reverting with the results. The resulting success boolean and
+     * return data are ABI-encoded in the revert data.
+     *
+     * Inspired by simulateAndRevert() from Gnosis' StorageSimulation contract.
+     */
+    function delegateCallAndRevert(address target, bytes memory data) external {
+        require(msg.sender == address(this), "StaticDelegateCall: caller must be self");
+        (bool success, bytes memory returndata) = target.delegatecall(data);
+        bytes memory result = abi.encode(success, returndata);
+        assembly {
+            revert(add(result, 0x20), mload(result))
+        }
+    }
+}
--- a/docs/modules/ROOT/pages/utilities.adoc
+++ b/docs/modules/ROOT/pages/utilities.adoc
@@ -188,3 +188,49 @@ await instance.multicall([
     instance.contract.methods.bar().encodeABI()
 ]);
 ----
+
+=== StaticDelegateCall
+
+The `StaticDelegateCall` abstract contract provides emulation of static delegatecalls for implementing view-only functions that execute another contract's code with the calling contract's context. It can be useful in implementing dynamic library support at runtime.
+
+Example:
+[source,solidity]
+----
+// contracts/DynamicGadget.sol
+// SPDX-License-Identifier: MIT
+
+interface IGadget {
+    function compute(uint256 c) external view returns (uint256);
+}
+
+contract GadgetStorage {
+    uint256 internal _x;
+}
+
+contract GadgetImplementationA is GadgetStorage, IGadget {
+    function compute(uint256 c) external view returns (uint256) {
+        return _x * c;
+    }
+}
+
+contract GadgetImplementationB is GadgetStorage, IGadget {
+    function compute(uint256 c) external view returns (uint256) {
+        return _x / c;
+    }
+}
+
+contract DynamicGadget is GadgetStorage, StaticDelegateCall {
+    address private _implementation;
+
+    ...
+
+    function compute(uint256 c) external view returns (uint256) {
+        (bool success, bytes memory result) = _staticDelegateCall(
+            _implementation,
+            abi.encodeCall(IGadget.compute, (c))
+        );
+        require(success);
+        return abi.decode(result, (uint256));
+    }
+}
+----
--- a/.changeset/neat-jobs-think.md
+++ b/.changeset/neat-jobs-think.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`StaticDelegateCall`: Add utility contract that emulates static delegatecalls.
--- a/test/utils/StaticDelegateCall.test.js
+++ b/test/utils/StaticDelegateCall.test.js
@@ -1,0 +1,33 @@
+const { expectRevert } = require('@openzeppelin/test-helpers');
+const { expect } = require('chai');
+
+const StaticDelegateCallImplementationMock = artifacts.require('StaticDelegateCallImplementationMock');
+const StaticDelegateCallMock = artifacts.require('StaticDelegateCallMock');
+
+contract('StaticDelegateCall', function () {
+  beforeEach(async function () {
+    this.implementation = await StaticDelegateCallImplementationMock.new();
+    this.mock = await StaticDelegateCallMock.new('5', this.implementation.address);
+  });
+
+  describe('proxied implementation', function () {
+    it('processes input', async function () {
+      expect(await this.mock.process('2')).to.be.bignumber.equal('7');
+    });
+    it('reverts on out of bounds parameter', async function () {
+      await expectRevert(this.mock.process('10'), 'Implementation reverted');
+    });
+  });
+
+  describe('delegateCallAndRevert', function () {
+    it('reverts on invalid caller', async function () {
+      await expectRevert(
+        this.mock.delegateCallAndRevert(
+          this.implementation.address,
+          this.implementation.contract.methods.process('2').encodeABI(),
+        ),
+        'StaticDelegateCall: caller must be self',
+      );
+    });
+  });
+});


contracts/utils/StaticDelegateCall.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev Provides emulation of static delegatecalls for implementing view-only
+ * functions that execute another contract's code with the calling contract's
+ * context.
+ */
+abstract contract StaticDelegateCall {
```
ernestognw, line 10: Can this be a library? Every file in /utils is and I think this fits into that same category.

contracts/utils/StaticDelegateCall.sol
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * @dev Simulate a delegatecall to the specified `target` with calldata
+     * `data` by reverting with the results. The resulting success boolean and
+     * return data are ABI-encoded in the revert data.
+     *
+     * Inspired by simulateAndRevert() from Gnosis' StorageSimulation contract.
+     */
+    function delegateCallAndRevert(address target, bytes memory data) external {
```
ernestognw, line 30: Am I understanding correctly that this file is an `abstract contract` because this function needs to be external?

If that's the case I'd make this call internally and let the users expose it, so we can make this a library.
Note that we generally avoid libraries with external calls, so [library linking](https://docs.soliditylang.org/en/latest/using-the-compiler.html#library-linking) is not needed.

I understand making it `internal` conflicts with the need for `_staticDelegateCall` to call `delegateCallAndRevert`, but still makes sense to me that an implementer would be required to expose it.

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -188,3 +188,49 @@ await instance.multicall([
...
+    function compute(uint256 c) external view returns (uint256) {
+        return _x * c;
+    }
+}
+
+contract GadgetImplementationB is GadgetStorage, IGadget {
+    function compute(uint256 c) external view returns (uint256) {
+        return _x / c;
+    }
+}
```
ernestognw, line 220: Why do we need two?

In case we don't need them, wouldn't be a clearer example something like:?

```solidity
contract GadgetImplementation {
   uint256 internal _x;

    function compute(uint256 c) external view returns (uint256) {
        return _x * c;
    }
}
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -188,3 +188,49 @@ await instance.multicall([
...
 ----
+
+=== StaticDelegateCall
+
+The `StaticDelegateCall` abstract contract provides emulation of static delegatecalls for implementing view-only functions that execute another contract's code with the calling contract's context. It can be useful in implementing dynamic library support at runtime.
+
+Example:
+[source,solidity]
+----
+// contracts/DynamicGadget.sol
```
ernestognw, line 199: Although its purpose is to showcase how the static delegatecall works, I'd say is difficult to follow because the purpose of `Gadget` is unclear.

Do you think we can provide a cleaner example? Maybe a Proxy would work here but I understand if it's too complex

contracts/mocks/StaticDelegateCallMock.sol
```diff
@@ -0,0 +1,34 @@
...
+        _x = x;
+        _implementation = implementation;
+    }
+
+    function process(uint256 c) external view returns (uint256) {
+        (bool success, bytes memory result) = _staticDelegateCall(
+            _implementation,
+            abi.encodeCall(StaticDelegateCallImplementationMock.process, (c))
+        );
+        if (!success) revert("Implementation reverted");
```
ernestognw, line 31:
```suggestion
        require(success, "Implementation reverted");
```

ernestognw (commented):
Hey @vsergeev, sorry for the late answer.

We've been working on the next 4.9 release, and it's our current priority. Since this is an addition that might require discussion, I'm just adding some comments.

contracts/utils/StaticDelegateCall.sol
Amxx: It can't because it relied on the contract exposing `delegateCallAndRevert` publicly.

vsergeev: This one was modeled after `Multicall`, but I can look into it. I had a few questions below.

contracts/utils/StaticDelegateCall.sol
vsergeev: I'm a little unsure of how this would work as a library -- which target would `_staticDelegateCall()` encode? Would the callsite get passed in somehow?

docs/modules/ROOT/pages/utilities.adoc
vsergeev: Yeah, I can simplify it to one if it gets the point across.

docs/modules/ROOT/pages/utilities.adoc
vsergeev: In our case, we were instantiating a contract with a possibly user specified "dynamic library" at runtime (i.e. not inlined at compile time, nor externally linked at deploy time), that would act on contract storage but do implementation-specific things, while still conforming to an interface like `IGadget`. But we ran into the issue of not being able to write any view functions.

Our use case might eventually be solved at the language level, e.g. ethereum/solidity#8490, ethereum/solidity#2469, ethereum/solidity#8353, but there are other uses for a static delegate call, like emulating external logic, providing generic access to contract storage, and from my research it appears other people were running into this issue in the past: https://ethereum.stackexchange.com/questions/71037/anything-like-a-static-delegatecall-from-a-view-function https://ethereum.stackexchange.com/questions/82342/how-to-perform-delegate-call-inside-of-view-call-staticall . I'll try to think of a better example.

contracts/utils/StaticDelegateCall.sol
ernestognw: Just realized Multicall is an abstract contract as well, so that'd make a case for keeping this one abstract.
I'll answer your questions in each thread.

contracts/utils/StaticDelegateCall.sol
ernestognw: What I see is that these two functions are strictly tied together while they belong to the same action, any reason why they should be decoupled?

What I'm saying is that, if we can couple them together into a single function we might be able to either publish it as an `internal` function in a library or as an `external` one in an `abstract contract`, similar to what we do with Multicall:

```solidity
    function staticDelegateCall(
        address target,
        bytes memory data,
        bool delegated
    ) external returns (bool, bytes memory) {
        return
            delegated
                ? _delegateCallAndRevert(target, data)
                : _staticDelegateCall(target, data);
    }

    function _staticDelegateCall(
        address target,
        bytes memory data
    ) private view returns (bool, bytes memory) {
        /* This staticcall always reverts, so we can ignore the success boolean */
        (, bytes memory returndata) = address(this).staticcall(
            abi.encodeCall(this.staticDelegateCall, (target, data, true))
        );
        return abi.decode(returndata, (bool, bytes));
    }

    function _delegateCallAndRevert(
        address target,
        bytes memory data
    ) private returns (bool, bytes memory) {
        require(
            msg.sender == address(this),
            "StaticDelegateCall: caller must be self"
        );
        (bool success, bytes memory returndata) = target.delegatecall(data);
        bytes memory result = abi.encode(success, returndata);
        assembly {
            revert(add(result, 0x20), mload(result))
        }
    }
```

Although I'm not a big fan of the pattern, it makes clear what to use if we export this as an abstract contract.
Does something like this make sense to you?

docs/modules/ROOT/pages/utilities.adoc
ernestognw: > But we ran into the issue of not being able to write any view functions.

What do you mean with that? Can you provide an example? I think the example could be simplified to only using a `staticDelegateCall` regardless of the use case and then listing the use cases after the example.

Unless there's a relevant security consideration to account, simple would be better at first sight.

contracts/utils/StaticDelegateCall.sol
vsergeev: As a clarification regarding coupling -- the original coupling is necessary to have a view entry point (using `staticcall`) and a non-view executor (using `delegatecall`).

I'm trying to understand this approach -- are you thinking that the `external staticDelegateCall()` would go into `StaticDelegateCall` abstract contract, while `internal _staticDelegateCall()` and `internal _delegateCallAndRevert()` would go into the existing `Address` library, and the `StaticDelegateCall` abstract contract would then use the `Address` library as `Multicall` does?

I'm not sure this approach makes things cleaner overall, because the helpers `_staticDelegateCall()` and `_delegateCallAndRevert()` aren't useful on their own, and this arrangement imposes a few additional costs:
1) The footprint grows from two functions in a single abstract contract, to three functions spread across an abstract contract and a library, while still being coupled
2) There is an additional trampoline `_staticDelegateCall()` -> `staticDelegateCall()` -> `_delegateCallAndRevert()`
3) The `delegated` boolean parameter in `staticDelegateCall()` is not necessary and potentially misleading

There are also may be a few things missing with the posted implementation, i.e. `_staticDelegateCall()` and `_delegateCallAndRevert()` need to be at least internal for it to work, but maybe I'm not understanding where these functions are supposed to live.

Unless you think there may be another use case for `_delegateCallAndRevert()` in `Address`, I still think the original implementation is cleaner, easier to use, and easier to understand.

docs/modules/ROOT/pages/utilities.adoc
vsergeev: > What do you mean with that? Can you provide an example? I think the example could be simplified to only using a staticDelegateCall regardless of the use case and then listing the use cases after the example.

It's simply that you can't write a `view` function that uses `delegatecall`, so there's no way to write a getter that uses a contract's storage with another contract's logic at runtime. (Externally linked libraries can, but they have to be linked at deploy time.)

docs/modules/ROOT/pages/utilities.adoc
vsergeev: Addressed in 007f271d.

frangio:
I think this makes sense as a primitive, but we always ask for concrete use cases. Would you mind sharing a few?

contracts/utils/StaticDelegateCall.sol
frangio: I agree that the original implementation is cleaner. Even though this proposal is able to put all the logic inside a single external `staticDelegateCall` function, it does this at the expense of adding an extraneous boolean argument that is only used for internal logic. Additionally, the API is compromised because the user has to use `this.staticDelegateCall(..., false)`. If the end goal is to provide functionality to be used internally in a contract, an internal `_staticDelegateCall(...)` function is the best fit.

vsergeev:
> I think this makes sense as a primitive, but we always ask for concrete use cases. Would you mind sharing a few?
> 
> > In our case, we were instantiating a contract with a possibly user specified "dynamic library" at runtime (i.e. not inlined at compile time, nor externally linked at deploy time), that would act on contract storage but do implementation-specific things, while still conforming to an interface like IGadget.
> 
> Can you share a little more context about this?

The two use cases that come to mind are 1) dynamic libraries and 2) external storage access, which differ primarily in whether the `StaticDelegateCall` is used internally, or exposed and called externally.

The dynamic library use case is similar to having an externally linked library with a well-defined API, except that the library can be specified at runtime, and even changed at runtime. 

In our case, we had pools with interest rates that are computed from the current liquidity state in storage. The liquidity state is complex and relatively large. As there are many different ways to build good (and bad) interest rate models, we wanted the flexibility to develop and experiment with new interest rate models, and allow for user-defined ones, so that pools could compete on the merits of their interest rate models, among other things. We also wanted to do this without redeploying the pool implementation for every new interest rate model. In some cases, the interest rate models may have some of their own state too.

Basically, the contracts looked something like this:

``` solidity
interface IInterestRateModel {
  ...
  function rate() external view returns (uint256);
}

contract Pool {
  Liquidity storage _liquidity; /* complex */
  IInterestRateModel _interestRateModel;
  ...
  function rate() external view returns (uint256) {
    /* gets rate from interest rate model, based on current liquidity */
  }
  
  function doStuffWithLiquidity() external {
    /* uses rate to price liquidity */
  }
}
```

There are a couple of different approaches to the interest rate model computing a rate based on the Pool's liquidity state:

1) Deploy the interest rate model as an independent instance and pass liquidity state to it in calldata with `call`/`staticcall`
    - This requires copying the liquidity state from storage to memory to make it available as calldata, which is fine if the liquidity state is simple and small, but in our case was too expensive. It also requires a separate interest rate model instance for each pool.
2) Inline the interest rate model as a library at compile time, or link it as an external library at deploy time
    - This is not ideal, as we would like to have cheap clones of a single pool logic contract and do not want to burden a user with redeploying the entire pool contract when instantiating a permissionless pool.
3) Deploy the interest rate model as a "dynamic library" and `delegatecall` it from the pool

Option 3 allows for the interest rate models to be deployed once as logic contracts that can then be used by different pools. However, since there is no Solidity language support for "dynamic libraries" (see ethereum/solidity#8490, ethereum/solidity#2469, ethereum/solidity#8353), this kind of interaction must be done manually with `delegatecall`, and it's currently not possible to create simple view getters (like `rate()` in `Pool` above) without emulating them with something like `StaticDelegateCall`.

Another application of this use case might be a generic `Governor` that collects votes for a proposal, but uses a dynamic library to decide the weighting of the votes, similar to how Snapshot has [voting strategies](https://github.com/snapshot-labs/snapshot-strategies/tree/master/src/strategies). The `Governor` contract can be a standardized deployment, while projects can specify their own voting strategy at deploy time or runtime — without compile-time inheritance or overrides — with the potential of upgrading the voting strategy with time or even using different voting strategies for different proposals.

The other use case is external storage access. In this use case, a contract would expose the `StaticDelegateCall` to external contracts and allow for arbitrary access to its storage. This was done in the Gnosis v2 contracts with their [`StorageAccessible` mixin](https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/mixins/StorageAccessible.sol) (also the inspiration for this implementation) and attachable [readers](https://github.com/gnosis/gp-v2-contracts/tree/main/src/contracts/reader) for interpreting contract state. For contract developers, this might be done to future-proof the contract API (at least for view functions), in case the API was missing a getter to some internal state, or to outsource the logic for more complicated view functions. More interestingly, it could be used to expose internal state onchain to other contracts for downstream logic. One example of this would be an [Ante Test](https://docs.ante.finance/ante-docs/) that asserts some internal invariant in a contract that may not have been exposed at all by the contract's public API.

I still think the ideal version of this would be at the language level and/or with EVM support, rather than the `delegateCallAndRevert` mechanism, but I thought this might be useful to some projects in the meanwhile.

Amxx:
> I still think the ideal version of this would be at the language level and/or with EVM support, rather than the `delegateCallAndRevert` mechanism, but I thought this might be useful to some projects in the meanwhile.

This should probably be proposed as a new EIP. I know it takes ages to get one of these in a network upgrade, so the sooner its proposed the better.

While I agree that some project might use such a mechanism, I also believe that it is way more subtle and complex than something like the `Multicall` contract. IMO, anyone that needs that is knowledgeable enough to write it and use it. On the other hand what message do we give by providing it in the library? That this is a recommended/good practice? Its not just providing the code (which in fairness is not that complex). Its also putting out documentation, and supporting users that may want to use that. I'm personally not feeling comfortable with that last part.

contracts/utils/StaticDelegateCall.sol
```diff
@@ -0,0 +1,38 @@
...
+        require(msg.sender == address(this), "StaticDelegateCall: caller must be self");
```
Amxx, line 31: Out of curiosity, what is the attack/issue that this require protect against?

vsergeev: In this case, it's to ensure that the feature doesn't introduce any unexpected side effects that the user didn't opt into. Without the `msg.sender` check, `delegateCallAndRevert()` exposes all of the contract's storage to any other contract onchain, which isn't necessarily a security issue (in many ways it's a feature), but is a substantial side effect that didn't exist prior to inheriting the contract. If the user does want that effect, they would opt into knowingly by creating an external wrapper for `_staticDelegateCall()` that explicitly exposes it to other contracts.

PR #4041, opened 2023-02-11
Make transpilation setup local to this repo

frangio:
This PR migrates the transpilation of contracts for upgradeability from [the upgadeable repo](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable) to this repo. The result should be a lot simpler to use and understand.

### DIFF
--- a/hardhat/env-artifacts.js
+++ b/hardhat/env-artifacts.js
@@ -1,0 +1,20 @@
+const { HardhatError } = require('hardhat/internal/core/errors');
+
+extendEnvironment(env => {
+  const artifactsRequire = env.artifacts.require;
+
+  env.artifacts.require = (name) => {
+    for (const suffix of ['UpgradeableWithInit', 'Upgradeable', '']) {
+      try {
+        return artifactsRequire(name + suffix);
+      } catch (e) {
+        if (HardhatError.isHardhatError(e) && e.number === 700 && suffix !== '') {
+          continue;
+        } else {
+          throw e;
+        }
+      }
+    }
+    throw new Error('Unreachable');
+  };
+});
--- a/scripts/upgradeable/patch-apply.sh
+++ b/scripts/upgradeable/patch-apply.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+DIRNAME="$(dirname -- "${BASH_SOURCE[0]}")"
+PATCH="$DIRNAME/upgradeable.patch"
+
+error() {
+  echo Error: "$*" >&2
+  exit 1
+}
+
+if ! git diff --quiet contracts; then
+  error "Changes in contracts"
+fi
+
+if ! git apply -3 "$PATCH"; then
+  error "Fix conflicts and run $DIRNAME/patch-save.sh"
+fi
--- a/scripts/upgradeable/transpile.sh
+++ b/scripts/upgradeable/transpile.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+set -euo pipefail -x
+
+DIRNAME="$(dirname -- "${BASH_SOURCE[0]}")"
+
+bash "$DIRNAME/patch-apply.sh"
+
+npm run compile
+
+build_info=($(jq -r '.input.sources | keys | if any(test("^contracts/mocks/.*\\bunreachable\\b")) then empty else input_filename end' artifacts/build-info/*))
+build_info_num=${#build_info[@]}
+
+if [ $build_info_num -ne 1 ]; then
+  echo "found $build_info_num relevant build info files but expected just 1"
+  exit 1
+fi
+
+# -D: delete original and excluded files
+# -b: use this build info file
+# -i: use included Initializable
+# -x: exclude all proxy contracts except Clones library
+# -p: emit public initializer
+npx @openzeppelin/upgrade-safe-transpiler@latest -D \
+  -b "$build_info" \
+  -i contracts/proxy/utils/Initializable.sol \
+  -x 'contracts-exposed/**/*' \
+  -x 'contracts/proxy/**/*' \
+  -x '!contracts/proxy/Clones.sol' \
+  -x '!contracts/proxy/ERC1967/ERC1967Storage.sol' \
+  -x '!contracts/proxy/ERC1967/ERC1967Upgrade.sol' \
+  -x '!contracts/proxy/utils/UUPSUpgradeable.sol' \
+  -x '!contracts/proxy/beacon/IBeacon.sol' \
+  -p 'contracts/**/presets/**/*'
+
+# this is currently no longer used but could be useful again in the future
+# for p in scripts/upgradeable/patch/*.patch; do
+#   git apply "$p"
+# done
--- a/scripts/upgradeable/upgradeable.patch
+++ b/scripts/upgradeable/upgradeable.patch
@@ -80,10 +80,10 @@ index 00000000..3315e7bd
 +    }
 +}
 diff --git a/contracts/governance/extensions/GovernorVotes.sol b/contracts/governance/extensions/GovernorVotes.sol
-index b328c25d..68bca95e 100644
+index 64431711..885f0e42 100644
 --- a/contracts/governance/extensions/GovernorVotes.sol
 +++ b/contracts/governance/extensions/GovernorVotes.sol
-@@ -10,6 +10,8 @@ import "../utils/IVotes.sol";
+@@ -10,6 +10,8 @@ import "../../interfaces/IERC5805.sol";
   * @dev Extension of {Governor} for voting weight extraction from an {ERC20Votes} token, or since v4.5 an {ERC721Votes} token.
   *
   * _Available since v4.3._
@@ -91,9 +91,9 @@ index b328c25d..68bca95e 100644
 + * @custom:storage-size 51
   */
  abstract contract GovernorVotes is Governor {
-     IVotes public immutable token;
+     IERC5805 public immutable token;
 diff --git a/contracts/governance/extensions/GovernorVotesComp.sol b/contracts/governance/extensions/GovernorVotesComp.sol
-index 8c4a1f8c..0ea4a891 100644
+index 17250ad7..1d26b72e 100644
 --- a/contracts/governance/extensions/GovernorVotesComp.sol
 +++ b/contracts/governance/extensions/GovernorVotesComp.sol
 @@ -10,6 +10,8 @@ import "../../token/ERC20/extensions/ERC20VotesComp.sol";
@@ -247,7 +247,7 @@ index a357199b..9dc8e894 100644
  abstract contract ERC20Permit is ERC20, IERC20Permit, EIP712 {
      using Counters for Counters.Counter;
 diff --git a/contracts/token/ERC20/extensions/ERC20Wrapper.sol b/contracts/token/ERC20/extensions/ERC20Wrapper.sol
-index 8b153ffc..63f7c6ce 100644
+index ce515693..39e93513 100644
 --- a/contracts/token/ERC20/extensions/ERC20Wrapper.sol
 +++ b/contracts/token/ERC20/extensions/ERC20Wrapper.sol
 @@ -14,6 +14,8 @@ import "../utils/SafeERC20.sol";
@@ -258,7 +258,7 @@ index 8b153ffc..63f7c6ce 100644
 + * @custom:storage-size 51
   */
  abstract contract ERC20Wrapper is ERC20 {
-     IERC20 public immutable underlying;
+     IERC20 private immutable _underlying;
 diff --git a/contracts/token/ERC20/utils/TokenTimelock.sol b/contracts/token/ERC20/utils/TokenTimelock.sol
 index ed855b7b..3d30f59d 100644
 --- a/contracts/token/ERC20/utils/TokenTimelock.sol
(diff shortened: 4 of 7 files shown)


scripts/upgradeable/upgradeable.patch
```diff
@@ -0,0 +1,350 @@
...
++    // ERC1967Upgrade._functionDelegateCall is private so we reproduce it here.
++    // An extra underscore prevents a name clash error.
++    function __functionDelegateCall(address target, bytes memory data) private returns (bytes memory) {
++        require(Address.isContract(target), "Address: delegate call to non-contract");
++
++        // solhint-disable-next-line avoid-low-level-calls
++        (bool success, bytes memory returndata) = target.delegatecall(data);
++        return Address.verifyCallResult(success, returndata, "Address: low-level delegate call failed");
++    }
+ }
```
Amxx, line 23: this should not be necessary anymore

scripts/upgradeable/upgradeable.patch
```diff
@@ -0,0 +1,350 @@
...
```
Amxx, line 144: same

scripts/upgradeable/upgradeable.patch
```diff
@@ -0,0 +1,350 @@
...
++     * _Available since v3.4._
++     */
++    function _functionDelegateCall(address target, bytes memory data) private returns (bytes memory) {
++        require(Address.isContract(target), "Address: delegate call to non-contract");
++
++        // solhint-disable-next-line avoid-low-level-calls
++        (bool success, bytes memory returndata) = target.delegatecall(data);
++        return Address.verifyCallResult(success, returndata, "Address: low-level delegate call failed");
++    }
+ }
```
Amxx, line 209: same

scripts/upgradeable/upgradeable.patch
```diff
@@ -0,0 +1,350 @@
...
+-        bytes memory data,
+-        string memory errorMessage
+-    ) internal returns (bytes memory) {
+-        (bool success, bytes memory returndata) = target.delegatecall(data);
+-        return verifyCallResultFromTarget(target, success, returndata, errorMessage);
+-    }
+-
+     /**
+      * @dev Tool to verify that a low level call to smart-contract was successful, and revert (either by bubbling
+      * the revert reason or using the provided one) in case of unsuccessful call or if target was not a contract.
```
Amxx, line 310: should no longer be removed

scripts/upgradeable/upgradeable.patch
```diff
@@ -0,0 +1,350 @@
...
++     * _Available since v3.4._
++     */
++    function _functionDelegateCall(address target, bytes memory data) private returns (bytes memory) {
++        require(Address.isContract(target), "Address: delegate call to non-contract");
++
++        // solhint-disable-next-line avoid-low-level-calls
++        (bool success, bytes memory returndata) = target.delegatecall(data);
++        return Address.verifyCallResult(success, returndata, "Address: low-level delegate call failed");
++    }
+ }
```
Amxx, line 338: same

Amxx (commented):
patch could be simplified

hardhat/env-artifacts.js
```diff
@@ -0,0 +1,20 @@
+const { HardhatError } = require('hardhat/internal/core/errors');
+
+extendEnvironment(env => {
```
Amxx, line 3: IMO, there should be a comment explaining what this actually does and why its needed.

hardhat/task-test-get-files.js
```diff
@@ -0,0 +1,36 @@
+// Ignores the proxy tests after they are removed by the transpiler
+
+const { internalTask } = require('hardhat/config');
+const { TASK_TEST_GET_TEST_FILES } = require('hardhat/builtin-tasks/task-names');
+
+internalTask(TASK_TEST_GET_TEST_FILES).setAction(async ({ testFiles }, { config }) => {
```
Amxx, line 6: same, we need some documentation here

scripts/upgradeable/patch-apply.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+DIRNAME="$(dirname -- "${BASH_SOURCE[0]}")"
+PATCH="$DIRNAME/upgradeable.patch"
```
Amxx, line 6: I'm wondering if we should have multiple patch files:
```$DIRNAME/upgradeable-patches/*.patch```

The advantage is that we could more effectivelly segment the different things that are patched (instead of having a single difficult to read patch file). The downside is that patch generation is more complexe. Maybe that is something we can delay until we actually need it (if ever)

Comment (frangio) [reply scripts/upgradeable/patch-apply.sh:6]: Yeah the reason for a single patch is that `patch-save.sh` can be super simple.

File: hardhat/task-test-get-files.js (hunk 7 above)
Comment (frangio) [reply hardhat/task-test-get-files.js:6]: There was a comment at the top of the file. But I moved it down and rewrote it a little.

**Review by Hasinafidahussain — COMMENTED** [review]
- 

> #Sir easy pisa ya bank ka account number 03443616250 per Apne name k  sath send kare 

>

File: scripts/upgradeable/transpile-onto.sh
Comment (frangio) [inline scripts/upgradeable/transpile-onto.sh:1]: I'm sorry that this file has no comments... I wrote it a long time ago and I should've written the comments back then. :grimacing:

**Comment by frangio** [comment]
Moved this back into draft because after merging the latest `master` I found an issue with the EIP712 tests caused by the transpiler.

**Comment by frangio** [comment]
I addressed the EIP712 issues and this should be good now. It turned out to be tricky so here are some notes to help understand.

The EIP712 contract uses immutable variables quite extensively, and it's even designed to work well behind a proxy. I think we initially intended that code to go as-is in the upgradeable contracts, however, once we transpile the rest of the codebase we run into conflicts because those immutable variables need to be set in a constructor whereas everything else will be changed to initializers, and the transpiler doesn't support this kind of hybrid transformation.

To fix this, the [upgradeable patch](https://github.com/frangio/openzeppelin-contracts/blob/transpile/scripts/upgradeable/upgradeable.patch#L272) now removes most of the immutable variables (the remaining are converted into storage variables by the transpiler), including the cached domain separator. The new code after transpilation can be seen in [this branch in my fork](https://github.com/frangio/openzeppelin-contracts/blob/transpile-result/contracts/utils/cryptography/EIP712Upgradeable.sol), and is very similar to the [`EIP712`](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/release-v4.8/contracts/utils/cryptography/EIP712Upgradeable.sol) we have in v4.8.

One of the changes in v4.9 is that we now need the name and version for `eip712Domain()`, so some internal functions that were previously added exclusively in the upgradeable variant had to be refactored: instead of overrideable [`_EIP712NameHash` and `_EIP712VersionHash`](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/release-v4.8/contracts/utils/cryptography/EIP712Upgradeable.sol#L95-L113) we now have overrideable [`_EIP712Name` and `_EIP712Version`](https://github.com/frangio/openzeppelin-contracts/blob/transpile-result/contracts/utils/cryptography/EIP712Upgradeable.sol#L137-L155).

Another addition I made, that may be controversial, is that since there are new storage variables (name and version) that have to be initialized, I added code to deal with an upgraded but accidentally uninitialized contract. [`_EIP712NameHash` and `_EIP712VersionHash`](https://github.com/frangio/openzeppelin-contracts/blob/transpile-result/contracts/utils/cryptography/EIP712Upgradeable.sol#L157-L197) will read from the old storage variables if it detects that the new ones were not initialized, and `eip712Domain()` will revert if accurate name and version values can't be returned due to uninitialized storage.

File: scripts/upgradeable/README.md [hunk 9]
```diff
@@ -0,0 +1,21 @@
+The upgradeable variant of OpenZeppelin Contracts is automatically generated from the original Solidity code. We call this process "transpilation" and it is implemented by our [Upgradeability Transpiler](https://github.com/OpenZeppelin/openzeppelin-transpiler/).
+
+When the `master` branch or `release-v*` branches are updated, the code is transpiled and pushed to [OpenZeppelin/openzeppelin-contracts-upgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable) by the `upgradeable.yml` workflow.
+
+## `transpile.sh`
+
+Applies patches and invokes the transpiler with the necessary command line flags.
```
ernestognw, line 7: Feel free to adapt this suggestion as you wish, I just feel it's not explicit enough to just say `the necessary contracts`
```suggestion
Applies patches and invokes the transpiler with necessary command line flags to include or exclude certain purpose-specific contracts
```

scripts/upgradeable/transpile-onto.sh
```diff
@@ -0,0 +1,38 @@
...
+git checkout --quiet --detach
+
+if git rev-parse -q --verify "$target"; then
+  git reset --soft "$target"
+  git checkout "$target"
+else
+  git checkout --orphan "$target"
+  if [ -n "$base" ] && git rev-parse -q --verify "$base"; then
+    git reset --soft "$base"
+  fi
```
ernestognw, line 31: I have the feeling that we should add an `else` here if there's no `base` and `target` doesn't exist. My concern is that the next check will fail because the transpiler runs `patch-apply.sh`, adding changes:

```bash
if ! git diff --quiet --cached; then
  git commit -m "Transpile $commit"
fi
```

test/utils/cryptography/EIP712.test.js
```diff
@@ -60,11 +60,11 @@ contract('EIP712', function (accounts) {
...
             const cloneDomain = { ...this.domain, verifyingContract: clone.address };
 
-            const expectedSeparator = await domainSeparator(cloneDomain);
-            expect(await clone.$_domainSeparatorV4()).to.equal(expectedSeparator);
-
             const reportedDomain = await getDomain(clone);
             expect(mapValues(reportedDomain, String)).to.be.deep.equal(mapValues(cloneDomain, String));
+
+            const expectedSeparator = await domainSeparator(cloneDomain);
+            expect(await clone.$_domainSeparatorV4()).to.equal(expectedSeparator);
```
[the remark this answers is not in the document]
frangio: Previously the first assertion was about the domain separator being as expected. It's a hash, so seeing it fail is very opaque, just that a bytes32 value is not the one you wanted. I moved that assertion second, and the first one now compares the domain fields in cleartext so the errors are more understandable. It's only related to this PR in that I ran into this difficulty while working on the EIP712 upgradeability changes.

contracts/utils/cryptography/EIP712.sol
```diff
@@ -30,7 +30,7 @@ import "../../interfaces/IERC5267.sol";
  *
  * _Available since v3.4._
  *
- * @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
+ * @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment constructor
```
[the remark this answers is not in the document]
frangio: I should remove this, I think I added it in a previous iteration.

test/utils/cryptography/EIP712.test.js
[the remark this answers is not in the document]
ernestognw: The rationale of letting a more clear error to show up first is enough. Thanks!

scripts/upgradeable/transpile.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ $build_info_num -ne 1 ]; then
+  echo "found $build_info_num relevant build info files but expected just 1"
+  exit 1
+fi
+
+# -D: delete original and excluded files
+# -b: use this build info file
+# -i: use included Initializable
+# -x: exclude all proxy contracts except Clones library
```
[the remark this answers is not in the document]
frangio: This comment is outdated. I think it was Clones first and afterwards we added UUPS, etc. I've updated the comment.

scripts/upgradeable/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## `transpile-onto.sh`
+
+```
+bash scripts/upgradeable/transpile-onto.sh <target> [base]
```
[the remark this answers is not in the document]
frangio: I was confused. Checking [docopt](http://docopt.org/) for reference and it seems to use `[<base>]` as you suggest. Changed.

hardhat/task-test-get-files.js
```diff
@@ -0,0 +1,36 @@
...
+
+  return await glob(path.join(config.paths.tests, '**/*.js'), {
+    ignore: hasProxies
+      ? []
+      : [
+          'proxy/beacon/BeaconProxy.test.js',
+          'proxy/beacon/UpgradeableBeacon.test.js',
+          'proxy/ERC1967/ERC1967Proxy.test.js',
+          'proxy/transparent/ProxyAdmin.test.js',
+          'proxy/transparent/TransparentUpgradeableProxy.behaviour.js',
```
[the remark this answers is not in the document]
frangio: I'm not sure why I excluded this initially but it doesn't seem to be necessary.

scripts/upgradeable/transpile-onto.sh
frangio: I'm not sure I understand the comment/concern.

I've pushed a commit adding some comments in this file. Let me know if those clear things up.

ernestognw: My concern is that the transpile will add changes even if there's no `target` or `base`, in other words, there's no `else` case in which we fail (`exit 1`) from the script. In such case, we'll commit without the actual transpilation and then push to the `contracts-upgradeable` with a commit not including the upgradeable contracts.

Note that it'll only happen if we don't provide `target` or `base`, which can happen due to a bug or human error changing the scripts. Right now it works so I'll suggest to just add a comment, but I wouldn't block the merge since we know it works as of now.

scripts/upgradeable/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+Applies patches and invokes the transpiler with the command line flags we need for our requirements (for example, excluding certain files).
+
+## `transpile-onto.sh`
+
+```
+bash scripts/upgradeable/transpile-onto.sh <target> [<base>]
+```
+
+Transpiles the contents of the current git branch and commits the result as a new commit on branch `<target>`. If branch `<target>` doesn't exist, it will copy the commit history of `[base]` (this is used in GitHub Actions, but is usually not necessary locally).
```
ernestognw, line 15:
```suggestion
Transpiles the contents of the current git branch and commits the result as a new commit on branch `<target>`. If branch `<target>` doesn't exist, it will copy the commit history of `[<base>]` (this is used in GitHub Actions, but is usually not necessary locally).
```

ernestognw (approved):
Just a nit, but feel free to merge if accepting the suggestion dismisses my review

ernestognw (approved):
Just a nit, but feel free to merge if accepting the suggestion dismisses my review

scripts/upgradeable/README.md
ernestognw: This is fine, I also don't think we need to be exhaustive but the original writing was not saying much 😅 thanks!

Amxx (commented):
that is a lot to review, I'm about 50% done

scripts/upgradeable/upgradeable.patch
```diff
@@ -0,0 +1,481 @@
...
++
++    /**
++     * @dev The version parameter for the EIP712 domain.
++     *
++     * NOTE: This function reads from storage by default, but can be redefined to return a constant value if gas costs
++     * are a concern.
++     */
++    function _EIP712Version() internal virtual view returns (string memory) {
++        return _version;
++    }
```
Amxx, line 394: IMO this part should be in the vanilla contract. That would reduce the patch size, and improve "compatibility" between both versions.

(or be private?)

Amxx (commented):
I'll @frangio and @ernestognw process these and merge with my implicit approval that they are going to do things right.

scripts/upgradeable/upgradeable.patch
Amxx: The argument that we don't want to ship that with the "vanila" version without having it audited doesn't really hold if we and up shipping it with the upgradeable version (without it going through the audit)

.github/workflows/upgradeable.yml
```diff
@@ -7,17 +7,24 @@ on:
...
-            -d '{ "event_type": "Update", "client_payload": { "ref": "${{ github.ref }}" } }'
+          repository: OpenZeppelin/openzeppelin-contracts-upgradeable
+          fetch-depth: 0
+          token: ${{ secrets.GH_TOKEN_UPGRADEABLE }}
+      - name: Fetch current non-upgradeable branch
+        run: |
+          git fetch "https://github.com/${{ github.repository }}.git" "$REF"
+          git checkout FETCH_HEAD
+        env:
+          REF: ${{ github.ref }}
```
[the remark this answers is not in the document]
frangio: The reason for doing it this way is to avoid command injection. I believe `github.ref` is potentially attacker controlled so if we just interpolate it into the command like you suggest it could result in command injection. Usign an environment variable prevents this attack vector.

`github.repository` is known so we don't need to do the same thing.

See https://securitylab.github.com/research/github-actions-untrusted-input for more about this.

test/utils/Checkpoints.test.js
```diff
@@ -6,137 +6,161 @@ const { batchInBlock } = require('../helpers/txpool');
...
-            expect(await getAtBlockVariant(this.mock, this.tx2.receipt.blockNumber)).to.be.bignumber.equal('2');
+            expect(
+              await this.methods.history[getAtBlockVariant](this.tx1.receipt.blockNumber - 1),
+            ).to.be.bignumber.equal('0');
+            expect(await this.methods.history[getAtBlockVariant](this.tx1.receipt.blockNumber)).to.be.bignumber.equal(
+              '1',
+            );
+            expect(await this.methods.history[getAtBlockVariant](this.tx2.receipt.blockNumber)).to.be.bignumber.equal(
+              '2',
+            );
```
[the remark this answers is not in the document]
frangio: Yeah, it's the linter... I tried a few things but I don't think the results are better.

ernestognw (approved):

PR #4043, opened 2023-02-13
Restrict ERC721Wrapper wrap by direct transfer

ernestognw:
The current version of `ERC721Wrapper` can be used to wrap tokens by directly transferring to the wrapper contract, potentially minting the wrapped token to an account other than the sender, by encoding an address in the transfer `data` payload. In combination with wallets that do not surface this `data` payload to the user, this opens the door to phishing attacks where the victim unknowingly transfers their token to an attacker who inserts their address into this invisible payload.

This PR fixes this attack vector by restricting "wrap by direct transfer" so that it can only mint the transferred token to its original owner.

Fixes LIB-690

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/beige-plants-swim.md
+++ b/.changeset/beige-plants-swim.md
@@ -1,5 +1,0 @@
----
-'openzeppelin-solidity': patch
----
-
-`ERC721Wrapper`: Restrict custom `data` encoded wrapping to trusted `this`
--- a/.changeset/beige-plants-swim.md
+++ b/.changeset/beige-plants-swim.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`ERC721Wrapper`: Restrict custom `data` encoded wrapping to trusted `this`
--- a/contracts/token/ERC721/extensions/ERC721Wrapper.sol
+++ b/contracts/token/ERC721/extensions/ERC721Wrapper.sol
@@ -27,7 +27,7 @@ abstract contract ERC721Wrapper is ERC721, ERC721Holder {
     function depositFor(address account, uint256[] memory tokenIds) public virtual returns (bool) {
         uint256 length = tokenIds.length;
         for (uint256 i = 0; i < length; ++i) {
-            underlying().safeTransferFrom(_msgSender(), address(this), tokenIds[i], abi.encodePacked(account));
+            underlying().safeTransferFrom(_msgSender(), address(this), tokenIds[i], data);
         }
 
         return true;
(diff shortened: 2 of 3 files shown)


Amxx (approved):
It mostly look good, even though that is not how I would have done it.

The only way I can see that not work is if the contract is extended, the extension does transfer of the underlying tokens (not sure why it would do that) and that operation takes arbitrary data from the user.
This is all very unlikelly IMO.

contracts/token/ERC721/extensions/ERC721Wrapper.sol
```diff
@@ -29,11 +25,9 @@ abstract contract ERC721Wrapper is ERC721, ERC721Holder {
      * @dev Allow a user to deposit underlying tokens and mint the corresponding tokenIds.
      */
     function depositFor(address account, uint256[] memory tokenIds) public virtual returns (bool) {
-        bytes memory data = abi.encodePacked(WRAPPER_ACCEPT_MAGIC, account);
-
         uint256 length = tokenIds.length;
         for (uint256 i = 0; i < length; ++i) {
-            underlying().safeTransferFrom(_msgSender(), address(this), tokenIds[i], data);
+            underlying().safeTransferFrom(_msgSender(), address(this), tokenIds[i], abi.encodePacked(account));
```
Amxx, line 30: What about keeping the existing syntax, and do 
```bytes memory data = abi.encodePacked(account);``` 

That would minimize the changes, and avoid re-encoding multiple time in the loop.

```suggestion
            underlying().safeTransferFrom(_msgSender(), address(this), tokenIds[i], data);
```

ernestognw: > avoid re-encoding multiple time in the loop.

Thanks for the heads up! I didn't consider it was inside a loop, just committed

ernestognw:
> It mostly look good, even though that is not how I would have done it.

Are there changes you think are worth reviewing for another PR? I still think it's worth proposing a simpler design if possible.

> The only way I can see that not work is if the contract is extended, the extension does transfer of the underlying tokens (not sure why it would do that) and that operation takes arbitrary data from the user. This is all very unlikelly IMO.

Do you think it's worth keeping the MAGIC value at least for identifying valid operations within the wrapper context?
In that case, an extension would've to comply with the magic value, and keeping it would also make the changes less verbose.

We definitely can keep that.

.changeset/beige-plants-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`ERC721Wrapper`: Restrict custom `data` encoded wrapping to trusted `this`
```
frangio, line 5: The changeset is not needed because the feature was never released. The user-facing change in the next release will be simply the addition of ERC721Wrapper.

frangio:
>Do you think it's worth keeping the MAGIC value

I don't think it's worth it.

frangio (commented):
The code looks good.

frangio (approved):
LGTM. The code hasn't changed since @Amxx's approval.

frangio:
Merging based on prior approval by @Amxx. Since then the code only changed according to his suggestion.

PR #4045, opened 2023-02-14
Add publishing integrity check after releasing

ernestognw:
## Description

Add an integrity check to ensure contracts were packed exactly as they're in the repository.

## Checklist

- [x] Test release on a dummy package


.github/workflows/release-cycle.yml
```diff
@@ -130,6 +130,11 @@ jobs:
         run: bash scripts/release/workflow/pack.sh
         env:
           PRERELEASE: ${{ needs.state.outputs.is_prerelease }}
+      - name: Upload tarball artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ format('{0}{1}', github.ref_name, '.tgz') }}
```
frangio, line 136: Isn't this preferable?

```suggestion
          name: ${{ github.ref_name }}.tgz
```

.github/workflows/release-cycle.yml
```diff
@@ -146,6 +151,24 @@ jobs:
           PRERELEASE: ${{ needs.state.outputs.is_prerelease }}
         with:
           script: await require('./scripts/release/workflow/github-release.js')({ github, context })
+    outputs:
+      tarball_path: ${{ steps.pack.outputs.tarball }}
```
frangio, line 155: No longer used I think.

```suggestion
```

.github/workflows/release-cycle.yml
```diff
@@ -146,6 +151,24 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: Download tarball artifact
+        id: artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ github.ref_name }}.tgz
+      - name: Check integrity
+        run: bash scripts/release/workflow/integrity-check.sh
+        env:
+          TARBALL: ${{ format('{0}/{1}{2}', steps.artifact.outputs.download-path, github.ref_name, '.tgz') }}
```
frangio, line 171: Same as above.

```suggestion
          TARBALL: ${{ steps.artifact.outputs.download-path }}/${{ github.ref_name}}.tgz
```

frangio (commented):
Looks good. Minor style comments:

ernestognw:
Action lint is failing because the `download-path` is not declared in the `artifact/download` action `.yaml`.
[The fix is already merged](https://github.com/actions/download-artifact/pull/194) but we might want to wait until it's released. Another option is to ignore the error somehow but I don't like the idea.

What do you think? @frangio

frangio:
We can use the latest commit even if it's not released. `actions/download-artifact@e9ef242655d12993efdcda9058dee2db83a2cb9b`.

.github/workflows/release-cycle.yml
```diff
@@ -142,6 +142,11 @@ jobs:
         run: bash scripts/release/workflow/pack.sh
         env:
           PRERELEASE: ${{ needs.state.outputs.is_prerelease }}
+      - name: Upload tarball artifact
+        uses: actions/upload-artifact@v3
```
frangio, line 146:
```suggestion
        uses: actions/upload-artifact@e9ef242655d12993efdcda9058dee2db83a2cb9b
```

ernestognw: I'll accept the suggestion and add a comment.
There'd be already 2 technical debt items here:

1. To remove the comment for the [Changesets Action `body` parameter PR](https://github.com/changesets/action/pull/250) when they merge it
2. A comment to use `actions/upload-artifact@v3` when it's released.

Does this sound acceptable to you?

frangio: It's not ideal but it's okay with me. We may get an automatic PR to update the action from Renovatebot.

frangio (approved):

PR #4046, opened 2023-02-14
Fix flaky timestamp tests 

frangio:
They are due to an how OpenZeppelin Test Helpers implements `time.increaseTo`. We should prefer Hardhat's own network helpers (eventually we should migrate the rest of the test suite).

I also included a small refactor of `.then` in the Governor helper.

Fixes #4045

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -26,7 +26,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       FORCE_COLOR: 1
-      GAS: true
     steps:
       - uses: actions/checkout@v3
       - name: Set up environment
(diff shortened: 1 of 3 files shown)


Amxx (dismissed):
Awesome !

frangio:
The test suite is going a lot slower than before I think. But I don't think it's due to this PR.

frangio:
Test suite took twice the time it usually takes. I'm disabling gas reporting to see if it helps.

Amxx:
> Test suite took twice the time it usually takes. I'm disabling gas reporting to see if it helps.

gas report definitelly making all tests WAY slower

frangio:
| Gas Reporter | OZ Helpers | Nomic Helpers |
|-|--|--|
| on | 20m | TBD |
| off | 10m | 22m |

Amxx:
It made sens that #d79c2ecb6d31f82650fe71fef08364c208a4abea would fix the issue of flaky timestamps, but I'm not sure how we expect the current version of the PR to do anything since the behavior for timestamps in `helper/time.js` is no longer part of the diff.

frangio:
@Amxx I'm just trying things out to debug the test time degradation.

PR #4047, opened 2023-02-14
Add linter rule to prefer await instead of .then

frangio:
I believe we need to stick to `await` instead of `.then`. I find `.then` acceptable when it's used in one-liners, but elsewhere `await` is clearly more readable.

```js
x.then(y =>
  f(y)
);

// versus

const y = await x;
f(y);
```

Anyway we can use this PR to discuss. I'm proposing a linter rule so we can automate the guideline going forward.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,6 +4,9 @@
     "eslint:recommended",
     "prettier",
   ],
+  "plugins": [
+    "avoid-then"
+  ],
   "env": {
     "es2022": true,
     "browser": true,
@@ -16,5 +19,8 @@
     "web3": "readonly",
     "extendEnvironment": "readonly",
     "expect": "readonly",
-  }
+  },
+  "rules": {
+    "avoid-then/avoid-then": "error"
+  },
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "chai": "^4.2.0",
         "eslint": "^8.30.0",
         "eslint-config-prettier": "^8.5.0",
+        "eslint-plugin-avoid-then": "github:frangio/eslint-plugin-avoid-then",
         "eth-sig-util": "^3.0.0",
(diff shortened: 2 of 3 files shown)


Amxx:
> ```js
> x.then(y =>
>   f(y)
> );
> 
> // versus
> 
> const y = await x;
> f(y);
> ```

IMO the simplest, cleanest, option is
```
x.then(f);
```

Amxx:
Also, I think we should avoid nested await

I (strongly) prefer
```js
x.then(f).then(g)
```

to

```js
g(await f(await x))
```

frangio:
> IMO the simplest, cleanest, option is
> ```js
> x.then(f);
> ```

This is missing the point. I used a small example, you need to imagine that the body (`f`) is multiple lines long. 


> ```js
> g(await f(await x))
> ```

IMO this is only slightly less readable than `g(f(x))`. Function calls execute from right to left, this is standard JavaScript. When this becomes a problem, we can define intermediate variables.

In any case, I do see some value for simple one-liners, so the linter rule in this PR allows `.then` in one-liners.

frangio:
@Amxx You can see the new linter errors in this PR.

There is no reason to prefer

```js
await this.mock.proposalVotes(this.proposal.id).then(results => {
  expect(results.forVotes).to.be.bignumber.equal(web3.utils.toWei('17'));
  expect(results.againstVotes).to.be.bignumber.equal(web3.utils.toWei('5'));
  expect(results.abstainVotes).to.be.bignumber.equal(web3.utils.toWei('2'));
});
```

instead of 

```js
const results = await this.mock.proposalVotes(this.proposal.id);
expect(results.forVotes).to.be.bignumber.equal(web3.utils.toWei('17'));
expect(results.againstVotes).to.be.bignumber.equal(web3.utils.toWei('5'));
expect(results.abstainVotes).to.be.bignumber.equal(web3.utils.toWei('2'));
```

PR #4050, opened 2023-02-16
Add comment on unchecked arithmetic (division by zero) in `Math.sol`

pcaversaccio:
As title. Let me know if this requires a Changeset entry.


Amxx:
Interesting to note that the div opcode will return 0 when dividing by 0. Using assembly without a "manual" check would be incorrect.

pcaversaccio:
After a quick chat with @Amxx I would like to quickly explain why it's worth documenting. When I first looked at the code I was concerned about where the revert comes from because of the `unchecked` block and the usage of assembly? So the underlying reason is that the `DIV` opcode returns 0 for a denominator of 0 (hello footgun :)). That's why Uniswap, for example, uses a [check](https://github.com/Uniswap/v3-core/blob/412d9b236a1e75a98568d49b1aeb21e3a1430544/contracts/libraries/FullMath.sol#L34) beforehand:

image

Maybe it's just a personal thing, but imho it's worth highlighting that Solidity does add here an additional check.

contracts/utils/math/Math.sol
```diff
@@ -67,6 +67,8 @@ library Math {
 
             // Handle non-overflow cases, 256 by 256 division.
             if (prod1 == 0) {
+                // Will revert if denominator == 0. It is not possible to disable the check for division by zero using the unchecked block.
```
frangio, line 70: I think this expresses the message better:

```suggestion
                // Will revert if denominator == 0. The surrounding unchecked block does not change this fact.
```

pcaversaccio: Agree with the wording change.

frangio (dismissed):
LGTM. Thanks!

frangio:
Expanded the wording based on feedback from @Amxx.

Amxx (approved):

PR #4053, opened 2023-02-17
Improve cancelling with GovernorTimelockControl

### LINKED ISSUE #4052: Include missing _cancel(uint256 proposal_id) in GovernorTimelockControl
**🧐 Motivation**
When using the Governor contract with the GovernorTimelockControl, one needs to override ```cancel(uint256 proposalId)``` because both Pending and Queued proposals must be cancellable. The problem is that it's not possible to invoke the inherited function ```_cancel(address[] memory targets, uint256[] memory values, bytes[] memory calldatas, bytes32 descriptionHash``` from inside the overidden public cancel function.
    )

**📝 Details**
A solution would be to let the GovernorTimelockControl override both the internal _cancel functions it inherits from Governor, in which case the four-argument function simply forwards to the single-argument function (as is the case in the Governor contract). The logic currently inside _cancel would move to the new single-argument function.

The downside is that it's a breaking change for anyone using the GovernorTimelockControl.
I will submit a PR with a link to this issue.

sebsylvester:
Fixes #4052

The update adds an overridden ```_cancel(uint256 proposal_id) internal``` to GovernorTimelockControl, making it accessible from an overridden ```cancel(uint256 proposal_id) public```.

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/six-oranges-repair.md
+++ b/.changeset/six-oranges-repair.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+Include \_cancel(uint256 proposal_id) in GovernorTimelockControl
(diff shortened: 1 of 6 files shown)


Amxx:
Hello @sebsylvester 

Issue is more complex than that, we'll need to have some discussion about how to fix that (in #4052). 

Closing for now. We'll reopen (or do a different PR) when we have consensus on how to fix.

PR #4055, opened 2023-02-17
Fix ERC4906 compliance of ERC721URIStorage

Amxx:
As pointed out by @pcaversaccio in [this post](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4012#issuecomment-1434665993)

#4012 missed the ERC165 part of ERC4906. Its unclear how this value was produced, and why an ERC that doesn't include any function (only event) would have an ERC165 interfaceId.

Still, this PR makes `ERC721URIStorage` compliant with the ERC165 part of ERC4906.

#### PR Checklist

- [x] Tests

### DIFF
--- a/test/utils/introspection/SupportsInterface.behavior.js
+++ b/test/utils/introspection/SupportsInterface.behavior.js
@@ -100,20 +100,22 @@ function shouldSupportInterfaces(interfaces = []) {
 
     it('supportsInterface uses less than 30k gas', async function () {
       for (const k of interfaces) {
-        const interfaceId = INTERFACE_IDS[k];
+        const interfaceId = INTERFACE_IDS[k] ?? k;
         expect(await this.contractUnderTest.supportsInterface.estimateGas(interfaceId)).to.be.lte(30000);
       }
     });
 
     it('all interfaces are reported as supported', async function () {
       for (const k of interfaces) {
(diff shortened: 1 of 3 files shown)


test/utils/introspection/SupportsInterface.behavior.js
```diff
@@ -100,20 +100,22 @@ function shouldSupportInterfaces(interfaces = []) {
...
-        const interfaceId = INTERFACE_IDS[k];
+        const interfaceId = INTERFACE_IDS[k] ?? k;
         expect(await this.contractUnderTest.supportsInterface(interfaceId)).to.equal(true);
       }
     });
 
     it('all interface functions are in ABI', async function () {
       for (const k of interfaces) {
+        // skip interfaces for which we don't have a function list
+        if (INTERFACES[k] == undefined) continue;
```
frangio, line 118:
```suggestion
        if (INTERFACES[k] === undefined) continue;
```

frangio, line 118:
```suggestion
        if (INTERFACES[k] === undefined) continue;
```

frangio (approved):

pcaversaccio:
> Its unclear how this value was produced

Well, the `0x49064906` comes from the EIP number lol. Twice `4906` simply.

PR #4060, opened 2023-02-21
Fix comment in `ERC4626` on decimals default

pcaversaccio:
As title. No `CHANGELOG` entry is added since I consider this a simple correction.


Amxx:
Thank you  @pcaversaccio for spotting that. 

Considering the recent addition of `_decimalsOffset`, I think the change should be deeper than what you propose.

pcaversaccio:
You are right, but I wasn't sure whether you want to duplicate the content already written above in the contract and details `_decimalsOffset`:

```solidity
 * Since v4.9, this implementation uses virtual assets and shares to mitigate that risk. The `_decimalsOffset()`
 * corresponds to an offset in the decimal representation between the underlying asset's decimals and the vault
 * decimals. This offset also determines the rate of virtual shares to virtual assets in the vault, which itself
 * determines the initial exchange rate. While not fully preventing the attack, analysis shows that the default offset
 * (0) makes it non-profitable, as a result of the value being captured by the virtual shares (out of the attacker's
 * donation) matching the attacker's expected gains. With a larger offset, the attack becomes orders of magnitude more
 * expensive than it is profitable. More details about the underlying math can be found
 * xref:erc4626.adoc#inflation-attack[here].
```

To make the process as simple as possible, would like to commit what you have exactly in mind to this PR?

Amxx:
> To make the process as simple as possible, would like to commit what you have exactly in mind to this PR?

I don't have anything specific in mind yet. I'll have to think about it!

pcaversaccio:
Sure, actually I think the comment in `_convertToShares` is also not anymore correct:

```
/**
 * Will revert if assets > 0, totalSupply > 0 and totalAssets = 0. That corresponds to a case where any asset
 * would represent an infinite amount of shares.
 */
```

We now got:
```solidity
return assets.mulDiv(totalSupply() + 10 ** _decimalsOffset(), totalAssets() + 1, rounding);
```

The denominator will never be `0`, and there will not revert. So it will return simply `assets * totalSupply() + 10 ** _decimalsOffset()`. Theoretically, it could revert due to the checked arithmetic (addition), but the comment doesn't specify this.

Amxx:
Let me know if you agree with the current formulation for decimals.

pcaversaccio (commented):
LGTM - quickly fixed a typo.

Amxx (approved):
LGTM, we also need @frangio approval

frangio (approved):
Looks good to me! Thanks!

PR #4063, opened 2023-02-21
Implement `0x00` version of EIP-191 in ECDSA Library

### LINKED ISSUE #3772: Implement `0x00` version of EIP191 in ECDSA Library
### 🧐 Motivation
In the current `ECDSA` library, there is support to construct a message to sign according to the EIP191 Standard with the following:
- The version `0x45` (E) with `toEthSignedMessage(..)` function 
- The version `0x01` with the `toTypedDataHash(..)` function. 
But we are missing the `0x00` version.

Screen Shot 2022-10-19 at 10 17 27 PM

### 📝 Details
After looking at [this closed issue](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1810), saw a comment from @frangio  stating that `0x00` version is not that used, which is true because people are misusing the standard, they are using the `0x45` version for everything: signing message for offchain verification, for smart contract execution based on signatures, which is dangerous.

As people could be easily tricked to sign a normal message, we should have a different mechanism, then different handling for execution based on signatures, and that's why we should make people use the `0x00` version for this case. + some projects are starting to use it like [xenium](https://github.com/Arachnid/xenium) and the [lsp-smart-contracts](https://github.com/lukso-network/lsp-smart-contracts) (planning to use it). 

I am suggesting to add a new function `toDataWithIntendedValidator(...)` to be compatible with the version `0x00` taking 2 parameters, `<address validator>` and `<bytes dataToSign>`.

The function would be:

```solidity
function toDataWithIntendedValidator(address validator, bytes memory dataToSign) internal pure returns (bytes32) {
        return keccak256(abi.encodePacked("\x19\x00", validator, dataToSign));
}
```

If you're okay with it I am happy to open the PR, otherwise you can close the issue 😄

YamenMerhi:
## What does this PR introduce ?

- Add a new function `toDataWithIntendedValidator(...)` to be compatible with the version `0x00` of EIP-191 taking 2 parameters `<address validator>` and `<bytes dataToSign>`.

Fixes #3772 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/small-cars-appear.md
+++ b/.changeset/small-cars-appear.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-Implement 0x00 version of EIP-191 in ECDSA Library
+`ECDSA`: Add a function `toDataWithIntendedValidatorHash` that encodes data with version 0x00 following EIP-191.
--- a/.changeset/small-cars-appear.md
+++ b/.changeset/small-cars-appear.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Implement 0x00 version of EIP-191 in ECDSA Library
--- a/test/helpers/sign.js
+++ b/test/helpers/sign.js
@@ -5,6 +5,21 @@ function toEthSignedMessageHash(messageHex) {
 }
 
 /**
+ * Create a signed data with intended validator according to the version 0 of EIP-191
+ * @param validatorAddress The address of the validator
+ * @param dataHex The data to be concatened with the prefix and signed
+ */
+function toDataWithIntendedValidatorHash(validatorAddress, dataHex) {
+  const validatorBuffer = Buffer.from(web3.utils.hexToBytes(validatorAddress));
+  const dataBuffer = Buffer.from(web3.utils.hexToBytes(dataHex));
+  const preambleBuffer = Buffer.from('\x19');
+  const versionBuffer = Buffer.from('\x00');
+  const ethMessage = Buffer.concat([preambleBuffer, versionBuffer, validatorBuffer, dataBuffer]);
+
+  return web3.utils.sha3(ethMessage);
+}
+
+/**
  * Create a signer between a contract and a signer for a voucher of method, args, and redeemer
  * Note that `method` is the web3 method, not the truffle-contract method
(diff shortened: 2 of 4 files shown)


.changeset/small-cars-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Implement 0x00 version of EIP-191 in ECDSA Library
```
Amxx, line 5:
```suggestion
`ECDSA`: Add a function `toDataWithIntendedValidatorHash` that encodes data with version 0x00 following EIP-191.
```

test/helpers/sign.js
```diff
@@ -4,6 +4,21 @@ function toEthSignedMessageHash(messageHex) {
   return web3.utils.sha3(Buffer.concat([prefix, messageBuffer]));
 }
 
+/**
+ * Create a signed data with intended validator according to the version 0 of EIP-191
+ * @param validatorAddress The address of the validator
+ * @param dataHex The data to be concatened with the prefix and signed
```
Amxx, line 10:
```suggestion
 * @param dataHex The data to be concatenated with the prefix and signed
```

test/utils/cryptography/ECDSA.test.js
```diff
@@ -8,6 +8,7 @@ const ECDSA = artifacts.require('$ECDSA');
 const TEST_MESSAGE = web3.utils.sha3('OpenZeppelin');
 const WRONG_MESSAGE = web3.utils.sha3('Nope');
 const NON_HASH_MESSAGE = '0x' + Buffer.from('abcd').toString('hex');
+const RANDOM_ADDRESS = '0xA9d8C83D404d3397aDE08321E7551c8B36dbF4Ab';
```
Amxx, line 11: lets really make it random.

```suggestion
const RANDOM_ADDRESS = web3.utils.toChecksumAddress(web3.utils.randomHex(20));
```

Internal note: we have this defined as a function in `test/crosschain/CrossChainEnabled.test.js`, maybe we should make it a helper.

Amxx (approved):
LGTM,

we also need @frangio 's review

frangio (approved):
LGTM. Thanks!

PR #4065, opened 2023-02-22
Make `MinimalForwarder` production-ready

### LINKED ISSUE #3664: Loss of unused Ether in meta-transaction
In the `MinimalForwarder.execute(ForwardRequest req, bytes signature)` payable function, the `msg.value`   may contain more Ether than specified in `req.value`. In this case (since only `req.value` is transferred further), the unused amount of Ether will remain on the contract address.

### LINKED ISSUE #3884: Consider removing MinimalForwarder
Even though the documentation states that this contract is not suitable for production use, that message isn't getting across and I believe we should just remove the contract entirely.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3d7a93876a2e5e1d7fe29b5a0e96e222afdc4cfa/contracts/metatx/MinimalForwarder.sol#L12-L15

ernestognw:
> **Warning**
> This PR is currently on hold, meaning it's not considered yet safe to use any of the updates here until merged. 

Fixes #3884
Fixes #3664

Currently, `MinimalForwarder` is not considered production-ready since it was designed mainly for testing purposes in the library. However, it's not that far from a MinimalForwarder that's something we may recommend.

The new MinimalForwarder should guarantee:

- Backwards compatibility
- Security (and thorough documentation about it)
- A deadline for expiring transactions

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/metatx/MinimalForwarder.sol
+++ b/contracts/metatx/MinimalForwarder.sol
@@ -5,16 +5,21 @@ pragma solidity ^0.8.19;
 
 import "../utils/cryptography/ECDSA.sol";
 import "../utils/cryptography/EIP712.sol";
+import "../utils/Nonces.sol";
 
 /**
- * @dev Simple minimal forwarder to be used together with an ERC2771 compatible contract. See {ERC2771Context}.
+ * @dev A minimal implementation of a production-ready forwarder compatible with ERC2771 contracts.
  *
- * MinimalForwarder is mainly meant for testing, as it is missing features to be a good production-ready forwarder. This
- * contract does not intend to have all the properties that are needed for a sound forwarding system. A fully
- * functioning forwarding system with good properties requires more complexity. We suggest you look at other projects
- * such as the GSN which do have the goal of building a system like that.
+ * This forwarder operates on forward request that include:
+ * * `from`: An address to operate on behalf of. It is required to be equal to `msg.sender`.
+ * * `to`: An address destination to call within the request.
+ * * `value`: The amount of ETH to attach within the requested call.
+ * * `gas`: The amount of gas limit that will be forwarded within the requested call.
+ * * `nonce`: A unique transaction ordering identifier to avoid replayability and request invalidation.
+ * * `deadline`: A timestamp after which the request is not executable anymore.
+ * * `data`: Encoded `msg.calldata` to send within the requested call.
  */
-contract MinimalForwarder is EIP712 {
+contract MinimalForwarder is EIP712, Nonces {
     using ECDSA for bytes32;
 
     struct ForwardRequest {
@@ -23,13 +28,14 @@ contract MinimalForwarder is EIP712 {
         uint256 value;
         uint256 gas;
         uint256 nonce;
+        uint48 deadline;
         bytes data;
     }
 
-    bytes32 private constant _TYPEHASH =
-        keccak256("ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data)");
-
-    mapping(address => uint256) private _nonces;
+    bytes32 private constant _FORWARD_REQUEST_TYPEHASH =
+        keccak256(
+            "ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,uint48 deadline,bytes data)"
+        );
 
     /**
      * @dev The request `from` doesn't match with the recovered `signer`.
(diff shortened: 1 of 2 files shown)


contracts/metatx/MinimalForwarder.sol
```diff
@@ -26,29 +21,64 @@ contract MinimalForwarder is EIP712 {
...
+        uint256 gas;
+        uint256 nonce;
+        bytes data;
+    }
+
+    bytes32 private constant _FORWARD_REQUEST_TYPEHASH =
         keccak256("ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data)");
+    bytes32 private constant _EXPIRING_FORWARD_REQUEST_TYPEHASH =
+        keccak256(
+            "ExpiringForwardRequest(uint48 deadline,address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data)"
```
Amxx, line 38: why put the deadline as the first argument? I would personnaly put it after the nonce.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -26,29 +21,64 @@ contract MinimalForwarder is EIP712 {
...
+        );
 
     mapping(address => uint256) private _nonces;
 
-    constructor() EIP712("MinimalForwarder", "0.0.1") {}
+    constructor() EIP712(name(), version()) {}
+
+    function version() public view virtual returns (string memory) {
```
Amxx, line 45: Why do we have this function ?

contracts/metatx/MinimalForwarder.sol
```diff
@@ -26,29 +21,64 @@ contract MinimalForwarder is EIP712 {
...
+        return "0.0.1";
+    }
 
-    function getNonce(address from) public view returns (uint256) {
+    function name() public view virtual returns (string memory) {
```
Amxx, line 49: Why do we have this function ?

contracts/metatx/MinimalForwarder.sol
```diff
@@ -26,29 +21,64 @@ contract MinimalForwarder is EIP712 {
...
         keccak256("ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data)");
+    bytes32 private constant _EXPIRING_FORWARD_REQUEST_TYPEHASH =
+        keccak256(
+            "ExpiringForwardRequest(uint48 deadline,address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data)"
+        );
 
     mapping(address => uint256) private _nonces;
 
-    constructor() EIP712("MinimalForwarder", "0.0.1") {}
+    constructor() EIP712(name(), version()) {}
```
Amxx, line 43: why change from `constructor() EIP712("MinimalForwarder", "0.0.1") {}` ?

contracts/metatx/MinimalForwarder.sol
```diff
@@ -26,29 +21,64 @@ contract MinimalForwarder is EIP712 {
...
+    function getNonce(address from) public view virtual returns (uint256) {
         return _nonces[from];
     }
 
-    function verify(ForwardRequest calldata req, bytes calldata signature) public view returns (bool) {
-        address signer = _hashTypedDataV4(
-            keccak256(abi.encode(_TYPEHASH, req.from, req.to, req.value, req.gas, req.nonce, keccak256(req.data)))
-        ).recover(signature);
-        return _nonces[req.from] == req.nonce && signer == req.from;
+    function verify(ForwardRequest calldata req, bytes calldata signature) public view virtual returns (bool) {
```
Amxx, line 57: That is not what we agreed. If we do that old signature are no longer valid. We need to continue verifying `ForwardRequest` using the `ForwardRequest` type.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -69,4 +99,51 @@ contract MinimalForwarder is EIP712 {
...
+
+    function _verify(ExpiringForwardRequest memory req, bytes memory signature) public view virtual returns (bool) {
+        bool expired = block.timestamp >= req.deadline;
+        return
+            getNonce(req.from) == req.nonce && // Signed nonce corresponds with current nonce
+            _signer(req, signature) == req.from && // From is signer
+            !expired;
+    }
+
+    function _signer(ForwardRequest memory req, bytes memory signature) private view returns (address) {
```
Amxx, line 111: this function is currently unused, because you rewrite everything as `ExpiringForwardRequest`.

test/metatx/MinimalForwarder.test.js
```diff
@@ -55,8 +61,9 @@ contract('MinimalForwarder', function (accounts) {
           expect(await this.forwarder.getNonce(this.req.from)).to.be.bignumber.equal(web3.utils.toBN(this.req.nonce));
         });
 
-        it('success', async function () {
-          expect(await this.forwarder.verify(this.req, this.sign())).to.be.equal(true);
+        it.only('success', async function () {
```
Amxx, line 64: Its ok to change to `it.only` for local tests, but that should never be commited.

@frangio can we add a rule so that CI detects that if we are not carefull?

Amxx (commented):
Left some comments

test/metatx/MinimalForwarder.test.js
frangio: Keeping `.only` will cause coverage checks to fail, and we will notice it drop to like 1%.

contracts/metatx/MinimalForwarder.sol
frangio: Customizing EIP712 version, similar to Governor:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7b3e7b7055d6e1226d4920ca7d7c446c0c0bfff5/contracts/governance/Governor.sol#L85

I can't see immediate reasons why someone would override it, but I feel it makes sense to leave the option open? What do you think?

contracts/metatx/MinimalForwarder.sol
frangio: I believe the reason was to customize the EIP712 name, but here I don't agree that it makes sense to make it customizable. It should be fine to make it always `"MinimalForwarder"`.

contracts/metatx/MinimalForwarder.sol
frangio: Replied in your comments about `version()` and `name()`.

contracts/metatx/MinimalForwarder.sol
ernestognw: Made sense to me to pack them this way. It's not a bid deal since it's calldata, but I thought it was better to put the `uint48` next to an `address`

contracts/metatx/MinimalForwarder.sol
ernestognw: I can't neither think about a reason why someone would need it, but I'd say the same logic of why we set `virtual` in public functions should apply here.

"Why would someone override `version` and `name`?" is kinda the same question as "Why would someone override X function?" from my perspective.

No strong opinion if we remove it.

contracts/metatx/MinimalForwarder.sol
ernestognw: I was exploring how to arrange them together since we need to keep backward compatibility. I'll solve this thread since I feel the focus is on the incorrect verification I pushed as a draft last night and that's solved.

Please open another thread if you think there's something to discuss related to the interface.

contracts/metatx/MinimalForwarder.sol
ernestognw: Used now.

test/metatx/MinimalForwarder.test.js
ernestognw: Solving since the `only` was removed and the coverage check is already there.

contracts/metatx/MinimalForwarder.sol
ernestognw: Same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4065#discussion_r1115184979

contracts/metatx/MinimalForwarder.sol
ernestognw: Same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4065#discussion_r1115184979

contracts/metatx/MinimalForwarder.sol
Amxx: What I am saying is, by forcing the verification of `ForwardRequest` to be done using the verification of `ExpiringForwardRequest` (which include a different hashing mechanism) you are actually breaking backward compatibility. Not keeping it.

Amxx: It is **not** ok to verify everything using the `ExpiringForwardRequest` typed struct. We need to keep the verfication of "old" `ForwardRequest` using the existing signature scheme.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -69,4 +99,51 @@ contract MinimalForwarder is EIP712 {
 
         return (success, returndata);
     }
+
+    function _verify(ExpiringForwardRequest memory req, address signer) private view returns (bool) {
+        bool expired = block.timestamp >= req.deadline;
+        return
+            getNonce(req.from) == req.nonce && // Signed nonce corresponds with current nonce
+            signer == req.from && // From is signer
+            !expired;
```
Amxx, line 108:
```suggestion
        return
            getNonce(req.from) == req.nonce && // Signed nonce corresponds with current nonce
            signer == req.from && // From is signer
            block.timestamp < req.deadline;
```

contracts/metatx/MinimalForwarder.sol
frangio: I believe ABI encoding does not pack values. Every value takes up 32 bytes.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -69,4 +99,51 @@ contract MinimalForwarder is EIP712 {
 
         return (success, returndata);
     }
+
+    function _verify(ExpiringForwardRequest memory req, address signer) private view returns (bool) {
```
frangio, line 103: Note that we're missing `require(msg.value == req.value)`.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -26,29 +21,64 @@ contract MinimalForwarder is EIP712 {
...
+
+    function execute(
+        ExpiringForwardRequest calldata req,
+        bytes calldata signature
+    ) public payable virtual returns (bool, bytes memory) {
+        require(verify(req, signature), "MinimalForwarder: signature does not match request");
+        return _execute(req);
+    }
+
+    function _execute(ExpiringForwardRequest memory req) private returns (bool, bytes memory) {
```
MerlinEgalite, line 81: Is the fact that `_execute` is not reverting if the call fails intended and left to the contract inheriting from `MinimalForwarder`?

If yes and given that this version will be considered as production ready, it would be nice to add a warning about this to avoid silent failure on the implementation side. Else, I think we should revert while bubbling up the error.

Amxx: Yes, its an intended behavior.

This PR is on hold. The code it contains should not be seen as a production-ready version of the forwarder. For sure we will want to challenge that design choice (IMO there are good reason for this behavior), and document it.

MerlinEgalite: Clear, thx for the answer.

ernestognw: ➕  to @Amxx explanation.

I'll leave this open and I added a note at the top of the PR indicating this is currently on hold.

PR #4066, opened 2023-02-22
Mark ERC777 and ERC1820 as deprecated

### LINKED ISSUE #2620: Deprecate ERC777 implementation

Amxx:
Fixes #2620 
Fixes LIB-640

#### PR Checklist

- [x] Documentation
  - [x] direct changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 
 - `ERC20Permit`: Added the file `IERC20Permit.sol` and `ERC20Permit.sol` and deprecated `draft-IERC20Permit.sol` and `draft-ERC20Permit.sol` since [EIP-2612](https://eips.ethereum.org/EIPS/eip-2612) is no longer a Draft. Developers are encouraged to update their imports. ([#3793](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3793))
 - `Timers`: The `Timers` library is now deprecated and will be removed in the next major release. ([#4062](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4062))
-- `ERC777`: The `ERC777` token standard is no longer supported by OpenZeppelin. Our implementation is now deprecated and will be removed in the next major release. The corresponding stardard interfaces remain available. ([#](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/))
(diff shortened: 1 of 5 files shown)


frangio (approved):

ThisCoolDude:
ERC777 token standard is longer supported by OpenZeppelin. Our implementation is now deprecated and will be removed in the next major release. The corresponding standard interfaces remain available. Is this only for now or forever. 

There isn't much information on why it's being deprecated

ThisCoolDude:
ERC777 token standard is longer supported by OpenZeppelin. Our implementation is now deprecated and will be removed in the next major release. The corresponding standard interfaces remain available. Is this only for now or forever. 

There isn't much information on why it's being deprecated

frangio:
@ThisCoolDude Please refer to the extended discussion at #2620.

PR #4073, opened 2023-02-24
Document "duplicate" proposal storing in GovernorCompatibilityBravo

Amxx:
Fixes #4073
Fixes LIB-625

#### PR Checklist

- [x] Documentation


contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -60,7 +62,13 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
     }
 
     /**
-     * @dev See {IGovernorCompatibilityBravo-propose}.
+     * @dev Alternative interface with separate function signatures. Stores the full proposal and fallback to the
```
frangio, line 65: I'm not sure this should be user-facing docs. I think it should be more like a comment inside the function, would you agree?

frangio (approved):

PR #4079, opened 2023-02-28
Make AccessControlDefaultAdminRules delay configurable

ernestognw:
Makes the delay of transferring the `DEFAULT_ADMIN_ROLE` configurable even after deployment.
It follows the next rules:

1. The delay can only be changed if there's no default admin transfer in between to avoid confusion on how it affects a scheduled default admin change
2. The delay can be canceled or set again
3. The delay change is effective after a time that is at least the previous delay set, so that the original delay guarantee is preserved
4. If a delay has been changed, a default admin transfer can't start until the scheduled change is met
5. The delay change is accepted and applied on the next `beginAdminTransfer`

#### PR Checklist

- [x] Tests
- [x] Documentation


ernestognw:
## Notes

###`schedule` naming
I changed `delayedUntil` to scheduled after several conversations and not finding a better alternative. Still open to suggestions but it was needed with this PR to avoid a naming such as `delayChangeDelayedUntil`.

Amxx (commented):
I'll need to review this more in depth, but my first point is that the data packing might not be optimal. We can put two uint48 and one address in the same slot. Right now we have 2 slots:

- Slot 0
```solidity
uint48 _curentDefaultAdminDelay
uint48 _pendingDefaultAdminDelay
address _currentDefaultAdmin
```
- slot 1
```solidity
address _pendingDefaultAdmin
uint48 _defaultAdminTransferSchedule
uint48 _defaultAdminDelayChangeSchedule
```

when you reset using, `_resetDefaultAdminDelayChange` you write to the 2 slots. I think this should be avoided if possible.
We should probably lay out all the operation's read and write, and think what is the most effective layout

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -190,15 +220,27 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+
+        uint48 currentDefaultAdminDelay = defaultAdminDelay();
+
+        // Schedules the current delay if the delay is increased.
+        // Otherwise, wait the difference between current and new delay to guarantee the delay change schedule + a default admin change
+        // is effectively the current delay. For example, if delay is reduced from 10 days to 3 days, it's needed to wait 7 days
+        // before starting the new 3 days delayed transfer suming up to 10 days, which is the current delay.
+        uint48 changeDelay = newDefaultAdminDelay >= currentDefaultAdminDelay
+            ? newDefaultAdminDelay
+            : currentDefaultAdminDelay - newDefaultAdminDelay; // This is 1 in worst-case scenario. Can't be 0.
```
frangio, line 238: What does it matter if it's 0?

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -190,15 +220,27 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
-        _resetDefaultAdminTransfer();
+    function _scheduleDefaultAdminDelayChange(uint48 newDefaultAdminDelay) internal virtual {
+        require(defaultAdminTransferSchedule() != 0, "AccessControl: default admin transfer pending");
+
+        uint48 currentDefaultAdminDelay = defaultAdminDelay();
+
+        // Schedules the current delay if the delay is increased.
+        // Otherwise, wait the difference between current and new delay to guarantee the delay change schedule + a default admin change
+        // is effectively the current delay. For example, if delay is reduced from 10 days to 3 days, it's needed to wait 7 days
+        // before starting the new 3 days delayed transfer suming up to 10 days, which is the current delay.
```
frangio, line 235:
```suggestion
        // before starting the new 3 days delayed transfer summing up to 10 days, which is the current delay.
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -190,15 +220,27 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+        uint48 changeDelay = newDefaultAdminDelay >= currentDefaultAdminDelay
+            ? newDefaultAdminDelay
```
frangio, line 237: This seems like a mistake. It should be the current delay, no?
```suggestion
            ? currentDefaultAdminDelay
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -68,7 +70,21 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @inheritdoc IAccessControlDefaultAdminRules
      */
     function defaultAdminDelay() public view virtual returns (uint48) {
-        return _defaultAdminDelay;
+        return _curentDefaultAdminDelay;
```
frangio, line 73: The "pending" delay becomes the "current" delay automatically (in `_beginDefaultAdminTransfer`).

This function returns the current delay even if the pending delay has already gone into effect. This presents a confusing view of the contract state, but most importantly it could mislead about the rules of the contract.

This pattern where some change is supposed to happen automatically is a little difficult because you can't encapsulate it in an internal function that gets the state and updates it, because you wouldn't be able to use it in a `view` context. But we should be able to find a clean way to implement it.

ernestognw:
> I'll need to review this more in depth, but my first point is that the data packing might not be optimal.

I see what you mean! Thanks for the clarity. I'll also review it in depth once we settle on the design.
Currently no strong opinion, but definitely will refactor it if we find a better layout.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: > Why does it matter if it's 0? 

It matters because we don't know what's going to be used for. Since there's a `pendingDefaultAdminDelay()` getter, someone implementing on top of this may rely on that not being 0.

I'd rather keep it in 1 at least to avoid accidental allowance of bypassing the `block.timestamp` in the same transaction.

Let me know if this shouldn't even be a concern.

.changeset/tall-hotels-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`AccessControlDefaultAdminRules`: Make the `defaultAdminDelay` configurable after construction.
```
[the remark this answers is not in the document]
ernestognw: Removed

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: The mistake was in the comments. I just updated. Let me explain the rationale:

If a user sets an accidental high delay and wait for only the previous delay, there's a risk if they begin an admin transfer and got locked. This is a way to ensure they didn't set a high value, because they will always need to wait for it, and it's also cancellable

frangio:
> once we settle on the design.

What are the remaining design questions?

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: We agreed to make the change automatic for the `defaultAdminDelay()` and the `pendingAdminDelay()`. I just implemented it in the last commit, however, [I added this warning.](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079/files#diff-f3b8263b4d61197159b628b24b617c1b96139f713bdf96704e3cb30de436e4f0R32)

I think the admin's power to cancel the delay if it hasn't been written to `_currentDefaultAdminDelay` is non-trivial, what do you think?

ernestognw:
> > once we settle on the design.
> 
> What are the remaining design questions?

For the record, the remaining question was if making the `pendingDefaultAdminDelay()` and `defaultAdminDelay()` change automatically.

We agreed to do so.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -190,15 +223,27 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+
+        uint48 currentDefaultAdminDelay = defaultAdminDelay();
+
+        // Schedules the new delay if the delay is increased, this is done so the user can't accidentally set a high new delay.
+        // Otherwise, wait the difference between current and new delay to guarantee the delay change schedule + a default admin change
+        // is effectively the current delay. For example, if delay is reduced from 10 days to 3 days, it's needed to wait 7 days
+        // before starting the new 3 days delayed transfer summing up to 10 days, which is the current delay.
+        uint48 changeDelay = newDefaultAdminDelay >= currentDefaultAdminDelay
+            ? newDefaultAdminDelay
+            : currentDefaultAdminDelay - newDefaultAdminDelay; // This is 1 in worst-case scenario. Can't be 0.
```
frangio, line 241:
```suggestion
        uint48 changeDelay = newDefaultAdminDelay > currentDefaultAdminDelay
            ? currentDefaultAdminDelay
            : currentDefaultAdminDelay - newDefaultAdminDelay;
```

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Fixed in [this commit](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079/commits/ff7a4d63c9f77ad45ec8c29e9f9438370045ae83)

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: After discussing, we decided to tolerate the 0.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: We decided to avoid another delay change to start or cancel the previous one if the schedule had already passed. In this way, it's guaranteed that the value will be kept when `defaultAdminDelay` is toggled.

ernestognw: Implemented [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079/commits/e99b5a41419aa6aa45b2d83ec957b586ee0c0bcd)

ernestognw:
> I'll need to review this more in depth, but my first point is that the data packing might not be optimal. We can put two uint48 and one address in the same slot. Right now we have 2 slots:
> 
> * Slot 0
> 
> ```solidity
> uint48 _curentDefaultAdminDelay
> uint48 _pendingDefaultAdminDelay
> address _currentDefaultAdmin
> ```
> 
> * slot 1
> 
> ```solidity
> address _pendingDefaultAdmin
> uint48 _defaultAdminTransferSchedule
> uint48 _defaultAdminDelayChangeSchedule
> ```
> 
> when you reset using, `_resetDefaultAdminDelayChange` you write to the 2 slots. I think this should be avoided if possible. We should probably lay out all the operation's read and write, and think what is the most effective layout

Following this comment, I reordered the variables in this way:

```solidity
address private _currentDefaultAdmin;

// Delay change
uint48 private _pendingDefaultAdminDelay;
uint48 private _defaultAdminDelayChangeSchedule; // 0 == unset

uint48 private _currentDefaultAdminDelay;

// Admin transfer
address private _pendingDefaultAdmin;
uint48 private _defaultAdminTransferSchedule; // 0 == unset
```

They're still packed but prioritizing those variables together in the `_reset*` operations.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -222,19 +303,26 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
     }
 
     /**
-     * @dev Checks if a {defaultAdminTransferDelayedUntil} has been set and passed.
+     * @dev Checks if a {defaultAdminTransferSchedule} has been set and passed.
      */
-    function _hasDefaultAdminTransferDelayPassed() private view returns (bool) {
-        uint48 delayedUntil = defaultAdminTransferDelayedUntil();
-        return delayedUntil > 0 && delayedUntil < block.timestamp;
+    function _hasScheduledPassed(uint48 schedule) private view returns (bool) {
```
frangio, line 325: There is a typo in `Scheduled`, it should say `Schedule`, I believe.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -190,15 +234,31 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
-    function _acceptDefaultAdminTransfer() internal virtual {
-        require(_hasDefaultAdminTransferDelayPassed(), "AccessControl: transfer delay not passed");
-        _revokeRole(DEFAULT_ADMIN_ROLE, defaultAdmin());
-        _grantRole(DEFAULT_ADMIN_ROLE, pendingDefaultAdmin());
-        _resetDefaultAdminTransfer();
+    function _beginDefaultAdminDelayChange(uint48 newDefaultAdminDelay) internal virtual {
+        require(
+            !_hasScheduledPassed(defaultAdminDelayChangeSchedule()),
+            "AccessControl: can't change past scheduled change"
+        );
```
frangio, line 245: `defaultAdminDelayChangeSchedule` is only cleared on `_reset`, `_reset` is only called on `cancel`. It seems like it is only possible to change the delay once.

frangio:
Apologies for bringing this up again, but I remain pretty concerned about the naming in this contract and I think we should make a stronger effort to simplify it. We also need to remove the terms "defaultAdmin" from where it is evident from context.

Here are my suggestions:

```diff
-    uint48 private _currentDefaultAdminDelay;
+    uint48 private _currentDelay;
 
-    uint48 private _pendingDefaultAdminDelay;
+    uint48 private _pendingDelay;
 
-    uint48 private _defaultAdminDelayChangeSchedule;
+    uint48 private _pendingDelaySchedule;
 
     address private _currentDefaultAdmin; // (no change)
     address private _pendingDefaultAdmin;

-    uint48 private _defaultAdminTransferSchedule;
+    uint48 private _pendingDefaultAdminSchedule;
 
-    event DefaultAdminDelayChangeStarted(uint48 indexed newDefaultAdminDelay, uint48 defaultAdminDelaySchedule);
+    event DefaultAdminDelayChangeScheduled(uint48 newDelay, uint48 effectSchedule);
 
     // Renaming RoleChange -> Transfer
-    event DefaultAdminRoleChangeStarted(address indexed newDefaultAdmin, uint48 defaultAdminTransferSchedule);
+    event DefaultAdminTransferScheduled(address indexed newDefaultAdmin, uint48 acceptSchedule);
 
     // Here we need "defaultAdmin" because the function will appear among the full contract interface
-    function increasedDelayWait() external view returns (uint48);
+    function defaultAdminDelayIncreaseWait() external view returns (uint48);
 
-    function beginDefaultAdminDelayChange(uint48 newDefaultAdminDelay) external;
+    function changeDefaultAdminDelay(uint48 newDelay) external;
 
-    function cancelDefaultAdminDelayChange() external;
+    function rollbackDefaultAdminDelay() external;
```

Additionally, I would recommend merging each pair of "pendingX" and "XSchedule" getters into a single function with 2 return values. This is a bit of a trick to avoid having to come up with a name... but I also think it makes sense. For example:

```diff
-    function pendingDefaultAdmin() public view virtual returns (address);
-    function defaultAdminTransferSchedule() public view virtual returns (uint48);
+    function pendingDefaultAdmin() external view returns (uint48 address, uint48 acceptSchedule);
```

This also applies to the delay change.

I've also realized we are missing events for cancellations. We could add the following:

```diff
+    event DefaultAdminDelayChangeCancelled(uint48 restoredDelay);
+    event DefaultAdminTransferCancelled(address indexed restoredAdmin);
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -207,9 +267,30 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
     function _beginDefaultAdminTransfer(address newAdmin) internal virtual {
-        _defaultAdminTransferDelayedUntil = SafeCast.toUint48(block.timestamp) + defaultAdminDelay();
+        uint48 delayChangeSchedule = defaultAdminDelayChangeSchedule();
+
+        if (delayChangeSchedule != 0) {
+            require(delayChangeSchedule < block.timestamp, "AccessControl: Delay change in progress");
+            // Must read the raw private value to avoid getting a 0 from `pendingDefaultAdminDelay()`
+            _currentDefaultAdminDelay = _pendingDefaultAdminDelay;
+            _resetDefaultAdminDelayChange();
+        }
```
frangio, line 277: This is key logic to compute the correct admin delay and update the storage variables that we are placing in a function that deals with a different concern, which is transferring the admin. We should separate these concerns, and design internal abstractions so that this function can focus on the admin transfer.

With the changes I've suggested in other comments it should not be necessary to update the storage variables. We should be able to replace these lines with just `require(defaultAdminDelayChangeSchedule() == 0)`.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -21,7 +21,8 @@ import "../interfaces/IERC5313.sol";
  * * Only one account holds the `DEFAULT_ADMIN_ROLE` since deployment until it's potentially renounced.
  * * Enforce a 2-step process to transfer the `DEFAULT_ADMIN_ROLE` to another account.
  * * Enforce a configurable delay between the two steps, with the ability to cancel in between.
```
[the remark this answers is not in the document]
ernestognw: Makes sense to me, thanks for simplifying!

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: I changed it to just `_hasPassed` [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079/commits/9e2bda997a56d26657d02df81f910af65b723304). I think the parameter is indicative of what has actually passed.

ernestognw:
## Update

Turned this back into a draft.

I applied the naming suggestion along with the interface changes. I currently refactored the tests to reflect those changes. However, I turned this into a draft since I have the following TODOs:

- [ ] **Improve testing:** I'm making my head around edge cases but definitely there are improvements in the docs, including coverage for view functions
- [ ] **Updater documentation**: Current docs correspond to the previous version, I need to update
- [ ] **Review storage layout**: Now that both schedule and pending values are returned in `pending*()` functions, it's better to pack variables optimizing for that packaging, so they're read from the same storage slot. Haven't looked into it yet
- [ ] **Review use of private variables**: Some private variables are directly used, which is fine since the interface view functions doesn't directly correspond to the private variables, but I just want to make sure I use the interface the most to honor users overriding the getters.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Solving since I also needed to split this function in two: `_hasPassed` and `_isSet`, which is ultimately removing the initial concern here. Feel free to reopen if the naming still sounds bad

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: After refactoring with the suggested style changes, this is still sort of a problem. However, my recent updates include the following:

- I renamed this error to `AccessControl: can't change virtual delay` instead. Signaling that the value is virtual and not yet materialized
- Although a `_reset*` is required to make the current delay materialized, the reset is now also called from `cancel*`, removing the possibility for an actual cancellation.
- Under the same logic, this `require(..., ...)` can be replaced by a `_materializeDefaultAdminDelayChange` by splitting the current `_resetDefaultAdminDelayChange` in two.

We either keep the `require` or add `_materialize*()`. I'm ok for the latter, but will think more about it.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -68,7 +70,24 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+     */
+    function pendingDefaultAdminDelay() public view virtual returns (uint48) {
+        return _hasScheduledPassed(defaultAdminDelayChangeSchedule()) ? 0 : _pendingDefaultAdminDelay;
+    }
+
+    /**
+     * @inheritdoc IAccessControlDefaultAdminRules
+     */
+    function defaultAdminDelayChangeSchedule() public view virtual returns (uint48) {
+        return _defaultAdminDelayChangeSchedule;
```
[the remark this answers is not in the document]
ernestognw: Now it does via `pendingDefaultAdminDelay()`, solving.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -99,6 +118,31 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+        _beginDefaultAdminDelayChange(newDefaultAdminDelay);
+    }
+
+    /**
+     * @inheritdoc IAccessControlDefaultAdminRules
+     */
+    function cancelDefaultAdminDelayChange() public virtual onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(
+            !_hasScheduledPassed(defaultAdminDelayChangeSchedule()),
+            "AccessControl: can't cancel a passed schedule"
```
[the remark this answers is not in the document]
ernestognw: I removed this require statement and made `_resetDefaultAdminDelayChange` to "materialize" the changes.

I'm solving.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: I think this is now solved since the `_resetDefaultAdminDelayChange` is treated now as a noop, materializing the value new delay into storage whenever is needed.

Can you take a look, I think it may be cleaner?

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -222,19 +282,37 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
         super._revokeRole(role, account);
     }
 
+    /**
+     * @dev Sets a pending delay into effect if its delay has passed
+     */
+    function _materializeDefaultAdminDelayChange() private {
+        uint48 delaySchedule = _pendingDelaySchedule;
+        if (_isSet(delaySchedule) && _hasPassed(delaySchedule)) _currentDelay = _pendingDelay;
+    }
```
frangio, line 291: I think this function should delete the pending variables as well to leave a consistent state. We can add a boolean return value and we can make both `_materialize` and `_rollback` make sense on their own.

My motivation to propose this is that I was very confused when I saw "rollback delay change" in the body of "begin admin transfer".

```suggestion
    function _materializeDefaultAdminDelayChange() private returns (bool materialized) {
        uint48 delaySchedule = _pendingDelaySchedule;
        if (_isSet(delaySchedule) && _hasPassed(delaySchedule)) {
            _currentDelay = _pendingDelay;
            delete _pendingDelay;
            delete _pendingDelaySchedule;
            return true;
        } else {
            return false;
        }
    }
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -222,19 +282,37 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
      * @dev Resets the pending default admin and delayed until.
      */
-    function _resetDefaultAdminTransfer() private {
-        delete _pendingDefaultAdmin;
-        delete _defaultAdminTransferDelayedUntil;
+    function _rollbackDefaultAdminDelay() private {
+        _materializeDefaultAdminDelayChange();
+
+        delete _pendingDelay;
+        delete _pendingDelaySchedule;
```
frangio, line 300: Following my other suggestion:

```suggestion
        if (!_materializeDefaultAdminDelayChange()) {
            delete _pendingDelay;
            delete _pendingDelaySchedule;
        }
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -207,9 +249,27 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * Internal function without access restriction.
      */
     function _beginDefaultAdminTransfer(address newAdmin) internal virtual {
-        _defaultAdminTransferDelayedUntil = SafeCast.toUint48(block.timestamp) + defaultAdminDelay();
+        (, uint48 delaySchedule) = pendingDefaultAdminDelay();
+        if (_isSet(delaySchedule)) _rollbackDefaultAdminDelay();
```
frangio, line 253: Following my other suggestion, here we would just write:
```suggestion
        _materializeDefaultAdminDelay();
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -207,9 +249,27 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+
+        _pendingDefaultAdminSchedule = SafeCast.toUint48(block.timestamp) + defaultAdminDelay();
         _pendingDefaultAdmin = newAdmin;
-        emit DefaultAdminRoleChangeStarted(pendingDefaultAdmin(), defaultAdminTransferDelayedUntil());
+
+        (address newDefaultAdmin, uint48 transferSchedule) = pendingDefaultAdmin();
+        emit DefaultAdminTransferScheduled(newDefaultAdmin, transferSchedule);
```
frangio, line 259: IMO this is taking virtual functions too far. We've just set the values, here we should use the values we just set. If someone is overriding this function, it's on them to make sure their override is consistent with this logic.

```suggestion
        emit DefaultAdminTransferScheduled(newAdmin, _pendingDefaultAdminSchedule);
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -222,19 +282,37 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+    function _resetDefaultAdminTransfer() private {
+        delete _pendingDefaultAdmin;
+        delete _pendingDefaultAdminSchedule;
+    }
+
+    function _hasPassed(uint48 schedule) private view returns (bool) {
+        return schedule < block.timestamp;
+    }
+
+    function _isSet(uint48 schedule) private pure returns (bool) {
```
frangio, line 315: I would rename this to `_isScheduleSet`, and `_hasPassed` to `_hasSchedulePassed`.

test/access/AccessControl.behavior.js
```diff
@@ -215,13 +215,21 @@ function shouldBehaveLikeAccessControlEnumerable(errorPrefix, admin, authorized,
 function shouldBehaveLikeAccessControlDefaultAdminRules(errorPrefix, delay, defaultAdmin, newDefaultAdmin, other) {
   shouldSupportInterfaces(['AccessControlDefaultAdminRules']);
 
-  it('has a default disabled delayed until', async function () {
-    expect(await this.accessControl.defaultAdminTransferDelayedUntil()).to.be.bignumber.equal(web3.utils.toBN(0));
-  });
+  // it('has an admin transfer schedule disabled by default', async function () {
```
frangio, line 218: I think this is covered in your to-do list, but so we don't forget I'll say that these tests are commented out.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -207,9 +249,27 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * Internal function without access restriction.
      */
     function _beginDefaultAdminTransfer(address newAdmin) internal virtual {
```
frangio, line 251: Initially we said that an admin transfer wouldn't be allowed if there was a delay change in progress, but I don't see that in this function. Did you remove this on purpose? Can you share your reasoning? Honestly I don't think I ever had a good intuition for why we are/were not allowing these two things to happen in parallel. I see that initiating a delay change is still not allowed if there is an admin transfer in progress.

frangio (commented):
This is looking pretty good. :pray:

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Done :)

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Initially it was because of assumptions we eventually dropped. I didn't want to make this more complex because it was hard to reason what delay was in effect since it required a "materialization".

This last iteration is way more straightforward and well-encapsulated, so I decided to relax the restriction of both changes not occurring in parallel. It's easy to see what is the delay change in every moment.

test/access/AccessControl.behavior.js
ernestognw: Yes, updated. I'm solving since there are replacing tests.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Yes, I also realized by taking a look again for the docs. Just updated.

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: I took my time thinking about the best way to split responsibilities and I ended up with a setup that I think comes simple to avoid unnecessary reads in most scenarios, I think you'd like it, but it basically makes this function non-existent.

Can you take a look? I think it's clear enough.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -189,52 +247,98 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+        uint48 newSchedule = SafeCast.toUint48(block.timestamp) + _delayChangeWait(newDelay);
+
+        _pendingDelay = newDelay;
+        _pendingDelaySchedule = newSchedule;
+
+        // An `oldSchedule` from `pendingDefaultAdminDelay()` is only set if it hasn't passed
+        // So this is equivalent to a `_isScheduleSet(_pendingDelaySchedule) && !_hasSchedulePassed(_pendingDelaySchedule)`
+        if (_isScheduleSet(oldSchedule))
+            // Emit for implicit cancelations when another delay was scheduled.
+            emit DefaultAdminDelayChangeCanceled();
```
frangio, line 310:
```suggestion
        if (_isScheduleSet(oldSchedule)) {
            // Emit for implicit cancelations when another delay was scheduled.
            emit DefaultAdminDelayChangeCanceled();
        }
```

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: We discussed this in a meeting and agreed on the new approach

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: No longer applies

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: No longer applies

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -161,6 +185,44 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+        return
+            newDefaultAdminDelay > currentDelay ? defaultAdminDelayIncreaseWait() : currentDelay - newDefaultAdminDelay;
+    }
+
+    /**
+     * @dev Setter of the tuple for pending admin and its schedule.
+     *
+     * May emit a DefaultAdminTransferCanceled event.
+     */
+    function _setPendingDefaultAdmin(address newAdmin, uint48 newSchedule) internal virtual {
```
ernestognw, line 213: Should we consider a `_setPendingDelay`? It'll be used only once but it makes sense to me for keeping a consistent `internal` API, I just don't know if that's important.

ernestognw: Changed it, looks cleaner.

ernestognw: image

frangio:
The `tests` check is [failing due to storage layout checks](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/4482615873/jobs/7880879756?pr=4079), but this is code we haven't released yet so we are free to change its layout.

frangio (approved):
Ok, this looks good to me.

contracts/access/IAccessControlDefaultAdminRules.sol
```diff
@@ -12,62 +12,137 @@ import "./IAccessControl.sol";
...
+     *
+     * May emit a DefaultAdminTransferCanceled event.
      */
     function cancelDefaultAdminTransfer() external;
+
+    /**
+     * @dev Begins a {defaultAdminDelay} change by scheduling the change in a way in which the
+     * current delay is still guaranteed to be respected.
+     *
+     * The {pendingDefaultAdminDelay}'s schedule is defined such that `(its schedule + a delayed default admin transfer)`
```
Amxx, line 122: > its schedule + a delayed default admin transfer

That is not very clear

contracts/access/IAccessControlDefaultAdminRules.sol
```diff
@@ -12,62 +12,137 @@ import "./IAccessControl.sol";
...
+     * takes at least the current {defaultAdminDelay}, following that:
+     *
+     * - The schedule is `block.timestamp + (current delay - new delay)` if the delay is reduced.
+     * - The schedule is `block.timestamp + {defaultAdminDelayIncreaseWait}` if the delay is increased.
+     *
+     * A {pendingDefaultAdminDelay} that never got into effect will be canceled in favor of a new scheduled change.
```
Amxx, line 128: don't understand that line

contracts/access/IAccessControlDefaultAdminRules.sol
```diff
@@ -12,62 +12,137 @@ import "./IAccessControl.sol";
...
+
+    /**
+     * @dev Begins a {defaultAdminDelay} change by scheduling the change in a way in which the
+     * current delay is still guaranteed to be respected.
+     *
+     * The {pendingDefaultAdminDelay}'s schedule is defined such that `(its schedule + a delayed default admin transfer)`
+     * takes at least the current {defaultAdminDelay}, following that:
+     *
+     * - The schedule is `block.timestamp + (current delay - new delay)` if the delay is reduced.
+     * - The schedule is `block.timestamp + {defaultAdminDelayIncreaseWait}` if the delay is increased.
```
Amxx, line 126: you need to explain the rational. 

Is it that `current delay` might be too short, which would not allow the admin to cancel if the new delay is too long ?

Amxx (commented):
That is just the interface review. There is already a lot

contracts/access/IAccessControlDefaultAdminRules.sol
```diff
@@ -12,62 +12,137 @@ import "./IAccessControl.sol";
...
+     *
+     * Requirements:
+     *
+     * - Only can be called by the {pendingDefaultAdmin}'s `newAdmin` and if its `scheduled` has passed.
      */
     function acceptDefaultAdminTransfer() external;
 
     /**
-     * @dev Cancels a `DEFAULT_ADMIN_ROLE` transfer.
+     * @dev Cancels a {defaultAdmin} transfer if there was a {pendingDefaultAdmin}.
```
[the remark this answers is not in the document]
ernestognw: I'm referring to a NOOP here. I'm definitely removing the `if part` since it is misleading, but not adding to the requirements as it doesn't revert.

contracts/access/IAccessControlDefaultAdminRules.sol
```diff
@@ -12,62 +12,137 @@ import "./IAccessControl.sol";
...
+     * A zero value in both indicates there's no pending admin transfer.
+     *
+     * NOTE: A zero value only for `newAdmin` means that {defaultAdmin} is being renounced.
      */
-    function pendingDefaultAdmin() external view returns (address);
+    function pendingDefaultAdmin() external view returns (address newAdmin, uint48 schedule);
 
     /**
-     * @dev Returns the timestamp after which the pending default admin can claim the `DEFAULT_ADMIN_ROLE`.
+     * @dev Time in seconds to wait before a delay is increased after calling {changeDefaultAdminDelay}.
```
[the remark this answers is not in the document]
ernestognw: Yes, I added a direct reference to `{defaultAdminDelay}`

contracts/access/IAccessControlDefaultAdminRules.sol
```diff
@@ -12,62 +12,137 @@ import "./IAccessControl.sol";
...
-    function defaultAdminTransferDelayedUntil() external view returns (uint48);
+    function defaultAdminDelayIncreaseWait() external view returns (uint48);
 
     /**
-     * @dev Starts a `DEFAULT_ADMIN_ROLE` transfer by setting a pending default admin
-     * and a timer to pass.
+     * @dev Returns the delay between each {defaultAdmin} transfer.
      *
-     * Requirements:
+     * A scheduled delay change will take effect as soon as the schedule passes, returning the new delay.
```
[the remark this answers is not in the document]
ernestognw: I replaced with:

```
     * NOTE: If a delay change has been scheduled, it will take effect as soon as the schedule passes, making this
     * function returns the new delay. See {changeDefaultAdminDelay}.
```

Clearer and has a reference for the relevant function to change the delay

contracts/access/IAccessControlDefaultAdminRules.sol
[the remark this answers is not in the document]
ernestognw: Removed the `if` and improved the docs

contracts/access/IAccessControlDefaultAdminRules.sol
```diff
@@ -12,62 +12,137 @@ import "./IAccessControl.sol";
...
-    function pendingDefaultAdmin() external view returns (address);
+    function pendingDefaultAdmin() external view returns (address newAdmin, uint48 schedule);
 
     /**
-     * @dev Returns the timestamp after which the pending default admin can claim the `DEFAULT_ADMIN_ROLE`.
+     * @dev Time in seconds to wait before a delay is increased after calling {changeDefaultAdminDelay}.
+     * Default to 5 days.
+     *
+     * Used whenever {changeDefaultAdminDelay} is called with a `newDelay` higher
+     * than the current {defaultAdminDelay}.
```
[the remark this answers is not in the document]
ernestognw: I changed how this function is presented and increased the details in `changeDefaultAdminDelay` for reference, so you first read what the change does and then read what the `defaultAdminDelayIncreaseWait`, now your question is explained there with better clarity.

contracts/access/IAccessControlDefaultAdminRules.sol
ernestognw: I improved the way is described. I'm solving since I think it should be clear now

contracts/access/IAccessControlDefaultAdminRules.sol
ernestognw: I tried to express that if a `{pendingDefaultAdminDelay}` never reaches its schedule before the delay is changed again, the pending one will be lost in favor of the newly set.

```suggestion
     * By calling this function, a {pendingDefaultAdminDelay} that never got into effect will be canceled in favor of a new scheduled change.
```

Could be?

contracts/access/IAccessControlDefaultAdminRules.sol
ernestognw: I improved the way is described, but I moved the rationale to `defaultAdminDelayIncreaseWait`.

```
     * With the purpose of avoiding doubly delayed transfers (waiting the new delay to start another delayed transfer),
     * an increased delay can relax the effect schedule, however, when the original delay is low, waiting only the old
     * one can pose a risk of accidentally locking the contract (i.e. increasing from 1 second to 10 years).
     *
     * For both avoiding double-delayed transfers or fast-effect increases, this function can be configured for a custom
     * {defaultAdminDelay} increase scheduling.
     *
```

Do you think this is clear?

ernestognw:
> Overall I fell like there are two many references to two many long and complexe names.

The rationale for the current naming was to be as clear as possible and avoid clashes between concepts like:

- An admin vs a proxy admin
- A transfer vs a token transfer
- A delay vs other delayed systems

>IMO we should refocus on a small number of concept:
>
>admin
>pending admin
>delay

I'm afraid these break the clarity in some of the examples I shared. I'm still open to improvements, but so far I think the current naming has clearly improved since the beginning.

> and try to formulate everything using sort and simple sentences. Avoid indeterminate (ex: "a delay") and use precise (ex: "the delay" or "the delay applied to admin updates")
 
I rewrote part of the docs for better clarity, I think it's now better

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -161,6 +185,67 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
```
Amxx, line 213: If we don't make this function private, then we need to do the following:
- `require(newAdmin == address(0) || newSchedule == 0);`
- `emit DefaultAdminTransferScheduled(newDelay, newSchedule)` here

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -161,6 +185,67 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+        (, uint48 oldSchedule) = pendingDefaultAdmin();
+
+        _pendingDefaultAdmin = newAdmin;
+        _pendingDefaultAdminSchedule = newSchedule;
+
+        // An `oldSchedule` from `pendingDefaultAdmin()` is only set if it hasn't been accepted.
+        if (_isScheduleSet(oldSchedule)) {
+            // Emit for implicit cancellations when another default admin was scheduled.
+            emit DefaultAdminTransferCanceled();
+        }
```
Amxx, line 223:
```suggestion
        }
        if (newSchedule != 0) {
            emit DefaultAdminTransferScheduled(newDelay, newSchedule);
        }
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -161,6 +185,67 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+            _currentDelay = _pendingDelay; // Materialize a virtual delay
+        }
+
+        _pendingDelay = newDelay;
+        _pendingDelaySchedule = newSchedule;
+
+        if (set && !passed) {
+            // Emit for implicit cancellations when another delay was scheduled.
+            emit DefaultAdminDelayChangeCanceled();
+        }
```
Amxx, line 246: Same here, if it is to be internal and not private, than we need to emit the event
```suggestion
        }
        if (_isScheduleSet(newSchedule)) {
            emit DefaultAdminDelayChangeScheduled(newDelay, newSchedule);
        }
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -189,52 +271,81 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
      */
-    function _revokeRole(bytes32 role, address account) internal virtual override {
-        if (role == DEFAULT_ADMIN_ROLE) {
-            delete _currentDefaultAdmin;
-        }
-        super._revokeRole(role, account);
+    function _changeDefaultAdminDelay(uint48 newDelay) internal virtual {
+        uint48 newSchedule = SafeCast.toUint48(block.timestamp) + _delayChangeWait(newDelay);
+        _setPendingDelay(newDelay, newSchedule);
+        emit DefaultAdminDelayChangeScheduled(newDelay, newSchedule);
```
Amxx, line 326: let `_setPendingDelay` handle the event
```suggestion
```

contracts/access/IAccessControlDefaultAdminRules.sol
Amxx: That's better. 

I still think using `{pendingDefaultAdminDelay}` instead of something like `a pending change to the {defaultAdminDelay}` is just asking people to do the rewrite in their head ... but we'll address that in a further PR.

contracts/access/IAccessControlDefaultAdminRules.sol
Amxx: Maybe if the new delay is longer than the current delay, but smaller than the "i don't know what its called delay that is 5 days by default" ... then we can just use do accelerate the change

contracts/access/AccessControlDefaultAdminRules.sol
Amxx: replaced by https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079/files#r1147649767

ernestognw:
IIUC, the recommendations are made based on exposing the setter `_set*` functions to the `internal` interface, so I decided to make the functions only private to keep the same abstraction, and also @frangio agrees this makes it more intuitive what can be done with the contract.

Although I agree that there may be some tweaks to make the `_set*` functions `internal`, that's something we can discuss further since it doesn't change the interface and any change to these private functions will be backwards compatible.

contracts/access/AccessControlDefaultAdminRules.sol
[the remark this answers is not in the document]
ernestognw: This is what I don't feel comfortable with:

```solidity
if (_isScheduleSet(newSchedule)) {
    emit DefaultAdminDelayChangeScheduled(newDelay, newSchedule);
}
```

Since it requires an `if` that can be inferred from the parent function.
I'm solving because this function was marked as private and this shouldn't be an issue anymore

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Solved with the same rationale as in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079#issuecomment-1483295774

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -189,52 +271,81 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
...
+
+    /**
+     * @dev See {beginDefaultAdminTransfer}.
+     *
+     * Internal function without access restriction.
+     */
+    function _beginDefaultAdminTransfer(address newAdmin) internal virtual {
+        uint48 newSchedule = SafeCast.toUint48(block.timestamp) + defaultAdminDelay();
+        _setPendingDefaultAdmin(newAdmin, newSchedule);
+        emit DefaultAdminTransferScheduled(newAdmin, newSchedule);
```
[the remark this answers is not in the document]
ernestognw: Solved with the same rationale as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079#issuecomment-1483295774

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: Solving with the same rational as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4079#issuecomment-1483295774

contracts/access/IAccessControlDefaultAdminRules.sol
ernestognw: Are you proposing a `Math.min(defaultAdminDelayIncreaseWait(), _currentDefaultAdminDelay)` instead? I definitely would like to do that but it carries the same problem as in a delay change from 1 second to 10 years.

I agree with your comments, so I'm open to ideas.

Amxx:
- making the function private and not internal breaks some of our guidelines.
- honestly, if they are not internal, we may as well in-line (and simplify) them to improve the overall readability of the contract.

frangio (dismissed):

frangio (approved):
LGTM

PR #4081, opened 2023-03-01
Improvement. Make sure the new owner isn't the old owner, which will reduce misoperations.

FrankLiu007:
Fixes #???? 

check if the new owner is the old owner before transferOwnership which will reduce misoperations.

> before
```
    function _transferOwnership(address newOwner) internal virtual {
        address oldOwner = _owner;
        _owner = newOwner;
        emit OwnershipTransferred(oldOwner, newOwner);
    }
```
> after
```
    function _transferOwnership(address newOwner) internal virtual {
        require(_owner!=newOwner, "Ownable: the newOwner is the same with the old owner");
        address oldOwner = _owner;
        _owner = newOwner;
        emit OwnershipTransferred(oldOwner, newOwner);
    }
```

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -76,6 +76,7 @@ abstract contract Ownable is Context {
      * Internal function without access restriction.
      */
     function _transferOwnership(address newOwner) internal virtual {
+        require(_owner!=newOwner, "Ownable: the newOwner is the same with the old owner");
         address oldOwner = _owner;
         _owner = newOwner;
         emit OwnershipTransferred(oldOwner, newOwner);


Amxx:
Hello @FrankLiu007 

We don't have any reason to prevent transferring to the current owner. Will it can be debated why someone would do that, its a valid no-op that doesn't need reverting.

Changing that behavior might break some usecases/applications. That is something we will want to avoid.

Why do you think transferring to the current owner is problematic and should cause a revert?

Amxx:
If you want to add this behavior to a contract of your own, you can do it like this


```solidity
function _transferOwnership(address newOwner) internal virtual override {
    require(owner() != newOwner, "Ownable: the newOwner is the same with the old owner");
}
```

FrankLiu007:
@Amxx Hello. No one want to transferring ownership to the current owner. If this happens, it means that something unwanted happened. Reverting the transcation gives the developer a seconder chance to review the code. Security should always be the top consideration.

Amxx:
> @Amxx Hello. No one want to transferring ownership to the current owner. If this happens, it means that something unwanted (maybe bad) happened.

You may not want it, but saying no-one want it is an exaggeration. If you think of tokens like erc20 and erc721, it is valid to

- transfer token to yourself
- transfer 0 tokens.

It's even valid to transfer 0 tokens to yourself!

These are all valid no-op that will happen (particularly with contract composition). We don't want to block these usecases for everyone. We also don't want to add opinionated behavior that are not explicit in the ERCs.

Still, if you want your contracts to not allow such a transfer, that is your choice to make, and it's easy for you to add that.

FrankLiu007:
@Amxx I think you are right. Sorry for the inconveniency.

PR #4082, opened 2023-03-01
Fix an upgrade replay bug in Governor.propose

Amxx:
The `proposer` is not set for proposal created before a smart contract upgrade.

This would allow anyone to re-propose a proposal created before the upgrade. The timers + flags (execute/cancel) would be reset, but the votes would not. Old vote count (if any) still hold. Accounts that already voted cannot re-vote to oppose. New vote use a new snapshot.

This is a fix to an bug introduced in unreleased code. Do we need a changeset?

#### PR Checklist


frangio:
>Do we need a changeset?

No, this is fine.

frangio (approved):
LGTM.

PR #4083, opened 2023-03-01
Formal verification using Certora

Amxx:
This PR introduces the core tooling for formal verification. The scripts and specs should be copied for the `formal-verification` branch when they run cleanly.

This include a refactor of the Makefile:

- `make apply`: copy the `contracts` folder to `certora/patched` and apply all patches found in `certora/diffs`
- `make record`: compare the `certora/patched` folder to `contracts` and store the diff as patches in `certora/diffs`
- there is one diff per source file to improve the discoverability of changes/patches, and to avoid merge conflicts.

The github action runs on all commit on the `master` and `release-v*` branches. For other branches (and PR) it must be triggered manually. Maybe we could have a label that turn on the formal verification workflow on a specific PR.


.github/workflows/formal-verifiation.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  list-scripts:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: echo "matrix=$(ls certora/scripts/{,**}/*.sh | grep -v '\WnoCI\W' | jq -Rsc 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
Amxx: We can get rid of the `noCI` part and assume from know on all scripts must pass

Amxx, line 19:
```suggestion
        run: echo "matrix=$(echo ../certora-old/scripts/{,**}/*.sh | jq -Rsc 'split(" ")[:-1]' | jq -Rsc 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
```

Amxx: that does not work

Amxx, line 19:
```suggestion
        run: echo "matrix=$(echo -n certora/scripts/**/*.sh | jq -Rsc 'split(" ")')" >> $GITHUB_OUTPUT
```

.github/workflows/formal-verifiation.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  list-scripts:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: echo "matrix=$(echo -n certora/scripts/**/*.sh | jq -Rsc 'split(" ")')" >> $GITHUB_OUTPUT
```
Amxx, line 19:
```suggestion
        run: echo "matrix=$(echo -n certora/scripts/*.sh | jq -Rsc 'split(" ")')" >> $GITHUB_OUTPUT
```

.github/workflows/formal-verifiation.yml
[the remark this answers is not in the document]
frangio: The `ls` here is redundant, we could use `echo`, yes.

But I would agree with this:
>We can get rid of the noCI part and assume from know on all scripts must pass

.github/workflows/formal-verifiation.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  list-scripts:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: echo "matrix=$(echo -n certora/scripts/*.sh | jq -Rsc 'split(" ")')" >> $GITHUB_OUTPUT
```
Amxx, line 19: So I managed to make that work without ls/grep, which the linter complains about, but now it asks me to `Double quote to prevent globbing and word splitting` when globbing is exactly what I want.

.github/workflows/formal-verification.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  list-scripts:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: set-matrix
+        run: echo "matrix=$(echo -n certora/scripts/*.sh | jq -Rsc 'split(" ")')" >> $GITHUB_OUTPUT
```
Amxx, line 19:
```suggestion
        run: echo "matrix=$(echo -n certora/scripts/*.sh | jq -Rsc 'split(" ")')" >> "$GITHUB_OUTPUT"
```

PR #4084, opened 2023-03-01
Formal verification using Certora

Amxx:
Replaces #4083 

---

This PR introduces the core tooling for formal verification. The scripts and specs should be copied for the formal-verification branch when they run cleanly.

This include a refactor of the Makefile:

- make apply: copy the contracts folder to certora/patched and apply all patches found in certora/diffs
- make record: compare the certora/patched folder to contracts and store the diff as patches in certora/diffs
- there is one diff per source file to improve the discoverability of changes/patches, and to avoid merge conflicts.

The github action runs on all commit on the master and release-v* branches. For other branches (and PR) it must be triggered manually. Maybe we could have a label that turn on the formal verification workflow on a specific PR.

---

This PR currently includes specs for:

- [x] AccessControl
- [x] ERC20
  - [x] ERC20Permit
  - [x] ERC20FlashMint


certora/specs/methods/IERC3156.spec
```diff
@@ -0,0 +1,5 @@
+methods {
+    maxFlashLoan(address)                    returns (uint256) envfree => DISPATCHER(true)
+    flashFee(address,uint256)                returns (uint256) envfree => DISPATCHER(true)
+    flashLoan(address,address,uint256,bytes) returns (bool)            => DISPATCHER(true)
+}
```
frangio, line 5: What exactly does it mean to include this file? I had a look at the [`methods`](https://docs.certora.com/en/latest/docs/cvl/methods.html) docs but could get a good intuition.

certora/specs/ERC20FlashMint.spec
```diff
@@ -0,0 +1,43 @@
...
+*/
+ghost mapping(address => uint256) trackedMintAmount;
+ghost mapping(address => uint256) trackedBurnAmount;
+
+function specMint(address account, uint256 amount) returns bool { trackedMintAmount[account] = amount; return true; }
+function specBurn(address account, uint256 amount) returns bool { trackedBurnAmount[account] = amount; return true; }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: When doing a flashLoan, "amount" is minted and "amount + fee" is burnt                                        │
```
frangio, line 27: It's not accurate to say that `amount + fee` is burnt, it's not true when the receiver is non-zero.

Should we also test that the fee is transferred to the receiver?

certora/specs/ERC20FlashMint.spec
```diff
@@ -0,0 +1,43 @@
...
+import "methods/IERC20.spec"
+import "methods/IERC3156.spec"
+
+methods {
+    // non standard ERC3156 functions
+    flashFeeReceiver() returns (address) envfree
+
+    // patched - function summaries below
+    _mint(address account, uint256 amount) => specMint(account, amount)
+    _burn(address account, uint256 amount) => specBurn(account, amount)
```
frangio, line 11: Does this also replace the behavior when the function is called internally?

certora/specs/ERC20FlashMint.spec
```diff
@@ -0,0 +1,43 @@
...
+    uint256 amount;
+    bytes data;
+
+    uint256 feeBefore = flashFee(token, amount);
+    address recipient = flashFeeReceiver();
+
+    flashLoan(e, receiver, token, amount, data);
+
+    assert to_mathint(amount)                                    == trackedMintAmount[receiver];
+    assert to_mathint(amount + (recipient == 0 ? feeBefore : 0)) == trackedBurnAmount[receiver];
```
frangio, line 42: Do you know why this rule isn't defined based on the return values of `totalSupply()`?

frangio:
Could we have a check in all PRs that the patches apply?

certora/specs/ERC20.spec
```diff
@@ -0,0 +1,414 @@
...
+    uint256 toBalanceBefore    = balanceOf(to);
+    uint256 otherBalanceBefore = balanceOf(other);
+    uint256 totalSupplyBefore  = totalSupply();
+
+    // run transaction
+    _mint@withrevert(e, to, amount);
+
+    // check outcome
+    if (lastReverted) {
+        assert to == 0 || totalSupplyBefore + amount > to_uint256(max_uint256);
```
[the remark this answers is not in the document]
Amxx: Because here we are not in solidity. Number are not limited by uint256

certora/specs/ERC20.spec
```diff
@@ -0,0 +1,414 @@
...
+        // allowance is valid & updated
+        assert allowanceBefore            >= amount;
+        assert allowance(holder, spender) == (allowanceBefore == max_uint256 ? to_uint256(max_uint256) : allowanceBefore - amount);
+
+        // balances of holder and recipient are updated
+        assert balanceOf(holder)          == holderBalanceBefore    - (holder == recipient ? 0 : amount);
+        assert balanceOf(recipient)       == recipientBalanceBefore + (holder == recipient ? 0 : amount);
+
+        // no other balance is modified
+        assert balanceOf(other) != otherBalanceBefore => (other == holder || other == recipient);
```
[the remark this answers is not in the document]
Amxx: That would not work.

The property says: if balance X changes, than X is either the sender or the recipient.

What you say is : if balance X changes, than X is the sender or sender and recipient are equal. Which is wrong.

certora/specs/ERC20.spec
```diff
@@ -0,0 +1,414 @@
...
+    require nonceBefore      < max_uint256;
+    require otherNonceBefore < max_uint256;
+
+    // run transaction
+    permit@withrevert(e, holder, spender, amount, deadline, v, r, s);
+
+    // check outcome
+    if (lastReverted) {
+        // Without formally checking the signature, we can't verify exactly the revert causes
+        assert deadline < e.block.timestamp || true;
```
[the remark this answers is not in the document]
Amxx: Yes. The thing is we have no good way to check the signature validity

certora/specs/ERC20FlashMint.spec
Amxx: Yes.

certora/specs/ERC20FlashMint.spec
Amxx: Afaik there are many reason, on of which is that the flash loan recipient can do anything, including minting and burning.

certora/specs/ERC20FlashMint.spec
Amxx: Short answer is I'll try to address that. But there is A LOT I'd like to say about this rule so we should probably get on a call for that

requirements.txt
```diff
@@ -0,0 +1 @@
+certora-cli==3.6.3
```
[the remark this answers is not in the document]
Amxx: This acts like a package.json for pip. Should we manually install that directly in the yaml workflow ?

certora/specs/methods/IERC3156.spec
Amxx: AFAIK, declaring the methods is optional, but it can bring some added values

- One of the value is that if you use one of this methods in a spec file, but the contract targeted doesn't implement them, then the specs checking will not fail, it will just skip the tests.
  - That is something we can see in ERC20.spec. [This run](https://prover.certora.com/output/71959/2513b40ab3864c428cb5e3ed44a504cb/?anonymousKey=c8e9bd5b655e025c994c5182bcc41be711444c64) shows what happen it the contract doesn't publicly expose mint and burn, the tests are skipped and the rest of the property is verified
  - IMO this is great for test reuse
  
- Another value, is that it allows us to specify that the method is envfree (the whole concept of envfree is very strange to me. A function that reads from storage can be envfree, but not one that reads block.timestamp) 

- Lastly it enable "summaries" ... Now that is a concept I find very hard to understand, but certora that `DISPATCHER(true)` should be used for "standard interfaces such as ERC20" ... so I follow their guideline.

certora/specs/ERC20FlashMint.spec
Amxx: Lets try:

- This rule was written before we had a fee recipient, so the idea of transfer to the recipient was not included
- This rule was targeting a "bad" harness, the devs did not realize it but the fee they was always 0 for the FV point of view.
- The core of the flash loan can do anything. Including minting/burning/transfering ... so any check that is balance or supply related should fail. If it doesn't that means the FV is not working properly.
- The workaround Certora found is to "capture" the `burn` call, and register it locally. I duplicated that for the `mint`.
  - Technically, we could do that for transfer as well. That would capture the transfer call, and we would be able to check that such a transfer was indeed executed. 
  - However, that would require the `_transfer` function to be patched to public (I don't think we can do all that on internal methods). If we do so, all the ERC20 spec will have to consider this new function as an entrypoint. Which raise an even bigger question:  **what are we formally verifying ?**. Do we want to formally verify the behavior of the contract with regard to its public/external interface. Do we want to formally verify the internal interface as well?

Amxx: I was wrong about one thing: it is possible to summaries internal function. So no need to make `_transfer` public to check the transfer to the recipient.

Question of knowing if `_transfer` should be part of the FV spec remains open.

certora/specs/ERC20.spec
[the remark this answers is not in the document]
Amxx: blocks cannot be empty, we need to have "something".
(or we change how the if is written, but I'd like to keep it consistent with the other rules)

Amxx:
> Could we have a check in all PRs that the patches apply?

done

requirements.txt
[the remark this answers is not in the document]
Amxx: also, the `actions/setup-python@v2` action will fail if there is no file matching `**/requirements.txt` in the repo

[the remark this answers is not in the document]
frangio: It makes a lot more sense now if we do `pip install -r requirements.txt`.

.github/workflows/formal-verification.yml
```diff
@@ -13,8 +14,16 @@ env:
...
+  apply-diff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Apply patches
+        run: make -C certora apply
+
   list-scripts:
     runs-on: ubuntu-latest
+    # if: github.event_name != 'pull_request'
```
frangio, line 26: Are we uncommenting this line before merging?

Amxx: yes

certora/specs/ERC20.spec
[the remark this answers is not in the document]
frangio: I would prefer to have `assert true`.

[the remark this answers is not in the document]
Amxx: ok

certora/specs/ERC20.spec
```diff
@@ -0,0 +1,414 @@
...
```
Amxx, line 401:
```suggestion
        assert true;
```

certora/run.js
```diff
@@ -0,0 +1,40 @@
...
+
+const fs = require('fs');
+const proc = require('child_process');
+const { PassThrough } = require('stream');
+const events = require('events');
+const limit = require('p-limit')(MAX_PARALLEL);
+
+for (const [spec, { files, contract }] of Object.entries(specs)) {
+  limit(run, files, contract, spec);
+}
```
Amxx, line 15: We'll need that soon, lets plan for it.

```suggestion
for (const [spec, { files, contract, args = [] }] of Object.entries(specs)) {
  limit(run, files, contract, spec, args);
}
```

frangio: We need different arguments for different specs?

certora/run.js
```diff
@@ -0,0 +1,46 @@
...
+
+console.log(process.cwd());
+
+for (const { spec, contract, files, args = [] } of Object.values(specs)) {
+  limit(run, spec, contract, files, args);
+}
+
+// Run certora, aggregate the output and print it at the end
+async function run(spec, contract, files, args = []) {
+  args = [...files, '--verify', `${contract}:certora/specs/${spec}.sol`, '--optimistic_loop', ...args]
```
Amxx, line 20:
```suggestion
  args = [...files, '--verify', `${contract}:certora/specs/${spec}.spec`, '--optimistic_loop', ...args]
```

certora/run.js
```diff
@@ -0,0 +1,46 @@
...
+  limit(run, spec, contract, files, args);
+}
+
+// Run certora, aggregate the output and print it at the end
+async function run(spec, contract, files, args = []) {
+  const child = proc.spawn('certoraRun', [
+    ...files,
+    '--verify',
+    `${contract}:certora/specs/${spec}.spec`,
+    '--optimistic_loop',
```
Amxx, line 22:
```suggestion
    '--short_output',
```

frangio (approved):

frangio (approved):

PR #4088, opened 2023-03-03
Fix empty short string encoding

### LINKED ISSUE #4087: `toStringWithFallback` misbehaves for an empty string
I'm sorry for the informal format of this issue, but it's a really small one I've randomly found while reading code.

`toStringWithFallback` checks if the short string encoding failed by running `if (length(value) > 0) {`, but this doesn't take into account a genuinely empty string, which encodes in `toShortStringWithFallback` just fine, without a fallback: `if (bytes(value).length < 32) {` ` return toShortString(value);`. The result is that if a genuinely empty ShortString is used with a non-empty fallback storage string, the storage will be read. This wastes gas, but may also return an invalid value if storage isn't properly cleaned.

To solve this IMO we should encode differently a fallen-back ShortString in `toShortStringWithFallback`, e.g. by setting its length to 255.

Amxx:
Fixes #4087

#### PR Checklist

- [x] Tests

### DIFF
--- a/contracts/utils/ShortStrings.sol
+++ b/contracts/utils/ShortStrings.sol
@@ -32,7 +32,7 @@ type ShortString is bytes32;
  * ```
  */
 library ShortStrings {
-    bytes32 constant _INVALID_SHORT_STRING = 0x00000000000000000000000000000000000000000000000000000000000000FF;
+    bytes32 constant _FALLBACK_SENTINEL = 0x00000000000000000000000000000000000000000000000000000000000000FF;
 
     error StringTooLong(string str);
     error InvalidShortString();
@@ -84,7 +84,7 @@ library ShortStrings {
             return toShortString(value);
         } else {
             StorageSlot.getStringSlot(store).value = value;
-            return ShortString.wrap(_INVALID_SHORT_STRING);
+            return ShortString.wrap(_FALLBACK_SENTINEL);
         }
     }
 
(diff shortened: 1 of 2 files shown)


contracts/utils/ShortStrings.sol
```diff
@@ -44,7 +45,7 @@ library ShortStrings {
         if (bstr.length > 31) {
             revert StringTooLong(str);
         }
-        return ShortString.wrap(bytes32(uint256(bytes32(bstr)) | bstr.length));
+        return ShortString.wrap(bytes32(uint256(bytes32(bstr)) | (bstr.length + 1)));
```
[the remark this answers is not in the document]
Amxx: done

frangio:
This presents a question. What should be the value of an uninitialized `ShortString`? This PR is currently making that an invalid state (attempt to decode results in `InvalidShortString`), and I think intuitively people would expect it to be an empty string like the native `string` type.

I think a ShortString should only become invalid should be when you use `setWithFallback` and the function has to fall back to storage. `setWithFallback` should set a sentinel value (e.g. length = 255) to indicate that the value must be read from storage with `getWithFallback`.

contracts/utils/ShortStrings.sol
```diff
@@ -32,7 +32,10 @@ type ShortString is bytes32;
  * ```
  */
 library ShortStrings {
+    bytes32 constant _INVALID_SHORT_STRING = 0x00000000000000000000000000000000000000000000000000000000000000FF;
```
frangio, line 35:
```suggestion
    bytes32 private constant _INVALID_SHORT_STRING = 0x00000000000000000000000000000000000000000000000000000000000000FF;
```

frangio, line 35: I would call this something like `_FALLBACK_SENTINEL`.

CodeSandwich:
Thank you for fixing it! From my side it looks good, but obviously I'm not a maintainer.

frangio (approved):

frangio:
Looks good to me.

The new logic makes it more evident that `lengthWithFallback` might make sense. But I wouldn't add that in this PR.

PR #4089, opened 2023-03-03
Add `lengthWithFallback` to the ShortStrings library

Amxx:
`ShortStrings` is not yet released, so we don't need more changeset than what we already have.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/ShortStrings.sol
+++ b/contracts/utils/ShortStrings.sol
@@ -98,4 +98,19 @@ library ShortStrings {
             return store;
         }
     }
+
+    /**
+     * @dev Return the length of a string that was encoded to `ShortString` or written to storage using {setWithFallback}.
+     */
+    function lengthWithFallback(ShortString value, string storage store) internal view returns (uint256) {
+        if (ShortString.unwrap(value) != _FALLBACK_SENTINEL) {
+            return length(value);
+        } else {
+            uint256 result;
+            assembly {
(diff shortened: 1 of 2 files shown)


contracts/utils/ShortStrings.sol
```diff
@@ -98,4 +98,19 @@ library ShortStrings {
...
+     */
+    function lengthWithFallback(ShortString value, string storage store) internal view returns (uint256) {
+        if (ShortString.unwrap(value) != _FALLBACK_SENTINEL) {
+            return length(value);
+        } else {
+            uint256 result;
+            assembly {
+                result := div(sload(store.slot), 2)
+            }
+            return result;
```
Amxx, line 113: Note: it would be great if we could do 

```solidity
return store.length;
``` 
or
```solidity
return StorageSlot.getStringSlot(store).value.length;
```

but neither will compile.

frangio: I think that's just because `length` is not available for the `string` type. You should be able to do `bytes(store).length`.

By the way, the reason why length is not available is that users might expect length to be the character length of the string, but since they are UTF-8 strings the length operation is not O(1). `bytes(str).length` will return the byte length.

frangio (approved):
LGTM, but I renamed `length` to `byteLength` based on the same reasoning, let me know if you agree with that.

PR #4090, opened 2023-03-06
Clarify Governor Bravo compatibility scope

### LINKED ISSUE #3877: Missing GovernorBravo compatibility
The GovernorCompatibilityBravo.sol contract seems to be missing alot of the the public interface members for bravo

such as:

    /// @notice The latest proposal for each proposer
    mapping (address => uint) public latestProposalIds;

ernestognw:
Fixes #3877 

Currently, `GovernorCompatibilityBravo` extends the interface of `Governor` for compatibility with Bravo versions. However, there's still missing compatibility for some functions that are not in scope or are directly not needed. This also includes a special consideration related to how `proposalId`s work internally.


docs/modules/ROOT/pages/governance.adoc
```diff
@@ -20,7 +20,9 @@ The ERC20 extension to keep track of votes and vote delegation is one such case.
 
 === Governor & GovernorCompatibilityBravo
 
-An OpenZeppelin Governor contract is by default not interface-compatible with GovernorAlpha or Bravo, since some of the functions are different or missing, although it shares all of the same events. However, it’s possible to opt in to full compatibility by inheriting from the GovernorCompatibilityBravo module. The contract will be cheaper to deploy and use without this module.
+An OpenZeppelin Governor contract is by default not interface-compatible with GovernorAlpha or Bravo, since some of the functions are different or missing, although it shares all of the same events. However, it’s possible to opt in for compatibility by inheriting from the GovernorCompatibilityBravo module, which is designed for covering the events and the voting ABI. The contract will be cheaper to deploy and use without this module.
```
frangio, line 23: I think it is "opt in _to_". Also maybe we should add "some" to be super clear.

```suggestion
An OpenZeppelin Governor contract is by default not interface-compatible with GovernorAlpha or Bravo, since some of the functions are different or missing, although it shares all of the same events. However, it’s possible to opt in to some compatibility by inheriting from the GovernorCompatibilityBravo module, which is designed for covering the events and the voting ABI. The contract will be cheaper to deploy and use without this module.
```

Can you elaborate a bit more (in the docs) about "designed for covering the events and the voting ABI"? I thought all events were already compatible in the core Governor, as stated in the previous sentence.

docs/modules/ROOT/pages/governance.adoc
```diff
@@ -20,7 +20,9 @@ The ERC20 extension to keep track of votes and vote delegation is one such case.
...
+
+Note that the compatibility provided doesn't override the OpenZeppelin Governor implementation, that comes with differences in the way `proposalId`s are calculated. Governor stores the hash of the proposal with the purpose of keeping its data off-chain by event indexing, while the original Bravo implementation stores sequential `proposalId`s.
```
frangio, line 25:
```suggestion
Note that the compatibility provided doesn't override the OpenZeppelin Governor implementation, that comes with differences in the way `proposalId`s are calculated. Governor stores the hash of the proposal with the purpose of keeping its data off-chain by event indexing, while the original Bravo implementation uses sequential `proposalId`s.
```

docs/modules/ROOT/pages/governance.adoc
ernestognw: I'm trying to synthesize [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3877#issuecomment-1387088832) from @Amxx, I think it made sense to me in the context of the issue conversation, but I'll try to port it to here, will update.

Amxx: There might be a misunderstanding here.

Event & voting ABI compatibility is built in the core, and does not require the `GovernorCompatibilityBravo` module. The module adds some more compatibility (in particular the propose/execute functions, and the getters)

PR #3812, opened 2022-11-13
Adds certora spec for Ownable2Step

spalladino:
Adds a certora spec for formally verifying `Ownable2Step`. Includes the rules:
- `transferOwnershipSetsPendingOwner`
- `onlyCurrentOwnerCanCallOnlyOwner`: verifies that transfer, renounce, and user functions decorated with onlyOwner can indeed only be called by the current owner
- `onlyPendingOwnerCanAccept`
- `renounceDoesNotRequireTwoSteps`

PR #4094, opened 2023-03-07
Add FV specs for Ownable and Ownable2Steps

Amxx:
Replaces & extend #3812
Replaces #4092


certora/specs/Ownable.spec
```diff
@@ -0,0 +1,75 @@
...
+    transferOwnership@withrevert(e, newOwner);
+    bool success = !lastReverted;
+
+    assert success <=> (e.msg.sender == current && newOwner != 0), "unauthorized caller or invalid arg";
+    assert success => owner() == newOwner, "current owner changed";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Function correctness: renounceOwnership removes the owner and the pendingOwner                                      │
```
frangio, line 28:
```suggestion
│ Function correctness: renounceOwnership removes the owner                                                           │

```

certora/specs/Ownable.spec
```diff
@@ -0,0 +1,75 @@
...
+    calldataarg args;
+    f(e, args);
+
+    address newCurrent = owner();
+
+    // If owner changes, must be either transferOwnership or renounceOwnership
+    assert oldCurrent != newCurrent => (
+        (e.msg.sender == oldCurrent && newCurrent != 0 && f.selector == transferOwnership(address).selector) ||
+        (e.msg.sender == oldCurrent && newCurrent == 0 && f.selector == renounceOwnership().selector)
+    );
```
frangio, line 74: Could this be refactored into `if newCurrent == 0 { ... } else { ... }`? It sounds like it would be easier to understand.

certora/specs/Ownable2Step.spec
```diff
@@ -0,0 +1,106 @@
...
+    assert success => pendingOwner() == 0, "pending owner not cleared";
+    assert success => owner() == 0, "owner not cleared";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Function correctness: acceptOwnership changes owner and reset pending owner                                         │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule acceptOwnership(env e){
```
frangio, line 50:
```suggestion
rule acceptOwnership(env e) {

```

certora/specs/Ownable2Step.spec
```diff
@@ -0,0 +1,106 @@
...
+    calldataarg args;
+    f(e, args);
+
+    address newCurrent = owner();
+    address newPending = pendingOwner();
+
+    // If owner changes, must be either acceptOwnership or renounceOwnership
+    assert oldCurrent != newCurrent => (
+        (e.msg.sender == oldPending && newCurrent == oldPending && newPending == 0 && f.selector == acceptOwnership().selector) ||
+        (e.msg.sender == oldCurrent && newCurrent == 0          && newPending == 0 && f.selector == renounceOwnership().selector)
```
frangio, line 97: Do we need to state `newPending == 0` here? I'd assume it's already verified in the `acceptOwnership` rule.

Amxx: I added it here for exaustivity, but true, its not really necessary. Should we remove it here?

certora/specs/Ownable.spec
Amxx: Can you propose the change? I'm not sure what you have in mind.

certora/specs/Ownable2Step.spec
frangio: I would remove it.

Amxx: just in case we ever tamper with acceptOwnership, I'd rather have the check here. 

It doesn't cost must, and I think it capture the essence that "if the owner ever changes, then the pending owner is reset"

To have the same "proof", you otherwize need to combine stuff spread over 3 rules:
- if ownership changes, that means either acceptOwnership or renounceOwnership was called
  - if acceptOwnership was called, pending owner is reset
  - if renounceOwnership was called, pending owner is reset

certora/specs/Ownable.spec
frangio: I was thinking:

```
if (newCurrent == 0) { 
  assert oldCurrent != newCurrent => e.msg.sender == oldCurrent && f.selector == renounceOwnership().selector;
} else {
  assert oldCurrent != newCurrent => e.msg.sender == oldCurrent && f.selector == transferOwnership(address).selector
}
```

Based on your feedback I'm okay with the current version.

frangio (approved):
LGTM!

Amxx:
@gitpoap-bot @spalladino

PR #4097, opened 2023-03-07
Ability to set starting token id for ERC721Consecutive

kfishnchips:
Many of our contracts start with token id 1, ERC721Consecutive currently assumes the starting token id to be zero.

Add a constructor which allows setting the starting token id.

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx:
Hello @kfishnchips 

There is something troublesome with you implementation: `_ownerOf` will revert for tokenId below the `_startingTokenId`. This causes the corresponding tokens to not be mintable at all, including by `_mint`  later in the lifecycle of the contract.

Additionally, reading `_startingTokenId` makes all token operation that read the owner cost one more `sload`.

IMO, if this is a feature we want, we should implement it by allowing `_mintConsecutive` to address(0):
- we would not call the hook
- we would not increase the balance
- we would not emit an event
- we would still push a "checkpoint" to the ownership trace

That way you could create gaps in your allocation, not only before the first `_mintConsecutive` but also between multiple `_mintConsecutive` operation.

kfishnchips:
> Hello @kfishnchips
> 
> There is something troublesome with you implementation: `_ownerOf` will revert for tokenId below the `_startingTokenId`. This causes the corresponding tokens to not be mintable at all, including by `_mint` later in the lifecycle of the contract.
> 
> Additionally, reading `_startingTokenId` makes all token operation that read the owner cost one more `sload`.
> 
> IMO, if this is a feature we want, we should implement it by allowing `_mintConsecutive` to address(0):
> 
> * we would not call the hooks
> * we would not increase the balance
> * we would not emit an event
> * we would still push a "checkpoint" to the ownership trace
> 
> That way you could create gaps in your allocation, not only before the first `_mintConsecutive` but also between multiple `_mintConsecutive` operation.

Was it naive from me to think that minted tokens would only have token ids larger than the batch? Now that I think of it, you could mint batches with gaps and then have `_mint` fill those gaps.

kfishnchips:
`mintConsecutive` to `address(0)` would effectively "burn" the unwanted token ids, makes sense

kfishnchips:
@Amxx is this last commit somewhere along the lines of what you were thinking?

Amxx:
Thanks for changing your PR so fast. Usually we discuss the feature before going into an actual implementation. 

@frangio how do you feel about it?

test/token/ERC721/extensions/ERC721Consecutive.t.sol
```diff
@@ -28,21 +31,41 @@ contract ERC721ConsecutiveTarget is StdUtils, ERC721Consecutive {
     function burn(uint256 tokenId) public {
         _burn(tokenId);
     }
+
+    function maxBatchSize() public view returns (uint256) {
+        return _maxBatchSize();
+    }
```
Amxx, line 37: this isn't actually needed

test/token/ERC721/extensions/ERC721Consecutive.t.sol
```diff
@@ -16,7 +16,10 @@ function toSingleton(address account) pure returns (address[] memory) {
 contract ERC721ConsecutiveTarget is StdUtils, ERC721Consecutive {
     uint256 public totalMinted = 0;
 
-    constructor(address[] memory receivers, uint256[] memory batches) ERC721("", "") {
+    constructor(address[] memory receivers, uint256[] memory batches, uint96 startingTokenId) ERC721("", "") {
+        if (startingTokenId > 0) {
+            _mintConsecutive(address(0), startingTokenId);
+        }
```
Amxx, line 22: Afaik, this is covered by "normal" batches. We just need to make sure that receiver = 0 is considered

Amxx (commented):
Tests needs more checks, but please wait until we discuss that with @frangio and @ernestognw before putting work into it.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -87,24 +89,27 @@ abstract contract ERC721Consecutive is IERC2309, ERC721 {
         // minting a batch of size 0 is a no-op
         if (batchSize > 0) {
             require(!Address.isContract(address(this)), "ERC721Consecutive: batch minting restricted to constructor");
-            require(to != address(0), "ERC721Consecutive: mint to the zero address");
             require(batchSize <= _maxBatchSize(), "ERC721Consecutive: batch too large");
```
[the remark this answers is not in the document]
kfishnchips: We want to use this for an upcoming contract, hence the speed :) 

Is the batch size necessary since we are essentially minting the offset to the zero address?

Let me know if there is anything else I can do to help test.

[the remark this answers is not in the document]
Amxx: The max batch size makes sure batches can be processed by offchain tolling. Since these batches to 0 are not actually minted, and don't trigger any event, there is no issue with offchain tolling here.

frangio:
Allowing this makes sense to me.

However, I would not reuse `_mintConsecutive` for this. I would instead introduce a new function `_offsetConsecutive` (or similar name) that would only create the checkpoint. @Amxx what do you think?

Amxx:
`_offsetConsecutive` works for me, provided its documented well.

The performance implication is that the the trace array will be longer, making `ownerOf` lookup more expensive (growth is logarithmic in the number of checkpoints)

frangio:
That seems like a big price to pay for a feature that appears to be implementable essentially for free with a single immutable variable (or pure virtual function) to mark the start of the consecutive mints.

Amxx:
> That seems like a big price to pay for a feature that appears to be implementable essentially for free with a single immutable variable (or pure virtual function) to mark the start of the consecutive mints. Is there a reason I'm not seeing why that wouldn't work?

See the first part of [this message](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4097#issuecomment-1458619237).

TLDR:
- Depending on how the offset is implemented, we must make sure that all tokenId in the range are not blocked forever, and can later be minted using the normal `_mint` function. Also, I expect calls to `_ownerOf` to not revert. It might be possible to implement that cleanly, but the first draft of this PR was not clean in that regard.

> It's not exactly the same as _offsetConsecutive because it doesn't allow leaving gaps in between batches, but the feature that was requested by @kfishnchips was just to set the starting id.

While we are at it, don't we want to support gaps ? That sounds like a reasonable feature.

frangio:
Gaps sound "reasonable" but no one has requested it and I feel that we should favor the more efficient option here.

Amxx:
> Gaps sound "reasonable" but no one has requested it and I feel that we should favor the more efficient option here.

To me, it felt the efficient in the sens that it reuses existing logic, and doesn't add more logic to this contract.
An immutable initial offset might be more gas effective, but it also adds code complexity.

frangio:
Generally I agree but when we're talking about removing SSTORE/SLOAD, I think additional code complexity is justifiable. In particular this contract has the specific goal of being efficient.

Amxx:
Ok @kfishnchips lets go back to implementing this with a 
```solidity
function _consecutiveStart() internal view virtual returns (uint96) { return 0; }
```

We need to make sure that `_ownerOf` doesn't revert for tokens before the first batch

test/token/ERC721/extensions/ERC721Consecutive.t.sol
ernestognw: Although multiple tests use `vm.assume(receiver != address(0))`, they'll fail for `ownerOf` since [address zero is not a valid owner](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/eedca5d873a559140d79cc7ec674d0e28b2b6ebd/contracts/token/ERC721/ERC721.sol#L63), according to the ERC721 implementation.

I don't see an easy way around that, so the constructor if may be the easiest.

test/token/ERC721/extensions/ERC721Consecutive.t.sol
```diff
@@ -31,18 +34,34 @@ contract ERC721ConsecutiveTarget is StdUtils, ERC721Consecutive {
 }
 
 contract ERC721ConsecutiveTest is Test {
+    function test_starting_id(address receiver, uint256[] calldata batches, uint96 startingTokenId) public {
+        vm.assume(receiver != address(0));
+        vm.assume(startingTokenId > 0);
+        vm.assume(startingTokenId < 5000);
```
ernestognw, line 40: Can we use [bound](https://book.getfoundry.sh/reference/forge-std/bound?highlight=bound#bound) instead?

ernestognw (commented):
Just leaving some small suggestions, but I agree with the approach of an overrideable function so far.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -141,8 +141,15 @@ abstract contract ERC721Consecutive is IERC2309, ERC721 {
         super._afterTokenTransfer(from, to, firstTokenId, batchSize);
     }
 
+    /**
+     * @dev Used to offset the first token id in {_totalConsecutiveSupply}
+     */
+    function _firstConsecutiveId() internal view virtual returns (uint96) {
```
ernestognw, line 147: We need to make the `_firstConsecutiveId` doesn't affect boundaries in other functions.

For example, there's a range check of the consecutive range in [`ownerOf(...)`](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4097/files#diff-cf3d2b9b88440ac0f2ca0e32a16251bbd411cd98ea845549eb347668894212b7R59), that will be affected by the offset `_firstConsecutiveId` is adding.

We may need something like:

```solidity
if (owner != address(0) || tokenId > (type(uint96).max + _firstConsecutiveId()) || tokenId < _firstConsecutiveId()) {
  return owner;
}
```

Another example is the `_afterTokenTransfer()` function in [line 135](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4097/files#diff-cf3d2b9b88440ac0f2ca0e32a16251bbd411cd98ea845549eb347668894212b7R135), we'll need to make sure the `firstTokenId` was indeed a batch mint.

Maybe:

```solidity
if (
    to == address(0) && // if we burn
    firstTokenId >= _firstConsecutiveId() && firstTokenId - _firstConsecutiveId() < _totalConsecutiveSupply() // and the tokenId is in the batch range
    !_sequentialBurn.get(firstTokenId) // and the token was never marked as burnt
)
```

Although being honest, I don't like the amount of reads it needs

test/token/ERC721/extensions/ERC721Consecutive.t.sol
```diff
@@ -119,4 +152,29 @@ contract ERC721ConsecutiveTest is Test {
...
+        address receiver,
+        uint256[2] calldata unboundedBatches,
+        uint256[2] calldata unboundedTokenId,
+        uint96 startingId
+    ) public {
+        vm.assume(receiver != address(0));
+
+        uint256 startingTokenId = bound(startingId, 1, 5000);
+
+        // We assume _maxBatchSize is 5000 (the default). This test will break otherwise.
```
ernestognw, line 166: Why is this breaking? If it's breaking because of the `ownerOf` range check, we definitely need to address that.

ernestognw (commented):
We still need to address the consistency of the range checks when there's a `_firstConsecutiveId()` and add tests for that. Currently, although the `_maxBatchSize()` is 5000, there are missing edge cases to consider:

1. What happens if a regular `_mint(...)` is before the `_firstConsecutiveId()`?
2. Are we testing correctly the effects on `_afterTokenTransfer` and `ownerOf(...)`?

I added detailed comments with my thoughts, but still, thank you @kfishnchips for the fast response.

test/token/ERC721/extensions/ERC721Consecutive.t.sol
kfishnchips: The comment about the breaking test is from a previous commit which I still can't decipher

contracts/token/ERC721/extensions/ERC721Consecutive.sol
Amxx: > Although being honest, I don't like the amount of reads it needs

If the value is not in storage there is no read.

ernestognw: But there are still `JUMP`s, adding overhead in a contract meant for efficiency. Even if it's a few units of gas, in a batch of 5000 we'll see the impact.

Amxx: And yes, we definitelly need to skip the lookup in `_ownerOf` if the tokenId is < offset.
The part in `_afterTokenTransfer()` is a welcome optimisation but its not as critical (just to be clear we still want it!)

kfishnchips: > We need to make the `_firstConsecutiveId` doesn't affect boundaries in other functions.
> 
> For example, there's a range check of the consecutive range in [`ownerOf(...)`](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4097/files#diff-cf3d2b9b88440ac0f2ca0e32a16251bbd411cd98ea845549eb347668894212b7R59), that will be affected by the offset `_firstConsecutiveId` is adding.
> 
> We may need something like:
> 
> ```solidity
> if (owner != address(0) || tokenId > (type(uint96).max + _firstConsecutiveId()) || tokenId < _firstConsecutiveId()) {
>   return owner;
> }
> ```
> 
> Another example is the `_afterTokenTransfer()` function in [line 135](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4097/files#diff-cf3d2b9b88440ac0f2ca0e32a16251bbd411cd98ea845549eb347668894212b7R135), we'll need to make sure the `firstTokenId` was indeed a batch mint.
> 
> Maybe:
> 
> ```solidity
> if (
>     to == address(0) && // if we burn
>     firstTokenId >= _firstConsecutiveId() && firstTokenId - _firstConsecutiveId() < _totalConsecutiveSupply() // and the tokenId is in the batch range
>     !_sequentialBurn.get(firstTokenId) // and the token was never marked as burnt
> )
> ```
> 
> Although being honest, I don't like the amount of reads it needs

I ended up having to cast to not get an arithmetic overflow/underflow error 
```solidity
if (owner != address(0) || tokenId > (uint256(type(uint96).max) + _firstConsecutiveId()) || tokenId < _firstConsecutiveId()) {
        return owner;
}
```

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -141,8 +141,15 @@ abstract contract ERC721Consecutive is IERC2309, ERC721 {
...
+        return 0;
+    }
+
     function _totalConsecutiveSupply() private view returns (uint96) {
         (bool exists, uint96 latestId, ) = _sequentialOwnership.latestCheckpoint();
-        return exists ? latestId + 1 : 0;
+        return exists ? latestId + 1 : _firstConsecutiveId();
```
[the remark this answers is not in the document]
kfishnchips: Would requiring that the tokenId be greater than the last batch minted one in the `_mint` function be a viable option? It might be too inefficient... 
```solidity
    function _mint(address to, uint256 tokenId) internal virtual override {
        require(Address.isContract(address(this)), "ERC721Consecutive: can't mint during construction");
        (, uint96 latestId, ) = _sequentialOwnership.latestCheckpoint();
        require(tokenId > latestId, "ERC721Consecutive: token id must be greater than latest id");
        super._mint(to, tokenId);
    }
```

kfishnchips:
```solidity
    function _ownerOf(uint256 tokenId) internal view virtual override returns (address) {
        address owner = super._ownerOf(tokenId);

        // If token is owned by the core, or beyond consecutive range, return base value
        if (owner != address(0) || tokenId > (uint256(type(uint96).max) + _firstConsecutiveId()) || tokenId < _firstConsecutiveId()) {
            return owner;
        }

        // Otherwise, check the token was not burned, and fetch ownership from the anchors
        // Note: no need for safe cast, we know that tokenId <= type(uint96).max
        return _sequentialBurn.get(tokenId) ? address(0) : address(_sequentialOwnership.lowerLookup(uint96(tokenId)));
    }
```

if `address owner = super._ownerOf(tokenId);` returns `address(0)` shouldn't we revert with `"ERC721: invalid token ID"`?

```solidity
    function tokenURI(uint256 tokenId) public view virtual override returns (string memory) {
        _requireMinted(tokenId);

        string memory baseURI = _baseURI();
        return bytes(baseURI).length > 0 ? string(abi.encodePacked(baseURI, tokenId.toString())) : "";
    }
```

for some reason calling `tokenURI` with any token id is returning a string even for non existent IDs, I suspect it's because `_exists` calls `_ownerOf` from ERC721Consecutive since we have it overriden

```solidity
function _exists(uint256 tokenId) internal view virtual returns (bool) {
        return _ownerOf(tokenId) != address(0);
    }
```

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -56,7 +56,11 @@ abstract contract ERC721Consecutive is IERC2309, ERC721 {
         address owner = super._ownerOf(tokenId);
 
         // If token is owned by the core, or beyond consecutive range, return base value
-        if (owner != address(0) || tokenId > type(uint96).max) {
+        if (
+            owner != address(0) ||
+            tokenId > (uint256(type(uint96).max) + _firstConsecutiveId()) ||
```
Amxx, line 61: this feels wrong. The lookup checkpoint (`_sequentialOwnership`)  only support id that are uint96. So if a tokenId is greater than that, we should not try to do the lookup.

said otherwize, `_firstConsecutiveId` should be reducing the range of tokens that can be sequentially minted, not moving it.

kfishnchips: in that case we should not need `+ _firstConsecutiveId()` since it overflows, I'll remove that and test

Amxx:
IMO things would be way simpler if we rename `_totalConsecutiveSupply` to `_nextConsecutiveId` and implement it as 

```
    function _totalConsecutiveSupply() private view returns (uint96) {
        (bool exists, uint96 latestId, ) = _sequentialOwnership.latestCheckpoint();
        return exists ? latestId + 1 : _firstConsecutiveId();
    }
```

This is a private function, so that is a change we can do.

kfishnchips:
By the way, thanks for all the reviews you guys rock!

Amxx:
I fixed a small thing an updated the unit tests (not the fuzzing though)

contracts/token/ERC721/extensions/ERC721Consecutive.sol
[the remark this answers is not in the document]
Amxx: That is something we don't want, because it adds 2 (cold) sload to the mint operation, which we want to keep as cheap as possible.

Also, if we do that, it would be impossible to `_mint` a token that was first minted as part of a batch then burnt.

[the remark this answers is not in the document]
Amxx: note that mint already check if the tokenId has an owner, and revert if that is the case.

.changeset/spotty-hotels-type.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`ERC721Consecutive`: Add support for setting a starting token id
```
Amxx, line 5:
```suggestion
`ERC721Consecutive`: Add a `_firstConsecutiveId` internal function that can be overridden to change the id of the first token minted through `_mintConsecutive`.
```

test/token/ERC721/extensions/ERC721Consecutive.t.sol
Amxx: The test tries to mint a batch that is too large. too large means bigger than `_maxBatchSize`. Since the contract is not yet constructed, we assume the value is 5000. If you ever override `_maxBatchSize` in `ERC721ConsecutiveTarget` , you should make sure it is reflected here.

kfishnchips: Ah, understood. Thanks for clarifying!

contracts/token/ERC721/extensions/ERC721Consecutive.sol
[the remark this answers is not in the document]
kfishnchips: I see, I'm still wrapping my head around the internals of this but I think I got it.

kfishnchips:
We are successfully using this version of ERC721Consecutive in a collection we are developing and so far it has passed our test suite which is composed of about 120 tests using forge and fuzzing, we are continuing to add tests and checking for edge cases.

Amxx:
> > Ok @kfishnchips lets go back to implementing this with a
> > 
> > ```solidity
> > function _firstConsecutiveId() internal view virtual returns (uint96) { return 0; }
> > ```
> > 
> > We need to make sure that `_ownerOf` doesn't revert for tokens before the first batch
> 
> Wouldn't we need to revert for `_ownerOf` of tokens before the starting token id? Say if the `_firstConsecutiveId()` returns 1, would checking ownership of token id 0 revert?

No! _ownerOf should never revert !

OwnerOf must either return the owner (if the token is minted) or 0 (if the token is not minted). 

The way it works in the context of ERC721Consecutive is pretty simple:
- check if the token is controled by the core (super call)
- if that is the case, use that value
- if that is not the case, and if the token was minted in a batch, fallback to the checkpoint system.

If the token is before the first consecutive, we should do the super call, and return that because the token is for sure not part of a batch.

Amxx (dismissed):
It looks mostly good for me it its current format. We need a second approval from the team.

kfishnchips:
Just FYI we launched our contract with this and it went super well.

kfishnchips:
@ernestognw @Amxx what is missing so we can merge this? Anything I can do?

kfishnchips:
bump

kfishnchips:
@ernestognw @Amxx wondering if you folks are  still interested in this PR

ernestognw:
Hey @kfishnchips, thanks for bumping. We've been focused on improving our advanced testing processes, launching 4.9, and currently working on 5.0 so this completely went out of my radar. My approval is still there and previously @Amxx approved with further relevant changes so I'm merging.

PR #4098, opened 2023-03-07
Add FV specs for TimelockController

Amxx:


certora/specs/TimelockController.spec
```diff
@@ -0,0 +1,278 @@
...
+
+    assert !isExecutorOrOpen => lastReverted, "Only executor can execute";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Access control: only users with the CANCELLER_ROLE can call cancel                                                  │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule onlyCanceler(env e, bytes32 id, method f) filtered { f ->
```
frangio, line 269: "Canceler" is missing an `l` in this rule.

frangio (commented):
It feels like this spec is missing "liveness" checks. If an operation is unset it can be scheduled, if an operation is ready it can be executed.

certora/specs/TimelockController.spec
```diff
@@ -0,0 +1,278 @@
...
+    (isDone(id)    <=> (!isUnset(id)   && !isPending(id))) &&
+    isOperationReady(e, id) => isPending(id)
+    filtered { f -> !f.isView }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: state transition rules                                                                                        │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule stateTransition(bytes32 id, env e, method f, calldataarg args) {
```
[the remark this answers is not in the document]
Amxx: That is a possibility. I considered many ways of writing it. If you feel like the current syntax is not clear enough and you have something to propose, fell free!

Amxx:
I rewrote the "schedule", "execute" and "cancel" rules to also do liveness. They check under which circumstances a revert is acceptable. This includes missing roles, so I removed the "access control" rules that are now covered here.

certora/specs/TimelockController.spec
```diff
@@ -128,151 +166,91 @@ rule minDelayOnlyChange(env e) {
...
+    bool  isProposerBefore  = hasRole(PROPOSER_ROLE(), e.msg.sender);
+
+    helperScheduleWithRevert(e, f, id, delay);
+
+    if (lastReverted) {
+        assert stateBefore != UNSET() || !isDelaySufficient || !isProposerBefore, "Unexpected revert";
+    } else {
+        assert stateBefore == UNSET() && state(id) == PENDING(), "State transition violation";
+        assert isProposerBefore, "Unauthorized proposal";
+        assert isDelaySufficient, "Minimum delay violation";
```
frangio: I believe that this is equivalent, but in my opinion it expresses the property more explicitly and clearly. What do you think?

```suggestion
    bool success = !lastRevert;

    assert success <=> stateBefore == UNSET() && isDelaySufficient && isProposerBefore, "Unexpected revert";

    if (success) {
        assert state(id) == PENDING(), "State transition violation";
```

certora/specs/TimelockController.spec
```diff
@@ -128,151 +166,91 @@ rule minDelayOnlyChange(env e) {
...
 
-    calldataarg args;
-    f@withrevert(e, args);
+    cancel@withrevert(e, id);
 
-    assert !isCanceler => lastReverted, "Only canceler can cancel";
+    if (lastReverted) {
+        assert stateBefore != PENDING() || !isCanceller, "Unexpected revert";
+    } else {
+        assert stateBefore == PENDING() && state(id) == UNSET(), "State transition violation";
```
frangio: Same question/suggestion as above.

```suggestion
    bool success = !lastReverted;

    assert success <=> (stateBefore == PENDING() && isCanceller), "Unexpected revert";

    if (success) {
```

certora/specs/TimelockController.spec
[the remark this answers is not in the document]
frangio: The current syntax is very clear and I like it, and as far as I can tell it is exhaustive. But the tool isn't helping us to check that we are exhaustively checking all possible before and after state combinations.

I was thinking that if we write something like this, we can be sure because of all the final `else` statements that there are no combinations that we missed. I definitely prefer the current spec aesthetically though.

```
if (stateBefore == UNSET()) {
    if (stateAfter == PENDING()) {
        // UNSET → PENDING: schedule or scheduleBatch
        assert (
            f.selector == schedule(address, uint256, bytes, bytes32, bytes32, uint256).selector ||
            f.selector == scheduleBatch(address[], uint256[], bytes[], bytes32, bytes32, uint256).selector
        );
    } else {
        assert stateAfter == UNSET();
    }
} else if (stateBefore == PENDING()) {
    if (stateAfter == UNSET()) {
        // PENDING → UNSET: cancel
        assert stateBefore == PENDING() && stateAfter == UNSET() => (
            f.selector == cancel(bytes32).selector
        );
    } else if (stateAfter == DONE()) {
        // PENDING → DONE: execute or executeBatch
        assert stateBefore == PENDING() && stateAfter == DONE() => (
            f.selector == execute(address, uint256, bytes, bytes32, bytes32).selector ||
            f.selector == executeBatch(address[], uint256[], bytes[], bytes32, bytes32).selector
        );
    } else {
        assert stateAfter == PENDING();
    }
} else if (stateBefore == DONE()) {
    // DONE is final
    assert stateAfter == DONE();
} else {
    assert false;
}
```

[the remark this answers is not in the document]
Amxx: If we had `switch` it would be so clean

certora/specs/TimelockController.spec
```diff
@@ -26,10 +27,44 @@ methods {
...
+        address target; uint256 value; bytes data; bytes32 predecessor; bytes32 salt;
+        require hashOperation(target, value, data, predecessor, salt) == id; // Correlation
+        schedule@withrevert(e, target, value, data, predecessor, salt, delay);
+    } else if (f.selector == scheduleBatch(address[], uint256[], bytes[], bytes32, bytes32, uint256).selector) {
+        address[] targets; uint256[] values; bytes[] payloads; bytes32 predecessor; bytes32 salt;
+        require hashOperationBatch(targets, values, payloads, predecessor, salt) == id; // Correlation
+        scheduleBatch@withrevert(e, targets, values, payloads, predecessor, salt, delay);
+    } else {
+        calldataarg args;
+        f@withrevert(e, args);
```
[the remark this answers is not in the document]
Amxx: I used to have `assert false` here, and it caused the specs to fail. I also find it weird, but this is how certora did it in the GovernorBase.spec.

certora/specs/TimelockController.spec
```diff
@@ -0,0 +1,275 @@
...
```
Amxx, line 45:
```suggestion
        require false;
```

certora/specs/TimelockController.spec
```diff
@@ -0,0 +1,274 @@
...
+        address target; uint256 value; bytes data; bytes32 salt;
+        require hashOperation(target, value, data, predecessor, salt) == id; // Correlation
+        execute@withrevert(e, target, value, data, predecessor, salt);
+    } else if (f.selector == executeBatch(address[], uint256[], bytes[], bytes32, bytes32).selector) {
+        address[] targets; uint256[] values; bytes[] payloads; bytes32 salt;
+        require hashOperationBatch(targets, values, payloads, predecessor, salt) == id; // Correlation
+        executeBatch@withrevert(e, targets, values, payloads, predecessor, salt);
+    } else {
+        calldataarg args;
+        f@withrevert(e, args);
```
Amxx, line 60:
```suggestion
        require false;
```

frangio:
I've been thinking about the `require false` thing and I don't feel comfortable with it, because it can cause vacuity if we're not careful...

Even though it doesn't make sense to us, I think we should revert back to the previous code. At least it will not remove branches from verification.

frangio (approved):

PR #4099, opened 2023-03-08
Add effect checks on the accesscontrol specs

Amxx:
During today's call, we noticed that the rules did not actually verify that the effect was applied when calls are successfull.

This PR adds that missing property, and rewrite the other for improved readability.

### DIFF
--- a/certora/specs/AccessControl.spec
+++ b/certora/specs/AccessControl.spec
@@ -16,20 +16,20 @@ methods {
 rule onlyGrantCanGrant(env e, bytes32 role, address account) {
     method f; calldataarg args;
 
-    bool hasRoleBefore = hasRole(role, account);
+    bool hasOtherRoleBefore = hasRole(role, account);
     f(e, args);
-    bool hasRoleAfter = hasRole(role, account);
+    bool hasOtherRoleAfter = hasRole(role, account);
 
     assert (
-        !hasRoleBefore &&
-        hasRoleAfter
+        !hasOtherRoleBefore &&
+        hasOtherRoleAfter
     ) => (
         f.selector == grantRole(bytes32, address).selector
     );
 
     assert (
-        hasRoleBefore &&
-        !hasRoleAfter
+        hasOtherRoleBefore &&
+        !hasOtherRoleAfter
     ) => (
         f.selector == revokeRole(bytes32, address).selector ||
         f.selector == renounceRole(bytes32, address).selector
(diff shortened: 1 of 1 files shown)


certora/specs/AccessControl.spec
```diff
@@ -44,22 +44,25 @@ rule onlyGrantCanGrant(env e, bytes32 role, address account) {
...
-    assert (
-        hasRoleBefore != hasRoleAfter
-    ) => (
-        hasRoleAfter == true && role1 == role2 && account1 == account2
-    );
+    // effect
+    assert success => hasRole(role, account);
+
+    // no side effect
+    assert hasRoleBefore != hasRole(otherRole, otherAccount) => (otherRole == role && otherAccount == account);
```
frangio, line 65: I think this is clearer:

```suggestion
    bool hasRoleAfter = hasRole(otherRole, otherAccount);
    assert hasRoleBefore != hasRoleAfter => (otherRole == role && otherAccount == account);

```

Amxx: addressed

certora/specs/AccessControl.spec
```diff
@@ -44,22 +44,27 @@ rule onlyGrantCanGrant(env e, bytes32 role, address account) {
...
 
-    bool isCallerAdmin = hasRole(getRoleAdmin(role2), e.msg.sender);
-    bool hasRoleBefore = hasRole(role1, account1);
+    bool isCallerAdmin = hasRole(getRoleAdmin(role), e.msg.sender);
+    bool hasOtherRoleBefore = hasRole(otherRole, otherAccount);
 
-    grantRole@withrevert(e, role2, account2);
-    assert !lastReverted <=> isCallerAdmin;
+    grantRole@withrevert(e, role, account);
+    bool success = !lastReverted;
```
ernestognw, line 56: Is `!lastReverted` used for meaning "no reversion"?

I just want to confirm, feels weird that: `no reversion = !lastReverted`

certora/specs/AccessControl.spec
```diff
@@ -44,22 +44,27 @@ rule onlyGrantCanGrant(env e, bytes32 role, address account) {
...
-        hasRoleAfter == true && role1 == role2 && account1 == account2
-    );
+    // liveness
+    assert success <=> isCallerAdmin;
+
+    // effect
+    assert success => hasRole(role, account);
+
+    // no side effect
+    assert hasOtherRoleBefore != hasOtherRoleAfter => (role == otherRole && account == otherAccount);
```
ernestognw, line 67: Should we to some degree rely on naming for _commenting_ on the assertions?

```suggestion
    bool hasOtherRoleChanged = hasOtherRoleBefore != hasOtherRoleAfter
    assert hasOtherRoleChanged => (role == otherRole && account == otherAccount);
```

certora/specs/AccessControl.spec
frangio: You should read "last reverted" as "has the last call reverted?". If the answer is "false", then it means the last call succeeded. Thus, `success = !lastReverted`.

certora/specs/AccessControl.spec
frangio: In this particular case, in my opinion `hasOtherRoleBefore != hasOtherRoleAfter` is sufficiently clear.

frangio (approved):
LGTM

PR #4100, opened 2023-03-08
Add FV specification for ERC20Wrapper

Amxx:
Certora specs for the ERC20 wrapper, rewritten almost from scratch

### DIFF
--- a/certora/run.js
+++ b/certora/run.js
@@ -64,10 +64,7 @@ async function runCertora(spec, contract, files, options = []) {
   stream.end();
 
   // write results in markdown format
-  const url = (await output).match(/https:\S*/)?.[0];
-  if (url) {
-    writeEntry(spec, contract, code || signal, url);
-  }
+  writeEntry(spec, contract, code || signal, (await output).match(/https:\S*/)?.[0]);
 
   // write all details
   console.error(`+ certoraRun ${args.join(' ')}\n` + (await output));
@@ -105,8 +102,8 @@ function writeEntry(spec, contract, success, url) {
       spec,
       contract,
       success ? ':x:' : ':heavy_check_mark:',
-      `[link](${url})`,
-      `[link](${url?.replace('/jobStatus/', '/output/')})`,
+      url ? `[link](${url})`: 'error',
+      url ? `[link](${url?.replace('/jobStatus/', '/output/')})` : 'error',
     ),
   );
 }
--- a/certora/run.js
+++ b/certora/run.js
@@ -31,7 +31,7 @@ if (request) {
 }
 
 for (const { spec, contract, files, options = [] } of Object.values(specs)) {
-  limit(runCertora, spec, contract, files, [...options.split(' '), ...extraOptions]);
+  limit(runCertora, spec, contract, files, [...options.flatMap(opt => opt.split(' ')), ...extraOptions]);
 }
 
 // Run certora, aggregate the output and print it at the end
--- a/certora/run.js
+++ b/certora/run.js
@@ -102,7 +102,7 @@ function writeEntry(spec, contract, success, url) {
       spec,
       contract,
       success ? ':x:' : ':heavy_check_mark:',
-      url ? `[link](${url})`: 'error',
+      url ? `[link](${url})` : 'error',
       url ? `[link](${url?.replace('/jobStatus/', '/output/')})` : 'error',
     ),
   );
--- a/certora/run.js
+++ b/certora/run.js
@@ -24,7 +24,7 @@ const [reqSpec, reqContract] = request.split(':').reverse();
 
 for (const { spec, contract, files, options = [] } of Object.values(specs)) {
   if ((!reqSpec || reqSpec === spec) && (!reqContract || reqContract === contract)) {
-    limit(runCertora, spec, contract, files, [...options, ...extraOptions]);
+    limit(runCertora, spec, contract, files, [...options.flatMap(opt => opt.split(' ')), ...extraOptions]);
   }
 }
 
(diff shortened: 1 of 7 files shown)


certora/specs/ERC20Wrapper.spec
```diff
@@ -0,0 +1,190 @@
...
+    totalSupply() <= underlyingBalanceOf(currentContract)  &&
+    underlyingBalanceOf(currentContract) <= underlyingTotalSupply() &&
+    underlyingTotalSupply() <= max_uint256
+    {
+        preserved {
+            requireInvariant totalSupplyIsSumOfBalances;
+            require underlyingBalanceOf(currentContract) <= underlyingTotalSupply();
+        }
+        preserved depositFor(address account, uint256 amount) with (env e) {
+            require e.msg.sender != currentContract;
```
frangio, line 45: I believe we should encode this assumption into the Solidity code. Gas cost shouldn't be significant, but we can check in the gas report.

certora/specs/ERC20Wrapper.spec
```diff
@@ -0,0 +1,190 @@
...
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant totalSupplyIsSmallerThanUnderlyingBalance()
+    totalSupply() <= underlyingBalanceOf(currentContract)  &&
+    underlyingBalanceOf(currentContract) <= underlyingTotalSupply() &&
+    underlyingTotalSupply() <= max_uint256
+    {
+        preserved {
+            requireInvariant totalSupplyIsSumOfBalances;
+            require underlyingBalanceOf(currentContract) <= underlyingTotalSupply();
```
frangio, line 42: Can't we write:

```suggestion
            requireInvariant totalSupplyIsSmallerThanUnderlyingBalance;
```

certora/specs/ERC20Wrapper.spec
```diff
@@ -0,0 +1,190 @@
...
+}
+
+use invariant totalSupplyIsSumOfBalances
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Helper: consequence of `totalSupplyIsSumOfBalances` applied to underlying                                           │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+function sumOfUnderlyingBalancesLowerThanUnderlyingSupply(address a, address b) {
```
frangio, line 22: Can we call this helper `requireSumOf...`?

It's very important to make it explicit that it's adding an assumption.

certora/specs/ERC20Wrapper.spec
```diff
@@ -0,0 +1,190 @@
...
+        receiver != 0                             && // invalid receiver
+        amount   <= senderUnderlyingBalanceBefore && // deposit doesn't exceed balance
+        amount   <= senderUnderlyingAllowanceBefore  // deposit doesn't exceed allowance
+    );
+
+    // effects
+    assert success => balanceOf(receiver)                  == balanceBefore                  + amount;
+    assert success => totalSupply()                        == supplyBefore                   + amount;
+    assert success => underlyingBalanceOf(currentContract) == wrapperUnderlyingBalanceBefore + amount;
+    assert success => underlyingBalanceOf(sender)          == senderUnderlyingBalanceBefore  - amount;
```
frangio, line 95: Why write this as multiple asserts as opposed to a single `assert success => (... && ...)` ?

frangio, line 95: I don't think the alignment is helping. It's a lot of spacing. This just looks better to me:

```suggestion
    assert success => balanceOf(receiver) == balanceBefore + amount;
    assert success => totalSupply() == supplyBefore + amount;
    assert success => underlyingBalanceOf(currentContract) == wrapperUnderlyingBalanceBefore + amount;
    assert success => underlyingBalanceOf(sender) == senderUnderlyingBalanceBefore - amount;
```

certora/run.js
```diff
@@ -44,7 +44,7 @@ async function runCertora(spec, contract, files, options = []) {
     const urls = data.toString('utf8').match(/https?:\S*/g);
     for (const url of urls ?? []) {
       if (url.includes('/jobStatus/')) {
-        console.error(`[${spec}] ${url}`);
+        console.error(`[${spec}] ${url.replace('/jobStatus/', '/output/')}`);
```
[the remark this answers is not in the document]
Amxx: yes, and it really is the go to page to see what is going on.

certora/specs/ERC20Wrapper.spec
Amxx: 👍🏻

certora/specs/ERC20Wrapper.spec
Amxx: Having multiple ones allow to give description to each one if they fail. Here the two are equivalent.

Amxx: It honestly looks unreadeable to me

Amxx: feel free to change the spacing/linting

certora/specs/ERC20Wrapper.spec
Amxx: we could do

```sumOfUnderlyingBalancesLowerThanUnderlyingSupply(currentContract, 0)```

certora/specs/ERC20Wrapper.spec
Amxx: we can, but its also important to notice that this assumption is a direct consequence of the `totalSupplyIsSumOfBalances` invariant on `underlying`. So its not an additional assumption

Amxx: I made it return a bool, so that you can do

`require sumOfUnderlyingBalancesLowerThanUnderlyingSupply(a, b);`

certora/specs/ERC20Wrapper.spec
frangio: > It a bit like using a variable inside its own declaration.

No, that is not the case according to the [docs on invariants](https://docs.certora.com/en/latest/docs/cvl/invariants.html#invariants-and-induction). The `preserve` block is included as a precondition, so by using a "recursive" `requireInvariant` in its own preserve block, we're just stating that this is an inductive property.

frangio (dismissed):

frangio (approved):

PR #4101, opened 2023-03-08
Throw error when requested specs are not found

frangio:
`node certora/run.js Foo:Bar` will now throw an error if `Foo` is not a valid harness or `Bar` is not a known spec.


Amxx (approved):
LGTM

PR #4102, opened 2023-03-09
Bump and pin Forge Std submodule

PaulRBerg:
Bumps and also pins `forge-std` to `v1`.

The benefit of pinning is that it leads to more predictable repo checkouts.

We use the `release-v4.8` branch (`solidity-generators`) in a project with many git submodules. Pinning your version of `forge-std` to `v1` helps reduce the installation time when pulling a fresh copy of the repo for the first time, because it leads to the same commit for `forge-std`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,5 @@
 [submodule "lib/forge-std"]
+	branch = v1
 	path = lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
 [submodule "lib/erc4626-tests"]


frangio:
I'm not sure I understand why this helps your setup, unless we always keep the commit in sync with the one you have checked out in your own repo. What is the command that you use to initialize the git submodules? Isn't the problem that you are doing a "recursive" clone of submodules?

PaulRBerg:
> unless we always keep the commit in sync with the one you have checked out in your own repo

This is exactly the goal here; pinning the submodule to the `v1` branch helps with that. Updating submodules will not change `forge-std` in your repo if you pin the branch to `v1` in `.gitmodules`.

There are fewer changes made to the `v1` branch of Forge Std than there are on `master`:

https://github.com/foundry-rs/forge-std/issues/275

frangio:
I see. Git submodules are so bad for dependency management... :sweat_smile:

I will merge this in `master` and cherry-pick on the release branch.

frangio (approved):

frangio:
Cherry-picked in 9eee01c5a2757fa9bf8421a2810776b885995d2f.

PaulRBerg:
Thanks for merging, @frangio!

>  I hope the v1 branch remains updated. They don't seem to have automated that.

Yeah @mds1 and I will make sure to do that 🫡

PR #4103, opened 2023-03-09
Credit YieldBox for virtual offset

frangio:
This PR adds a line in our ERC-4626 guide referencing YieldBox as the source of the virtual offset idea.

Fixes LIB-704.


Amxx (approved):
LGTM

PR #4106, opened 2023-03-10
Formal verification for the Governor system

Amxx:
WIP

### DIFF
--- a/certora/run.js
+++ b/certora/run.js
@@ -8,26 +8,27 @@
 
 const MAX_PARALLEL = 4;
 
-let specs = require(__dirname + '/specs.json');
-
 const proc = require('child_process');
 const { PassThrough } = require('stream');
 const events = require('events');
 const limit = require('p-limit')(MAX_PARALLEL);
+
+const strToRegex = str => new RegExp(`^${str.replace(/[.+?^${}()|[\]\\]/g, '\\$&').replace(/[*]/g, '.$&')}$`);
 
 let [, , request = '', ...extraOptions] = process.argv;
 if (request.startsWith('-')) {
   extraOptions.unshift(request);
   request = '';
 }
+const [reqSpec, reqContract] = request.split(':').reverse();
 
-if (request) {
-  const [reqSpec, reqContract] = request.split(':').reverse();
-  specs = Object.values(specs).filter(s => reqSpec === s.spec && (!reqContract || reqContract === s.contract));
-  if (specs.length === 0) {
-    console.error(`Error: Requested spec '${request}' not found in specs.json`);
-    process.exit(1);
-  }
+const specs = require(__dirname + '/specs.js')
+  .filter(entry => !reqSpec || strToRegex(reqSpec).test(entry.spec))
+  .filter(entry => !reqContract || strToRegex(reqContract).test(entry.contract));
+
+if (specs.length === 0) {
+  console.error(`Error: Requested spec '${request}' not found in specs.json`);
+  process.exit(1);
 }
 
 for (const { spec, contract, files, options = [] } of Object.values(specs)) {
--- a/certora/specs.js
+++ b/certora/specs.js
@@ -1,4 +1,7 @@
-[
+/// This helper will be handy when we want to do cross product. Ex: all governor specs on all variations of the clock mode.
+const product = (...arrays) => arrays.reduce((a, b) => a.flatMap(ai => b.map(bi => [ai, bi].flat())));
+
+module.exports = [
   {
     "spec": "AccessControl",
     "contract": "AccessControlHarness",
@@ -45,5 +48,18 @@
     "spec": "Initializable",
     "contract": "InitializableHarness",
     "files": ["certora/harnesses/InitializableHarness.sol"]
-  }
-]
+  },
+  ...[ "GovernorBase", "GovernorInvariants", "GovernorStates", "GovernorFunctions" ].map(spec => ({
+    spec,
+    "contract": "GovernorHarness",
+    "files": [
+      "certora/harnesses/GovernorHarness.sol",
+      "certora/harnesses/ERC20VotesHarness.sol"
+    ],
+    "options": [
+      "--link GovernorHarness:token=ERC20VotesHarness",
+      "--optimistic_loop",
+      "--optimistic_hashing"
+    ]
+  }))
+];
--- a/certora/specs.js
+++ b/certora/specs.js
@@ -1,4 +1,3 @@
-/// This helper will be handy when we want to do cross product. Ex: all governor specs on all variations of the clock mode.
 const product = (...arrays) => arrays.reduce((a, b) => a.flatMap(ai => b.map(bi => [ai, bi].flat())));
 
 module.exports = [
@@ -49,15 +48,18 @@ module.exports = [
     "contract": "InitializableHarness",
     "files": ["certora/harnesses/InitializableHarness.sol"]
   },
-  ...[ "GovernorBase", "GovernorInvariants", "GovernorStates", "GovernorFunctions" ].map(spec => ({
+  ...product(
+    [ "GovernorBase", "GovernorInvariants", "GovernorStates", "GovernorFunctions" ],
+    [ "ERC20VotesBlocknumberHarness", "ERC20VotesTimestampHarness" ],
+  ).map(([ spec, token ]) => ({
     spec,
     "contract": "GovernorHarness",
     "files": [
       "certora/harnesses/GovernorHarness.sol",
-      "certora/harnesses/ERC20VotesHarness.sol"
+      `certora/harnesses/${token}.sol`
     ],
     "options": [
-      "--link GovernorHarness:token=ERC20VotesHarness",
+      `--link GovernorHarness:token=${token}`,
       "--optimistic_loop",
       "--optimistic_hashing"
     ]
--- a/certora/specs/Governor.helpers.spec
+++ b/certora/specs/Governor.helpers.spec
@@ -45,7 +45,7 @@ definition voting(method f) returns bool =
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 function helperVoteWithRevert(env e, method f, uint256 pId, address voter, uint8 support) returns uint256 {
-    string reason; bytes params; uint8 v; bytes32 s; bytes32 r;
+    string reason; bytes params;
 
     if (f.selector == castVote(uint256,uint8).selector)
     {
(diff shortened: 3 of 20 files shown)


certora/run.js
```diff
@@ -8,26 +8,27 @@
...
 const MAX_PARALLEL = 4;
 
-let specs = require(__dirname + '/specs.json');
-
 const proc = require('child_process');
 const { PassThrough } = require('stream');
 const events = require('events');
 const limit = require('p-limit')(MAX_PARALLEL);
 
+const strToRegex = str => new RegExp(`^${str.replace(/[.+?^${}()|[\]\\]/g, '\\$&').replace(/[*]/g, '.$&')}$`);
```
frangio, line 16: We can just use the `micromatch` dependency instead of this. It's already declared in our dev dependencies.

certora/specs.js
```diff
@@ -0,0 +1,51 @@
+const product = (...arrays) => arrays.reduce((a, b) => a.flatMap(ai => b.map(bi => [ai, bi].flat())));
```
frangio, line 1:
```suggestion
const product = (...arrays) => arrays.reduce((a, b) => a.flatMap(ai => b.map(bi => [ai, bi])));
```

frangio: Ok I see it's for n > 2...

certora/run.js
Amxx: I'll check it out !

certora/specs.js
Amxx: what you propose limit the number of array to 2. The current version goes beyond:

```javascript
> product([1,2],[true,false],['a','b'])
[
  [ 1, true, 'a' ],
  [ 1, true, 'b' ],
  [ 1, false, 'a' ],
  [ 1, false, 'b' ],
  [ 2, true, 'a' ],
  [ 2, true, 'b' ],
  [ 2, false, 'a' ],
  [ 2, false, 'b' ]
]
```

Amxx: that can be usefull if you ever want to combine "specs" with "contract" with some link (like the token used)

frangio:
We need to merge `master` being careful to include the new specs, which at the moment seem to be `DoubleEndedQueue` and `ERC721`.

certora/specs/GovernorFunctions.spec
```diff
@@ -0,0 +1,277 @@
...
+*/
+rule propose_liveness(uint256 pId, env e) {
+    require nonpayable(e);
+    require clockSanity(e);
+
+    uint8 stateBefore = state(e, pId);
+
+    address[] targets; uint256[] values; bytes[] calldatas; string descr;
+    require pId == hashProposal(targets, values, calldatas, descr);
+    //require validString(descr);
```
frangio, line 21: Can this line be removed?

certora/specs/GovernorFunctions.spec
```diff
@@ -0,0 +1,277 @@
...
+    address[] targets; uint256[] values; bytes[] calldatas; string descr;
+    require pId == hashProposal(targets, values, calldatas, descr);
+
+    propose(e, targets, values, calldatas, descr);
+
+    // no side-effect
+    assert state(e, otherId)         != otherStateBefore => otherId == pId;
+    assert proposalSnapshot(otherId) == otherVoteStart;
+    assert proposalDeadline(otherId) == otherVoteEnd;
+    assert proposalProposer(otherId) == otherProposer;
```
frangio, line 60: Slightly confused, why do these work even for `otherId == pId`?

certora/specs/Governor.helpers.spec
```diff
@@ -0,0 +1,163 @@
...
+definition PENDING()   returns uint8 = 0;
+definition ACTIVE()    returns uint8 = 1;
+definition CANCELED()  returns uint8 = 2;
+definition DEFEATED()  returns uint8 = 3;
+definition SUCCEEDED() returns uint8 = 4;
+definition QUEUED()    returns uint8 = 5;
+definition EXPIRED()   returns uint8 = 6;
+definition EXECUTED()  returns uint8 = 7;
+
+function safeState(env e, uint256 pId) returns uint8 {
```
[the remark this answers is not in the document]
Amxx: added a comment

certora/specs/Governor.helpers.spec
```diff
@@ -0,0 +1,163 @@
...
+
+definition proposalCreated(uint256 pId) returns bool =
+    proposalSnapshot(pId) > 0 && proposalDeadline(pId) > 0 && proposalProposer(pId) != 0;
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Filters                                                                                                             │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+definition skip(method f) returns bool =
```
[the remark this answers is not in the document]
Amxx: Agree, but I don't have a better name on the top of my head.

This is a list of function that we don't consider when doing a "any function" rule. The point is to reduce the search space avoid some timeouts.

Example of usage:
```
rule immutableFieldsAfterProposalCreation(uint256 pId, env e, method f, calldataarg args)
    filtered { f -> !skip(f) }
{
...
}
```

certora/specs/Governor.helpers.spec
```diff
@@ -0,0 +1,163 @@
...
+    }
+    else
+    {
+        calldataarg args;
+        f(e, args);
+        return 0;
+    }
+}
+
+function helperFunctionsWithRevert(env e, method f, uint256 pId) {
```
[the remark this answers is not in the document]
Amxx: added a long comment

certora/specs/GovernorInvariants.spec
```diff
@@ -0,0 +1,151 @@
...
+*/
+invariant votesImplySnapshotPassed(env e, uint256 pId)
+    (
+        getAgainstVotes(pId) > 0 ||
+        getForVotes(pId)     > 0 ||
+        getAbstainVotes(pId) > 0
+    ) => proposalSnapshot(pId) < clock(e)
+    {
+        preserved with (env e2) {
+            require clock(e) == clock(e2);
```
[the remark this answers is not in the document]
Amxx: I'm not 100% sure, I'm starting to winder if it makes sens.

The issue is that all the envfree functions are not tied to a clock :/

certora/specs/GovernorStates.spec
```diff
@@ -0,0 +1,176 @@
...
+            (stateAfter == QUEUED()   && f.selector == queue(address[],uint256[],bytes[],bytes32).selector) ||
+            (stateAfter == EXECUTED() && f.selector == execute(address[],uint256[],bytes[],bytes32).selector)
+        ) &&
+        stateBefore == QUEUED() => (
+            (stateAfter == EXECUTED() && f.selector == execute(address[],uint256[],bytes[],bytes32).selector)
+        ) &&
+        stateBefore == ACTIVE()    => false &&
+        stateBefore == CANCELED()  => false &&
+        stateBefore == DEFEATED()  => false &&
+        stateBefore == EXECUTED()  => false
```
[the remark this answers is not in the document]
Amxx: I'm not sure about the precedence of
```
A && B || C && D || E && F
```
So I added additional `(`, `)`

certora/specs/GovernorStates.spec
```diff
@@ -0,0 +1,176 @@
...
+
+    uint48  currentClock  = clock(e);
+    uint8   currentState  = state(e, pId);
+    uint256 snapshot      = proposalSnapshot(pId);
+    uint256 deadline      = proposalDeadline(pId);
+    bool    quorumSuccess = quorumReached(pId);
+    bool    voteSuccess   = voteSucceeded(pId);
+
+    // Pending: before vote starts
+    assert currentState == PENDING() => (
```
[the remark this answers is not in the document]
Amxx: the goal here is to prove that some properties are true about each state. 
- If you are in state A, than X and Y
- If you are in state B, than Z 
- ...

There might indeed be cases where we can say
- If X and Y, than you are in state A
but where you cannot say 
- If Z, ... might be state B but might also be something else.

Should this rule remain consistent in how it treats all states, of should we prove more ?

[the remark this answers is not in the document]
Amxx: We can't say much about `quorumSuccess` because the quorum might be `= 0` (in which case no vote = success) or `> 0` (in which case no vote = no success)

For the `voteSuccess` its more complex. I order to say we don't have success, we need to "prove" that the votes counters are empty. This is done by `votesImplyCreated` with is part of `GovernorInvariants.spec`. If we want to prove it here, we need to import the invariant and require it. Note that this will cause all the runs of this spec file to re-verify the invariant.

certora/specs/GovernorFunctions.spec
Amxx: the real answer is: I don't know, this rule timeout so I can't really say much about it.

certora/specs/GovernorFunctions.spec
Amxx: In this rule we do `state(e, otherId)`. This means that `otherId` has a valid state, which means its been proposed (because state reverts if the proposal doesn't exist).

So we know `otherId` exists, and as been propose. 

`propose_liveness` is supposed (if it was working) to prove that `propose(...)` reverts if the proposal already exist. Since this rule does call propose without `@withrevert`, we know the propose worked, which implies `otherId == pId`.

So technically, we (expect to) know for sure that `otherId == pId`.

Amxx: I think these (failing) rules should be rewritten using `safeState`

certora/specs/GovernorStates.spec
```diff
@@ -68,18 +57,26 @@ rule stateTransitionWait(uint256 pId, env e1, env e2) {
...
+        (stateBefore == PENDING() && stateAfter == DEFEATED() ) ||
+        (stateBefore == ACTIVE()  && stateAfter == SUCCEEDED()) ||
+        (stateBefore == ACTIVE()  && stateAfter == DEFEATED() ) ||
+        // Strange consequence of the timelock binding:
+        // When transitioning from ACTIVE to SUCCEEDED (because of the clock moving forward) the proposal state in
+        // the timelock is suddenly considered. Prior state set in the timelock can cause the proposal to already be
+        // queued, executed or canceled.
+        (stateBefore == ACTIVE()  && stateAfter == CANCELED()) ||
+        (stateBefore == ACTIVE()  && stateAfter == EXECUTED()) ||
+        (stateBefore == ACTIVE()  && stateAfter == QUEUED())
```
frangio, line 79: Not sure what we should do here. This rule is currently not looking good, as we would never expect a proposal to go (for example) from pending to defeated automatically, I think...

If the rule was working before (using `=>`) we should probably keep it like that.

certora/specs/GovernorStates.spec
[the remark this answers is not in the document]
frangio: Let's keep the rule as it is now so we don't introduce new complexities at this point.

certora/specs/GovernorStates.spec
Amxx: I needed to add an additional "helper" invariant, but I managed to fix that part :)

Amxx:
Note: I wanted to do
```
node certora/run.js GovernorStates --options="--rule stateTransitionWait"
```
But that caused issues with options including spaces. So I moved the `.flatMap(opt => opt.split(' '))` to inside the `runCertora` function

@frangio can you see any issue with that ?

frangio:
No, that makes sense. I don't think it's ideal the way we're handling options but it works.

Amxx:
FYI: main (only?) remaining blocker is the `stateTransitionFn` function timming out.

It appears that it only reason it used to pass was because of poorly written formula and mistakes in the associativity of operators. Rewritting them cleanly (or adding parenthesis) causes it to timeout. An option could be to split that rule into multiple rules with limited logic in each one.

PR #4114, opened 2023-03-13
Update certora/README.md

ernestognw:
Updates documentation for using the Certora tooling after #4084

### DIFF
--- a/certora/Makefile
+++ b/certora/Makefile
@@ -3,9 +3,9 @@ default: help
 SRC   := ../contracts
 DST   := patched
 DIFF  := diff
-SRCS  := $(sh find $(SRC) -type f)
-DSTS  := $(sh find $(DST) -type f)
-DIFFS := $(sh find $(DIFF) -type f)
+SRCS  := $(shell find $(SRC) -type f)
+DSTS  := $(shell find $(DST) -type f)
+DIFFS := $(shell find $(DIFF) -type f)
 
 ###############################################################################
 # Apply all patches in the $DIFF folder to the $DST folder
--- a/certora/Makefile
+++ b/certora/Makefile
@@ -3,9 +3,9 @@ default: help
 SRC   := ../contracts
 DST   := patched
 DIFF  := diff
-SRCS  := $(shell find $(SRC) -type f)
-DSTS  := $(shell find $(DST) -type f)
-DIFFS := $(shell find $(DIFF) -type f)
+SRCS  := $(sh find $(SRC) -type f)
+DSTS  := $(sh find $(DST) -type f)
+DIFFS := $(sh find $(DIFF) -type f)
 
 ###############################################################################
 # Apply all patches in the $DIFF folder to the $DST folder
--- a/certora/README.md
+++ b/certora/README.md
@@ -6,7 +6,7 @@ Documentation for CVT and the specification language are available [here](https:
 
 ## Prerequisites
 
-Follow the [Certora installation guide](https://docs.certora.com/en/latest/docs/user-guide/getting-started/install.html) in order to get the Certora Prover Package and the `solc` executable folder in your path
+Follow the [Certora installation guide](https://docs.certora.com/en/latest/docs/user-guide/getting-started/install.html) in order to get the Certora Prover Package and the `solc` executable folder in your path.
 
 > **Note**
 > An API Key is required for local testing. Although the prover will run on a Github Actions' CI environment on each Pull Request.
--- a/certora/README.md
+++ b/certora/README.md
@@ -56,5 +56,5 @@ If the original contracts change, it is possible to create a conflict with the p
 For more information about the `make` scripts available, run:
 
 ```bash
-make -C certore help
+make -C certora help
 ```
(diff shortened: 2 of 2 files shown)


certora/Makefile
```diff
@@ -3,9 +3,9 @@ default: help
 SRC   := ../contracts
 DST   := patched
 DIFF  := diff
-SRCS  := $(shell find $(SRC) -type f)
-DSTS  := $(shell find $(DST) -type f)
-DIFFS := $(shell find $(DIFF) -type f)
+SRCS  := $(sh find $(SRC) -type f)
+DSTS  := $(sh find $(DST) -type f)
+DIFFS := $(sh find $(DIFF) -type f)
```
frangio, line 8: https://www.gnu.org/software/make/manual/html_node/Shell-Function.html

```suggestion
SRCS  := $(shell find $(SRC) -type f)
DSTS  := $(shell find $(DST) -type f)
DIFFS := $(shell find $(DIFF) -type f)
```

Did you run into an error with the original code?

certora/README.md
```diff
@@ -1,56 +1,60 @@
 # Running the certora verification tool
 
-These instructions detail the process for running CVT on the OpenZeppelin (Wizard/Governor) contracts.
+These instructions detail the process for running Certora Verification Tool on the OpenZeppelin contracts.
```
frangio, line 3:
```suggestion
These instructions detail the process for running Certora Verification Tool on OpenZeppelin Contracts.
```

certora/README.md
```diff
@@ -1,56 +1,60 @@
...
 
-Documentation for CVT and the specification language are available
-[here](https://certora.atlassian.net/wiki/spaces/CPD/overview)
+Documentation for CVT and the specification language are available [here](https://certora.atlassian.net/wiki/spaces/CPD/overview)
```
frangio, line 5:
```suggestion
Documentation for CVT and the specification language are available [here](https://certora.atlassian.net/wiki/spaces/CPD/overview).
```

certora/README.md
```diff
@@ -1,56 +1,60 @@
...
+
+## Prerequisites
+
+Follow the [Certora installation guide](https://docs.certora.com/en/latest/docs/user-guide/getting-started/install.html) in order to get the Certora Prover Package and the `solc` executable folder in your path
```
frangio, line 9:
```suggestion
Follow the [Certora installation guide](https://docs.certora.com/en/latest/docs/user-guide/getting-started/install.html) in order to get the Certora Prover Package and the `solc` executable folder in your path.
```

certora/README.md
```diff
@@ -1,56 +1,60 @@
...
+
+> **Note**
+> An API Key is required for local testing. Although the prover will run on a Github Actions' CI environment on each Pull Request.
```
frangio, line 12: We are not running it in every pull request to avoid too high resource usage.

```suggestion
> An API Key is required for local testing. Although the prover will run on a Github Actions' CI environment on select Pull Requests.
```

certora/README.md
```diff
@@ -1,56 +1,60 @@
...
 
-The most important of these is `verifyAll.sh`, which checks
-all of the harnessed contracts (`certora/harness/Wizard*.sol`) against all of
-the specifications (`certora/spec/*.spec`).
+Where:
 
-The other scripts run a subset of the specifications or the contracts.  You can
-verify different contracts or specifications by changing the `--verify` option,
-and you can run a single rule or method with the `--rule` or `--method` option.
+- `CONTRACT_NAME` matches the `contract` key in the `./spec.json` file and may be empty. It will run the all matching contracts if not provided.
```
frangio, line 28:
```suggestion
- `CONTRACT_NAME` matches the `contract` key in the `./spec.json` file and may be empty. It will run all matching contracts if not provided.
```

frangio (commented):
Thanks! Minor comments.

certora/Makefile
ernestognw: I did. I think this is equivalent and doesn't require extra configuration from my side.
I can do it on local if needed.

certora/README.md
ernestognw: Wouldn't be `selected Pull Requests`?

frangio: Both work grammatically but `selected` seems to have the right meaning.

certora/Makefile
frangio: Can you check what implementation and version of make you have? Is it GNU make?

ernestognw: Yes it is.

```shell
$ make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.
```

What doesn't seem to work is the find though:

```shell
$ make -C certora apply
find: diff: No such file or directory
```

frangio: I see, so that's unrelated, I've seen the same message in the logs for the workflow. Let's look at that in another PR.

frangio (approved):

frangio:
Thank you!

PR #4117, opened 2023-03-14
Add Pausable FV

ernestognw:
Fixes LIB-743

Adds Formal Verification rules for Pausable.


certora/specs/Pausable.spec
```diff
@@ -0,0 +1,93 @@
...
+
+    bool pausedBefore = paused();
+
+    pause@withrevert(e);
+    bool success = !lastReverted;
+
+    bool pausedAfter = paused();
+    
+    // liveness
+    assert success => !pausedBefore, "unpause call must succeed if the contract was unpaused before";
```
Amxx, line 25: liveness property should be equivalence `<=>`. 

Right now you are proving that is the call was succesfull, it wasn't paused before, but you are not proving that is the contract wasn't paused before, then the call doesn't revert

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,93 @@
...
+
+    // effect
+    assert success => (pausedBefore != pausedAfter && pausedAfter == true), "contract must be paused after a successful call";
```
Amxx, line 28: we already know `!pausedBefore` from line 25, so this only needs to check 

``assert success => pausedAfter;```

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,93 @@
...
+
+    // no side effect
+    assert !success => (pausedBefore == pausedAfter && pausedAfter == true), "contract must kept paused after a failed call";
```
Amxx, line 31: there is no such thing a side effect here. Side effect is if you wanted to check something else beside the paused status. if you fix the liveness, then this becomes irrelevant.

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,93 @@
...
+
+    bool pausedBefore = paused();
+
+    unpause@withrevert(e);
+    bool success = !lastReverted;
+
+    bool pausedAfter = paused();
+    
+    // liveness
+    assert success => pausedBefore, "pause call must succeed if the contract was paused before";
```
Amxx, line 50: same

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,93 @@
...
+
+    // effect
+    assert success => (pausedBefore != pausedAfter && pausedAfter == false), "contract must be unpaused after a successful call";
```
Amxx, line 53: same

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,93 @@
...
+
+    // no side effect
+    assert !success => (pausedBefore == pausedAfter && pausedAfter == false), "contract must kept unpaused after a failed call";
```
Amxx, line 56: same

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,93 @@
...
+    method f;
+    calldataarg args;
+
+    require paused();
+
+    bool pausedBefore = paused();
+    f(e, args);
+    bool pausedAfter = paused();
+
+    assert pausedBefore != pausedAfter => f.selector == unpause().selector, "contract must only be unpaused by _unpause()";
```
Amxx, line 92: these two rules can be merged into something like this

```
rule noUnpause(env e) {
    bool pausedBefore = paused();
    method f; calldataarg args; f(e, args);
    bool pausedAfter = paused();

    assert pausedBefore != pausedAfter => (
        (!pausedAfter && f.selector == unpause().selector) ||
        (pausedAfter && f.selector == pause().selector)
    );

File: certora/specs/Pausable.spec [hunk 8]
```diff
@@ -0,0 +1,93 @@
...
+│ Rules: only _unpause can unpause                                                                                    │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule noUnpause(env e) {
+    method f;
+    calldataarg args;
+
+    require paused();
+
+    bool pausedBefore = paused();
```
Amxx, line 88: if you require something, then there really is no need to assign it to a variable.

Amxx (changes requested):
Your on the right track, but you are missing specs for the `whenNotPaused` and `whenPaused` modifiers.

I would add some public functions like

```solidity
function onlyWhenPaused() external whenPaused() {}
function onlyWhenNotPaused() external whenNotPaused() {}
```

And show that they can only be called when the state is correct (liveness)

certora/specs/Pausable.spec
ernestognw: Got it, I didn't make it work at first and didn't try again. Now it does!

Is this "liveness property should be equivalence `<=>`" a common rule or something that just applies here?

certora/specs/Pausable.spec
ernestognw: Thanks!

Solving since I took the suggestion exactly as proposed.

certora/specs/Pausable.spec
ernestognw: I removed it but left the `no side effect` comment for keeping the same format, and explicitly indicating there are no checks.

certora/specs/Pausable.spec
ernestognw: Applied, thanks!

certora/specs/Pausable.spec
ernestognw: Just merged them

https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4117/commits/66f18ffb66f38480cdfa443262c7f49e038d8abc

ernestognw:
> I would add some public functions like
> 
> ```solidity
> function onlyWhenPaused() external whenPaused() {}
> function onlyWhenNotPaused() external whenNotPaused() {}
> ```

Just did, although, I set two assertions for liveness, but feels off to me. Is that okay?

.zip-output-url.txt
```diff
@@ -0,0 +1 @@
+https://prover.certora.com/zipOutput/71959/e1abf6518a0b48d280435917dc2174c3?anonymousKey=62ee1a919071c2ec4a3244879d76cd6f7b4df0db
```
Amxx, line 1: Please don't commit that. If you need, add it to `.gitignore`

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,114 @@
+import "helpers.spec"
+import "methods/IPausable.spec"
+
+methods {
+    pause()
+    unpause()
+}
```
Amxx, line 7:
```suggestion
methods {
    pause()
    unpause()
    onlyWhenPaused()
    onlyWhenNotPaused()
}

```

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,114 @@
...
+
+    bool pausedAfter = paused();
+    
+    // liveness
+    assert success <=> !pausedBefore, "works if and only if the contract was not paused before";
+
+    // effect
+    assert success => pausedAfter, "contract must be paused after a successful call";
+
+    // no side effect
```
Amxx, line 30:
```suggestion

```

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,114 @@
...
+
+    bool pausedAfter = paused();
+    
+    // liveness
+    assert success <=> pausedBefore, "works if and only if the contract was paused before";
+
+    // effect
+    assert success => !pausedAfter, "contract must be unpaused after a successful call";
+
+    // no side effect
```
Amxx, line 54:
```suggestion

```

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,114 @@
...
+*/
+rule whenPaused(env e) {
+    require nonpayable(e);
+
+    onlyWhenPaused@withrevert(e);
+    bool success = !lastReverted;
+    
+    // liveness
+    assert success <=> paused(), "works if and only if the contract is paused";
+    assert !success <=> !paused(), "fails if and only if the contract is not paused";
```
Amxx, line 70: This is redundant, you are writting the same thing twice

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,114 @@
...
+
+    // effect
+
+    // no side effect
+}
```
Amxx, line 75: We can compress all that to improve redeability
```suggestion
rule whenPaused(env e) {
    require nonpayable(e);

    onlyWhenPaused@withrevert(e);
    assert lastReverted <=> !paused(), "works if and only if the contract is paused";
}

```

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,114 @@
...
+    bool success = !lastReverted;
+    
+    // liveness
+    assert success <=> !paused(), "works if and only if the contract is not paused";
+    assert !success <=> paused(), "fails if and only if the contract is paused";
+
+    // effect
+
+    // no side effect
+}
```
Amxx, line 95: same
```suggestion
rule whenNotPaused(env e) {
    require nonpayable(e);
    
    onlyWhenNotPaused@withrevert(e);
    assert lastReverted <=> paused(), "works if and only if the contract is not paused";
}

```

certora/specs/Pausable.spec
```diff
@@ -0,0 +1,114 @@
...
+rule noPauseChange(env e) {
+    method f;
+    calldataarg args;
+
+    bool pausedBefore = paused();
+    f(e, args);
+    bool pausedAfter = paused();
+
+    assert pausedBefore != pausedAfter => (
+        !pausedAfter && f.selector == unpause().selector ||
```
Amxx, line 111: Just to be sure
```suggestion
        (!pausedAfter && f.selector == unpause().selector) ||

```

Amxx (commented):
Thats good! 

I made some minor cosmetic comments, but that is (mostly) about readeability. From a specification point of view, I think everything is here.

certora/specs/Pausable.spec
Amxx: If its easier for you, think of `<=>` like its a typed version of `==` that only compares boolean values.

certora/specs/methods/IPausable.spec
```diff
@@ -0,0 +1,3 @@
+methods {
+    paused() returns (bool) envfree
+}
```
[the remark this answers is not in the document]
ernestognw: I also think it's not needed, so I removed it and it's working as you suggested.

One question I have is, what are these files about if we're defining the interface inside each spec?

My feeling is that we should add `I*.spec` for standard interfaces (eg. `IERC20.spec` or `IERC5313.spec`) and import them where needed. 

An example would be: 

- I see `owner()` defined both in `IOwnable.spec` and `IOwnable2Step`, while they both can just import `IERC5313.spec`
- Pausable should not have an `IPausable.spec` because it doesn't adhere to any standard interface

Just a proposal, do you see this logic something we should follow more carefully?

certora/specs/Pausable.spec
ernestognw: I added the previous suggestion, and just got it while answering you.

Both instructions are exactly the same because of the "if and only if", right?

I'm solving

ernestognw:
> Thats good!
> 
> I made some minor cosmetic comments, but that is (mostly) about readeability. From a specification point of view, I think everything is here.

Thanks @Amxx, let's do another review round!

certora/specs/Pausable.spec
Amxx: The issue is that sometimes the operation can fail for "conditions" that are hard/impossible to formalize. For example, in a TimelockController, the execution will revert if the timelocked call (which is an arbitrary call) reverts. In that case its very hard to prove lifeness, and we are limited to proving the first point.

certora/specs/methods/IPausable.spec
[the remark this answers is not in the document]
Amxx: IMO the real question here is weither or not the interface is going to be reused. 

I'm not sure ERC5313 needs an interface here. On the other hand, having one for TimelockController makes sens if governor also needs it.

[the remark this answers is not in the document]
ernestognw: Yes, I'm not suggesting adding the `IERC5313.spec`, but wanted to confirm what we expect before moving forward.

So far:

- If the interface is going to be reused, we create an `I*.spec`
- If it's not going to be reused, we put it at the top of its `*.spec` file

certora/specs/Pausable.spec
ernestognw: Thanks for confirming, so far I've seen `success <=> ...` for liveness across other examples in this repository, but wasn't clear to me if that was a general rule.

I'll use it moving forward with the DoubleEndedQueue.

> the execution will revert if the timelocked call (which is an arbitrary call) reverts

Isn't this most of the reasons is difficult to fully formally verify the library? Now I see why generalization gets hard in this context.

frangio (approved):
LGTM!

PR #4118, opened 2023-03-14
Cleanup timelockId on execution for gas refund

Amxx:
When calling execute, the state is checked. We expect to be in `Queued`. That means the `state` call done by `execute` is going to read `_timelockIds[proposalId]` and check the operation status. This sload makes the slot "hot".

As a consequence, cleaning it will cost `2900` and refund `4800`, resulting in a theoretical `1900` gas savings. Because of the hash necessary to compute storage slot, this saving is reduced to `1825`, which is still "nice to have".

- This does not change the behavior of the `state(uint256)` function 
  - `execute()` sets the `executed` flag, which means the state will be returns without even looking at `_timelockIds[proposalId]`
- This does not change the behavior of the  `proposalEta(uint256)` function
  - before this change, calling `proposalEta` on an executed proposal would lookup the deadline on the timelock. That returns `_DONE_TIMESTAMP` (1) which is rewritten to 0 by the governor
  - after this change, calling `proposalEta` on an executed proposal would lookup the deadline of `0x0` on the timelock, which should be 0.
  

#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx:
Capture d’écran du 2023-03-14 11-41-50

Amxx:
With the reordering
Capture d’écran du 2023-03-14 12-13-27

frangio:
This looks good, but there is a conflict that needs to be resolved.

contracts/governance/extensions/GovernorTimelockControl.sol
```diff
@@ -64,7 +64,11 @@ abstract contract GovernorTimelockControl is IGovernorTimelock, Governor {
             return ProposalState.Executed;
         } else if (_timelock.isOperationPending(queueid)) {
             return ProposalState.Queued;
+        } else if (_timelock.isOperationDone(queueid)) {
+            // This can happen if the proposal is executed directly on the timelock.
+            return ProposalState.Executed;
```
frangio, line 69: The merge wasn't done properly. `isOperationDone` is already called 2 lines above. Which one should we keep?

contracts/governance/extensions/GovernorTimelockControl.sol
```diff
@@ -117,12 +121,15 @@ abstract contract GovernorTimelockControl is IGovernorTimelock, Governor {
...
     function _execute(
-        uint256 /* proposalId */,
+        uint256 proposalId,
         address[] memory targets,
         uint256[] memory values,
         bytes[] memory calldatas,
         bytes32 descriptionHash
     ) internal virtual override {
+        // cleanup for refund
+        delete _timelockIds[proposalId];
```
frangio, line 131: I would feel more comfortable deleting the id after the execute call for cases of reentrancy.

contracts/governance/extensions/GovernorTimelockControl.sol
Amxx: `isOperationPending` and `isOperationDone` are exclusive, so the order is irrelevant.

There are "public" calls that have a Queued state in their path (execute) but none that have an Executed state in their path. So I would optimize for queued

contracts/governance/extensions/GovernorTimelockControl.sol
Amxx: I was thinking the opposite: doing all the local changes before having an external call and a possible reentrancy

contracts/governance/extensions/GovernorTimelockControl.sol
Amxx: Fixed

contracts/governance/extensions/GovernorTimelockControl.sol
frangio: I see the reasoning... But in this case the local change is not "mark it as done", it is "erase the way to know that it's already scheduled".

Amxx: Its more than that ... by cleaning that we are changing the way `state(uint256)` behaves. 

I'm not saying I'm doing it write, but we need to consider what we want the state to be is a reentrancy call happens. At first glance, it feels more correct to return Queued than Succeeded... so we should probably clean after.

frangio: I still would feel more comfortable deleting after the call.

contracts/governance/extensions/GovernorTimelockControl.sol
```diff
@@ -117,12 +119,15 @@ abstract contract GovernorTimelockControl is IGovernorTimelock, Governor {
...
+        // execute
         _timelock.executeBatch{value: msg.value}(targets, values, calldatas, 0, descriptionHash);
```
Amxx, line 131:
```suggestion
        // execute
        _timelock.executeBatch{value: msg.value}(targets, values, calldatas, 0, descriptionHash);
        // cleanup for refund
        delete _timelockIds[proposalId];
```

contracts/governance/extensions/GovernorTimelockControl.sol
```diff
@@ -140,10 +145,13 @@ abstract contract GovernorTimelockControl is IGovernorTimelock, Governor {
...
         uint256 proposalId = super._cancel(targets, values, calldatas, descriptionHash);
+        bytes32 timelockId = _timelockIds[proposalId];
 
-        if (_timelockIds[proposalId] != 0) {
-            _timelock.cancel(_timelockIds[proposalId]);
+        if (timelockId != 0) {
+            // cleanup
             delete _timelockIds[proposalId];
+            // cancel
+            _timelock.cancel(timelockId);
```
Amxx, line 154:
```suggestion
            // cancel
            _timelock.cancel(timelockId);
            // cleanup
            delete _timelockIds[proposalId];
```

frangio (approved):

PR #4121, opened 2023-03-16
Add AccessManager contracts

frangio:
Implements a pattern of access control where permissions for a multi-contract system are managed in a central location (an AccessManager instance) and with function granularity.


contracts/access/manager/AccessManaged.sol
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @dev This contract module makes available a {restricted} modifier. Functions decorated with this modifier will be
+ * permissioned according to an "authority": a contract like {AccessManager} that follows the {IAuthority} interface,
+ * implementing a policy that allows certain callers access to certain functions.
+ */
+contract AccessManaged {
+    event AuthorityUpdated(IAuthority indexed oldAuthority, IAuthority indexed newAuthority);
+
+    IAuthority private _authority;
```
Amxx, line 15: I know that is restrictive, but making this immutable would save a non insignificant amount of gas.
Should we have two versions ?

In the case of proxy/clones created by a factory, would it make sens that all implementation use the same authority? I think yes.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,317 @@
...
+ *
+ * Permissioning schemes and team and contract group assignments can be configured by the default admin. The contract
+ * includes {AccessControlDefaultAdminRules} by default to enforce security rules on this account. Additionally, it is
+ * expected that the account will be highly secured (e.g., a multisig or a well-configured DAO) as all the permissions
+ * of the managed system can be modified by it.
+ */
+contract AccessManager is IAccessManager, AccessControlDefaultAdminRules {
+    bytes32 _createdTeams;
+    mapping(address => bytes32) private _userTeams;
+    mapping(bytes32 => mapping(bytes4 => bytes32)) private _allowedTeams;
```
Amxx, line 76: IMO we should use 0.8.18 "named parameters in mapping types" here. That would improve the understanding that one bytes32 is a group and the other is a set of teams

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,317 @@
...
+    mapping(address => bytes32) private _contractGroup;
+
+    uint8 private constant _TEAM_ALL = 255;
+    bytes32 private constant _GROUP_UNSET = 0;
+    bytes32 private constant _GROUP_OPEN = "group:open";
+    bytes32 private constant _GROUP_CLOSED = "group:closed";
```
Amxx, line 82: I would have gone for hashes like we do everywhere else. Sure that syntax is nice because you can understand the value without doing a preimage, but that might brings confusion as to why we use it here and not for roles such as PROPOSER_ROLE (in the timelock).

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,317 @@
...
+        emit TeamUpdated(team, name);
+    }
+
+    /**
+     * @dev Updates an existing team's name. The caller must be the default admin.
+     */
+    function updateTeamName(uint8 team, string calldata name) public virtual onlyDefaultAdmin {
+        require(team != _TEAM_ALL && _teamExists(team));
+        emit TeamUpdated(team, name);
+    }
```
Amxx, line 123: What was the rational for needing to explicitly create teams?

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,317 @@
...
+        }
+    }
+
+    /**
+     * @dev Returns the {AccessControl} role id that corresponds to a team.
+     *
+     * This role id starts with the ASCII characters `team:`, followed by zeroes, and ends with the single byte
+     * corresponding to the team number.
+     */
+    function _encodeTeamRole(uint8 team) internal virtual returns (bytes32) {
```
Amxx, line 270:
```suggestion
    function _encodeTeamRole(uint8 team) internal virtual view returns (bytes32) {
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,317 @@
...
+
+    /**
+     * @dev Decodes a role id into a team, if it is a role id of the kind returned by {_encodeTeamRole}.
+     */
+    function _decodeTeamRole(bytes32 role) internal virtual returns (bool, uint8) {
+        bytes32 tagMask = hex"ffffffff";
+        bytes32 teamMask = hex"ff";
+        bool isTeam = (role & tagMask) == bytes32("team:");
+        uint8 team = uint8(uint256(role & teamMask));
+        return (isTeam, team);
```
Amxx, line 282:
```suggestion
        bytes32 tag = (role >> 8) << 8;
        uint8 team  = uint8(role[31]);
        return (tag == bytes32("team:"), team);
```

contracts/access/manager/AccessManaged.sol
frangio: > Should we have two versions ?

You know I hate multiple versions... :smile:

The main reason why I left this as a mutable variable is because I think it can be used to freeze the permissions in a contract by moving it off of an AccessManager onto an immutable Authority. We could also implement freezing in the AccessManager itself, by having frozen contract groups in which permissions can't be altered, and whose contracts can't be moved out of that group. I felt that having mutability of the authority at the managed contract was the simpler option, but it's true that it makes it more expensive.

contracts/access/manager/AccessManager.sol
frangio: In retrospect I think it was a mistake to use hashes and I think this is a clearly superior option. I think you're understating the benefit of not needing the preimage!

frangio:
Should this contract support batching? If so, should it be "native" or simply through Multicall?

Note that the default admin should almost always be a multisig or DAO which tend to have batching built in anyway.

contracts/access/manager/AccessManager.sol
frangio: This gives the assurance that a team is "clean" and has no pre-existing members. I'm not entirely convinced that this is important. I was thinking of the situation where you're auditing a governance proposal or multisig proposal where there is a new team being defined that is granted permissions to a set of functions. You want to make sure that it is actually a new team that doesn't have pre-existing members and seeing the `createTeam` function call ensures that that's true (as long as the batch is atomic).

Do you agree that this is a benefit of having explicit creation? Do you see good reasons to remove explicit creation?

Something that could be argued is that contract groups have a similar problem with respect to being "clean", and those have no explicit creation, although they could have it too, but I think that would be more cumbersome operationally because many contracts will have ad hoc groups.

contracts/access/manager/AccessManager.sol
Amxx: that is not correct though.

- `(role >> 8) << 8` removes the rightmost byte
- `role & ~hex"ff"` removes the leftmost byte

Amxx:
> Should this contract support batching? If so, should it be ad hoc or simply through Multicall?
> 
> Note that the default admin should almost always be a multisig or DAO which tend to have batching built in anyway.

It would go with multicall. Simpler, cleaner

contracts/access/manager/AccessManager.sol
frangio: You're right, but I meant `role & ~0xff`.

frangio:
@ggonzalez94 suggested simply making the `selector` argument an array of selectors in `setFunctionAllowedTeam`, since that is likely to be the main thing people do in batch. I think that's simple and it makes sense.

I'm not a fan of adding `Multicall` because usability and auditability are not great. For example, a Multicall-batched transaction in Gnosis Safe becomes unauditable without extra steps to manually decode calldata.

contracts/access/manager/AccessManager.sol
Amxx: `~bytes32(uint256(0xff))` might work.

Amxx: the fact that even we get confused by it says a lot about how messy this is

Amxx:
> For example, a Multicall-batched transaction in Gnosis Safe becomes unauditable without extra steps to manually decode calldata.

That is a UI issue that gnosis should fix!

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+        return (tag == bytes32("team:"), team);
+    }
+
+    /**
+     * @dev Returns the isolate group id for a target contract.
+     *
+     * The group id consists of the ASCII characters `group:isolate:` followed by the contract address bytes.
+     */
+    function _encodeIsolateGroup(address target) internal virtual pure returns (bytes32) {
+        return _GROUP_ISOLATE_PREFIX | (bytes20(target) >> _GROUP_ISOLATE_PREFIX.length);
```
ernestognw, line 336: I think you meant:

```suggestion
        return _GROUP_ISOLATE_PREFIX | (bytes32(bytes20(target)) >> _GROUP_ISOLATE_PREFIX.length * 8);
```

In which case, there are 2 bytes left out of the target address.

I wouldn't be worried about this since it's removing 2 bytes from the address collision security (20 bytes), reducing it to 2**18, which is still high enough. However, there might be some vanity addresses out there with multiple zeroes at the beginning.

According to a vanity address generator on an M1 Macbook Pro 16', getting a 7-zeros prefix is 5 months, reducing the security to 20 - 2 - 7 = 11 bytes. That starts to feel dangerously close of being non-negligible.

Moreover, some chains use predeploys or precompiles with hardcoded addresses that might imply a security concern if we trim the last two bytes. Consider the [Optimism predeploys](https://github.com/ethereum-optimism/optimism/blob/develop/specs/predeploys.md), where almost everyone starts with `0x42...` and the significant bits are the last two. 

Would be fixed if we replace the prefix with something shorter like `group:single` or `group:islte`.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    /**
+     * @dev Returns the isolate group id for a target contract.
+     *
+     * The group id consists of the ASCII characters `group:custom:` followed by the group name.
+     */
+    function _encodeCustomGroup(string calldata name) internal virtual pure returns (bytes32) {
```
ernestognw, line 344: Can we restrict `name` length? I think it's not clear that the last part gets trimmed, and dependent frontends (or even other contracts) may make mistakes

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+        require(bytes(name).length + _GROUP_CUSTOM_PREFIX.length < 32);
+        return _GROUP_CUSTOM_PREFIX | (bytes32(bytes(name)) >> _GROUP_CUSTOM_PREFIX.length);
```
ernestognw, line 346: Same comment about the length as with `_GROUP_ISOLATE_PREFIX`

```suggestion
        return _GROUP_CUSTOM_PREFIX | (bytes32(bytes(name)) >> _GROUP_CUSTOM_PREFIX.length * 8);
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+     * @dev Returns true if the group is one of "open", "closed", or unset (zero).
+     */
+    function _isSpecialGroup(bytes32 group) private pure returns (bool) {
+        return group == _GROUP_OPEN || group == _GROUP_CLOSED;
+    }
+
+    /**
+     * @dev Returns a bit mask where the only non-zero bit is the team number bit.
+     */
+    function _teamMask(uint8 team) private pure returns (bytes32) {
```
ernestognw, line 359: 🧠

ernestognw (commented):
So far I really like the design, leaving a bunch of comments but I know it's still a drat, so feel free to disregard.

contracts/access/README.adoc
```diff
@@ -25,3 +25,13 @@ This directory provides ways to restrict who can access the functions of a contr
 {{AccessControlEnumerable}}
 
 {{AccessControlDefaultAdminRules}}
+
+== AccessManager
```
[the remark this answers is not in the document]
frangio: Yeah we can add a small intro. Usually we put intro at the top level but there should be something here as well.

contracts/access/manager/AccessManaged.sol
frangio: > now a managed contract may have the right to update itself without the Manager's approval.

I don't think this is true. Only the current manager should be able to "eject" a managed contract from itself.

contracts/access/manager/AccessManaged.sol
```diff
@@ -0,0 +1,36 @@
...
+    constructor(IAuthority initialAuthority) {
+        _authority = initialAuthority;
+    }
+
+    function authority() public view virtual returns (IAuthority) {
+        return _authority;
+    }
+
+    function setAuthority(IAuthority newAuthority) public virtual {
+        require(msg.sender == address(_authority));
```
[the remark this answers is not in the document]
frangio: Seems risky. I don't think this ability would/should ever be granted to a team.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+        _setTeam(_createdTeams, team, true);
+        emit TeamUpdated(team, name);
+    }
+
+    /**
+     * @dev Updates an existing team's name. The caller must be the default admin.
+     */
+    function updateTeamName(uint8 team, string calldata name) public virtual onlyDefaultAdmin {
+        require(team != _TEAM_PUBLIC && _teamExists(team));
+        emit TeamUpdated(team, name);
```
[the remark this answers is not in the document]
frangio: Yes, changed this after your initial feedback.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+     * human-readable name. The caller must be the default admin.
+     *
+     * Team numbers are not auto-incremented in order to avoid race conditions, but administrators can safely use
+     * sequential numbers.
+     *
+     * Emits {TeamUpdated}.
+     */
+    function createTeam(uint8 team, string memory name) public virtual onlyDefaultAdmin {
+        require(!_teamExists(team));
+        _setTeam(_createdTeams, team, true);
```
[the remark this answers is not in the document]
frangio: Yes, good catch, I noticed this afterwards myself too and renamed the function to `_withUpdatedTeam`. Very error prone otherwise.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+    // function transferContractAuthority(address target, address newAuthority) external;
+}
+
+/**
+ * @dev AccessManager is a central contract that stores the permissions of a system. It is an AccessControl contract,
+ * i.e. it has roles and all the standard functions like `grantRole` and `revokeRole`, but it defines a particular set
+ * of roles, with a particular structure.
+ *
+ * Users are grouped in "teams". Teams must be created before users can be assigned into them, up to a maximum of 255
+ * teams. A user can be assigned to multiple teams. Each team defines an AccessControl role, identified by a role id
```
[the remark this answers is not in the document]
frangio: What do you mean by "team 0 is the default"? There is no "default team" in any sense as far as I can tell.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+
+    // TODO: consider sanitizing name
+    /**
+     * @dev Creates a new team with a team number that can be chosen arbitrarily but must be unused, and gives it a
+     * human-readable name. The caller must be the default admin.
+     *
+     * Team numbers are not auto-incremented in order to avoid race conditions, but administrators can safely use
+     * sequential numbers.
+     *
+     * Emits {TeamUpdated}.
```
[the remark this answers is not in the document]
frangio: Yeah I want to look into that after we freeze the Solidity code.

contracts/access/manager/AccessManager.sol
frangio: The improvement that I see is that you are able to get a value (role id, group id) and more easily interpret it. With hashes, it's very difficult to map a value to its intended "meaning", you need a rainbow table to get an actual label that explains that meaning in a human friendly way.

That said, this benefit is not fully realized because standard tooling will automatically display a `bytes32` value as a hex string. But I still think it's worth doing without hashes for applications that integrate with AccessManager. For example, if we add a new special group in the future called `group:paused` or something, those applications would be automatically able to show `group:paused` in the UI without needing an update to the rainbow table.

contracts/access/manager/AccessManager.sol
frangio: Oh interesting find. I definitely wouldn't want to erase any of the address bits so we should choose a shorter prefix.

contracts/access/manager/AccessManager.sol
frangio: The length is restricted in `_encodeCustomGroup`.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+ * Contracts in the system may be grouped as well. These are simply called "contract groups". There can be an arbitrary
+ * number of groups. Each contract can only be in one group at a time. In the simplest setup, each contract is assigned
+ * to its own separate group, but groups can also be shared among similar contracts.
+ *
+ * All contracts in a group share the same permissioning scheme. A permissioning scheme consists of a mapping between
+ * functions and allowed teams. Each function can be allowed to multiple teams, meaning that if a user is in at least
+ * one of the allowed teams they can call that function.
+ *
+ * Note that a function in a target contract may become permissioned only when: 1) said contract is {AccessManaged} and
+ * is connected to this contract as its manager, and 2) said function is decorated with the `restricted` modifier.
```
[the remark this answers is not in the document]
frangio: I prefer the current style personally.

frangio:
I've renamed "teams" to "badges". As I was writing tests I realized that often a team might just contain one account and the term wasn't very fitting. I think "badges" are a very fitting concept for the domain.

frangio:
Unfortunately the new mapping syntax is not compatible with solidity-coverage. :confused: 

- https://github.com/sc-forks/solidity-coverage/issues/782

Amxx:
> Yes, 100%, but I think a selectors array even helps with that use case too.

A selector array will help if you want to batch multiple instances of one function call. What if you want to include another operation in the batch ? Multicall allows you to batch arbitrary function calls. IMO its strictly more powerfull.

Amxx:
> I think "badges" are a very fitting concept for the domain.

I honestly preferred team. 

It makes sens to me that I can be part of a team, and that it can be a one-person-team.
Having a badge on the other hand is confusing. Maybe I just need to get used to it ... but I'd say its less intuitive.

frangio:
Ok, I've removed contract groups which we don't see a very strong immediate use case for, and renamed "badge" to "group".

hardhat.config.js
```diff
@@ -40,7 +40,7 @@ const argv = require('yargs/yargs')()
     compiler: {
       alias: 'compileVersion',
       type: 'string',
-      default: '0.8.13',
+      default: '0.8.18',
```
ernestognw, line 43: We may want to return this back to `0.8.13` and discuss in another PR, I would personally prefer to check it more carefully.

ernestognw (commented):
Pending to review tests

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    /**
+     * @dev Sets the target contract to be in "closed" mode. All restricted functions in the target contract will be
+     * closed down and disallowed to all. The caller must be the default admin.
+     */
+    function setContractModeClosed(address target) public virtual onlyDefaultAdmin {
+        _contractMode[target] = RestrictedMode.Closed;
+        emit RestrictedModeUpdated(target, RestrictedMode.Closed);
+    }
```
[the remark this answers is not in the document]
frangio: Agree.

ernestognw:
> > Edit: also gnosis safe support batches, so if they want great transaction auditability, they should just use that.
> 
> Yes, 100%, but I think a selectors array even helps with that use case too.

I'd be in favor of this, but as an addition after merging this. I don't think we can increase the scope right now

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,339 @@
...
+
+    /**
+     * @dev Augmented version of {AccessControl-_grantRole} that keeps track of user group bitmaps.
+     */
+    function _grantRole(bytes32 role, address user) internal virtual override {
+        super._grantRole(role, user);
+        (bool isGroup, uint8 group) = _decodeGroupRole(role);
+        if (isGroup) {
+            require(hasGroup(group), "AccessManager: unknown group");
```
ernestognw, line 268: I think this `require` should be a noop instead. The probability of `isGroup` being hit by accident is `2 ** (8 * 6)`, which I consider non-negligible.

Specially because some people use the `abi.encode(bytes("..."))`, which may have `group:` as prefix in some setups. 

Look at [this one](https://github.com/OpenZeppelin/accesscontrol-explorer/blob/master/src/lib/role-dict.json#L5270) that starts with `gr` for `grantor`.

ernestognw: Nevermind, I realized removing this restriction will make easy to impersonate a group by crafting a role that starts with the 6 bytes of `group:`. Think about cases like the following may not be easy to spot:

```
bytes32 SOME_ROLE = abi.encodePacked(bytes("group:..."));
```

And we've seen such cases https://github.com/OpenZeppelin/accesscontrol-explorer/blob/master/src/lib/role-dict.json#L5270

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,339 @@
...
+     * The caller must be the default admin.
+     */
+    function setFunctionAllowedGroup(
+        address target,
+        bytes4[] calldata selectors,
+        uint8 group,
+        bool allowed
+    ) public virtual onlyDefaultAdmin {
+        require(_contractMode[target] == RestrictedMode.Custom, "AccessManager: target in special mode");
+        for (uint256 i = 0; i < selectors.length; i++) {
```
ernestognw, line 205: We may want to restrict selectors length by some `MAX_SELECTORS`, similar to `_maxBatchSize`in `ERC721Consecutive` if either:

1. It may cause issues to emit multiple `GroupAllowed` at once.
2. This function can be DoS

I would discard number 2, but at least we should add a not for other contracts relying on this function.

test/access/manager/AccessManager.test.js
```diff
@@ -0,0 +1,523 @@
...
+
+    beforeEach('deploying managed contract', async function () {
+      await this.manager.createGroup(group, '', { from: admin });
+      await this.manager.grantGroup(groupMember, group, { from: admin });
+      this.managed = await AccessManaged.new(this.manager.address);
+    });
+
+    it('non-admin cannot change allowed groups', async function () {
+      await expectRevert(
+        this.manager.methods['setFunctionAllowedGroup(address,bytes4[],uint8,bool)'](
```
ernestognw, line 214: Can we put all `setFunctionAllowedGroup(address,bytes4[],uint8,bool)` in a variable?

ernestognw (commented):
Will re-review tomorrow, so far looks good to me, but there are some pending comments.

Amxx (commented):
I'm mostly good with the behavior of the contract, with some minor concerns.

However I still have 2 significant concerns:

- The packaging of the manager in a "preset" format. IMO that goes against our the policies. I'm starting to consider that we should have two repo. One with the unopinionated libraries & utils, and another (that depends on the first one) with ready to use instances that can be deployed out of the box (vesting/timelock/manager/...)

- This contract is all about clear and precise role management, yet a single admin is in charge of everything. In particular, in a context of progressive decentralisation, it is difficult for an admin to renounce part (an only part) of its role. I would like more time to experiment on a An AccessManager that is is managing itself.

Final comment is that I'm really not comfortable with use working on deadline. I know why its the case today, but I also feels like this is the negation of the slow pace that we used to follow, and that greatly contributed to the quality of our code. We took it slow for weeks, and now we have to rush the review? This looks like French politics... which is not a good sign.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,339 @@
...
+     * corresponding to the group number.
+     */
+    function _encodeGroupRole(uint8 group) internal pure virtual returns (bytes32) {
+        return bytes32("group:") | bytes32(uint256(group));
+    }
+
+    /**
+     * @dev Decodes a role id into a group, if it is a role id of the kind returned by {_encodeGroupRole}.
+     */
+    function _decodeGroupRole(bytes32 role) internal pure virtual returns (bool isGroup, uint8 group) {
```
Amxx, line 307:
```suggestion
    function _decodeGroupRole(bytes32 role) internal pure virtual returns (bool, uint8) {
```

hardhat.config.js
Amxx, line 43: This unfortunately break the gas estimation workflow. I'll look into that.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,339 @@
...
+
+    uint8 private constant _GROUP_PUBLIC = type(uint8).max;
+
+    /**
+     * @dev Initializes an AccessManager with initial default admin and transfer delay.
+     */
+    constructor(
+        uint48 initialDefaultAdminDelay,
+        address initialDefaultAdmin
+    ) AccessControlDefaultAdminRules(initialDefaultAdminDelay, initialDefaultAdmin) {
```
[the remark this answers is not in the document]
frangio: > This doesn't feel like OZ contract library, it fells like something else.

I agree that there is a small difference in spirit but I think it goes in the right direction. This whole contract is "opinionated"! It's a recommendation on how to manage permissions.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,339 @@
...
+        _setContractMode(target, RestrictedMode.Open);
+    }
+
+    /**
+     * @dev Sets the target contract to be in "closed" mode. All restricted functions in the target contract will be
+     * closed down and disallowed to all. The caller must be the default admin.
+     */
+    function setContractModeClosed(address target) public virtual onlyDefaultAdmin {
+        _setContractMode(target, RestrictedMode.Closed);
+    }
```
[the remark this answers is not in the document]
frangio: The thinking is that a single function with an enum argument is more error-prone, and the consequences are critical. Do you disagree?

contracts/access/manager/AccessManaged.sol
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @dev This contract module makes available a {restricted} modifier. Functions decorated with this modifier will be
+ * permissioned according to an "authority": a contract like {AccessManager} that follows the {IAuthority} interface,
+ * implementing a policy that allows certain callers access to certain functions.
+ *
+ * IMPORTANT: The `restricted` modifier should never be used on `internal` functions, judiciously used in `public`
+ * functions, and ideally only used in `external` functions. See {restricted}.
+ */
+contract AccessManaged {
+    event AuthorityUpdated(IAuthority indexed oldAuthority, IAuthority indexed newAuthority);
```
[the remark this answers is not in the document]
Amxx: In solmate the first argument is the user, not the `oldAuthority`. So if we keep this event, we have two identical events, that are indistinguishable without looking at the code, that have different semantics. 

Remember that there is an internal `_setAuthority` that can called by someone that is not the oldAuthority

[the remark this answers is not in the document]
Amxx: I'd go for exaclty matching Solmate's semantics

contracts/access/AccessControl.sol
```diff
@@ -72,6 +72,11 @@ abstract contract AccessControl is Context, IAccessControl, ERC165 {
         _;
     }
 
+    modifier onlyDefaultAdmin() {
+        _checkRole(DEFAULT_ADMIN_ROLE);
+        _;
+    }
```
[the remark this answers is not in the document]
frangio: Just improving syntax. Would you prefer we remove it?

test/access/manager/AccessManager.test.js
```diff
@@ -0,0 +1,523 @@
...
+const AccessManaged = artifacts.require('$AccessManagedMock');
+
+const Ownable = artifacts.require('$Ownable');
+const AccessControl = artifacts.require('$AccessControl');
+
+const groupUtils = {
+  mask: group => 1n << BigInt(group),
+  decodeBitmap: hexBitmap => {
+    const m = BigInt(hexBitmap);
+    const allGroups = new Array(256).fill().map((_, i) => i.toString());
```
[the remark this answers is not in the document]
frangio: The reason is that we need to use a string here:

```js
        await expectEvent.inConstruction(this.manager, 'GroupUpdated', {
          group: PUBLIC_GROUP, // <- must be string, otherwise expectEvent fails
          name: 'public',
        });
```

test/access/manager/AccessManager.test.js
```diff
@@ -0,0 +1,523 @@
...
+    return allGroups.filter(i => (m & groupUtils.mask(i)) !== 0n);
```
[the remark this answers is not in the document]
frangio: I don't see how this suggestion makes sense. We need `map` first because filtering from `new Array(256).fill()` will leave a shorter array without any indication of what indices were removed.

contracts/access/manager/AccessManaged.sol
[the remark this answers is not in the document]
Amxx: > Is it just renaming `oldAuthority` to user and making sure we use `msg.sender`?

yes

test/access/manager/AccessManager.test.js
```diff
@@ -0,0 +1,523 @@
...
+      await this.manager.grantGroup(user1, group, { from: admin });
+      this.adapter = await AccessManagerAdapter.new(this.manager.address);
+    });
+
+    it('with ownable', async function () {
+      const target = await Ownable.new();
+      await target.transferOwnership(this.adapter.address);
+
+      const { data } = await target.$_checkOwner.request();
+      const selector = data.slice(0, 10);
```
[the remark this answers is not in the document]
Amxx: But you in `AccessManaged.test.js` you use `const selector = web3.eth.abi.encodeFunctionSignature('restrictedFunction()');`

contracts/access/manager/AccessManaged.sol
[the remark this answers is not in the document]
frangio: Ok to use `sender` instead of `user`? We have precedent for `sender` in `AccessControl.RoleGranted` and so on.

test/access/manager/AccessManager.test.js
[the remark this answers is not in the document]
frangio: The thing is that since `_checkRole` has an argument we need the full encoded calldata at least for AccessControl.

What do you suggest?

contracts/access/manager/AccessManaged.sol
frangio: IMO it's fine if the target contract deliberately inserts a custom-gated `_setAuthority` function call. I wouldn't expect it normally, but I can imagine some sort of emergency mechanism.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,339 @@
...
+ * `restricted` modifier.
+ *
+ * There is a special group defined by default named "public" which all accounts automatically have.
+ *
+ * Contracts can also be configured in two special modes: 1) the "open" mode, where all functions are allowed to the
+ * "public" group, and 2) the "closed" mode, where no function is allowed to any group.
+ *
+ * Since all the permissions of the managed system can be modified by the admins of this instance, it is expected that
+ * it will be highly secured (e.g., a multisig or a well-configured DAO). Additionally, {AccessControlDefaultAdminRules}
+ * is included to enforce security rules on this account.
```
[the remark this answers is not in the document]
frangio: Agree on reviewing this but I think we need do it after this PR with more time.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,339 @@
...
+        return bytes32("group:") | bytes32(uint256(group));
+    }
+
+    /**
+     * @dev Decodes a role id into a group, if it is a role id of the kind returned by {_encodeGroupRole}.
+     */
+    function _decodeGroupRole(bytes32 role) internal pure virtual returns (bool isGroup, uint8 group) {
+        bytes32 tagMask = ~bytes32(uint256(0xff));
+        bytes32 tag = role & tagMask;
+        uint8 group = uint8(role[31]);
```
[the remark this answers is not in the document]
ernestognw: It was replaced. Solving

contracts/access/manager/AccessManaged.sol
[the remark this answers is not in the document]
ernestognw: > Ok to use sender instead of user?

If `sender == msg.sender` for every case. Yes.

ernestognw (approved):
LGTM after addressing the `virtual` suggestion I added for `canCall`.

These are the assumptions I'm taking (before the contract is released) for approving:

- Batching is something we shall discuss more in-depth to make improvements
- It's pending to add better documentation or at least an introduction/guide to the system, following [this discussion](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4121#discussion_r1141496472)
- We need to research more about making `AccessManager` also `AccessManaged`.
- The idea of contract groups is still an open question
- We'll review and discuss if `AccessControlDefaultAdminRules` should be default instead of just `AccessControl`
- We'll also check alternatives for `setFunctionAllowedGroup` being `onlyRole(DEFAULT_ADMIN_ROLE)` and potential risk mitigations

contracts/access/README.adoc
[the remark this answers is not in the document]
ernestognw: Referenced in [my LGTM](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4121#pullrequestreview-1353256465).

Solving

contracts/access/manager/AccessManaged.sol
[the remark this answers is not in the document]
ernestognw: Also referenced in my last review. Solving.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,381 @@
...
+
+    function revokeTeam(address user, uint8 team) external;
+
+    function renounceTeam(address user, uint8 team) external;
+
+    function getFunctionAllowedTeams(address target, bytes4 selector) external view returns (bytes32 teams);
+    function getFunctionAllowedTeams(string calldata group, bytes4 selector) external view returns (bytes32 teams);
+
+    function setFunctionAllowedTeam(address target, bytes4[] calldata selectors, uint8 team, bool allowed) external;
+    function setFunctionAllowedTeam(string calldata group, bytes4[] calldata selectors, uint8 team, bool allowed) external;
```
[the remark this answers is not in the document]
ernestognw: The concept of `teams` was removed. Solving

test/access/manager/AccessManager.test.js
```diff
@@ -0,0 +1,506 @@
...
+  describe('adapter', function () {
+    const group = '0';
+
+    beforeEach('deploying adapter', async function () {
+      await this.manager.createGroup(group, 'a group', { from: admin });
+      await this.manager.grantGroup(group, user1, { from: admin });
+      this.adapter = await AccessManagerAdapter.new(this.manager.address);
+    });
+
+    it('with ownable', async function () {
```
ernestognw, line 468: There's no test case for the manager's `_DEFAULT_ADMIN_ROLE ` relay case

frangio:
Merging into feature branch.

PR #4123, opened 2023-03-17
Bump forge std to v1.5.2

PaulRBerg:
This PR bumps `forge-std` to [v1.5.2](https://github.com/foundry-rs/forge-std/releases/tag/v1.5.2).


frangio:
Hi @PaulRBerg, as I mentioned previously we need to merge things into `master` and from there we can cherry pick them onto the release branch.

What is in this release? Unfortunately I don't think it's realistic for you to assume that this repo will always be on the latest version of `forge-std`.

frangio:
Ok, I'll close then!

PR #4125, opened 2023-03-20
Updated owner.sol

gaurav-singh7092:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx:
Hello @gaurav-singh7092 

What do you need this function for? `owner()` is already public. It feels like this function doesn't add much, increase the deployment cost, and can be confusing because of the dependency on msg.sender.

I'm personally not a fan of context sensitive view functions.

PR #206, opened 2017-04-24
Create and test PausableToken Contract

DavidKnott:
Creates PausableToken contract that inherits from the StandardTokenContract.
Change Pauable contract to use ```pause``` as opposed to ```emergency```
PausableToken has an owner who is able to:

Pause all token transfers
Unpause all token transfers

This contract was created in response to issue #194 
It's fully tested in PausableToken.js and all tests are passing.

PR #4132, opened 2023-03-22
Add custom linting rules

### LINKED ISSUE #3919: Create custom linting rules for conventions
There are a few things that we should be checking for automatically. We should create custom linter rules for them. We are currently using Solhint as a linter and it has the ability to define custom rules. Solhint seems to be mostly unmaintained and if it continues like that we might want to switch to some other linter, but unfortunately there seems to be no alternative at the moment. (Although one alternative could be Slither, which we're already using as well.) The effort of writing custom rules for Solhint is probably small enough that it's still worth doing even if we switch to some other linter soon.

Here we can collect the different rules that we should have:

- Check all state variables are private.
- Check there are no "external virtual" functions (they can't be overridden and call `super`).
- Check all internal and private variables and functions are prefixed with underscore.

lhemerly:
Fixes #3919 

Adds a custom plugin for Solhint that implements the project's Solidity guidelines.


frangio:
Nice @lhemerly! This is appreciated.

Is it possible to have the "plugin" as a JS file in this repository, rather than a separate npm package? I would like to do that as it will be easier for us to maintain and evolve.

lhemerly:
As far as I know it is not possible right now as you can't set-up the plugin path in solhint:
[https://github.com/protofire/solhint/issues/206]

It is hardcoded to look up for `solhint-plugin-${pluginName}` in the node_modules folder:
[https://github.com/protofire/solhint/blob/6b3c7ea2d5b36636a2d5799e280d577518249324/lib/rules/index.js#L73]

I may try a PR on Solhint for custom plugins paths later

lhemerly:
Just a follow up, PR in solhint created:
[Solhint PR for Issue #206](https://github.com/protofire/solhint/pull/424)

frangio:
Maybe we could make it work by defining the package locally in the repo? Not exactly but sort of monorepo style.

lhemerly:
The way I can imagine making it work as solhint is right now is to take the "solhint-plugin-ozcontracts" folder inside node_modules out of the gitignore and work with the .js file there as part of the repo. The way npm (or yarn) install works it won't change that folder as long as we remove and keep it out of package.json.

If you think it is ok to go with that I can make the necessary changes

frangio:
Putting in `node_modules` is not ideal. We should be able to put it in a subdirectory somewhere in the repo and add `"solhint-plugin-custom": "file:./solhint-plugin"` in the dependencies.

frangio:
@lhemerly I've vendored your plugin into this repository like I suggested.

This should be a great start to begin to build more custom linting rules. Thanks!

frangio:
@Amxx @ernestognw Note that this will allow us to have internal constants without leading underscore moving forward.

scripts/solhint-custom/index.js
```diff
@@ -0,0 +1,82 @@
...
+      }
+      if (node.visibility === 'internal' && node.parent.kind === 'library') {
+        if (/^_/.test(node.name)) {
+          this.error(node, 'Library internal functions should not be prefixed with underscore');
+        }
+      }
+    }
+  },
+
+  // TODO: re-enable and fix
```
frangio, line 72: We have some code that causes this rule to fail so I'm proposing to merge and enable it later.

solhint.config.js
```diff
@@ -0,0 +1,19 @@
...
+  'const-name-snakecase',
+  'contract-name-camelcase',
+  'event-name-camelcase',
+  'func-name-mixedcase',
+  'func-param-name-mixedcase',
+  'modifier-name-mixedcase',
+  'var-name-mixedcase',
+  'imports-on-top',
+  ...customRules.map(r => `openzeppelin/${r}`),
+];
```
Amxx:
```suggestion
const customRules = require('./scripts/solhint-custom');

const rules = [
  'no-unused-vars',
  'const-name-snakecase',
  'contract-name-camelcase',
  'event-name-camelcase',
  'func-name-mixedcase',
  'func-param-name-mixedcase',
  'modifier-name-mixedcase',
  'var-name-mixedcase',
  'imports-on-top',
  ...customRules.map(r => `openzeppelin/${r.ruleId}`),
];
```

solhint.config.js
```diff
@@ -0,0 +1,20 @@
...
+  'contract-name-camelcase',
+  'event-name-camelcase',
+  'func-name-mixedcase',
+  'func-param-name-mixedcase',
+  'modifier-name-mixedcase',
+  'var-name-mixedcase',
+  'imports-on-top',
+  'no-global-import',
+  ...customRules.map(r => `openzeppelin/${r}`),
+];
```
Amxx, line 15:
```suggestion
const customRules = require('./scripts/solhint-custom');

const rules = [
  'no-unused-vars',
  'const-name-snakecase',
  'contract-name-camelcase',
  'event-name-camelcase',
  'func-name-mixedcase',
  'func-param-name-mixedcase',
  'modifier-name-mixedcase',
  'var-name-mixedcase',
  'imports-on-top',
  'no-global-import',
  ...customRules.map(r => `openzeppelin/${r.ruleId}`),
];
```

scripts/solhint-custom/index.js
```diff
@@ -0,0 +1,82 @@
...
+    static ruleId = 'underscore-prefix';
+
+    VariableDeclaration(node) {
+      const constantOrImmutable = node.isDeclaredConst || node.isImmutable;
+      if (node.isStateVar && !constantOrImmutable && node.visibility === 'private') {
+        if (!/^_/.test(node.name)) {
+          this.error(node, 'Private variables must be prefixed with underscore');
+        }
+      }
+    }
```
Amxx, line 56: Should we also have some rule that if the stateVar is constantOrImmutable, then it must NOT start with an underscore ? or do we accept both ?

scripts/solhint-custom/index.js
```diff
@@ -0,0 +1,82 @@
...
+  class extends Base {
+    static ruleId = 'underscore-prefix';
+
+    VariableDeclaration(node) {
+      const constantOrImmutable = node.isDeclaredConst || node.isImmutable;
+      if (node.isStateVar && !constantOrImmutable && node.visibility === 'private') {
+        if (!/^_/.test(node.name)) {
+          this.error(node, 'Private variables must be prefixed with underscore');
+        }
+      }
```
Amxx, line 55:
```suggestion
      }
      if (node.isStateVar && constantOrImmutable) {
        if (/^_/.test(node.name)) {
          this.error(node, 'Constant and immutable variables should not be prefixed with underscore');
        }
      }
```

lhemerly:
Just an update, development branch in solhint adjusted plugins to be in any path:
https://github.com/protofire/solhint/pull/424#event-9745473144

scripts/solhint-custom/index.js
frangio: I agree, but this causes the rule to fail in a few places.

I added the rule but commented it out, and I'd propose to do another PR next week fixing those things. Does this sound ok?

Amxx: So we will have 2 follow up PRs that :
- enable the rule 
- fix the code
?

That sounds ok

frangio: Yeah or just one PR doing the two.

Amxx: Maybe the rule should also check that the name is capitalized.

frangio: That should be already handled by the casing rules in `solhint.config.js`, specifically `const-name-snakecase`.

solhint.config.js
```diff
@@ -0,0 +1,20 @@
...
+  'modifier-name-mixedcase',
+  'var-name-mixedcase',
+  'imports-on-top',
+  'no-global-import',
+  ...customRules.map(r => `openzeppelin/${r.ruleId}`),
+];
+
+module.exports = {
+  plugins: ['openzeppelin'],
+  rules: Object.fromEntries(rules.map(r => [r, 'error'])),
```
ernestognw, line 19: Why not define it like an object?

```
const rules = {
  'no-unused-vars': 'error',
  ...
}
```

Not a blocker because everything shows an "error", I just wanted to point out that the array doesn't seem to be needed

ernestognw (approved):

solhint.config.js
frangio: We need to use `fromEntries` in some way because we get the names of the custom rules dynamically:

```
customRules.map(r => `openzeppelin/${r.ruleId}`)
```

The rest could be an object, but I had a feeling that this was clearer? Let me know if it's too confusing.

PR #4133, opened 2023-03-23
ReentrancyGuard gas optimization

simon-perriard:
By setting `_NOT_ENTERED` to `1` and `_ENTERED` to `0`, and modifying the `_status` check to be`==`, this saves 51 gas.

#### PR Checklist

- [X] Tests - no new test needed
- [X] Documentation - no documentation update needed

### DIFF
--- a/contracts/security/ReentrancyGuard.sol
+++ b/contracts/security/ReentrancyGuard.sol
@@ -55,7 +55,7 @@ abstract contract ReentrancyGuard {
 
     function _nonReentrantBefore() private {
         // On the first call to nonReentrant, _status will be _NOT_ENTERED
-        require(_status > _ENTERED, "ReentrancyGuard: reentrant call");
+        require(_status == _NOT_ENTERED, "ReentrancyGuard: reentrant call");
 
         // Any calls to nonReentrant after this point will fail
         _status = _ENTERED;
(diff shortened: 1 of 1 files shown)


contracts/security/ReentrancyGuard.sol
```diff
@@ -55,7 +55,7 @@ abstract contract ReentrancyGuard {
 
     function _nonReentrantBefore() private {
         // On the first call to nonReentrant, _status will be _NOT_ENTERED
-        require(_status != _ENTERED, "ReentrancyGuard: reentrant call");
+        require(_status == _NOT_ENTERED, "ReentrancyGuard: reentrant call");
```
Amxx, line 58: This causes an issue if the module is added through an upgrade. If for any reason the constructor/initialized is not set, `_status` will be 0, and the system will be blocked

Amxx:
Hello @simon-perriard 

Doing `1 → 2 → 1` and `1 → 0 → 1` should cost the same

- In the first case:
  - `1 → 2`: 5000 (cold slot)
  - `2 → 1`: 2200 + 4900 refund
  - Total: 7200 + 4900 refund

- In the second case:
  - `1 → 0`: 5000 (cold slot) + 4800 refund
  - `0 → 1`: 2200 + 100 refund
  - Total: 7200 + 4900 refund
  
  So AFAIK, the saving only comes from changing a `==` in a `!=`. 
  Our checks shows it saves 3 gas, which is exactly the cost of the added `NOT` opcode.
  
  As discussed in the other comment, using `!=` is on purpose, and the cost of 3 gas sounds reasonable. Where does the `51` gas come from ?

PR #4134, opened 2023-03-24
Improve `ERC4626` test coverage

### LINKED ISSUE #4075: Improve `ERC4626` test coverage
After my extensive `ERC4626` review, I would like to quickly discuss whether to add the following missing unit tests. As a reference, I refer to the existing tests at [`ERC4626.test.js`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/token/ERC20/extensions/ERC4626.test.js). The coverage issue is specifically related to the `private` function `_tryGetAssetDecimals`:

```solidity
/**
 */
function _tryGetAssetDecimals(IERC20 asset_) private view returns (bool, uint8) {
    (bool success, bytes memory encodedDecimals) = address(asset_).staticcall(
        abi.encodeWithSelector(IERC20Metadata.decimals.selector)
    );
    if (success && encodedDecimals.length >= 32) {
        uint256 returnedDecimals = abi.decode(encodedDecimals, (uint256));
        if (returnedDecimals <= type(uint8).max) {
            return (true, uint8(returnedDecimals));
        }
    }
    return (false, 0);
}
```

- Add a unit test that calls a simple EOA as `asset_` at construction time (checks for `encodedDecimals.length >= 32`).
- Add a unit test that calls a contract as `asset_` at construction time that returns a higher `decimals` than `type(uint8).max` (checks the condition `returnedDecimals <= type(uint8).max`). For that you can use a simple mock like that:
```solidity
pragma solidity ^0.8.0;

contract ERC20DecimalsExcessMock {
    function decimals() public pure returns (uint256) {
        return type(uint256).max;
    }
}
```

- Add a unit test that checks the overflow of `_underlyingDecimals + _decimalsOffset()`. This might be a non-issue right now, but if you add such a check and do an `unchecked` refactor at some point it should immediately throw :)

Let me know what you think.

pcaversaccio:
Fixes #4075. I added loops for the tests to iterate over multiple values.

- [x] Tests
- [ ] Documentation

### DIFF
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,0 +1,1 @@
+openzeppelin/=contracts/
--- a/test/token/ERC20/extensions/ERC4626.t.sol
+++ b/test/token/ERC20/extensions/ERC4626.t.sol
@@ -19,10 +19,10 @@ contract ERC4626VaultOffsetMock is ERC4626OffsetMock {
 }
 
 contract ERC4626StdTest is ERC4626Test {
-    ERC20 private underlying = new ERC20Mock();
+    ERC20 private _underlying = new ERC20Mock();
 
     function setUp() public override {
-        _underlying_ = address(underlying);
+        _underlying_ = address(_underlying);
         _vault_ = address(new ERC4626Mock(_underlying_));
         _delta_ = 0;
         _vaultMayBeEmpty = true;
@@ -33,9 +33,9 @@ contract ERC4626StdTest is ERC4626Test {
      * @dev Check the case where calculated `decimals` value overflows the `uint8` type.
      */
     function testFuzzDecimalsOverflow(uint8 offset) public {
-        /// @dev Remember that the `underlying` exhibits a `decimals` value of 18.
+        /// @dev Remember that the `_underlying` exhibits a `decimals` value of 18.
         offset = uint8(bound(uint256(offset), 238, uint256(type(uint8).max)));
-        ERC4626VaultOffsetMock erc4626VaultOffsetMock = new ERC4626VaultOffsetMock(underlying, offset);
+        ERC4626VaultOffsetMock erc4626VaultOffsetMock = new ERC4626VaultOffsetMock(_underlying, offset);
         vm.expectRevert();
         erc4626VaultOffsetMock.decimals();
     }
--- a/test/token/ERC20/extensions/ERC4626.t.sol
+++ b/test/token/ERC20/extensions/ERC4626.t.sol
@@ -1,18 +1,42 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import "erc4626-tests/ERC4626.test.sol";
+import {ERC4626Test} from "erc4626-tests/ERC4626.test.sol";
 
-import {SafeCast} from "../../../../contracts/utils/math/SafeCast.sol";
-import {ERC20Mock} from "../../../../contracts/mocks/ERC20Mock.sol";
-import {ERC4626Mock} from "../../../../contracts/mocks/ERC4626Mock.sol";
+import {SafeCast} from "openzeppelin/utils/math/SafeCast.sol";
+import {ERC20} from "openzeppelin/token/ERC20/ERC20.sol";
+import {ERC4626} from "openzeppelin/token/ERC20/extensions/ERC4626.sol";
+
+import {ERC20Mock} from "openzeppelin/mocks/ERC20Mock.sol";
+import {ERC4626Mock} from "openzeppelin/mocks/ERC4626Mock.sol";
+import {ERC4626OffsetMock} from "openzeppelin/mocks/token/ERC4626OffsetMock.sol";
+
+contract ERC4626VaultOffsetMock is ERC4626OffsetMock {
+    constructor(
+        ERC20 underlying_,
+        uint8 offset_
+    ) ERC20("My Token Vault", "MTKNV") ERC4626(underlying_) ERC4626OffsetMock(offset_) {}
+}
 
 contract ERC4626StdTest is ERC4626Test {
+    ERC20 private underlying = new ERC20Mock();
+
     function setUp() public override {
-        _underlying_ = address(new ERC20Mock());
+        _underlying_ = address(underlying);
         _vault_ = address(new ERC4626Mock(_underlying_));
         _delta_ = 0;
         _vaultMayBeEmpty = true;
         _unlimitedAmount = true;
+    }
+
+    /**
+     * @dev Check the case where calculated `decimals` value overflows the `uint8` type.
+     */
+    function testFuzzDecimalsOverflow(uint8 offset) public {
+        /// @dev Remember that the `underlying` exhibits a `decimals` value of 18.
+        offset = uint8(bound(uint256(offset), 238, uint256(type(uint8).max)));
+        ERC4626VaultOffsetMock erc4626VaultOffsetMock = new ERC4626VaultOffsetMock(underlying, offset);
+        vm.expectRevert();
+        erc4626VaultOffsetMock.decimals();
     }
 }
--- a/test/token/ERC20/extensions/ERC4626.test.js
+++ b/test/token/ERC20/extensions/ERC4626.test.js
@@ -5,6 +5,7 @@ const ERC20Decimals = artifacts.require('$ERC20DecimalsMock');
 const ERC4626 = artifacts.require('$ERC4626');
 const ERC4626OffsetMock = artifacts.require('$ERC4626OffsetMock');
 const ERC4626FeesMock = artifacts.require('$ERC4626FeesMock');
+const ERC20ExcessDecimalsMock = artifacts.require('ERC20ExcessDecimalsMock');
 
 contract('ERC4626', function (accounts) {
   const [holder, recipient, spender, other, user1, user2] = accounts;
@@ -18,6 +19,30 @@ contract('ERC4626', function (accounts) {
       const token = await ERC20Decimals.new('', '', decimals);
       const vault = await ERC4626.new('', '', token.address);
       expect(await vault.decimals()).to.be.bignumber.equal(decimals);
+    }
+  });
+
+  it('asset has not yet been created', async function () {
+    for (let i = 0; i < 6; i++) {
+      const vault = await ERC4626.new('', '', accounts[i]);
+      expect(await vault.decimals()).to.be.bignumber.equal(decimals);
+    }
+  });
+
+  it('underlying excess decimals', async function () {
+    const token = await ERC20ExcessDecimalsMock.new();
+    const vault = await ERC4626.new('', '', token.address);
+    expect(await vault.decimals()).to.be.bignumber.equal(decimals);
+  });
+
+  it('decimals overflow', async function () {
+    for (const offset of [243, 250, 255].map(web3.utils.toBN)) {
+      const token = await ERC20Decimals.new('', '', decimals);
+      const vault = await ERC4626OffsetMock.new(name + ' Vault', symbol + 'V', token.address, offset);
+      await expectRevert(
+        vault.decimals(),
+        'reverted with panic code 0x11 (Arithmetic operation underflowed or overflowed outside of an unchecked block)',
+      );
     }
   });
 
--- a/test/token/ERC20/extensions/ERC4626.test.js
+++ b/test/token/ERC20/extensions/ERC4626.test.js
@@ -23,10 +23,8 @@ contract('ERC4626', function (accounts) {
   });
 
   it('asset has not yet been created', async function () {
-    for (let i = 0; i < 6; i++) {
-      const vault = await ERC4626.new('', '', accounts[i]);
-      expect(await vault.decimals()).to.be.bignumber.equal(decimals);
-    }
+    const vault = await ERC4626.new('', '', other);
+    expect(await vault.decimals()).to.be.bignumber.equal(decimals);
   });
 
   it('underlying excess decimals', async function () {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- `ERC4626`: Improved test coverage by adding further unit tests covering the internal function `_tryGetAssetDecimals` and an overflow test for `decimals`. ([#]())
+
 ### Breaking changes
 
 - `EIP712`: Addition of ERC5267 support requires support for user defined value types, which was released in Solidity version 0.8.8. This requires a pragma change from `^0.8.0` to `^0.8.8`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- `ERC4626`: Improved test coverage by adding further unit tests covering the internal function `_tryGetAssetDecimals` and an overflow test for `decimals`. ([#]())
+- `ERC4626`: Improved test coverage by adding further unit tests covering the internal function `_tryGetAssetDecimals` and an overflow test for `decimals`. ([#4134](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4134))
 
 ### Breaking changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # Changelog
 
-## Unreleased
-
-- `ERC4626`: Improved test coverage by adding further unit tests covering the internal function `_tryGetAssetDecimals` and an overflow test for `decimals`. ([#4134](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4134))
-
 ### Breaking changes
 
 - `EIP712`: Addition of ERC5267 support requires support for user defined value types, which was released in Solidity version 0.8.8. This requires a pragma change from `^0.8.0` to `^0.8.8`.
--- a/contracts/mocks/token/ERC20ExcessDecimalsMock.sol
+++ b/contracts/mocks/token/ERC20ExcessDecimalsMock.sol
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+contract ERC20ExcessDecimalsMock {
+    function decimals() public pure returns (uint256) {
+        return type(uint256).max;
+    }
+}


pcaversaccio:
@ernestognw thanks for the info - makes sense! Will remove accordingly my `CHANGELOG` entry in this case. Btw, the Slither CI seems broken...

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -21,6 +22,30 @@ contract('ERC4626', function (accounts) {
     }
   });
 
+  it('asset has not yet been created', async function () {
+    for (let i = 0; i < 6; i++) {
+      const vault = await ERC4626.new('', '', accounts[i]);
```
Amxx, line 27: why do we need to check for multiple EOA. Isn't one enough ?

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -21,6 +22,30 @@ contract('ERC4626', function (accounts) {
...
+    const vault = await ERC4626.new('', '', token.address);
+    expect(await vault.decimals()).to.be.bignumber.equal(decimals);
+  });
+
+  it('decimals overflow', async function () {
+    for (const offset of [243, 250, 255].map(web3.utils.toBN)) {
+      const token = await ERC20Decimals.new('', '', decimals);
+      const vault = await ERC4626OffsetMock.new(name + ' Vault', symbol + 'V', token.address, offset);
+      await expectRevert(
+        vault.decimals(),
```
Amxx, line 43: In that case, we should probably revert in the constructor.

test/token/ERC20/extensions/ERC4626.test.js
pcaversaccio: since `accounts` is available I thought we can do a simple "EOA" fuzzing test, but generally, yes one should be enough. Lmk your preferences.

test/token/ERC20/extensions/ERC4626.test.js
pcaversaccio: Is it really worth the additional deployment costs?

test/token/ERC20/extensions/ERC4626.test.js
ernestognw: Since we're already using the [ERC4626 test suite](https://github.com/a16z/erc4626-tests/blob/8b1d7c2ac248c33c3506b1bff8321758943c5e11/ERC4626.test.sol) with Foundry, we can add that fuzzing case for the accounts.

pcaversaccio: I had to force myself this morning to write `JS` tests lol - I actually wanted to first just simply open a PR with the `forge` tests but I thought you will reject that in any case :). Happy to add some fuzz tests if you guys agree on this path.

ernestognw: I think [this reference ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7b3e7b7055d6e1226d4920ca7d7c446c0c0bfff5/test/token/ERC20/extensions/ERC4626.t.sol) might help you. It'd be a matter of extending `ERC4626StdTest`

> Happy to add some fuzz tests if you guys agree on this path.

We'd appreciate those tests if they make sense. At least this one does for me.

pcaversaccio: Thx, am actually familiar with it since I used it for my Vyper `ERC4626` tests [here](https://github.com/pcaversaccio/snekmate/blob/main/test/extensions/ERC4626.t.sol) as well (yes I use Solidity tests to test Vyper :)).

pcaversaccio: Actually writing my own tests has led to opening an issue here (I was really paranoid about everything tbh).

pcaversaccio: @Amxx I refactored it into using only one EOA which seems fine.

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -21,6 +22,30 @@ contract('ERC4626', function (accounts) {
...
+  });
+
+  it('underlying excess decimals', async function () {
+    const token = await ERC20ExcessDecimalsMock.new();
+    const vault = await ERC4626.new('', '', token.address);
+    expect(await vault.decimals()).to.be.bignumber.equal(decimals);
+  });
+
+  it('decimals overflow', async function () {
+    for (const offset of [243, 250, 255].map(web3.utils.toBN)) {
```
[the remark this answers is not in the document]
pcaversaccio: I added a fuzz test for this case - the above one I refactored using only one EOA.

pcaversaccio:
I slightly adjusted the existing Foundry tests as well as added a remapping file that simplifies the internal imports imho.

test/token/ERC20/extensions/ERC4626.test.js
ernestognw: I'd argue that if it fails because of a `decimals()` overflow, it's behaving as expected (that's why we're adding the test). I stand with @pcaversaccio and don't think it should go on the constructor

test/token/ERC20/extensions/ERC4626.t.sol
```diff
@@ -1,18 +1,42 @@
...
+        uint8 offset_
+    ) ERC20("My Token Vault", "MTKNV") ERC4626(underlying_) ERC4626OffsetMock(offset_) {}
+}
 
 contract ERC4626StdTest is ERC4626Test {
+    ERC20 private _underlying = new ERC20Mock();
+
     function setUp() public override {
-        _underlying_ = address(new ERC20Mock());
+        _underlying_ = address(_underlying);
```
ernestognw, line 25: I think this has to be `new ERC20Mock`. Otherwise, it makes further tests dependant on each other.
Not a problem right now.

ernestognw (approved):
This LGTM, but someone else's approval is needed.
Thanks @pcaversaccio!

remappings.txt
```diff
@@ -0,0 +1 @@
+openzeppelin/=contracts/
```
ernestognw, line 1: Oh, we also have to discuss if we want `openzeppelin`, `@openzeppelin`, or something else. Right now I'm fine with this remapping.

test/token/ERC20/extensions/ERC4626.test.js
Amxx: Also, we are talking about a valut where is offset is so big that `10**offset` doesn't fix in a uint256. This means `_convertToShares` and `_convertToAssets` will always revert. That means this contract is completely unusable.

I'm thinking there should be an upper bound on the offset. Anything beyond 64 is really dangerous. I personally don't think it should ever be more than 32

pcaversaccio: Well, the functions we're talking about here (`decimals`, `_convertToShares`, and `_convertToAssets`) are all `virtual`. So let's say we add a check in the constructor. Someone inherits from the `ERC4626` abstract contract and implements a new `_convertToShares` and `_convertToAssets` function (for whatever reason) without a `super` call to the parent function. In these new functions anything can be possible, e.g. `1**offset` or `100**offset`, which could lead to a non-overflowing or overflowing behaviour even with having a check. If we're talking about that high of an offset, I think the problem itself is a PEBCAK problem.

Amxx: Lets not get confused here:

The fact that devs might override in way that are unsafe is beyond our control, and the scope of what we can check.

What I believe we should be concerned about is the fact that we currently have a contract that, without modification from the dev, can be completely broken if the constructor parameters are not within a certain range.

Right now, all it takes is a deployer to not fully understand the way the decimal offset work, use a big value, and end up with a bricked contract. I don't think this is OK. I believe its better to revert during the deployment than to end up with a contract where every meaningful operation will revert.

This is already something we do in the `VestingWallet` contract. If the `beneficiaryAddress` is 0, we revert during construction.

Amxx: > What I'm concerned of is the corresponding revert string to get written on the bytecode if we set an upper bound. Do you think it's worth it?

You are more concerned with a revert string in th deployment bytecode (it shouldn't even make it into the actual bytecode of the contract) than with broken contracts being deployed ?

pcaversaccio: >  Right now, all it takes is a deployer to not fully understand the way the decimal offset work, use a big value, and end up with a bricked contract.

First of all, I do fully agree that this is not ok - that's why I added this test in the first line. But, just for me to understand this argument: someone who is able to understand what a `virtual` function is and who is able to override the default `_decimalsOffset` in a way that breaks the contract does not understand how the offset works? What if we document this properly in the function `_decimalsOffset`, because this is what people will read first in case they override it?

> What I believe we should be concerned about is the fact that we currently have a contract that, without any modification from the dev, can be completely broken if the constructor parameters are not within a certain range.

I don't think this is true. The testing mock is configured in a customisable way, yes, but the `ERC4626` itself is an abstract contract that does not have an `offset` value set in the constructor (the max value for `decimals` in the default setting is already constrained to `type(uint8).max`). It's set by default to 0. Thus, what I want to emphasise here, is that the `ERC4626` (abstract) contract is not vulnerable by default to the overflow. What it takes is to customise the offset itself. Please correct me if I'm wrong here.

Amxx: > but the `ERC4626` itself is an abstract contract that does not have an `offset` value set in the constructor

Yes indeed. My mistake. I confused the mock constructor with the contract constructor.

Amxx: I was confused by the fact that at some point `offset` was an immutable variable set in ERC4626 constructor. All good !

test/token/ERC20/extensions/ERC4626.t.sol
ernestognw: I see, so your approach is favoring execution time, right? (Deploy a mock ERC20 once and not on each fuzz run)

My worry comes from the following scenario:

1. We add a test for the `mint` function that requires approval of the underlying asset
2. Later, if we need the `mint` for another test, we might be adding more approval than what's needed (eg. an old test left some approval)

Might not be a big issue, but I'm not clear if the tradeoff between this and test performance is worth it.

remappings.txt
ernestognw: We haven't added a remappings file before, so we can have this discussion later outside of this PR. For now all good, thanks @pcaversaccio

test/token/ERC20/extensions/ERC4626.t.sol
Amxx: We have a limited understanding of how foundry testing works. If there are multiple `testXxxx` functions in a `XxxTest` contract, which of the following is true:

- A new (fresh) `XxxTest` is deployed for running each test
- A single `XxxTest` is deployed, but snapshot are used to ensure the state is "clean" at the begining of each `testXxx` function
- A single `XxxTest` contract is deployed, and some function (setUp?) is used to clean the contract between tests?

It looks like its not option 3.

Amxx: With option 1 or 2, what happens is:

- the `_underlyin = new ERC20Mock()` is executed as part of the implicit constructor.
- the `setUp()` function is then called, and initializes the `ERC4646Test` variable.
- we are ready to test

---

It feel like we could even do
```solidity
    ERC4626 private _vault = newERC4626Mock(new ERC20Mock());
    
    function setUp() public override {
        _underlying_ = _vault.asset();
        _vault_ = address(_vault);
        _delta_ = 0;
        _vaultMayBeEmpty = true;
        _unlimitedAmount = true;
    }
```
Would that work just as fine ?

ernestognw: They look the same to me

Captura de pantalla 2023-03-30 a la(s) 12 50 36

These are the sources:

```solidity
contract CounterTest is Test {
    Counter public counter;

    function setUp() public {
        counter = new Counter();
        counter.setNumber(0);
    }

    function testSetNumber(uint256 x) public {
        counter.setNumber(x);
        assertEq(counter.number(), x);
    }
}
```

```solidity
pragma solidity ^0.8.13;

import "forge-std/Test.sol";
import "../src/Counter.sol";

contract CounterTest is Test {
    Counter public counter = new Counter();

    function setUp() public {
        counter.setNumber(0);
    }

    function testSetNumber(uint256 x) public {
        counter.setNumber(x);
        assertEq(counter.number(), x);
    }
}
```

test/token/ERC20/extensions/ERC4626.t.sol
```diff
@@ -1,18 +1,42 @@
...
+    /**
+     * @dev Check the case where calculated `decimals` value overflows the `uint8` type.
+     */
+    function testFuzzDecimalsOverflow(uint8 offset) public {
+        /// @dev Remember that the `_underlying` exhibits a `decimals` value of 18.
+        offset = uint8(bound(uint256(offset), 238, uint256(type(uint8).max)));
+        ERC4626VaultOffsetMock erc4626VaultOffsetMock = new ERC4626VaultOffsetMock(_underlying, offset);
+        vm.expectRevert();
+        erc4626VaultOffsetMock.decimals();
+    }
```
Amxx, line 41: I'm not really seeing the value of this test. IMO fuzzing is great to:

- check correctness of functions (for math)
- check liveness (whatever we throw at it, function does not revert)

Here you are showing that the function revert for offset between 238 and 255 ... but unit tests already shows that. What else do we want to prove ?

test/token/ERC20/extensions/ERC4626.t.sol
ernestognw: They look the same to me

Captura de pantalla 2023-03-30 a la(s) 12 50 36

These are the sources:

```solidity
contract CounterTest is Test {
    Counter public counter;

    function setUp() public {
        counter = new Counter();
        counter.setNumber(0);
    }

    function testSetNumber(uint256 x) public {
        counter.setNumber(x);
        assertEq(counter.number(), x);
    }
}
```

```solidity
pragma solidity ^0.8.13;

import "forge-std/Test.sol";
import "../src/Counter.sol";

contract CounterTest is Test {
    Counter public counter = new Counter();

    function setUp() public {
        counter.setNumber(0);
    }

    function testSetNumber(uint256 x) public {
        counter.setNumber(x);
        assertEq(counter.number(), x);
    }
}
```

Amxx (commented):
1. I like the new unit tests, in particular the one about `ERC20ExcessDecimalsMock`
2. I'm not sure the fuzzing tests are bringing anything that the unit test don't already cover.
3. We won't be able to merge into master until 4.9 is at least in RC. Maybe we sould set a `4.10` branch for PR like this one.

test/token/ERC20/extensions/ERC4626.t.sol
pcaversaccio: Well, `setUp` is an optional function invoked before **each** test case (i.e. a function prepended with `test`) is run. This is not something specific to Foundry but has its roots in hevm. You can see this [here](https://github.com/dapphub/dapptools/blob/master/src/hevm/src/EVM/UnitTest.hs#L125). So what happens in our case is the following:
1. The state of `ERC20 private _underlying = new ERC20Mock();` and any modification thereof via the _implicit_ constructor (i.e. part of the contract-level state variable definitions) remain the same across test functions.
2. Any modification of `_underlying` or any other variable within the `setUp` call is invoked before each test function. This means if you for instance deploy within the `setUp` function an `Ownable` contract, for which the `msg.sender` equals the `owner` by design, and you have a unit test that transfers the ownership multiple times, the state is reset if you call another test function, i.e. the `msg.sender` will be again the `owner` in the new test function at the beginning. All unit tests are one transaction and start with the post-`setUp` call state. Btw, that's one of the reasons why you can't test properly `selfdestruct` since the code is deleted at the end of a transaction. 

The reason why I didn't choose to use ` ERC4626 private _vault = new ERC4626Mock(new ERC20Mock());` directly, is that maybe we want to re-use `new ERC20Mock()` at some point.

test/token/ERC20/extensions/ERC4626.t.sol
pcaversaccio: Well, fuzzing is random input generation and for the above test case, the domain space is very small indeed but also means we can cover the full range (which is not the case for e.g. a domain space of `uint16`). Personally, I like adding such fuzzing tests to ensure that the contracts are somehow future-proof (i.e. if I make a small change, it still checks for all possible values) and can be easily extended if needed. I might be too paranoid here, so if you think in the current state it doesn't make sense, please remove it. But in the near future, such a test might come handy since it can be easily extended.

pcaversaccio:
> 1. I like the new unit tests, in particular the one about `ERC20ExcessDecimalsMock`
> 2. I'm not sure the fuzzing tests are bringing anything that the unit test don't already cover.
> 3. We won't be able to merge into master until 4.9 is at least in RC. Maybe we sould set a `4.10` branch for PR like this one.

Re 2., see me comment [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4134#discussion_r1153829404).

Re 3., I think this test must belong to the `4.9` release, since you introduce there the decimals offset to prevent the inflation attack.

remappings.txt
n012judge420: **__**

test/token/ERC20/extensions/ERC4626.t.sol
n012judge420: Good

test/token/ERC20/extensions/ERC4626.t.sol
Amxx: `new ERC20Mock()` would be accessible via `_vault.asset()`

Amxx: But the burning question is:

If we do what you propose AND if one of the tests mints some underlying tokens, will the following tests also have the token minted? We want all test to start with a "blank" state, and that includes having a brand new underlying token with no balances (in particular, no balance for the vault)

pcaversaccio: > new ERC20Mock() would be accessible via _vault.asset()

But this returns an `address` type directly and not the contract object itself. Maybe we want to do something like that once:

```solidity
_underlying.mint(holder, type(uint16).max);
_underlying.approve(spender, type(uint256).max);
```

> But the burning question is:
> 
> If we do what you propose AND if one of the tests mints some underlying tokens, will the following tests also have the token minted? We want all test to start with a "blank" state, and that includes having a brand new underlying token with no balances (in particular, no balance for the vault)

The short answer is yes, it will have again the clean state (i.e. the post-`setUp` call state). I will make an example taken from my own unit tests [here](https://github.com/pcaversaccio/snekmate/blob/main/test/tokens/ERC721.t.sol) to illustrate this.

Take for instance [this](https://github.com/pcaversaccio/snekmate/blob/main/test/tokens/ERC721.t.sol#LL595-L604C6) unit test:

```solidity
function testBalanceOfCase1() public {
    address owner = makeAddr("owner");
    string memory uri1 = "my_awesome_nft_uri_1";
    string memory uri2 = "my_awesome_nft_uri_2";
    vm.startPrank(deployer);
    ERC721Extended.safe_mint(owner, uri1);
    ERC721Extended.safe_mint(owner, uri2);
    assertEq(ERC721Extended.balanceOf(owner), 2);
    vm.stopPrank();
}
```

Now further down, take the [following](https://github.com/pcaversaccio/snekmate/blob/main/test/tokens/ERC721.t.sol#LL1422-L1444C24) unit test which does another couple of NFT mints for instance.

```solidity
function testSafeMintSuccess() public {
        address owner = makeAddr("owner");
        string memory uri1 = "my_awesome_nft_uri_1";
        string memory uri2 = "my_awesome_nft_uri_2";
        string memory uri3 = "my_awesome_nft_uri_3";
        uint256 tokenId = 0;
        vm.startPrank(deployer);
        vm.expectEmit(true, true, true, false);
        emit Transfer(zeroAddress, owner, tokenId);
        vm.expectEmit(false, false, false, true);
        emit MetadataUpdate(tokenId);
        ERC721Extended.safe_mint(owner, uri1);
        vm.expectEmit(true, true, true, false);
        emit Transfer(zeroAddress, owner, tokenId + 1);
        vm.expectEmit(false, false, false, true);
        emit MetadataUpdate(tokenId + 1);
        ERC721Extended.safe_mint(owner, uri2);
        vm.expectEmit(true, true, true, false);
        emit Transfer(zeroAddress, owner, tokenId + 2);
        vm.expectEmit(false, false, false, true);
        emit MetadataUpdate(tokenId + 2);
        ERC721Extended.safe_mint(owner, uri3);
        vm.stopPrank();

        vm.startPrank(owner);
        vm.expectEmit(true, true, true, false);
        emit Transfer(owner, zeroAddress, tokenId + 2);
        ERC721Extended.burn(tokenId + 2);
        vm.stopPrank();

        vm.startPrank(deployer);
        vm.expectEmit(true, true, true, false);
        emit Transfer(zeroAddress, owner, tokenId + 3);
        vm.expectEmit(false, false, false, true);
        emit MetadataUpdate(tokenId + 3);
        ERC721Extended.safe_mint(owner, "");
        vm.stopPrank();
        assertEq(ERC721Extended.balanceOf(owner), 3);
        assertEq(ERC721Extended.totalSupply(), 3);
        assertEq(ERC721Extended.ownerOf(tokenId), owner);
        assertEq(ERC721Extended.ownerOf(tokenId + 1), owner);
        assertEq(ERC721Extended.ownerOf(tokenId + 3), owner);
    }
```

The above test would fail if it had the post `testBalanceOfCase1` state (since the NFTs would already exist). What I usually do is a `testInitalSetup` unit test that checks the post-`setUp` call. For instance, for my `ERC4626` (Vyper) implementation see [here](https://github.com/pcaversaccio/snekmate/blob/main/test/extensions/ERC4626.t.sol#LL182-L311C6). That's also where I e.g. test for the overflow problem. Does this clarify your question?

pcaversaccio:
> code that is being audited

I hope they return feedback that the test coverage is missing there ;-), otherwise am disappointed by the auditors lol

test/token/ERC20/extensions/ERC4626.t.sol
Amxx: That is also what I observed. Its just that this part is essential and is often overlooked. This is what @ernestognw meant when he said "makes further tests dependant on each other".

I guess snapshot are involved here.

Amxx:
> > code that is being audited
> 
> I hope they return feedback that the test coverage is missing there ;-), otherwise am disappointed by the auditors lol

I think they priority goes toward finding issues in code that is potentially covered rather than for correct code that is not covered.

pcaversaccio:
> > > code that is being audited
> > 
> > I hope they return feedback that the test coverage is missing there ;-), otherwise am disappointed by the auditors lol
> 
> I think they priority goes toward finding issues in code that is potentially covered rather than for correct code that is not covered.

Of course, I was joking 🙃

test/token/ERC20/extensions/ERC4626.t.sol
pcaversaccio: Fyi, if you want to handle snapshots directly yourself, you can use the `snapshot` cheatcode as well: https://book.getfoundry.sh/cheatcodes/snapshots. Do you feel overall uncomfortable with this change?

frangio (approved):
Thanks!

PR #4138, opened 2023-03-25
Moved 'to != owner' from ERC721.approval to ERC721._approval

### LINKED ISSUE #4136: Move `to != owner` check from `ERC721.approval` to `ERC721._approval`
In `ERC721.approve` (public) we do
```solidity
require(to != owner, "ERC721: approval to current owner");
```
we don't do it in `ERC721._approve` (internal)

That means its not ok to approve the owner, unless you do it internally ? IMO the check should move from the public to the internal function. (note that there is already a call to `ERC721.ownerOf` inside the internal version, so its not like we are saving an `sload` anyway)

Saksham010:
Fixes #4136  

'to!=owner' check has been moved from ERC721.approve to ERC721._approve internal function

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -365,8 +365,8 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
      * Emits an {Approval} event.
      */
     function _approve(address to, uint256 tokenId) internal virtual {
-         address owner = ERC721.ownerOf(tokenId);
-         require(to != owner, "ERC721: approval to current owner");
+        address owner = ERC721.ownerOf(tokenId);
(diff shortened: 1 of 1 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -364,6 +363,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
      * Emits an {Approval} event.
      */
     function _approve(address to, uint256 tokenId) internal virtual {
+         require(to != owner, "ERC721: approval to current owner");
         _tokenApprovals[tokenId] = to;
         emit Approval(ERC721.ownerOf(tokenId), to, tokenId);
```
Amxx, line 368: Please make sure your code compile and passes the tests.

PR #4139, opened 2023-03-25
Custom Errors For ERC20 and unit tests

### LINKED ISSUE #2839: Solidity 0.8.4 Custom Errors
Solidity 0.8.4 introduced "custom errors", which are named errors that can be thrown similar to how events are emitted.

These bring many benefits: primarily the gas savings of not needing to embed revert strings in the bytecode, as well as the ability to include data in the error.

More details:  
https://blog.soliditylang.org/2021/04/21/custom-errors/

I would love to see these errors added to OZ contracts.

lhemerly:
Fixes #2839

Add feature custom error to ERC20

Testing is raising errors for multicall, ERC777 and ERC 4626 because I have not added the feature to them yet, and they are using ERC20 behavior. Wanted to make sure the design decisions I made for the feature and testing are OK so I can move forward to the other contracts

- [ X ] Tests 
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -1,11 +1,41 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v4.8.0) (token/ERC20/ERC20.sol)
 
-pragma solidity ^0.8.0;
+pragma solidity ^0.8.4;
 
 import "./IERC20.sol";
 import "./extensions/IERC20Metadata.sol";
 import "../../utils/Context.sol";
+
+/**
+ * @dev Implementation of solidity 0.8.4 custom errors.
+ *
+ * This implementation adopted the convention of errors before and outside the contract
+ * code.
+ *
+ * The following are EIP-6093 implementations:
+ * https://eips.ethereum.org/EIPS/eip-6093
+ */
+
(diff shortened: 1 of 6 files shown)


ernestognw:
Hey @lhemerl, thanks for the contribution.

For context, EIP-6093 was published for community feedback before we implement it into OpenZeppelin Contracts for the following reasons:

1. We don't want to make arbitrary decisions on the way custom errors should look, and we're rather asking for feedback (this is the [official discussion](https://ethereum-magicians.org/t/eip-6093-custom-errors-for-commonly-used-tokens))
2. We'll try to use the same rationale there for migrating all the contracts in 5.0
3. Making the changes as of now will be breaking for the library as stated in our [API Stability docs](https://docs.openzeppelin.com/contracts/4.x/releases-stability#api-stability)

Also, [we'll deprecate `ERC777` in 4.9](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4066) so the testing issues on it can wait.

I'll close since it's too soon for 5.0 changes, but feel free to join the 6093 conversation, especially if you saw any particular use case we didn't consider while you were working on this.

ThisCoolDude:
ERC777 token standard is longer supported by OpenZeppelin. Our implementation is now deprecated and will be removed in the next major release. The corresponding standard interfaces remain available.

Is this only for now or forever. There isn't much information on why it's being deprecated

PR #4140, opened 2023-03-26
Check Address Zero and Custom Errors

### LINKED ISSUE #3862: Reject address zero in AccessControl
The AccessControl contract was implemented without checking that address arguments are not the zero address. This is because we have [gone back and forth](https://forum.openzeppelin.com/t/removing-address-0x0-checks-from-openzeppelin-contracts/2222) on the actual relevance of these checks. My current thinking is that the checks do provide value as a layer of protection against using address 0 accidentally, either after reading it from an uninitialized address in storage or as the return value of ecrecover.

Thus, I would suggest rejecting address zero when granting and revoking roles in AccessControl.

lhemerly:
Fixes #3862

add checks for zero address in AccessControl.sol

Refactor requires to custom errors in AccessControl.sol and dependencies TimeLockController.sol, GovernorTimelockControl.sol and AccessControlDefaultAdminRules

Draft as I'm working on updating tests to reflect the custom errors.

- [ ] Tests - WIP
- [ ] Documentation

### DIFF
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -8,6 +8,13 @@ import "../utils/Context.sol";
 import "../utils/Strings.sol";
 import "../utils/introspection/ERC165.sol";
 
+/** @dev Custom errors introduced in solidity 0.8.4 */
+
+error AccessControlCanOnlyRenounceRolesForSelf(address sender, bytes32 role);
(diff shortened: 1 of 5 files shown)


lhemerly:
From the feedback on custom errors PR from issue #2839   I'm closing this because it contains custom error features. Will redo the PR without it.

PR #4141, opened 2023-03-28
Add requires for account != zero address

### LINKED ISSUE #3862: Reject address zero in AccessControl

lhemerly:
Fixes #3862

Add require that address is different from zero address for grant and renounce role of access control contract.

Didn't identify the need to change the docs, but included the reverts on comments.

Have a failed test on migrate-imports which already happens in master.

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/two-pandas-pump.md
+++ b/.changeset/two-pandas-pump.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+Add reject zero address on grant and revoke roles in AccessControl
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -133,8 +133,8 @@ abstract contract AccessControl is Context, IAccessControl, ERC165 {
     /**
      * @dev Grants `role` to `account`.
      *
-     * If `account` had not been already granted `role`, emits a {RoleGranted}
-     * event.
+     * If `account` had not been already granted `role`, and it's not the
+     * zero address emits a {RoleGranted} event.
      *
      * Requirements:
      *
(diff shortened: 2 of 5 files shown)


Amxx:
Hello @lhemerly 

I think there are many issues with this change, and this is why we need the change to be discussed in the issue BEFORE making a PR to implement them.

First issue is specific to your implementation: 
You check that the address is not 0 in the public functions and not in the internal ones. That means that an internal operation can give a role to the zero address.  But the admin for that role won't be able to remove it using the public interface. (similarly, an internal function can remove, and the admin cannot re-grant using the public interface)

A second issue is backward compatibility:
This is breaking a behavior that is currently present and that people might rely one.

A third issue is a consequence of issue number two:
We actively use "granting role to address 0" in the `onlyRoleOrOpenRole` modifier in `TimelockController`. So this change is not only breaking a behavior that might be used by other devs, its breaking a behavior that we use in our contracts.

My personnal opinion is that this change breaks things for no good reason. And it should not be applied. As I said, this needs to be discussed, and we should only move to an implementation after this concerns addressed, if we still think its a good idea.

PR #4142, opened 2023-03-28
Add internal getter for pure token URI

### LINKED ISSUE #3023: ERC721URIStorage: change _tokenURIs visibility from private to internal

lhemerly:
Fixes #3023

Add getter for pure token URI

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/hip-vans-crash.md
+++ b/.changeset/hip-vans-crash.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`ERC721URIStorage`: added internal getter for retrieving token URI without base URI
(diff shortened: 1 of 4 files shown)


Amxx:
duplicate of #3878

PR #4143, opened 2023-03-28
Add Payment Splitter Payees Length Getter

### LINKED ISSUE #3818: Add PaymentSplitter.payeesLength()
**🧐 Motivation**

I want to list all payees of a `PaymentSplitter` contract.

**📝 Details**

`PaymentSplitter` has a `.payee(index)` method but it is hard to use it without knowing how many payees are there.

Consider adding this to the `PaymentSplitter` contract:
```solidity
function payeesLength() public view returns (uint256) {
  return _payees.length;
}
```

lhemerly:
Fixes #3818

Add payment splitter payess length getter

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/dirty-clocks-smoke.md
+++ b/.changeset/dirty-clocks-smoke.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-Payment Splitter: add payees length getter
+`PaymentSplitter`: Add `getPayeesLength()` function returning the 
--- a/.changeset/dirty-clocks-smoke.md
+++ b/.changeset/dirty-clocks-smoke.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-`PaymentSplitter`: Add `getPayeesLength()` function returning the 
+`PaymentSplitter`: Add `totalPayees()` function returning the length of the `_payees` array
--- a/.changeset/dirty-clocks-smoke.md
+++ b/.changeset/dirty-clocks-smoke.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-`PaymentSplitter`: Add `totalPayees()` function returning the length of the `_payees` array
+`PaymentSplitter`: Add `totalPayees()` function returning the length of the `_payees` array.
--- a/.changeset/dirty-clocks-smoke.md
+++ b/.changeset/dirty-clocks-smoke.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Payment Splitter: add payees length getter
--- a/contracts/finance/PaymentSplitter.sol
+++ b/contracts/finance/PaymentSplitter.sol
@@ -184,12 +184,9 @@ contract PaymentSplitter is Context {
         emit ERC20PaymentReleased(token, account, payment);
     }
 
-    /**
-     *
-     * @dev getter for payees length.
-     *
-     */
-
+/**
+ * @dev Getter for the amount of payees.
+ */
     function getPayeesLength() public view returns (uint256) {
         return _payees.length;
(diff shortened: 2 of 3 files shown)


contracts/finance/PaymentSplitter.sol
```diff
@@ -184,6 +184,16 @@ contract PaymentSplitter is Context {
         emit ERC20PaymentReleased(token, account, payment);
     }
 
+    /**
+     *
+     * @dev getter for payees length.
+     *
+     */
+
```
ernestognw, line 192: This is compiled into the documentation so it's important to present it formatted

```suggestion
/**
 * @dev Getter for the amount of payees.
 */
```

contracts/finance/PaymentSplitter.sol
```diff
@@ -184,6 +184,16 @@ contract PaymentSplitter is Context {
...
+    function getPayeesLength() public view returns (uint256) {
```
ernestognw, line 193: I think it makes sense to call it `totalPayees` or just `payees` to follow the contract convention. I'd wait for comments from @Amxx

.changeset/dirty-clocks-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Payment Splitter: add payees length getter
```
ernestognw, line 5:
```suggestion
`PaymentSplitter`: Add `getPayeesLength()` function returning the 
```

ernestognw (commented):
Thanks @lhemerly! I'll wait for comments from @Amxx

contracts/finance/PaymentSplitter.sol
Amxx: `totalPayees()` or `payeeCount()` would either be fine by me.

Amxx: Also, this function should move to line 123, just after `payee(uint256)`

.changeset/dirty-clocks-smoke.md
ernestognw: Sorry, just realized I didn't finish this suggestion. I'll send a commit with adjustments

contracts/finance/PaymentSplitter.sol
ernestognw: I went for `totalPayees` and put the function after `totalReleased` following the order of `total*` getters followed by their `*` counterparts in the same order.

ernestognw (approved):
This looks good to me. Thanks, @lhemerly.
As per policy, we require 2 people to approve a PR before merging, so you'd expect either @Amxx approval or more follow-up comments (if any).

Amxx:
With 4.9 audit being under way, we don't want to add anything in the master branch that would get in the release without having been audited. Either we merge that into another branch (v4.10?) of we freeze all merges until 4.9 has been released.

contracts/finance/PaymentSplitter.sol
```diff
@@ -92,6 +92,13 @@ contract PaymentSplitter is Context {
         return _erc20TotalReleased[token];
     }
 
+    /**
+     * @dev Getter for the amount of payees.
+     */
+    function totalPayees() public view returns (uint256) {
+        return _payees.length;
+    }
+
```
Amxx, line 101: Also, as I said before (somehow my comment was missed), this new function should go just after `function payee(uint256)`.

Function ordering is important for readability / auditability

contracts/finance/PaymentSplitter.sol
Amxx: it has nothing to do with `totalReleased`. This function works in pair with `function payee(uint256)`, so they should go together.

frangio:
PaymentSplitter has been removed for now in 5.0.

PR #4144, opened 2023-03-28
Moved owner check from public to internal function in ERC721 contract…

ViharGandhi:
… - resolves issue #4136

Updated ERC721._approve function to include a check for the owner before approval is granted. This ensures that the owner is not accidentally approved and prevents unnecessary sload calls. Moved the check from the public ERC721.approve function to the internal ERC721._approve function. Resolves issue #4136.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -364,9 +364,10 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
      * Emits an {Approval} event.
      */
     function _approve(address to, uint256 tokenId) internal virtual {
+        address owner = ERC721.ownerOf(tokenId);
         require(to != owner, "ERC721: approval to current owner");
         _tokenApprovals[tokenId] = to;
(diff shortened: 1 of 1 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -111,7 +111,7 @@ contract ERC721 is Context, ERC165, IERC721, IERC721Metadata {
      */
     function approve(address to, uint256 tokenId) public virtual override {
         address owner = ERC721.ownerOf(tokenId);
-        require(to != owner, "ERC721: approval to current owner");
+
```
Amxx, line 114:
```suggestion
```

ernestognw:
For reference, there's also [another PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4138) for this. 👍🏻 to @Amxx comments.

I'd also suggest adding a Changelog entry, it's technically breaking and I think it's valuable to have an entry

PR #4145, opened 2023-03-30
Update IERC3156FlashBorrower.sol

aviggiano:
Fix documentation as per ERC-3156

> The lender MUST verify that the onFlashLoan callback returns the keccak256 hash of “ERC3156FlashBorrower.onFlashLoan”.


Amxx (approved):
Thanks @aviggiano for spotting this!

PR #4146, opened 2023-03-30
Add fuzz tests for the Checkpoints library

Amxx:
#### PR Checklist

- [x] Tests


test/utils/Checkpoints.t.sol
```diff
@@ -0,0 +1,276 @@
...
+
+            // push
+            _ckpts.push(key, value);
+
+            // check length & latest
+            assertTrue(_ckpts.length() == i + 1 - duplicates);
+            assertTrue(_ckpts.latest() == value);
+            _assertLatestCheckpoint(true, key, value);
+        }
+    }
```
ernestognw, line 240: I have the feeling this function is checking multiple things:

- Initial state
- Duplicates
- `length` and `latest` consistency

I think this is a similar conversation to the one we had for the Certora rules (smaller tests vs a single big one). However, in the case of fuzzing, I see value in making individual tests so we can configure the fuzzer on purpose. For example, not bounding the `keys` may make sense for testing duplicates.

Do you think makes sense splitting the tests?

test/utils/Checkpoints.t.sol
```diff
@@ -0,0 +1,276 @@
...
+        (bool _exist, uint96 _key, uint160 _value) = _ckpts.latestCheckpoint();
+        assertTrue(_exist == exist);
+        assertTrue(_key == key);
+        assertTrue(_value == value);
+    }
+
+    // tests
+    function testPush(uint96[] memory keys, uint160[] memory values) public {
+        vm.assume(values.length > 0);
+        _prepareKeys(keys, 64);
```
ernestognw, line 219: Why 64 in both tests? I see it makes sense in the lookup tests because of the binary search but I don't see a good reason for using it here.

test/utils/Checkpoints.t.sol
```diff
@@ -0,0 +1,276 @@
...
+            // track expected result of lookups
+            if (key <= lookup) {
+                upper = value;
+            }
+            if (key >= lookup && (i == 0 || keys[i - 1] < lookup)) {
+                lowerKey = key;
+            }
+            if (key == lowerKey) {
+                lower = value;
+            }
```
ernestognw, line 268:
```suggestion
            if (key >= lookup && (i == 0 || keys[i - 1] < lookup)) {
                lowerKey = key;
                lower = value;
            }
```

ernestognw:
If we're generating tests in Solidity via script, I think it'll start conflicting with `.prettierc`.
Should we ignore the `tests` folder for prettier? Otherwise we may need to run formatting on the generated tests

test/utils/Checkpoints.t.sol
Amxx: No that would not work. 
```solidity
if (key >= lookup && (i == 0 || keys[i - 1] < lookup)) {
    lowerKey = key;
}
```
is here to identify the key of the "lower lookup" checkpoint, and  
```
if (key == lowerKey) {
    lower = value;
}
```
is here to get the last value pushed to that checkpoint (in case of duplicated key)

test/utils/Checkpoints.t.sol
Amxx: There is no real point to having crazy gaps between checkpoints, and I we need the keys to not overflow uint32. I also want to get some key duplication, which is not likely to happen if the allowed space is too big.

test/utils/Checkpoints.t.sol
Amxx: I'm not sure how splitting that test would make it better. I believe it would decrease readability. 

This is about checking that the `push` method updates the structure correctly. We do the same in FV:
- run an arbitrary push function
- check that the affected view functions are updated correctly. (`length()`, `latest()`, `latestCheckpoint()`)

scripts/generate/run.js
```diff
@@ -35,3 +29,21 @@ for (const [file, template] of Object.entries({
...
 }
+
+// Contracts
+for (const [file, template] of Object.entries({
+  'utils/math/SafeCast.sol': './templates/SafeCast.js',
+  'utils/structs/EnumerableSet.sol': './templates/EnumerableSet.js',
+  'utils/structs/EnumerableMap.sol': './templates/EnumerableMap.js',
+  'utils/Checkpoints.sol': './templates/Checkpoints.js',
+  'utils/StorageSlot.sol': './templates/StorageSlot.js',
+})) {
```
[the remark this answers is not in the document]
Amxx: Also, if you check the diff carefully, you'll see that this syntax was already present. I did not change/introduce it.

scripts/generate/run.js
```diff
@@ -35,3 +29,21 @@ for (const [file, template] of Object.entries({
...
+  generateFromTemplate(file, template, './contracts/');
+}
+
+// Tests
+for (const [file, template] of Object.entries({
```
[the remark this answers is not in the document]
Amxx: same answer

scripts/generate/templates/Checkpoints.test.js
frangio, line 1: I would rename this file to `Checkpoints.t.js` to match the Solidity file naming.

test/utils/Checkpoints.t.sol
```diff
@@ -0,0 +1,276 @@
...
+    function _assertLatestCheckpoint(bool exist, uint32 key, uint224 value) internal {
+        (bool _exist, uint32 _key, uint224 _value) = _ckpts.latestCheckpoint();
+        assertTrue(_exist == exist);
+        assertTrue(_key == key);
+        assertTrue(_value == value);
+    }
+
+    // tests
+    function testPush(uint32[] memory keys, uint224[] memory values) public {
+        vm.assume(values.length > 0);
```
frangio, line 38: Do we want to add this? Feels weird that without this condition Foundry will invest time into generating cases that we will not fully consider, and will count them as successful fuzz runs, and we may have wanted to use those runs to explore more of the "useful" argument space.

This applies to `testLookup` as well.

```suggestion
        vm.assume(values.length > 0 && values.length <= keys.length);
```

test/utils/Checkpoints.t.sol
frangio, line 1: We are writing `assertTrue(a == b)` but ideally we should be using `assertEq(a, b)` (it produces better error messages).

scripts/generate/templates/Checkpoints.test.js
```diff
@@ -0,0 +1,207 @@
...
+        ${opts.valueTypeName} value = values[i % values.length];
+
+        // push
+        _ckpts.push(key, value);
+
+        // track expected result of lookups
+        if (key <= lookup) {
+            upper = value;
+        }
+        if (key >= lookup && (i == 0 || keys[i-1] < lookup)) {
```
frangio, line 108:
```suggestion
        // store the first key that is not smaller than the lookup key
        if (key >= lookup && (i == 0 || keys[i-1] < lookup)) {
```

test/utils/Checkpoints.t.sol
```diff
@@ -0,0 +1,276 @@
...
+            if (i > 0 && key == keys[i - 1]) ++duplicates;
+
+            // push
+            _ckpts.push(key, value);
+
+            // check length & latest
+            assertTrue(_ckpts.length() == i + 1 - duplicates);
+            assertTrue(_ckpts.latest() == value);
+            _assertLatestCheckpoint(true, key, value);
+        }
```
frangio, line 151: Perhaps we can add a final push after the loop, with a random key lower than the latest checkpoint, and check that it reverts?

frangio (commented):
Looks pretty good!

test/utils/Checkpoints.t.sol
frangio: It's true that this function is checking for a lot of things but I think it makes sense to keep them coupled in one test. Because we want to test a sequence of pushes and assert properties at each step.

test/utils/Checkpoints.t.sol
frangio: 64 seems fine to me. But I would define it as a constant at the top of the file and explain the rationale.

test/utils/Checkpoints.t.sol
frangio: Yes, this is not obvious from the code, but the separate `if` is needed to process duplicate keys.

frangio:
>If we're generating tests in Solidity via script, I think it'll start conflicting with .prettierc.
>Should we ignore the tests folder for prettier? Otherwise we may need to run formatting on the generated tests

@ernestognw The Solidity generation scripts run Prettier on the generated code.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cf86fd9962701396457e50ab0d6cc78aa29a5ebc/scripts/generate/run.js#L36

Does this address your concern?

I don't think there will be conflicts?

scripts/generate/run.js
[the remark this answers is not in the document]
ernestognw: > but it would be easy to get a `[` wrong.

This argument is strong enough, let's keep it!

test/utils/Checkpoints.t.sol
ernestognw: > @ernestognw The Solidity generation scripts run Prettier on the generated code.
> 
> https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cf86fd9962701396457e50ab0d6cc78aa29a5ebc/scripts/generate/run.js#L36
> 
> Does this address your concern?

Yes it does, thanks!

ernestognw:
> @ernestognw The Solidity generation scripts run Prettier on the generated code.
> 
> https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cf86fd9962701396457e50ab0d6cc78aa29a5ebc/scripts/generate/run.js#L36
> 
> Does this address your concern?

Yes it does, thanks!

test/utils/Checkpoints.t.sol
ernestognw: > Because we want to test a sequence of pushes and assert properties at each step.

Okay now I see the point, but still, feel the initial assertions are unnecessary. Both keeping them or removing is fine for me so I'm solving

test/utils/Checkpoints.t.sol
ernestognw: Agree, let's make it a constant with a comment.

test/utils/Checkpoints.t.sol
frangio: That's true, the initial state is kind of unnecessary to check every time. I'd let @Amxx decide whether to move that to a separate test (which wouldn't be fuzzed since it's just the initial state).

Amxx: The first check is also a sanity check, that makes we are starting with a clean state. Given our limited understanding of how foundry setup/checkpoint the contract, I felt it was better to have it.

frangio (approved):

frangio:
`vm.expectRevert` was being used wrong in this PR and leading to very strange behavior. It's important to remember that `vm.expectRevert` only works with external calls. In this case it was used prior to `ckpts.push(...)` which is an internal call. I've refactor that to use an auxiliary external function. @ernestognw please review that part.

ernestognw (approved):

PR #4147, opened 2023-03-31
Add DoubleEndedQueue FV

ernestognw:
Fixes LIB-744

Adds Formal Verification rules for Double Ended Queue.

### DIFF
--- a/certora/specs/DoubleEndedQueue.spec
+++ b/certora/specs/DoubleEndedQueue.spec
@@ -81,7 +81,31 @@ invariant lengthConsistency()
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 invariant outOfBoundsDefault(int128 key)
-    key < begin() && key >= end() && at_(to_uint256(key)) == 0
+    (
+        key < begin() &&  // Is before beginning
+        key >= end() // Is after ending
+    ) => at_(to_uint256(key)) == 0
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: empty() is length 0 and no element exists                                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant emptiness(int128 key)
+    empty() <=> length() == 0 && at_(to_uint256(key)) == 0
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: front() and back() point to the [begin, end) range limits                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant pointerCorrectness()
+    back() == at_(to_uint256(begin())) && front() == at_(to_uint256(end()) - 1)
+    {
+        preserved {
+            requireInvariant boundariesConsistency;
+        }
+    }
 
 /*
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
@@ -89,19 +113,19 @@ invariant outOfBoundsDefault(int128 key)
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 rule pushFront {
+    // Sanity
     requireInvariant boundariesConsistency;
-
+    
     bytes32 value;
 
     int128 beginBefore = begin();
     int128 beginNext = beginBefore - 1;
-
-    bytes32 valueBefore = at_(0);
+    bytes32 frontBefore = front();
 
     pushFront@withrevert(value);
 
     int128 beginAfter = begin();
-    bytes32 valueAfter = at_(0);
+    bytes32 frontAfter = front();
 
     bool success = !lastReverted;
 
@@ -109,10 +133,13 @@ rule pushFront {
     assert success;
 
     // effect
-    assert success => valueAfter == value && beginAfter == beginNext, "value set, and begin adjusted";
+    assert success => (
+        frontAfter == value && // Front set to value
+        beginAfter == beginNext // Begin is adjusted
+    ), "front set to value and begin is adjusted";
 
     // no side effect
-    assert valueBefore == valueAfter => valueBefore == value, "value doesn't change only if it was already set";
+    assert frontBefore == frontAfter <=> frontBefore == value, "front doesn't change only if it was already set";
 }
 
 /*
@@ -121,19 +148,19 @@ rule pushFront {
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 rule pushBack {
+    // Sanity
     requireInvariant boundariesConsistency;
 
     bytes32 value;
 
     int128 endBefore = end();
     int128 endNext = endBefore + 1;
-    
-    bytes32 valueBefore = at_(to_uint256(end() - 1));
+    bytes32 backBefore = back();
 
     pushBack@withrevert(value);
 
     int128 endAfter = end();
-    bytes32 valueAfter = at_(to_uint256(end() - 1));
+    bytes32 backAfter = back();
 
     bool success = !lastReverted;
 
@@ -141,9 +168,73 @@ rule pushBack {
     assert success;
 
     // effect
-    assert success => valueAfter == value && endAfter == endNext, "value set, and end adjusted";
+    assert success => (
+        backAfter == value && // Back set to true
+        endAfter == endNext // End is adjusted
+    ), "back set to value and begin is adjusted";
 
     // no side effect
-    assert valueBefore == valueAfter => valueBefore == value, "value doesn't change only if it was already set";
+    assert backBefore == backAfter <=> backBefore == value, "back doesn't change only if it was already set";
 }
 
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Function correctness: popFront removes an element from the beginning of the queue                                   │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popFront {
+    requireInvariant boundariesConsistency;
+    bytes32 value;
+    require !empty();
+
+    int128 beginBefore = begin();
+    int128 beginNext = beginBefore + 1;
+    bytes32 frontBefore = front();
+
+    popFront@withrevert();
+
+    int128 beginAfter = begin();
+    bytes32 frontAfter = front();
+
+    bool success = !lastReverted;
+
+    // liveness
+    assert success;
+
+    // effect
+    assert success => beginAfter == beginNext, "begin is adjusted ";
+
+    // no side effect
+    assert frontBefore == frontAfter <=> frontBefore == value, "front doesn't change only if it was already set";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Function correctness: popBack removes an element from the end of the queue                                          │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popBack {
+    requireInvariant boundariesConsistency;
+    bytes32 value;
+    require !empty();
+
+    int128 endBefore = end();
+    int128 endNext = endBefore - 1;
+    bytes32 backBefore = back();
+
+    popBack@withrevert();
+
+    int128 endAfter = end();
+    bytes32 backAfter = back();
+
+    bool success = !lastReverted;
+
+    // liveness
+    assert success;
+
+    // effect
+    assert success => endAfter == endNext, "begin is adjusted ";
+
+    // no side effect
+    assert backBefore == backAfter <=> backBefore == value, "back doesn't change only if it was already set";
+}
--- a/certora/specs/DoubleEndedQueue.spec
+++ b/certora/specs/DoubleEndedQueue.spec
@@ -242,7 +242,7 @@ rule popBackConsistency(uint256 key) {
     requireInvariant boundariesConsistency();
     require boundedQueue();
 
-    // Read (any) value that is not the front (this assert the value exist / the queue is long enough)
+    // Read (any) value that is not the back (this asserts the value exists / the queue is long enough)
     require key < length() - 1;
     bytes32 before = at_(key);
 
--- a/certora/specs/DoubleEndedQueue.spec
+++ b/certora/specs/DoubleEndedQueue.spec
@@ -110,16 +110,16 @@ rule pushFront(bytes32 value) {
 │ Rule: pushFront preserves the previous values in the queue with a +1 offset                                         │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule pushFrontConsistency(uint256 key) {
+rule pushFrontConsistency(uint256 index) {
     require boundedQueue();
 
-    bytes32 beforeAt = at_(key);
+    bytes32 beforeAt = at_(index);
 
     bytes32 value;
     pushFront(value);
 
     // try to read value
-    bytes32 afterAt = at_@withrevert(key + 1);
+    bytes32 afterAt = at_@withrevert(index + 1);
 
     assert !lastReverted, "value still there";
     assert afterAt == beforeAt, "data is preserved";
@@ -150,16 +150,16 @@ rule pushBack(bytes32 value) {
 │ Rule: pushBack preserves the previous values in the queue                                                           │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule pushBackConsistency(uint256 key) {
+rule pushBackConsistency(uint256 index) {
     require boundedQueue();
 
-    bytes32 beforeAt = at_(key);
+    bytes32 beforeAt = at_(index);
 
     bytes32 value;
     pushBack(value);
 
     // try to read value
-    bytes32 afterAt = at_@withrevert(key);
+    bytes32 afterAt = at_@withrevert(index);
 
     assert !lastReverted, "value still there";
     assert afterAt == beforeAt, "data is preserved";
@@ -193,18 +193,18 @@ rule popFront {
 │ Rule: at(x) is preserved and offset to at(x - 1) after calling popFront                                             |
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule popFrontConsistency(uint256 key) {
+rule popFrontConsistency(uint256 index) {
     requireInvariant boundariesConsistency();
     require boundedQueue();
 
     // Read (any) value that is not the front (this asserts the value exists / the queue is long enough)
-    require key > 1;
-    bytes32 before = at_(key);
+    require index > 1;
+    bytes32 before = at_(index);
 
     popFront();
 
     // try to read value
-    bytes32 after = at_@withrevert(key - 1);
+    bytes32 after = at_@withrevert(index - 1);
     
     assert !lastReverted, "value still exists in the queue";
     assert before == after, "values are offset and not modified";
@@ -238,18 +238,18 @@ rule popBack {
 │ Rule: at(x) is preserved after calling popBack                                                                     |
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule popBackConsistency(uint256 key) {
+rule popBackConsistency(uint256 index) {
     requireInvariant boundariesConsistency();
     require boundedQueue();
 
     // Read (any) value that is not the back (this asserts the value exists / the queue is long enough)
-    require key < length() - 1;
-    bytes32 before = at_(key);
+    require index < length() - 1;
+    bytes32 before = at_(index);
 
     popBack();
 
     // try to read value
-    bytes32 after = at_@withrevert(key);
+    bytes32 after = at_@withrevert(index);
     
     assert !lastReverted, "value still exists in the queue";
     assert before == after, "values are offset and not modified";
@@ -298,16 +298,16 @@ rule onlyEmptyRevert(env e) {
 
 /*
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
-│ Rule: at(key) only reverts if key is out of bounds                                                                  |
+│ Rule: at(index) only reverts if index is out of bounds                                                                  |
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule onlyOutOfBoundsRevert(uint256 key) {
+rule onlyOutOfBoundsRevert(uint256 index) {
     requireInvariant boundariesConsistency();
     require boundedQueue();
 
-    at_@withrevert(key);
+    at_@withrevert(index);
 
-    assert lastReverted <=> key >= length(), "only reverts if key is out of bounds";
+    assert lastReverted <=> index >= length(), "only reverts if index is out of bounds";
 }
 
 /*
@@ -347,16 +347,16 @@ rule noDataChange(env e) {
     method f;
     calldataarg args;
 
-    uint256 key;
-    bytes32 atBefore = at_(key);
+    uint256 index;
+    bytes32 atBefore = at_(index);
     f(e, args);
-    bytes32 atAfter = at_@withrevert(key);
+    bytes32 atAfter = at_@withrevert(index);
     bool atAfterSuccess = !lastReverted;
 
     assert !atAfterSuccess <=> (
         f.selector == clear().selector ||
-        (f.selector == popBack().selector && key == length()) ||
-        (f.selector == popFront().selector && key == length())
+        (f.selector == popBack().selector && index == length()) ||
+        (f.selector == popFront().selector && index == length())
     ), "indexes of the queue are only removed by clear or pop";
 
     assert atAfterSuccess && atAfter != atBefore => (
--- a/certora/specs/DoubleEndedQueue.spec
+++ b/certora/specs/DoubleEndedQueue.spec
@@ -337,7 +337,7 @@ rule noLengthChange(env e) {
 
 /*
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
-│ Rule: only push/pop can change values bounded in the queue (outters aren't cleared)                                 │
+│ Rule: only push/pop can change values bounded in the queue (outside values aren't cleared)                          │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 rule noDataChange(env e) {
@@ -350,12 +350,17 @@ rule noDataChange(env e) {
     uint256 key;
     bytes32 atBefore = at_(key);
     f(e, args);
-    bytes32 atAfter = at_(key);
+    bytes32 atAfter = at_@withrevert(key);
+    bool atAfterSuccess = !lastReverted;
 
-    assert atAfter != atBefore => (
-        f.selector == clear().selector || // Although doesn't change values, outters are symbolic, so `* != *`
-        (f.selector == popBack().selector && key == length()) || // at_ only changes if it's left out, becoming symbolic 
+    assert !atAfterSuccess <=> (
+        f.selector == clear().selector ||
+        (f.selector == popBack().selector && key == length()) ||
+        (f.selector == popFront().selector && key == length())
+    ), "indexes of the queue are only removed by clear or pop";
+
+    assert atAfterSuccess && atAfter != atBefore => (
         f.selector == popFront().selector ||
         f.selector == pushFront(bytes32).selector
-    ), "values of the queue are only changed by a clear, pop or push operation";
+    ), "values of the queue are only changed by popFront or pushFront";
 }
--- a/certora/specs/DoubleEndedQueue.spec
+++ b/certora/specs/DoubleEndedQueue.spec
@@ -90,26 +90,19 @@ invariant queueEndings()
 │ Function correctness: pushFront adds an element at the beginning of the queue                                       │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule pushFront {
+rule pushFront(bytes32 value) {
     require boundedQueue();
+
+    uint256 lengthBefore = length();
     
-    bytes32 value;
-
-    bytes32 frontBefore = front();
-
     pushFront@withrevert(value);
-    bool success = !lastReverted;
-
-    bytes32 frontAfter = front();
-
+    
     // liveness
-    assert success, "never reverts";
-
+    assert !lastReverted, "never reverts";
+    
     // effect
-    assert success => frontAfter == value, "front set to value";
-
-    // no side effect
-    assert frontBefore == frontAfter <=> frontBefore == value, "front doesn't change only if it was already set";
+    assert front() == value, "front set to value";
+    assert length() == lengthBefore + 1, "queue extended";
 }
 
 /*
@@ -136,26 +129,19 @@ rule pushFrontConsistency {
 │ Function correctness: pushBack adds an element at the end of the queue                                              │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule pushBack {
+rule pushBack(bytes32 value) {
     require boundedQueue();
 
-    bytes32 value;
-
-    bytes32 backBefore = back();
-
-    pushBack(value);
-    bool success = !lastReverted;
-
-    bytes32 backAfter = back();
-
+    uint256 lengthBefore = length();
+    
+    pushBack@withrevert(value);
+    
     // liveness
-    assert success, "never reverts";
-
+    assert !lastReverted, "never reverts";
+    
     // effect
-    assert success => backAfter == value, "back set to value";
-
-    // no side effect
-    assert backBefore == backAfter <=> backBefore == value, "back doesn't change only if it was already set";
+    assert back() == value, "back set to value";
+    assert length() == lengthBefore + 1, "queue increased";
 }
 
 /*
@@ -204,17 +190,21 @@ rule popFront {
 │ Rule: at(x) is preserved and offset to at(x - 1) after calling popFront                                             |
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule popFrontConsistency {
+rule popFrontConsistency(uint256 key) {
     requireInvariant boundariesConsistency();
     require boundedQueue();
 
-    uint256 key;
+    // Read (any) value that is not the front (this assert the value exist / the queue is long enough)
+    require key > 1;
     bytes32 before = at_(key);
 
-    popFront@withrevert();
-    bool success = !lastReverted;
+    popFront();
+
+    // try to read value
+    bytes32 after = at_@withrevert(key - 1);
     
-    assert !lastReverted <=> before == at_(key - 1), "values are offset and not modified";
+    assert !lastReverted, "value still exists in the queue";
+    assert before == after, "values are offset and not modified";
 }
 
 /*
@@ -244,17 +234,21 @@ rule popBack {
 │ Rule: at(x) is preserved after calling popBack                                                                     |
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-rule popBackConsistency {
+rule popBackConsistency(uint256 key) {
     requireInvariant boundariesConsistency();
     require boundedQueue();
 
-    uint256 key;
+    // Read (any) value that is not the front (this assert the value exist / the queue is long enough)
+    require key < length() - 1;
     bytes32 before = at_(key);
 
-    popBack@withrevert();
-    bool success = !lastReverted;
+    popBack();
+
+    // try to read value
+    bytes32 after = at_@withrevert(key);
     
-    assert !lastReverted <=> before == at_(key), "values are not modified";
+    assert !lastReverted, "value still exists in the queue";
+    assert before == after, "values are offset and not modified";
 }
 
 /*
@@ -332,23 +326,13 @@ rule noLengthChange(env e) {
     f(e, args);
     uint256 lengthAfter = length();
 
-    assert lengthAfter > lengthBefore => (
-        f.selector == pushFront(bytes32).selector ||
-        f.selector == pushBack(bytes32).selector
-    ), "length increases only with push operations";
-    
-    assert lengthAfter < lengthBefore => (
-        f.selector == popBack().selector ||
-        f.selector == popFront().selector ||
-        f.selector == clear().selector
-    ), "length decreases only with clear/pop operations";
-
-    assert (
-        lengthAfter - lengthBefore != 1 &&
-        lengthAfter - lengthBefore == 1
-    ) => (
-        f.selector == clear().selector
-    ), "length changes by more than one only with clear operations";
+    assert lengthAfter != lengthBefore => (
+        (f.selector == pushFront(bytes32).selector && lengthAfter == lengthBefore + 1) ||
+        (f.selector == pushBack(bytes32).selector && lengthAfter == lengthBefore + 1) ||
+        (f.selector == popBack().selector && lengthAfter == lengthBefore - 1) ||
+        (f.selector == popFront().selector && lengthAfter == lengthBefore - 1) ||
+        (f.selector == clear().selector && lengthAfter == 0)
+    ), "length is only affected by clear/pop/push operations";
 }
 
 /*
--- a/certora/specs/DoubleEndedQueue.spec
+++ b/certora/specs/DoubleEndedQueue.spec
@@ -63,9 +63,23 @@ invariant lengthConsistency()
 │ Invariant: empty() is length 0 and no element exists                                                                │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-invariant emptiness(uint256 idx)
+invariant emptiness()
     empty() <=> (length() == 0 && begin() == end())
     { preserved { require boundedQueue(); } }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: front points to the first index and back points to the last one                                          │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant queueEndings()
+    at_(length() - 1) == back() && at_(0) == front()
+    { 
+        preserved { 
+            requireInvariant boundariesConsistency(); 
+            require boundedQueue(); 
+        } 
+    }
 
 /*
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
@@ -287,7 +301,7 @@ rule clear {
 
 /*
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
-│ Rule: front/back access reverts only if the queue is empty                                                      │
+│ Rule: front/back access reverts only if the queue is empty                                                          │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 rule onlyEmptyRevert(env e) {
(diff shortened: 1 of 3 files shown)


certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+function absolute_diff(mathint a, mathint b) returns mathint {
+    return a > b ? a - b : b - a; 
+}
+
+function min_int128() returns mathint {
+    return max_uint256 / 2 + 1;
+}
+
+function max_int128() returns mathint {
+    return max_uint256 / 2;
```
Amxx, line 37: This should be available
```suggestion
    return max_uint128;
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+│ Helpers                                                                                                             │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+
+function absolute_diff(mathint a, mathint b) returns mathint {
+    return a > b ? a - b : b - a; 
+}
+
+function min_int128() returns mathint {
+    return max_uint256 / 2 + 1;
```
Amxx, line 33: This is not `min_int128`, its `neg_min_int128`
```suggestion
    return max_uint128 + 1;
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+}
+
+function max_int128() returns mathint {
+    return max_uint256 / 2;
+}
+
+// Could be broken in theory, but not in practice
+function beginLimited(int128 begin) returns bool {
+    return min_int128() < to_mathint(begin); // not inclusive so it can pushFront
```
Amxx, line 42: there is an issue here. `min_int128` is positive, and is bigger than the max value of `int128`. If you `require` that you are probably creating vacuity.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: length is end minus begin                                                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant lengthConsistency()
+    length() == absolute_diff(to_mathint(end()), to_mathint(begin()))
```
Amxx, line 71: do we need the absolute here ? `boundariesConsistency` is `end() >= begin()` ... so it should be good without the absolute value.

I would expect that to work.
```suggestion
    length() == to_mathint(end() - begin())
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: out of bounds is always 0                                                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant outOfBoundsDefault(int128 key)
+    (
+        key < begin() &&  // Is before beginning
+        key >= end() // Is after ending
+    ) => at_(to_uint256(key)) == 0
```
Amxx, line 87: Why not directly work with a uint256 key ?

```suggestion
invariant outOfBoundsDefault(uint256 key)
    (
        key < begin() ||  // Is before beginning
        key >= end() // Is after ending
    ) => at_(key) == 0
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: empty() is length 0 and no element exists                                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant emptiness(int128 key)
+    empty() <=> length() == 0 && at_(to_uint256(key)) == 0
```
Amxx, line 95: be carefull with the associativity of `<=>` !

Here I would remove the `at_(to_uint256(key))` part, since its already covered by `outOfBoundsDefault`.
Instead I would check 

```suggestion
invariant emptiness()
    (empty() <=> length() == 0) && (empty() <=> begin() == end())
```

certora/specs/DoubleEndedQueue.spec
Amxx: Also, I think the `&&` should be a `||`  !!!

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    int128 beginBefore = begin();
+    int128 beginNext = beginBefore - 1;
+    bytes32 frontBefore = front();
+
+    pushFront@withrevert(value);
+
+    int128 beginAfter = begin();
+    bytes32 frontAfter = front();
+
+    bool success = !lastReverted;
```
Amxx, line 130: Here you are checking that the last call did not revert, but the last call is `front()` on line 128. You need to reorder
```suggestion
    pushFront@withrevert(value);
    bool success = !lastReverted;

    int128 beginAfter = begin();
    bytes32 frontAfter = front();

```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+
+    // liveness
+    assert success;
+
+    // effect
+    assert success => (
+        frontAfter == value && // Front set to value
+        beginAfter == beginNext // Begin is adjusted
```
Amxx, line 138: I think this is clearer

```suggestion
        beginAfter == beginBefore - 1 // Begin is adjusted
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    ), "front set to value and begin is adjusted";
+
+    // no side effect
+    assert frontBefore == frontAfter <=> frontBefore == value, "front doesn't change only if it was already set";
```
Amxx, line 142: It would be nice to check that `end()` and `back()` are not affected.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    int128 endBefore = end();
+    int128 endNext = endBefore + 1;
+    bytes32 backBefore = back();
+
+    pushBack@withrevert(value);
+
+    int128 endAfter = end();
+    bytes32 backAfter = back();
+
+    bool success = !lastReverted;
```
Amxx, line 165: same (ordering)

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    assert success;
+
+    // effect
+    assert success => (
+        backAfter == value && // Back set to true
+        endAfter == endNext // End is adjusted
+    ), "back set to value and begin is adjusted";
+
+    // no side effect
+    assert backBefore == backAfter <=> backBefore == value, "back doesn't change only if it was already set";
```
Amxx, line 177: same (check that begin() and front() are not affected).

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    bool success = !lastReverted;
+
+    // liveness
+    assert success;
+
+    // effect
+    assert success => beginAfter == beginNext, "begin is adjusted ";
+
+    // no side effect
+    assert frontBefore == frontAfter <=> frontBefore == value, "front doesn't change only if it was already set";
```
Amxx, line 208: I don't get what you are checking here.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+rule popFront {
+    requireInvariant boundariesConsistency;
+    bytes32 value;
+    require !empty();
+
+    int128 beginBefore = begin();
+    int128 beginNext = beginBefore + 1;
+    bytes32 frontBefore = front();
+
+    popFront@withrevert();
```
Amxx, line 194: you are not checking the value returned by `popFront()`

You whould probably be doing 
```
bytes32 value = popFront@withrevert();
bool success = !lastReverted;

assert success => value == frontBefore();
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+rule popBack {
+    requireInvariant boundariesConsistency;
+    bytes32 value;
+    require !empty();
+
+    int128 endBefore = end();
+    int128 endNext = endBefore - 1;
+    bytes32 backBefore = back();
+
+    popBack@withrevert();
```
Amxx, line 225: same here (check value)

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    bool success = !lastReverted;
+
+    // liveness
+    assert success;
+
+    // effect
+    assert success => endAfter == endNext, "begin is adjusted ";
+
+    // no side effect
+    assert backBefore == backAfter <=> backBefore == value, "back doesn't change only if it was already set";
```
Amxx, line 239: same question

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    int128 beginBefore = begin();
+    int128 beginNext = beginBefore + 1;
+    bytes32 frontBefore = front();
+
+    popFront@withrevert();
+
+    int128 beginAfter = begin();
+    bytes32 frontAfter = front();
+
+    bool success = !lastReverted;
```
Amxx, line 199: same (ordering)

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    int128 endBefore = end();
+    int128 endNext = endBefore - 1;
+    bytes32 backBefore = back();
+
+    popBack@withrevert();
+
+    int128 endAfter = end();
+    bytes32 backAfter = back();
+
+    bool success = !lastReverted;
```
Amxx, line 230: same (ordering)

Amxx (commented):
A lot of things to fix, but this is going in the right direction. Maybe add a check that does arbitrary call and check that:
- front() and begin() should only change if pop or push to front
- back() and end() should only change if pop or push to back


Also I'm not sure if we really need to go into all the trouble of looking being() and end():
- if we prove the length is changing correctly
- if we prove that `at` return 0 for anything beyond the length
- if we prove the correct behavior of pushFront, pushBack, popFront, popBack (they push/pop the right value and they update the length correctly)

Then we really don't care about being() and end(), do we ?

certora/specs/DoubleEndedQueue.spec
Amxx: Last but not least, you are confusing the `uint256` position for `at` with the `int256` position in the storage mapping. Lot at the code for `at`, we apply an offset. 

This rule should be something like 
```key >= length <=> at_(key) reverts```
which you could write as
```key < length <=> at_(key) >= 0```
(`at_(key) >= 0` being a way to say `at_(key)` returns something)

File: certora/specs/DoubleEndedQueue.spec [hunk 18]
```diff
@@ -0,0 +1,240 @@
...
+invariant emptiness(int128 key)
+    empty() <=> length() == 0 && at_(to_uint256(key)) == 0
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: front() and back() point to the [begin, end) range limits                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant pointerCorrectness()
+    back() == at_(to_uint256(begin())) && front() == at_(to_uint256(end()) - 1)
```
Amxx, line 103: You are missing the logic for `at`

```suggestion
    back() == at_(0) && front() == at_(length() - 1)
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+    {
+        preserved {
+            requireInvariant boundariesConsistency;
+        }
+    }
```
Amxx, line 108: I would expect this to not be needed

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+*/
+rule pushFront {
+    // Sanity
+    requireInvariant boundariesConsistency;
+    
+    bytes32 value;
+
+    int128 beginBefore = begin();
+    int128 beginNext = beginBefore - 1;
+    bytes32 frontBefore = front();
```
Amxx, line 123: Be carefull about this. Here you are telling the prover "`front()` must return something without reverting" ... which is implying that the queue must not be empty. So anything you prove here will be on non-empty queues.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popFront {
+    requireInvariant boundariesConsistency;
+    bytes32 value;
+    require !empty();
+
+    int128 beginBefore = begin();
+    int128 beginNext = beginBefore + 1;
+    bytes32 frontBefore = front();
```
Amxx, line 192: same (queue not empty)

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,240 @@
...
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popBack {
+    requireInvariant boundariesConsistency;
+    bytes32 value;
+    require !empty();
+
+    int128 endBefore = end();
+    int128 endNext = endBefore - 1;
+    bytes32 backBefore = back();
```
Amxx, line 223: same (queue not empty)

Amxx (commented):
There are a lot of small thing about the language that you miss, which reduces the effectiveness of the specs you write. Lets get on a call and discuss the rules we want to prove. On the top of my head I have:

- `empty <=> length == 0` (you already have that)
- `front`, `back`, `pop_front` and `pop_back` should all revert iff queue is empty
- `push_front` and `push_back` should never revert
- if the length changes, then either a push or a pop operation was call (check that the length moved correctly)
- if `at(x)` should not change if doing any operation that is not a pop, a push or clear 
- if doing push_front, `at(x)` before should be `at(x+1)` after
- if doing push back, `at(x)` before should be `at(x)` after
- if doing pop_front, the result should be the same as pop() / at(0). Any at(x) with x > 0 before should be equal to at(x-1) after
- if doing pop_back, the result should be the same as back() / at(length() - 1). Any at(x) with x < length() -2 before should be equal to at(x) after

Maybe others...

certora/specs/DoubleEndedQueue.spec
ernestognw: Thanks for this one!

certora/specs/DoubleEndedQueue.spec
ernestognw: I didn't manage to get this working. Turns out that Certora makes an implicit casting and this one doesn't overflow.
I ended up hardcoding the values so they're compatible with `mathint` everywhere.

certora/specs/DoubleEndedQueue.spec
ernestognw: I'm solving since this was completely changed, let me know if the vacuity is still there but I think we're now good

certora/specs/DoubleEndedQueue.spec
ernestognw: You're correct, `absolute_diff` was removed so I'm solving

certora/specs/DoubleEndedQueue.spec
ernestognw: This comment was particularly insightful, thank you!
I applied the suggestion and I'm also still using `at_` in some places without the `@revert` but I made sure to prove `at_` only reverts when there's a query for an out of bounds value.

certora/specs/DoubleEndedQueue.spec
ernestognw: Both are correct, I left multiple mistakes trying to make everything fit. Should be clearer now.

certora/specs/DoubleEndedQueue.spec
ernestognw: Thanks!

certora/specs/DoubleEndedQueue.spec
ernestognw: Ended up following your advice and not checking `_begin` and `_end` pointers but proving the related invariants. This check went away during the process.

certora/specs/DoubleEndedQueue.spec
ernestognw: Regarding `end()`, I don't think it's necessary since it's not part of the interface as we spoke last Friday.

Also, I ended up with several rules that prove `end` and `back` are not affected.

- `lengthConsistency` guarantees length is always `end() - begin()`
- `noLengthChange`  guarantees length is only changed by 1
- `noDataChange` proves the only operations that can change the values in the queue
- `(push/pop)(Front/Back)Consistency` prove that no values in the queue were modified

Therefore, there's no way back is affected, and even if `end` is affected, is guaranteed to be consistent.

certora/specs/DoubleEndedQueue.spec
ernestognw: This one is for checking that if the `front` didn't change, it can only be because the next value after popping the old `front` is equal to the older value

certora/specs/DoubleEndedQueue.spec
ernestognw: Wasn't compiling at the beginning but now it should be added.

ernestognw:
I did another iteration following your recommendations, and I think it's in better shape. I also answered and solved the issues I think were solved with the new iteration.

The only thing I'm not sure is about [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4147#discussion_r1154409317) regarding `front`, `at` and `back` without the `@withrevert`. I think I'm leaving the `@withrevert` out on purpose in some places since I think it's fine to exclude non-empty and (non-empty + 1) cases.

Is it a general rule to always aim for using `@withrevert` everywhere, or is it not to do so in some situations?

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+*/
+rule popFrontExpected {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+    
+    bytes32 nextFront = at_(1);
+    popFront@withrevert();
+
+    assert !lastReverted <=> front() == nextFront, "front is adjusted";
+}
```
Amxx, line 215: This is a consequence of `popFrontConsistency` + `queueEndings`. I'd remove that rule.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: at(x) is preserved and offset to at(x - 1) after calling popFront                                             |
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popFrontConsistency {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+
+    uint256 key;
+    bytes32 before = at_(key);
```
Amxx, line 227: Note that the `at_` call on line 227 is not `@withrevert` ... so it forces the queue to be non empty, which means the `popFront` will never revert.

Also, the `at_` on line 232 adds additional contraints to the rule (queue is not empty after the `popFront`)

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+
+    popFront@withrevert();
+    bool success = !lastReverted;
+    
+    assert !lastReverted <=> before == at_(key - 1), "values are offset and not modified";
+}
```
Amxx, line 233: Suggestion

```suggestion
rule popFrontConsistency(uint256 key) {
    requireInvariant boundariesConsistency();
    require boundedQueue();

    // Read (any) value that is not the front (this assert the value exist / the queue is long enough)
    require at > 1;
    bytes32 before = at_(key);

    // pop front
    popFront();
    
    // try to read value
    bytes32 after = at_@withrevert(key);
    
    assert !lastReverted, "value still exist in queue";
    assert before == after, "values are offset and not modified";
}
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+*/
+rule popBackExpected {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+    
+    bytes32 nextBack = at_(length() - 2);
+    popBack@withrevert();
+
+    assert !lastReverted => back() == nextBack, "back is adjusted";
+}
```
Amxx, line 270: This is a consequence of `popBackConsistency` + `queueEndings`. I'd remove that rule.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+    require boundedQueue();
+
+    uint256 key;
+    bytes32 before = at_(key);
+
+    popBack@withrevert();
+    bool success = !lastReverted;
+    
+    assert !lastReverted <=> before == at_(key), "values are not modified";
+}
```
Amxx, line 288: symetrical to `popFrontConsistency`

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+*/
+rule clear {
+    clear@withrevert();
+    bool success = !lastReverted;
+    
+    // liveness
+    assert success, "never reverts";
+    
+    // effect
+    assert success <=> length() == 0, "sets length to 0";
```
Amxx, line 303: While technically correct, this is over complicated. You already `assert success`. So you can just 
```suggestion
    assert length() == 0, "sets length to 0";
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+
+    f@withrevert(e, args);
+    bool success = !lastReverted;
+
+    assert !success && (
+        f.selector == front().selector ||
+        f.selector == back().selector ||
+        f.selector == popFront().selector ||
+        f.selector == popBack().selector
+    ) => emptyBefore, "only revert if empty";
```
Amxx, line 329: I would write it as

```suggestion
    f@withrevert(e, args);

    assert lastReverted => (
        (f.selector == front().selector && emptyBefore) ||
        (f.selector == back().selector && emptyBefore) ||
        (f.selector == popFront().selector  && emptyBefore) ||
        (f.selector == popBack().selector  && emptyBefore) ||
        f.selector == at_()
    ), "only revert if empty our out of bound";
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: at(key) only reverts if key is out of bounds                                                                  |
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule onlyOutOfBoundsRevert(env e) {
```
Amxx, line 337: is `e` used anywhere ?

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+        f.selector == popFront().selector ||
+        f.selector == clear().selector
+    ), "length decreases only with clear/pop operations";
+
+    assert (
+        lengthAfter - lengthBefore != 1 &&
+        lengthAfter - lengthBefore == 1
+    ) => (
+        f.selector == clear().selector
+    ), "length changes by more than one only with clear operations";
```
Amxx, line 381: I would do, so that the length change are tightly associated to the function that cause them

```suggestion
    assert lengthAfter != lengthBefore => (
        (f.selector == pushFront(bytes32).selector && lengthAfter == lengthBefore + 1) ||
        (f.selector == pushBack(bytes32).selector && lengthAfter == lengthBefore + 1) ||
        (f.selector == popBack().selector && lengthAfter == lengthBefore - 1) ||
        (f.selector == popFront().selector && lengthAfter == lengthBefore - 1) ||
        (f.selector == clear().selector && lengthAfter == 0)
    );    
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+rule noDataChange(env e) {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+
+    method f;
+    calldataarg args;
+
+    uint256 key;
+    bytes32 atBefore = at_(key);
+    f@withrevert(e, args);
```
Amxx, line 398: no need for `@withevert` here

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+*/
+rule noLengthChange(env e) {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+
+    method f;
+    calldataarg args;
+
+    uint256 lengthBefore = length();
+    f@withrevert(e, args);
```
Amxx, line 362: no need for `@withrevert` here

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+    calldataarg args;
+
+    uint256 key;
+    bytes32 atBefore = at_(key);
+    f@withrevert(e, args);
+    bytes32 atAfter = at_(key);
+
+    assert atAfter != atBefore => (
+        f.selector == clear().selector || // Although doesn't change values, outters are symbolic, so `* != *`
+        f.selector == popBack().selector ||
```
Amxx, line 403: WDYT? 

```suggestion
        (f.selector == popBack().selector && key == length()) ||
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+        f.selector == popFront().selector ||
+        f.selector == pushBack(bytes32).selector ||
```
Amxx, line 405: How can a `pushBack` change a `at_` ? I'm not expecting this here

Amxx (commented):
Great work! You are definitely on the right track.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Helpers                                                                                                             │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+
+function min_int128() returns mathint {
+    return (2 ^ 128) / 2 * -1;
```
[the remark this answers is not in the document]
Amxx: Also, is there any reason you use `^` and not `<<`? In some languages `^` is power, but in other its xor (and `**` is power). To avoid the confusion i's use shifts.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+}
+
+function max_int128() returns mathint {
+    return (2 ^ 128) / 2 - 1;
```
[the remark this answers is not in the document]
Amxx: same remark about `^` vs `<<`

certora/specs/DoubleEndedQueue.spec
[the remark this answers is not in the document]
ernestognw: Added the <<, I think it's now clearer. I decided not doing it at first just to play it safe

certora/specs/DoubleEndedQueue.spec
ernestognw: That's correct, thanks!

certora/specs/DoubleEndedQueue.spec
ernestognw: I agree, thanks!

certora/specs/DoubleEndedQueue.spec
ernestognw: No, good catch!

certora/specs/DoubleEndedQueue.spec
ernestognw: Make sense, thank you!

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,408 @@
...
+
+    // liveness
+    assert success, "never reverts";
+
+    // effect
+    assert success => frontAfter == value, "front set to value";
+
+    // no side effect
+    assert frontBefore == frontAfter <=> frontBefore == value, "front doesn't change only if it was already set";
+}
```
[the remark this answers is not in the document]
ernestognw: Thanks for the suggestion, also applied to `pushBack`

certora/specs/DoubleEndedQueue.spec
ernestognw: I think you meant `require key > 1;` instead of `require at > 1;` so I added the change with this modification to your suggestion. Let me know if it's not ok.

certora/specs/DoubleEndedQueue.spec
ernestognw: I think I lost this reference by updating so I'm not sure where it's the suggestion pointing to, but I'm assuming is for `popBackConsistency` and `pushFrontConsistency`, I think it's not a problem anymore after adding your suggestion (eg. attempting to check `at_@wtihRevert` after popping)

certora/specs/DoubleEndedQueue.spec
ernestognw: I see, I had the initial thought it was symbolic as in `clear()`, but it's not. Removing

certora/specs/DoubleEndedQueue.spec
ernestognw: Yes, I like it, although it's difficult to rationalize so I added a comment

ernestognw:
Comments addressed, but I need to make a last review myself to make sure there's no rule left out.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,363 @@
...
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Function correctness: clear sets length to 0                                                                        │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule clear {
+    clear@withrevert();
+    bool success = !lastReverted;
+    
+    // liveness
+    assert success, "never reverts";
```
Amxx, line 266:
```suggestion
    
    // liveness
    assert !lastReverted, "never reverts";
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,363 @@
...
+rule onlyOutOfBoundsRevert {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+
+    uint256 key;
+
+    at_@withrevert(key);
+    bool success = !lastReverted;
+
+    assert success <=> key < length(), "only reverts if key is out of bounds";
```
Amxx, line 312:
```suggestion

    assert lastReverted <=> key >= length(), "only reverts if key is out of bounds";
```

Amxx (commented):
This is really good !

ernestognw:
> This is really good !
> 
> Some very minor things.

Agree with all the comments. I batched them in the last commit with extra small changes, you can review them.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,363 @@
...
+rule pushFrontConsistency {
+    require boundedQueue();
+
+    uint256 key;
+    bytes32 beforeAt = at_(key);
+
+    bytes32 value;
+    pushFront(value);
+
+    bytes32 afterAt = at_(key + 1);
```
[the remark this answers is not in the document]
ernestognw: Definitely valuable, I'm wrapping my head around what it means to have (or not) the `@withrevert`, thanks!

Amxx (dismissed):
LGTM

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,361 @@
...
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: at(x) is preserved and offset to at(x - 1) after calling popFront                                             |
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popFrontConsistency(uint256 key) {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+
+    // Read (any) value that is not the front (this assert the value exist / the queue is long enough)
```
frangio, line 200:
```suggestion
    // Read (any) value that is not the front (this asserts the value exists / the queue is long enough)
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,361 @@
...
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: at(x) is preserved after calling popBack                                                                     |
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popBackConsistency(uint256 key) {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+
+    // Read (any) value that is not the front (this assert the value exist / the queue is long enough)
```
frangio, line 245:
```suggestion
    // Read (any) value that is not the back (this asserts the value exists / the queue is long enough)
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,361 @@
...
+    bool emptyBefore = empty();
+
+    f@withrevert(e, args);
+
+    assert lastReverted => (
+        (f.selector == front().selector && emptyBefore) ||
+        (f.selector == back().selector && emptyBefore) ||
+        (f.selector == popFront().selector  && emptyBefore) ||
+        (f.selector == popBack().selector  && emptyBefore) ||
+        f.selector == at_(uint256).selector
```
frangio, line 295:
```suggestion
        f.selector == at_(uint256).selector // revert conditions are verified in onlyOutOfBoundsRevert
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,361 @@
...
+    uint256 lengthBefore = length();
+    f(e, args);
+    uint256 lengthAfter = length();
+
+    assert lengthAfter != lengthBefore => (
+        (f.selector == pushFront(bytes32).selector && lengthAfter == lengthBefore + 1) ||
+        (f.selector == pushBack(bytes32).selector && lengthAfter == lengthBefore + 1) ||
+        (f.selector == popBack().selector && lengthAfter == lengthBefore - 1) ||
+        (f.selector == popFront().selector && lengthAfter == lengthBefore - 1) ||
+        (f.selector == clear().selector && lengthAfter == 0)
```
frangio, line 334: The length postconditions are already checked in each correctness rule.

```suggestion
        f.selector == pushFront(bytes32).selector ||
        f.selector == pushBack(bytes32).selector ||
        f.selector == popBack().selector ||
        f.selector == popFront().selector ||
        f.selector == clear().selector
```

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,361 @@
...
+    ), "length is only affected by clear/pop/push operations";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: only push/pop can change values bounded in the queue (outters aren't cleared)                                 │
```
frangio, line 340: "Outters" doesn't seem to be a word, do you mean "outer"? It also appears in a comment below.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,361 @@
...
+    bytes32 atBefore = at_(key);
+    f(e, args);
+    bytes32 atAfter = at_(key);
+
+    assert atAfter != atBefore => (
+        f.selector == clear().selector || // Although doesn't change values, outters are symbolic, so `* != *`
+        (f.selector == popBack().selector && key == length()) || // at_ only changes if it's left out, becoming symbolic 
+        f.selector == popFront().selector ||
+        f.selector == pushFront(bytes32).selector
+    ), "values of the queue are only changed by a clear, pop or push operation";
```
frangio, line 360: In some of these cases, like `clear`, the second `at_` shohuld revert. The comments about values being symbolic don't sound right to me, those cases would not be considered for verification.

I think this needs to be something like this:

```suggestion
    bytes32 atAfter = at_@withrevert(key);
    bool atAfterSuccess = !lastReverted;

    assert !atAfterSuccess <=> (
        f.selector == clear().selector ||
        (f.selector == popBack().selector && key == length())
    ), "indexes of the queue are only removed by clear or popBack";

    assert atAfterSuccess && atAfter != atBefore => (
        f.selector == popFront().selector ||
        f.selector == pushFront(bytes32).selector
    ), "values of the queue are only changed by popFront or pushFront";
```

frangio (commented):
This is a really nice spec! I only have an issue with the rule `noDataChange`.

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,361 @@
...
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: end is larger or equal than begin                                                                        │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant boundariesConsistency()
+    end() >= begin()
+    filtered { f -> !f.isView }
```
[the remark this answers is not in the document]
ernestognw: That's correct, if you remove it they will keep passing. It is a pattern I took from the ERC721 but I don't have a strong opinion for keeping it.

certora/specs/DoubleEndedQueue.spec
ernestognw: I mean everything that it's out of bounds (the values from outside - outters). Outers is the correct word?

certora/specs/DoubleEndedQueue.spec
ernestognw: That was sort of the previous layout but Hadrien suggested the current pattern, so that the length is associated with the function that causes it. I think it makes sense to have both although it's implicit in each function correctness check. What do you think?

certora/specs/DoubleEndedQueue.spec
ernestognw: The `at_` function is proven to fail iff a key is out of bounds by the `onlyOutOfBoundsRevert` rule and also each `(pop/push)(Front/Back)Consistency` rule makes sure `afterAt == beforeAt`.

In this case, not having the `@withrevert` was on purpose since it guarantees there would be at least one value left, but I think you're right that the case should be accounted.

Regarding your suggestion, I think the first assertion is not checking the same thing, it just checks the implication in the case failure, which is when the queue is left empty, and we want to check that for every `popBack()`

This is the rule originally (the `@withrevert` was incorrectly left out now that I think about it):

```
rule noDataChange(env e) {
    requireInvariant boundariesConsistency();
    require boundedQueue();

    method f;
    calldataarg args;

    uint256 key;
    bytes32 atBefore = at_(key);
    f(e, args);
    bytes32 atAfter = at_@withrevert(key);

    assert atAfter != atBefore => (
        f.selector == clear().selector || // Although doesn't change values, outters are symbolic, so `* != *`
        (f.selector == popBack().selector && key == length()) || // at_ only changes if it's left out, becoming symbolic 
        f.selector == popFront().selector ||
        f.selector == pushFront(bytes32).selector
    ), "values of the queue are only changed by a clear, pop or push operation";
}
```

Having a `@withrevert` there will indeed include the revert cases, but in those cases is when `atAfter` becomes symbolic. I'd rather only add `@withrevert` to the second `at_` as in the snippet above. What do you think?

certora/specs/DoubleEndedQueue.spec
frangio: I would prefer "outside values".

certora/specs/DoubleEndedQueue.spec
frangio: I don't know, I would avoid the redundancies, given that verification is slow and expensive.

certora/specs/DoubleEndedQueue.spec
frangio: > The `at_` function is proven to fail iff a key is out of bounds by the `onlyOutOfBoundsRevert` rule

I think this is different than that. `onlyOutOfBoundsRevert` is a property of a single `at_` call. Here what I was trying to express is that if an `at_` call succeeds, the only way that the same `at_` call can be made to fail is through `clear` or `popBack` with `key == length`. I think this is valuable information that we lose without `@withrevert`, but it seems we agree that we should add it.

I wasn't sure what the return value of a reverted call would be. Is it documented somewhere that it's a "symbolic" value? I'm not sure what that implies.

frangio: I guess your proposal also works, but my intuition is that it's weaker than what I proposed, which is also showing exactly when `at(key)` can go from returning to reverting.

certora/specs/DoubleEndedQueue.spec
Amxx: > I don't know, I would avoid the redundancies, given that verification is slow and expensive.

In this case, the check is done on values that are already computed/explored. So its doesn't really increase the complexity.

Equivalent in solidity would be doing a require on things that are already in the stack (cheap) vs doing a require on data that we need to read from storage and process.

Also, this entire spec file takes ~30sec. So its not really a concern here.

certora/specs/DoubleEndedQueue.spec
Amxx: > In some of these cases, like `clear`, the second `at_` should revert. The comments about values being symbolic don't sound right to me, those cases would not be considered for verification without `@withrevert`.
> 
> I think this needs to be something like this:

👍🏻 

This feels better (if it passes)

certora/specs/DoubleEndedQueue.spec
ernestognw: Done, but I want to mention I have the same feeling about the `iff` although I've used so I'm guilty too, he

certora/specs/DoubleEndedQueue.spec
ernestognw: Yes, there's no perceivable difference when testing one vs the other, but I can run a benchmark if you like.

Also, I think it's valuable as a programmer to look at the verification rules and get insights from them. Although these rules are redundant, it's easier to see the length changes instead of looking for answers on other rules.

certora/specs/DoubleEndedQueue.spec
ernestognw: > What I proposed would show exactly when at(key) can go from returning to reverting.

I see. Since that's the objective, then I added:

```
(f.selector == popFront().selector && key == length())
```

to the first rule, since it's not only popBack reverting.

Sent a commit with your suggestions.

frangio (dismissed):
LGTM

certora/specs/DoubleEndedQueue.spec
```diff
@@ -0,0 +1,366 @@
...
+│ Rule: at(x) is preserved and offset to at(x - 1) after calling popFront                                             |
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule popFrontConsistency(uint256 key) {
+    requireInvariant boundariesConsistency();
+    require boundedQueue();
+
+    // Read (any) value that is not the front (this asserts the value exists / the queue is long enough)
+    require key > 1;
+    bytes32 before = at_(key);
```
Amxx, line 202: I just realized that here (and almost everywhere else) you use `key` ... but this is more of an `index`. 

Looking at DoubleEndedQueue alone its not obvious, but when considering EnumerableSet and EnumerableMap ... I think we should be consistent and use `index`

ernestognw: Done :D

frangio (approved):

PR #4150, opened 2023-03-31
Refactor DoubleEndedQueue 

Amxx:
The use of signed integer is confusing, the cast decrease readability, and that is not actually needed. 
Using uint128 is cleaner IMO.

We can discuss narrowing down the scope of the `unchecked` block. Also, we could remove possibly the `Full()` check.

This not upgrade-compatible, so I'm targeting 5.0. Otherwise, the interface is backward compatible.

Note that there is currently an comment saying
 ```// We also assume there are at most int256.max items in the queue.```
 But since the queue is stored in a `mapping(int128 → uint256)` the size limit is 2**128, which is WAY smaller than int256.max

Fixes LIB-953

#### PR Checklist

- [x] Tests
- [ ] Documentation


frangio:
I don't really see a clear benefit in this. We can equally say that the use of unsinged integer wraparound is confusing.

Amxx:
The points was that it would avoid some (problematic?) casting in `length()` while also reducing the cost of `empty()`. 

The wraparound would not only be verified by FV, but the FV would also become cleaner because the "full" case is clearer. We would need to make fewer "value does not overflow" assumptions in the specs.

frangio:
> The points was that it would avoid some (problematic?) casting in `length()` while also reducing the cost of `empty()`.

These are very marginal benefits IMO.

That said, if this makes FV cleaner I would support it, but that also shows that we would need to update the spec.

I analyzed whether this could be a backwards compatible change, but it turns out that when computing the hash to locate a value in a mapping with integer keys, the key is _sign-extended_ rathert than _zero-padded_, so the upgrade from `mapping(int128 => bytes32`  to `mapping(uint128 => bytes32)` would be breaking.

Amxx:
> I analyzed whether this could be a backwards compatible change, but it turns out that when computing the hash to locate a value in a mapping with integer keys, the key is _sign-extended_ rathert than _zero-padded_, so the upgrade from `mapping(int128 => bytes32` to `mapping(uint128 => bytes32)` would be breaking.

Yes, that is why I wanted to do it in 5.0

Amxx:
Note: this change overall reduces the gas cost

Capture d’écran du 2023-06-29 16-35-17

ernestognw:
I can confirm this would make FV simpler and shouldn't be a big change there because the contract behaves the same. Also, with the overall gas reduction I think we have good reasons (still marginally good) to include this change in 5.0 with no caveat as far as I know.

frangio:
Let's move ahead with this then.

Amxx:
This is an internal change that doesn't affect the library interface, does it need a changelog?

I updated the Certora specs. I bumped it to Certora 4 (CVL2).

frangio:
> This is an internal change that doesn't affect the library interface, does it need a changelog?

Yes, according to the guidelines.

requirements.txt
```diff
@@ -1 +1 @@
-certora-cli==3.6.4
+certora-cli==4.3.1
```
frangio, line 1: If we merge this will all the existing specs stop working?

Amxx: They already don't work ... because 3.6.4 is no longer supported

Amxx: We'll have to do a "bump FV to new format" task during the quality control sprint. I can take care of it.

contracts/utils/structs/DoubleEndedQueue.sol
```diff
@@ -152,17 +155,16 @@ library DoubleEndedQueue {
      * @dev Returns the number of items in the queue.
      */
     function length(Bytes32Deque storage deque) internal view returns (uint256) {
-        // The interface preserves the invariant that begin <= end so we assume this will not overflow.
-        // We also assume there are at most int256.max items in the queue.
+        // We also assume there are at most uint128.max items in the queue.
```
frangio, line 158: Is this true? It previously said "we assume at most int256.max items", but now we are not really assuming anything, and as far as I can tell there could be as many as uint256.max items (theoretically)?

This comment can be removed I think.

frangio, line 158: I don't think this code is assuming anything so the comment can be removed.

frangio (commented):
This looks good to me. Sorry, minor comment though. I am ready to merge basically, beyond that comment.

Can this be moved out of draft?

contracts/utils/structs/DoubleEndedQueue.sol
```diff
@@ -40,18 +45,19 @@ library DoubleEndedQueue {
      * data[end - 1].
      */
     struct Bytes32Deque {
-        int128 _begin;
-        int128 _end;
-        mapping(int128 => bytes32) _data;
+        uint128 _begin;
+        uint128 _end;
+        mapping(uint128 => bytes32) _data;
```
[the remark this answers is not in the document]
Amxx: `begin` and `end` are the "borders" of the queue. That limits the length of the queue to `2**128-1` elements

We could use `uint256` instead of `uint128`, but that would result in `_begin` and `_end` being in two different slots ... which makes many operations more expensive.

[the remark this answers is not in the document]
Amxx: The only SafeCast i see is in `at`. We could keep the uint128 for `_begin` and `_end`, and use a uint256 key for `_data` but we will still need `at` to do the "uint128 wrapping". 

So if the key is uint256, then we would be doing 
```
deque._data[(uint256(deque._begin) + index) % (1<<128) ]
```
or
```
deque._data[(uint256(deque._begin) + index) & type(uint128).max ]
```

We can also make the argument that:
- `length` is always returning a uint128, by construction
- `if (index >= length(deque)) revert QueueOutOfBounds();` checks that index is smaller than a length, which is a uint128, so index is a uint128
- we can do a direct cast.

contracts/utils/structs/DoubleEndedQueue.sol
Amxx:
```suggestion
```

Amxx: done

contracts/utils/structs/DoubleEndedQueue.sol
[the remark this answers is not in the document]
frangio: That makes sense. I didn't realize before that the length was always uint128, since the code used to say "we assume there are at most int256.max items in the queue", but apparently it was already provably uint128 at most.

Let's do that and remove the safe cast.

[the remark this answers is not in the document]
Amxx: done

frangio (approved):
LGTM

ernestognw (approved):
I couldn't run the FV locally because of the syntax changes, but everything else LGTM.
I may be slow to review again if there are changes, you can merge without another approval from my side.

PR #4154, opened 2023-04-03
Fix TransparentUpgradeableProxy's transparency

Amxx:
Todo

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
+++ b/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
@@ -50,19 +50,23 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
      */
     function _fallback() internal virtual override {
         if (msg.sender == _getAdmin()) {
+            bytes memory ret;
             bytes4 selector = msg.sig;
             if (selector == ITransparentUpgradeableProxy.admin.selector) {
-                _internalAdmin();
+                ret = _dispatchAdmin();
             } else if (selector == ITransparentUpgradeableProxy.implementation.selector) {
-                _internalImplementation();
+                ret = _dispatchImplementation();
             } else if (selector == ITransparentUpgradeableProxy.changeAdmin.selector) {
-                _internalChangeAdmin();
+                ret = _dispatchChangeAdmin();
             } else if (selector == ITransparentUpgradeableProxy.upgradeTo.selector) {
-                _internalUpgradeTo();
+                ret = _dispatchUpgradeTo();
             } else if (selector == ITransparentUpgradeableProxy.upgradeToAndCall.selector) {
-                _internalUpgradeToAndCall();
+                ret = _dispatchUpgradeToAndCall();
             } else {
                 revert('TransparentUpgradeableProxy: admin cannot fallback to proxy target');
+            }
+            assembly {
+                return(add(ret, 0x20), mload(ret))
             }
         } else {
             super._fallback();
@@ -76,13 +80,11 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
      * https://eth.wiki/json-rpc/API#eth_getstorageat[`eth_getStorageAt`] RPC call.
      * `0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103`
      */
-    function _internalAdmin() private {
+    function _dispatchAdmin() private returns (bytes memory) {
         _requireZeroValue();
+
         address admin = _getAdmin();
-        assembly {
-            mstore(0x00, admin)
-            return(0, 0x20)
-        }
+        return abi.encode(admin);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -5,6 +5,16 @@ pragma solidity ^0.8.0;
 
 import "../ERC1967/ERC1967Proxy.sol";
 
+interface ITransparentUpgradeableProxy {
```
frangio, line 8: We should add some docs explaining why this is defined here.

I don't like that we have to redefine the events here. They're not strictly necessary either but I can imagine it was necessary for the test suite to run. What do you think about definining an `IERC1967` interface with the events, and using that interface here and also in `ERC1967Upgrade`?

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -36,75 +46,95 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
     }
 
     /**
-     * @dev Modifier used internally that will delegate the call to the implementation unless the sender is the admin.
+     * @dev If caller is the admin process the call internally, otherwise transparently fallback to the proxy behavior
      */
-    modifier ifAdmin() {
```
frangio, line 41: Are there strong reasons to remove the `ifAdmin` modifier? I believe it may be better to keep it with a deprecation notice.

I imagine an argument for removing it is that users who may be extending the contract and using `ifAdmin` should also consider removing it in favor of the manual dispatch approach we follow here. But it doesn't feel necessary to force that change. If the function using `ifAdmin` has no arguments there is no issue.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -114,14 +144,6 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
         return _getAdmin();
     }
 
-    /**
-     * @dev Makes sure the admin cannot access the fallback function. See {Proxy-_beforeFallback}.
-     */
-    function _beforeFallback() internal virtual override {
```
frangio, line 120: And what do you think about keeping this? It's not necessary to remove it, and by removing it we are changing the semantics of `_fallback`.

Amxx: Then it raises the question of what we do in `_fallback`.

- currently, in `_fallback` we have an
```
if (msg.sender == _getAdmin()) {
   ...
} else {
    super._fallback(); // here we should probably revert with "TransparentUpgradeableProxy: admin cannot fallback to proxy target"
}
```

- If we put the check in `_beforeFallback` and also keep it in `_fallback`, we are wasting an sload. That is 100gas on every call that goes through.
- If we put the check in `_beforeFallback` and remove it from `_fallback`, we are safe ... until someone overrides `_beforeFallback` without calling super and removes it. 

Note that the second option is already possible today. If someone overrides `_beforeFallback` and does not call super, the function selector that are not implemented in the proxy proxy become available to the admin. Is that something we want to keep ?

Amxx: That causes an inefficiency. 

For all "transparent" calls by users, here is what happens:
- `Proxy::fallback()` calls `_fallback()`
  - `TransparentUpgradeableProxy::_fallback()` (overidden) check the caller. It is not the admin, so call `super`
  - `Proxy::_fallback()` calls `_beforeFallback()`
    - `TransparentUpgradeableProxy::_beforeFallback()` (overidden) check the caller. If it is the admin we revert
    - `Proxy::_beforeFallback()` is empty
  - `Proxy::_fallback()` does the delegation

my issue is that `msg.sender != _getAdmin()` is checked twice: once in `TransparentUpgradeableProxy::_fallback()` and once in `TransparentUpgradeableProxy::_beforeCallback()`. This extra sload cost 100gas on ALL the user calls.

frangio (commented):
The fix looks good to me but all of the checks are failing.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -37,6 +62,9 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
 
     /**
      * @dev Modifier used internally that will delegate the call to the implementation unless the sender is the admin.
+     *
+     * CAUTION: This modifier is deprecated, as it could cause issues if the modified function has arguments, and the
+     * implementation provides a function with the same selector.
      */
     modifier ifAdmin() {
```
ernestognw, line 69: Can this just?

```solidity
...
{ else {
  super._fallback();
}
```

We know `msg.sender != _getAdmin()` so the `_fallback` will check that again

ernestognw (approved):
LGTM, just a minor comment

.changeset/thirty-shrimps-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`TransparentUpgradeableProxy`: fix transparency in case of selector clash with non-decodable calldata
```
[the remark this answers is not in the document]
ernestognw:
```suggestion
`TransparentUpgradeableProxy`: Fix transparency in case of selector clash with non-decodable calldata.
```

I'd stick to the PR title guidelines here and use `Fix` rather than `fixed`

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
frangio: Using `super._fallback` here has the usual problem we've discussed lately that it may skip an override.

I wouldn't worry about the inefficiency given that the modifier is being deprecated.

ernestognw: Gotcha! Thanks

contracts/interfaces/IERC1967.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @dev ERC-1967: Proxy Storage Slots. This interface contains the events defined in the ERC.
+ *
+ * _Available since v4.9._
```
pcaversaccio, line 8: This is not correct since it's released as part of `v4.8.3`. Will open a PR @Amxx.

pcaversaccio: FYI https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4183

PR #4158, opened 2023-04-04
Initializable should use high-slot storage fields

drortirosh:
Currently, Initializable uses a slot with 2 bytes. Since it is a base-contract to be extended, it most likely takes the 2st two bytes of slot-zero.

As a result, a contract that adds "Initializable" as a base class and has (say) an "address" field will have that field moved up few bytes, making its access slower (need to shift it down on each read/write) worse - if a contract attempted to "byte-align" several fields into a single slot, the Initializable adds 2 bytes into the first slot, which could easily break such alignment.

### Proposed fix

The Initializable should use storage at high offset, so that it won't interfere with any other contract

NOTE: We can't easily use directly the StorageSlot, since it can't use arbitrary struct (of two fields in the same storage slot), and splitting this two fields into separate slots is undesireable.
Instead, we use the same pattern.

### DIFF
--- a/contracts/proxy/utils/Initializable.sol
+++ b/contracts/proxy/utils/Initializable.sol
@@ -56,16 +56,29 @@ import "../../utils/Address.sol";
  * ====
  */
 abstract contract Initializable {
-    /**
-     * @dev Indicates that the contract has been initialized.
-     * @custom:oz-retyped-from bool
-     */
-    uint8 private _initialized;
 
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @drortirosh 

This way of separating the storage space of the different contracts is something we plan to use in v5 of contracts. This is unfortunately not something we can do today as it would be a breaking change.

PR #4160, opened 2023-04-04
Fix spurious CI check failures

frangio:
It's unclear why but we are seeing fuzz tests fail more often due to too many rejects.

Slither is also failing and it seems to be fixed by using Node 18.


ernestognw (approved):
I tried some options for checking if there's a dependency requiring a lower node version but seems there's no issue

frangio:
@ernestognw Yes I really don't know what causes the Slither failures but after multiple reruns it seems like it does fix it.

PR #4161, opened 2023-04-04
Reenable skipped TransparentUpgradeableProxy test

RenanSouza2:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4164

This is a fix for the 'uses the storage of the proxy' test in TransparentUpgradeableProxy.behaviour.js

The tests expects a different layout than described in the contracts


frangio (approved):
Thanks! Looks good to me.

PR #4163, opened 2023-04-05
Refactor PaymentSplitter require condition

launganik:
There is no need to check for payees and shares length mismatch if `payees.length` is 0


Amxx:
There is also no need to check that `payee.length > 0` if `payee.length` does not equal `shares.length`.

In any case we need to check both, and successful construction will have to pay for both. We should only reorder if we believe that one of the require is going to fail way more often than the other. Do we have a reason to believe that is the case?

ernestognw:
PaymentSplitter was removed in 5.0
We'll consider readding in the future, but for now, I think we can close this.

PR #4165, opened 2023-04-05
Merge changesets for transparency improvements

frangio:
Merging changesets from #3977 and #4154. The latter subsumes the former, which no longer applies in the way it was originally described (by reference to `ifAdmin`).

### DIFF
--- a/.changeset/many-panthers-hide.md
+++ b/.changeset/many-panthers-hide.md
@@ -1,5 +1,0 @@
----
-'openzeppelin-solidity': minor
----
-
-`TransparentUpgradeableProxy`: support value passthrough for all ifAdmin function.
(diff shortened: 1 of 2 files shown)


ernestognw (approved):
I don't think we should do this frequently since it may break the Changesets release automation (not badly). The Changelog entries are associated with the PR that merged the change, so the Changesets PR will link this PR instead of the old one.

This is why we kept the PR links in the Changesets when we migrated and will require manual intervention on the Changesets PR.

LGTM still, this is something to care about when using the automation.

frangio:
Ah interesting. I can tolerate a wrong PR link every once in a while, but we should definitely make it an infrequent occurrence.

Amxx (approved):

PR #4167, opened 2023-04-06
Add formal verification specs for EnumerableSet & EnumerableMap

Amxx:
Fixes LIB-774
replaces #4166

- [x] EnumerableSet
- [x] EnumerableMap


certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+            requireInvariant atUniqueness(index2, to_uint256(length() - 1));
+        }
+    }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: index <> value relationship is consistent                                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant consistencyIndex(uint256 index)
```
ernestognw, line 57: We discussed this on slack, but these two invariants (`consistencyIndex` and `consistencyKey`) are _arguably_ equivalent. The rationale for leaving the two is that sometimes we want to check `key → index → key` and others `index → key → index`.

Mentioning for the record

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+    {
+        preserved remove(bytes32 key) {
+            requireInvariant atUniqueness(index1, to_uint256(length() - 1));
+            requireInvariant atUniqueness(index2, to_uint256(length() - 1));
+        }
+    }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: index <> value relationship is consistent                                                                │
```
ernestognw, line 54:
```suggestion
│ Invariant: index <> key relationship is consistent                                                                │
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+*/
+invariant consistencyIndex(uint256 index)
+    _indexOf(at_(index)) == index + 1
+    {
+        preserved remove(bytes32 key) {
+            requireInvariant consistencyIndex(to_uint256(length() - 1));
+        }
+    }
+
+invariant consistencyKey(bytes32 key)
```
ernestognw, line 65: The suggestion with the complete header doesn't fit here.

```suggestion
/*
...
| Invariant: index <> key relationship is consistent        |
...
*/
invariant consistencyKey(bytes32 key)
```

On a side comment, I find the convention for commenting on top of rules a bit problematic, any reason to use it?

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+            requireInvariant atUniqueness(
+                to_uint256(_indexOf(key) - 1),
+                to_uint256(_indexOf(otherKey) - 1)
+            );
+        }
+    }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rules: state changes                                                                                                │
```
ernestognw, line 79:
```suggestion
│ Rule: state only changes by adding or removing elements                                                                                             │
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+    bool    containsBefore      = contains(key);
+    bool    containsOtherBefore = contains(otherKey);
+
+    bool added = add@withrevert(key);
+    bool success = !lastReverted;
+
+    // liveness & immediate effect
+    assert success && contains(key);
+
+    // return value: added iff not contained
```
ernestognw, line 134:
```suggestion
    // return value: added if not contained
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+    assert added <=> !containsBefore;
+
+    // effect: length increass iff added
```
ernestognw, line 137:
```suggestion
    // effect: length increases if added
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+    assert length() == lengthBefore + to_uint256(added ? 1 : 0);
```
ernestognw, line 138: Official [Certora's recommendation](https://docs.certora.com/en/latest/docs/cvl/mathops.html?highlight=casting#mathematical-operations) is to _use `mathint` where possible_, so:

```suggestion
    assert length() == lengthBefore + to_mathint(added ? 1 : 0);
```

Also, I think there's an implicit casting by just:

```suggestion
    assert length() == lengthBefore + added ? 1 : 0;
```

But I'm not sure if it works given the ternary

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+
+    // effect: add at the end
+    assert added => at_(lengthBefore) == key;
+
+    // side effect: other keys are not affected
+    assert containsOtherBefore != contains(otherKey) => (added && key == otherKey);
```
ernestognw, line 144: Question, why not adding the comment as?:

```suggestion
    assert containsOtherBefore != contains(otherKey) => (added && key == otherKey), "side effect: other keys are not affected";
```

I've found this pattern helpful for me since I feel the strings are more tied to the rule than as in a comment, but just personal preference.

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+    bool    containsBefore      = contains(key);
+    bool    containsOtherBefore = contains(otherKey);
+
+    bool removed = remove@withrevert(key);
+    bool success = !lastReverted;
+
+    // liveness & immediate effect
+    assert success && !contains(key);
+
+    // return value: removed iff contained
```
ernestognw, line 166:
```suggestion
    // return value: removed if contained
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+    assert removed <=> containsBefore;
+
+    // effect: length increass iff removed
```
ernestognw, line 169:
```suggestion
    // effect: length increases if removed
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,223 @@
...
+    assert length() == lengthBefore - to_uint256(removed ? 1 : 0);
```
ernestognw, line 170: Same remark as the ternary

ernestognw (commented):
Finishing with EnumerableSet, will add more comments to EnumerableMap

certora/specs/EnumerableMap.spec
```diff
@@ -0,0 +1,308 @@
...
+    bytes32 atKeyBefore = at_key(index);
+    bytes32 atValueBefore = at_value(index);
+
+    set(key, value);
+
+    bytes32 atKeyAfter = at_key@withrevert(index);
+    assert !lastReverted;
+
+    bytes32 atValueAfter = at_value@withrevert(index);
+    assert !lastReverted;
```
ernestognw, line 257: I think we can remove the `@withrevert` here because it's proven that they only revert if there's no value

```suggestion
    bytes32 atKeyAfter = at_key(index);
    bytes32 atValueAfter = at_value(index);
```

ernestognw (commented):
Pretty much just typos and few suggestions, looking good!

certora/specs/EnumerableSet.spec
Amxx: Note:

They are equivalent if you assume the function `at` and `indexOf` are bijections between the `keys` and the `indices`.

Having both properties proven is a proof of the bijections.

I would argue that we should have both, and we should document that having both proves the bijection.

certora/specs/EnumerableSet.spec
Amxx: [iff is a way to say "if and only if"](https://simple.wikipedia.org/wiki/If_and_only_if)

Not a typo.

certora/specs/EnumerableSet.spec
Amxx: It work without `to_mathint` or `to_uint256` ... but we still need to put `()`

certora/specs/EnumerableMap.spec
Amxx: Its only proven that it revert if the index is out of bound. And true, since we are doing set, the length increasses and so the function should not revert.

Strictly speaking this is already proven somewhere else. I still think its nice to include it here, just so that this rule fully verify that if a value was accessible before, it remains accessible at the same place. (without falling back to another rule)

certora/specs/EnumerableMap.spec
```diff
@@ -0,0 +1,308 @@
...
+
+    bool added = set@withrevert(key, value);
+    bool success = !lastReverted;
+
+    // liveness & immediate effect
+    assert success
+        && contains(key)
+        && get(key) == value;
+
+    // return value: added iff not contained
```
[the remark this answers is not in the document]
frangio: `iff` is intentional, it means "if and only if".

certora/specs/EnumerableMap.spec
```diff
@@ -0,0 +1,312 @@
...
+invariant consistencyIndex(uint256 index)
+    _indexOf(at_key(index)) == index + 1
+    {
+        preserved remove(bytes32 key) {
+            requireInvariant consistencyIndex(to_uint256(length() - 1));
+        }
+    }
+
+invariant consistencyKey(bytes32 key)
+    at_key(to_uint256(_indexOf(key) - 1)) == key
```
frangio, line 87: Here I would be more concerned about `_indexOf(key)` being 0. Shouldn't we be writing the following?

```suggestion
    _indexOf(key) > 0 => at_key(to_uint256(_indexOf(key) - 1)) == key
```

Otherwise it seems writing `requireInvariant consistencyKey(key)` may in some cases be requiring a condition that shouldn't be assumed.

frangio (commented):
Just leaving some initial comments.

certora/specs/EnumerableMap.spec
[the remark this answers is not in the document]
ernestognw: Interesting, isn't the fact I misunderstood it a problem by itself?

certora/specs/EnumerableSet.spec
ernestognw: Just realized, I didn't know.
I still think the misunderstanding shouldn't be infrequent, or maybe I'm just new to the terminology.

certora/specs/EnumerableMap.spec
```diff
@@ -0,0 +1,312 @@
...
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: All indexed keys are contained                                                                           │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant indexedContained(uint256 index)
+    contains(at_key(index))
+    {
+        preserved {
+            requireInvariant consistencyIndex(index);
+            requireInvariant consistencyIndex(to_uint256(length() - 1));
```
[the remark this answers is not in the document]
Amxx: That is an interesting point.

If the length is 0, `at_key` (now renamed to `key_at`) reverts, so basically we have no precondition, so nothing to check. Maybe that whole invariant is pointless ...

certora/specs/EnumerableMap.spec
Amxx: If `_indexOf(key)` is 0, `to_uint256(_indexOf(key) - 1)` should revert, and nothing is checked.

Indeed `requireInvariant consistencyKey(key)` requires that the key exist. This is clear to me, but might not be to everyone.

Amxx: The same applies to `invariant consistencyIndex(uint256 index)`. If the index is not in scope, `at_key(index)` reverts

frangio: > `to_uint256(_indexOf(key) - 1)` should revert

What is the part that should revert exactly? I expect `0 - 1` to evaluate to `-1`, and `to_uint256(-1)` evaluates to `max_uint256`.

certora/specs/EnumerableSet.spec
frangio: > I find the convention for commenting on top of rules a bit problematic

@ernestognw What do you mean?

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,231 @@
...
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: index <> key relationship is consistent                                                                  │
+│                                                                                                                     │
+│ Note that the two consistencyXxx invariants, put together, prove that at_ and _indexOf are inverse of one another.  │
+│ This proves that we have a bijection between indices (the enumerability part) and keys (the entries that are added  │
+│ and removed from the EnumerableSet).                                                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant consistencyIndex(uint256 index)
+    index < length() => _indexOf(at_(index)) == index + 1
```
frangio, line 62: As mentioned in the other spec.
```suggestion
    index < length() => _indexOf(at_@withrevert(index)) == index + 1
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,231 @@
...
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule stateChange(env e, bytes32 key) {
+    require sanity();
+    requireInvariant consistencyKey(key);
+
+    uint256 lengthBefore   = length();
+    bool    containsBefore = contains(key);
+
+    method f; calldataarg args; f(e, args);
```
frangio, line 97:
```suggestion
    method f;
    calldataarg args;
    f(e, args);
```

certora/specs/EnumerableSet.spec
```diff
@@ -0,0 +1,231 @@
...
+    assert success && contains(key),
+        "liveness & immediate effect";
+
+    assert added <=> !containsBefore,
+        "return value: added iff not contained";
+
+    assert length() == lengthBefore + to_mathint(added ? 1 : 0),
+        "effect: length increases iff added";
+
+    assert added => at_(lengthBefore) == key,
```
frangio, line 148: What are your thoughts on this suggestion? It feels unnecessary, but why?

```suggestion
    assert added => at_@withrevert(lengthBefore) == key,
```

certora/specs/EnumerableSet.spec
ernestognw: /*
┌─
|
└─
*/

This block. I see it problematic

certora/harnesses/EnumerableMapHarness.sol
```diff
@@ -0,0 +1,59 @@
...
+        return value;
+    }
+
+    function get(bytes32 key) public view returns (bytes32) {
+        return _map.get(key);
+    }
+
+    // function keys() public view returns (bytes32[] memory) {
+    //     return _map.keys();
+    // }
```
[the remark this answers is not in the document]
Amxx: Not really. I wanted to check that in FV, but function that return array are not supported. So I commented it, but kept it because at some point we may be able to cover that ?

We could remove it

certora/specs/EnumerableMap.spec
Amxx: I guess `at_key(max_uint256)` is the part that reverts. Will change the way to propose it

certora/specs/EnumerableMap.spec
```diff
@@ -0,0 +1,315 @@
...
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: index <> value relationship is consistent                                                                │
+│                                                                                                                     │
+│ Note that the two consistencyXxx invariants, put together, prove that at_ and _indexOf are inverse of one another.  │
+│ This proves that we have a bijection between indices (the enumerability part) and keys (the entries that are set    │
+│ and removed from the EnumerableMap).                                                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant consistencyIndex(uint256 index)
+    index < length() => _indexOf(key_at(index)) == index + 1
```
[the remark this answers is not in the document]
Amxx: If We make a liveness rule for key_at / length / others, than would you be ok not making that change ?

Amxx:
I modified the "outOfBound" rules into "liveness" rules that check the domain of all view function.

certora/specs/EnumerableSet.spec
Amxx: In `added` case, we checked just above that `length() == lengthBefore + 1` so we know that `at_(lengthBefore)` is in the bounds defined by `liveness_2`

frangio:
I've removed the commented lines. If we keep commented lines I think we should have a comment explaining the reason.

frangio (approved):

PR #4168, opened 2023-04-06
Alternative fix  for the TransparentUpgradeableProxy "decoding" bug

Amxx:
Partly revert and replace #4154.

This alternative fix is motivated by the goal to make more localized changes, since #4154 is a larger refactor.

Since the first commit in this PR is a revert of 5523c14 (#4154), it's better to see the [diff against its parent](https://github.com/OpenZeppelin/openzeppelin-contracts/compare/5523c14~1...Amxx:openzeppelin-contracts:fix/proxy/TransparentUpgradeableProxy).

### DIFF
--- a/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
+++ b/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
@@ -49,7 +49,7 @@ interface ITransparentUpgradeableProxy is IERC1967 {
  * NOTE: The real interface of this proxy is that defined in `ITransparentUpgradeableProxy`. This contract does not
  * inherit from that interface, and instead you will see some admin functions are implemented by oddly named functions.
  * These functions are designed to match the selectors of `ITransparentUpgradeableProxy`, but without any arguments.
- * This is necessary to fully implement transparency without decoding errors caused by selector clashes between the
+ * This is necessary to fully implement transparency without decoding reverts caused by selector clashes between the
  * proxy and the implementation.
  */
 contract TransparentUpgradeableProxy is ERC1967Proxy {
--- a/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
+++ b/contracts/proxy/transparent/TransparentUpgradeableProxy.sol
@@ -6,24 +6,6 @@ pragma solidity ^0.8.0;
 import "../ERC1967/ERC1967Proxy.sol";
 
 /**
- * @dev Interface for the {TransparentUpgradeableProxy}. This is useful because {TransparentUpgradeableProxy} uses a
- * custom call-routing mechanism, the compiler is unaware of the functions being exposed, and cannot list them. Also
- * {TransparentUpgradeableProxy} does not inherit from this interface because it's implemented in a way that the
- * compiler doesn't understand and cannot verify.
- */
-interface ITransparentUpgradeableProxy is IERC1967 {
-    function admin() external view returns (address);
-
-    function implementation() external view returns (address);
-
-    function changeAdmin(address) external;
-
-    function upgradeTo(address) external;
-
-    function upgradeToAndCall(address, bytes memory) external payable;
-}
-
-/**
  * @dev This contract implements a proxy that is upgradeable by an admin.
  *
  * To avoid https://medium.com/nomic-labs-blog/malicious-backdoors-in-ethereum-proxies-62629adf3357[proxy selector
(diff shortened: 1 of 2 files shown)


contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -44,12 +46,11 @@ interface ITransparentUpgradeableProxy is IERC1967 {
...
- * implemented using a custom call-routing mechanism in `_fallback`. Consequently, the compiler will not produce an
- * ABI for this contract. Also, if you inherit from this contract and add additional functions, the compiler will not
- * check that there are no selector conflicts. A selector clash between any new function and the functions declared in
- * {ITransparentUpgradeableProxy} will be resolved in favor of the new one. This could render the admin operations
- * inaccessible, which could prevent upgradeability.
+ * WARNING: This contract does not inherit from {ITransparentUpgradeableProxy}, and the admin functions are
+ * implemented by functions that have signature designed to match the selector of the {ITransparentUpgradeableProxy}
+ * interface, but without any arguments. This allows use to decode the argument manually after the `ifAdmin` modifier
+ * has had a change to forward the call. This is done so that the argument decoding does not fail here in case the
+ * proxy and the implementation have "selector clash".
```
ernestognw:
```suggestion
 * WARNING: This contract does not inherit from {ITransparentUpgradeableProxy} because its functions are implemented
 * in this contract with alternative signatures designed to match the selectors of the {ITransparentUpgradeableProxy}
 * interface, but without any arguments. This allows manual argument decoding after the `ifAdmin` modifier
 * has had a chance to forward the call, so that the argument decoding does not fail here in case the
 * proxy and the implementation have clashing selectors.
```

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -175,6 +154,14 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
         return _getAdmin();
     }
 
+    /**
+     * @dev Makes sure the admin cannot access the fallback function. See {Proxy-_beforeFallback}.
+     */
+    function _beforeFallback() internal virtual override {
+        require(msg.sender != _getAdmin(), "TransparentUpgradeableProxy: admin cannot fallback to proxy target");
+        super._beforeFallback();
+    }
```
ernestognw, line 163: Any particular concern for adding this?

ernestognw (commented):
I reviewed this alternative with Seppi and his main concern was related to extensibility. I don't think we should expect issues with overriding (agree they shouldn't be frequent and not recommended), but any modification to this contract will face the issue of needing to mine a clashing selector.

I share the same concern for legitimate possible use cases.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
ernestognw: Ah, you replaced it first @frangio.
I'm solving but leaving it in case you think it's useful

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
frangio: Note that this was here originally, it's being "added" only because #4154 is being reverted.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
frangio: Yeah I thought this text was describing implementation detail too much. Please look at the new text.

frangio:
> any modification to this contract will face the issue of needing to mine a clashing selector.

In my view, this is not something that is introduced by this PR, it's just implied by the transparency requirement.

The alternative is https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4154, which makes extending the proxy a little more cumbersome.

frangio:
I've updated the docs a little.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
ernestognw: Looks good, only comment might be:

```
 * This is necessary to fully implement transparency without decoding *reverts* caused by selector clashes between the
```

"errors" sound very vague to me when the issue is an actual decoding revert, what do you think?

frangio:
We've decided to keep the fix from https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4154.

Moved the docs improvements to its own PR: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4181

PR #4171, opened 2023-04-09
Next v5.0 erc721 update hook

RenanSouza2:
This is proposal to remove before and after hooks from ERC721 and replace them with an update hook


frangio:
Thank you @RenanSouza2, we will review this shortly.

Amxx:
FYI, we already have a branch (that has been sitting for the last 3 months) with the `_update` refactor for ERC721.

here is what I found (https://github.com/Amxx/openzeppelin-contracts/blob/refactor/erc721-mint-burn-transfer/contracts/token/ERC721/ERC721.sol) ... there was something on julissa's fork that I pushed to mine.

AFAIK, we were worried about the duplicate sload when doing a transferFrom:
- in `approvedOrOwner` we need to get the current owner to check authorisation
- in `_update` we need to get the the current owner to check that `from` is correct.
I proposed a solution that uses fnPointers to put the "constraint" check inside `_update`.

It was supposed to get a review ... but I guess we forgot about it ?

RenanSouza2:
I've expanded the code in the PR in this other branch 
https://github.com/RenanSouza2/openzeppelin-contracts/blob/update-ERC721Votes/contracts/token/ERC721/extensions/ERC721Votes.sol

There is a function _increaseBalance that is also used as a hook. This function is called in the _mintBatch function in Consecutive and is overriden in Enumerable to avoit batch sizes bigger than one.

In this branch specifically it is used to in the Votes to distribute voting power

PR #4175, opened 2023-04-10
Add fuzz tests for ShortString

frangio:


test/utils/ShortStrings.t.sol
```diff
@@ -0,0 +1,55 @@
...
+        _fallback = fallbackInitial;
+        ShortString short = ShortStrings.toShortStringWithFallback(input, _fallback);
+        string memory output = ShortStrings.toStringWithFallback(short, _fallback);
+        assertEq(input, output);
+    }
+
+    function testRevertLong(string memory input) external {
+        vm.assume(!_isShort(input));
+        vm.expectRevert(abi.encodeWithSelector(ShortStrings.StringTooLong.selector, input));
+        this.toShortString(input);
```
ernestognw, line 29: Any reason to put this function aside?

ernestognw (approved):
LGTM but one comment:

- I think it's not obvious why you're setting `_fallback` since it's not used, but I'm assuming is for fuzzing `StorageSlot.getStringSlot(store)`. I'd add a comment

test/utils/ShortStrings.t.sol
ernestognw: Oh, I see

From https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4146#issuecomment-1502313546
> It's important to remember that this cheatcode only works with external calls

ernestognw:
On a side comment, I like camel case on the fuzz tests, but we already have snake case. We must decide what style to go with before we get more mixing.

frangio:
I went with snake case initially but the linter wants camel case. What would you suggest?

---

> I think it's not obvious why you're setting _fallback since it's not used, but I'm assuming is for fuzzing StorageSlot.getStringSlot(store). I'd add a comment

Do you mean why I defined a `fallbackInitial` parameter? That value should have no effect at all, but treating the implementation as a black box that's not something that we woul dknow.

ernestognw:
> I went with snake case initially but the linter wants camel case. What would you suggest?

camelCase definitely, but just wanted to make consensus so we can flag it in other test

> Do you mean why I defined a `fallbackInitial` parameter? That value should have no effect at all, but treating the implementation as a black box that's not something that we woul dknow.

Yeah, that was my thought, with the comment you added should be enough.

ernestognw (approved):

PR #4177, opened 2023-04-11
Fix AccessControlDefaultAdminRules admin consistency

ernestognw:
The current `AccessControlDefaultAdminRules` implementation inherits from the `AccessControl` behavior, which allows any account to renounce any role even if it hasn't been granted. This is consistent with our policy of allowing NOOPs. 

However, when a user renounces the `DEFAULT_ADMIN_ROLE` without holding it, the action will reset the `defaultAdmin()` and `owner()` variables, breaking the invariant that assumes that if an account `hasRole(...) == true`, then that same account should be both `owner()` and `defaultAdmin()`

This PR ensures consistency for the stated invariant.

### DIFF
--- a/contracts/access/AccessControlDefaultAdminRules.sol
+++ b/contracts/access/AccessControlDefaultAdminRules.sol
@@ -136,7 +136,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @dev See {AccessControl-_revokeRole}.
      */
     function _revokeRole(bytes32 role, address account) internal virtual override {
-        if (role == DEFAULT_ADMIN_ROLE) {
+        if (role == DEFAULT_ADMIN_ROLE && account == _currentDefaultAdmin) {
             delete _currentDefaultAdmin;
         }
         super._revokeRole(role, account);
--- a/test/access/AccessControl.behavior.js
+++ b/test/access/AccessControl.behavior.js
@@ -643,12 +643,22 @@ function shouldBehaveLikeAccessControlDefaultAdminRules(errorPrefix, delay, defa
       );
     });
 
+    it.only('keeps defaultAdmin consistent with hasRole if another non-defaultAdmin user renounces the DEFAULT_ADMIN_ROLE', async function () {
+      await time.setNextBlockTimestamp(delayPassed);
+
+      // This passes because it's a noop
+      await this.accessControl.renounceRole(DEFAULT_ADMIN_ROLE, other, { from: other });
+
+      expect(await this.accessControl.hasRole(DEFAULT_ADMIN_ROLE, defaultAdmin)).to.be.true;
+      expect(await this.accessControl.defaultAdmin()).to.be.equal(defaultAdmin);
+    });
+
     it('renounces role', async function () {
       await time.setNextBlockTimestamp(delayPassed);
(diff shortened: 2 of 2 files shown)


frangio (dismissed):
LGTM

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -136,7 +136,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @dev See {AccessControl-_revokeRole}.
      */
     function _revokeRole(bytes32 role, address account) internal virtual override {
-        if (role == DEFAULT_ADMIN_ROLE) {
+        if (role == DEFAULT_ADMIN_ROLE && account == _currentDefaultAdmin) {
```
Amxx, line 139: Why not do 
```
bool revoked = super._revokeRole(role, account);
if (role == DEFAULT_ADMIN_ROLE && revoked) {
  delete _currentDefaultAdmin;
}
```
?

(feels better "integrated" to me)

frangio: Both options work for me.

ernestognw: I honestly don't see how this might work. `super._revokeRole()` doesn't return anything.

Amxx: > I honestly don't see how this might work. `super._revokeRole()` doesn't return anything.

Indeed ... I got confused with `EnumerableSet._add` and `EnumerableSet._remove` that return bool.
In 5.0 maybe we want this functions to return a bool

Amxx (commented):
LGTM, except for some details in the tests:

there are many `renounceRole(DEFAULT_ADMIN_ROLE, defaultAdmin, { from })`, which confused me because `let from = defaultAdmin;` is declared higher. Should we be more explicit using 

```
renounceRole(DEFAULT_ADMIN_ROLE, defaultAdmin, { from: defaultAdmin })
```

test/access/AccessControl.behavior.js
```diff
@@ -643,12 +643,22 @@ function shouldBehaveLikeAccessControlDefaultAdminRules(errorPrefix, delay, defa
...
+    });
+
     it('renounces role', async function () {
       await time.setNextBlockTimestamp(delayPassed);
-      const receipt = await this.accessControl.renounceRole(DEFAULT_ADMIN_ROLE, from, { from });
+      const receipt = await this.accessControl.renounceRole(DEFAULT_ADMIN_ROLE, defaultAdmin, { from });
 
       expect(await this.accessControl.hasRole(DEFAULT_ADMIN_ROLE, defaultAdmin)).to.be.false;
-      expect(await this.accessControl.hasRole(constants.ZERO_ADDRESS, defaultAdmin)).to.be.false;
+      expect(await this.accessControl.hasRole(DEFAULT_ADMIN_ROLE, defaultAdmin)).to.be.false;
```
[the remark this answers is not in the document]
ernestognw: A mistake, I think I was originally intending to:

```js
expect(await this.accessControl.defaultAdmin()).to.be.equal(constants.ZERO_ADDRESS);
```

I added it but it was also checked by the `owner()`

frangio (approved):

PR #4178, opened 2023-04-11
Implement suggestions from audit of AccessManager

ernestognw:
Audit report to be published soon.

### DIFF
--- a/contracts/access/manager/AccessManaged.sol
+++ b/contracts/access/manager/AccessManaged.sol
@@ -13,7 +13,7 @@ import "./IAuthority.sol";
  * IMPORTANT: The `restricted` modifier should never be used on `internal` functions, judiciously used in `public`
  * functions, and ideally only used in `external` functions. See {restricted}.
  */
-contract AccessManageable is Context {
+contract AccessManaged is Context {
     event AuthorityUpdated(address indexed sender, IAuthority indexed newAuthority);
 
     IAuthority private _authority;
@@ -67,7 +67,7 @@ contract AccessManageable is Context {
      * @dev Transfers control to a new authority. The caller must be the current authority.
      */
     function setAuthority(IAuthority newAuthority) public virtual {
-        require(_msgSender() == address(_authority), "AccessManageable: not current authority");
+        require(_msgSender() == address(_authority), "AccessManaged: not current authority");
         _setAuthority(newAuthority);
     }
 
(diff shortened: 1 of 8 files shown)


contracts/access/manager/AccessManager.sol
```diff
@@ -18,7 +17,7 @@ interface IAccessManager is IAuthority, IAccessControlDefaultAdminRules {
 
     event GroupAllowed(address indexed target, bytes4 indexed selector, uint8 indexed group, bool allowed);
 
-    event AccessModeUpdated(address indexed target, AccessMode indexed mode);
+    event AccessModeUpdated(address indexed target, AccessMode indexed previousMode, AccessMode indexed mode);
```
frangio, line 20: I don't think the previous one needs to be indexed. If you're looking up by mode you probably want to find the event that changed into a specific mode so you would look up by `mode`?

contracts/access/manager/AccessManaged.sol
```diff
@@ -30,6 +30,19 @@ contract AccessManaged is Context {
...
      * function at the bottom of the call stack, and not the function where the modifier is visible in the source code.
      * ====
+     *
+     * [NOTE]
+     * ====
+     * Although a selector collision is prevented by scoping permissions per contract, consider a selector with access granted
+     * to a group may clash with unintended functions in the following cases:
+     *
+     * * If the https://docs.soliditylang.org/en/latest/contracts.html#receive-ether-function[`receive()`] function is restricted,
+     * any other function with a `bytes4(0)` selector will share permissions with `receive()`.
```
frangio, line 40:
```suggestion
     * any other function with a `0x00000000` selector will share permissions with `receive()`.
```

contracts/access/manager/AccessManaged.sol
```diff
@@ -30,6 +30,19 @@ contract AccessManaged is Context {
      * implications! This is because the permissions are determined by the function that entered the contract, i.e. the
      * function at the bottom of the call stack, and not the function where the modifier is visible in the source code.
      * ====
+     *
+     * [NOTE]
+     * ====
+     * Although a selector collision is prevented by scoping permissions per contract, consider a selector with access granted
+     * to a group may clash with unintended functions in the following cases:
```
frangio, line 37:
```suggestion
     * Selector collisions are mitigated by scoping permissions per contract, but some edge cases must be considered:
```

PR #4179, opened 2023-04-12
Make sendValue suggest a fix on revert

martinvol:
I spent a couple of hours figuring out why my contract didn't work, this error message would have save me quite some time.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

I just changed a string

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -65,7 +65,7 @@ library Address {
         require(address(this).balance >= amount, "Address: insufficient balance");
 
         (bool success, ) = recipient.call{value: amount}("");
-        require(success, "Address: unable to send value, recipient may have reverted");
+        require(success, "Address: unable to send value, recipient may have reverted. Maybe you forgot to implement receive()?");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ernestognw:
Hey @martinvol, thanks for your contribution. However, I'm afraid this increases the bytecode size since the revert strings are embedded into it, increasing deployment and revert execution gas costs.

In this case, this is the effect in size increase:

```diff
- 0x416464726573733a20756e61626c6520746f2073656e642076616c75652c20726563697069656e74206d61792068617665207265766572746564000000000000
+ 0x416464726573733a20756e61626c6520746f2073656e642076616c75652c20726563697069656e74206d617920686176652072657665727465642e204d6179626520796f7520666f72676f7420746f20696d706c656d656e74207265636569766528293f00000000000000000000000000000000000000000000000000000000
```

Generally, our priority is short revert strings so they occupy as least space as possible.
On the other hand, we're planning to support custom errors in the next 5.0 release later this year, here's the discussion #2839

ernestognw:
Hey @martinvol, we recently merged #4261 with a [bit more of an explanation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b425a722409fdf4f3e0ec8f42d686f3c0522af19/contracts/utils/Address.sol#L21) for this error.

I'm closing this since the new state doesn't depend on revert strings.

PR #4180, opened 2023-04-12
Add AccessControlDefaultAdminRules FV

ernestognw:
Fixes LIB-756


ernestognw:
I want to adjust some details and review grammar, but it's ready for review since what I'll change later shouldn't change the rules

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,379 @@
...
+//   role -> role != DEFAULT_ADMIN_ROLE()
+// }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Helpers                                                                                                             │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+
+function max_uint48() returns mathint {
```
frangio, line 24: Is there a difference between `function` and `definition`?

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,379 @@
...
+    return (1 << 48) - 1;
+}
+
+// defaultAdmin() returns the zero address when there's no default admin
```
frangio, line 28: This comment feels out of place here. I'd  move it to `defaultAdminConsistency`.

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,379 @@
...
+  hasRole(DEFAULT_ADMIN_ROLE(), account) && hasRole(DEFAULT_ADMIN_ROLE(), another) => another == account
+  filtered { f -> f.selector != acceptDefaultAdminTransfer().selector }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: DEFAULT_ADMIN_ROLE's admin is always the zero address                                                    │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant defaultAdminRoleAdminConsistency()
+  getRoleAdmin(DEFAULT_ADMIN_ROLE()) == 0
```
frangio, line 79: `getRoleAdmin` returns a role id, not an address. So this should be more like this:

```suggestion
│ Invariant: DEFAULT_ADMIN_ROLE's admin is always DEFAULT_ADMIN_ROLE                                                    │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
*/
invariant defaultAdminRoleAdminConsistency()
  getRoleAdmin(DEFAULT_ADMIN_ROLE()) == DEFAULT_ADMIN_ROLE()
```

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,379 @@
...
+│ Rule: defaultAdmin is only affected by accepting an admin transfer or renoucing                                     │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule noDefaultAdminChange(env e, method f, calldataarg args) {
+  require nonZeroAccount(e.msg.sender);
+  requireInvariant defaultAdminConsistency(defaultAdmin());
+  requireInvariant singleDefaultAdmin(e.msg.sender, defaultAdmin());
+  
+  address adminBefore = defaultAdmin();
+  f@withrevert(e, args);
```
frangio, line 100: This shouldn't be necessary right? It should be removed in all of the "no change" rules. If `f` reverts, we know there was no change.

```suggestion
  f(e, args);
```

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,379 @@
...
+
+  assert adminBefore != adminAfter => (
+    (f.selector == acceptDefaultAdminTransfer().selector && adminAfter != 0) ||
+    (f.selector == renounceRole(bytes32,address).selector && adminAfter == 0)
+  ), "default admin is only affected by accepting an admin transfer or renoucing";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: pendingDefaultAdmin is only affected by beginning, accepting or cancelling an admin transfer                  │
```
frangio, line 111:
```suggestion
│ Rule: pendingDefaultAdmin is only affected by beginning, accepting or canceling an admin transfer                  │
```

:stuck_out_tongue:

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,379 @@
...
+  requireInvariant defaultAdminConsistency(defaultAdmin());
+  requireInvariant singleDefaultAdmin(e.msg.sender, defaultAdmin());
+  
+  address adminBefore = defaultAdmin();
+  f@withrevert(e, args);
+  address adminAfter = defaultAdmin();
+
+  assert adminBefore != adminAfter => (
+    (f.selector == acceptDefaultAdminTransfer().selector && adminAfter != 0) ||
+    (f.selector == renounceRole(bytes32,address).selector && adminAfter == 0)
```
frangio, line 105: This spec is large, so I would favor simplicity in the "no change" rules and would not include any postconditions:

```suggestion
    f.selector == acceptDefaultAdminTransfer().selector ||
    f.selector == renounceRole(bytes32,address).selector
```

These are all checked already by the respective rules for each function.

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,379 @@
...
+  ),
+    "pending delay schedule is set in the future unless accepted edge cases";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: pending delay wait is set depending on increasing or decreasing the delay                                     │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule pendingDelayWait(env e, uint48 newDelay) {
```
frangio, line 321: This rule is proving our code correct against a reference model, but it would be nice to have a more abstract rule that proves instead _the property that we were aiming for_: that if you change the delay wait and then change the admin, the old delay must have passed in between. Do you see what I mean? I feel it should be possible to do this with two separate `env` variables.

certora/harnesses/AccessControlDefaultAdminRulesHarness.sol
```diff
@@ -0,0 +1,47 @@
...
+        uint48 initialDelay,
+        address initialDefaultAdmin,
+        uint48 delayIncreaseWait
+    ) AccessControlDefaultAdminRules(initialDelay, initialDefaultAdmin) {
+        _delayIncreaseWait = delayIncreaseWait;
+    }
+
+    // FV
+    function _pendingDefaultAdmin() external view returns (address) {
+        (address newAdmin, ) = pendingDefaultAdmin();
```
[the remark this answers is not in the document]
ernestognw: Unfortunately, tuples are not valid in Certora, this syntax doesn't compile:

```
(address pendingAdminBefore, scheduleBefore) = pendingDefaultAdmin(e);
```

You can see they mention "tuples are not CVL types" [here](https://docs.certora.com/en/latest/docs/cvl/ghosts.html?highlight=tuple#declaring-ghost-variables), whereas the official recommendation is to use [Harnessing to workaround these issues](https://docs.certora.com/en/latest/docs/cvl/types.html?highlight=harnessing#solidity-types)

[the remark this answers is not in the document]
ernestognw: Will change the underscore, I added `delayChangeWait()` at the very end and it screwed up the pattern 😆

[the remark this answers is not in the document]
ernestognw: I renamed all of the `FV` exposed functions to use a trailing `_` suffix instead.

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: Nice catch, thanks!

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: You're right, removing from all of them

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: Now that you mention it, I don't think so. At least for these particular examples.
I'd keep it like this for consistency at the moment but we can do a PR after changing these kind of functions for definitions.

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: Yes, I see what you mean. Will keep playing with this and I'll update if I can come up with an example for that property.

frangio: Check this out in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4106:

image

certora/specs/methods/IAccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,36 @@
+import "./IERC5313.spec"
+
+methods {
```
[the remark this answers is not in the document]
ernestognw: I removed the functions and it's working, however, I found interesting that I kept coming back to the `methods` block for reference. Although I'm okay removing them, I think it somewhat improves my developer experience because I know in what order I'm trying to make a spec.

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -203,6 +269,24 @@ rule beginDefaultAdminTransfer(env e, address newAdmin) {
     "pending default admin delay is set";
 }
 
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: A default admin can't change in less than the applied schedule                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule pendingDefaultAdminDelayEnforced(env e1, env e2, method f, calldataarg args, address newAdmin) {
```
frangio, line 277: Seeing this rule now I'm realizing that my initial suggestion didn't make a lot of sense. This rule doesn't convince me that there is no way to change the admin faster, because it's only looking at ways that consist of invoking a single function (`f`). What if there was another way to violate the property?

I have some ideas for how we might try to prove that using an invariant and a ghost variable, but I don't know if they make sense fully...

We can just leave this rule as is with a comment saying that there may be ways to generalize it.

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -203,6 +269,24 @@ rule beginDefaultAdminTransfer(env e, address newAdmin) {
...
+  require intervalSanity(e1, e2);
+
+  uint48 delayBefore = defaultAdminDelay(e1);
+  address adminBefore = defaultAdmin();
+  beginDefaultAdminTransfer(e1, newAdmin);
+  f(e1, args);
```
frangio, line 283: Did you mean `e2` here? Otherwise `e2` is basically not used in the rule. I think as writen this rule is vacuous, because `adminAfter == newAdmin` will always be false.

certora/harnesses/AccessControlDefaultAdminRulesHarness.sol
```diff
@@ -0,0 +1,47 @@
...
+        (uint48 newDelay, ) = pendingDefaultAdminDelay();
+        return newDelay;
+    }
+
+    function pendingDelaySchedule_() external view returns (uint48) {
+        (, uint48 schedule) = pendingDefaultAdminDelay();
+        return schedule;
+    }
+
+    function delayChangeWait_(uint48 newDelay) external view returns (uint48) {
```
Amxx, line 39: why the trailing _ here ?

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant defaultAdminConsistency(address account)
+  defaultAdmin() == account <=> hasRole(DEFAULT_ADMIN_ROLE(), account) 
+  {
+    preserved {
+      // defaultAdmin() returns the zero address when there's no default admin
+      require nonZeroAccount(account);
+    }
+  }
```
Amxx, line 70: Not sure about the preserve here. I'm wondering if we should rather change the invariant like that

```suggestion
invariant defaultAdminConsistency(address account)
  (account == defaultAdmin() && account != 0) <=> hasRole(DEFAULT_ADMIN_ROLE(), account)
```

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule noDefaultAdminDelayIncreaseWaitChange(env e, method f, calldataarg args) {
+  uint48 delayIncreaseWaitBefore = defaultAdminDelayIncreaseWait();
+  f(e, args);
+  uint48 delayIncreaseWaitAfter = defaultAdminDelayIncreaseWait();
+
+  assert delayIncreaseWaitBefore == delayIncreaseWaitAfter, 
+    "delay increase wait can't be changed atomically by any function";
+}
```
Amxx, line 245: same comment as for `noDefaultAdminDelayChange`. I'm a bit lost with all these names ... but still "can't be changed atomically" is not the right description.

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: A default admin can't change in less than the applied schedule                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule pendingDefaultAdminDelayEnforced(env e1, env e2, method f, calldataarg args, address newAdmin) {
+  require intervalSanity(e1, e2);
+
+  uint48 delayBefore = defaultAdminDelay(e1);
+  address adminBefore = defaultAdmin();
+  beginDefaultAdminTransfer(e1, newAdmin);
```
Amxx, line 282: I'm not sure what the point of this call is.

certora/specs/AccessControlDefaultAdminRules.spec
Amxx: I share the concern about e2 ... but I don't get the rest of the comment

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+  ) => newAdmin == pendingAdminBefore, "pending admin stays the same if the new admin set is the same";
+  assert (
+    pendingAdminBefore != pendingDefaultAdmin_(e) &&
+    scheduleBefore == pendingDefaultAdminSchedule_(e)
+  ) => (
+    // Schedule doesn't change if:
+    // - The previous schedule is block.timestamp and the delay is 0
+    (e.block.timestamp == scheduleBefore && defaultAdminDelay(e) == 0) ||
+    // - The defaultAdminDelay was reduced to a value such that added to the block.timestamp is equal to previous schedule
+    e.block.timestamp + defaultAdminDelay(e) == scheduleBefore
```
Amxx, line 316: `e.block.timestamp == scheduleBefore && defaultAdminDelay(e) == 0` 
is a special case of
`e.block.timestamp + defaultAdminDelay(e) == scheduleBefore`

so you can remove the first condition since the second part will cover it

certora/specs/methods/IAccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,7 @@
...
+    defaultAdmin() returns(address) envfree
+    pendingDefaultAdmin() returns(address, uint48) envfree
```
Amxx, line 5: You are not including some function 
```
defaultAdminDelay() returns (uint48)
pendingDefaultAdminDelay() returns (address, uint48)
...
beginDefaultAdminTransfer
cancelDefaultAdminTransfer
acceptDefaultAdminTransfer
changeDefaultAdminDelay
rollbackDefaultAdminDelay
```

plus the one for easy instrumentation
```
pendingDefaultAdmin_() returns (address)
pendingDefaultAdminSchedule_() returns (uint48)
pendingDelay_() returns (uint48)
pendingDelaySchedule_() returns (uint48)
delayChangeWait_(uint48 newDelay) returns (uint48)
```

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -136,7 +136,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @dev See {AccessControl-_revokeRole}.
      */
     function _revokeRole(bytes32 role, address account) internal virtual override {
-        if (role == DEFAULT_ADMIN_ROLE) {
+        if (role == DEFAULT_ADMIN_ROLE && account == _currentDefaultAdmin) {
             delete _currentDefaultAdmin;
         }
         super._revokeRole(role, account);
```
Amxx, line 142: why not

```
bool success = super._revokeRole(role, account);
if (role == DEFAULT_ADMIN_ROLE && success) {
  delete _currentDefaultAdmin;
}
```
?

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+│ Rule: A delay can't change in less than the applied schedule                                                        │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule pendingDelayWaitEnforced(env e1, env e2, method f, calldataarg args, uint48 newDelay) {
+  require intervalSanity(e1, e2);
+
+  uint48 delayBefore = defaultAdminDelay(e1);
+  changeDefaultAdminDelay(e1, newDelay);
+  f(e1, args);
+  uint48 delayAfter = defaultAdminDelay(e2);
```
Amxx, line 415: why e2 here but not in f?

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+│ Rule: pending delay wait is set depending on increasing or decreasing the delay                                     │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule pendingDelayWait(env e, uint48 newDelay) {
+  changeDefaultAdminDelay(e, newDelay);
+
+  assert newDelay > defaultAdminDelay(e) => pendingDelaySchedule_(e) == increasingDelaySchedule(e, newDelay),
+    "Delay wait is the minimum between the new delay and a threshold when the delay is increased";
+  assert newDelay <= defaultAdminDelay(e) => pendingDelaySchedule_(e) == decreasingDelaySchedule(e, newDelay),
+    "Delay wait is the difference between the current and the new delay when the delay is decreased";
```
Amxx, line 434:
```suggestion
  uint48 oldDelay = defaultAdminDelay(e);
  changeDefaultAdminDelay(e, newDelay);

  assert newDelay > oldDelay => pendingDelaySchedule_(e) == increasingDelaySchedule(e, newDelay),
    "Delay wait is the minimum between the new delay and a threshold when the delay is increased";
  assert newDelay <= oldDelay => pendingDelaySchedule_(e) == decreasingDelaySchedule(e, newDelay),
    "Delay wait is the difference between the current and the new delay when the delay is decreased";
```

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: It fails when removing the preserve because `defaultAdmin()` always reports the `address(0)` when there's no admin (eg. has been renounced). So the counterexample is that `defaultAdmin()` can be `address(0)` when `hasRole(DEFAULT_ADMIN_ROLE, address(0)) == false`. If we do your suggestion, we'll end up with a more complex rule:

```
invariant defaultAdminConsistency(address account)
  (account == defaultAdmin() && account != 0 || defaultAdmin() == 0 && account == 0) <=> hasRole(DEFAULT_ADMIN_ROLE(), account)
```

Also, it affects all of the following

```diff
+ assert success <=> hasRole(DEFAULT_ADMIN_ROLE, e.msg.sender)
- assert success <=> e.msg.sender == defaultAdmin()
```

I think is way cleaner keeping the `preserved` since the assumption of non-zero address is reasonable, but I'm open to discussion

certora/specs/methods/IAccessControlDefaultAdminRules.spec
ernestognw: I was including them previously, but we decided to remove them [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4180#discussion_r1166067453). The conclusion after a call we had is that the blocks method doesn't help in anything unless you specify `envfree` or any other summary.

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+  defaultAdmin() == owner()
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Function correctness: revokeRole only affects the specified user/role combo                                         │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule revokeRoleEffect(env e, bytes32 role) {
+    require nonpayable(e);
+    require role != DEFAULT_ADMIN_ROLE();
```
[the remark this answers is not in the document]
ernestognw: I did, thanks!

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Function correctness: renounceRole only affects the specified user/role combo                                       │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule renounceRoleEffect(env e, bytes32 role) {
+    require nonpayable(e);
+    require role != DEFAULT_ADMIN_ROLE();
```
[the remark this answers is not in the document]
ernestognw: Also done

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: pendingDefaultAdmin is only affected by beginning, accepting or canceling an admin transfer                   │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule noPendingDefaultAdminChange(env e, method f, calldataarg args) {
+  requireInvariant defaultAdminConsistency(defaultAdmin());
+  requireInvariant singleDefaultAdmin(e.msg.sender, defaultAdmin());
+
+  address pendingAdminBefore = pendingDefaultAdmin_(e);
+  address scheduleBefore = pendingDefaultAdminSchedule_(e);
```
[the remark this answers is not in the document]
ernestognw: Nice catch, changed

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -0,0 +1,484 @@
...
+    "pending default admin delay is set";
+}
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: A default admin can't change in less than the applied schedule                                                │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule pendingDefaultAdminDelayEnforced(env e1, env e2, method f, calldataarg args, address newAdmin) {
+  require intervalSanity(e1, e2);
```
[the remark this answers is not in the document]
ernestognw: Replaced in all the places where used

contracts/access/AccessControlDefaultAdminRules.sol
ernestognw: I don't get this comment. I'm applying the fix from #4177

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: Accidentally, thanks for noticing!

certora/specs/AccessControlDefaultAdminRules.spec
Amxx:
```
invariant defaultAdminConsistency(address account)
  (account == defaultAdmin() && account != 0 || defaultAdmin() == 0 && account == 0) <=> hasRole(DEFAULT_ADMIN_ROLE(), account)
```
???

if defaultAdmin and account are both 0, `hasRole(DEFAULT_ADMIN_ROLE(), account)` is false.
Also, mixing `&&` and `||` without proper parenthesis is really confusing.

Amxx: Note that in my proposal, 
```(account == defaultAdmin() && account != 0) <=> hasRole(DEFAULT_ADMIN_ROLE(), account)```
if the admin is renounced, and account is 0, the left side is false. I don't see how that is a counterexample

Comment (ernestognw) [reply certora/specs/AccessControlDefaultAdminRules.spec:66]: The invariant you proposed is still broken for the acceptDefaultAdmin case. What I'm saying is that it ends up in worse readability issues.

We'll keep playing around with this to see if I find a better alternative, but I'd prefer to keep it as it is rn

File: certora/specs/AccessControlDefaultAdminRules.spec [hunk 25]
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+/*
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Invariant: Only one account holds the DEFAULT_ADMIN_ROLE                                                            │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+invariant singleDefaultAdmin(address account, address another)
+  hasRole(DEFAULT_ADMIN_ROLE(), account) && hasRole(DEFAULT_ADMIN_ROLE(), another) => another == account
+  filtered { f -> f.selector != acceptDefaultAdminTransfer().selector }
```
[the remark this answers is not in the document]
Amxx: I don't understand. We are talking about the fact there is suppose to be only one account that is the default admin. Its essential we verify that `acceptDefaultAdminTransfer` doesn't break that invariant.

What your invariant is currently saying is :
> we know that if accounts are both admin, then they are the same, because there is only a single admin ...
> ... but we only check that this is true until someone calls `acceptDefaultAdminTransfer`.

So implicitly you are saying 
> if someone calls `acceptDefaultAdminTransfer`, then anything can happen, there might be multiple admins

Its ok to say "when processing the function, I'm going to help you with some `preserve`", but IMO its not ok to just say "lets not consider this function"

certora/specs/AccessControlDefaultAdminRules.spec
Amxx: When doing a rule, the state before the rule can be anything. 

You don't need to set the state (like with fuzzing) ... because unless you explicitly required that there is no pending admin, then the system will consider that there might be one.

certora/specs/methods/IAccessControlDefaultAdminRules.spec
Amxx: Technically it helps if you ever run the rules on a contract that fail to implement them.

contracts/access/AccessControlDefaultAdminRules.sol
Amxx: Commented there instead

certora/specs/methods/IAccessControlDefaultAdminRules.spec
ernestognw: How? We only found it helpful for `envfree` methods.

ernestognw: How? We only found it helpful for `envfree` methods.

ernestognw: And?

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: > Did you mean e2 here? 
Yes.

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: Would be helpful if you tell me why is not the right description

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: This still doesn't work. I've spent some time trying it out and I don't think it's worth it, the counterexample is well defined and explained, so I'm solving.

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: I think `can't be changed **atomically**` is the right description in the context of a transaction. A single transaction can't change this value and that's all that this rule is proving.

I'm solving since there's no clear recommendation or constructive question

certora/specs/methods/IAccessControlDefaultAdminRules.spec
Amxx: IMO declaring all the function we use as some advantages
- an exhaustive list of function makes it clear which ones are envfree, and which ones are not, without having to dive into the contract
- if someone wants to reuse the rule on a contract that is missing some functions (like an ERC20 that doesn't have a burn function) it would be possible to just plug our rules on top of their contract and get (partial) coverage
- in case we ever want to change the dispatcher settings, that would reduce the changes needed
- just be explicit about what it is the specs needs to fully run

The drawback I see is the maintenance that it requires. That is not really an issue if we add a function to the contract (it will still be considered when doing `method f`) ... it gets bad if we remove a function (in that case the rules are disabled with no error). Removing functions is not something we really do.

ernestognw: Can you wait until I request a review? Please

certora/specs/AccessControlDefaultAdminRules.spec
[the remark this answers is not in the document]
ernestognw: I've spent the whole day on this and I don't find an easy way to make this invariant applicable to the `acceptDefaultAdminTransfer()`, I've tried with different ghost and hooks setups, as long as different variations of the same invariant, an example of how complex (and still not passing) is:

```
ghost defaultAdmins() returns uint256;

hook Sstore _roles[KEY bytes32 role].(offset 0)[KEY address account] bool granted (bool oldGranted) STORAGE {
  mathint countChange = 
    granted && !oldGranted ? 1 : // Granted
    !granted && oldGranted ? -1 : // Revoked
    0; // Anything else
    
  havoc defaultAdmins assuming defaultAdmins@new() == defaultAdmins@old() + countChange;
}

...

invariant singleDefaultAdmin(address account, address another)
  defaultAdmins() <= 1 &&
  (hasRole(DEFAULT_ADMIN_ROLE(), account) && hasRole(DEFAULT_ADMIN_ROLE(), another) => another == account)
  {
    preserved {
      requireInvariant singleDefaultAdmin(another, account);
      requireInvariant defaultAdminConsistency(defaultAdmin());
      require nonZeroAccount(account);
    }
  }
```

The main reason for it not pass is that I don't find a way of forcing the contract not to consider cases in which `role[DEFAULT_ADMIN_ROLE].members` only have one member at most since the beginning. The prover assumes this is a valid initial state:

Captura de pantalla 2023-05-03 a la(s) 19 03 29

certora/specs/methods/IAccessControlDefaultAdminRules.spec
ernestognw: So, overall I agree with keeping them. Initially, they were removed because we didn't see value in having but we also agreed that consistency is better.

I readded them because I also see high value on keeping them. Solving since the suggestion is accepted.

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: The problem when removing the `beginDefaultAdminTransfer` is that any other state is valid so that the rule won't be attached to the `delayBefore`, so we will be proving only that the admin changed without a strict relation with the delay.

If I remove beginDefaultAdminTransfer, then the rule basically becomes the same as in `noDefaultAdminChange`.

If you find a way to simplify those two, please let me know, or feel free to go and implement it, would appreciate it :)

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: My answer to this is basically [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4180#discussion_r1184005586). TLDR; is that generalizing the rule detaches the change from the `delayBefore`.

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -484,3 +432,50 @@ rule rollbackDefaultAdminDelay(env e) {
   assert success => pendingDelaySchedule_(e) == 0,
     "Pending default admin delay is reset";
 }
+
+/* 
+┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
+│ Rule: pending default admin and the delay can only change along with their corresponding schedules                  │
+└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+*/
+rule pendingValueAndScheduleCoupling(env e, address newAdmin, uint48 newDelay) {
```
frangio, line 446: Personally I don't really understand why this rule verifies a property for both default admin and delay, which are separate values. IMO, it would be more correct to verify them separately so that the first one doesn't constrain the state on which the second one is verified.

That said, I don't consider this a blocker.

frangio (dismissed):

frangio (approved):
I've fixed an error in the specs due to #4211.

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: I see your point. We [agreed on merging](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4180#discussion_r1178014239) them to have fewer rules, but the state constraints kinda worry me.

I also don't consider it a blocker

Amxx (approved):
I made a small change to clarify associativity in a place where I was not sure how `||` and `&&` were resolved.

I have a list of remarques / comments / suggestion to further improve this specs file in the future, but this is good to merge IMO !

frangio (approved):

PR #4182, opened 2023-04-13
move loop invariant variables out from loop

MollyHe5:
Moving loop invariant codes out from loop can save gas.
In the following demo, it can save 765 units of gas when the length of ```arr``` is 100.

```
    function test1(uint256[] calldata arr) public {     
        for(uint256 i = 0; i < arr.length; i++) {
        }
    }

    function test2(uint256[] calldata arr) public {
        uint256 len = arr.length;
        for(uint256 i = 0; i < len; i++) {
        }
    }
```

#### PR Checklist

- [ ] Tests
- [ ] Documentation


songlh (commented):
All changes can preserve the original functionality, while saving some gases.

songlh:
@nventuro @frangio @Amxx Can you help take a look at this pull request?

frangio:
>it can save 765 units of gas when the length of arr is 100.

First of all, I can't reproduce this benchmark. I only see 5 gas difference in total.

Moreover, if it was 765 gas, it would mean less than 8 gas per iteration. This is marginal, we are not interested in such a small optimization.

PR #4183, opened 2023-04-14
Fix release note in `IERC1967`

pcaversaccio:
As title. If you think this requires a CHANGESET, I can create one.


ernestognw (approved):
LGTM, although there's an overflow in our Checkpoints fuzzing templates, looking into it. No need for a changeset though, thanks!

ernestognw:
> I guess this PR #4184 blocks the merging of this PR, correct @ernestognw?

Not really, they can be merged separately. I'm waiting for a second review as per policy, I'm adding review requests so thanks for the heads up.

PR #4184, opened 2023-04-14
Fix Checkpoints fuzz overflow

ernestognw:
There was an edge case on the property tests for `Checkpoints.sol` when checking for reverts when setting past keys. When the key is 0, substracting 1 results on an Arithmetic overflow/underflow error.

### DIFF
--- a/scripts/generate/templates/Checkpoints.t.js
+++ b/scripts/generate/templates/Checkpoints.t.js
@@ -176,11 +176,13 @@ function testPush(
     // Can't push any key in the past
     if (keys.length > 0) {
         ${opts.keyTypeName} lastKey = keys[keys.length - 1];
-        pastKey = _bound${capitalize(opts.keyTypeName)}(pastKey, 0, lastKey - 1);
-
-        vm.roll(pastKey);
-        vm.expectRevert();
-        this.push(values[keys.length % values.length]);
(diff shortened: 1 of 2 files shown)


scripts/generate/templates/Checkpoints.t.js
```diff
@@ -86,10 +86,13 @@ function testPush(
 
     if (keys.length > 0) {
         ${opts.keyTypeName} lastKey = keys[keys.length - 1];
-        pastKey = _bound${capitalize(opts.keyTypeName)}(pastKey, 0, lastKey - 1);
-
-        vm.expectRevert();
-        this.push(pastKey, values[keys.length % values.length]);
+        if (lastKey > 0) {
+            ${opts.keyTypeName} maxKey = lastKey == 0 ? 0 : lastKey - 1;
+            pastKey = _bound${capitalize(opts.keyTypeName)}(pastKey, 0, maxKey);
```
frangio, line 91: The `maxKey` variable is redundant because of the conditional.

```suggestion
            pastKey = _bound${capitalize(opts.keyTypeName)}(pastKey, 0, lastKey - 1);
```

frangio (changes requested):

Amxx (approved):

PR #4188, opened 2023-04-21
Merge release-v4.8

frangio:
Merges deleted changesets and updated file headers from 4.8.3.


ernestognw (commented):
Why are we removing the changesets if they didn't make it into the `CHANGELOG.md`?

ernestognw (changes requested):
My impression is that conflicts from https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4188/commits/0da9c50dc32088ece130f72202efa8ba24cbb869 where not correctly solved, am I missing something?

frangio:
You're right. I ran a command like this one:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/91df66c4a9dfd0425ff923cbeb3a20155f1355ea/scripts/release/workflow/prepare-release-merge.sh#L15

I think the merge error was because of `-X theirs`.

We may need to fix this in that script.

ernestognw:
> You're right. I ran a command like this one:
> 
> https://github.com/OpenZeppelin/openzeppelin-contracts/blob/91df66c4a9dfd0425ff923cbeb3a20155f1355ea/scripts/release/workflow/prepare-release-merge.sh#L15
> 
> I think the merge error was because of `-X theirs`.
> 
> We may need to fix this in that script.

I see, but this is because the update was made on this branch, whereas a CHANGELOG update should be done on the v4.8 branch. Shouldn't happen when using the automation

ernestognw (approved):

PR #4191, opened 2023-04-24
Improve Address tests

ernestognw:
Coverage for address increased to 100% by adding the test cases needed to cover all branches.
Also updated a reference to the solidity docs.

### DIFF
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -59,7 +59,7 @@ library Address {
      * IMPORTANT: because control is transferred to `recipient`, care must be
      * taken to not create reentrancy vulnerabilities. Consider using
      * {ReentrancyGuard} or the
-     * https://solidity.readthedocs.io/en/v0.5.11/security-considerations.html#use-the-checks-effects-interactions-pattern[checks-effects-interactions pattern].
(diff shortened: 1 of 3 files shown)


contracts/utils/Address.sol
```diff
@@ -59,7 +59,7 @@ library Address {
      * IMPORTANT: because control is transferred to `recipient`, care must be
      * taken to not create reentrancy vulnerabilities. Consider using
      * {ReentrancyGuard} or the
-     * https://solidity.readthedocs.io/en/v0.5.11/security-considerations.html#use-the-checks-effects-interactions-pattern[checks-effects-interactions pattern].
+     * https://solidity.readthedocs.io/en/latest/security-considerations.html#use-the-checks-effects-interactions-pattern[checks-effects-interactions pattern].
```
Amxx, line 62: Here we fixed the version because we can't be sure that this documentation will still be in "latest" in the future. 

Said otherwise:
- the content of latest can change
- the content of 0.5.11 cannot

ernestognw: Let's fix it to `v0.8.0`.

We use `latest` in multiple parts of the code, and I don't think it's ok sending people to the `v0.5.11` docs

frangio (approved):
LGTM

Amxx (approved):

PR #4195, opened 2023-04-26
Only run FV on new or updated specs

Amxx:
Fixes LIB-788


Amxx:
So the FV fails because this branch comes from my repo, and we are missing the certora apikey. However, the good thing is that it fails only once, for Ownable (which was modified in the PR). Other specs are not run.

image

certora/run.js
```diff
@@ -1,37 +1,65 @@
...
 const { PassThrough } = require('stream');
 const events = require('events');
-const limit = require('p-limit')(MAX_PARALLEL);
 
-let [, , request = '', ...extraOptions] = process.argv;
-if (request.startsWith('-')) {
-  extraOptions.unshift(request);
-  request = '';
+const argv = require('yargs')
+  // .env('')
```
frangio, line 15:
```suggestion
```

certora/run.js
```diff
@@ -1,37 +1,65 @@
...
-  const [reqSpec, reqContract] = request.split(':').reverse();
-  specs = Object.values(specs).filter(s => reqSpec === s.spec && (!reqContract || reqContract === s.contract));
-  if (specs.length === 0) {
-    console.error(`Error: Requested spec '${request}' not found in specs.json`);
-    process.exit(1);
+const specs = require(argv.spec).filter(s => argv.all || argv._.some(r => match(s, r)));
+const limit = require('p-limit')(argv.parallel);
+
+if (argv._.length == 0 && !argv.all) {
+  console.error(`Warning: No specs requested. Did you forgot to toggle '--all'`);
```
frangio, line 47:
```suggestion
  console.error(`Warning: No specs requested. Did you forgot to toggle '--all'?`);
```

certora/run.js
```diff
@@ -1,37 +1,65 @@
...
+    parallel: {
+      alias: 'p',
+      type: 'number',
+      default: 4,
+    },
+    options: {
+      alias: 'o',
+      type: 'array',
+      default: [],
+    },
```
frangio, line 35: This doesn't seem to work as expected. E.g., `-o --optimistic_loop` results in `{ options: [], optimistic_loop: true }`.

We may need a special way to run a single spec with options...

Amxx: I'll check that !

maybe `-o "--optimistic_loop"` is the way to go

Amxx: Note that
```
node ./certora/run.js -o=--optimistic_loop -o=--optimistic_hashing ERC20
```
will not work and that you should do either
```
node ./certora/run.js ERC20 -o=--optimistic_loop -o=--optimistic_hashing
node ./certora/run.js -o=--optimistic_loop -o=--optimistic_hashing -- ERC20
```

frangio: Ok, I don't think this is a good interface but let's just merge it and see how it works for us.

frangio (approved):

frangio:
@Amxx Please update the PR title according to the guidelines before merging.

frangio:
Verification seems to be failing with:

```
CRITICAL: Encountered an error running Certora Prover:
You are using a demo version of the tool in a CI environment.
                     If you have a premium Certora key, please set it as the environment variable CERTORAKEY.
```

PR #4197, opened 2023-04-26
Add reentrancy test cases for ERC4626

ernestognw:
Checked that:

- Reentrancy on deposit doesn't change the shares price
- Reentrancy on withdrawal doesn't change the assets price

### DIFF
--- a/contracts/mocks/ERC4626ReentrantAsset.sol
+++ b/contracts/mocks/ERC4626ReentrantAsset.sol
@@ -3,7 +3,6 @@ pragma solidity ^0.8.0;
 
 import "../token/ERC20/ERC20.sol";
 import "../token/ERC20/extensions/ERC4626.sol";
-import "hardhat/console.sol";
 
 contract ERC4626ReentrantAsset is ERC20("TEST", "TST") {
     bool _entered;
--- a/contracts/mocks/ERC4626ReentrantAsset.sol
+++ b/contracts/mocks/ERC4626ReentrantAsset.sol
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../token/ERC20/ERC20.sol";
+import "../token/ERC20/extensions/ERC4626.sol";
+import "hardhat/console.sol";
+
+contract ERC4626ReentrantAsset is ERC20("TEST", "TST") {
+    bool _entered;
+    bool _canReenter;
+
+    function setReenter(bool reenter) external {
+        _canReenter = reenter;
+    }
+}
+
+contract ERC4626DepositReentrantAsset is ERC4626ReentrantAsset {
+    function _beforeTokenTransfer(address, address to, uint256) internal override {
+        if (_canReenter && !_entered) {
+            _entered = true;
+            this.approve(msg.sender, 1);
+            IERC4626(msg.sender).deposit(1, to);
+        }
+    }
+}
+
+contract ERC4626WithdrawReentrantAsset is ERC4626ReentrantAsset {
+    /**
+     * @dev Simulates a deposit before enabling the withdraw reentrancy
+     */
+    function depositTo(address vault, uint256 amount) external {
+        this.approve(vault, amount);
+        IERC4626(vault).deposit(amount, address(this));
+    }
+
+    function _afterTokenTransfer(address, address to, uint256) internal override {
+        if (_canReenter && !_entered) {
+            _entered = true;
+            IERC4626(msg.sender).withdraw(1, to, address(this));
+        }
+    }
+}
--- a/contracts/mocks/ERC4626ReentrantAsset.sol
+++ b/contracts/mocks/ERC4626ReentrantAsset.sol
@@ -5,37 +5,39 @@ import "../token/ERC20/ERC20.sol";
 import "../token/ERC20/extensions/ERC4626.sol";
 
 contract ERC4626ReentrantAsset is ERC20("TEST", "TST") {
-    bool _entered;
-    bool _canReenter;
-
-    function setReenter(bool reenter) external {
-        _canReenter = reenter;
+    enum Type {
+        No,
+        Before,
+        After
     }
-}
 
-contract ERC4626DepositReentrantAsset is ERC4626ReentrantAsset {
-    function _beforeTokenTransfer(address, address to, uint256) internal override {
-        if (_canReenter && !_entered) {
-            _entered = true;
-            this.approve(msg.sender, 1);
-            IERC4626(msg.sender).deposit(1, to);
+    Type reenterType;
+    address reenterTarget;
+    bytes reenterData;
+
+    function scheduleReenter(Type when, address target, bytes calldata data) external {
+        reenterType = when;
+        reenterTarget = target;
+        reenterData = data;
+    }
+
+    function functionCall(address target, bytes memory data) public returns (bytes memory) {
+        return Address.functionCall(target, data);
+    }
+
+    function _beforeTokenTransfer(address from, address to, uint256 amount) internal override {
+        if (reenterType == Type.Before) {
+            reenterType = Type.No;
+            functionCall(reenterTarget, reenterData);
         }
-    }
-}
-
-contract ERC4626WithdrawReentrantAsset is ERC4626ReentrantAsset {
-    /**
-     * @dev Simulates a deposit before enabling the withdraw reentrancy
-     */
-    function depositTo(address vault, uint256 amount) external {
-        this.approve(vault, amount);
-        IERC4626(vault).deposit(amount, address(this));
+        super._beforeTokenTransfer(from, to, amount);
     }
 
-    function _afterTokenTransfer(address, address to, uint256) internal override {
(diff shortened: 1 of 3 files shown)


test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -44,6 +46,49 @@ contract('ERC4626', function (accounts) {
...
+    await token.$_mint(holder, amount);
+    await token.$_mint(token.address, amount); // Requires balance to reenter
+
+    await token.approve(vault.address, constants.MAX_UINT256, { from: holder });
+
+    await token.setReenter(true);
+
+    const sharesBefore = await vault.previewDeposit(amount, { from: holder });
+    await vault.deposit(amount, holder, { from: holder });
+    const sharesAfter = await vault.previewDeposit(amount, { from: holder });
```
frangio, line 64: It's not really clear to me why the price would be changed after the deposit is done. I think the key thing is that the attacker should not be able to get/observe a low price _during_ the reentrancy, and that's something we could test more directly by looking at the price via `convertToAssets` or the preview functions.

ernestognw: The reason why the price changes is because during reentrancy there's a different proportion between assets and shares after the first `safeTransferFrom` or `mint`, which is used for the next deposit/withdraw.

The problem I see is that _during_ the reentrancy, the price is the same. You need to reenter `deposit` at least once so that you use the same price twice. Just observing the price during reentrancy is not enough

Amxx (commented):
I'm a bit confused by the mocks "hardcoding" the reentrancy scenario, doing approves and everything. What would you think of having a more generic mock ... where the reentrancy would be programmed in the unit test?

```solidity
contract ERC4626ReentrantAsset is ERC20("TEST", "TST") {
    enum Type {
        No,
        Before,
        After
    };

    Type    reenterType;
    address reenterTarget;
    bytes   reenterData;

    function scheduleReenter(Type t, address target, bytes calldata data) external {
        reenterType   = t;
        reenterTarget = target;
        reenterData   = data;
    }

    function _beforeTokenTransfer(address from, address to, uint256 amount) internal override {
        if (reenterType == Type.Before) {
            reenterType = Type.No;
            reenterTarget.call(reenterData);
        }
    }

    function _afterTokenTransfer(address from, address to, uint256 amount) internal override {
        super._afterTokenTransfer(from, to, amount);
        if (reenterType == Type.After) {
            reenterType = Type.No;
            reenterTarget.call(reenterData);
        }
    }
}
```

ernestognw:
Just added your proposal @Amxx, looks way better, thanks!

test/token/ERC20/extensions/ERC4626.test.js
ernestognw: I added extra comments so I'd say now is pretty clear since you commented on Slack that we might want to just add comments of how it works.,

frangio (approved):
LGTM. Thanks!

contracts/mocks/ERC4626ReentrantAsset.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../token/ERC20/ERC20.sol";
+import "../token/ERC20/extensions/ERC4626.sol";
+
+contract ERC4626ReentrantAsset is ERC20("TEST", "TST") {
```
Amxx, line 7: Can be find a better name? This is not ERC4626.

Sure it is used in the context of ERC4626, but its more of a `ERC20ArbitraryCallMock` or something like that.

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -44,6 +45,81 @@ contract('ERC4626', function (accounts) {
     }
   });
 
+  describe('reentrancy', async function () {
+    const reenterType = {
+      Before: 1,
+      After: 2,
+    };
```
Amxx, line 52: we have an helper for that in `test/helpers/enums.js`. Doesn't make a big difference, but for consistency it might be better to use it.

```suggestion
    const reenterType = Enum('No', 'Before', 'After');
```

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -44,6 +45,81 @@ contract('ERC4626', function (accounts) {
...
+    });
+
+    // During a `_deposit`, the vault does `transferFrom(depositor, vault, assets)` -> `_mint(receiver, shares)`
+    // such that a reentrancy BEFORE the transfer guarantees the price is kept the same.
+    // If the order of transfer -> mint is changed to mint -> transfer, the reentrancy could be triggered on an
+    // intermediate state in which the ratio of assets/shares has been decreased (more shares than assets).
+    it('reenters before deposit without changing the price', async function () {
+      // Schedules a reentrancy from the token contract
+      const reenterAmount = 1;
+      await token.functionCall(token.address, vault.contract.methods.approve(vault.address, reenterAmount).encodeABI());
```
Amxx, line 74: You are targetting `token.address` but you are encoding using `vault.contract.methods` ?

IMO that would be simpler (if it is what you are trying to do)
```suggestion
      await token.$_approve(token.address, vault.address, reenterAmount);
```

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -44,6 +45,81 @@ contract('ERC4626', function (accounts) {
...
+      amount = web3.utils.toBN(10).pow(decimals);
+      token = await ERC4626ReentrantAsset.new();
+
+      vault = await ERC4626.new('', '', token.address);
+      await token.$_mint(holder, amount);
+      await token.$_mint(token.address, amount); // Requires balance to reenter
+    });
+
+    // During a `_deposit`, the vault does `transferFrom(depositor, vault, assets)` -> `_mint(receiver, shares)`
+    // such that a reentrancy BEFORE the transfer guarantees the price is kept the same.
```
Amxx, line 68: I'm not sure about this test for two reason:

- you check the price before and after the deposit that includes a re-entrancy... and there is no difference. But that is partly because the reentrant part does not change the price of the asset (a deposit doesn't affect price)
- what you want to check is that the price applied to the user is the one that was previewed... that is about it

What I would do:
- remove the `approval` part on line 74
- program the `scheduleReenter` so that it calls `token._mint(vault, some amount)` → this is an operation that actually mess up the price.
- do the preview deposit like you do
- do the deposit, and check the the event matched the preview
- do another preview deposit and observe that the price was modified by the reentrancy (but it did not affect the deposit)

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -44,6 +45,81 @@ contract('ERC4626', function (accounts) {
...
+      // Assert prices is kept
+      const sharesAfter = await vault.previewDeposit(amount, { from: holder });
+      expect(sharesBefore).to.be.bignumber.eq(sharesAfter);
+    });
+
+    // During a `_withdraw`, the vault does `_burn(owner, shares)` -> `transfer(receiver, assets)`
+    // such that a reentrancy AFTER the transfer guarantees the price is kept the same.
+    // If the order of burn -> transfer is changed to transfer -> burn, the reentrancy could be triggered on an
+    // intermediate state in which the ratio of shares/assets has been decreased (more assets than shares).
+    it('reenters after withdraw without changing the price', async function () {
```
Amxx, line 96: similar to the other test, I would do it differently

- setup the vault (deposit). Not that you'll either need more user than just holder, of you'll need holder to do a partial withdrawal
- remove the `functionCall` on line 99 and 100
- program the scheduleReenter so that it calls token._burn(vault, some resonable amount) → this is an operation that actually mess up the price.
- do the preview withdraw like you do
- do the withdraw, and check the the event matched the preview
- do another preview withdraw and observe that the price was modified by the reentrancy (but it did not affect the withdraw)

Amxx (commented):
I think these tests are a great addition, but IMO the are not properly designed to check what we want to check. 

I proposed a redesign that would IMO be cleaner/simpler to understand. 

Let me know if you want to discuss that alternative, or if you want me to code it.

test/token/ERC20/extensions/ERC4626.test.js
```diff
@@ -197,7 +197,7 @@ contract('ERC4626', function (accounts) {
       );
 
       // Price before
-      const sharesBefore = await vault.previewWithdraw(amount);
+      const assetsBefore = await vault.previewWithdraw(amount);
```
Amxx, line 199: `Withdraw` (and `previewWithdraw`) take assets as input, and output shares.

In the case of preview, its equivalent to "I want to get X underlying token, how many shares do I need to burn". So this is NOT `assetsBefore`

Also in lines 210 and 211 you are confusing the two.

It happens to works because the rate is 1:1, but it won't it the rates are modified (for example if we run this tests with an offset)

Amxx (approved):

frangio (approved):
Looks good, pretty thorough. I've just renamed the mock since it was not specific to ERC4626.

PR #4199, opened 2023-04-27
Formal verification for ERC20 votes

Amxx:
WIP


Hasinafidahussain:
Ada ap apna easy pisa ya bank ka account number mere 03443616250 per send Karo main apko apko raqam send karta hon

frangio:
Since this is proving challenging we've decided to come back to it in a future iteration.

PR #4200, opened 2023-04-27
Add reentrancy test cases for TimelockController

ernestognw:
Checked that:

- Reentrancy can't re-execute a single operation
- Reentrancy can't re-execute a batched operation

### DIFF
--- a/contracts/mocks/TimelockReentrant.sol
+++ b/contracts/mocks/TimelockReentrant.sol
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../utils/Address.sol";
+import "hardhat/console.sol";
+
+contract TimelockReentrant {
+    address private _reenterTarget;
+    bytes private _reenterData;
+    bool _reentered;
+
+    function disableReentrancy() external {
+        _reentered = true;
+    }
+
+    function enableRentrancy(address target, bytes calldata data) external {
+        _reenterTarget = target;
+        _reenterData = data;
+    }
+
+    function reenter() external {
+        if (!_reentered) {
+            _reentered = true;
+            Address.functionCall(_reenterTarget, _reenterData);
+        }
+    }
+}
--- a/contracts/mocks/TimelockReentrant.sol
+++ b/contracts/mocks/TimelockReentrant.sol
@@ -2,7 +2,6 @@
 pragma solidity ^0.8.0;
 
 import "../utils/Address.sol";
-import "hardhat/console.sol";
 
 contract TimelockReentrant {
     address private _reenterTarget;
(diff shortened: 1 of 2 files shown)


test/governance/TimelockController.test.js
```diff
@@ -632,6 +709,84 @@ contract('TimelockController', function (accounts) {
...
               );
             });
+
+            it('prevents reentrancy execution', async function () {
+              // Create operation
+              const reentrant = await TimelockReentrant.new();
+              const reentrantBatchOperation = genOperationBatch(
+                Array(8).fill(reentrant.address),
+                Array(8).fill(0),
+                Array(8).fill(reentrant.contract.methods.reenter().encodeABI()),
```
frangio, line 719: I think I would use a batch of 1, it's simpler and seems more "likely" to be vulnerable to reentrancy.

```suggestion
                [reentrant.address],
                [0],
                [reentrant.contract.methods.reenter().encodeABI()],
```

frangio (approved):

contracts/mocks/TimelockReentrant.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../utils/Address.sol";
+
+contract TimelockReentrant {
```
Amxx, line 6: This contract is not timelock specific. IMO it should have a generic name like "ReentrancyTester"

Also, it looks like its "just" a contract that exposes a `run` (the one you call `reenter`) function which itself does an arbitrary call. Unless you put the address of this particular contract in the `target` ... this doesn't actually reenter at all.

IMO we should find a better name / design for this helper contract.

ernestognw: I don't think it'll work. If we generalize this contract we'll end up with logic shared across tests without context. Similarly to [this previous comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4177#pullrequestreview-1405421223), it may become difficult to parse a generic "Reentrant" contract considering that we may need different behaviors depending on the case.

For example, sometimes we may need to emit an event in the reentrancy to check it (reentrancy may not fail as in ERC4626), or we may need to catch a specific error (eg. using `Address.functionCall(...)` with fallback error), or maybe checking a value that was changed.

In this particular contract, I used `_reenterData` because I saw the comment about generalizing coming. I'd be okay with any of the following options:

1. Keep it as it is rn
2. Instead of `_reenterData`, use the Timelock interface to make a call directly to both `execute` or `executeBatch`
3.  Making a `Reentrant.mock.sol` with kinds of reentrancy in which the current `TimelockReentrant` will become one of them

Amxx: ok, lets merge that, and I'll propose a more generic helper in another PR at some point.

Amxx: > Instead of _reenterData, use the Timelock interface to make a call directly to both execute or executeBatch
You'll have a fixpoint problem if you try to do that without an helper (the proposal data needs to contain itself)

PR #4206, opened 2023-05-02
uncheck ’+1‘ operation of large integer to save gas

MollyHe5:
Large integers such as uint128, uint256 are unlikely to overflow in the ```+1``` operation. So adding ```unchecked``` to these statements can save much gas.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/governance/compatibility/GovernorCompatibilityBravo.sol
+++ b/contracts/governance/compatibility/GovernorCompatibilityBravo.sol
@@ -150,10 +150,13 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
     ) private pure returns (bytes[] memory) {
         bytes[] memory fullcalldatas = new bytes[](calldatas.length);
 
-        for (uint256 i = 0; i < signatures.length; ++i) {
+        for (uint256 i = 0; i < signatures.length;) {
             fullcalldatas[i] = bytes(signatures[i]).length == 0
                 ? calldatas[i]
                 : abi.encodePacked(bytes4(keccak256(bytes(signatures[i]))), calldatas[i]);
+            unchecked{
+                ++i;
+            }
         }
 
         return fullcalldatas;
--- a/contracts/mocks/InitializableMock.sol
+++ b/contracts/mocks/InitializableMock.sol
@@ -105,7 +105,9 @@ contract ReinitializerMock is Initializable {
     }
 
     function doStuff() public onlyInitializing {
-        counter++;
+        unchecked {
+            counter++;
+        }
     }
 }
 
--- a/contracts/mocks/PausableMock.sol
+++ b/contracts/mocks/PausableMock.sol
@@ -14,7 +14,9 @@ contract PausableMock is Pausable {
     }
 
     function normalProcess() external whenNotPaused {
-        count++;
+        unchecked{
+            count++;
+        }
     }
 
     function drasticMeasure() external whenPaused {
--- a/contracts/mocks/ReentrancyMock.sol
+++ b/contracts/mocks/ReentrancyMock.sol
@@ -38,7 +38,9 @@ contract ReentrancyMock is ReentrancyGuard {
     }
 
     function _count() private {
-        counter += 1;
+        unchecked{
+            counter += 1;
+        }
     }
 
     function guardedCheckEntered() public nonReentrant {
--- a/contracts/mocks/VotesMock.sol
+++ b/contracts/mocks/VotesMock.sol
@@ -21,7 +21,9 @@ abstract contract VotesMock is Votes {
     }
 
     function _mint(address account, uint256 voteId) internal {
-        _balances[account] += 1;
+        unchecked{
+            _balances[account] += 1;
+        }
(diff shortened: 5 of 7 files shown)


contracts/utils/math/Math.sol
```diff
@@ -139,7 +139,9 @@ library Math {
     function mulDiv(uint256 x, uint256 y, uint256 denominator, Rounding rounding) internal pure returns (uint256) {
         uint256 result = mulDiv(x, y, denominator);
         if (rounding == Rounding.Up && mulmod(x, y, denominator) > 0) {
-            result += 1;
+            unchecked{
+                result += 1;
+            }
```
Amxx, line 144: when dealing with a counter that starts from 0 and increases slowly, it is true that an overflow is impossible in practice. It is however not the case here.

Result is not the result of a (slow) incrementation, but the result of a math operation. While i don't have an immediate counter example, I'm really not sure we can guarantee this won't overflow.

contracts/mocks/VotesMock.sol
```diff
@@ -21,7 +21,9 @@ abstract contract VotesMock is Votes {
     }
 
     function _mint(address account, uint256 voteId) internal {
-        _balances[account] += 1;
+        unchecked{
+            _balances[account] += 1;
+        }
```
Amxx, line 26: we don't care about the gas efficiency of mocks. Lets leave them as they are for the sake of readability.

contracts/mocks/ReentrancyMock.sol
```diff
@@ -38,7 +38,9 @@ contract ReentrancyMock is ReentrancyGuard {
     }
 
     function _count() private {
-        counter += 1;
+        unchecked{
+            counter += 1;
+        }
```
Amxx, line 43: we don't care about the gas efficiency of mocks. Lets leave them as they are for the sake of readability.

contracts/mocks/PausableMock.sol
```diff
@@ -14,7 +14,9 @@ contract PausableMock is Pausable {
     }
 
     function normalProcess() external whenNotPaused {
-        count++;
+        unchecked{
+            count++;
+        }
```
Amxx, line 19: we don't care about the gas efficiency of mocks. Lets leave them as they are for the sake of readability.

contracts/mocks/InitializableMock.sol
```diff
@@ -105,7 +105,9 @@ contract ReinitializerMock is Initializable {
     }
 
     function doStuff() public onlyInitializing {
-        counter++;
+        unchecked {
+            counter++;
+        }
```
Amxx, line 110: we don't care about the gas efficiency of mocks. Lets leave them as they are for the sake of readability.

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -154,6 +154,9 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
             fullcalldatas[i] = bytes(signatures[i]).length == 0
                 ? calldatas[i]
                 : abi.encodePacked(bytes4(keccak256(bytes(signatures[i]))), calldatas[i]);
+            unchecked{
+                ++i;
+            }
```
Amxx, line 159: here you added `unchecked { ++i; }` but forgot to remove the `++i` from the for loop. This is really characteristic of the little mistake that you tend to do when trying to (over) optimize ...

Amxx (commented):
Hello @MollyHe5 and thank you for your PR.

As discussed in our [contributing guideliness](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/CONTRIBUTING.md), we usually prefer opening an issue before submitting a PR, so we get a change to discuss the feature that is being proposed.

But since you opened a PR, lets discuss here.

IMO the impact of this changes is really minimal. For example, in the case of the MerkleProof library, it represent a few units of gas per round. We would have to measure that in a realistic usecase, but I expect the change to be minimal.

On the other hand, doing that change does affect the readability of the code. This is true for humans, but it might also be true of code analyzers (I don't have any facts to back that, its just a fear that it will bite back at some point). 

It also easy to make mistake when you go out of your way to do this micro-optimisation (your PR being an example of that). As we value security over anything else, we tend to go in favor of the less "risky" / less "likely to mess up" syntax. 

Ideally, the compiler would optimize that (with our without human help) without moving away from the normal "for" syntax.

contracts/mocks/VotesMock.sol
vincentmatt: Hello, please with all due respect, I have been away from my office for a while now due to my health issues, please what do I have to do immediately

vincentmatt: If you can help me I will be very grateful, 
Few months ago I had an accident and I have a memory lost, it took me more often time than not before I could also remember all

vincentmatt: Mr Amxx

Amxx:
Closing for inactivity + solidity compiler should address that nativelly "soon"

PR #4208, opened 2023-05-03
Disable automatic formal verification workflow on push

Amxx:


frangio (approved):

ernestognw:
We need 2 approvals as per policy.

PR #4212, opened 2023-05-04
Add more test cases for EIP712

frangio:
Improves test coverage of `EIP712` by 1) adding a test case for using the contract behind a proxy, which should affect the domain separator, and 2) considering long strings for name and version to test for correct use of the `ShortStrings` library.

I recommend viewing the diff with "ignore whitespace" on.

I've also added an "Available since" comment that was missing.

Fixes LIB-776

### DIFF
--- a/contracts/utils/cryptography/EIP712.sol
+++ b/contracts/utils/cryptography/EIP712.sol
@@ -111,6 +111,8 @@ abstract contract EIP712 is IERC5267 {
 
     /**
      * @dev See {EIP-5267}.
+     *
+     * _Available since v4.9._
      */
     function eip712Domain()
         public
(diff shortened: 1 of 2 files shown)


Amxx (approved):
LGTM

Its interesting that short name & version make it proxy-compatible... but long name or version breaks that. Its a very specific corner case, and I wonder if anyone will ever run into it.

frangio:
Yes, however, that test should work fine if the upgradeable variant is used and the proxy is properly initialized. Using the non-upgradeable version behind a proxy is an advanced use case so I'd expect the user to look into it carefully.

frangio:
One comment I forgot to make is that for testing chain forks we're blocked by https://github.com/NomicFoundation/hardhat/issues/3074.

PR #4218, opened 2023-05-05
Fix lookup documentation in ERC20Votes and Checkpoints

ToonVanHove:
The comments for `_checkpointsLookup()` in `ERC20Votes.sol` describe that it returns the earliest checkpoint **after** the given `timepoint`. However the function actually returns the latest checkpoint **before** the given `timepoint`

e.g.:
```Solidity
Checkpoint[] checks;

function foo() public returns (uint256){
    checks.push(Checkpoint(90, 10));
    checks.push(Checkpoint(110, 20));
    return _checkpointsLookup(checks, 100); //returns 10
}
```


Amxx (approved):

frangio (approved):
Thank you @ToonVanHove! This was a good find.

PR #4222, opened 2023-05-06
Improve FV specifications for AccessControlDefaultAdminRules

Amxx:
Changeset:
- [x] Rewrite `defaultAdminConsistency` to have a better scope, and make smaller assumptions
- [x] Rewrite `singleDefaultAdmin` to have better coverage / make assumptions
- [x] Reduce the use of require by removing superflous ones.
- [x] Remove the `pendingValueAndScheduleCoupling` rule that is superfluous
  - everything it proves is already proven in a simpler form in the  `beginDefaultAdminTransfer` and `changeDefaultAdminDelay` rules)
  - see discussion bellow
- [x] replace `nonZeroAccount(address)`  with `nonzerosender(env) in `helpers/helpers.spec`
- [x] move `max_uint48` and `min` to `helpers/helpers.spec`
- [x] use definition instead of function to force inlining
- [x] improve `renounceRoleEffect`
- [x] improve `pendingDefaultAdminDelayEnforced` (limitations remain, see comment bellow)
- [x] improve `pendingDelayWaitEnforced` (limitations remain, see comment bellow)


Amxx:
For the `pendingValueAndScheduleCoupling`, here is an example of why the assert it does are superfluous. 

I'm going to focus on 
```
  assert (
    pendingDelayScheduleBefore != pendingDelayScheduleAfter &&
    pendingDelayBefore == pendingDelayAfter
  ) => newDelay == pendingDelayBefore || pendingDelayBefore == 0, "pending delay stays the same if the new delay set is the same";
```
but the same logic applies to the other checks.

First, we need to observer that this rule is in the format `(A && B) → (C || D)`. 

This can be rewritten as: `(C || D) || ! (A && B)`
ie: `C || D || ! A || !B`

One interesting thing to remark here is that, since `B → C` is `C || !B`, we know that `B → C` is stronger than `(A && B) → (C || D)`. 
Said differently, if we have `B → C`, we automatically have `(A && B) → (C || D)` (for any `A` and `D`)


This means that if we have 
```(pendingDelayBefore == pendingDelayAfter) → newDelay == pendingDelayBefore```
than the above rule is an immediate consequence.

Know we can have a look at the `rule changeDefaultAdminDelay`. This rule says that after a call to `changeDefaultAdminDelay`:
```assert success => pendingDelay_(e) == newDelay```

Know that we know that after a sucessfull call `pendingDelayAfter == newDelay`, it is clear that is `pendingDelayBefore == pendingDelayAfter` than `newDelay == pendingDelayBefore` (by transitivity of the equality).

---

TLDR: the asserts in `pendingValueAndScheduleCoupling` are overly complex, and prove "less" than what is already proven in the `beginDefaultAdminTransfer` and `changeDefaultAdminDelay` rules. We can remove `pendingValueAndScheduleCoupling` without losing coverage.

Amxx:
About `pendingDefaultAdminDelayEnforced`: 

Technically, we would want any number of function being called between  e1 and e2. Currently the rule only checks:

> after a `beginDefaultAdminTransfer` call, a single function can only change the admin the call happens after `delayBefore`.

What we would want to check is that 

> after a `beginDefaultAdminTransfer` call, any sequence of calls that result in the admin changing must take at least `delayBefore`.

---

From the other rules we know that 
- admin can only change through `acceptDefaultAdminTransfer` or `renounceRole` (`rule noDefaultAdminChange`)
- `acceptDefaultAdminTransfer` can only be called after the schedule as passed (`rule acceptDefaultAdminTransfer` liveness)
- `renounceRole` can only be called after the schedule as passed (if role is 0) (`rule renounceRoleEffect` liveness)
- the schedule can only be modified by `beginDefaultAdminTransfer`, `acceptDefaultAdminTransfer` or `cancelDefaultAdminTransfer` (`rule noPendingDefaultAdminChange`)
- `acceptDefaultAdminTransfer` sets the schedule to 0 (`rule acceptDefaultAdminTransfer`)
- `cancelDefaultAdminTransfer` sets the schedule to 0 (`rule cancelDefaultAdminTransfer`)
- `beginDefaultAdminTransfer` sets the delay further enough in the future (`rule beginDefaultAdminTransfer`)

IMO that is good if on top you add the rules about changing the delay... 
Its hard to "compile" because the proof is spread between different rules, but the result is stronger.

We are missing the part where `renounceRole` only changes the admin if called with the correct role though

PR #4223, opened 2023-05-07
 Improve FV specifications for AccessControlDefaultAdminRules 

Amxx:
Replaces #4222  (change for the FV workflow to run)

--- 

**Changeset**
- [x] Rewrite `defaultAdminConsistency` to have a better scope, and make smaller assumptions
- [x] Rewrite `singleDefaultAdmin` to have better coverage / make assumptions
- [x] Reduce the use of require by removing superflous ones.
- [x] Remove the `pendingValueAndScheduleCoupling` rule that is superfluous
  - everything it proves is already proven in a simpler form in the  `beginDefaultAdminTransfer` and `changeDefaultAdminDelay` rules)
  - see discussion bellow
- [x] replace `nonZeroAccount(address)`  with `nonzerosender(env) in `helpers/helpers.spec`
- [x] move `max_uint48` and `min` to `helpers/helpers.spec`
- [x] use definition instead of function to force inlining
- [x] improve `renounceRoleEffect`
- [x] improve `pendingDefaultAdminDelayEnforced` (limitations remain, see comment bellow)
- [x] improve `pendingDelayWaitEnforced` (limitations remain, see comment bellow)


Amxx:
**copy from https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4222#issuecomment-1537211740**

---

For the `pendingValueAndScheduleCoupling`, here is an example of why the assert it does are superfluous. 

I'm going to focus on 
```
  assert (
    pendingDelayScheduleBefore != pendingDelayScheduleAfter &&
    pendingDelayBefore == pendingDelayAfter
  ) => newDelay == pendingDelayBefore || pendingDelayBefore == 0, "pending delay stays the same if the new delay set is the same";
```
but the same logic applies to the other checks.

First, we need to observer that this rule is in the format `(A && B) → (C || D)`. 

This can be rewritten as: `(C || D) || ! (A && B)`
ie: `C || D || ! A || !B`

One interesting thing to remark here is that, since `B → C` is `C || !B`, we know that `B → C` is stronger than `(A && B) → (C || D)`. 
Said differently, if we have `B → C`, we automatically have `(A && B) → (C || D)` (for any `A` and `D`)


This means that if we have 
```
(pendingDelayBefore == pendingDelayAfter) → newDelay == pendingDelayBefore
```
than the above rule is an immediate consequence.

If we look the `rule changeDefaultAdminDelay`, we can see that after a call to `changeDefaultAdminDelay`:
```
assert success => pendingDelay_(e) == newDelay
```

Now that we know that after a sucessfull call `pendingDelayAfter == newDelay`, it is clear that if `pendingDelayBefore == pendingDelayAfter` than `newDelay == pendingDelayBefore` (by transitivity of the equality).


It is now clear that `changeDefaultAdminDelay` proves 
```
(pendingDelayBefore == pendingDelayAfter) → newDelay == pendingDelayBefore
```
and as a consequence also proves
```
  assert (
    pendingDelayScheduleBefore != pendingDelayScheduleAfter &&
    pendingDelayBefore == pendingDelayAfter
  ) => newDelay == pendingDelayBefore || pendingDelayBefore == 0;
```
making (this check of) `pendingValueAndScheduleCoupling` superfluous.

---

TLDR: the asserts in `pendingValueAndScheduleCoupling` are overly complex, and prove "less" than what is already proven in the `beginDefaultAdminTransfer` and `changeDefaultAdminDelay` rules. We can remove `pendingValueAndScheduleCoupling` without losing coverage.

Amxx:
**copy from https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4222#issuecomment-1537441793**

---

About `pendingDefaultAdminDelayEnforced` (logic also applies to `pendingDelayWaitEnforced`):

Technically, we would want any number of function being called between  e1 and e2. Currently the rule only checks:

> after a `beginDefaultAdminTransfer` call, a single function can only change the admin the call happens after `delayBefore`.

What we would want to check is that 

> after a `beginDefaultAdminTransfer` call, any sequence of calls that result in the admin changing must take at least `delayBefore`.

---

From the other rules we know that 
- admin can only change through `acceptDefaultAdminTransfer` or `renounceRole` (`rule noDefaultAdminChange`)
- `acceptDefaultAdminTransfer` can only be called after the schedule as passed (`rule acceptDefaultAdminTransfer` liveness)
- `renounceRole` can only be called after the schedule as passed (if role is 0) (`rule renounceRoleEffect` liveness)
- the schedule can only be modified by `beginDefaultAdminTransfer`, `acceptDefaultAdminTransfer` or `cancelDefaultAdminTransfer` (`rule noPendingDefaultAdminChange`)
- `acceptDefaultAdminTransfer` sets the schedule to 0 (`rule acceptDefaultAdminTransfer`)
- `cancelDefaultAdminTransfer` sets the schedule to 0 (`rule cancelDefaultAdminTransfer`)
- `beginDefaultAdminTransfer` sets the delay further enough in the future (`rule beginDefaultAdminTransfer`)

IMO that is good if on top you add the rules about changing the delay... 
Its hard to "compile" because the proof is spread between different rules, but the result is stronger.

We are missing the part where `renounceRole` only changes the admin if called with the correct role though (added to this PR)

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -309,17 +317,19 @@ rule pendingDefaultAdminDelayEnforced(env e1, env e2, method f, calldataarg args
 */
 rule acceptDefaultAdminTransfer(env e) {
   require nonpayable(e);
-  requireInvariant defaultAdminConsistency(defaultAdmin());
-  requireInvariant singleDefaultAdmin(e.msg.sender, defaultAdmin());
```
ernestognw, line 313: Have you run this rule? I think it's required because otherwise, the `e.msg.sender` may have the `DEFAULT_ADMIN_ROLE` without being the `defaultAdmin()`, otherwise the `acceptDefaultAdminTransfer` call will succeed without being called by the `pendingAdminBefore`

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -338,8 +348,8 @@ rule acceptDefaultAdminTransfer(env e) {
 */
 rule cancelDefaultAdminTransfer(env e) {
   require nonpayable(e);
-  requireInvariant defaultAdminConsistency(defaultAdmin());
-  requireInvariant singleDefaultAdmin(e.msg.sender, defaultAdmin());
+  require nonzerosender(e);
```
ernestognw, line 351: Why is this needed? Even if the zero address is the sender and the `defaultAdmin()`, I think it's something theoretically allowed

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -392,16 +404,21 @@ rule changeDefaultAdminDelay(env e, uint48 newDelay) {
...
   uint48 delayAfter = defaultAdminDelay(e2);
 
   mathint delayWait = newDelay > delayBefore ? increasingDelaySchedule(e1, newDelay) : decreasingDelaySchedule(e1, newDelay);
 
-  assert delayAfter == newDelay => (e2.block.timestamp >= delayWait || delayBefore == newDelay),
+  assert delayAfter != delayBefore => (
+    delayAfter == newDelay &&
+    e2.block.timestamp >= delayWait
+  ),
     "A delay can't change in less than applied schedule";
```
ernestognw, line 422:
```suggestion
    "A delay can only change after the applied schedule";
```

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -153,19 +138,39 @@ rule renounceRoleEffect(env e, bytes32 role) {
...
+      "role is renounced";
+
+    assert success => (
+      (
+        role    == DEFAULT_ADMIN_ROLE() &&
+        account == adminBefore
+      ) ? (
+        adminAfter        == 0 &&
+        pendingAdminAfter == 0 &&
+        scheduleAfter     == scheduleBefore // For some reason this is not reset. needs fix ?
```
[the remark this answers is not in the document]
Amxx: The other way around, we need to change this if we merge #4230

certora/specs/AccessControlDefaultAdminRules.spec
Amxx: Lets imagine both `msg.sender` and the `defaultAdmin()` are 0. We are in a case where the admin was renounced. In that case, even if `defaultAdmin()` is 0, `hasRole(DEFAULT_ADMIN_ROLE, address(0))` is false.

So the call to `changeDefaultAdminDelay` will fail because `onlyRole(DEFAULT_ADMIN_ROLE)` will not pass. This causes the liveness check to fail.

Either we do 
```
require nonzerosender(e)
```
or we change the liveness check to
```
assert success <=> (e.msg.sender == defaultAdmin() && e.msg.sender != 0),
    "only the current default admin can cancel a transfer";
```

I think the first option is simpler

ernestognw: Got it, thanks for the explanation!

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: I see. I assume then the prover is not considering the initial case of multiple default admins, which it does when specifying an `invariant`. Do you think that's the reason? I just don't want to assume that if it passes it's okay, I don't completely see why it passes.

Amxx: The `acceptDefaultAdminTransfer` is restricted to the `pendingDefaultAdmin`, not the `defaultAdmin`. It doesn't include any role check.

Then the `_acceptDefaultAdminTransfer` does :
- the schedule check
- `_revokeRole` 
  - if there is an inconsistency (default admin doesn't have the role) then this is a no-op that passes. Only thing we know for sure is the (old) defaultAdmin() doesn't have the role after
- `_grantRole`
  - if there is an inconsistency (pending default admin has the role) then this is a no-op that passes. Only thing we know for sure is the new defaultAdmin() has the role after
  
I don't see which part of the function call or the rule needs the defaultAdmin being consistent.

ernestognw: My bad, I mistaken the check of the `pendingDefaultAdmin` by the `defaultAdmin` 👍🏻

ernestognw (approved):
LGTM, but considering https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4230 needs discussion, we'll need to open a PR removing the comment in `renounceRoleEffect `

Amxx:
Updated to reflect the last changes to ACDAR (that how I'm calling AccessControlDefaultAdminRules for short)

certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -209,7 +209,8 @@ rule noPendingDefaultAdminChange(env e, method f, calldataarg args) {
   ) => (
     f.selector == beginDefaultAdminTransfer(address).selector ||
     f.selector == acceptDefaultAdminTransfer().selector ||
-    f.selector == cancelDefaultAdminTransfer().selector
+    f.selector == cancelDefaultAdminTransfer().selector ||
+    f.selector == renounceRole(bytes32,address).selector
```
ernestognw, line 213: As far as I can tell, `renounceRole(bytes32,address)` doesn't affect the pending admin, because it makes it 0 anyway. Was it failing?

```suggestion
f.selector == cancelDefaultAdminTransfer().selector
```

ernestognw: If we're keeping, we need to change the title of the rule and the error message for this assert to include renounce

ernestognw (changes requested):

certora/specs/AccessControlDefaultAdminRules.spec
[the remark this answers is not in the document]
ernestognw: Solving since #4230 was merged and this comment was addressed

certora/specs/AccessControlDefaultAdminRules.spec
frangio: Updated docs and message.

frangio (approved):

ernestognw (commented):
This [comment hasn't been solved](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4223#discussion_r1192535734), but I'm approving to unblock it under the assumption that if we remove the `renounceRole` it should still work.

ernestognw (commented):
This [comment hasn't been solved](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4223#discussion_r1192535734), but I'm approving to unblock it under the assumption that if we remove the `renounceRole` it should still work.

ernestognw (approved):
This [comment hasn't been solved](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4223#discussion_r1192535734), but I'm approving to unblock it under the assumption that if we remove the `renounceRole` it should still work.

certora/specs/AccessControlDefaultAdminRules.spec
ernestognw: Answered in Slack, the reason we need this is because #4230, the `renounceRole`affects now the pending admin schedule, so the alternative is to have two rules (one using `renounceRole`, and the other not):

```
assert pendingAdminBefore != pendingAdminAfter => ( ... );
assert scheduleBefore != scheduleAfter => ( ... );
```

But we considered that unnecessary. Solving

PR #4224, opened 2023-05-09
Implement extra suggestions from audit review of 4.9

ernestognw:
Audit report to be published soon.

# Requirements

Using audit report codes

- [ ] N-01: Added docstrings for `Trace224` and `Trace160` and comments for ShortStrings
- [ ] N-02: Added note for deprecation with a recommendation
- [x] N-03: Couldn't be addressed since the functions access `msg.value`, can't be view
- [ ] N-04: Changed `string(abi.encodePacked(...))` for `string.concat(...)`

### DIFF
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -42,7 +42,7 @@ library Strings {
      * @dev Converts a `int256` to its ASCII `string` decimal representation.
      */
     function toString(int256 value) internal pure returns (string memory) {
-        return string(abi.encodePacked(value < 0 ? "-" : "", toString(SignedMath.abs(value))));
+        return string.concat(value < 0 ? "-" : "", toString(SignedMath.abs(value)));
     }
 
     /**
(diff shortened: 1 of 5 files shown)


contracts/utils/Strings.sol
```diff
@@ -42,7 +42,7 @@ library Strings {
      * @dev Converts a `int256` to its ASCII `string` decimal representation.
      */
     function toString(int256 value) internal pure returns (string memory) {
-        return string(abi.encodePacked(value < 0 ? "-" : "", toString(SignedMath.abs(value))));
+        return string.concat(value < 0 ? "-" : "", toString(SignedMath.abs(value)));
```
Amxx, line 45: This is something we have planed for 5.0 but that we decided not to do so far for a simple reason:

`bytes.concat` was introduced in 0.8.4, and `string.concat` was properly introduced in 0.8.12. If we do this change, we should change the pragma for this file from `^0.8.0` to `^0.8.12` ... which is a breaking change.

Amxx (commented):
I believe we should not fix N-04 (see the discussion).

Everything else looks good to me.

contracts/utils/Strings.sol
frangio: Agree with this, let's remove this change and add it to the list of new Solidity features to use in 5.x.

ernestognw: Thanks for flagging this, I had no idea. I just reverted that particular change and I'll circle this back with the audit team

frangio (approved):
LGTM

Amxx (approved):
LGTM

PR #4225, opened 2023-05-09
Diff view between Amxx:next-v5.0 and master 

Amxx:
Fixes: LIB-858

This PR includes:
- work from previous PRs
  - that targeted the `next-v5.0` branch
  - that were reviewed 
- code added when resolving the conflicts between `master` and `next-v5.0`

The second part is what we need to review. Unfortunately, it has to be done by a throughout review of everything in this diff :/


.github/workflows/checks.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - next-v*
```
Amxx, line 7: Should we keep that for `next-v6`, or revert it ?

frangio: I'm good with keeping this.

frangio:
After we merge this, what is the plan for any potential 4.x patches, be it for the 4.9 release candidate or even after that? Would we use the `release-v4.9` branch for all future 4.x patches from then on?

Amxx:
> Would we use the `release-v4.9` branch for all future 4.x patches from then on?

I was thinking either that or we branch of something (`master-v4.9` / `develop-v4.9` ?) out of the release branch, and then we cherry-pick from the new branch to the release branch

docs/modules/ROOT/pages/erc20-supply.adoc
```diff
@@ -54,40 +54,12 @@ contract ERC20WithMinerReward is ERC20 {
...
-
-TIP: To learn more about roles and permissioned systems, head to our xref:access-control.adoc[Access Control guide].
-
 [[automating-the-reward]]
 == Automating the Reward
 
-So far our supply mechanisms were triggered manually, but `ERC20` also allows us to extend the core functionality of the token through the xref:api:token/ERC20.adoc#ERC20-_beforeTokenTransfer-address-address-uint256-[`_beforeTokenTransfer`] hook (see xref:extending-contracts.adoc#using-hooks[Using Hooks]).
+So far our supply mechanism was triggered manually, but `ERC20` also allows us to extend the core functionality of the token through the xref:api:token/ERC20.adoc#ERC20-_beforeTokenTransfer-address-address-uint256-[`_beforeTokenTransfer`] hook (see xref:extending-contracts.adoc#using-hooks[Using Hooks]).
```
Amxx, line 60: Need to replace `_beforeTokenTransfer` with `_update`.

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,184 +1,182 @@
...
 
-const { getChainId } = require('../../../helpers/chainid');
+const { clock, clockFromReceipt } = require('../../../helpers/time');
 
 const { shouldBehaveLikeVotes } = require('../../../governance/utils/Votes.behavior');
 
-const ERC721Votes = artifacts.require('$ERC721Votes');
+const MODES = {
+  blocknumber: artifacts.require('$ERC721Votes'),
+  // no timestamp mode for ERC20VotesComp yet
```
Amxx, line 12:
```suggestion
  // no timestamp mode for ERC721Votes yet
```

Do we want to add one ?

Amxx:
`TimelockControllerWith46Migration`, that is part of `upgradeable.patch` causes an error (it uses `TIMELOCK_ADMIN_ROLE` that was removed).

IMO we should remove it from the patch.

scripts/generate/templates/Checkpoints.js
```diff
@@ -133,13 +133,6 @@ function getAtProbablyRecentBlock(${opts.historyTypeName} storage self, uint256
     return pos == 0 ? 0 : _unsafeAccess(self.${opts.checkpointFieldName}, pos - 1).${opts.valueFieldName};
 }
 
-/**
- * @dev Returns checkpoint at given position.
- */
-function getAtPosition(History storage self, uint32 pos) internal view returns (Checkpoint memory) {
-    return self._checkpoints[pos];
-}
-
```
ernestognw: Why is this removed?

ernestognw (dismissed):
Didn't fully look into the tests and focused only on the conflicts I got by merging locally. LGTM

frangio:
A comment that is independent of the specific changes in the diffs.

I think it's important to preserve the commit history of `next-v5.0`. Rather than having a big squashed commit in `master` with all the changes, we need an actual merge commit. If we agree on that, I believe we need to change the approach that this PR is assuming.

Instead of merging `next-v5.0` into `master`, we should think of first merging current `master` into `next-v5`. Later we will merge `next-v5.0` into `master` after all conflicts have been resolved, ideally it's a simple fast-forward.

This also means we need to keep the commit history of `next-v5.0` tidy. I believe @Amxx may have pushed some commits to this repo's [`next-v5.0`](https://github.com/OpenZeppelin/openzeppelin-contracts/commits/next-v5.0) that should have gone to his fork, @Amxx can you look into that? We may have to roll back `next-v5.0`, and I'd suggest working on a separate branch to avoid that confusion.

After that, this PR should be retargeted towards `next-v5.0`, and ideally there would be a single merge commit plus one or a few other commits making further adjustments.

Sorry this might be confusing... Let me know what you think.

contracts/governance/utils/Votes.sol
```diff
@@ -28,23 +28,16 @@ import "../../utils/cryptography/EIP712.sol";
...
+abstract contract Votes is Context, EIP712, Nonces, IERC5805 {
     using Checkpoints for Checkpoints.Trace224;
-    using Counters for Counters.Counter;
 
     bytes32 private constant _DELEGATION_TYPEHASH =
         keccak256("Delegation(address delegatee,uint256 nonce,uint256 expiry)");
 
     mapping(address => address) private _delegation;
-
-    /// @custom:oz-retyped-from mapping(address => Checkpoints.History)
```
frangio, line 40: As far as I can tell this was not due to conflict resolution. I'm guessing you're removing these annotations because they're no longer needed for 5.0?

I would've preferred to have a clean separation between what is merge conflict resolution and what is further 5.0 work. I wouldn't go as far as to say we should revert this but let's please not add any more "new" things in this PR.

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -24,267 +22,31 @@ import "../../../utils/cryptography/ECDSA.sol";
...
-    function CLOCK_MODE() public view virtual override returns (string memory) {
-        // Check that the clock was not modified
-        require(clock() == block.number, "ERC20Votes: broken clock mode");
-        return "mode=blocknumber&from=default";
-    }
-
-    /**
-     * @dev Get the `pos`-th checkpoint for `account`.
-     */
-    function checkpoints(address account, uint32 pos) public view virtual returns (Checkpoint memory) {
```
frangio, line 60: This function was present in `next-v5.0` prior to the recent merges.

frangio:
I left some comments above but they are not exhaustive. I want us to first revert `next-v5.0` to a state that has been reviewed, as far as I can tell that should be 2d05db17, and then retarget this PR at `next-v5.0` so we can see more clearly what is being changed. For example, it would be clear then that `ERC20Votes.checkpoints` was removed at some point and I don't think that was reviewed. It's been difficult to make sense of the merges so I may be mistaken but that's my current understanding.

scripts/generate/templates/Checkpoints.js
Amxx: My understanding is it was added, not used, and then removed. If you look at the overall PR diff, this file (and the resulting solidity) are not touched

Amxx:
I reset `upstream:next-v5.0` to https://github.com/OpenZeppelin/openzeppelin-contracts/commit/2d05db171a99e93132abef98bda15534129f7805, and changed the target for this branch 

@frangio

contracts/token/ERC1155/ERC1155.sol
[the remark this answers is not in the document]
Amxx: In master, `ERC1155.sol` uses `^0.8.0`.

This pragma change was done in `https://github.com/OpenZeppelin/openzeppelin-contracts/commit/c5d040beb9a951b00e9cb57c4e7dd97cd04b45ac`, presumably to add support for the new `address.code.length` syntax.

I initially thought that 0.8.0 would be ok here, because the `address.code.length` syntax is available in 0.8.0. Now I realize that while its not strictly speaking necessary, using 0.8.1 forces use of `extcodesize` and save gas.

contracts/token/ERC20/extensions/ERC20Votes.sol
Amxx: The conflict that cause its removal is the following:

- `Checkpoint` is now available as a member of `Checkpoint.History`
- `Checkpoint.History` is (unofficially) deprecated in favor of Trace224, because the "vote with timestamp" needs to control the values used for checkpoints (history forces blockNumber)
- In Votes we use Trace224, so in an ERC20Votes that uses Votes, the type is not available

Here is a commit that re implements that, but that returns a Checkpoint224 (its the same layout, but with different fields name): https://github.com/OpenZeppelin/openzeppelin-contracts/commit/73cf027231de996c8583c002a50dd1ad7b782af7

Note that both `checkpoints(address,uint32)` and `numCheckpoints(address)` are leftover from Compound, are not available in ERC721Votes, and are AFAIK only used for testing. I'd like to discuss removing them in 5.0

scripts/generate/templates/Checkpoints.js
Amxx: re-added

PR #4230, opened 2023-05-10
Clean up pending admin schedule on renounce in DefaultAdminRules

frangio:
In `AccessControlDefaultAdminRules` we realized that we are not cleaning up the schedule value in storage after renouncing. This doesn't seem to have any consequences other than misrepresenting the state of the contract through one of its getters.

I'm also including an unrelated change in this PR but I noticed two of the descriptions in the test suite were not using the right wording (delay vs schedule).

Fixes LIB-862

### DIFF
--- a/contracts/access/AccessControlDefaultAdminRules.sol
+++ b/contracts/access/AccessControlDefaultAdminRules.sol
@@ -112,6 +112,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
                 newDefaultAdmin == address(0) && _isScheduleSet(schedule) && _hasSchedulePassed(schedule),
                 "AccessControl: only can renounce in two delayed steps"
             );
+            delete _pendingDefaultAdminSchedule;
         }
         super.renounceRole(role, account);
     }
--- a/.changeset/loud-wolves-promise.md
+++ b/.changeset/loud-wolves-promise.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`AccessControlDefaultAdminRules`: Clean up pending admin schedule on renounce.
--- a/test/access/AccessControl.behavior.js
+++ b/test/access/AccessControl.behavior.js
@@ -267,7 +267,7 @@ function shouldBehaveLikeAccessControlDefaultAdminRules(errorPrefix, delay, defa
         [0, 'exactly when'],
         [1, 'after'],
       ]) {
-        it(`returns pending admin and delay ${tag} delay schedule passes if not accepted`, async function () {
+        it(`returns pending admin and schedule ${tag} it passes if not accepted`, async function () {
           // Wait until schedule + fromSchedule
           const { schedule: firstSchedule } = await this.accessControl.pendingDefaultAdmin();
           await time.setNextBlockTimestamp(firstSchedule.toNumber() + fromSchedule);
@@ -279,7 +279,7 @@ function shouldBehaveLikeAccessControlDefaultAdminRules(errorPrefix, delay, defa
         });
       }
 
-      it('returns 0 after delay schedule passes and the transfer was accepted', async function () {
+      it('returns 0 after schedule passes and the transfer was accepted', async function () {
         // Wait after schedule
         const { schedule: firstSchedule } = await this.accessControl.pendingDefaultAdmin();
         await time.setNextBlockTimestamp(firstSchedule.addn(1));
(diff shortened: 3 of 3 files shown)


frangio:
How should we handle this in terms of changeset? :thinking: It sounds like we need a changeset for the release automation to create a new release candidate for us, but this would be a part of the existing changeset for AccessControlDefaultAdminRules.

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -112,6 +112,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
                 newDefaultAdmin == address(0) && _isScheduleSet(schedule) && _hasSchedulePassed(schedule),
                 "AccessControl: only can renounce in two delayed steps"
             );
+            delete _pendingDefaultAdminSchedule;
```
Amxx, line 115: Small comment:

If a dev calls `_revokeRole` internally, the schedule is not reset. This means that if someone does a `beginTransferAdmin` and then a `_revokeRole`, the contract appears to be administered by `address(0)`, but the pending owner can still claim the ownership.

IMO this delete should go (in addition to `delete _pendingDefaultAdmin`) in `_revokeRole`.

```solidity
        if (role == DEFAULT_ADMIN_ROLE && account == _currentDefaultAdmin) {
            delete _currentDefaultAdmin;
        }
    }
```

Amxx: I tested with 
```solidity
        if (role == DEFAULT_ADMIN_ROLE && account == _currentDefaultAdmin) {
            delete _currentDefaultAdmin;
            delete _pendingDefaultAdmin;
            delete _pendingDefaultAdminSchedule;
        }
    }
```
and the tests pass.

Amxx: (I'm increasingly worried by the number of corner case that this contract has, and that we are discovering quite late IMO)

ernestognw:
> How should we handle this in terms of changeset? 🤔 It sounds like we need a changeset for the release automation to create a new release candidate for us, but this would be a part of the existing changeset for AccessControlDefaultAdminRules.

Easiest is to add a changeset as a placeholder. We'll need to review and edit each changesets PR opened when cherry-picking this.

A more complex solution could be configuring the release cycle to ignore certain files when converting to the `Changelog`, so we can add some keyword like `@changesets:ignore`

In any of both cases, the Changelog processing will add an empty entry for `4.9.0-rc.1`, which will also be extracted into the GitHub Release page.

ernestognw (dismissed):
LGTM assuming no changes with the `_revokeRole`. I think that requires more discussion so far.

contracts/access/AccessControlDefaultAdminRules.sol
Amxx: > For example, a developer might want to extend the contract in a way in which a multisig has the ability to disable a DEFAULT_ADMIN before a new one enters into effect. That's a valid use case for using _revokeRole while keeping the already scheduled one.

That is something I raised on slack as being a possible attack: 
- you initiate a default admin transfer
- you use the internal `_revoke`. This emits the revoke event, and set the current admin to 0
  - at the point users will think the contract has no admin, and will not get a new one
- the pending admin accepts, and "resurrect" the admin rights
  - this goes against the expectation (by the user) that the admin was removed from the system.
  
If someone reported that on Immunefi, would we say its the expected behavior. As a user wouldn't expect this behavior.

I personally like the fact that my proposal makes `_acceptDefaultAdminTransfer` "cleaner"
- we cache the pending admin
- we check the schedule
- we use revoke to clear old state
- we use grant to set the new one

Reseting the state in revoke is clearly opinionated. I think what it comes down to is whether we consider the 
```
beginDefaultAdminTransfer → _revoke → acceptDefaultAdminTransfer
```
workflow being dangerous or not.

frangio: Yeah I also don't see a reason why `_revokeRole` should reset the schedule.

> if someone does a `beginTransferAdmin` and then a `_revokeRole`, the contract appears to be administered by `address(0)`, but the pending owner can still claim the ownership.

This is an interesting observation but I don't think it's specific to this situation with revoke. In general, anyone that wants to check the permissions should always also check if there's a pending/scheduled admin. For example, the same strategy could be used to mislead someone into thinking that account A is the admin when in fact account B is a pending admin that can claim at any moment.

Amxx (approved):
Lets merge that, and open an issue to continue the discussion

contracts/access/AccessControlDefaultAdminRules.sol
Amxx: > I agree with this statement. I'd be more in favor of an internal _renounceAdmin that does both the _revoke and the schedule reset, but _revoke resetting the schedule on its own sounds to me like a misleading assumption

Fair!

frangio:
Had to add a changeset first. Please re-approve.

ernestognw (approved):

Amxx (approved):

PR #4235, opened 2023-05-10
Add final PDF report for v4.9 audit

ernestognw:
Following up with #4227, this adds the final report with missing resolutions for the first notes, whose fixes were applied in #4224


frangio (changes requested):
The PDF seems broken, is it just me?

ernestognw:
> The PDF seems broken, is it just me?

No no, you're right, thanks for double checking! I dragged and dropped into the folder but something went off during the process. I just did the same again and now it looks fine

frangio (approved):
Yeah that looks good now. Thanks!

PR #4238, opened 2023-05-10
Fix bug allowing anyone to cancel an admin renounce

Amxx:
Changes introduced in #4230 include a bug allowing anyone to reset the schedule of an admin renounce:

- Alice (admin) starts a transfer to address(0)
- Delay passes
- Bob (not admin) does `renounceRole(0, bob)
  - `if (role == DEFAULT_ADMIN_ROLE)` passes
    - `newDefaultAdmin == address(0)` is true
    - schedule is good
      - `delete _pendingDefaultAdminSchedule`
  - super.renounceRole
    - caller is the account (bob)
    - caller doesn't have the role: nothing happens (no-op does not revert)

So Bob was able to `delete _pendingDefaultAdminSchedule`, which shouldn't have been possible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,4 +69,3 @@ contracts-exposed
 .last_confs
 certora_*
 .zip-output-url.txt
-.worktree
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,4 @@ contracts-exposed
 .last_confs
 certora_*
 .zip-output-url.txt
+.worktree
--- a/contracts/access/AccessControlDefaultAdminRules.sol
+++ b/contracts/access/AccessControlDefaultAdminRules.sol
@@ -106,7 +106,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * non-administrated role.
      */
     function renounceRole(bytes32 role, address account) public virtual override(AccessControl, IAccessControl) {
-        if (role == DEFAULT_ADMIN_ROLE) {
+        if (role == DEFAULT_ADMIN_ROLE && account == defaultAdmin()) {
             (address newDefaultAdmin, uint48 schedule) = pendingDefaultAdmin();
             require(
                 newDefaultAdmin == address(0) && _isScheduleSet(schedule) && _hasSchedulePassed(schedule),
@@ -138,7 +138,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @dev See {AccessControl-_revokeRole}.
      */
     function _revokeRole(bytes32 role, address account) internal virtual override {
-        if (role == DEFAULT_ADMIN_ROLE && account == _currentDefaultAdmin) {
+        if (role == DEFAULT_ADMIN_ROLE && account == defaultAdmin()) {
             delete _currentDefaultAdmin;
         }
         super._revokeRole(role, account);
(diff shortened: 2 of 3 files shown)


frangio:
We should have an additional unit test for this scenario.

frangio (approved):

test/access/AccessControl.behavior.js
```diff
@@ -639,6 +642,15 @@ function shouldBehaveLikeAccessControlDefaultAdminRules(errorPrefix, delay, defa
       );
     });
 
+    it('no op if renouncing when not having the role', async function () {
```
ernestognw, line 645: Small NIT:

```suggestion
    it('no-ops if renouncing when not having the role', async function () {
```

I personally like to read the tests as `it('<verb> ...` (eg. `it('does something ...')`)

contracts/access/AccessControlDefaultAdminRules.sol
```diff
@@ -138,7 +138,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @dev See {AccessControl-_revokeRole}.
      */
     function _revokeRole(bytes32 role, address account) internal virtual override {
-        if (role == DEFAULT_ADMIN_ROLE && account == _currentDefaultAdmin) {
```
ernestognw, line 141: My original rationale for using `_currentDefaultAdmin` instead of `defaultAdmin()` is that `_revokeRole()` is internal and thus, it's less obvious that an override will affect it. I think also @frangio flagged something similar in the original PR (#4009).

I'd revert this one unless we have a very good reason for changing it. The other one is good under my original rationale because `renounceRole` is public.

frangio: >it's less obvious that an override will affect it

Do you mean that for the developer that overrides `defaultAdmin` it is less obvious that the override will affect `_revokeRole`? I'm not convinced by this argument, I think the consistency is more valuable.

ernestognw (approved):
LGTM

contracts/access/AccessControlDefaultAdminRules.sol
frangio: >it may be excessive using the public methods everywhere

Yes, I recall saying this, it was not in this line. I still feel that way... But the solution that I see is to remove virtual, as long as we have virtual we have to more or less use the overrideable functions.

frangio (approved):

frangio (approved):
Note that the test description was updated but is still not following Ernesto's suggestion. But IMO we should come back to this later, there are many tests throughout that are following a different format.

ernestognw:
> Note that the test description was updated but is still not following Ernesto's suggestion. But IMO we should come back to this later, there are many tests throughout that are following a different format.

Agree, it's a minor readability improvement but a suggestion as well, not mandatory. Will try to enforce this in my reviews but don't feel we need to update every test written the same way

.gitignore
```diff
@@ -69,3 +69,4 @@ contracts-exposed
 .last_confs
 certora_*
 .zip-output-url.txt
+.worktree
```
ernestognw, line 72: We need to add a comment for this, every other ignored file is in its own category block and this one doesn't belong to `# Certora`

ernestognw (changes requested):

.gitignore
frangio: That was me, let me remove this.

ernestognw (approved):

ernestognw (approved):

ernestognw (approved):

PR #4239, opened 2023-05-11
Prevent attempt to publish to npm

frangio:
Commits in the release branches when there are no new changesets will currently result in an attempt to publish to npm, even though the version has already been published and it would fail. This should prevent that error.


Amxx (dismissed):

ernestognw (approved):
LGTM, let me know if you need a reapproval if we're changing to `undici`

ernestognw (approved):

PR #4241, opened 2023-05-11
Add a bool return to _grantRole and _revokeRole

Amxx:
Similarly to `EnumerableSet`'s `add` and `remove`, I believe `AccessControl`'s `_grantRole` and `_revokeRole` should return a boolean value. This would help overrides version handle no-ops.

I believe the public versions should not return the boolean, but that is open to discussion.

#### PR Checklist

- [x] Tests
- [ ] Documentation


frangio:
This would need to be rebased. However, I would also like us to consider the alternative of making the functions revert in the cases where they would return false, as an alternative to adding a return value.

The argument so far has been that we want to allow no-ops, but we've seen how in this case it can actually lead to bugs when you override the function, and you very easily assume that there is an actual revoke ongoing. I think this is stronger for revoke than it is for grant.

Amxx:
@frangio @ernestognw 

This got lost, I think we should do it in 5.0

contracts/access/AccessControlEnumerable.sol
```diff
@@ -49,16 +49,22 @@ abstract contract AccessControlEnumerable is IAccessControlEnumerable, AccessCon
...
+    function _grantRole(bytes32 role, address account) internal virtual override returns (bool) {
+        bool granted = super._grantRole(role, account);
+        if (granted) {
+            _roleMembers[role].add(account);
+        }
+        return granted;
     }
 
     /**
      * @dev Overload {_revokeRole} to track enumerable memberships
```
ernestognw, line 61:
```suggestion
     * @dev Overload {AccessControl-_revokeRole} to track enumerable memberships
```

contracts/access/AccessControlEnumerable.sol
```diff
@@ -49,16 +49,22 @@ abstract contract AccessControlEnumerable is IAccessControlEnumerable, AccessCon
     /**
      * @dev Overload {_grantRole} to track enumerable memberships
```
ernestognw, line 50:
```suggestion
     * @dev Overload {AccessControl-_grantRole} to track enumerable memberships
```

contracts/access/AccessControl.sol
```diff
@@ -190,10 +190,13 @@ abstract contract AccessControl is Context, IAccessControl, ERC165 {
      *
      * May emit a {RoleGranted} event.
      */
-    function _grantRole(bytes32 role, address account) internal virtual {
+    function _grantRole(bytes32 role, address account) internal virtual returns (bool) {
```
ernestognw, line 193: This function (and `_revokeRole`) needs to mention what the return value means in the NatSpec. I can't make a suggestion more than 3 lines further than here but I'd propose:

```solidity
* @dev Attempts to grant `role` to `account` and returns a boolean indicating if `role` was granted.
```

contracts/access/AccessControl.sol
```diff
@@ -204,10 +207,13 @@ abstract contract AccessControl is Context, IAccessControl, ERC165 {
      *
      * May emit a {RoleRevoked} event.
      */
-    function _revokeRole(bytes32 role, address account) internal virtual {
+    function _revokeRole(bytes32 role, address account) internal virtual returns (bool) {
```
ernestognw, line 210: Similar to `_grantRole`, I'd propose:

```solidity
* @dev Attempts to revoke `role` to `account` and returns a boolean indicating if `role` was revoked.
```

ernestognw:
I'd like to consider this change for 5.0, but we didn't address @frangio's comments.

> I would also like us to consider the alternative of making the functions revert in the cases where they would return false, as an alternative to adding a return value.

Sounds like a good idea but it'll require us to check how it behaves with AccessControl extensions. For example, `AccessControlDefaultAdminRules` does a `_revokeRole` and `_grantRole` together because it assumes no-reverts. If a user extends the contract to add an alternative mechanism for revoking the `DEFAULT_ADMIN_ROLE` after a change was scheduled via `beginDefaultAdminTransfer`, there won't be way of accepting the role anymore.

imo we don't have time to check for those cases if we decide to do reverts. I'm more in favor of the current PR proposal.

ernestognw:
btw, should we add tests for the internal function return values?

Amxx:
> btw, should we add tests for the internal function return values?

Added

Amxx:
@frangio @ernestognw

frangio:
> However, I would also like us to consider the alternative of making the functions revert in the cases where they would return false, as an alternative to adding a return value.

Following up on this question. Our current thinking is that if these functions reverted, it would be possible to frontrun a batch of `revokeRole` operations with a single `renounceRole` and cause the entire batch to fail, which doesn't seem good either.

If the concern is that `revokeRole` may appear to succeed when no roles have been revoked, the recommendation should be to check the transaction events to confirm the desired end result.

We should write this in the contract.

frangio (approved):

ernestognw (approved):
LGTM but we also discussed making `TimelockController` return the `proposalId`, I'd like to merge this and add that in another PR if it's not already out there.

Amxx:
> LGTM but we also discussed making `TimelockController` return the `proposalId`, I'd like to merge this and add that in another PR if it's not already out there.

In `TimelockController` I was talking about public function. Changing the return type of public functions is an issue because of composability. If the new implementations of TimelockController return the proposalId when doing propose, would we use that return value in the Governor modules ? If yes, then the module would not be compatible with pre-existing instances.

PR #4245, opened 2023-05-12
Resolve conflicts between master and next-v5.0

Amxx:
Same origin as #4225, but this PR targets `OpenZeppelin:next-v5.0`.

My understanding is that the goal is to merge this PR (not 4225) to resolve all conflict in `next-v5.0`, and then fast forward `master` to `next-v5.0`

Fixes LIB-858


frangio:
Ok I accidentally reverted `Amxx:next-v5.0` to `OpenZeppelin:next-v5.0` for a second, but now I've done what I intended which was just to rebase the contents of this PR into a single merge commit that you can find at https://github.com/Amxx/openzeppelin-contracts/pull/4245/commits/f933f3533073771074031ac73d0f4d62f4f0e2c9. You can verify that this commit contains exactly the same as before here: [`ec3f4f64..f933f353`](https://github.com/OpenZeppelin/openzeppelin-contracts/compare/ec3f4f64692c83bd06ead31dd36643d888d0243b..f933f3533073771074031ac73d0f4d62f4f0e2c9).

The best way to review this diff is to check it out locally and run `git show --remerge-diff`, this will show the resolved conflicts only.

scripts/generate/templates/Checkpoints.js
```diff
@@ -177,6 +195,13 @@ function length(${opts.historyTypeName} storage self) internal view returns (uin
     return self.${opts.checkpointFieldName}.length;
 }
 
+/**
+ * @dev Returns checkpoint at given position.
+ */
+function at(${opts.historyTypeName} storage self, uint32 pos) internal view returns (${opts.checkpointTypeName} memory) {
```
frangio, line 201: I just want to acknowledge that this was essentially renamed from `getAtPosition` to `at` and that was somewhat opinionated but I'm okay with the change.

frangio (commented):
I feel like I can vouch for almost everything in here, only except for some of the governance tests. It seems like they have differences with both branches (next-v5.0 and master). In master we had the timestamp refactor, what happened in next-v5.0 that affected these tests?

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -87,9 +120,23 @@ library SafeERC20 {
...
+     *
+     * This is a variant of {_callOptionalReturn} that silents catches all reverts and returns a bool instead.
+     */
+    function _callOptionalReturnBool(IERC20 token, bytes memory data) private returns (bool) {
+        // We need to perform a low level call here, to bypass Solidity's return data size checking mechanism, since
+        // we're implementing it ourselves. We cannot use {Address-functionCall} here since this should return false
+        // and not revert is the subcall reverts.
+
+        (bool success, bytes memory returndata) = address(token).call(data);
+        return success && (returndata.length == 0 || abi.decode(returndata, (bool))) && address(token).code.length > 0;
```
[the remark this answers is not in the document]
Amxx: Fixed in #4256

scripts/generate/templates/Checkpoints.js
Amxx: The transition to clock() in Votes requires this function to be available in Trace224 and not just in History. So I aligned the naming wit other functions in that "common" template

Amxx:
Replaced by #4256

PR #4267, opened 2023-05-22
Make Ownable's initial owner explicit

### LINKED ISSUE #2639: Ownable should support setting the initial owner through its constructor

Amxx:
Fixes #2639
Fixes LIB-835

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/clever-pumas-beg.md
+++ b/.changeset/clever-pumas-beg.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': major
 ---
 
-Add an `initialOwner` parameter to Ownable's constructor, making the ownership's initialization explicit.
+`Ownable`: Add an `initialOwner` parameter to the constructor, making the ownership initialization explicit.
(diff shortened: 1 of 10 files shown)


ernestognw (dismissed):
LGTM, thanks. Just a minor nit in the format of the changelog, I think it's more consisten with what we've done

frangio (approved):
LGTM!

PR #4268, opened 2023-05-22
Pack Governor's ProposalCore into a single slot

### LINKED ISSUE #3208: Governor: Use smaller integers for timing parameters

Amxx:
Fixes #3208
Fixes LIB-803

Note: we could merge `canceled` and `executed` into a single byte using an enum. That would increase the space for duration to `uint40`.

This is purely internal, no interface change:
- no test changes
- no documentation change ?

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/governance/Governor.sol
```diff
@@ -161,7 +156,8 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
      * @dev See {IGovernor-state}.
      */
     function state(uint256 proposalId) public view virtual override returns (ProposalState) {
-        ProposalCore storage proposal = _proposals[proposalId];
+        // ProposalCore is just one slot
```
frangio, line 159: What's the intention with this comment? Is it that we're reading it into memory because it's one slot and that will be more efficient? If so, let's write that explicitly.

contracts/governance/Governor.sol
```diff
@@ -207,14 +203,16 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
      * @dev See {IGovernor-proposalSnapshot}.
      */
     function proposalSnapshot(uint256 proposalId) public view virtual override returns (uint256) {
-        return _proposals[proposalId].voteStart;
+        ProposalCore memory proposal = _proposals[proposalId];
+        return proposal.voteStart;
```
frangio, line 207: Why would this be better? It implies an additional memory allocation. I think the previous code would definitely be more efficient.

contracts/governance/Governor.sol
```diff
@@ -207,14 +203,16 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
     }
 
     /**
      * @dev See {IGovernor-proposalDeadline}.
      */
     function proposalDeadline(uint256 proposalId) public view virtual override returns (uint256) {
-        return _proposals[proposalId].voteEnd;
+        ProposalCore memory proposal = _proposals[proposalId];
+        return proposal.voteStart + proposal.voteDuration;
```
frangio, line 215: This makes me wonder if Solidity would optimize the following:

```
return proposal.voteStart + proposal.voteDuration;
```

into a single SLOAD.... That would be better because it saves an allocation.

contracts/governance/Governor.sol
```diff
@@ -287,16 +285,14 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
         uint256 snapshot = currentTimepoint + votingDelay();
-        uint256 deadline = snapshot + votingPeriod();
+        uint256 duration = votingPeriod();
 
         _proposals[proposalId] = ProposalCore({
             proposer: proposer,
-            voteStart: SafeCast.toUint64(snapshot),
-            voteEnd: SafeCast.toUint64(deadline),
+            voteStart: SafeCast.toUint48(snapshot),
+            voteDuration: SafeCast.toUint32(duration),
```
frangio, line 293: Are we documenting these bounds anywhere?

frangio:
Let's change GovernorSettings in here as well.

contracts/governance/Governor.sol
Amxx: Yes, that is the intention.

contracts/governance/Governor.sol
Amxx: Here it doesn't bring anything. I did it for consistency with the other functions, but we ca revert to direct read.

contracts/governance/Governor.sol
Amxx: Not yet.

contracts/governance/Governor.sol
Amxx: I don't trust the compiler to go that, but yes we need to check.

That might be one of these case where the "optimal" solidity code depends on the compiler settings.

contracts/governance/IGovernor.sol
```diff
@@ -175,6 +178,10 @@ abstract contract IGovernor is IERC165, IERC6372 {
      *
      * NOTE: The {votingDelay} can delay the start of the vote. This must be considered when setting the voting
      * duration compared to the voting delay.
+     *
+     * NOTE: This value is saved when the proposal is saved, so that possible changes to the value do not affect
+     * proposals that have already been submitted. The type used to saving it is a uint32. Consequently, while this
```
Amxx, line 183:
```suggestion
     * proposals that have already been submitted. The type used to save it is a uint32. Consequently, while this
```

contracts/governance/Governor.sol
Amxx: Documentation added in IGovernor (in te natspec for getDelay and getDuration)

contracts/governance/IGovernor.sol
```diff
@@ -232,6 +235,10 @@ abstract contract IGovernor is IERC165, IERC6372 {
      *
      * NOTE: The {votingDelay} can delay the start of the vote. This must be considered when setting the voting
      * duration compared to the voting delay.
+     *
+     * NOTE: This value is saved when the proposal is saved, so that possible changes to the value do not affect
```
ernestognw, line 239: `saved` is kinda ambiguous, I think it might be better to:

```suggestion
     * NOTE: This value is stored when the proposal is submitted so that possible changes to the value do not affect
```

contracts/governance/IGovernor.sol
```diff
@@ -222,6 +222,9 @@ abstract contract IGovernor is IERC165, IERC6372 {
      *
      * This can be increased to leave time for users to buy voting power, or delegate it, before the voting of a
      * proposal starts.
+     *
+     * NOTE: While this interface returns a uint256, timepoint are stored as uint48 following the ERC-6372 clock type.
+     * Consequently this value must fit in a uint48 (when added to the current clock).
```
ernestognw, line 227:
```suggestion
     * NOTE: While this interface returns a uint256, timepoint are stored as uint48 following the ERC-6372 clock type.
     * Consequently this value must fit in a uint48 (when added to the current clock). See {IERC6372-clock}.
```

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -24,7 +24,7 @@ abstract contract GovernorTimelockCompound is IGovernorTimelock, Governor {
     ICompoundTimelock private _timelock;
 
     /// @custom:oz-retyped-from mapping(uint256 => GovernorTimelockCompound.ProposalTimelock)
```
ernestognw, line 26: We don't need this anymore after 5.0, right? Non-blocking and we shouldn't remove it yet but I think we may need to create a ticket for removing these. CC @frangio

contracts/governance/Governor.sol
Amxx:
```suggestion
        return _proposals[proposalId].voteStart;
```

contracts/governance/Governor.sol
Amxx:
```suggestion
        return _proposals[proposalId].voteStart + _proposals[proposalId].voteDuration;
```

contracts/governance/IGovernor.sol
```diff
@@ -222,6 +222,9 @@ abstract contract IGovernor is IERC165, IERC6372 {
      *
      * This can be increased to leave time for users to buy voting power, or delegate it, before the voting of a
      * proposal starts.
+     *
+     * NOTE: While this interface returns a uint256, timepoint are stored as uint48 following the ERC-6372 clock type.
```
frangio, line 226:
```suggestion
     * NOTE: While this interface returns a uint256, timepoints are stored as uint48 following the ERC-6372 clock type.
```

frangio (approved):

frangio:
LGTM except for the comments above.

frangio (approved):

PR #4261, opened 2023-05-18
Replace revert strings with custom errors

ernestognw:
Fixes LIB-839

#### Files to replace errors

- [x] ERC20
- [x] TokenTimelock
- [x] SafeERC20
- [x] ECDSA
- [x] ERC20Permit
- [x] ERC20Pausable
- [x] ERC20Votes
- [x] ERC20Wrapper
- [x] ERC20Snapshot
- [x] ERC20FlashMint
- [x] ERC20Capped
- [x] ERC4626
- [ ] ERC777 [DEPRECATED]
- [x] ERC2981
- [x] ERC721
- [x] ERC721URIStorage
- [x] ERC721Pausable
- [x] ERC721Enumerable
- [x] ERC721Consecutive
- [x] ERC721Wrapper
- [x] ERC721Burnable
- [x] ERC1155
- [x] ERC1155Supply
- [x] ERC1155Pausable
- [x] ERC1155Burnable
- [x] Clones
- [x] Initializable
- [x] UUPSUpgradeable
- [x] ERC1967Upgrade
- [x] UpgradeableBeacon
- [x] TransparentUpgradeableProxy
- [x] Ownable
- [x] AccessControlDefaultAdminRules
- [x] Ownable2Step
- [x] AccessControl
- [x] ReentrancyGuard
- [x] Pausable
- [ ] LibArbitrumL1 [DEPRECATED]
- [x] Checkpoints
- [x] MerkleProof
- [ ] Counters [DEPRECATED]
- [x] Address
- [ ] SafeMath [DEPRECATED]
- [x] Math
- [x] SafeCast
- [x] EnumerableMap
- [x] Create2
- [x] Strings
- [x] RefundEscrow
- [x] ConditionalEscrow
- [x] MinimalForwarder
- [x] GovernorCompatibilityBravo
- [x] Governor
- [x] Votes
- [x] GovernorTimelockCompound
- [x] GovernorCountingSimple
- [x] GovernorTimelockControl
- [x] GovernorSettings
- [x] TimelockController
- [x] PaymentSplitter
- [x] VestingWallet

### Tests

- [x] Access
- [x] Finance
- [x] Governance
- [x] Meta-tx
- [x] Proxy
- [x] Security
- [x] Token
- [x] Utils

### General

- [x] Tests
- [ ] Documentation
- [x] Changeset entry (run `npx changeset add`)

PR #4269, opened 2023-05-23
Replace safety transfer acceptance check calls with Address.functionCall

### LINKED ISSUE #4264: Replace safety transfer acceptance check calls with `Address.functionCall`
**📝 Details**
Both [ERC1155](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/token/ERC1155/ERC1155.sol#L460) and [ERC721](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/token/ERC721/ERC721.sol#L411) make transfer acceptance checks and they catch errors to bubble them up. However, this behavior is already implemented in [Address.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/utils/Address.sol#L231).

For consistency, it's required to use `Address.functionCall(address target, bytes memory data)` instead

qiweiii:
Fixes #4264  

> 📝 Details
> Both [ERC1155](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/token/ERC1155/ERC1155.sol#L460) and [ERC721](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/token/ERC721/ERC721.sol#L411) make transfer acceptance checks and they catch errors to bubble them up. However, this behavior is already implemented in [Address.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96b95592c300a1b38953b276f0b6c77017f5c942/contracts/utils/Address.sol#L231).
> 
> For consistency, it's required to use Address.functionCall(address target, bytes memory data) instead

Replaced direct `onERC721Received` and `onERC1155BatchReceived` calls with `Address.functionCall`.

No test changes since does not modify existing behaviour.

#### PR Checklist

- [x] Tests
- [x] Documentation


ernestognw (commented):
Hey @qiweiii, thank you so much for taking the issue 🙌 I think we also need to change the other acceptance check function in ERC1155 (`_doSafeTransferAcceptanceCheck`) aside from that looks good to me.

ernestognw:
As a side note, surprisingly the gas costs seem to have increased consistently.

https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/5055405317

I'd expect otherwise, but I'd like to double check, if that's the case, this may need fuether discussion

Amxx:
Note that we are doing some changes to the way `Address.functionCall` handles the default revert. We should wait for that refactor to be complete and then further analyse the impact of that change.

Also, we'll want to use `abi.encodeCall`

qiweiii:
@ernestognw I found a possible [answer](https://ethereum.stackexchange.com/a/117734) for increase in gas:
> The gas cost comes from the state lookup. To load the code that should be run for a call or delegatecall it is required to perform a lookup in the state tree. If these costs are removed it is easy to perform a denial of service attack on Ethereum.

@Amxx sure, let's wait for the refactor first.

ernestognw:
Hey @qiweiii, sorry for the late reply, I think the repo is now in good shape for adding this.

The conflicts are because of changes introduced to Address after #4261 and #4399. Also thanks for updating to `abi.encodeCall`

I'm requesting a review from the team. Please consider this is a 5.0 stretch goal so we might put some priority on other items first.

.changeset/hot-dodos-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Replace safety transfer acceptance check calls with `Address.functionCall`
```
ernestognw, line 5: I'm not sure if we'll require a changeset because this is not changing anything for the users or developers. CC: @frangio

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -353,16 +354,17 @@ contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI {
...
+            bytes memory payload = abi.encodeCall(
+                IERC1155Receiver.onERC1155BatchReceived,
+                (operator, from, ids, amounts, data)
+            );
+            bytes memory response = Address.functionCall(
+                to,
+                payload,
+                "ERC1155: transfer to non-ERC1155Receiver implementer"
+            );
+            if (bytes4(response) != IERC1155Receiver.onERC1155BatchReceived.selector) {
```
frangio, line 366: Pretty sure this should be `abi.decode(response, (bytes4))`. Same in ERC721.

.changeset/hot-dodos-flow.md
frangio: The guidelines say:

>All changes to the core codebase (excluding tests, auxiliary scripts, etc.) must be documented in a changelog, except for purely cosmetic or documentation changes.

This is not cosmetic or documentation. Refactors should also be described in the changelog. The reasoning  is that users may see that a contract was updated in X.Y.Z and may want to see exactly what changed.

frangio:
> Note that we are doing some changes to the way `Address.functionCall` handles the default revert. We should wait for that refactor to be complete and then further analyse the impact of that change.

@Amxx What were you referring to here?

Amxx:
> > Note that we are doing some changes to the way `Address.functionCall` handles the default revert. We should wait for that refactor to be complete and then further analyse the impact of that change.
> 
> @Amxx What were you referring to here?

This PR uses `Address.functionCall(address, bytes, string)`. That is something we recently remove

Amxx:
I updated this PR following the changes to the `Address` library and to `ERC721Utils`/`ERC1155Utils`.

This PR changes one behavior: if the receiver reverts without a reason, the old behavior used the specific custom error (`revert IERC1155Errors.ERC1155InvalidReceiver(to);` in the case of ERC1155), but this change will trigger an `revert Errors.FailedCall()` instead.

Also I'm worried this PR allocates memory buffer that are not freed, increassing the gas cost.

Waiting for the test to produce numbers, but I think I'll close it.

PR #4271, opened 2023-05-23
Avoid accessing SELFBALANCE opcode for zero-value function calls

forshtat:
'functionCallWithValue' is often called from 'functionCall' which means 'value' is often zero and getting 'address(this).balance' is excessive.

Additionally, the ERC-4337 explicitly forbids using 'balance' opcode in the verification context which prevents using SafeERC20 library: https://github.com/eth-infinitism/account-abstraction/pull/286#discussion_r1200947018

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/gorgeous-jars-yawn.md
+++ b/.changeset/gorgeous-jars-yawn.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Avoid accessing SELFBALANCE opcode for zero-value function calls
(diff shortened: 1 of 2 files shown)


huaweigu:
👍

naveed-klaytn:
nice pick (Y)

naveed949 (commented):
nice pick (Y)

frangio:
Note that this is not really an optimization. SELFBALANCE costs 5 gas, so the 10 gas cost of JUMPI implied by `||` will be larger than the cost of SELFBALANCE itself.

However, I find this an interesting point:

>ERC-4337 explicitly forbids using 'balance' opcode in the verification context which prevents using SafeERC20 library

Perhaps the solution is not to add `value == 0 || `, but instead to implement `functionCall` with separate code instead of dispatching to `functionCallWithValue(..., value = 0)`.

PR #4284, opened 2023-05-26
 Use ERC721Holder & ERC1155Holder in the TimelockController

### LINKED ISSUE #3936: Use ERC721Holder & ERC1155Holder for Governor, Timelock
As discussed before in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3174, the Governor and TimelockController contracts are able to receive ERC721 and ERC1155 tokens, so they should implement the respective Receiver interfaces. This was fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3230, but due to storage layout compatibility (https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3174#issuecomment-1033731672) we didn't use `ERC721Holder` and `ERC1155Holder`. We should do that now.

Amxx:
Fixes #3936
Fixes LIB-838

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/governance/Governor.sol
```diff
@@ -25,7 +25,7 @@ import "./IGovernor.sol";
  *
  * _Available since v4.3._
  */
-abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receiver, IERC1155Receiver {
+abstract contract Governor is Context, ERC165, EIP712, IGovernor, ERC721Holder, ERC1155Holder {
```
frangio, line 28: What is our convention for ordering of interfaces within inheritance lists? I would like to include this in the guidelines doc.

I would always put interfaces first but I think you always put them last. In this case there is an added weirdness though because `IGovernor` is an abstract contract.

scripts/checks/inheritance-ordering.js
```diff
@@ -13,7 +13,7 @@ for (const artifact of artifacts) {
   const linearized = [];
 
   for (const source in solcOutput.contracts) {
-    if (source.includes('/mocks/')) {
+    if (['contracts-exposed', 'contracts/mocks'].some(pattern => source.match(pattern))) {
```
frangio, line 16: Not a big deal but keeping the slashes and using `startsWith` makes the pattern more robust. `match` will cast the argument to a regex so better to avoid unless it's an actual regex.

```suggestion
    if (['contracts-exposed/', 'contracts/mocks/'].some(pattern => source.startsWith(pattern))) {
```

contracts/governance/Governor.sol
Amxx: My main goal was to have the inheritance ordering tests pass. I first I wanted to reorder some things, and I spent ~2h down the rabit hole of fixing ordering when the solution was staring at me from the start.

Amxx, line 28:
```suggestion
abstract contract Governor is Context, ERC165, EIP712, ERC721Holder, ERC1155Holder, IGovernor {
```

frangio:
Wow, I've been trying for a while to get the inheritance ordering to be consistent and I can't get it to work.

Amxx:
> Wow, I've been trying for a while to get the inheritance ordering to be consistent and I can't get it to work.

Yes, I sent over 2h on that too. We need a good guideline (IMO interfaces should come before implementations)

ernestognw (commented):
I don't see a major issue with this, but I see there's a conversation around the inheritance order, can you explain me real quick why is the inheritance an issue here?

Consider this an LGTM if you sort it out.

Amxx:
I'm starting to doubt the value of this PR if the hooks are overridden by #4314 anyway.

ernestognw:
> I'm starting to doubt the value of this PR if the hooks are overridden by #4314 anyway.

Just got back to this issue after reviewing tests for #4348, why would be the problem of being overridden by #4314? We'll still need it for Timelock independently I'd say

Amxx:
> > I'm starting to doubt the value of this PR if the hooks are overridden by #4314 anyway.
> 
> Just got back to this issue after reviewing tests for #4348, why would be the problem of being overridden by #4314? We'll still need it for Timelock independently I'd say

Right, the timelock part (that is not actually difficult) makes sens. The governor part (which is the one we struggle with) not so much.

Should we only keep the timelock part of this PR and merge it?

frangio:
>Should we only keep the timelock part of this PR and merge it?

This sounds good to me.

frangio (approved):
LGTM. Storage layout check is failing but this is expected.

ernestognw (dismissed):
LGTM, thanks @Amxx!

Amxx:
This PR was still targeting next-v5.0.

I had to retarget and resolve conflicts, which dismissed the reviews :/

frangio (approved):

Issue #3931, opened 2023-01-05
Disallow Address Poisoning in ERC20


alex-ppg:
**💻 Environment**

Environment is not applicable to this bug report as it pertains to the smart contracts themselves and how they behave on-chain.

**📝 Details**

The current token contract implementation of the [EIP-20](https://eips.ethereum.org/EIPS/eip-20) standard does not satisfy all `SHOULD` clauses of the standard definition and as such permits the commonly-known address poisoning attack by executing `transferFrom` instructions from arbitrary addresses with an `amount` of `0`.

The problem arises from how `_spendAllowance` evaluates the approval between the caller and the sender of the funds. In the EIP-20 standard, the following statement is present:

> The function **SHOULD** throw unless the `_from` account has deliberately authorized the sender of the message via some mechanism

This condition is not validated for zero-value transfers as no "deliberate" approval is evaluated. To ensure we remain compliant with the EIP-20 standard in full, we cannot disallow zero-value transfers altogether. 

As a workaround, we propose that the `require` check within `_spendAllowance` is refactored to evaluate a non-zero approval between the sender of funds and the `msg.sender` in case the `amount` is being specified as `0`.

This change will ensure maximal compatibility with existing contract systems, conform to the EIP-20 standard to a greater degree, and address the address poisoning attack we have seen being extensively exploited in recent times.

**🔢 Code to reproduce bug**

A simple Solidity smart contract to illustrate the bug in action:

```solidity
import "@openzeppelin/contracts/token/ERC20/ERC20.sol";

contract Poisoning {
    ERC20 public immutable token;

    constructor() {
        token = new ERC20("Test", "TST");
    }

    function poison(address from_, address to_) external {
        require(token.allowance(from_, address(this)) == 0);
        token.transferFrom(from_, to_, 0);
    }
}
```

Invoking the `poison` function with arbitrary `from_` and `to_` arguments will successfully perform a `transferFrom` invocation even when the approval between `from_` (the sender of funds) and `address(this)` (the caller of the `transferFrom` function) has been set to `0`.

This behaviour permits polluting the transfer history of an account on blockchain explorers which in turn can cause users to be misled and copy incorrect addresses when performing their own valid transfers.

frangio:
Thanks @alex-ppg. This was on our radar as well.

[...]

PR #4285, opened 2023-05-27
fix: avoid zero-value transfer attack

aitimate:
As shown in the question


frangio:
>As shown in the question

What question?

Please see the discussion in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3931.

PR #4289, opened 2023-05-29
Remove utils/Counters.sol

### LINKED ISSUE #4233: Remove `Counters.sol`
**🧐 Motivation**

The `Counters.sol` contract initially served as an abstraction for a number that guarantees two main properties:

- Provide a number that can only be incremented, decremented or reset
- Abstract its behavior into a utility

Also, it allows to use `unchecked` arithmetic assuming no overflow

However, its usage has been reduced across OpenZeppelin Contracts and there are some relevant reasons to remove it, such as:

- New announced Solidity features might allow replicating the behavior without the `Counters.sol` contract (eg. User Defined Types)
- It apparently doesn't add much value against using a regular variable
- It increments the maintenance surface
- Can be confusing for newcomers

**📝 Details**

We're proposing removing it for Contracts 5.0, but we want to hear comments against removing it (if any), so we can have a clear view of what to do in the future. An alternative we're strongly considering is removing it for 5.0 but re-adding in the future some variation if people consider it useful and it makes sense.

Amxx:
Fixes #4233 
Fixes LIB-890

#### PR Checklist

- [ ] Tests
- [x] Documentation


contracts/utils/Nonces.sol
```diff
@@ -1,31 +1,28 @@
...
      * @dev Consumes a nonce.
      *
      * Returns the current value and increments nonce.
      */
-    function _useNonce(address owner) internal virtual returns (uint256 current) {
-        Counters.Counter storage nonce = _nonces[owner];
-        current = nonce.current();
-        nonce.increment();
+    function _useNonce(address owner) internal virtual returns (uint256) {
+        unchecked {
```
frangio, line 23: We need to add a comment justifying why `unchecked` is safe.

frangio (approved):
LGTM

Amxx:
Do we want to add natspec comments so the upgradeability layout test passes?

frangio:
No, we don't need those.

H3liaw:
why did you remove the CountersUpgradeable ?

Amxx:
@H3liaw It was all discussed in the linked issue https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4233

PR #4296, opened 2023-06-01
Replace some uses of abi.encodePacked with more explicit alternatives

Amxx:
Some occurences of abi.encodePacked were purposefully left:
- when its part of a "vendored" contracted that we use for testing
- when the different blocks being encoded have various types (string+bytes+...)

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/flat-bottles-wonder.md
+++ b/.changeset/flat-bottles-wonder.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-Replace some uses of `abi.encodePacked` with clearer alternatives
+Replace some uses of `abi.encodePacked` with clearer alternatives (e.g. `bytes.concat`, `string.concat`).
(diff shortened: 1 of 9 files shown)


frangio (approved):
Very nice thanks!

.changeset/flat-bottles-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Replace some uses of `abi.encodePacked` with clearer alternatives
```
frangio, line 5:
```suggestion
Replace some uses of `abi.encodePacked` with clearer alternatives (e.g. `bytes.concat`, `string.concat`).
```

ernestognw (approved):
LGTM.

Side note but we might want to update the recommendation we're making for Base64 in the [Utilities page](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/2ee1da12c4498fc130b7770905d0b4c4f91f6a9d/docs/modules/ROOT/pages/utilities.adoc?plain=1#L130)

Issue #3883, opened 2022-12-16
Stop using hardcoded resolution of functions


frangio:
In a few places we hardcode the resolution of functions, skipping over any potential overrides. This is done to preserve the internal consistency of the code. In the example below, we need owner to be the actual owner that the contract stores in the ownership mapping in storage.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3d7a93876a2e5e1d7fe29b5a0e96e222afdc4cfa/contracts/token/ERC721/ERC721.sol#L113

The result is confusing though, so I think we should look for some alternative. Our policy anyway is that if a developer overrides a function with custom logic it is their responsibility to make sure that other things don't break.

fulldecent:
Sounds good.

Here is all of them to review:

- [ ] `SafeERC20.safeTransfer(token, account, payment);`
- [ ] `SafeERC20.safeTransfer(IERC20(token), beneficiary(), amount);`
- [ ] `return ERC1155URIStorage.uri(tokenId);`
- [ ] `return ERC2771Context._msgSender();`
- [ ] `return ERC2771Context._msgData();`
- [ ] `SafeERC20.safePermit(IERC20Permit(address(_token)), owner, spender, value, deadline, v, r, s);`
- [ ] `ERC1967Upgrade._upgradeToAndCall(newImplementation, bytes(""), false);`
- [ ] `ERC1967Upgrade._upgradeToAndCall(newImplementation, data, false);`
- [ ] `return ERC1967Upgrade._getImplementation();`
- [ ] `return ERC1155Supply.totalSupply(id) > 0;`
- [ ] `require(ERC20.totalSupply() + amount <= cap(), "ERC20Capped: cap exceeded");`
- [ ] `return token == address(this) ? type(uint256).max - ERC20.totalSupply() : 0;`
- [ ] `SafeERC20.safeTransferFrom(underlying, _msgSender(), address(this), amount);`
- [ ] `SafeERC20.safeTransfer(underlying, account, amount);`
- [ ] `SafeERC20.safeTransferFrom(_asset, caller, address(this), assets);`
- [ ] `SafeERC20.safeTransfer(_asset, receiver, assets);`
- [ ] `address owner = ERC721.ownerOf(tokenId);`
- [ ] `address owner = ERC721.ownerOf(tokenId);`
- [ ] `address owner = ERC721.ownerOf(tokenId);`
- [ ] `owner = ERC721.ownerOf(tokenId);`
[...]

PR #4299, opened 2023-06-02
Remove hardcoded function resolution

frangio:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3883

This should be removing all hardcoded function resolution.

As a consequence, in ERC721 I have switched it to checked arithmetic for `_balances[from] -= 1`, to avoid overflow if `_ownerOf` is overriden and out of sync with the mapping.

### DIFF
--- a/.changeset/red-dots-fold.md
+++ b/.changeset/red-dots-fold.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': major
 ---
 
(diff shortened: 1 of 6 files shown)


Amxx:
> As a consequence, in ERC721 I have switched it to checked arithmetic for `_balances[from] -= 1`, to avoid overflow if `_ownerOf` is overriden and out of sync with the mapping.

That part I'm not sure about. We considered that option in `ERC721Consecutive`, With an additional balance
```solidity
mapping(address => uint256) private _balanceFromConsecutiveMints;

function _balanceOf(address user) internal virtual view returns (uint256) {
    unchecked {
        return super._balanceOf(user) + _balanceFromConsecutiveMints[user];
    }
}
```
Which overflow balancing out. If we make the balance update in `_transfer` checked, we kill that pattern.

Amxx (approved):

PR #4300, opened 2023-06-02
Optimize array access in ERC1155

### LINKED ISSUE #3940: Array optimizations for ERC1155
This PR refers to the `next-v5.0` branch. In https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3876 two potential optimizations were identified.

1. `_asSingletonArray`: Throughout the contract we use:
```solidity
        uint256[] memory ids = _asSingletonArray(id);
        uint256[] memory amounts = _asSingletonArray(amount);
```
We could combine this into a single function `_asSigletonArrays(id, amount)` and using assembly we would be able to do a single allocation and possibly avoid a couple of bounds checks.

2. `_unsafeAccess`: In places where we loop over arrays of ids and amounts, we don't need to do bounds checking inside the for loop, so we could define a function to skip it.

Given that these are arrays in memory, it's unclear how much of an improvement these optimizations will make, so we have to benchmark to know for sure.

Based on the results of using `_unsafeAccess` for memory arrays we should consider applying it in other places

clauBv23:
Fixes #3940 

Add a function to access memory unsafely and refactor ERC1155 using it.

NOTE: I would recommend extending the current function to other datatypes as it's currently being done with the `unsafeAccess` one, it was out of this task scope so is not being done on this PR.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/hot-coins-judge.md
+++ b/.changeset/hot-coins-judge.md
@@ -2,4 +2,5 @@
 'openzeppelin-solidity': patch
 ---
 
-Array optimization in ERC1155
+`Arrays`: Add `unsafeMemoryAccess` helpers to read from a memory array without checking the length.
+`ERC1155`: optimize memory accesses
--- a/.changeset/hot-coins-judge.md
+++ b/.changeset/hot-coins-judge.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Array optimization in ERC1155
--- a/.changeset/hot-coins-judge.md
+++ b/.changeset/hot-coins-judge.md
@@ -1,6 +1,5 @@
 ---
-'openzeppelin-solidity': patch
+'openzeppelin-solidity': minor
 ---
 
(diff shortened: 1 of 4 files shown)


Amxx (approved):
Nice! LGTM

Amxx:
There are likelly other parts of the code that could use that access pattern.

Amxx (approved):

.changeset/hot-coins-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Array optimization in ERC1155
```
Amxx, line 5:
```suggestion
`Arrays`: Add `unsafeMemoryAccess` helpers to read from a memory array without checking the length.
`ERC1155`: optimize memory accesses
```

Amxx (dismissed):

frangio (approved):
LGTM

PR #4301, opened 2023-06-02
`ECDSA`: Use unchecked arithmetic for the `tryRecover` function

pcaversaccio:
This PR adds the usage of unchecked arithmetic for the `tryRecover` function that receives the `r` and `vs` short-signature fields separately. The addition can't overflow (in my understanding) by design since the shift operation results in 0 or 1. 

#### PR Checklist
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -98,9 +98,12 @@ library ECDSA {
      * _Available since v4.3._
      */
     function tryRecover(bytes32 hash, bytes32 r, bytes32 vs) internal pure returns (address, RecoverError) {
-        bytes32 s = vs & bytes32(0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff);
-        uint8 v = uint8((uint256(vs) >> 255) + 27);
-        return tryRecover(hash, v, r, s);
+        unchecked {
+            bytes32 s = vs & bytes32(0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff);
+            // We do not check for an overflow here since the shift operation results in 0 or 1.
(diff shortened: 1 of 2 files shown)


contracts/utils/cryptography/ECDSA.sol
```diff
@@ -98,9 +98,12 @@ library ECDSA {
      * _Available since v4.3._
      */
     function tryRecover(bytes32 hash, bytes32 r, bytes32 vs) internal pure returns (address, RecoverError) {
-        bytes32 s = vs & bytes32(0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff);
-        uint8 v = uint8((uint256(vs) >> 255) + 27);
-        return tryRecover(hash, v, r, s);
+        unchecked {
+            bytes32 s = vs & bytes32(0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff);
```
Amxx, line 102: Does it change anything to have this line inside the unchecked block ?

pcaversaccio: No but I thought it would be cleaner like that.

Amxx: We can't have it only around `uint8 v = uint8((uint256(vs) >> 255) + 27);`, because that would limit the visibility of `v` and we would not be able to return it.

pcaversaccio: I mean, as an alternative, if you're open to it we can use inline assembly. Something like:

```solidity
uint8 v;
assembly {
    v := add(shr(255, vs), 27)
}
```

frangio (approved):
LGTM

ernestognw (approved):
Thanks @pcaversaccio

PR #4303, opened 2023-06-02
Add EIP-712 `name` and `version` getters

### LINKED ISSUE #4226: Add internal name and version getters for EIP712
Note that we have these getters in the [upgradeable version of EIP712](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/cryptography/EIP712Upgradeable.sol).

RenanSouza2:
Fixes #4226

Add internal name and version getters for EIP712

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/utils/cryptography/EIP712.sol
```diff
@@ -139,4 +139,26 @@ abstract contract EIP712 is IERC5267 {
...
+
+    /**
+     * @dev The name parameter for the EIP712 domain.
+     *
+     * NOTE: This function reads from storage by default, but can be redefined to return a constant value if gas costs
+     * are a concern.
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function _EIP712Name() internal view virtual returns (string memory) {
+        return ShortStrings.toStringWithFallback(_name, _nameFallback);
```
Amxx, line 151:
```suggestion
        return _name.toStringWithFallback(_nameFallback);
```

contracts/utils/cryptography/EIP712.sol
```diff
@@ -139,4 +139,26 @@ abstract contract EIP712 is IERC5267 {
...
+
+    /**
+     * @dev The version parameter for the EIP712 domain.
+     *
+     * NOTE: This function reads from storage by default, but can be redefined to return a constant value if gas costs
+     * are a concern.
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function _EIP712Version() internal view virtual returns (string memory) {
+        return ShortStrings.toStringWithFallback(_version, _versionFallback);
```
Amxx, line 162:
```suggestion
        return _version.toStringWithFallback(_versionFallback);
```

Amxx (commented):
IMO, `eip712Domain() public` should use these two new internal function (in case they are ever overriden)

EDIT: actually, if the values ever change (because of an override) that would cause a lot of trouble ... so I'll remove the virtual keyword for the 2 internal functions.

RenanSouza2:
> IMO, `eip712Domain() public` should use these two new internal function (in case they are ever overriden)
> 
> EDIT: actually, if the values ever change (because of an override) that would cause a lot of trouble ... so I'll remove the virtual keyword for the 2 internal functions.

Yes, agreed

how about an internal setFunction(atring) so it can also be overriden?

Amxx:
I don't think this values should ever be changed. If they were, the caches would be invalid ... and we can't update the immutable cache.

Also I don't think frontends are designed to handle such changes

RenanSouza2:
Makes sense, I commited with the getters used internaly and there are no more changes to be done, right?

Also the checks / tests-upgradeable is failling but I gess it it some conflict with the upgradeable contract that already have this getters, I`ll see how to fix this

contracts/utils/cryptography/EIP712.sol
```diff
@@ -131,12 +131,34 @@ abstract contract EIP712 is IERC5267 {
...
             address(this),
             bytes32(0),
             new uint256[](0)
         );
     }
+
+    /**
+     * @dev The name parameter for the EIP712 domain.
+     *
+     * NOTE: This function reads from storage by default, but can be redefined to return a constant value if gas costs
```
Amxx, line 146: A few things:
- By default we read `_name` with is an immutable value. We only read from storage if necessary (in case the value is too large to fit in a ShortString.
- Since this is not virtual, its not possible to redefine it to return a constant value.

RenanSouza2: I've updated the comments

contracts/utils/cryptography/EIP712.sol
```diff
@@ -143,19 +143,19 @@ abstract contract EIP712 is IERC5267 {
...
+     * It only reads from storage if necessary (in case the value is too large to fit in a ShortString).
      */
     // solhint-disable-next-line func-name-mixedcase
     function _EIP712Name() internal view returns (string memory) {
         return _name.toStringWithFallback(_nameFallback);
     }
 
     /**
-     * @dev The version parameter for the EIP712 domain.
+     * @dev The name parameter for the EIP712 domain.
```
Amxx, line 155:
```suggestion
     * @dev The version parameter for the EIP712 domain.
```

Amxx:
It Looks good to me, but I'd like @ernestognw and @frangio 's opinion on whether the function should be virtual.

Its critical that the value returned by the getters is constant in time, because the `eip721domain` must match the (immutable) cached values.

.changeset/little-falcons-build.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Add internal name and version getters for EIP712
```
Amxx, line 5:
```suggestion
`EIP712`: Add internal getters for the name and version strings
```

RenanSouza2: Done

Amxx (dismissed):
LGTM

ernestognw:
> It looks good to me, but I'd like @ernestognw and @frangio 's opinion on whether the getters should be virtual.
> 
> Its critical that the values returned by the getters are constant in time, because the `eip721domain` must match the (immutable) cached values.

Perhaps there's a use case in overriding any of those values. For example, if a contract wants to invalidate a domain within an upgrade (eg. reject signatures based on the version).

Still, I'm not convinced it's worth the risk of letting users override them.

Are you really sure those values will never change? @Amxx

frangio:
@jbcarpanelli suggested making all functions `virtual` but adding an annotation saying whether overriding is either fully okay, or okY under certain conditions (invoking super), or probably never okay. The idea is to still allow these overrides for the shadowy coders who want that but to document the potential implications.

E.g.:

```
@custom:overrideability { unrestricted | restricted | discouraged }
```

ernestognw:
I had to update the upgradeable scripts. However, there were a few issues related to previous changes we've merged.
I think it makes sense but since it's my first time using the upgradeable patch I'll require a review from @frangio and @Amxx again.

Almost there!

Amxx:
> Perhaps there's a use case in overriding any of those values. For example, if a contract wants to invalidate a domain within an upgrade (eg. reject signatures based on the version).

If the internal getters are override, and the value changes during the contract's life, the `eip712domain` will be updated but the cache will not be modified. The domain separator used for recovery will stay the old one.

On the upgradeable version of this contract, the storage is very different. It doesn't have a cache, and recomputes the domainSeparator every time. The upgradeable version already includes getters, [and they are virtual](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/cryptography/EIP712Upgradeable.sol).

So any decision we make here should NOT be motivated by contract upgradeability.

frangio (approved):
LGTM.

PR #4309, opened 2023-06-03
Remove further hardcoded function resolution

frangio:
For some reason I missed these in #4299.

The one in `ERC721Enumerable._removeTokenFromOwnerEnumeration` is interesting and deserves attention. We were using `ERC721.balanceOf` because we are more confident that it's the length of the enumeration than when we use the potentially overridden `balanceOf`. I've just made it call `balanceOf` now because I don't see a way around this, other than introducing a storage inefficiency by storing the length which is redundant with the balance mapping.

---

For the record here is how I am checking that these changes are exhaustive. (`rg` = ripgrep)

```
rg '^(abstract )?contract (\w+).*' contracts -I -g '*.sol' --replace '\b$2\.\w+\(' > patterns
rg -f patterns -g '*.sol'
```

The only remaining ones are `ERC2771Context._msgData()` and `ERC2771Context._msgSender()` in a mock, but without this we get a compiler warning.

### DIFF
--- a/.changeset/red-dots-fold.md
+++ b/.changeset/red-dots-fold.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': major
 ---
 
-Overrides are now used internally for a number of functions that were previously hardcoded to their default implementation in certain locations: `ERC1155Supply.totalSupply`, `ERC721.ownerOf`, `ERC721.balanceOf` in `ERC721Enumerable`, and `ERC20.totalSupply` in `ERC20FlashMint`.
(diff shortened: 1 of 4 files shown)


Amxx (dismissed):
LGTM.

There are many restrictions to combining the ERC721Enumerable module with "custom" balanceOf. This is why this module is not compatible with ERC721Concecutive (for example).
Maybe we will want to clarify the assumption that this module makes.

ernestognw (approved):
I also agree with with adding a warning about the risks of using custom `balanceOf`

ernestognw (approved):
Merging since it already implements the suggestion @Amxx and me mentioned.

PR #4312, opened 2023-06-03
Update ERC20.sol to enable effective _transfer overriding.

Alexander-Herranz:
Updated "_balances" mapping's visibility from private, to internal, in order to enable _transfer function overriding when inheriting.  Ie. to apply taxes, burning, etc. in each transfer.

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -35,7 +35,7 @@ import "../../utils/Context.sol";
  * allowances. See {IERC20-approve}.
  */
 contract ERC20 is Context, IERC20, IERC20Metadata {
-    mapping(address => uint256) private _balances;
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @Alexander-Herranz 

The storage mappings are private by design. This is to ensure that any changes comes with the corresponding events. You should be able to update the balances the way you like, using "super" calls.

For example, if you want all transfers that are not mints or burn to automatically transfer 1% to the owner of the contract, you can do it without touching the storage as follows

```solidity
function _update(address from, address to, uint256 amount) internal virtual {
    if (from != address(0) && to != address(0)) {
        uint256 fee = amount / 100;
        super._update(from, owner(), fee);
        amount -= fee;
    }
    super._update(from, to, amount);
}
```

PR #4313, opened 2023-06-04
Minor update Multicall.sol

nazarkhatsko:
Hello, I was studying your code base and noticed a mistake in the written code. Here is the modified code in the Multicall contract method:

```solidity
function multicall(bytes[] calldata data) external virtual returns (bytes[] memory) {
    bytes[] memory results = new bytes[](data.length);
    for (uint256 i = 0; i < data.length; i++) {
        results[i] = Address.functionDelegateCall(address(this), data[i]);
    }
    return results;
}
```

The change in the code is that the variable results is now declared directly in the function signature instead of being assigned later. This change improves readability and ensures that the variable is initialized correctly.

### DIFF
--- a/contracts/utils/Multicall.sol
+++ b/contracts/utils/Multicall.sol
@@ -15,8 +15,8 @@ abstract contract Multicall {
      * @dev Receives and executes a batch of function calls on this contract.
      * @custom:oz-upgrades-unsafe-allow-reachable delegatecall
      */
-    function multicall(bytes[] calldata data) external virtual returns (bytes[] memory results) {
-        results = new bytes[](data.length);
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @nazarkhatsko 

> This change improves readability and ensures that the variable is initialized correctly.

The readability can be discussed. Its probably a matter of personnal preference.
What do you mean by "ensures that the variable is initialized correctly"? How is the current syntax not initializing the variable correctly ?

nazarkhatsko:
The original version can be used because the changes are not essential, but the original version does not need the last line to return the change, or we can declare and initialize it in a function in my code example.

nazarkhatsko:
From the technical side, there are no problems with this code during initialization variable and beyond, I meant only the readable part.

PR #4317, opened 2023-06-05
`ECDSA`: Use hexadecimal literals

pcaversaccio:
As title. Since this is a simple refactor, I assume no `Changeset` is required.


ernestognw:
I agree it shouldn't require a changeset. I'm requesting review from the team

ernestognw:
Thanks as always @pcaversaccio!, I'm merging

PR #4324, opened 2023-06-07
Remove Context

### LINKED ISSUE #3929: Consider removing Context
`Context` provides `_msgSender()` and the reason it exists and is used across the codebase is so that the logic for getting the function caller can be replaced via inheritance, for example using `ERC2771Context`. It is necessary if we want to support overriding that logic for any piece of code in the library. While `Context` is fairly simple, it is pervasive and the vast majority of use cases do not need the functionality at all, so it's a great candidate for removal to simplify the code.

We would like to understand first if Context alternatives such as `ERC2771Context` are actively used, as well as what are the alternative recommendations we can make for those use cases.

balajipachai:
Fixes #3929 

#### PR Checklist

- [All Passing] Tests
- [No Changes] Documentation
- [Not needed] Changeset entry (run `npx changeset add`)


Amxx:
Hello @balajipachai 

Issue #3929 is "**Consider** removing Context". The truth is we've been discussing that a lot lately, and the consensus so far is that we will very probably not remove it.

frangio:
@balajipachai Your willingness to contribute is appreciated, but please ask for confirmation before submitting an issue. We don't want you to do work that we have to discard afterwards. In this case, the issue was meant as discussion, it was not at all ready to be implemented.

balajipachai:
Thank You @frangio and @Amxx , got your point.
I understand you guys are running tight on deadline thus I am trying to pick up the ones and start working, having been said that if either of you could direct/assign some issues to me that would be really helpful and at the same time the efforts that I put in will indeed be useful for the project overall.
Thanks again.

PR #4325, opened 2023-06-07
Make ERC1967Upgrades a library instead of an abstract contract

### LINKED ISSUE #4204: Turn ERC1967Upgrade into a library
Solidity 0.8.20 should include https://github.com/ethereum/solidity/pull/10996, which would finally allow `ERC1967Upgrade` to be written as a library (instead of an abstract contract).

This raises the question of what to do with `IERC1967`.

Amxx:
Fixes #4204
Fixes LIB-880

This PR requires 0.8.20, and forces it in `ERC1967Upgrades.sol` for the following bugfix:
> ABI: Include events in the ABI that are emitted by a contract but defined outside of it.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


frangio:
Sorry for asking this now but can you remind me why we want this to be a library? I don't see a clear benefit at the moment.

Amxx:
- It doesn't need to be a contract, because it does not include any public/external function. 
- It's just tooling. Close example is SafeERC20. Should SafeERC20 be an abstract contracts 
- if it's a contract, all the events it declares (or its parents declare) are included in the ABI. You end up with a uups proxy having BeaconUpgraded in its abi.
- when we do partial transpilation, we will not want to transpile this one.

frangio:
> * if it's a contract, all the events it declares (or its parents declare) are included in the ABI. You end up with a uups proxy having BeaconUpgraded in its abi.

Ok, this seems like one of the strongest reasons for me.

My main concern was that the functions implicitly access storage, which is not something I'd usually expect from a library. What do you think about this?

Amxx:
I don't see that as an issue. 

IMO it's explicit that the ERC1967Upgrade library provides helper to interact with the ERC1967 storage slots.

It's not like the abstract contracts was declaring the storage explicitely either.

contracts/proxy/ERC1967/ERC1967Upgrade.sol
```diff
@@ -114,7 +132,8 @@ abstract contract ERC1967Upgrade is IERC1967 {
      * This is the keccak-256 hash of "eip1967.proxy.admin" subtracted by 1, and is
      * validated in the constructor.
      */
-    bytes32 internal constant _ADMIN_SLOT = 0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103;
+    // solhint-disable-next-line private-vars-leading-underscore
```
ernestognw, line 135: I want to raise a question about this. It's not the first time I found cases where this warning shows and I actually think the constants should not have a leading underscore, but it's becoming quite common.

Can we take a look a the rule configuration? I see in the docs there's [a `strict`](https://github.com/protofire/solhint/blob/master/docs/rules/naming/private-vars-leading-underscore.md) flag that might help, but only apply to [non-state vars](https://github.com/protofire/solhint/blob/master/lib/rules/naming/private-vars-leading-underscore.js#L82), which I'm not sure if constants fall into that category.

Non blocking in case this convo gets bigger.

contracts/proxy/ERC1967/ERC1967Upgrade.sol
```diff
@@ -134,30 +153,31 @@ abstract contract ERC1967Upgrade is IERC1967 {
...
             revert ERC1967InvalidAdmin(address(0));
         }
-        StorageSlot.getAddressSlot(_ADMIN_SLOT).value = newAdmin;
+        StorageSlot.getAddressSlot(ADMIN_SLOT).value = newAdmin;
     }
 
     /**
      * @dev Changes the admin of the proxy.
      *
      * Emits an {AdminChanged} event.
```
ernestognw, line 162: Perhaps we want to point to `IERC1967` events reference with:

```suggestion
     * Emits an {IERC1967-AdminChanged} event.
```

Apply for the rest of the events. Non-blocking either.

.changeset/grumpy-worms-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+Change definition of ERC1967Upgrades from "abstract contract" to "library"
```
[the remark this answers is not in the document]
Amxx: Since we are renaming, what do you think of:
```suggestion
`ERC1967Utils`: Refactor the `ERC1967Upgrade` abstract contract as a library.
```

contracts/proxy/ERC1967/ERC1967Upgrade.sol
```diff
@@ -90,22 +108,22 @@ abstract contract ERC1967Upgrade is IERC1967 {
      *
      * Emits an {Upgraded} event.
      */
-    function _upgradeToAndCallUUPS(address newImplementation, bytes memory data, bool forceCall) internal {
+    function upgradeToAndCallUUPS(address newImplementation, bytes memory data, bool forceCall) internal {
         // Upgrades from old implementations will perform a rollback test. This test requires the new
         // implementation to upgrade back to the old, non-ERC1822 compliant, implementation. Removing
         // this special case will break upgrade paths from old UUPS implementation to new ones.
         if (StorageSlot.getBooleanSlot(_ROLLBACK_SLOT).value) {
             _setImplementation(newImplementation);
```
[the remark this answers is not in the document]
Amxx: I implemented this change but I think it deserve a dedicated PR 
(two small PR are easier to review then one big IMO)

WDYT?

contracts/proxy/ERC1967/ERC1967Upgrade.sol
frangio: 100% we don't want this rule, we can't have these comments everywhere. Solhint is not working well for us...

Amxx: Lets create an issue for that, and resolve in another PR

contracts/proxy/ERC1967/ERC1967Upgrade.sol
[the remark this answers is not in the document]
frangio: Sounds good.

contracts/proxy/ERC1967/ERC1967Upgrade.sol
frangio: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4355

frangio (approved):

scripts/upgradeable/transpile.sh
```diff
@@ -29,7 +29,7 @@ npx @openzeppelin/upgrade-safe-transpiler@latest -D \
   -x 'contracts/proxy/**/*' \
   -x '!contracts/proxy/Clones.sol' \
   -x '!contracts/proxy/ERC1967/ERC1967Storage.sol' \
-  -x '!contracts/proxy/ERC1967/ERC1967Upgrade.sol' \
+  -x '!contracts/proxy/ERC1967/ERC1967Utils.sol' \
```
ernestognw, line 32: Aren't we ignoring libraries for transpilation already?

If so, let's also create an issue with good-first-issue label

ernestognw (approved):
LGTM 🙌

scripts/upgradeable/transpile.sh
Amxx: Basically the library are needed by the contracts, so if we don't transpile them we need to get them from somewhere.

The partial transpilation proposal is to not transpile the libraries and the interfaces, and import the vanila version instead. (the creates a dependency from the upgradeable contracts to the vanilla contracts)

PR #4340, opened 2023-06-11
`IERC1155`: Fix `safeBatchTransferFrom` documentation

pcaversaccio:
As title. The function `safeBatchTransferFrom` invokes `onERC1155BatchReceived` and not `onERC1155Received`. No `Changeset` is required since it's a simple documentation correction.


frangio (approved):
Thanks!

PR #4346, opened 2023-06-13
Make `MinimalForwarder` production-ready for 5.0

### LINKED ISSUE #3664: Loss of unused Ether in meta-transaction

### LINKED ISSUE #3884: Consider removing MinimalForwarder

ernestognw:
Fixes #3884
Fixes #3664
Replaces #4065 

Currently, `MinimalForwarder` is not considered production-ready since it was designed mainly for testing purposes in the library. However, it's not that far from a MinimalForwarder that's something we may recommend.

The new MinimalForwarder should have:

- [x] A deadline for expiring transactions
- [x] A `msg.value` mismatch check
- [ ] Redesign of gas forwarding check (in discussion)

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/metatx/MinimalForwarder.sol
+++ b/contracts/metatx/MinimalForwarder.sol
@@ -5,16 +5,21 @@ pragma solidity ^0.8.19;
 
 import "../utils/cryptography/ECDSA.sol";
 import "../utils/cryptography/EIP712.sol";
+import "../utils/Nonces.sol";
 
 /**
- * @dev Simple minimal forwarder to be used together with an ERC2771 compatible contract. See {ERC2771Context}.
+ * @dev A minimal implementation of a production-ready forwarder compatible with ERC2771 contracts.
  *
- * MinimalForwarder is mainly meant for testing, as it is missing features to be a good production-ready forwarder. This
- * contract does not intend to have all the properties that are needed for a sound forwarding system. A fully
- * functioning forwarding system with good properties requires more complexity. We suggest you look at other projects
- * such as the GSN which do have the goal of building a system like that.
+ * This forwarder operates on forward request that include:
+ * * `from`: An address to operate on behalf of. It is required to be equal to `msg.sender`.
+ * * `to`: An address destination to call within the request.
+ * * `value`: The amount of ETH to attach within the requested call.
+ * * `gas`: The amount of gas limit that will be forwarded within the requested call.
+ * * `nonce`: A unique transaction ordering identifier to avoid replayability and request invalidation.
+ * * `deadline`: A timestamp after which the request is not executable anymore.
+ * * `data`: Encoded `msg.calldata` to send within the requested call.
  */
-contract MinimalForwarder is EIP712 {
+contract MinimalForwarder is EIP712, Nonces {
     using ECDSA for bytes32;
 
     struct ForwardRequest {
@@ -23,13 +28,14 @@ contract MinimalForwarder is EIP712 {
         uint256 value;
         uint256 gas;
         uint256 nonce;
+        uint48 deadline;
         bytes data;
     }
 
-    bytes32 private constant _TYPEHASH =
-        keccak256("ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,bytes data)");
-
-    mapping(address => uint256) private _nonces;
+    bytes32 private constant _FORWARD_REQUEST_TYPEHASH =
+        keccak256(
+            "ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,uint48 deadline,bytes data)"
+        );
 
     /**
      * @dev The request `from` doesn't match with the recovered `signer`.
@@ -37,45 +43,70 @@ contract MinimalForwarder is EIP712 {
     error MinimalForwarderInvalidSigner(address signer, address from);
 
     /**
-     * @dev The request nonce doesn't match with the `current` nonce for the request signer.
+     * @dev The request `value` doesn't match with the `msg.value`.
      */
-    error MinimalForwarderInvalidNonce(address signer, uint256 current);
+    error MinimalForwaderMismatchedValue(uint256 msgValue, uint256 value);
 
-    constructor() EIP712("MinimalForwarder", "0.0.1") {}
+    /**
+     * @dev The list of requests length doesn't match with the list of signatures length.
+     */
+    error MinimalForwaderInvalidBatchLength(uint256 requestsLength, uint256 signaturesLength);
 
-    function getNonce(address from) public view returns (uint256) {
-        return _nonces[from];
+    /**
+     * @dev The request `deadline` has expired.
+     */
+    error MinimalForwarderExpiredRequest(uint256 deadline);
+
+    /**
+     * @dev See {EIP7712-constructor}.
+     */
+    constructor(string memory name, string memory version) EIP712(name, version) {}
+
+    /**
+     * @dev Returns `true` if a request is valid for a provided `signature` at the current block.
+     */
+    function verify(ForwardRequest calldata request, bytes calldata signature) public view virtual returns (bool) {
+        (bool alive, bool signerMatch, bool nonceMatch) = _validate(request, signature);
+        return alive && signerMatch && nonceMatch;
     }
 
-    function verify(ForwardRequest calldata req, bytes calldata signature) public view returns (bool) {
-        address signer = _recover(req, signature);
-        (bool correctFrom, bool correctNonce) = _validateReq(req, signer);
-        return correctFrom && correctNonce;
-    }
-
+    /**
+     * @dev Executes a `request` on behalf of `signature`'s signer.
+     */
     function execute(
-        ForwardRequest calldata req,
+        ForwardRequest calldata request,
         bytes calldata signature
-    ) public payable returns (bool, bytes memory) {
-        address signer = _recover(req, signature);
-        (bool correctFrom, bool correctNonce) = _validateReq(req, signer);
-
-        if (!correctFrom) {
-            revert MinimalForwarderInvalidSigner(signer, req.from);
-        }
-        if (!correctNonce) {
-            revert MinimalForwarderInvalidNonce(signer, _nonces[req.from]);
+    ) public payable virtual returns (bool, bytes memory) {
+        if (request.deadline < block.number) {
+            revert MinimalForwarderExpiredRequest(request.deadline);
         }
 
-        _nonces[req.from] = req.nonce + 1;
+        address signer = _recoverForwardRequestSigner(request, signature);
+        if (signer != request.from) {
+            revert MinimalForwarderInvalidSigner(signer, request.from);
+        }
 
-        (bool success, bytes memory returndata) = req.to.call{gas: req.gas, value: req.value}(
-            abi.encodePacked(req.data, req.from)
+        if (msg.value != request.value) {
+            revert MinimalForwaderMismatchedValue(msg.value, request.value);
+        }
+
+        _useCheckedNonce(request.from, request.nonce);
+
+        // As a consequence of EIP-150, the maximum gas forwarded to a call is 63/64 of the remaining gas. So:
+        // - At most `gasleft() - floor(gasleft() / 64)` is passed.
+        // - At least `floor(gasleft() / 64)` is kept in the caller.
+        // The current gas available is saved for later checking.
+        uint256 gasBefore = gasleft();
+        (bool success, bytes memory returndata) = request.to.call{gas: request.gas, value: request.value}(
+            abi.encodePacked(request.data, request.from)
         );
 
-        // Validate that the relayer has sent enough gas for the call.
-        // See https://ronan.eth.limo/blog/ethereum-gas-dangers/
-        if (gasleft() <= req.gas / 63) {
+        // To avoid gas griefing attacks, as referenced in https://ronan.eth.limo/blog/ethereum-gas-dangers/
+        // A malicious relayer can attempt to manipulate the gas forwarded so that the underlying call reverts and
+        // the top-level call still passes.
+        // Such manipulation can be prevented by checking if `gasleft() < floor(gasBefore / 64)`. If so, we
+        // can assume an out of gas error was forced in the subcall. There's no need to process such transactions.
+        if (gasleft() < gasBefore / 64) {
             // We explicitly trigger invalid opcode to consume all gas and bubble-up the effects, since
             // neither revert or assert consume all gas since Solidity 0.8.0
             // https://docs.soliditylang.org/en/v0.8.0/control-structures.html#panic-via-assert-and-error-via-require
@@ -88,17 +119,50 @@ contract MinimalForwarder is EIP712 {
         return (success, returndata);
     }
 
-    function _recover(ForwardRequest calldata req, bytes calldata signature) internal view returns (address) {
-        return
-            _hashTypedDataV4(
-                keccak256(abi.encode(_TYPEHASH, req.from, req.to, req.value, req.gas, req.nonce, keccak256(req.data)))
-            ).recover(signature);
+    /**
+     * @dev Validates if the provided request can be executed at `blockNumber` with `signature`.
+     */
+    function _validateAt(
+        uint256 blockNumber,
+        ForwardRequest calldata request,
+        bytes calldata signature
+    ) internal view virtual returns (bool alive, bool signerMatch, bool nonceMatch) {
+        address signer = _recoverForwardRequestSigner(request, signature);
+        return (request.deadline >= blockNumber, signer == request.from, nonces(request.from) == request.nonce);
     }
 
-    function _validateReq(
-        ForwardRequest calldata req,
-        address signer
-    ) internal view returns (bool correctFrom, bool correctNonce) {
-        return (signer == req.from, _nonces[req.from] == req.nonce);
+    /**
+     * @dev Same as {_validateAt} but for the current block.
+     */
+    function _validate(
+        ForwardRequest calldata request,
+        bytes calldata signature
+    ) internal view virtual returns (bool alive, bool signerMatch, bool nonceMatch) {
+        return _validateAt(block.number, request, signature);
+    }
+
+    /**
+     * @dev Recovers the signer of an EIP712 message hash for a forward `request`
+     * and its corresponding `signature`. See {ECDSA-recover}.
+     */
+    function _recoverForwardRequestSigner(
+        ForwardRequest calldata request,
+        bytes calldata signature
+    ) internal view returns (address) {
+        return
+            _hashTypedDataV4(
+                keccak256(
+                    abi.encode(
+                        _FORWARD_REQUEST_TYPEHASH,
+                        request.from,
+                        request.to,
+                        request.value,
+                        request.gas,
+                        request.nonce,
+                        request.deadline,
+                        keccak256(request.data)
+                    )
+                )
+            ).recover(signature);
     }
 }
--- a/test/metatx/MinimalForwarder.test.js
+++ b/test/metatx/MinimalForwarder.test.js
@@ -3,7 +3,7 @@ const Wallet = require('ethereumjs-wallet').default;
 const { getDomain, domainType } = require('../helpers/eip712');
 const { expectRevertCustomError } = require('../helpers/customError');
 
-const { constants, expectRevert } = require('@openzeppelin/test-helpers');
+const { constants, expectRevert, time } = require('@openzeppelin/test-helpers');
 const { expect } = require('chai');
 
 const MinimalForwarder = artifacts.require('MinimalForwarder');
@@ -11,7 +11,7 @@ const CallReceiverMock = artifacts.require('CallReceiverMock');
 
 contract('MinimalForwarder', function (accounts) {
   beforeEach(async function () {
-    this.forwarder = await MinimalForwarder.new();
+    this.forwarder = await MinimalForwarder.new('MinimalForwarder', '0.0.1');
 
     this.domain = await getDomain(this.forwarder);
     this.types = {
@@ -22,6 +22,7 @@ contract('MinimalForwarder', function (accounts) {
         { name: 'value', type: 'uint256' },
         { name: 'gas', type: 'uint256' },
         { name: 'nonce', type: 'uint256' },
+        { name: 'deadline', type: 'uint48' },
         { name: 'data', type: 'bytes' },
       ],
     };
@@ -34,43 +35,50 @@ contract('MinimalForwarder', function (accounts) {
       value: web3.utils.toWei('1'),
       nonce: 1234,
       data: '0x1742',
+      deadline: 0xabcdef,
     };
 
     beforeEach(async function () {
       this.wallet = Wallet.generate();
       this.sender = web3.utils.toChecksumAddress(this.wallet.getAddressString());
-      this.req = {
+      this.blockNumber = await time.latestBlock();
+      this.request = {
         from: this.sender,
         to: constants.ZERO_ADDRESS,
         value: '0',
         gas: '100000',
-        nonce: Number(await this.forwarder.getNonce(this.sender)),
+        nonce: Number(await this.forwarder.nonces(this.sender)),
         data: '0x',
+        deadline: this.blockNumber.toNumber() + 2, // Next + 1
       };
-      this.forgeData = req => ({
+      this.forgeData = request => ({
         types: this.types,
         domain: this.domain,
         primaryType: 'ForwardRequest',
-        message: { ...this.req, ...req },
+        message: { ...this.request, ...request },
       });
-      this.sign = req =>
+      this.sign = request =>
         ethSigUtil.signTypedMessage(this.wallet.getPrivateKey(), {
-          data: this.forgeData(req),
+          data: this.forgeData(request),
         });
     });
 
     context('verify', function () {
       context('valid signature', function () {
         beforeEach(async function () {
-          expect(await this.forwarder.getNonce(this.req.from)).to.be.bignumber.equal(web3.utils.toBN(this.req.nonce));
+          expect(await this.forwarder.nonces(this.request.from)).to.be.bignumber.equal(
+            web3.utils.toBN(this.request.nonce),
+          );
         });
 
         it('success', async function () {
-          expect(await this.forwarder.verify(this.req, this.sign())).to.be.equal(true);
+          expect(await this.forwarder.verify(this.request, this.sign())).to.be.equal(true);
         });
 
         afterEach(async function () {
-          expect(await this.forwarder.getNonce(this.req.from)).to.be.bignumber.equal(web3.utils.toBN(this.req.nonce));
+          expect(await this.forwarder.nonces(this.request.from)).to.be.bignumber.equal(
+            web3.utils.toBN(this.request.nonce),
+          );
         });
       });
 
@@ -86,13 +94,22 @@ contract('MinimalForwarder', function (accounts) {
         it('returns false with tampered signature', async function () {
           const tamperedsign = web3.utils.hexToBytes(this.sign());
           tamperedsign[42] ^= 0xff;
-          expect(await this.forwarder.verify(this.req, web3.utils.bytesToHex(tamperedsign))).to.be.equal(false);
+          expect(await this.forwarder.verify(this.request, web3.utils.bytesToHex(tamperedsign))).to.be.equal(false);
         });
 
         it('returns false with valid signature for non-current nonce', async function () {
           const req = {
-            ...this.req,
-            nonce: this.req.nonce + 1,
+            ...this.request,
+            nonce: this.request.nonce + 1,
+          };
+          const sig = this.sign(req);
+          expect(await this.forwarder.verify(req, sig)).to.be.equal(false);
+        });
+
+        it('returns false with valid signature for expired deadline', async function () {
+          const req = {
+            ...this.request,
+            nonce: this.request.nonce + 1,
           };
           const sig = this.sign(req);
           expect(await this.forwarder.verify(req, sig)).to.be.equal(false);
@@ -103,16 +120,18 @@ contract('MinimalForwarder', function (accounts) {
     context('execute', function () {
       context('valid signature', function () {
         beforeEach(async function () {
-          expect(await this.forwarder.getNonce(this.req.from)).to.be.bignumber.equal(web3.utils.toBN(this.req.nonce));
+          expect(await this.forwarder.nonces(this.request.from)).to.be.bignumber.equal(
+            web3.utils.toBN(this.request.nonce),
+          );
         });
 
         it('success', async function () {
-          await this.forwarder.execute(this.req, this.sign()); // expect to not revert
+          await this.forwarder.execute(this.request, this.sign()); // expect to not revert
         });
 
         afterEach(async function () {
-          expect(await this.forwarder.getNonce(this.req.from)).to.be.bignumber.equal(
-            web3.utils.toBN(this.req.nonce + 1),
+          expect(await this.forwarder.nonces(this.request.from)).to.be.bignumber.equal(
+            web3.utils.toBN(this.request.nonce + 1),
           );
         });
       });
@@ -133,33 +152,54 @@ contract('MinimalForwarder', function (accounts) {
           const tamperedSig = web3.utils.hexToBytes(this.sign());
           tamperedSig[42] ^= 0xff;
           await expectRevertCustomError(
-            this.forwarder.execute(this.req, web3.utils.bytesToHex(tamperedSig)),
+            this.forwarder.execute(this.request, web3.utils.bytesToHex(tamperedSig)),
             'MinimalForwarderInvalidSigner',
-            [ethSigUtil.recoverTypedSignature({ data: this.forgeData(), sig: tamperedSig }), this.req.from],
+            [ethSigUtil.recoverTypedSignature({ data: this.forgeData(), sig: tamperedSig }), this.request.from],
           );
         });
 
         it('reverts with valid signature for non-current nonce', async function () {
           const req = {
-            ...this.req,
-            nonce: this.req.nonce + 1,
+            ...this.request,
+            nonce: this.request.nonce + 1,
           };
           const sig = this.sign(req);
-          await expectRevertCustomError(this.forwarder.execute(req, sig), 'MinimalForwarderInvalidNonce', [
-            this.req.from,
-            this.req.nonce,
+          await expectRevertCustomError(this.forwarder.execute(req, sig), 'InvalidAccountNonce', [
+            this.request.from,
+            this.request.nonce,
           ]);
+        });
+
+        it('reverts with valid signature for expired deadline', async function () {
+          const req = {
+            ...this.request,
+            deadline: this.blockNumber.toNumber() - 1,
+          };
+          const sig = this.sign(req);
+          await expectRevertCustomError(this.forwarder.execute(req, sig), 'MinimalForwarderExpiredRequest', [
+            this.blockNumber.toNumber() - 1,
+          ]);
+        });
+
+        it('reverts with valid signature but mismatched value', async function () {
+          const value = 100;
+          const req = {
+            ...this.request,
+            value,
+          };
+          const sig = this.sign(req);
+          await expectRevertCustomError(this.forwarder.execute(req, sig), 'MinimalForwaderMismatchedValue', [0, value]);
         });
       });
 
       it('bubble out of gas', async function () {
         const receiver = await CallReceiverMock.new();
         const gasAvailable = 100000;
-        this.req.to = receiver.address;
-        this.req.data = receiver.contract.methods.mockFunctionOutOfGas().encodeABI();
-        this.req.gas = 1000000;
+        this.request.to = receiver.address;
+        this.request.data = receiver.contract.methods.mockFunctionOutOfGas().encodeABI();
+        this.request.gas = 1000000;
 
-        await expectRevert.assertion(this.forwarder.execute(this.req, this.sign(), { gas: gasAvailable }));
+        await expectRevert.assertion(this.forwarder.execute(this.request, this.sign(), { gas: gasAvailable }));
 
         const { transactions } = await web3.eth.getBlock('latest');
         const { gasUsed } = await web3.eth.getTransactionReceipt(transactions[0]);


contracts/metatx/MinimalForwarder.sol
```diff
@@ -5,16 +5,21 @@ pragma solidity ^0.8.19;
...
 /**
- * @dev Simple minimal forwarder to be used together with an ERC2771 compatible contract. See {ERC2771Context}.
+ * @dev A minimal implementation of a production-ready forwarder compatible with ERC2771 contracts.
  *
- * MinimalForwarder is mainly meant for testing, as it is missing features to be a good production-ready forwarder. This
- * contract does not intend to have all the properties that are needed for a sound forwarding system. A fully
- * functioning forwarding system with good properties requires more complexity. We suggest you look at other projects
- * such as the GSN which do have the goal of building a system like that.
+ * This forwarder operates on forward request that include:
+ * * `from`: An address to operate on behalf of. It is required to be equal to `msg.sender`.
```
frangio, line 14: >It is required to be equal to `msg.sender`.

This seems like a typo, I think you mean that it should be equal to the signer.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -5,16 +5,21 @@ pragma solidity ^0.8.19;
...
+ * * `to`: An address destination to call within the request.
+ * * `value`: The amount of ETH to attach within the requested call.
+ * * `gas`: The amount of gas limit that will be forwarded within the requested call.
```
frangio, line 17: The word "within" sounds like it is somewhere in the context of the requested call, but what we mean is the requested call itself, so just "with" is clearer.

```suggestion
 * * `to`: The address that should be called.
 * * `value`: The amount of ETH to attach with the requested call.
 * * `gas`: The amount of gas limit that will be forwarded with the requested call.
```

contracts/metatx/MinimalForwarder.sol
```diff
@@ -5,16 +5,21 @@ pragma solidity ^0.8.19;
...
+ * * `nonce`: A unique transaction ordering identifier to avoid replayability and request invalidation.
+ * * `deadline`: A timestamp after which the request is not executable anymore.
+ * * `data`: Encoded `msg.calldata` to send within the requested call.
```
frangio, line 20:
```suggestion
 * * `data`: Encoded `msg.data` to send with the requested call.
```

contracts/metatx/MinimalForwarder.sol
```diff
@@ -23,59 +28,85 @@ contract MinimalForwarder is EIP712 {
...
+    /**
+     * @dev The request `deadline` has expired.
+     */
+    error MinimalForwarderExpiredRequest(uint256 deadline);
 
-    function getNonce(address from) public view returns (uint256) {
-        return _nonces[from];
-    }
+    /**
+     * @dev See {EIP7712-constructor}.
```
frangio, line 61: I assume that EIP7712 is a typo and you meant instead EIP2771?

I don't understand though, there is no EIP2771 constructor.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -23,59 +28,85 @@ contract MinimalForwarder is EIP712 {
...
+     */
+    constructor(string memory name, string memory version) EIP712(name, version) {}
```
frangio, line 63: In the documentation we should explain what the parameters are so the user can understand how to choose them.

I can see value in customizing the name, but I don't know why a user would want control over the version parameter for the contract, and I think it makes more sense for us to control it. We can hardcode it to an initial version `"1"` now, and if in a future version we want to change the behavior of the contract in some incompatible way, bumping this version to the next number could be a nice way to do that.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -23,59 +28,85 @@ contract MinimalForwarder is EIP712 {
...
-     * @dev The request nonce doesn't match with the `current` nonce for the request signer.
+     * @dev The request `value` doesn't match with the `msg.value`.
+     */
+    error MinimalForwaderMismatchedValue(uint256 msgValue, uint256 value);
+
+    /**
+     * @dev The list of requests length doesn't match with the list of signatures length.
      */
-    error MinimalForwarderInvalidNonce(address signer, uint256 current);
+    error MinimalForwaderInvalidBatchLength(uint256 requestsLength, uint256 signaturesLength);
```
frangio, line 53: This error is not used at the moment.

Are we planning to add batching?

contracts/metatx/MinimalForwarder.sol
```diff
@@ -88,17 +119,50 @@ contract MinimalForwarder is EIP712 {
         return (success, returndata);
     }
 
-    function _recover(ForwardRequest calldata req, bytes calldata signature) internal view returns (address) {
+    /**
+     * @dev Validates if the provided request can be executed at `blockNumber` with `signature`.
+     */
+    function _validateAt(
```
frangio, line 125: What was the reasoning to separate `_validate` and `_validateAt`? I would keep just one, to simplify.

contracts/metatx/MinimalForwarder.sol
```diff
@@ -23,59 +28,85 @@ contract MinimalForwarder is EIP712 {
...
+     */
     function execute(
-        ForwardRequest calldata req,
+        ForwardRequest calldata request,
         bytes calldata signature
-    ) public payable returns (bool, bytes memory) {
-        address signer = _recover(req, signature);
-        (bool correctFrom, bool correctNonce) = _validateReq(req, signer);
+    ) public payable virtual returns (bool, bytes memory) {
+        if (request.deadline < block.number) {
```
frangio, line 80: The formula for deadline validity is written twice, once in `execute` and once in `_validateAt`. Those two formulas are in sync now, but I'd prefer if it was written in only one place to avoid the possibility for off-by-one errors.

.changeset/blue-horses-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`MinimalForwarder`: Added `deadline` for expiring transactions and add `msg.value` mismatch check
```
frangio, line 5:
```suggestion
`MinimalForwarder`: Added `deadline` for expiring transactions and added `msg.value` mismatch check.
```

contracts/metatx/MinimalForwarder.sol
frangio: Oh I see now, it was EIP712.

ernestognw:
I think I've finally nailed the `gasLeft < request.gas / 63` and I also added batching.

Still need to adjust the tests (adding more) and I'm also thinking of fuzzing the execution.

contracts/metatx/MinimalForwarder.sol
ernestognw: Not sure if it's a good idea.

The validation consists of three checks:

1. Deadline
2. Signer
3. Nonce

We can keep the same deadline check in a single shared function, but that's not the case for the other two because:

- In case the signer fails, we need to throw an error with the signer. This means that a single shared function for both signature checks should share the signer as well (or we have to recover twice unnecessarily)
- I think it's reasonable to use `_useCheckedNonce`, but that already performs an SLOAD for the nonce. Using a single shared place would require a duplicated SLOAD

We can still have a deadline check function though, but I think it's better to fuzz the `verify` and `execute` functions to check they both work the same. What do you think?

contracts/metatx/MinimalForwarder.sol
ernestognw: None in particular

contracts/metatx/MinimalForwarder.sol
ernestognw: > Are we planning to add batching?

Did we decide not to do so? I think we agreed a while ago.

Batching added.

contracts/metatx/MinimalForwarder.sol
ernestognw: > In the documentation we should explain what the parameters are so the user can understand how to choose them.

Honestly, I don't see how this adds value when the documentation will add a link to the explanation in EIP-712. Considering [this comment's context](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4346#discussion_r1235518479), do you still think an explanation is needed here?

> I can see value in customizing the name, but I don't know why a user would want control over the version parameter for the contract, and I think it makes more sense for us to control it.

Sure thing.

contracts/metatx/MinimalForwarder.sol
frangio: I would rather address this at the code level than with fuzzing or testing.

The concerns you shared are valid though. We could have the check deadline function.

Amxx (commented):
I have a concern with the batch part. Since all the request have their own signature, anyone could extract one of the request from the batch, and execute it alone. I don't think we want that to be possible. 

The nonce would protect out of order execution, but by front running the relaying of the first one, an attacker would disrupt the relaying of the rest. Also, a user may want the request to be executed atomically.

This mostly applies to request that come from the same user. And I realise that it may be usefull for a relayer to batch requests from different users.

So my suggestion would be: 
- To avoid someone batching multiple request from the same user, and hopping they are executed attomically, maybe the `ForwardRequest` should be

```solidity
struct ForwardRequest {
        address from;
        address[] to;
        uint256[] value;
        bytes[] data;
        uint256[] gas;
        uint256 nonce;
        uint48 deadline;
}

**Comment by ernestognw** [comment]
> I have a concern with the batch part. Since all the request have their own signature, anyone could extract one of the request from the batch, and execute it alone. I don't think we want that to be possible.
> 
> The nonce would protect out of order execution, but by front running the relaying of the first one, an attacker would disrupt the relaying of the rest. Also, a user may want the request to be executed atomically.
> 
> This mostly applies to request that come from the same user. And I realise that it may be usefull for a relayer to batch requests from different users.
> 
> So my suggestion would be:
> 
>     * To avoid someone batching multiple request from the same user, and hopping they are executed attomically, maybe the `ForwardRequest` should be
> 
> 
> ```solidity
> struct ForwardRequest {
>         address from;
>         address[] to;
>         uint256[] value;
>         bytes[] data;
>         uint256[] gas;
>         uint256 nonce;
>         uint48 deadline;
> }
> ```

It's true that the current state allows an attacker to frontrun a relayer to force a revert on at least one of the requests, and I think that is the main problem to fix.

In regards to atomicity, you're right but I'm not sure if that requirement should fit into this contract (at least at the moment) since we don't have feedback yet. In any case, batching is a scalability strategy for the relayer more than an atomicity guarantee for a user.

According to what we discussed internally, we'll allow batching to continue even if there's an already processed nonce.

File: contracts/metatx/MinimalForwarder.sol [hunk 10]
```diff
@@ -23,82 +29,202 @@ contract MinimalForwarder is EIP712 {
...
-    ) public payable returns (bool, bytes memory) {
-        address signer = _recover(req, signature);
-        (bool correctFrom, bool correctNonce) = _validateReq(req, signer);
+    ) public payable virtual returns (bool, bytes memory) {
+        (bool success, bytes memory returndata) = _execute(request, signature);
 
-        if (!correctFrom) {
-            revert MinimalForwarderInvalidSigner(signer, req.from);
+        if (msg.value != request.value) {
+            revert MinimalForwarderMismatchedValue(request.value, msg.value);
```
Amxx, line 90: shouldn't we do this check before the `_execute()` call ?

contracts/metatx/MinimalForwarder.sol
```diff
@@ -23,82 +29,202 @@ contract MinimalForwarder is EIP712 {
...
-        return (success, returndata);
+    /**
+     * @dev Validates if the provided request can be executed at current block with `signature` on behalf of `signer`.
+     */
+    function _validate(
+        ForwardRequest calldata request,
+        bytes calldata signature
+    ) internal view virtual returns (bool alive, bool signerMatch, bool nonceMatch) {
+        address signer = _recoverForwardRequestSigner(request, signature);
+        return (request.deadline >= block.number, signer == request.from, nonces(request.from) == request.nonce);
```
Amxx, line 124: Do we have a rational for using `block.number` for the deadline ? Historically we always did that, but we also now today that they are not really predicable, particularly on L2s. Should we use block.timestamp instead ?

test/metatx/ERC2771Context.test.js
```diff
@@ -12,8 +12,10 @@ const ContextMockCaller = artifacts.require('ContextMockCaller');
 const { shouldBehaveLikeRegularContext } = require('../utils/Context.behavior');
 
 contract('ERC2771Context', function (accounts) {
+  const MAX_UINT48 = new BN('2').pow(new BN('48')).sub(new BN('1')).toString();
```
Amxx, line 15: IMO that is more readable (no need for all the `new BN` 
```suggestion
  const MAX_UINT48 = web3.utils.toBN(1).shln(48).subn(1).toString();
```

test/metatx/MinimalForwarder.test.js
```diff
@@ -22,150 +31,331 @@ contract('MinimalForwarder', function (accounts) {
...
+    this.alice = Wallet.generate();
+    this.alice.address = web3.utils.toChecksumAddress(this.alice.getAddressString());
+
+    this.blockNumber = await time.latestBlock();
+    this.request = {
+      from: this.alice.address,
+      to: constants.ZERO_ADDRESS,
+      value: '0',
+      gas: '100000',
+      nonce: Number(await this.forwarder.nonces(this.alice.address)),
```
Amxx, line 48: In [ERC2771Context.test.js](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4346/files?diff=split&w=0#diff-296c2f0ca2b388a964973951bfce6a90eefa2b24df65564389dd7f944b6d896e) you do `toString()` and here number. IMO we should do the same everywhere.

contracts/metatx/MinimalForwarder.sol
ernestognw: I like the idea but I think it makes sense to block because transactions are generally block-number sensitive (eg. Flashbots). I don't see a major reason to keep `block.number` so I'd go for `block.timestamp` since ERC20Permit uses timestamp as well.

Shall we consider a `clock()`? CC @frangio

contracts/metatx/MinimalForwarder.sol
ernestognw: Yes, the ordering is result from multiple maneuvering I've been doing here. Changing.

test/metatx/ERC2771Context.test.js
ernestognw: Ahh, perfect, thanks!

contracts/metatx/MinimalForwarder.sol
ernestognw: > I would rather address this at the code level than with fuzzing or testing.

Same but I don't see an easy way now that we are even checking different things in the batch version vs the single version. We're not checking nonce correctness in the batch version so now the checks are even more distributed across different functions

ernestognw: Ok I did this in `_validate`:

```solidity
function _validate(
    ForwardRequest calldata request,
    bytes calldata signature
) internal view virtual returns (bool alive, bool signerMatch, bool nonceMatch, address signer) {
    signer = _recoverForwardRequestSigner(request, signature);
    return (
        request.deadline >= block.number,
        signer == request.from,
        nonces(request.from) == request.nonce,
        signer
    );
}
```

Let me know if you like it, I think it looks pretty good tho

contracts/metatx/MinimalForwarder.sol
Amxx: I'm not sure how flashbots affect this. My understanding is that if you put a deadline its because you don't the message to sit around for weeks, and execute when the context is not favorable (for example a buy order at a price that is ok now, but that may be very low in 3 months).

When you want your messaged to be relayed, flashbot can happen before or after ... just like a normal transaction. I don't see how they are relevant here. Even if they are, how do you know in which block number a flashbot will operate ?

I may not see the full picture, but IMO the deadline is mostly a user thing: "if that is not done by sunday, then forget it". In that regard, timestamps are way better IMO.

Not sure about the clock. It would work, but it would also require the UI that prepare the request to do clock detection.

frangio: Agree on using timestamps because they have better usability. Using block number does not mitigate MEV as far as I can tell (it's still possible to frontrun and reorder within one block).

I don't see a need for a customizable clock either, and I agree it would make it harder for tooling since they would have to detect what clock is being used.

.changeset/blue-horses-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`ERC2771Forwarder`: Added `deadline` for expiring transactions and added `msg.value` mismatch check.
```
frangio, line 5:
```suggestion
`ERC2771Forwarder`: Added `deadline` for expiring transactions, batching, and more secure handling of `msg.value`.
```

contracts/metatx/MinimalForwarder.sol
ernestognw: > I'm not sure how flashbots affect this.

It doesn't, what I say is that transactions might be targeted to a block. For example, a request signer might want to limit its request to a certain block because there's an MEV concern in further blocks. For such cases, a timestamp doesn't offer the same guarantees.

Although I generally agree it's an edge case and timestamps are fine.

I'll update with the timestamp

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,240 @@
...
+ * * `value`: The amount of ETH to attach with the requested call.
+ * * `gas`: The amount of gas limit that will be forwarded with the requested call.
+ * * `nonce`: A unique transaction ordering identifier to avoid replayability and request invalidation.
+ * * `deadline`: A timestamp after which the request is not executable anymore.
+ * * `data`: Encoded `msg.data` to send with the requested call.
+ */
+contract ERC2771Forwarder is EIP712, Nonces {
+    using ECDSA for bytes32;
+
+    struct ForwardRequest {
```
frangio, line 26: There's some redundancy in this struct. It doesn't need to be the same as the contents of the typehash. Since this impacts the size and cost of calldata that the relayer will submit when invoking `forward`, it makes sense to try to reduce the size of the struct.

The one thing that I see that can be removed is the nonce, because it's implicit with `from`. Given `from`, we can fetch the expected nonce from `Nonces`.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,240 @@
...
+        uint256 value;
+        uint256 gas;
+        uint256 nonce;
+        uint48 deadline;
+        bytes data;
+    }
+
+    bytes32 private constant _FORWARD_REQUEST_TYPEHASH =
+        keccak256(
+            "ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,uint48 deadline,bytes data)"
```
frangio, line 38: I believe we agreed to add `address relayer` here. My proposal is to add this field, if it has a zero address anyone is allowed to relay it (include it in a `forward` function call), and if it has a non-zero address then it must be equal to the relayer (`msg.sender`).

Note that it doesn't need to be included in the struct, because we already have `msg.sender` available, but this forces us to try both `msg.sender` and `address(0)` as possible values, which will incur some additional cost so we need to make sure to implement it efficiently.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,240 @@
...
+
+        (bool success, bytes memory returndata) = _execute(request, signature);
+
+        return (success, returndata);
+    }
+
+    /**
+     * @dev Batch version of {execute}.
+     */
+    function executeBatch(ForwardRequest[] calldata requests, bytes[] calldata signatures) public payable virtual {
```
frangio, line 102: Is it necessary to submit this as two separate arrays? Could we do a single array with a struct inside with request + signature?

contracts/metatx/ERC2771Forwarder.sol
frangio: After discussing further, we agree to disregard the frontrunning (which only becomes an issue if relayers are incentivized -- not the case here), and not add the relayer parameter to forward requests.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,238 @@
...
+        (bool success, bytes memory returndata) = _execute(request, _useNonce(request.from), true);
+
+        return (success, returndata);
+    }
+
+    /**
+     * @dev Batch version of {execute} that also includes an `atomic` parameter to indicate whether all requests are
+     * executed or none of them.
+     */
+    function executeBatch(ForwardRequestData[] calldata requests, bool atomic) public payable virtual {
```
Amxx, line 101: I think we need to clearly document what the atomic boolean does, and what it doesn't do. 

I see the point for a relayer to use "false", so that the the rest of the batch passes if one of the request is invalid/expired/...

If you use true, you don't have such a clear upside. Yes your tx will fail if one the the request is not alive/signed ... but its still possible for some attacker to partially execute the requests. Said otherwise, atomic is a "best effort" that this tx will try do stuff atomically, but not a security guarantee that an attacker can't do a partial execution.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * @dev The requested `value` doesn't match with the available `msgValue`, leaving ETH stuck in the contract.
+     */
+    error ERC2771ForwarderMismatchedValue(uint256 value, uint256 msgValue);
+
+    /**
+     * @dev The list of requests length doesn't match with the list of signatures length.
+     */
+    error ERC2771ForwarderInvalidBatchLength(uint256 requestsLength, uint256 signaturesLength);
```
frangio, line 59: This error is no longer used. We need a linter to detect unused errors.

contracts/metatx/ERC2771Forwarder.sol
ernestognw: Given that now `ForwardRequest` is not the same as `ForwardRequestData`, we agreed to include the signature along with the request data.

ernestognw:
I've just realized the contract can leave ETH stuck by using {executeBatch}. This is because by optionally executing the valid requests, there may be ETH not sent. I'm considering the following two alternatives:

1. Guarantee the contract holds always 0 value by optionally returning the whole balance back to the `msg.sender` (or reverting) and adding a note to consider this when extending the contract
2. Return only the difference between the consumed value and the provided value for the requests

I don't like number 1 because it cuts the possibility of extending the contract to hold ETH and the problem with the second option is that there's no way to know exactly how much ETH to refund because the user can inflate the `msg.value` to extract value out of the contract if it's ever extended, so it comes with the cost of also counting the total requests' value in a loop.

frangio:
I see. The options seem complex, we may just want to remove the value field if it's hard to handle. To be honest it's difficult to see a use case for the relayer giving money to a requests beyond gas costs. I'm looking at [ERC-4337](https://eips.ethereum.org/EIPS/eip-4337#definitions) UserOperations and they don't seem to contain a `value` field, so I feel that this would justify a removal.

contracts/metatx/MinimalForwarder.sol
frangio: The new `_validate` looks pretty good. The nonces are nicely handled.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,276 @@
...
+                // some point due to insufficient balance.
+                unchecked {
+                    spentValue += requests[i].value;
+                }
+            }
+        }
+
+        if (requestsValue != spentValue) {
+            if (refundReceiver == address(0)) {
+                revert ERC2771ForwarderMismatchedValue(requestsValue, msg.value);
```
frangio, line 134: I think this is missing:

```suggestion
                revert ERC2771ForwarderMismatchedValue(requestsValue, spentValue + msg.value);
```

contracts/metatx/ERC2771Forwarder.sol
ernestognw: I agree, I increased the documentation around this and added notes across the contract to indicate how the flag affects other values (eg. those emitted in an event). I also added a note indicating that the atomicity is only guaranteed for the first-level subcalls.

contracts/metatx/ERC2771Forwarder.sol
ernestognw: No, `msg.value` is always the same during the whole call, what changes is the `address(this).balance` instead.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,278 @@
...
+
+        for (uint256 i; i < requests.length; ++i) {
+            requestsValue += requests[i].value;
+            bool success = _execute(requests[i], _useNonce(requests[i].from), atomic);
+            if (success) {
+                // Will never reallistically overflow given _execute will natively revert at
+                // some point due to insufficient balance.
+                unchecked {
+                    spentValue += requests[i].value;
+                }
```
Amxx, line 130:
```suggestion
            bool success = _execute(requests[i], _useNonce(requests[i].from), atomic);
            unchecked {
                // Will never reallistically overflow given _execute will natively revert at
                // some point due to insufficient balance.
                requestsValue += requests[i].value;
                spentValue += success ? requests[i].value : 0;
            }
```

contracts/metatx/ERC2771Forwarder.sol
Amxx: There used to be a check that `requestsValue` is equal to `msg.value` but now its gone.

We should keep in mind that no this contract can still get some value trough `selfdestruct` (or its replacement) or through block reward. If that ever happens, the forwarder has a balance.

In that case its possible to send a batch that has some `requestsValue` (the fowarder balance) and no `msg.value`. If that case the call reverts, `spentValue` is 0, and the caller can "drain" the forwarder. Is that something we want ?

Otherwize we can: 
- Not track `requestsValue`
- Track `spentValue`
- if `spendValue < msg.value` then "refund" the difference (`mg.value - spendValue`).

In that case the funds send through selfdestruct and block rewards are un-recoverable.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,278 @@
...
+        return
+            _hashTypedDataV4(
+                keccak256(
+                    abi.encode(
+                        _FORWARD_REQUEST_TYPEHASH,
+                        request.from,
+                        request.to,
+                        request.value,
+                        request.gas,
+                        nonce,
```
Amxx, line 178: I honestly think we should get `request.from`'s nounce here instead of passing it. 

Passing it might save one sload duplication when the nonce is consumed, but it also makes the function parameters more complex to follow IMO.

I'd value simplicity over a 100gas saving in an already super expensive operation.

Amxx (commented):
See comments:
- msg.value should be considered IMO
- nonce reading

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,278 @@
...
+        ForwardRequestData[] calldata requests,
+        bool atomic,
+        address payable refundReceiver
+    ) public payable virtual {
+        uint256 requestsValue;
+        uint256 spentValue;
+
+        for (uint256 i; i < requests.length; ++i) {
+            requestsValue += requests[i].value;
+            bool success = _execute(requests[i], _useNonce(requests[i].from), atomic);
```
Amxx, line 124: Here, if atomic is false:
- we first use the nonce
- we go inside the `_execute` function
  - inside `_execute`, `requireValidRequest` is false, so we don't check the "liveness" of the request.
    if the request is not live we do nothing
    
This means that we used the nonce of `requests[i].from` even though the request might not have been signed. It feels like this allows anyone to invalidate anyone elses nonce. This, combined with frontrunning is really bad. You could DOS someone, of worst. 

Lets say Alice wants to do operations A, B, C and D. I like B and D (they are to my advantage) but not A and C.
I could forge two invalid requests, with Alice as the from, that I would call A2 and C2. I then try to batch A2, B, C2, D with atomic to set to false. B and D will be executed, while A2 and C2 will increase the nonce without doing anything.

Amxx: I'm proposing this fix: https://github.com/Amxx/openzeppelin-contracts/commit/b2c8ec406333b2fb274e7995b5f28d814183506c

contracts/metatx/ERC2771Forwarder.sol
frangio: @ernestognw Yes, my bad.

---

It's fine IMO if funds received through selfdestruct/rewards are unrecoverable.

ernestognw: > There used to be a check that requestsValue is equal to msg.value but now its gone.

It's gone because in the even of an skipped request (or many), we don't know exactly how much to refund. Instead, we need to track both consumed and requested value to refund correctly.

> We should keep in mind that no this contract can still get some value trough selfdestruct (or its replacement) or through block reward. If that ever happens, the forwarder has a balance.

The intention is not that the contract holds 0 value, but rather to make the internal accounting works even if it's overridden to hold value.

> In that case its possible to send a batch that has some requestsValue (the fowarder balance) and no msg.value. If that case the call reverts, spentValue is 0, and the caller can "drain" the forwarder. Is that something we want ?

Definitely not, but is the same as if we use `msg.value` instead. Someone can drain the contract using the same logic and that's why I decided to use this instead. Now I see it's also not viable.

> In that case the funds send through selfdestruct and block rewards are un-recoverable.

They're still recoverable (and any other balance can be stolen). Imagine the contract has 1 ether in balance and sends a batch of requests with a total value of 1 ether. If none of them executes and the relayer sends a msg.value of 2 ether, the `msg.value - spendValue` will be 1 ether, thus extracting value out of the contract.

contracts/metatx/ERC2771Forwarder.sol
ernestognw: No. Putting `requestsValue` in an unchecked block allows a relayer to send a signed invalid request (skipped) with `type(uint256).max` and overflow the `requestsValue`.

contracts/metatx/ERC2771Forwarder.sol
ernestognw: We ended up removing the nonce for optimizing `calldata`. It's not worth keeping 2 SLOADS when we can add the nonce to `calldata` and validate. I'll revert back to those changes.

contracts/metatx/ERC2771Forwarder.sol
Amxx: What is the goal here? Currently value can be extracted from the contract. 

If the goal is to refund the value that was passed and not relayed, I believe we can do that more precisely and without tracking the requested value (using the difference between what was passed through msg.value and what was passed to successful calls)

If the goal is different, then I don't get it.

ernestognw: The goal is to correctly refund. Can you clarify how using `msg.value` achieved that since I already proved value can be extracted as well?

contracts/metatx/ERC2771Forwarder.sol
frangio: Comparing gas costs we see that 1 redundant SLOAD costs 100 gas, whereas adding nonce in calldata costs 150 gas. In rollups that post calldata to L1 the different will likely be greater in favor of using storage.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,284 @@
...
+     * will receive the requested gas and no ETH is left stuck in the contract.
+     *
+     * Requirements:
+     *
+     * - The request value should be equal to the provided `msg.value`.
+     * - The request should be valid according to {verify}.
+     */
+    function execute(ForwardRequestData calldata request) public payable virtual returns (bool) {
+        // This check can be before the call because _execute reverts with an invalid request.
+        // Otherwise, `msg.value` will be left in the contract's balance.
```
frangio, line 96: I find this comment confusing, I don't think it makes a difference where we put this check, does it?

ernestognw: I also don't think it makes a difference but @Amxx expressed interest in doing it first:

https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4346#discussion_r1239082628

frangio: It doesn't seem to make a difference to the behavior of the function, but we would still want the check before `_execute` to follow the principle of failing as early as possible. I just don't understand what this comment is saying so I would remove it unless we think it's saying something important then we should rephrase it.

ernestognw: The comment is here because the check depends on the `_execute` having `requireValidRequest = true`, so I wanted to express that dependency that's not enforceable in code.

ernestognw: Also, the check in `executeBatch` can also be done before, are we interested in consistency and following the early fail principle?

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,286 @@
...
+
+    /**
+     * @dev The request `from` doesn't match with the recovered `signer`.
+     */
+    error ERC2771ForwarderInvalidSigner(address signer, address from);
+
+    /**
+     * @dev The requested `value` doesn't match with the available `msgValue`.
+     */
+    error ERC2771ForwarderMismatchedValue(uint256 value, uint256 msgValue);
```
frangio, line 59:
```suggestion
    error ERC2771ForwarderMismatchedValue(uint256 requestsValue, uint256 msgValue);
```

?

ernestognw: Let's do `requestedValue` since it applies to both single and batched requests

contracts/metatx/ERC2771Forwarder.sol
frangio:
```suggestion
        // We make sure that msg.value and request.value match exactly.
        // If the request is invalid or the call reverts, requireValidRequest ensures this value won't be lost.
```

frangio: >the check in executeBatch can also be done before

We need to iterate through the requests array to accumulate `requestsValue` before we can do the check in `executeBatch`, so I think it makes sense where it is now.

ernestognw: > We need to iterate through the requests array to accumulate requestsValue before we can do the check in executeBatch

Correct

> so I think it makes sense where it is now.

Sounds good.

I just want to mention that the for loop can be split in two without affecting gas costs (change would be marginal) and we might want to favor the early fail.

Amxx:
This looks good to me. Let me know when that PR moves out of draft and it requires a formal approval.

ernestognw:
> This looks good to me. Let me know when that PR moves out of draft and it requires a formal approval.

Thanks! I didn't move it before because there are some issues with Slither but seems like false positives from my side

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,286 @@
...
+     * is provided.
+     */
+    function verify(ForwardRequestData calldata request) public view virtual returns (bool) {
+        (bool alive, bool signerMatch, ) = _validate(request);
+        return alive && signerMatch;
+    }
+
+    /**
+     * @dev Executes a `request` on behalf of `signature`'s signer guaranteeing that the forwarded call
+     * will receive the requested gas and no ETH is left stuck in the contract.
```
frangio, line 87:
```suggestion
     * @dev Executes a `request` on behalf of `signature`'s signer using the ERC-2771 protocol. The gas
     * provided to the requested call may not be exactly the amount requested, but the call will not run
     * out of gas. Will revert if the request is invalid or the call reverts, in this case the nonce is not consumed.
```

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,286 @@
...
+        return _execute(request, true);
+    }
+
+    /**
+     * @dev Batch version of {execute} with optional refunding and atomic execution.
+     *
+     * In case a batch contains at least one invalid request (see {verify}), the
+     * request will be skipped and the `refundReceiver` parameter will receive back the
+     * unused requested value at the end of the execution. This is done to prevent unexpected
+     * reverts when a batch includes a request that was already frontrunned by another relayer.
```
frangio, line 112:
```suggestion
     * unused requested value at the end of the execution. This is done to prevent reverting
     * the entire batch when a request is invalid or has already been submitted.
```

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,286 @@
...
+        // This means that if the subcall runs out of gas we are able to detect that insufficient gas was passed.
+        // We will now also see that req.gas / 63 > gasleft() implies that req.gas >= X * 63 / 64.
+        // The contract guarantees Y <= req.gas, thus gasleft() = X - Y >= X - req.gas.
+        // -    req.gas / 63 > gasleft()
+        // -    req.gas / 63 >= X - req.gas
+        // -    req.gas >= X * 63 / 64
+        //
+        // In other words if req.gas < X * 63 / 64 then req.gas / 63 <= gasleft(), thus if the relayer behaves honestly
+        // the relay does not revert.
+        if (gasleft() < request.gas / 63) {
```
frangio, line 278: We are missing a piece here:

```suggestion
        if (gasleft() < request.gas / 63 && !success) {
```

This is related to where the proof above says "If the subcall ran out of gas".

We need to add success as an argument.

test/metatx/ERC2771Forwarder.test.js
```diff
@@ -0,0 +1,431 @@
...
+      it('success', async function () {
+        expect(await this.forwarder.verify(this.requestData)).to.be.equal(true);
+      });
+
+      afterEach(async function () {
+        expect(await this.forwarder.nonces(this.requestData.from)).to.be.bignumber.equal(
+          web3.utils.toBN(this.requestData.nonce),
+        );
+      });
+    });
```
frangio, line 88: Why not combine all this into a single test?

```suggestion
    it('with valid signature', async function () {
      expect(await this.forwarder.nonces(this.requestData.from)).to.be.bignumber.equal(
        web3.utils.toBN(this.requestData.nonce),
      );
      expect(await this.forwarder.verify(this.requestData)).to.be.equal(true);
      expect(await this.forwarder.nonces(this.requestData.from)).to.be.bignumber.equal(
        web3.utils.toBN(this.requestData.nonce),
      );
    });
```

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,286 @@
...
+                revert ERC2771ForwarderExpiredRequest(request.deadline);
+            }
+
+            if (!signerMatch) {
+                revert ERC2771ForwarderInvalidSigner(signer, request.from);
+            }
+        }
+
+        // Avoid execution instead of reverting in case a batch includes an already executed request
+        if (signerMatch && alive) {
```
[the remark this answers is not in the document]
ernestognw: I went for the second because I think the first is incorrect because of the `else`, correct me if wrong.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,286 @@
...
+        keccak256(
+            "ForwardRequest(address from,address to,uint256 value,uint256 gas,uint256 nonce,uint48 deadline,bytes data)"
+        );
+
+    /**
+     * @dev Emitted when a `ForwardRequest` is executed.
+     *
+     * NOTE: A non successful forwarded request should not be assumed as non out of gas exception because of
+     * {_checkForwardedGas}. Such function doesn't guarantee an out of gas exception won't be thrown, but instead
+     * it guarantees a relayer provided enough gas to cover the signer requested gas.
```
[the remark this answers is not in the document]
ernestognw: Yes but there's no way the relayer provides less gas than the requested if it doesn't run out of gas. I'm accepting the suggestion because I think it's clearer and basically the same idea I had.

With "enough" I was referring as enough for the request.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,286 @@
...
+import "../utils/Address.sol";
+
+/**
+ * @dev An implementation of a production-ready forwarder compatible with ERC2771 contracts.
+ *
+ * This forwarder operates on forward requests that include:
+ *
+ * * `from`: An address to operate on behalf of. It is required to be equal to the request signer.
+ * * `to`: The address that should be called.
+ * * `value`: The amount of ETH to attach with the requested call.
```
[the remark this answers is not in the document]
ernestognw: Should we apply this same term to all other "ETH"s? I decided to use ETH because it felt weird saying native token all over the place

contracts/metatx/ERC2771Forwarder.sol
ernestognw: I am not sure about this, I think the point of the proof was actually proving we don't need to check the success. If it fails, the check will catch it, but if it succeeds there's no way the relayer manipulated the gas provided anyway.

test/metatx/ERC2771Forwarder.test.js
ernestognw: It's just the way it was before but no preference, I added your suggestion

test/metatx/ERC2771Forwarder.test.js
```diff
@@ -0,0 +1,431 @@
...
+          deadline: this.timestamp - 1,
+        };
+        req.signature = this.sign(this.alice.getPrivateKey(), req);
+        expect(await this.forwarder.verify(req)).to.be.equal(false);
+      });
+    });
+  });
+
+  context('execute', function () {
+    context('with valid requests', function () {
```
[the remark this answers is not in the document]
ernestognw: I added an extra test after the discussion we had so this suite no longer uses `afterEach` so I think it makes sense the way it is right now.

contracts/metatx/ERC2771Forwarder.sol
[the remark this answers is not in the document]
frangio: Yes it is wrong haha. (Just realized)

contracts/metatx/ERC2771Forwarder.sol
frangio: I got this idea at some point but I just reviewed the proof and you're correct.

contracts/metatx/ERC2771Forwarder.sol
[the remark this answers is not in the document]
frangio: I updated one more occurence of "ETH" in the user-facing docs. The other one is in a comment so I don't mint if it says ETH.

"Native token" is award, yes. Not sure what other options we have.

frangio (dismissed):
LGTM!

frangio (approved):

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -0,0 +1,289 @@
...
+        // Ignore an invalid request because requireValidRequest = false
+        if (signerMatch && alive) {
+            // Nonce should be used before the call to prevent reusing by reentrancy
+            uint256 currentNonce = _useNonce(signer);
+
+            (success, ) = request.to.call{gas: request.gas, value: request.value}(
+                abi.encodePacked(request.data, request.from)
+            );
+
+            _checkForwardedGas(request);
```
Amxx, line 239: are we sure this is inlined ? If having the private function means there are jumps in the code, that is not great :/

Amxx (approved):
Minor comment about code inlining. Otherwize LGTM.

contracts/metatx/ERC2771Forwarder.sol
frangio: We can inspect the output, but we have no guarantees. We could change this to `_checkForwardedGas(request, gasleft())` to make sure `gasleft()` is evaluated here.

Amxx: if calling a "subroutine" the calldata pointer might be injected using a `push` or a `dup`, right ?

its a few units of gas anyway.

ernestognw: I agree it's particularly difficult to check. I'd be okay inlining the function manually but if I'm not mistaken, in order to "exploit" the small difference, the user's requested gas must be extremely close to the actual gas used so an out-of-gas can be forced in a very tight gas space, which is unlikely to happen.

frangio: I would rater not inline the function because the way it's separated now gives good room to explain the reasoning in lengthy comments.

I would prefer to add `gasleft()` as an argument here, I think that can be justified. @ernestognw what do you think about that?

ernestognw: I agree, let's do that in the next PR #4396

PR #4348, opened 2023-06-14
Change behavior of ceilDiv(0, 0) and improve test coverage

ernestognw:
### Description

Addresses last pending comment in #4261 about testing coverage.

I fixed some uncovered branches but is difficult to check for exhaustivity, and also some new added branches are difficult to reach specially because they're errors.

### DIFF
--- a/.changeset/blue-scissors-design.md
+++ b/.changeset/blue-scissors-design.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Math`: Make `ceilDiv` to revert on 0 division even if the numerator is 0
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -114,6 +114,11 @@ library Math {
      * of rounding down.
      */
     function ceilDiv(uint256 a, uint256 b) internal pure returns (uint256) {
+        if (b == 0) {
+            // Guarantee the same behavior as in a regular Solidity division.
+            return a / b;
+        }
+
         // (a + b - 1) / b can overflow on addition, so we distribute.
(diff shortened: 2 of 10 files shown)


contracts/utils/math/Math.sol
```diff
@@ -114,6 +114,11 @@ library Math {
      * of rounding down.
      */
     function ceilDiv(uint256 a, uint256 b) internal pure returns (uint256) {
+        if (b == 0) {
```
Amxx, line 117: This is more than just test improvement. This is a behavior change.

We used to have `ceilDiv(0,0)` return 0, and now it will revert instead. IMO this requires a changeset.

Amxx (approved):
one small comment

contracts/utils/math/Math.sol
ernestognw: You're right, this deserves a changeset. I just added it. Thanks

frangio (approved):

.changeset/blue-scissors-design.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Math`: Make `ceilDiv` to revert on 0 division even if the numerator is 0
```
Amxx, line 5: Its actually a panic. But this is already merge so whatever.

frangio: I think the changeset is correct, it reverts with a Panic error.

pcaversaccio:
FYI, I got a question here https://github.com/OpenZeppelin/openzeppelin-contracts/commit/247753426034263e07ef5dc7f0cf3483fb6c53b7#r118328369.

PR #4352, opened 2023-06-14
Merge release-v4.9 branch

ernestognw:
Merging changes from release-v4.9


Amxx:
Most of the changeset removal are part of 5.0 and should be kept

test/governance/Governor.test.js
```diff
@@ -701,6 +701,89 @@ contract('Governor', function (accounts) {
         });
       });
 
+      describe('frontrun protection using description suffix', function () {
```
frangio, line 704: I think these tests are now twice in the JS file.

ernestognw:
The merge resolution was kind of difficult. My conclusion is that we can't correctly automate the merge back of this because there are multiple cases in which we need manual resolution and I actually think it's dangerous to even keep this last release cycle automation.

We can discuss it later. Currently is ready for review.

CHANGELOG.md
```diff
@@ -55,10 +55,14 @@ Users that were registering EIP-165 interfaces with `_registerInterface` from `E
 
 ```solidity
-    return interfaceId == type(MyInterface).interfaceId || super.supportsInterface(interfaceId);
+  return interfaceId == type(MyInterface).interfaceId || super.supportsInterface(interfaceId);
```
frangio, line 58: Not sure where this change came from it seems unrelated to the merge.

```suggestion
    return interfaceId == type(MyInterface).interfaceId || super.supportsInterface(interfaceId);
```

ernestognw: It went from formatting the `CHANGELOG.md` after conflict resolution

frangio (approved):
LGTM

PR #4353, opened 2023-06-14
Improve release merge script for deleted files

ernestognw:
## Description

Recently, the [release scripts failed](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/5260126331) trying to merge the `release-v4.9` branch back to master. The reason is that merging with `-X theirs` strategy doesn't solve conflicts for files modified in `release-v4.9` and deleted in `master`, which is the case for multiple contracts we've removed for 5.0

With this change, the conflicts will be solved by removing the files as they're removed in master.


ernestognw:
Put on hold since we may want to remove that part of the workflow. We've realized there are a few edge cases that require manual review and are frequent enough to prefer manual intervention over a fine-tuned automation.

frangio:
Superceded by https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4362.

PR #4354, opened 2023-06-15
Make `TransparentUpgradeableProxy` admin immutable

### LINKED ISSUE #2638: TransparentUpgradeableProxy could be optimized by making admin immutable
**🧐 Motivation**
Currently the Transparent proxy read the admin slot for every call to it. 
Since TransparentUpgradeableProxy are most likely to be managed through a admin proxy contract that itself handle admin changes (Ownable),
it would be great to have a variant of the Transparent proxy where the admin cannot change.
With that you can store the admin both on storage (for compatibility with EIP-1967) and in an immutable variable (so you can read it cheaply).

The Proxy Admin contract will then be assigned indefinitely to that proxy and the admin changes will be handled on the ProxyAdmin only.

**📝 Details**
See an example implementation here : https://github.com/wighawag/hardhat-deploy/blob/4d5b527e07a4742cf97f9648a6e1d2795edbfd22/solc_0.7/proxy/OptimizedTransparentUpgradeableProxy.sol

ernestognw:
Fixes #2638
Fixes LIB-805

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/mocks/DummyImplementation.sol
```diff
@@ -11,6 +14,9 @@ contract DummyImplementation {
     string public text;
     uint256[] public values;
 
+    // bytes32(uint256(keccak256('eip1967.proxy.admin')) - 1)
+    bytes32 internal constant _ADMIN_SLOT = 0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103;
```
ernestognw, line 18: This can be gotten from the `ERC1967Upgrade` library once #4204 is merged

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -46,12 +44,21 @@ interface ITransparentUpgradeableProxy is IERC1967 {
...
+ *
  * WARNING: It is not recommended to extend this contract to add additional external functions. If you do so, the compiler
  * will not check that there are no selector conflicts, due to the note above. A selector clash between any new function
  * and the functions declared in {ITransparentUpgradeableProxy} will be resolved in favor of the new one. This could
  * render the admin operations inaccessible, which could prevent upgradeability. Transparency may also be compromised.
  */
 contract TransparentUpgradeableProxy is ERC1967Proxy {
+    // An immutable address for the admin avoid unnecessary SLOADs before each call
+    // at the expense of removing the ability to change the admin once it's set.
+    address immutable _admin;
```
Amxx, line 60:
```suggestion
    address private immutable _admin;
```

Note: when storage is immutable, we may want to make it internal, but for now private is what our guidelines advocate.

contracts/proxy/ERC1967/ERC1967Upgrade.sol
```diff
@@ -123,7 +123,7 @@ abstract contract ERC1967Upgrade is IERC1967 {
      * https://eth.wiki/json-rpc/API#eth_getstorageat[`eth_getStorageAt`] RPC call.
      * `0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103`
      */
-    function _getAdmin() internal view returns (address) {
+    function _getAdmin() internal view virtual returns (address) {
```
Amxx, line 126: When `ERC1967Upgrade` becomes a library, this will be modifier will have to dissapear.

ernestognw: Yes, in the meantime are we ok?
I'd like to help getting #4204 merged if it becomes a blocker.

Amxx: It really depends which PR gets merged first. I'm ok with your approach, I'm just saying that whichever PR gets merge first, it will create conflicts on the other.

The alternative would be to make this non virtual, and use a different name for the internal getter for the immutable variable (or not have a getter and make the immutable variable internal)

contracts/mocks/DummyImplementation.sol
Amxx:
```suggestion
```

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
```diff
@@ -47,6 +47,25 @@ module.exports = function shouldBehaveLikeTransparentUpgradeableProxy(createProx
     });
   });
 
+  describe('proxyAdmin', function () {
```
frangio, line 50:
```suggestion
  describe('proxy admin', function () {
```

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -95,17 +101,12 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
...
+     * Overrides ERC1967's admin in favor of an immutable value to avoid unnecessary SLOADs on each proxy call.
      */
-    function _dispatchChangeAdmin() private returns (bytes memory) {
-        _requireZeroValue();
-
-        address newAdmin = abi.decode(msg.data[4:], (address));
-        _changeAdmin(newAdmin);
-
-        return "";
+    function _getAdmin() internal view virtual override returns (address) {
```
Amxx, line 108:
```suggestion
    function _getAdmin() internal view virtual returns (address) {
```

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
frangio, line 50: If I'm not mistaken we seem to be missing a test for the event emitted in the constructor.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -46,12 +44,21 @@ interface ITransparentUpgradeableProxy is IERC1967 {
  * fully implement transparency without decoding reverts caused by selector clashes between the proxy and the
  * implementation.
  *
+ * IMPORTANT: This contract avoids unnecessary storage reads by setting the admin only during construction as an immutable variable,
```
frangio, line 47: The docs for the contract need to be fully reviewed beyond this. There is a mention of "changing the admin above". We should emphasize that the admin is immutable, and the recommendation to use ProxyAdmin should now be even stronger, because it will be the only way to later change it.

Honestly I would even consider enforcing this via `extcodesize` but I know this will be pretty controversial. Cc: @Amxx

frangio (commented):
Looks pretty good.

contracts/proxy/ERC1967/ERC1967Upgrade.sol
ernestognw: Solving since #4204 went first. Thanks!

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
ernestognw: Correct. I added an extra test

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
ernestognw: > The docs for the contract need to be fully reviewed beyond this. There is a mention of "changing the admin" above.

I updated the previous part and I think there are no other places to update after checking across the codebase.

> We should emphasize that the admin is immutable, and the recommendation to use ProxyAdmin should now be even stronger, because it will be the only way to later change it.

Added a comment but you may want to review or propose a different layout. I don't want to extend the text much because I already think it's long.

> Honestly I would even consider enforcing this via extcodesize but I know this will be pretty controversial. Cc: @Amxx

I was thinking the same but I decided to avoid it because I don't think it changes anything:

- There's no way to use the `ERC1967Utils.changeAdmin(admin_)` function twice in the constructor within the same contract
- Even if there was a way, the implementation can always brick the contract by setting the admin slot.

Now I wonder if we might need a way to see if a Proxy is "healthy", meaning that the immutable address matches the admin storage slot. That could be done if we make `public` the `_admin` variable. Obviously with a different name or a getter.

Amxx: > Even if there was a way, the implementation can always brick the contract by setting the admin slot.

If the implementation changes the admin slot the contract is not bricked. You "just" need to figure out who the initial admin was (looking at the event emitted in the constructor) and use that. UIs may be broken, but the contract is not.

My understanding was, since its impossible to change, its unsafe to use an EOA for that, so we may want to enforce that the address corresponds to a contract already deployed, hoppinh it is a ProxyAdmin (or a mulsitig?).

At that point I'd just go for deploying our own ProxyAdmin.

Amxx: > That could be done if we make public the _admin variable. Obviously with a different name or a getter.

If you make anything public in the TransparentProxyAdmin, you are basically breaking the transparency. The idea is that ANY function call (that is not done by the admin) would be forwarded.

.changeset/happy-falcons-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`TransparentUpgradeableProxy`: Added an immutable admin set during construction to avoid unnecessary storage reads on every proxy call, and removed the ability to change the admin from both proxy and `ProxyAdmin`.
```
Amxx, line 5: I think splitting the sentence is clearer. Also, I don't sure what you mean by "remove the ability to change the admin from the ProxyAdmin. The ProxyAdmin is still ownable.

```suggestion
`TransparentUpgradeableProxy`: Admin is now stored in an immutable variable (set during construction) to avoid unnecessary storage reads on every proxy call. This removed the ability to ever change the admin. Transfer of the upgrade capability is exclusively handled though the ownership of the `ProxyAdmin`.
```

contracts/mocks/DummyImplementation.sol
```diff
@@ -44,6 +46,12 @@ contract DummyImplementation {
     function reverts() public pure {
         require(false, "DummyImplementation reverted");
     }
+
+    // Use for forcing an unsafe TransparentUpgradeableProxy admin override
+    // solhint-disable-next-line private-vars-leading-underscore
+    function _unsafeOverrideAdmin(address newAdmin) public {
```
Amxx, line 52: this is a mock. Do we really want to mess with the linter and add a rule so that this can start with an `_` ?
I'd just call it `unsafeOverrideAdmin`

.changeset/happy-falcons-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`TransparentUpgradeableProxy`: Admin is now stored in an immutable variable (set during construction) to avoid unnecessary storage reads on every proxy call. This removed the ability to ever change the admin. Transfer of the upgrade capability is exclusively handled though the ownership of the `ProxyAdmin`.
```
ernestognw, line 5:
```suggestion
`TransparentUpgradeableProxy`: Admin is now stored in an immutable variable (set during construction) to avoid unnecessary storage reads on every proxy call. This removed the ability to ever change the admin. Transfer of the upgrade capability is exclusively handled through the ownership of the `ProxyAdmin`.
```

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
Amxx: I'm not sure I get what the overall idea is. We can't expose any getter to anyone but to the admin. Adding a getting that returns the admin, but that only the admin can call doesn't really work for me. 

AFAIK, the only absolutely reliable and immutable way to get the admin is to look at event history.

As for the "contract is bricked", I think its a question of what do you mean by that. I personnaly think of a "bricked contract" as something that you absolutely cannot recover from. Here the "standard workflow" is (at least temporarily) broken, but there exist a (more complex) path toward recovery.

Amxx (approved):

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
ernestognw: > I'm not sure I get what the overall idea is. We can't expose any getter to anyone but to the admin. Adding a getting that returns the admin, but that only the admin can call doesn't really work for me.
 
The overall idea is pretty clear I'd say:

> > Now I wonder if we might need a way to see if a Proxy is "healthy", meaning that the immutable address matches the admin storage slot.

> AFAIK, the only absolutely reliable and immutable way to get the admin is to look at event history.

This is expensive for DApps and an edge case. Taking Defender as an example, it already takes time in the UI to know that the contract features are (including Upgradeability), I think it's inconvenient adding a required step such as reading the events story.

> I think it's a question of what you mean by that.

I don't think it matters, we both agree it's a problem, and that's all I meant.

frangio: Yes it's important to understand changing the slot does not brick the proxy. It's a relatively minor issue.

ernestognw: That's understood since the beginning, but I'm worried the conversation is about the use of "brick" instead of an actual review.

frangio (approved):
LGTM

ernestognw:
Merging since no substantial changes have been made since @Amxx approval

PR #4356, opened 2023-06-15
Move upgradeToAndCallUUPS to UUPSUpgradeable

Amxx:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4325#discussion_r1230335898

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/proxy/ERC1967/ERC1967Utils.sol
```diff
@@ -103,30 +97,6 @@ library ERC1967Utils {
...
 
-    /**
-     * @dev Perform implementation upgrade with security checks for UUPS proxies, and additional setup call.
-     *
-     * Emits an {IERC1967-Upgraded} event.
-     */
-    function upgradeToAndCallUUPS(address newImplementation, bytes memory data, bool forceCall) internal {
-        // Upgrades from old implementations will perform a rollback test. This test requires the new
-        // implementation to upgrade back to the old, non-ERC1822 compliant, implementation. Removing
-        // this special case will break upgrade paths from old UUPS implementation to new ones.
```
frangio, line 114: I think there is a test for this behavior that we can remove. `UUPSLegacy`?

test/proxy/utils/UUPSUpgradeable.test.js
```diff
@@ -71,12 +71,6 @@ contract('UUPSUpgradeable', function () {
     );
 
     const receipt = await legacyInstance.upgradeTo(this.implInitial.address);
-
-    const UpgradedEvents = receipt.logs.filter(
-      ({ address, event }) => address === legacyInstance.address && event === 'Upgraded',
-    );
-    expect(UpgradedEvents.length).to.be.equal(1);
```
frangio, line 78: Why are we removing just this instead of the whole test and the UUPSLegacy mock?

ernestognw: Is there a reason to keep explicit support for the old version in the first place? I see in the `UUPSUpgradeableLegacyMock` that is for contracts pre-4.5.

I see the use case of just using the upgrade utils, so I'm not strong about removing these tests but I'd suggest being clear and keeping the previous lines but checking for `2` events emitted instead.

contracts/proxy/ERC1967/ERC1967Utils.sol
```diff
@@ -5,7 +5,6 @@ pragma solidity ^0.8.20;
 
 import "../beacon/IBeacon.sol";
 import "../../interfaces/IERC1967.sol";
```
ernestognw, line 7: Just realized this is not needed here and we should rather move it to `TransparentUpgradeableProxy`

ernestognw (dismissed):

contracts/proxy/ERC1967/ERC1967Utils.sol
Amxx: The goal is to re-use it as soon as possible. Should we remove + re add, or just leave it

test/proxy/utils/UUPSUpgradeable.test.js
Amxx: I just removed it.

contracts/proxy/ERC1967/ERC1967Utils.sol
ernestognw: > The goal is to re-use it as soon as possible. 

What do you mean by "re-use it"?

> Should we remove + re add, or just leave it

The concrete suggestion would be to remove this import and add it to `TransparentUpgradeableProxy`.

Amxx: With 0.8.21 we would be able to remove the events declared here, and use the ones declared in the interface instead. When that happens, we'll need the import

Amxx: And technically `TransparentUpgradeableProxy` does not need to inherit that. This interface only contains events. By using this library, the compiler will include (in the ABI) the events declared here. In fact inheriting for the interface is something I would discourage, because it will include irrelevant events (`BeaconUpgraded` gets into `TransparentUpgradeableProxy`)

ernestognw: For the sake of clarity, I moved it and pushed a commit. Feel free to revert the commit if you don't think we should do this, but for me is clear that it'd be an information issue during an audit, and we can re-add it at will once 0.8.21 is done.

I'm solving

contracts/proxy/utils/UUPSUpgradeable.sol
```diff
@@ -35,12 +40,10 @@ abstract contract UUPSUpgradeable is IERC1822Proxiable {
...
-        if (address(this) == __self) {
-            // Must be called through delegatecall
-            revert UUPSUnauthorizedCallContext();
-        }
-        if (ERC1967Utils.getImplementation() != __self) {
-            // Must be called through an active proxy
+        if (
+            address(this) == __self || // Must be called through delegatecall
+            ERC1967Utils.getImplementation() != __self // Must be called through an active proxy
+        ) {
```
ernestognw, line 46: I thought it made more sense now the error is the same.

ernestognw (approved):
I just added tests to increase the coverage and pass the Codecov check. LGTM

ernestognw (dismissed):

contracts/proxy/utils/UUPSUpgradeable.sol
Amxx: yes

test/proxy/utils/UUPSUpgradeable.test.js
```diff
@@ -37,13 +40,64 @@ contract('UUPSUpgradeable', function () {
...
+      this.helper.$functionDelegateCall(
+        this.implUpgradeOk.address,
+        this.implUpgradeOk.contract.methods.upgradeTo(this.implUpgradeUnsafe.address).encodeABI(),
+      ),
+      'UUPSUnauthorizedCallContext',
+      [],
+    );
+  });
+
+  it('calling upgradeTo from a contract that is not a proxy reverts', async function () {
```
Amxx, line 78:
```suggestion
  it('calling upgradeToAndCall from a contract that is not a proxy reverts', async function () {
```

test/proxy/utils/UUPSUpgradeable.test.js
```diff
@@ -37,13 +40,64 @@ contract('UUPSUpgradeable', function () {
...
+      this.implInitial.upgradeToAndCall(
+        this.implUpgradeOk.address,
+        this.implUpgradeOk.contract.methods.increment().encodeABI(),
+      ),
+      'UUPSUnauthorizedCallContext',
+      [],
+    );
+  });
+
+  it('calling upgradeTo from a contract that is not a proxy reverts', async function () {
```
Amxx, line 67:
```suggestion
  it('calling upgradeTo from a contract that is not an ERC1967 proxy (with the right implementation) reverts', async function () {
```

test/proxy/utils/UUPSUpgradeable.test.js
```diff
@@ -37,13 +40,64 @@ contract('UUPSUpgradeable', function () {
...
+      this.helper.$functionDelegateCall(
+        this.implUpgradeOk.address,
+        this.implUpgradeOk.contract.methods.upgradeTo(this.implUpgradeUnsafe.address).encodeABI(),
+      ),
+      'UUPSUnauthorizedCallContext',
+      [],
+    );
+  });
+
+  it('calling upgradeToAndCall from a contract that is not a proxy reverts', async function () {
```
Amxx, line 78:
```suggestion
  it('calling upgradeToAndCall from a contract that is not n ERC1967 proxy (with the right implementation) reverts', async function () {
```

test/proxy/utils/UUPSUpgradeable.test.js
```diff
@@ -37,13 +40,64 @@ contract('UUPSUpgradeable', function () {
...
+      this.helper.$functionDelegateCall(
+        this.implUpgradeOk.address,
+        this.implUpgradeOk.contract.methods.upgradeTo(this.implUpgradeUnsafe.address).encodeABI(),
+      ),
+      'UUPSUnauthorizedCallContext',
+      [],
+    );
+  });
+
+  it('calling upgradeToAndCall from a contract that is not n ERC1967 proxy (with the right implementation) reverts', async function () {
```
Amxx, line 78:
```suggestion
  it('calling upgradeToAndCall from a contract that is not an ERC1967 proxy (with the right implementation) reverts', async function () {
```

frangio:
We really need to agree on the guidelines for test descriptions. @Amxx and @ernestognw seem to disagree on how they should be written, based on the commit history here.

frangio (approved):

ernestognw:
> We really need to agree on the guidelines for test descriptions. @Amxx and @ernestognw seem to disagree on how they should be written, based on the commit history here, and past conversations.

I prefer `it("does something` for semantics but I'm not opposed to `test("something`. I just prefer the test to be read as a spec with a list of `it` units because `it` refers to the subject in a corresponding block.

```
Contract: MerkleProof
    verify
      ✔ returns true for a valid Merkle proof (52ms)
      ✔ returns false for an invalid Merkle proof
      ✔ returns false for a Merkle proof of invalid length
```

Using `it('...')` as part of the description makes sense to me because of this

PR #4358, opened 2023-06-16
Add state getter in TimelockController using OperationState enum

### LINKED ISSUE #4316: Add `state` getter in TimelockController using `OperationState` enum
### Description

We introduced an enum of `OperationState`s for custom errors in #4261. However, it feels weird that such enum is only used for custom errors.

We should use the same enum for retrieving the state. We might also want to replace the state getters with this new `state(bytes32 id)` getter.

RenanSouza2:
Fixes #4316
Fixes LIB-899

TimelockController

- remove:
 - isOperationPending
 - isOperationReady
 - isOperationDone

- implement:
  - getOperationState

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/governance/TimelockController.sol
```diff
@@ -169,28 +169,6 @@ contract TimelockController is AccessControl, ERC721Holder, ERC1155Holder {
...
-        return timestamp > _DONE_TIMESTAMP && timestamp <= block.timestamp;
-    }
-
-    /**
-     * @dev Returns whether an operation is done or not.
-     */
-    function isOperationDone(bytes32 id) public view virtual returns (bool) {
-        return getTimestamp(id) == _DONE_TIMESTAMP;
-    }
-
```
Amxx, line 193: Maybe we should keep these function for backward compatibility. Some interfaces may rely on them.

@ernestognw do you know if defender uses that ?

contracts/governance/TimelockController.sol
```diff
@@ -199,6 +177,22 @@ contract TimelockController is AccessControl, ERC721Holder, ERC1155Holder {
         return _timestamps[id];
     }
 
+    /**
+     * @dev Returns operation state.
+     */
+    function getOperationState(bytes32 id) public view virtual returns (OperationState) {
+        uint timestamp = getTimestamp(id);
```
Amxx, line 184: We favor uint256 over uint to achieve maximum clarity. Some people can be confused about what a uint is.
```suggestion
        uint256 timestamp = getTimestamp(id);
```

contracts/governance/TimelockController.sol
```diff
@@ -199,6 +177,22 @@ contract TimelockController is AccessControl, ERC721Holder, ERC1155Holder {
...
+        if (timestamp == 0) {
+            return OperationState.Unset;
+        } else if (getTimestamp(id) > _DONE_TIMESTAMP && timestamp > block.timestamp) {
```
Amxx, line 187: No need to re-do the sload here.
```suggestion
        } else if (timestamp > _DONE_TIMESTAMP && timestamp > block.timestamp) {
```

contracts/governance/TimelockController.sol
```diff
@@ -199,6 +177,22 @@ contract TimelockController is AccessControl, ERC721Holder, ERC1155Holder {
...
+            return OperationState.Pending;
+        } else if (timestamp > _DONE_TIMESTAMP) {
+            return OperationState.Ready;
+        } else {
+            return OperationState.Done;
+        }
+    }
```
Amxx, line 194: We can reduce the number of checks, an avoid a duplicated `getTimestamp` call
```suggestion
        if (timestamp == 0) {
            return OperationState.Unset;
        } else if (timestamp == DONE_TIMESTAMP) {
            return OperationState.Done;
        } else if (timestamp > block.timestamp) {
            return OperationState.Pending;
        } else {
            return OperationState.Ready;
        }
    }
```

Amxx: Note that this slightly changes the semantics:

Before:
- Pending: is going to be executed at some point, may or may not be ready
- Ready: can be executed now

After:
- Pending: scheduled but now yet ready for execution
- Ready: can be executed now


So before Ready implied Pending, and after Ready implies not Pending.
If we keep the old getters, we will need to either change their semantics, or include an inconsistency between what they return and what is returned by state().

contracts/governance/extensions/GovernorTimelockControl.sol
```diff
@@ -60,9 +60,15 @@ abstract contract GovernorTimelockControl is IGovernorTimelock, Governor {
...
+        }
+
+        TimelockController.OperationState _operationState = _timelock.getOperationState(queueid);
+        if (_operationState == TimelockController.OperationState.Done) {
             return ProposalState.Executed;
-        } else if (_timelock.isOperationPending(queueid)) {
+        } else if (
+            _operationState == TimelockController.OperationState.Pending ||
+            _operationState == TimelockController.OperationState.Ready
+        ) {
```
Amxx, line 71: The issue with this change is that the new governors that may be deployed in the future are not going to be compatible with the old TimelockController that have been deployed for a while.

contracts/governance/TimelockController.sol
RenanSouza2: Done

contracts/governance/extensions/GovernorTimelockControl.sol
RenanSouza2: Retrocompatibility is something kept even in breaking cahnges?

contracts/governance/TimelockController.sol
RenanSouza2: I suggest the name 'pending' be replaced by 'blockded' or something similar

contracts/governance/extensions/GovernorTimelockControl.sol
Amxx: Not everything needs to be kept, but I would argue that allowing new governor to use existing TimelockController is important.

contracts/governance/TimelockController.sol
ernestognw: Indeed they use them, though I'm in favor of removing them. I think Defender can just wait to implement 5.0 feature detection. This is something we'll need to discuss next Monday on sync.

frangio:
To reiterate, what is missing in this PR is to add back `isOperationReady`, etc., and I would add: to implement those functions in terms of `getOperationState()` so that `getOperationState` is the "source of truth".

RenanSouza2:
And what is de decision behing the pending name?

I suggest keep the pending meaning of 'operation scheduled but not done or canceled' because of the previous method 'isOperationPending'. And to create another name fot the state between scheduled but before the time alloed to execute like 'blocked'

frangio:
> 1. In line 341 I wasn't sure in what to include in the error because Pending is no longer in the enum

Interesting. The issue is that cancel can be done in two states, and the `TimelockUnexpectedOperationState` error only has a single `OperationState expected` parameter. Should we do the same as with the `GovernorUnexpectedProposalState` error and use a bitmap? @Amxx @ernestognw What do you think?

ernestognw:
> Interesting. The issue is that cancel can be done in two states, and the `TimelockUnexpectedOperationState` error only has a single `OperationState expected` parameter. Should we do the same as with the `GovernorUnexpectedProposalState` error and use a bitmap? @Amxx @ernestognw What do you think?

I think what we did in `GovernorUnexpectedProposalState` is the easiest way of expressing multiple enum values so I agree. Alternatively, I'd use the earliest state (eg. blocked) as the required state because it's first in the lifecycle of an operation.

More in favor if a bitmap anyway.

Amxx:
ok with bitmap

RenanSouza2:
As of now I just copied the '_encodeStateBitmap' from 'Governor' to 'TimelockController' I would like to unify both implementation, maybe in a file in the utils folder in governance itself, but this woul affect too many files so I want to see yout take on that before doing it

frangio:
> I would like to unify both implementation

I don't think we can do this because it's a different enum type.

frangio:
I've renamed `Blocked` to `Waiting` because I felt Blocked could refer to the predecessor logic.

A few other minor changes as well. This LGTM just need someone else to approve.

contracts/governance/README.adoc
```diff
@@ -70,14 +70,20 @@ NOTE: Functions of the `Governor` contract do not include access control. If you
 
 {{GovernorTimelockCompound}}
 
+{{IGovernorTimelock}}
```
Amxx, line 73: Should interfaces be added here. I'm not sure.

Amxx (commented):
Slight comment about the docs. 

Code and Tests are good to me !

contracts/governance/README.adoc
RenanSouza2: other docs including this one have interfaces in them, if you want I can remove them

Amxx: right, in the lower part we do include the interfaces...

It should probably have gone in a separate PR, but if @frangio or @ernestognw approves that that is fine with me.

Amxx (dismissed):

contracts/governance/README.adoc
frangio: I noticed this and it's not ideal but I'm fine with doing it in this PR.

contracts/governance/TimelockController.sol
```diff
@@ -434,4 +453,19 @@ contract TimelockController is AccessControl, ERC721Holder, ERC1155Holder {
...
+     *
+     * 0x000...1000
+     *   ^^^^^^----- ...
+     *         ^---- Done
+     *          ^--- Ready
+     *           ^-- Waiting
+     *            ^- Unset
+     */
+    function _encodeStateBitmap(OperationState proposalState) internal pure returns (bytes32) {
+        return bytes32(1 << uint8(proposalState));
```
ernestognw, line 469: For consistency, we need to call the arguments with the contract-related concepts. In this case it's an operation although it's a proposal when it passes through governance.

```suggestion
    function _encodeStateBitmap(OperationState operationState) internal pure returns (bytes32) {
        return bytes32(1 << uint8(operationState));
```

contracts/governance/TimelockController.sol
```diff
@@ -434,4 +453,19 @@ contract TimelockController is AccessControl, ERC721Holder, ERC1155Holder {
         emit MinDelayChange(_minDelay, newDelay);
         _minDelay = newDelay;
     }
+
+    /**
+     * @dev Encodes a `OperationState` into a `bytes32` representation where each bit enabled corresponds to
+     * the underlying position in the `ProposalState` enum. For example:
```
ernestognw, line 459:
```suggestion
     * the underlying position in the `OperationState` enum. For example:
```

contracts/governance/TimelockController.sol
```diff
@@ -53,7 +53,7 @@ contract TimelockController is AccessControl, ERC721Holder, ERC1155Holder {
     /**
      * @dev The current state of an operation is not as required.
```
ernestognw, line 54:
```suggestion
     * @dev The current state of an operation is not as required.
     * The `expectedStates` is a bitmap with the bits enabled for each OperationState enum position
     * counting from right to left.
     *
     * See {_encodeStateBitmap}.
```

ernestognw (commented):
We're close, just want to confirm why are we removing the `virtual` from the getters, but that was changed by @frangio

ernestognw:
> As of now I just copied the '_encodeStateBitmap' from 'Governor' to 'TimelockController' I would like to unify both implementation, maybe in a file in the utils folder in governance itself, but this woul affect too many files so I want to see yout take on that before doing it

Let's make an issue for this, it needs discussion but my opinion is that the `proposalStatesToBitMap` function in `test/helpers/governance.js` should be in `test/helpers/enums.js` instead and be renamed to `encodedBitmapEnum` or similar

ernestognw (approved):
Just checked back with @frangio on the `virtual` topic and we agreed the way to override them is by overriding the `state` getter, which is the source of truth. Also, other functions use the getters instead of `state` so this is the way to guarantee they behave correctly.

contracts/governance/README.adoc
ernestognw: I think it's removed [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4358/commits/bd74728f5630baf36622c4782d08f8c8b31d1f69) so not much to add other than saying I agree the interface is not needed.

PR #4360, opened 2023-06-16
Remove GovernorCompatibilyBravo and add simpler GovernorStorage

Amxx:
As discussed.

Fixes LIB-946
Fixes LIB-944

On of the challenge is that in order to provide the `queue(uint256)` endpoint, I needed to hook into the `queue(address[], uint256[], bytes[], bytes32)` function, which is defined in IGovernorTimelock and not in IGovernor.

I did not move the propotype declaration, because that would mess up the interfaceId, and I just implemented a base version of the function (that just reverts for now) in Governor. This requires some new overrides. Moving the interface declaration would possibly reduce the need for override, but it would also either break the interfaceIds or require a lot of "xor tricks".

No tests yet, will do them if we agree on the solidity part.

### Also
- [x] refactor public/internal functions for `queue`/`_queue` and `cancel`/`_cancel`, following the `execute`/`_execute` pattern
  - public version computes the proposalId from the details AND check the current status of the proposal
  - internal version get the proposalId as a parameter
    - internal no longer recompute / return the proposalId
- [x] new `_validateStateBitmap` function to avoid code duplication when checking the current state.

### To discuss
Need for overrides would be improved if we moved the public declaration of the `queue` function from `IGovernorTimelock` to `IGovernor`. This was not (yet) implemented because of the ERC-165 consequences. However, [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4378/files#r1244535064) makes me think that the ERC165 will likely change anyway, and that we are ok with that. So maybe we should also do that change.

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/governance/README.adoc
```diff
@@ -36,7 +36,7 @@ Timelock extensions add a delay for governance decisions to be executed. The wor
 
 Other extensions can customize the behavior or interface in multiple ways.
 
-* {GovernorCompatibilityBravo}: Extends the interface to be fully `GovernorBravo`-compatible. Note that events are compatible regardless of whether this extension is included or not.
+* {GovernorStorage}: Stores the proposal detail onchain and provide enumerability of the proposals. This can be usefull for some L2 chains where storage is cheap compared to calldata.
```
Amxx, line 39:
```suggestion
* {GovernorStorage}: Stores the proposal detail onchain and provide enumerability of the proposals. This can be useful for some L2 chains where storage is cheap compared to calldata.
```

contracts/governance/Governor.sol
```diff
@@ -327,6 +327,20 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
+     *
+     * Note: This is an empty declaration for timelock modules to hook into.
+     */
+    function queue(
+        address[] memory /*targets*/,
+        uint256[] memory /*values*/,
+        bytes[] memory /*calldatas*/,
+        bytes32 /*descriptionHash*/
+    ) public virtual returns (uint256) {
+        revert();
```
ernestognw, line 341: Why not a `NonImplementedMethod` error or similar?

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+
+pragma solidity ^0.8.19;
+
+import "../Governor.sol";
+
+/**
+ * @dev Extension of {Governor} that implements storage of proposal details. This modules also provides primitives for
+ * the enumerability of proposals.
+ *
+ * Usecases for this module include:
```
ernestognw, line 11:
```suggestion
 * Use cases for this module include:
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+ * - UIs that expore the proposal state without relying on event indexing.
```
ernestognw, line 12:
```suggestion
 * - UIs that explore the proposal state without relying on event indexing.
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+ * - For some L2 chains where storage is compared to calldata, using a the proposalId version of queue and execute
```
ernestognw, line 13:
```suggestion
 * - Using only the proposalId as an argument in the {Governor-queue} and {Governor-execute} functions for L2 chains where storage is cheap compared to calldata.
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+        });
+        return proposalId;
+    }
+
+    /**
+     * @dev ProposalId version of {IGovernorTimelock-queue}.
+     */
+    function queue(uint256 proposalId) public virtual {
+        ProposalDetails storage proposalDetails = _proposalDetails[proposalId];
+        queue(
```
ernestognw, line 55: Is this equivalent to `super.queue(...)`? Asking because I genuinely don't know if it makes any difference in the inheritance order

contracts/governance/Governor.sol
```diff
@@ -327,6 +327,20 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
         return proposalId;
     }
 
+    /**
+     * @dev See {IGovernorTimelock-queue}.
+     *
+     * Note: This is an empty declaration for timelock modules to hook into.
+     */
+    function queue(
```
ernestognw, line 335: This changes the interface so we may need to update `supportsInterface`. I think we agreed not to do so but flagging for visibility in case we want to add a comment about it.

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+        cancel(
+            proposalDetails.targets,
+            proposalDetails.values,
+            proposalDetails.calldatas,
+            proposalDetails.descriptionHash
+        );
+    }
+
+    /**
+     * @dev Return the number of stored proposals.
```
ernestognw, line 90:
```suggestion
     * @dev Returns the number of stored proposals.
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+     */
+    function getProposalCount() public view virtual returns (uint256) {
+        return _proposalIds.length;
+    }
+
+    /**
+     * @dev Return the details of a proposalId.
```
ernestognw, line 97:
```suggestion
     * @dev Returns the details of a proposalId.
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+     */
+    function getProposalDetails(
+        uint256 proposalId
+    ) public view virtual returns (uint256, address[] memory, uint256[] memory, bytes[] memory) {
+        ProposalDetails storage details = _proposalDetails[proposalId];
+        return (proposalId, details.targets, details.values, details.calldatas);
+    }
+
+    /**
+     * @dev Return the details (including the proposalId) of a proposal given its sequential index.
```
ernestognw, line 107:
```suggestion
     * @dev Returns the details (including the proposalId) of a proposal given its sequential index.
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+            targets: targets,
+            values: values,
+            calldatas: calldatas,
+            descriptionHash: keccak256(bytes(description))
+        });
+        return proposalId;
+    }
+
+    /**
+     * @dev ProposalId version of {IGovernorTimelock-queue}.
```
ernestognw, line 51:
```suggestion
     * @dev Version of {IGovernorTimelock-queue} with only `proposalId` as an argument.
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+        queue(
+            proposalDetails.targets,
+            proposalDetails.values,
+            proposalDetails.calldatas,
+            proposalDetails.descriptionHash
+        );
+    }
+
+    /**
+     * @dev ProposalId version of {IGovernor-execute}.
```
ernestognw, line 64:
```suggestion
     * @dev Version of {IGovernor-execute} with only `proposalId` as an argument.
```

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+        string memory description
+    ) public virtual override returns (uint256) {
+        uint256 proposalId = super.propose(targets, values, calldatas, description);
+        // store
+        _proposalIds.push(proposalId);
+        _proposalDetails[proposalId] = ProposalDetails({
+            targets: targets,
+            values: values,
+            calldatas: calldatas,
+            descriptionHash: keccak256(bytes(description))
```
ernestognw, line 45: I'm concerned about this because we're double hashing (expensive operation) in `super.propose()` and also here.
Is it possible to split `Governor.sol`'s `propose` into an internal `_propose` that already receives a `descriptionHash`? This way we can reuse it without hashing twice

contracts/governance/README.adoc
```diff
@@ -36,7 +36,7 @@ Timelock extensions add a delay for governance decisions to be executed. The wor
 
 Other extensions can customize the behavior or interface in multiple ways.
 
-* {GovernorCompatibilityBravo}: Extends the interface to be fully `GovernorBravo`-compatible. Note that events are compatible regardless of whether this extension is included or not.
+* {GovernorStorage}: Stores the proposal detail onchain and provide enumerability of the proposals. This can be useful for some L2 chains where storage is cheap compared to calldata.
```
ernestognw, line 39:
```suggestion
* {GovernorStorage}: Stores the proposal details onchain and provides enumerability of the proposals. This can be useful for some L2 chains where storage is cheap compared to calldata.
```

contracts/governance/Governor.sol
```diff
@@ -327,6 +327,20 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
         return proposalId;
     }
 
+    /**
+     * @dev See {IGovernorTimelock-queue}.
+     *
+     * Note: This is an empty declaration for timelock modules to hook into.
```
ernestognw, line 333:
```suggestion
     * NOTE: This is an empty declaration for timelock modules to hook into.
```

contracts/governance/Governor.sol
```diff
@@ -327,6 +327,20 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
         return proposalId;
     }
 
+    /**
+     * @dev See {IGovernorTimelock-queue}.
```
ernestognw, line 331: I'm a bit confused. I think this NatSpec doesn't clear explain why this function is right here. Even if it's for Timelock modules, why are adding this function instead of the `IGovernorTimelock`? or, why can't we just declare it as a virtual function if the contract is already abstract?

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,114 @@
...
+            proposalDetails.values,
+            proposalDetails.calldatas,
+            proposalDetails.descriptionHash
+        );
+    }
+
+    /**
+     * @dev Return the number of stored proposals.
+     */
+    function getProposalCount() public view virtual returns (uint256) {
```
ernestognw, line 92: What's the rationale for naming the getters as `get`? We're naming other proposal getters as just `proposal*`, in this case, it'd be `proposalCount()`.

Non-blocking either.

ernestognw (commented):
We still need to remove the bravo module mentions from `Governor`, `IGovernor`, `GovernorCountingSimple`, and `governance.adoc`

contracts/governance/Governor.sol
Amxx: That would be ok with me.

contracts/governance/Governor.sol
Amxx: It does not. The interface ids are derived from the interface (type(IGovernor).interfaceId), which this PR does not modify.

contracts/governance/extensions/GovernorStorage.sol
Amxx: Just realized there is also {Governor-cancel}. Maybe enumerating them is too much ?

contracts/governance/extensions/GovernorStorage.sol
Amxx: I fell like splitting `propose` would add complexity ... for something that is not a big gain.

Hashing is not the expensive part here:

- Doing keccak for ~1000 bytes is <500 gas. 
- This operation writes about `4+3*l` slots, with `l` the "length" or the proposal ... and more if calldatas are multi-slot (ofter the case).
  - These slots are all expected to be empty, so that is 21k per slot
    - so that is `84k + 63k * length` 
Even if the proposal contains a single call, and that call has very little calldata, we are still way over 100k gas, so I wouldn't worry about saving 500.

Again, this function is intended to be used on chain where gas is cheap.

contracts/governance/extensions/GovernorStorage.sol
Amxx: I think its different. Lets say you inherit: `MyGovernor is Governor, GovernorStorage, GovernorTimelockControl`

- calling `queue(...)` will get the public function, which is the result of the `Governor → GovernorTimelockControl` override. So we get the function we want.
- calling `super.queue(...)` will get the version of the parent, which here is `Governor`. It would skip the one in `GovernorTimelockControl` ... and revert.

If you inherit `GovernorTimelockControl` before `GovernorStorage` both versions will work, but we can't count on that. I think this is the way to do it.

Nice question though!

contracts/governance/extensions/GovernorStorage.sol
Amxx: I'm open to renaming them.

contracts/governance/Governor.sol
```diff
@@ -424,37 +431,44 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
+     * NOTE: This reverts by default, and must be overriden (without calling super) to implement queuing.
+     */
+    function _queue(
+        uint256 /* proposalId */,
+        address[] memory /* targets */,
+        uint256[] memory /* values */,
+        bytes[] memory /* calldatas */,
+        bytes32 /*descriptionHash*/
+    ) internal virtual {
+        revert GovernorQueueNotImplemented();
```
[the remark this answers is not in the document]
Amxx: So the GovernorTimelock modules should override both the `_queue` function and the `_queuingEnabled` function ?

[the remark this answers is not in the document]
frangio: >(both will have to be "resolved" in the final contracts :/)

I think this is only if you explicitly inherit both Governor and GovernorTimelockControl. We have to experiment with that a bit...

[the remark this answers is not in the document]
Amxx: In a perfect world we would be able to detect if queuing was enabled by comparing function pointers:

```solidity
bool isQueuingEnabled = _queue == Governor._queue;
```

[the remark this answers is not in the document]
Amxx: I'm wondering, since `_queueEnabled` will not call super when overriden, do we not have a similar issue here? 

What about `_cancel` returning a bool ?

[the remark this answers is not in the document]
Amxx: I implemented the boolean return

contracts/governance/Governor.sol
Amxx: updated to `GovernorQueueNotImplemented`

Amxx:
I did an overall refactor of the public and internal functions

### Types of functions
- public
  - compute proposalId
  - check restrictions
  - call internal
- internal
  - check state is consistent
  - write flags
  - setup before operation is done
  - calls "internalCalls"
  - cleanup
  - emit event
- "internalCalls"
  - does the external call
  
### Rational
- you can override the public one if you want to add restrictions for the public invocations.
  - In that case you call super
- you can override the internal one if you want to insert extra logic (like storing more stuff in _propose)
  - In that case you call super
- you can override the internalCalls one to reconfigure the logic (execute trough a timelock).
  - In that case you do not call super
  
Also: you can call the internal version directly if you want to bypass the public checks (for example do a proposal without the threshold check). you should never call the "internalCalls" directly (can't enforce that)

WDYT @frangio @ernestognw

contracts/governance/Governor.sol
```diff
@@ -274,19 +255,86 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
+        }
 
-        // Avoid stack too deep
-        {
-            uint256 proposerVotes = getVotes(proposer, currentTimepoint - 1);
-            uint256 votesThreshold = proposalThreshold();
-            if (proposerVotes < votesThreshold) {
-                revert GovernorInsufficientProposerVotes(proposer, proposerVotes, votesThreshold);
-            }
+        return _propose(proposer, targets, values, calldatas, description);
```
frangio, line 269: Is there a reason why `propose` + `_propose` is the only combo where the internal function computes the hash rather than the public one?

contracts/governance/Governor.sol
```diff
@@ -389,39 +433,50 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
-                if (targets[i] == address(this)) {
-                    _governanceCall.pushBack(keccak256(calldatas[i]));
-                }
-            }
+        _validateStateBitmap(proposalId, _encodeStateBitmap(ProposalState.Succeeded));
+
+        uint256 eta = _queueCalls(proposalId, targets, values, calldatas, descriptionHash);
+
+        if (eta == 0) {
+            revert GovernorQueueNotImplemented();
```
frangio, line 453: This is clever, but it seems a little dangerous to assume that a timelock would never create an eta of 0. Our own TimelockController allows that, and in that case the call to schedule is required, it is not possible to go directly to execute.

We may have to return `(bool, uint)`.

contracts/governance/Governor.sol
```diff
@@ -323,59 +371,55 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
     /**
-     * @dev See {IGovernor-execute}.
+     * @dev Internal execution mechanism. Can be overridden (with a super call) to add logic before or after the
+     * execution.
+     *
+     * Emits a {IGovernor-ProposalExecuted} event.
      */
-    function execute(
+    function _execute(
+        uint256 proposalId,
```
frangio, line 380: In your explanation you say:

> you can call the internal version directly if you want to bypass the public checks

But one of the issues with this is that you can pass a `proposalId` that doesn't correspond to the proposal parameters. Is this a bug or a feature? What do you think?

contracts/governance/extensions/GovernorStorage.sol
frangio: It is different, here it's important to not use super.

contracts/governance/Governor.sol
Amxx: This was moved to `IGovernor`

contracts/governance/Governor.sol
Amxx: > But one of the issues with this is that you can pass a `proposalId` that doesn't correspond to the proposal parameters. Is this a bug or a feature? What do you think?

Tat is something I don't like, and I'd like to avoid, but I'm not sure how to do without re-hashing the proposalId. Are we ok with computing it twice ?

Amxx: Its true that currently `execute/_execute` and `queue/_queue` don't bring much in terms of features. I did it for clarity (same pattern everywhere), and to have clear guidelines as to what should be overriden depending on the usecase.

You want to add accesscontrol to queue and execute ? do it on the public one. You want to add internal tracking logic ? do it on the internal one.

I'm open to merging them if we fell that would be preferable.

frangio: I think the "weakness" about `proposalId` is acceptable. But the separation doesn't seem to work well to add access control. In the case of cancel if we want to add a new way to cancel that uses different "access control" we would be violating the guidelines I'm proposing in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4415.

In the case of cancel we could solve this by having a `_canCancel(msgSender, proposalId)` function. What do you think?

Amxx: When I said "add access control" I meant adding restriction. Propose / queue / execute are already very open (appart from the proposalThreshold part), and adding restriction is "easy".

The `_canCancel` is interresting ... but ideally it would take all the proposal details (maybe the rules depend on the contract being called).

contracts/governance/extensions/GovernorStorage.sol
ernestognw: Ok, I agree, seems like I overestimated what the double operation would be.

contracts/governance/Governor.sol
```diff
@@ -333,59 +336,133 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
+        _validateStateBitmap(proposalId, _encodeStateBitmap(ProposalState.Succeeded));
 
-        emit ProposalExecuted(proposalId);
+        (bool queued, uint48 eta) = _doQueue(proposalId, targets, values, calldatas, descriptionHash);
 
-        _beforeExecute(proposalId, targets, values, calldatas, descriptionHash);
-        _execute(proposalId, targets, values, calldatas, descriptionHash);
-        _afterExecute(proposalId, targets, values, calldatas, descriptionHash);
+        if (queued) {
+            _proposalsExtra[proposalId].eta = eta;
```
frangio, line 369: I believe we may need `require(eta > 0)`. We use this assumption in `state()`, if I'm not mistaken?

frangio: I suppose if `eta == 0` the `queue` operation is a bit of a no-op because it would remain in `Succeeded` state.

Perhaps there is no strong reason not to allow this, other than as a sanity check.

contracts/governance/extensions/GovernorStorage.sol
frangio: It seems like we should remove the `get` prefixes if we want to be consistent with the rest of the interface. I don't feel super strongly about this but I would do it.

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,117 @@
...
+        return _proposalIds.length;
+    }
+
+    /**
+     * @dev Returns the details of a proposalId.
+     */
+    function getProposalDetails(
+        uint256 proposalId
+    ) public view virtual returns (uint256, address[] memory, uint256[] memory, bytes[] memory) {
```
frangio, line 104: Why is the `proposalId` returned from this function? I would not return it here, but I would agree with keeping it in `getProposalDetailsAt`.

contracts/governance/extensions/GovernorStorage.sol
ernestognw: Also not super strong but I think a common feature of Solidity is having a getter for every public variable. I'd argue that the developers are used to not seeing the `get` prefix.

One could argue we shouldn't do Solidity specifics, but this doesn't affect external contracts and we're indeed coding it in Solidity.

frangio: Renamed without `get`.

contracts/governance/Governor.sol
```diff
@@ -333,59 +336,133 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
      */
-    function execute(
+    function queue(
         address[] memory targets,
         uint256[] memory values,
         bytes[] memory calldatas,
         bytes32 descriptionHash
-    ) public payable virtual override returns (uint256) {
+    ) public virtual override returns (uint256) {
+        return _queue(targets, values, calldatas, descriptionHash);
```
frangio, line 347: Honestly I'm very tempted to say we should merge `queue`+`_queue` and `execute`+`_execute`... I understand the "logic", but it feels wrong to have this separation if we don't use it for anything, it just adds lines of code and potential confusion.

frangio:
I made some changes above, please take a look.

The only other thing I want to review is the fact that clocks are in the `IGovernor` interface, and in particulart in the interface id. At the very least we need to remove them from the id.

contracts/governance/extensions/GovernorTimelockCompound.sol
```diff
@@ -146,8 +111,6 @@ abstract contract GovernorTimelockCompound is IGovernorTimelock, Governor {
 
         uint256 eta = proposalEta(proposalId);
         if (eta > 0) {
-            // update state first
-            delete _proposalTimelocks[proposalId];
```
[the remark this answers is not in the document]
Amxx: We could delete it in the core, that would possibly give us a refund. Note that `GovernorBravoDelegate` does not clean the eta when a proposal is executed

contracts/governance/Governor.sol
Amxx: If we require eta > 0, does it make sens to remove the `bool queued` and to have `_doQueue` only return a uint32, with 
```
        if (eta > 0) {
            _proposalsExtra[proposalId].eta = eta;
            emit ProposalQueued(proposalId, eta);
        } else {
            revert GovernorQueueNotImplemented();
        }
 ...

File: contracts/governance/Governor.sol (hunk 24 above)
Comment (Amxx) [reply contracts/governance/Governor.sol:347]: It also adds some consistency.

My concern is, if we merge them, and we later realize that for some reason we should split them, is that something we can do in a non breaking (minot) version ?

File: contracts/governance/Governor.sol (hunk 22 above)
Comment (frangio) [reply contracts/governance/Governor.sol:369]: I think that makes sense. Initially I wanted to allow eta = 0 because because I thought it was a delay, I think. But if it's a timestamp, it should never be 0, we can use it as a natural sentinel value to mean not queued.

File: contracts/governance/extensions/GovernorTimelockCompound.sol (hunk 25 above)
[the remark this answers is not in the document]
Comment (frangio) [reply contracts/governance/extensions/GovernorTimelockCompound.sol:150]: Yeah I think it makes sense to retain the value for proposalEta(proposalId). Let's keep it.

File: docs/modules/ROOT/pages/governance.adoc [hunk 26]
```diff
@@ -18,11 +18,11 @@ OpenZeppelin’s Governor system was designed with a concern for compatibility w
...
 
-Note that even with the use of this module, there will still be differences in the way that `proposalId`s are calculated. Governor uses the hash of the proposal parameters with the purpose of keeping its data off-chain by event indexing, while the original Bravo implementation uses sequential `proposalId`s. Due to this and other differences, several of the functions from GovernorBravo are not included in the compatibility module.
+Note that even with the use of this module, one important difference in the way that `proposalId`s are calculated. Governor uses the hash of the proposal parameters with the purpose of keeping its data off-chain by event indexing, while the original Bravo implementation uses sequential `proposalId`s.
```
Amxx, line 25:
```suggestion
Note that even with the use of this module, one important difference with Compound's GovernorBravo is the way that `proposalId`s are calculated. Governor uses the hash of the proposal parameters with the purpose of keeping its data off-chain by event indexing, while the original Bravo implementation uses sequential `proposalId`s.
```

test/helpers/governance.js
```diff
@@ -1,3 +1,4 @@
+const { web3 } = require('hardhat');
```
Amxx, line 1:
```suggestion
```

Amxx, line 1: Do we need that to be explicit ?

Amxx: There are many place where we use implicit import for things like `web3`, and a few places where we have explicit import. IMO we need some guideline (or we need to move the ethers.js)

frangio:
We need to add a function like `proposalMayNeedQueueing(uint256 proposalId) returns (bool)`. Name TBD. Will do in a follow up PR.

test/helpers/governance.js
frangio: > or we need to move the ethers.js

This doesn't change anything, as long as we use Hardhat we have to decide whether to use globals or not.

Personally I always prefer explicit imports.

contracts/governance/Governor.sol
frangio: The only concern, if we do that, is that overrides to the public function would not be present for the internal function, and depending on the logic they might have to.

I think it's a fairly small risk of that causing any issues for these functions. I think I'm comfortable documenting via changelog if we ever do that.

contracts/governance/extensions/GovernorTimelockControl.sol
```diff
@@ -73,52 +63,34 @@ abstract contract GovernorTimelockControl is IGovernorTimelock, Governor {
...
 
-        emit ProposalQueued(proposalId, block.timestamp + delay);
-
-        return proposalId;
+        return SafeCast.toUint48(block.timestamp + delay);
     }
 
     /**
-     * @dev Overridden execute function that run the already queued proposal through the timelock.
+     * @dev Overridden version of the {Governor-_doExecute} function that run the already queued proposal through
```
ernestognw, line 90:
```suggestion
     * @dev Overridden version of the {Governor-_doExecute} function that runs the already queued proposal through
```

ernestognw (commented):
We're close!

contracts/governance/Governor.sol
ernestognw: It's not really clear in my mind why would we need to have the 3 of the functions. I agree with the rationale you shared and I think it makes sense but having a `_doExecute` is something we can add in a minor version if I'm not mistaken. Please correct me if wrong.

My preference would be having only `execute` and `_execute` but not super strong. I'll be open to it if it's clearer for the user. An example of clarity is the `_update` and `_updateWithAcceptanceCheck` in ERC721. It shows clearly what exactly each one of those do.

frangio: It's the other way around. We need "`_doExecute`" now, it's used by the timelock extensions. What we don't know is about the separation between `execute` and `_execute`.

ernestognw: I know it's the other way around, but the `_doExecute` can be `_execute` and the original `_execute` can be replaced with `_executeWithValidation`. Just putting an example, please don't discard my idea just because you interpreted differently.

frangio: I'm not sure I understood the idea yet. Initially you said you prefer only `execute` and `_execute`, I assume you meant the current `execute` and `_doExecute` functions (the latter renamed to `_execute`). Would `_executeWithValidation` be a 3rd function with the semantics of the current `_execute`?

ernestognw: > My preference would be having only execute and _execute but not super strong.

Because my opinion is not super strong, I'm open to having 3 if we change the names. I put the ERC721's case as an example.

frangio: Ok, so I think we are aligned. We both prefer having only two functions, and if there are three functions they need to have better names.

`execute` and `_execute` (in the current PR) do exactly the same, so they shouldn't be named too differently, and similarly `_doExecute` does something different, so it should not be named `_execute`. If we want a better name we should find one for `_doExecute`, regardless of whether we keep `_execute` or not.

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * @dev Extension of {Governor} that implements storage of proposal details. This modules also provides primitives for
+ * the enumerability of proposals.
+ *
+ * Use cases for this module include:
+ * - UIs that explore the proposal state without relying on event indexing.
+ * - Using only the proposalId as an argument in the {Governor-queue} and {Governor-execute} functions for L2 chains where storage is cheap compared to calldata.
+ *   might me cheaper.
+ *
+ * _Available since v5.0._
```
frangio, line 16:
```suggestion
```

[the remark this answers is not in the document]
frangio: Removed.

test/governance/extensions/GovernorStorage.test.js
```diff
@@ -0,0 +1,150 @@
...
+
+        await web3.eth.sendTransaction({ from: owner, to: this.timelock.address, value });
+
+        // normal setup: governor is proposer, everyone is executor, timelock is its own admin
+        await this.timelock.grantRole(PROPOSER_ROLE, this.mock.address);
+        await this.timelock.grantRole(PROPOSER_ROLE, owner);
+        await this.timelock.grantRole(CANCELLER_ROLE, this.mock.address);
+        await this.timelock.grantRole(CANCELLER_ROLE, owner);
+        await this.timelock.grantRole(EXECUTOR_ROLE, constants.ZERO_ADDRESS);
+        await this.timelock.revokeRole(DEFAULT_ADMIN_ROLE, deployer);
```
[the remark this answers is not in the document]
frangio: I tried `Promise.all` and regardless of `await` Hardhat will mine a block for every transaction. It doesn't seem to make a difference.

test/governance/Governor.test.js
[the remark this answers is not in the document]
frangio: Fixed in bdddeda1.

contracts/governance/Governor.sol
```diff
@@ -284,32 +278,45 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         string memory description
     ) public virtual override returns (uint256) {
         address proposer = _msgSender();
-        require(_isValidDescriptionForProposer(proposer, description), "Governor: proposer restricted");
 
-        uint256 currentTimepoint = clock();
+        // check description restriction
+        require(_isValidDescriptionForProposer(proposer, description), "Governor: proposer restricted");
```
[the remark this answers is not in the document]
frangio: Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4360/commits/bdddeda1cb5605418de65665ffb29aec4f8732f1.

test/governance/extensions/GovernorStorage.test.js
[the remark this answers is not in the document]
ernestognw: I made a quick test and it was saving a considerable amount of time if instead of `Promise.all` we disabled Hardhat's automining. I created an util for it.

Take a look to https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4360/commits/b2445f8bd86e49ea8f87769f9337f5ff641ecb16

I only applied it to GovernorStorage because I don't want to add even more changes to this PR, but these were the setup time reductions:

| Test Case                    | With Batching (ms) | Without Batching (ms) |
|------------------------------|--------------------|-----------------------|
| queue and execute by id      | 5.987              | 108.955               |
| cancel by id, proposal indexing | 5.412            | 93.074                |
| before propose               | 5.306              | 92.509                |
| after propose                | 1.833              | 80.512                |
| queue and execute by id      | 1.429              | 74.976                |
| cancel by id, proposal indexing | 1.434            | 81.671                |
| before propose               | 1.343              | 90.931                |
| after propose                | 1.297              | 83.321                |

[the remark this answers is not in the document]
frangio: Seem very good, but we 100% need the `await` in each statement. The benefit probably comes from automine being off.

contracts/governance/Governor.sol
```diff
@@ -203,21 +190,28 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
     /**
      * @dev Returns the account that created a given proposal.
      */
     function proposalProposer(uint256 proposalId) public view virtual override returns (address) {
-        return _proposals[proposalId].proposer;
+        return _proposals[proposalId].core.proposer;
+    }
+
+    /**
+     * @dev Public accessor to check the eta of a queued proposal
```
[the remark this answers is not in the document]
frangio: Defined it in `IGovernor`: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4360/commits/92589ff6201f1660339681a0ed4388cc077527f2

test/governance/extensions/GovernorStorage.test.js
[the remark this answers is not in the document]
ernestognw: It fails because each `await` is waiting for a block to be mined apparently. Shall we revert?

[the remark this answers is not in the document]
frangio: Ah, I forgot Truffle did that. Let me see.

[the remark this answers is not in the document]
frangio: We should just use the existing `batchInBlock`.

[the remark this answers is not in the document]
frangio: It's not working, I don't understand. Let's revert.

contracts/governance/Governor.sol
frangio: Okay, since `_do{Queue,Execute}` was not great naming, I've renamed to `_{queue,execute}Operations`. I realize this is closer to the original name of `_{queue,execute}Calls`, but I think "operations" is better ("calls" is confusing to me because it's also a verb).

---

I would still remove `_queue` and `_execute` but I'll let @Amxx do it if he agrees. It's as simple as removing a few lines so that the body is moved to the public functions, since they're not even tested directly.

test/governance/extensions/GovernorStorage.test.js
```diff
@@ -53,19 +54,20 @@ contract('GovernorStorage', function (accounts) {
...
-        await this.timelock.grantRole(CANCELLER_ROLE, owner);
-        await this.timelock.grantRole(EXECUTOR_ROLE, constants.ZERO_ADDRESS);
-        await this.timelock.revokeRole(DEFAULT_ADMIN_ROLE, deployer);
-
-        await this.token.$_mint(owner, tokenSupply);
-        await this.helper.delegate({ token: this.token, to: voter1, value: web3.utils.toWei('10') }, { from: owner });
-        await this.helper.delegate({ token: this.token, to: voter2, value: web3.utils.toWei('7') }, { from: owner });
-        await this.helper.delegate({ token: this.token, to: voter3, value: web3.utils.toWei('5') }, { from: owner });
-        await this.helper.delegate({ token: this.token, to: voter4, value: web3.utils.toWei('2') }, { from: owner });
+        await mineWith(async () => {
```
Amxx: its nice, but is it needed ?

test/governance/extensions/GovernorStorage.test.js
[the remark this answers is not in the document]
Amxx: We resolved this problem a while ago, when working on the voting contract. We have `batchInBlock` for that:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/helpers/txpool.js#L10-L33

Is that not nice enough for what you want to do ?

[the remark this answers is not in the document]
Amxx: Note that I don't get why you want all that to happen in a single block. We have similar setups in other test files, do you also want to change them ?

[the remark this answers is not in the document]
frangio: The motivation was for performance reasons, allegedly tests run faster.

We should do this in another PR anyway.

frangio:
@Amxx @ernestognw This is ready to merge in my opinion. I've done final changes I wanted to make in the last few commits.

- Removed `_queue`, `_execute`: were not strictly necessary and practical use cases are not clear.
- Made `IGovernor` into an interface.

Regarding `Governor.proposalMayNeedQueuing`, we should add this along with the Governor module for AccessManager.

test/governance/extensions/GovernorStorage.test.js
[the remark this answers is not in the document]
ernestognw: Yes, let's do it in another PR, I don't want to shift the attention to this.

contracts/governance/extensions/GovernorTimelockControl.sol
```diff
@@ -39,27 +39,17 @@ abstract contract GovernorTimelockControl is IGovernorTimelock, Governor {
     }
 
     /**
-     * @dev See {IERC165-supportsInterface}.
+     * @dev Overridden version of the {Governor-state} function status check on the timelock if in ss`Queued`.
```
ernestognw, line 42: Not sure what was the intention here but it needs attention.

ernestognw (dismissed):
I've reviewed the new commits after my other review. LGTM with a small comment.

contracts/governance/extensions/GovernorTimelockControl.sol
frangio: Fixed.

Amxx:
@frangio Considering this PR https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4495, I believe caching to memory is not something efficient that we should encourage. I propose:
- don't split the ProposalFull into 2 anymore, and go back to a `Proposal` struct that contains everything (it will be more then one slot though)
- apply the recommandation of #4495.

WDYT ?

contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -0,0 +1,109 @@
...
+        return _proposalIds.length;
+    }
+
+    /**
+     * @dev Returns the details of a proposalId. Reverts if `proposalId` is not a known proposal.
+     */
+    function proposalDetails(
+        uint256 proposalId
+    ) public view virtual returns (address[] memory, uint256[] memory, bytes[] memory, bytes32) {
+        ProposalDetails memory details = _proposalDetails[proposalId];
```
Amxx, line 87: In queue, cancel and execute we use a storage ptr. Here we use memory. Does it make sens to not be consistant ? Need to test which version is better.

PR #4365, opened 2023-06-16
Change release cycle `prBackExists` definition

ernestognw:
Fix an incorrect definition in the release cycle state file

### DIFF
--- a/scripts/release/workflow/state.js
+++ b/scripts/release/workflow/state.js
@@ -47,7 +47,7 @@ function shouldRunMerge({
   hasPendingChangesets,
   prBackExists,
 }) {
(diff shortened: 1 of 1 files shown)


Amxx (commented):
It looks like you are doing the negation twice, and the outcome of `shouldRunMerge` is not modified. I there more to it that I'm missing ?

ernestognw:
> It looks like you are doing the negation twice, and the outcome of `shouldRunMerge` is not modified. I there more to it that I'm missing ?

That's correct. The variable is behaving as expected is just that its name it's representing the opposite to what it should. We either do the double negation or change the variable name but I'd like to keep the naming since it makes it much easier to read each of the state functions imo

ernestognw:
> I thought there was a behavior we wanted to change.

Good call anyway, I'm renaming the PR for more clarity

frangio (approved):

PR #4369, opened 2023-06-18
Update comment to reflect code logic in Ownable.sol

s-tikhomirov:
The initial owner is not automatically assigned to the deployer's address. Rather, the initial owner is set to the constructor argument provided by the deployer.

(The code was modified in commit 13d5e04 but the comment wasn't.)

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -33,7 +33,7 @@ abstract contract Ownable is Context {
     event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
 
     /**
-     * @dev Initializes the contract setting the deployer as the initial owner.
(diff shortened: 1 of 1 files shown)


ernestognw (commented):
Hey @s-tikhomirov, Github doesn't allow suggestions 3 lines further than the change, but I think we should also update the NatSpec on top of the contract itself where it reads:

> By default, the owner account will be the one that deploys the contract.

We've already received comments about this inaccuracy, so just flagging it before merging.

Amxx (approved):

ernestognw (approved):
Thanks @s-tikhomirov!

PR #4371, opened 2023-06-19
Use clones for testing non ERC1967 proxies

Amxx:
IMO this makes the test clearer / easier to understand

#### PR Checklist

- [x] Tests


ernestognw (approved):
Thanks!

PR #4372, opened 2023-06-19
Revert upgradeTo in favor of upgradeToAndCall

Amxx:
Depends on #4354 (conflicts will have to be fixed)

Remove the upgradeToAndCall to reduce the size (and deployment cost) of: 
- UUPSUpgradeable
- TransparentUpgradeableProxy
- ProxyAdmin

This includes a refactor of the upgradeToAndCall workflow: if the calldata for the call is empty, is call is no longer performed. 

#### PR Checklist

- [x] Tests
- [ ] Documentation


frangio:
By removing `upgradeTo` and changing the behavior of `upgradeToAndCall` this would seem to cause issues with OpenZeppelin Upgrades Plugins and other similar tooling. If we want to go ahead with this we would need a solution to detect which function to call, but it seems complicated.

frangio:
Closing in favor of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4382.

PR #4374, opened 2023-06-19
Fix attempt to delete nonexistent npm tag

frangio:
The release workflows for 4.9.1 and 4.9.2 [failed](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/5294042249/jobs/9582925626) because there's an attempt to delete the `next` tag, which is not present when we're releasing a patch. This PR will first check if the next tag exists, but also if it should be deleted. If we have `next: 5.0.0-rc.0` and we publish `latest: 4.9.4` we don't want to delete the `next` tag in that case.

### DIFF
--- a/scripts/release/workflow/publish.sh
+++ b/scripts/release/workflow/publish.sh
@@ -2,19 +2,25 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


ernestognw (approved):
LGTM, let's hope it doesn't break

Amxx (approved):

PR #4377, opened 2023-06-21
Use the _update mechanism in ERC721

Amxx:
Fixes LIB-628

Issues to fix:
- [x] when a batch is minted, no hook is called, so other modules can't react to that
  - [x] ERC721Votes must update the votes accordingly
  - [x] ERC721Enumerable must revert

I fixed these using `__unsafe_increaseBalance` as a hook. It actually works well for that usecase.

Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4136.
Closes https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4405 (superceded).
Closes https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4169 (superceded).
Closes https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4144 (superceded).

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/token/ERC721/ERC721.sol
```diff
@@ -412,6 +386,34 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     * @dev Contraint: revert if token is already minted
+     */
+    function _constraintNotMinted(address from, address, uint256) internal pure {
+        if (from != address(0)) {
+            revert ERC721InvalidSender(address(0));
+        }
+    }
+
+    /**
+     * @dev Contraint: revert if token is not yet minted
```
Amxx, line 399:
```suggestion
     * @dev Constraint: revert if token is not yet minted
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -412,6 +386,34 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     */
+    function _constraintMinted(address from, address, uint256 tokenId) internal pure {
+        if (from == address(0)) {
+            revert ERC721NonexistentToken(tokenId);
+        }
+    }
+
+    /**
+     * @dev Contraint: check that the caller is the current owner, or approved by the current owner
```
Amxx, line 408:
```suggestion
     * @dev Constraint: check that the caller is the current owner, or approved by the current owner
```

frangio:
I realized that the constraints are reading `msg.sender` and I don't think this is good. It goes against the design principle that `msg.sender` should be used in the external interface, but internally it should be passed as an explicit argument. Users should not and would not expect `_update` to act differently depending on `msg.sender`.

I think `_update` needs a `from` argument, but it doesn't have to be an address, and it doesn't have to be the owner necessarily. I am thinking that there should be a user defined type representing an ADT with one variant for each of the constraints we have, and for the "approved or owner" constraint it should embed the value of `msg.sender`. (I believe this is essentially a defunctionalization of the constraints.)

Amxx:
> I realized that the constraints are reading `msg.sender` and I don't think this is good. It goes against the design principle that `msg.sender` should be used in the external interface, but internally it should be passed as an explicit argument. Users should not and would not expect `_update` to act differently depending on `msg.sender`.
> 
> I think `_update` needs a "`from`" argument, but it doesn't have to be an address, and it doesn't have to be the owner necessarily. I am thinking that this argument should have a user defined type encoding an ADT with one variant for each of the constraints we have, and for the "approved or owner" constraint it should contain the value of `msg.sender`. (I believe this is essentially a defunctionalization of the constraints.)

I would argue that its the constrain that reads `msg.sender`, not `_update` itself. So yes, `msg.sender` can affect the behavior of `_update` ... but only if the dev asks for it (and it checks according to what the dev requess).

Now I currently see 4 constraints:
- None 
- RequireMinted
- RequireNotMinted
- RequireOwnerOrApproved.

We could encode these 4 into an enum, and inside `_update` do the corresponding logic (calling a private helper?)
It would be less powerfull in terms of which behaviors a dev could implement using constaints ... but maybe you'd be more confortable with that ?

I'm not sure I fully understand what you propose, but it appears to include preparing some encoding ... I wouldn't want the result to be too complex.

Amxx:
> It will be invisible in `_transfer` and the same argument works there, it's behavior should not depend on the value of `msg.sender`.

`_transfer` does `_update(to, tokenId, _constraintMinted)` ... and `_constraintMinted` does not do any msg.sender check.

The only constrain that does a `msg.sender` check is `_constraintApprovedOrOwner`, which is only called in the public `transferFrom`. This refactor does not add any msg.sender check to any function that did not already have one, so I'm not really seeing the issue.

contracts/token/ERC721/extensions/ERC721Royalty.sol
```diff
@@ -29,10 +29,19 @@ abstract contract ERC721Royalty is ERC2981, ERC721 {
     }
 
     /**
-     * @dev See {ERC721-_burn}. This override additionally clears the royalty information for the token.
+     * @dev See {ERC721-_update}. This override additionally clears the royalty information for the token.
```
ernestognw, line 32:
```suggestion
     * @dev See {ERC721-_update}. When burning, this override will additionally clear the royalty information for the token.
```

contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -64,15 +64,21 @@ abstract contract ERC721URIStorage is IERC4906, ERC721 {
     }
 
     /**
-     * @dev See {ERC721-_burn}. This override additionally checks to see if a
+     * @dev See {ERC721-_update}. This override additionally checks to see if a
```
ernestognw, line 67:
```suggestion
     * @dev See {ERC721-_update}. When burning, this override will additionally check if a
```

ernestognw:
We still need to update the Changeset at `.changeset/bright-tomatoes-sing.md` with this new `_update` mechanism.

frangio:
@ernestognw There is an alternative to this PR in #4419.

I would like to avoid the use of function pointers introduced in this PR.

frangio:
Closing in favor of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4419.

contracts/token/ERC721/ERC721.sol
```diff
@@ -412,6 +386,34 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+    }
+
+    /**
+     * @dev Constraint: check that the caller is the current owner, or approved by the current owner
+     */
+    function _constraintApprovedOrOwner(address owner, address, uint256 tokenId) internal view {
+        address spender = _msgSender();
+        if (spender != owner && !isApprovedForAll(owner, spender) && getApproved(tokenId) != spender) {
+            revert ERC721InsufficientApproval(_msgSender(), tokenId);
+        }
```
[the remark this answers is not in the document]
Amxx: _isApprovedOrOwner reads the owner. We don't want to duplicate that sload.

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,41 +344,14 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
         }
-
-        _beforeTokenTransfer(from, to, tokenId, 1);
-
-        // Check that tokenId was not transferred by `_beforeTokenTransfer` hook
-        owner = ownerOf(tokenId);
+        address owner = _update(to, tokenId, _constraintMinted);
         if (owner != from) {
             revert ERC721IncorrectOwner(from, tokenId, owner);
         }
```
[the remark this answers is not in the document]
Amxx: that is literally the only check that verifies if the `from` provided matches the previous owner the tokenId.

Without this check you could do `_transfer(anything, to, tokenId)` and that would be accepted.

contracts/token/ERC721/ERC721.sol
```diff
@@ -151,12 +151,14 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-    function transferFrom(address from, address to, uint256 tokenId) public virtual {
-        if (!_isApprovedOrOwner(_msgSender(), tokenId)) {
-            revert ERC721InsufficientApproval(_msgSender(), tokenId);
+    function transferFrom(address from, address to, uint256 tokenId) public virtual override {
+        if (to == address(0)) {
+            revert ERC721InvalidReceiver(address(0));
+        }
+        address owner = _update(to, tokenId, _constraintApprovedOrOwner);
+        if (owner != from) {
+            revert ERC721IncorrectOwner(from, tokenId, owner);
```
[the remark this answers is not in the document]
Amxx: same answer

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -109,7 +103,7 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
      * @param tokenId uint256 ID of the token to be added to the tokens list of the given address
      */
     function _addTokenToOwnerEnumeration(address to, uint256 tokenId) private {
-        uint256 length = balanceOf(to);
+        uint256 length = balanceOf(to) - 1;
```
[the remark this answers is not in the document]
Amxx: tokenId is now owned by to (this is running after the `super._update`) so the balance is at least one.

contracts/mocks/token/ERC721ConsecutiveEnumerableMock.sol
```diff
@@ -28,25 +28,19 @@ contract ERC721ConsecutiveEnumerableMock is ERC721Consecutive, ERC721Enumerable
...
     }
 
-    function _afterTokenTransfer(
-        address from,
-        address to,
-        uint256 firstTokenId,
-        uint256 batchSize
-    ) internal virtual override(ERC721, ERC721Consecutive) {
-        super._afterTokenTransfer(from, to, firstTokenId, batchSize);
+    // solhint-disable-next-line func-name-mixedcase
```
frangio, line 39: This can be removed, it was for `__unsafe_...`.

contracts/token/ERC721/ERC721.sol
```diff
@@ -446,47 +432,14 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
     /**
      * @dev Unsafe write access to the balances, used by extensions that "mint" tokens using an {ownerOf} override.
      *
      * WARNING: Anyone calling this MUST ensure that the balances remain consistent with the ownership. The invariant
      * being that for any address `a` the value returned by `balanceOf(a)` must be equal to the number of tokens such
      * that `ownerOf(tokenId)` is `a`.
      */
-    // solhint-disable-next-line func-name-mixedcase
-    function __unsafe_increaseBalance(address account, uint256 value) internal {
+    function _increaseBalance(address account, uint256 value) internal virtual {
```
frangio, line 442: Thoughts on naming this `_updateBalance` to indicate the analogy with `_update`? In some cases like ERC20Votes contracts have to override both functions so it makes sense to me to tie them together conceptually.

@ernestognw I see you had renamed `_updateBalance` to `_increaseBalance` in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4419, what makes you prefer that one?

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,41 +329,41 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     * - `tokenId` token must exist and be owned by `from`.
+     * - If `to` refers to a smart contract, it must implement {IERC721Receiver-onERC721Received}, which is called upon a safe transfer.
+     *
+     * Emits a {Transfer} event.
+     */
+    function _safeTransfer(address from, address to, uint256 tokenId, bytes memory data) internal virtual {
+        _transfer(from, to, tokenId);
+        if (!_checkOnERC721Received(from, to, tokenId, data)) {
+            revert ERC721InvalidReceiver(to);
         }
```
frangio, line 366: In https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4419 I had made it so that the check function reverts, and it looks really clean, we should do that here.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cdacda12173d540137d263ff3779985b0519ae9a/contracts/token/ERC721/ERC721.sol#L193-L194

contracts/token/ERC721/ERC721.sol
```diff
@@ -233,28 +212,45 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-     * @dev Same as {xref-ERC721-_safeMint-address-uint256-}[`_safeMint`], with an additional `data` parameter which is
-     * forwarded in {IERC721Receiver-onERC721Received} to contract recipients.
-     */
-    function _safeMint(address to, uint256 tokenId, bytes memory data) internal virtual {
-        _mint(to, tokenId);
-        if (!_checkOnERC721Received(address(0), to, tokenId, data)) {
-            revert ERC721InvalidReceiver(to);
+        // Perform (optional) operator check
+        if (operatorCheck != address(0) && from != operatorCheck && !isApprovedForAll(from, operatorCheck) && getApproved(tokenId) != operatorCheck) {
+            revert ERC721InsufficientApproval(operatorCheck, tokenId);
```
frangio, line 232: There is still `_isApprovedOrOwner` in the contract, which is overrideable, but it's not used here, so we need to do something about that function.

In https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4419 I had refactored that function in this way:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cdacda12173d540137d263ff3779985b0519ae9a/contracts/token/ERC721/ERC721.sol#L229-L246

I believe we should do the same here, without the `actualOwner` part.

contracts/token/ERC721/ERC721.sol
frangio, line 1: I think it might make sense to merge [#4405 Exclude address(0) in ERC721 approvals](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4405) into this PR. The general theme could be removing redundant SLOADs and redundant checks.

contracts/token/ERC721/ERC721.sol
frangio, line 442: We should use a smaller integer type for the `value` argument. If we do that, there is nothing unsafe about using this function. The warning should be placed in `_ownerOf`, which is dangerous when overridden. Here is how I documented these functions in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4419:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cdacda12173d540137d263ff3779985b0519ae9a/contracts/token/ERC721/ERC721.sol#L372-L378

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cdacda12173d540137d263ff3779985b0519ae9a/contracts/token/ERC721/ERC721.sol#L197-L206

contracts/token/ERC721/ERC721.sol
```diff
@@ -257,6 +259,46 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+        if (from != address(0)) {
+            delete _tokenApprovals[tokenId];
+            unchecked {
+                _balances[from] -= 1;
+            }
+        }
+
+        if (to != address(0)) {
+            unchecked {
+                _balances[to] += 1;
```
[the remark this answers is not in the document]
frangio: Resolving to focus the conversation on the other similar comment.

contracts/token/ERC721/ERC721.sol
```diff
@@ -257,6 +259,46 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+        function(address, address, uint256) view constraints
+    ) internal virtual returns (address) {
+        address from = _ownerOf(tokenId);
+
+        constraints(from, to, tokenId);
+
+        if (from != address(0)) {
+            delete _tokenApprovals[tokenId];
+            unchecked {
+                _balances[from] -= 1;
```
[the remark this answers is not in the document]
frangio: I personally prefer `-= 1` and `+= 1`. @Amxx what is your preference here?

contracts/token/ERC721/ERC721.sol
```diff
@@ -257,6 +259,46 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+        uint256 tokenId,
+        function(address, address, uint256) view constraints
+    ) internal virtual returns (address) {
+        address from = _ownerOf(tokenId);
+
+        constraints(from, to, tokenId);
+
+        if (from != address(0)) {
+            delete _tokenApprovals[tokenId];
+            unchecked {
```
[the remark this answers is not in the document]
frangio: Constraints are no longer used, but the concern about `_ownerOf` is correct. See this comment about that: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#discussion_r1260521379

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
frangio: >We don't want to duplicate that sload.

Agreed, but the function is overrideable so we need to do something about it. See my suggestion here: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#discussion_r1260515576

(Resolving to conversation there)

contracts/token/ERC721/extensions/ERC721Pausable.sol
```diff
@@ -21,20 +21,18 @@ import {Pausable} from "../../../security/Pausable.sol";
...
-        uint256 firstTokenId,
-        uint256 batchSize
-    ) internal virtual override {
-        super._beforeTokenTransfer(from, to, firstTokenId, batchSize);
-
+        uint256 tokenId,
+        function(address, address, uint256) view constraints
+    ) internal virtual override returns (address) {
         _requireNotPaused();
+        return super._update(to, tokenId, constraints);
```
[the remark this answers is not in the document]
frangio: My preference is for `_requireNotPaused` because it feels more explicit.

contracts/token/ERC721/extensions/ERC721Enumerable.sol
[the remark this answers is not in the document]
frangio: Can we add a comment explaining that in the code?

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -135,7 +129,7 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
         // To prevent a gap in from's tokens array, we store the last token in the index of the token to delete, and
         // then delete the last slot (swap and pop).
 
-        uint256 lastTokenIndex = balanceOf(from) - 1;
+        uint256 lastTokenIndex = balanceOf(from);
```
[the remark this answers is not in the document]
frangio: Is this different because the enumerability code used to execute in the before hook but now executes _after_ updates?

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
Amxx: `+= 1` / `-= 1` is what we currently have, and I kept it. If I was to write the contract from scratch today, I would use `--` and `++`, but I don't have a strong preference.

contracts/mocks/token/ERC721ConsecutiveEnumerableMock.sol
Amxx, line 39:
```suggestion
```

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
Amxx: I'd keep it open so we have the suggestion, and we don't forget about it. But yes, the discussion should happen in the other conversation.

contracts/token/ERC721/ERC721.sol
Amxx: I think in `_updateBalance`, one would expect the value to replace the new one (overwrite), whereas `_increaseBalance` makes it clear that value is an delta (increment). I'd stick with `_increaseBalance` for clarity

contracts/token/ERC721/extensions/ERC721Enumerable.sol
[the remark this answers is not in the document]
Amxx: > Is this different because the enumerability code used to execute in the before hook but now executes after updates?

exactly!

contracts/token/ERC721/ERC721.sol
Amxx: merged

Amxx: merging breaks the tests. Can we fix the tests in #4405 before merging it here ?

contracts/token/ERC721/ERC721.sol
Amxx: done

contracts/token/ERC721/extensions/ERC721Pausable.sol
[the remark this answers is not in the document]
Amxx: Either is good for me

contracts/token/ERC721/ERC721.sol
Amxx: I did that, let me know if it works for you

contracts/token/ERC721/ERC721.sol
Amxx: The issue was with one thing: code in 4405 doesn't allow setting approval to 0, which is IMO the way to "clean" approval. I changed that part of the code, merged it, and tests are looking good.

contracts/token/ERC721/extensions/ERC721Pausable.sol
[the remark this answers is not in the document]
frangio: ERC20 and ERC1155 are using the modifier so for consistency we should use it here too.

contracts/token/ERC721/ERC721.sol
```diff
@@ -170,39 +172,15 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * @dev See {IERC721-safeTransferFrom}.
      */
     function safeTransferFrom(address from, address to, uint256 tokenId, bytes memory data) public virtual {
```
frangio, line 174: Keeping this one virtual but `safeTransferFrom(address,address,uint256)`.

.changeset/eighty-lemons-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`ERC721`: `_approve` no longer allows approving the owner of the tokenId. `_setApprovalForAll` no londer allows setting address(0) as an operator.
```
ernestognw, line 5:
```suggestion
`ERC721`: `_approve` no longer allows approving the owner of the tokenId. `_setApprovalForAll` no longer allows setting address(0) as an operator.
```

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: We agreed to keep it as it is because it's not a highly meaningful change. We might reconsider once this PR is done so we can correctly measure the gas savings.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: We agreed to keep it as it is because it's not a highly meaningful change. We might reconsider once this PR is done so we can correctly measure the gas savings.

Keeping this convo open and closing the other one.

contracts/token/ERC721/extensions/ERC721Votes.sol
```diff
@@ -16,18 +16,16 @@ import {Votes} from "../../../governance/utils/Votes.sol";
...
-        uint256 batchSize
-    ) internal virtual override {
-        _transferVotingUnits(from, to, batchSize);
-        super._afterTokenTransfer(from, to, firstTokenId, batchSize);
+    function _update(address to, uint256 tokenId, address operatorCheck) internal virtual override returns (address) {
+        address from = super._update(to, tokenId, operatorCheck);
+
+        _transferVotingUnits(from, to, 1);
+
+        return from;
```
ernestognw, line 28: Because both `_update` or `_transferVotingUnits` can't be reentered, I think we would prefer:

```suggestion
        _transferVotingUnits(from, to, 1);
        return super._update(to, tokenId, operatorCheck);
```

The rationale is that we've previously expressed that we don't like passing the `from` around from the beginning to the end because it may lead to unintentional overwrites.

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,41 +342,48 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-        _owners[tokenId] = to;
-
-        emit Transfer(from, to, tokenId);
+    /**
+     * @dev Safely transfers `tokenId` token from `from` to `to`, checking first that contract recipients
+     * are aware of the ERC721 protocol to prevent tokens from being forever locked.
+     *
+     * `data` is additional data, it has no specified format and it is sent in call to `to`.
+     *
+     * This internal function is equivalent to {safeTransferFrom}, and can be used to e.g.
```
ernestognw, line 363: This is no longer the case because `safeTransferFrom` sends `_msgSender()` as `operatorCheck`, where `_transfer` doesn't. I'd suggest adding a note for this small difference or rewording this section.

contracts/token/ERC721/ERC721.sol
```diff
@@ -386,8 +392,12 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * Emits an {Approval} event.
      */
     function _approve(address to, uint256 tokenId) internal virtual {
+        address owner = ownerOf(tokenId);
```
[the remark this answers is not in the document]
frangio: IMO reverting if the token doesn't exist is the right thing here. Setting an approval prior to minting seems like a bug more than a feature to me.

contracts/token/ERC721/ERC721.sol
```diff
@@ -221,40 +199,73 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-     * Requirements:
-     *
-     * - `tokenId` must not exist.
-     * - If `to` refers to a smart contract, it must implement {IERC721Receiver-onERC721Received}, which is called upon a safe transfer.
+     * WARNING:
+     * - ownership is not checked.
+     * - spender = address(0) will lead to this function returning true in unexpected situations.
+     */
+    function _isApproved(address owner, address spender, uint256 tokenId) internal view virtual returns (bool) {
+        return owner == spender || isApprovedForAll(owner, spender) || getApproved(tokenId) == spender;
```
frangio, line 217: Note that `getApproved` is reading the token owner because it checks existence, so we should use `_getApproved`. But then we need to protect against return value 0.

From https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4405, I believe this should be the logic for this function:

```solidity
        return (owner != address(0) &&
            (spender == owner ||
                isApprovedForAll(owner, spender) ||
                (_getApproved(tokenId) == spender && spender != address(0))));
```

Formatting is ugly, perhaps we have to refactor it into a series of if/else.

contracts/token/ERC721/ERC721.sol
```diff
@@ -386,8 +392,12 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * Emits an {Approval} event.
      */
     function _approve(address to, uint256 tokenId) internal virtual {
```
frangio, line 394: It is annoying that `approve` is reading the owner twice (the second time in `_approve`). Should we use a similar construction as with `_update` for the operator check?

contracts/token/ERC721/ERC721.sol
```diff
@@ -386,8 +392,12 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+        address owner = ownerOf(tokenId);
+        if (to == owner) {
+            revert ERC721InvalidOperator(to);
```
frangio, line 397: This check came from https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4405, but I don't know why it was there in the first place? We allow approve to self in both ERC1155 and ERC20, so it doesn't seem like a matter of consistency.

I would just remove it.

contracts/token/ERC721/ERC721.sol
```diff
@@ -269,34 +280,37 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      *
      * Emits a {Transfer} event.
      */
-    function _mint(address to, uint256 tokenId) internal virtual {
+    function _mint(address to, uint256 tokenId) internal {
         if (to == address(0)) {
             revert ERC721InvalidReceiver(address(0));
         }
-        if (_exists(tokenId)) {
```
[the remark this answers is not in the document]
frangio: Yeah I don't really know why this function exists (:speak_no_evil:). Overriding it should have no effect either.

I'd be in favor of keeping `_requireMinted` only.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: > Setting an approval prior to minting seems like a bug more than a feature to me.

It depends on how you brand it 😎 (I'm joking)

Jokes aside, I agree, I was trying to make a case in which the `delete tokenApprovals[tokenId]` in `_update` can be used for skipping an approval deletion.

Let's keep it as it is right now, but we'll need to take a double look into the interface descriptions. See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#discussion_r1261806819

contracts/token/ERC721/ERC721.sol
ernestognw: Related to https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#discussion_r1261790066

contracts/token/ERC721/ERC721.sol
```diff
@@ -152,57 +150,37 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * @dev See {IERC721-transferFrom}.
      */
     function transferFrom(address from, address to, uint256 tokenId) public virtual {
-        if (!_isApprovedOrOwner(_msgSender(), tokenId)) {
-            revert ERC721InsufficientApproval(_msgSender(), tokenId);
+        if (to == address(0)) {
+            revert ERC721InvalidReceiver(address(0));
+        }
+        address owner = _update(to, tokenId, _msgSender());
```
frangio, line 156: Thoughts on `previousOwner` instead of `owner`?

contracts/token/ERC721/ERC721.sol
```diff
@@ -221,40 +199,73 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
      *
-     * Requirements:
-     *
-     * - `tokenId` must not exist.
-     * - If `to` refers to a smart contract, it must implement {IERC721Receiver-onERC721Received}, which is called upon a safe transfer.
+     * WARNING:
+     * - ownership is not checked.
+     * - spender = address(0) will lead to this function returning true in unexpected situations.
+     */
+    function _isApproved(address owner, address spender, uint256 tokenId) internal view virtual returns (bool) {
```
frangio, line 216: I'm concerned that the meaning of `_isApproved`/`_checkApproved` is very different from the meaning of `_getApproved`.


I would rename `_is/checkApproved` to `_is/checkOperator`.

contracts/token/ERC721/ERC721.sol
```diff
@@ -221,40 +199,73 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     * The `operatorCheck` argument is optional. If the value passed is non 0, then this function will check that
+     * `operatorCheck` is either the owner of the token, or approved to operate on the token (by the owner).
+     *
+     * Emits a {Transfer} event.
      */
-    function _safeMint(address to, uint256 tokenId, bytes memory data) internal virtual {
-        _mint(to, tokenId);
-        if (!_checkOnERC721Received(address(0), to, tokenId, data)) {
-            revert ERC721InvalidReceiver(to);
+    function _update(address to, uint256 tokenId, address operatorCheck) internal virtual returns (address) {
```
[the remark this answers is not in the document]
frangio: I am not convinced about either `operatorCheck` or `expectedOperator`. My issue with `expectedOperator` is that there isn't just one operator that we will check to be this one, there are potentially multiple operators and this is just the current one.

Perhaps the name should be something to indicate that the value is used for "authorization". `address auth`? It also gives a hint that it's an important variable that shouldn't be touched (in an override).

We could pair this with renaming `_is/checkApproved` to `_is/checkAuthorized` (instead of `_is/checkOperator` like I suggested elsewhere).

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: I'm still uncomfortable with an internal function using a public function. Here's another idea.

There's another comment asking for the `_exists(uint256)` function (https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377/files#r1261760016), I think we can make a `_checkExists(address owner, uint256 tokenId)` that doesn't check actual ownership (as in `_isApproved`) and reverts with `ERC721NonexistentToken`. We're making that check in multiple places and make sense having a function for that.

ernestognw (commented):
I think we still need tests for the following cases:

1. `_setApprovalForAll ` should revert when `operator == address(0)`
2. Both `_safeTransfer` functions aren't covered, including the `_transfer`
3. `transferFrom` should revert when `owner == address(0)`

contracts/token/ERC721/ERC721.sol
```diff
@@ -269,34 +280,37 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      *
      * Emits a {Transfer} event.
      */
-    function _mint(address to, uint256 tokenId) internal virtual {
+    function _mint(address to, uint256 tokenId) internal {
         if (to == address(0)) {
             revert ERC721InvalidReceiver(address(0));
         }
-        if (_exists(tokenId)) {
+        address from = _update(to, tokenId, address(0));
```
frangio, line 287: We should use the same variable name here, either `owner` or `previousOwner`.

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,41 +342,48 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     * - `from` cannot be the zero address.
+     * - `to` cannot be the zero address.
+     * - `tokenId` token must exist and be owned by `from`.
+     * - If `to` refers to a smart contract, it must implement {IERC721Receiver-onERC721Received}, which is called upon a safe transfer.
+     *
+     * Emits a {Transfer} event.
+     */
+    function _safeTransfer(address from, address to, uint256 tokenId) internal {
+        _transfer(from, to, tokenId);
+        _checkOnERC721Received(from, to, tokenId, "");
```
frangio, line 377:
```suggestion
        _safeTransfer(from, to, tokenId, "");
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -396,8 +406,8 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * Emits an {ApprovalForAll} event.
      */
     function _setApprovalForAll(address owner, address operator, bool approved) internal virtual {
-        if (owner == operator) {
-            revert ERC721InvalidOperator(owner);
+        if (operator == owner || operator == address(0)) {
```
frangio, line 409: Same as in my other comment I don't really know why we would disallow `operator == owner`. I think we need to remove that check.

contracts/token/ERC721/ERC721.sol
```diff
@@ -441,52 +447,20 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     * NOTE: the value is limited to type(uint128).max. This protect against _balance overflow. It is unrealistic that
+     * a uint256 would ever overflow from increments when these increments are bounded to uint128 values.
+     *
+     * WARNING: Increassing an account's balance using this function should go in pair with an override of the
+     * {_ownerOf} function that resolve the ownership of the corresponding tokens so that balances and ownerships
+     * remain consistent with one another.
      */
-    // solhint-disable-next-line func-name-mixedcase
-    function __unsafe_increaseBalance(address account, uint256 value) internal {
+    function _increaseBalance(address account, uint128 value) internal virtual {
```
frangio, line 463: Should be moved next to `_update` IMO.

contracts/token/ERC721/ERC721.sol
```diff
@@ -152,57 +150,37 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-        }
+        transferFrom(from, to, tokenId);
+        _checkOnERC721Received(from, to, tokenId, data);
     }
 
     /**
      * @dev Returns the owner of the `tokenId`. Does NOT revert if token doesn't exist
+     *
+     * WARNING: Any override of this function that allocates token to accounts following a custom logic should go in
+     * pair with a call to {_increaseBalance} so that balances and ownerships remain consistent with one another.
```
frangio, line 183: A little more detailed I think:

```suggestion
      * IMPORTANT: Any overrides to this function that add ownership of tokens not tracked by the 
      * core ERC721 logic MUST be matched with the use of {_increaseBalance} to keep balances 
      * consistent with ownership. The invariant to preserve is that for any address `a` the value returned by 
      * `balanceOf(a)` must be equal to the number of tokens such that `_ownerOf(tokenId)` is `a`. 
```

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
frangio: Yeah I don't really know why this function exists (:speak_no_evil:). Overriding it should have no effect either.

I'd be in favor of keeping `_requireMinted` only.

contracts/token/ERC721/ERC721.sol
```diff
@@ -221,40 +199,73 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+        return owner == spender || isApprovedForAll(owner, spender) || getApproved(tokenId) == spender;
+    }
+
+    /**
+     * @dev Checks whether `spender` was approved by `owner` to operate on `tokenId` or is the owner of `tokenId`.
+     * Reverts if `spender` is not approved nor owner of `tokenId`.
      *
-     * Emits a {Transfer} event.
+     * WARNING:
+     * - ownership is not checked.
```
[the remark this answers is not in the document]
frangio: Hm, "Requirements" tend to be "the function will revert if you don't follow the requirements". In this case, it's not a requirement, more like an assumption. I think the warning as it is written makes sense.

contracts/token/ERC721/ERC721.sol
```diff
@@ -441,52 +447,20 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
      *
-     * WARNING: Anyone calling this MUST ensure that the balances remain consistent with the ownership. The invariant
-     * being that for any address `a` the value returned by `balanceOf(a)` must be equal to the number of tokens such
-     * that `ownerOf(tokenId)` is `a`.
+     * NOTE: the value is limited to type(uint128).max. This protect against _balance overflow. It is unrealistic that
+     * a uint256 would ever overflow from increments when these increments are bounded to uint128 values.
+     *
+     * WARNING: Increassing an account's balance using this function should go in pair with an override of the
+     * {_ownerOf} function that resolve the ownership of the corresponding tokens so that balances and ownerships
+     * remain consistent with one another.
```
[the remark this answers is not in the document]
frangio: > This is generally true but in the case of ERC721Votes it's not.

I wouldn't say this. ERC721Votes is not invoking `_increaseBalance`, it's just overriding it to react to when it's invoked by another module.

That said, "should go" or "usually goes" both work for me.

contracts/token/ERC721/ERC721.sol
```diff
@@ -413,24 +423,20 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-     * @return bool whether the call correctly returned the expected magic value
-     */
-    function _checkOnERC721Received(
-        address from,
-        address to,
-        uint256 tokenId,
-        bytes memory data
-    ) private returns (bool) {
+     */
+    function _checkOnERC721Received(address from, address to, uint256 tokenId, bytes memory data) private {
```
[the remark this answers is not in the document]
frangio: At this point both PRs will need to adapt to the other. I would rather merge this one first, which is more complex, and adapt the other one later which will be easier.

contracts/token/ERC721/ERC721.sol
```diff
@@ -114,12 +114,10 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      */
     function approve(address to, uint256 tokenId) public virtual {
```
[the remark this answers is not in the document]
frangio: >However, we can no longer approve the zero address because of the ownerOf in _approve

I don't think this is true. `_approve(to=address(0), tokenId)` should work normally.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
frangio: I am not convinced about either `operatorCheck` or `expectedOperator`. My issue with `expectedOperator` is that there isn't just one operator that we will check to be this one, there are potentially multiple operators and this is just the current one.

Perhaps the name should be something to indicate that the value is used for "authorization". `address auth`? It also gives a hint that it's an important variable that shouldn't be touched (in an override).

We could pair this with renaming `_is/checkApproved` to `_is/checkAuthorized` (instead of `_is/checkOperator` like I suggested elsewhere).

contracts/token/ERC721/extensions/ERC721Votes.sol
ernestognw: Ok no closing, this was an incorrect suggestion, the problem is that we clearly need `from` first.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: You're right, I misinterpreted the semantics of `_approve`. I thought we were restricting `to` instead of `tokenId`'s owner.

Solving.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: I see, and agree with the thinking. My concern is that it's explained that any `ownerOf` override should go along with an `_increaseBalance` but an `_increaseBalance` on its own might not be obvious. And because this is the documentation of `_increaseBalance`, the NatSpec should suggest what to consider when overriding.

My concrete suggestion is now to add an `IMPORTANT` block in the NatSpec of `ownerOf`:

```solidity
/**
 * @dev See {IERC721-ownerOf}.
 *
 * IMPORTANT: Any override that resolves the ownership by grouping sequential `tokenId`s (i.e. batching)
 * should be in pair with an override of the {_increaseBalance} function to keep balances consistent with
 * the {balanceOf} internal accounting accounting,
 */
```

And change this one to the following `IMPORTANT` block:

```suggestion
     * IMPORTANT: The {_update} function doesn't internally depend on this function as it updates the balances
     * by a fixed value of `1`. When implementing any mechanism that is triggered on `_update` invocations,
     * consider also overriding this function to support nonstandard balance updates.
```

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: I like this idea really much! I would pick `checkAuthorized` from the options you gave. The rationale is that we've consistently named `_check` to the reverting functions (eg. `_checkOnERC721Received`) and `isAuthorized` might suggest a boolean returning value.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: Yeah let's do `_requireMinted` only, it's the same.

I would suggest `_checkMinted` to be consistent with the `_check` convention. Not strong about it but perhaps you might like it.

contracts/token/ERC721/ERC721.sol
ernestognw: There's already a `_checkApproved` so it will conflict.

What about the convention of?:

1. `_is*` for boolean returning values (eg. `_isApproved`)
2. `_check`for functions reverting when `_is*` functions return false (eg. `_checkApproved`)

contracts/token/ERC721/ERC721.sol
ernestognw: I'd be in favor of `previousOwner`

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
Amxx: So If I follow correctly:
- `_isApproved` → `_isAuthorised` (internal function)
- `_checkApproved` → `_checkAuthorised` (internal function)
- `operatorCheck` → `auth` (function parameter)

Do we all agree on that ?

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
Amxx: I guess there was a point to if when you wanted to do different logic (like calling `_mint` instead of `_transfer`) based on some logic. Also, at some point we did not have `_ownerOf` but only the public `ownerOf` that revert is 0 ... so that was literraly the only way to know if a token is minted or not that would not revert if it does not exist.

Fells like its no longer needed, so I'm removing it.

contracts/token/ERC721/ERC721.sol
Amxx: what about 
```
function _isApproved(address owner, address spender, uint256 tokenId) internal view virtual returns (bool) {
    return
        spender != address(0) &&
        (owner == spender || isApprovedForAll(owner, spender) || _getApproved(tokenId) == spender);
}
```
?

Amxx: if owner is 0, that would return false because:
- spender must be non 0, and so `owner != spender`
- address 0 should not be able to set approval for all (that is not something we check is `_setApprovalForAll`, but we could)
- non minted token should not have an approval, so if spender is 0, `_getApproved(tokenId) != spender`

contracts/token/ERC721/ERC721.sol
Amxx: changed to previousOwner

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
Amxx: A `_checkExists` that does not check the value given, and just reverts (with some custom) error if the first input is 0 doesn't feel like an helper function I'd want.

There are already so many in ERC721, i'd like the number to go down, not up.

@frangio WDYT?

contracts/token/ERC721/ERC721.sol
Amxx: I thought it was an ERC requirement, if its not we should indeed remove it.

contracts/token/ERC721/ERC721.sol
Amxx: Note: currently if you call _isApproved on with a tokenId that doesn't exist:
- the owner is fetched with `_ownerOf`, that is 0
- the owner is not the spender
- the owner has not approved the spender
- getApproved is called, and that reverts with a `ERC721NonexistentToken`

With this change the function returns false, and we get an `ERC721InsufficientApproval` returned. So to keep the errors, we'll have to add some checks higher in the call stack

contracts/token/ERC721/ERC721.sol
Amxx: what about in the extensions like ERC721Enumerable

```solidity
    function _update(address to, uint256 tokenId, address auth) internal virtual override returns (address) {
        address from = super._update(to, tokenId, auth);

        if (from == address(0)) {
            _addTokenToAllTokensEnumeration(tokenId);
        } else if (from != to) {
            _removeTokenFromOwnerEnumeration(from, tokenId);
        }
        if (to == address(0)) {
            _removeTokenFromAllTokensEnumeration(tokenId);
        } else if (from != to) {
            _addTokenToOwnerEnumeration(to, tokenId);
        }

        return from;
    }
```

contracts/token/ERC721/ERC721.sol
frangio: I couldn't find it, but please double check.

contracts/token/ERC721/ERC721.sol
frangio: I would rather be consistent in there too.

contracts/token/ERC721/ERC721.sol
Amxx: I "thought" (like yesterday). Could not find it either.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
frangio: Yes but `authorized` is spelled with a "z" in American English which is what we use.

contracts/token/ERC721/ERC721.sol
Amxx: done

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
Amxx: done

contracts/token/ERC721/ERC721.sol
Amxx: Assuming the other invariants of the contract hold (things like address(0) never approves anyone), I expect `_isAuthorized` to return false if owner is 0. So the extra `if` in `_checkAuthorized` is not a divergence from `_isAuthorized`.

Still we could rewrite it as:
```solidity
        if (!_isAuthorized(owner, spender, tokenId)) {
            if (owner == address(0)) {
                revert ERC721NonexistentToken(tokenId);
            } else {
                revert ERC721InsufficientApproval(spender, tokenId);
            }
        }
```

Amxx: I did the rewrite. For the record the previous version was
```solidity
        if (owner == address(0)) {
            revert ERC721NonexistentToken(tokenId);
        } else if (!_isAuthorized(owner, spender, tokenId)) {
            revert ERC721InsufficientApproval(spender, tokenId);
        }
```

Let me know if you prefer the previous version.

ernestognw (commented):
It's looking pretty good! We're close

Amxx:
Coverage is "broken" because the _transfer and _safeTransfer function are no longer triggered by the external version

contracts/mocks/token/ERC721ConsecutiveMock.sol
```diff
@@ -41,26 +41,17 @@ contract ERC721ConsecutiveMock is ERC721Consecutive, ERC721Pausable, ERC721Votes
...
     }
 
-    function _afterTokenTransfer(
-        address from,
-        address to,
-        uint256 firstTokenId,
-        uint256 batchSize
-    ) internal virtual override(ERC721, ERC721Votes, ERC721Consecutive) {
-        super._afterTokenTransfer(from, to, firstTokenId, batchSize);
+    // solhint-disable-next-line func-name-mixedcase
```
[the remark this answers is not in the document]
frangio: No, removed.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
frangio: Personally I would also prefer to avoid such a function. Though I think there is some credit to this concern:
>I'm still uncomfortable with an internal function using a public function.
The way that I justify it is that `ownerOf` is "expected to be" simply `_ownerOf` + nonzero check, and I can't think of any actual issues this (internal using a public function) might cause.

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
frangio: I added a note inspired by your last suggestion in `_update` which I think is the right place...

contracts/token/ERC721/ERC721.sol
frangio: Fixed in ca32b459e.

contracts/token/ERC721/ERC721.sol
frangio: Yes this is the convention we seem to be using.

contracts/token/ERC721/extensions/ERC721Burnable.sol
```diff
@@ -19,9 +19,8 @@ abstract contract ERC721Burnable is Context, ERC721 {
      * - The caller must own `tokenId` or be an approved operator.
      */
     function burn(uint256 tokenId) public virtual {
-        if (!_isApprovedOrOwner(_msgSender(), tokenId)) {
-            revert ERC721InsufficientApproval(_msgSender(), tokenId);
-        }
-        _burn(tokenId);
+        // Setting an "auth" arguments enables the `_isApproved` check which verifies that the token exists
+        // (from != 0). Therefore, it is not needed to verify that the return value is not 0 here.
+        _update(address(0), tokenId, _msgSender());
```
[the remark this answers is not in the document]
frangio: Even though `_update` is a lower level function it is still available and can be used, so I don't think we need the more complex `_burn` definition. If we added that function, it sounds like we should also add `_mint(..., auth)` and `_transfer(..., auth)`, etc., and I don't think we should.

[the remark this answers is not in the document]
ernestognw: Then I'd advocate for removing the `_burn` function. The fact we're not using the `_burn` function here because it doesn't meet our needs is imo strong evidence that the function would also be not enough for several use cases.

We don't need `auth` neither in `_transfer` or `_mint` because the former don't require operator check because it checks the `from`, and the later doesn't require operator check because it doesn't belong to anyone in the first place. I think `_burn` needs an `auth` because it's destroying someone else's property and there's no `from` check.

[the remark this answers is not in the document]
Amxx: The current `_mint`, `_burn` and `_transfer` are designed to copy the 4.X behavior. This is so the migration would be simplified. Indeed, it is not as powerfull as the new `_update`.

Its a balance betten simplifying migration through functions that are not suited for all usecases, or removing them and force users to addapt to the new pattern.

contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -55,7 +55,7 @@ abstract contract ERC721URIStorage is IERC4906, ERC721 {
      * - `tokenId` must exist.
```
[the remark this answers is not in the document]
Amxx: We replaced `if (!_exists(tokenId)) {` with `if (_ownerOf(tokenId) == address(0)) {`
My understanding is that is still true.

contracts/token/ERC721/ERC721.sol
```diff
@@ -339,65 +346,87 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     *
+     * This internal function is like {safeTransferFrom} in the sense that it invokes
+     * {IERC721Receiver-onERC721Received} on the receiver, and can be used to e.g.
+     * implement alternative mechanisms to perform token transfer, such as signature-based.
+     *
+     * Requirements:
+     *
+     * - `from` cannot be the zero address.
+     * - `to` cannot be the zero address.
+     * - `tokenId` token must exist and be owned by `from`.
```
Amxx, line 379:
```suggestion
     * - `tokenId` token must exist and be owned by `from`.
     * - `to` cannot be the zero address.
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -152,109 +143,137 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     * particular (ignoring whether it is owned by `owner`).
      *
-     * Tokens can be managed by their owner or approved accounts via {approve} or {setApprovalForAll}.
-     *
-     * Tokens start existing when they are minted (`_mint`),
-     * and stop existing when they are burned (`_burn`).
+     * WARNING: This function doesn't check that `owner` is the actual owner of the specified
+     * `tokenId`. Moreover, it returns true if `owner == spender` in order to skip the check where
+     * needed. Consider checking for cases where `spender == address(0)` since they could lead to
+     * unexpected behavior.
```
Amxx, line 198:
```suggestion
     * needed.
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -152,109 +143,137 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
     /**
-     * @dev Returns whether `spender` is allowed to manage `tokenId`.
+     * @dev Checks if `spender` can operate on `tokenId`, assuming the provided `owner` is the actual owner.
+     * Reverts if `spender` has not approval for all assets of the provided `owner` nor the actual owner approved the `spender` for the specific `tokenId`.
      *
-     * Requirements:
-     *
-     * - `tokenId` must exist.
+     * WARNING: This function relies on {_isAuthorized}, so it doesn't check whether `owner` is the
+     * actual owner of `tokenId` and will skip the check if `spender == owner`
```
Amxx, line 211:
```suggestion
     * actual owner of `tokenId`.
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -339,65 +343,85 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * Requirements:
      *
      * - `to` cannot be the zero address.
-     * - `tokenId` token must be owned by `from`.
+     * - `tokenId` token must exists and be owned by `from`.
```
ernestognw, line 346: This is implied because `to` cannot be the zero address.

```suggestion
     * - `tokenId` token must be owned by `from`.
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -339,65 +343,85 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     * `data` is additional data, it has no specified format and it is sent in call to `to`.
+     *
+     * This internal function is like {safeTransferFrom} in the sense that it invokes
+     * {IERC721Receiver-onERC721Received} on the receiver, and can be used to e.g.
+     * implement alternative mechanisms to perform token transfer, such as signature-based.
+     *
+     * Requirements:
+     *
+     * - `tokenId` token must exist and be owned by `from`.
+     * - `to` cannot be the zero address.
```
ernestognw, line 375:
```suggestion
     * - `to` cannot be the zero address.
     * - `from` cannot be the zero address.
```

This is true because `_transfer` checks the following:

```solidity
if (previousOwner == address(0)) {
    revert ERC721NonexistentToken(tokenId);
} else if (previousOwner != from) {
    revert ERC721IncorrectOwner(from, tokenId, previousOwner);
}
```

Both cases completely exclude `from == address(0)`.

contracts/token/ERC721/ERC721.sol
```diff
@@ -192,10 +192,7 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * @dev Returns whether `spender` is allowed to manage `owner`'s tokens, or `tokenId` in
      * particular (ignoring whether it is owned by `owner`).
      *
-     * WARNING: This function doesn't check that `owner` is the actual owner of the specified
-     * `tokenId`. Moreover, it returns true if `owner == spender` in order to skip the check where
-     * needed. Consider checking for cases where `spender == address(0)` since they could lead to
-     * unexpected behavior.
```
ernestognw: Why did we simplified this writing? I think it's still important as someone may think by providing a correct `owner` the `spender` doesn't matter. It make sense to me advising for not defaulting to `spender` to `owner`

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: Solving agreeing with keeping it.

ernestognw (approved):
Just nits, this is pretty much done imo

ernestognw (approved):
LGTM

vladyan18:
I find this line causing some inconvenience:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/token/ERC721/ERC721.sol#L254

Taking into account the fact that both the process of [approving ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/token/ERC721/ERC721.sol#L399) and [getting an approved address](_getApproved) are placed in separate virtual functions and can be overriden. In this case, if the internal mechanism is somehow replaced, the above line will remain (or you need to redefine the entire `_update `function), creating unnecessary gas waste.

Sorry for posting here in a merged pull request. If needed, I can open a new issue.

Amxx:
[We used to do `_approve(address(0), tokenId);`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.7/contracts/token/ERC721/ERC721.sol#L341) but this was changed to avoid emiting the `Approval` event.

vladyan18:
> [We used to do `_approve(address(0), tokenId);`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.7/contracts/token/ERC721/ERC721.sol#L341) but this was changed to avoid emiting the `Approval` event.


Yes, I found what I described above when switching to version 4.8.0.

> We looked for, and could not find updates to the approval mechanism.

Example of such a approval machanism change (albeit for the old version):
https://github.com/Uniswap/v3-periphery/blob/6cce88e63e176af1ddb6cc56e029110289622317/contracts/NonfungiblePositionManager.sol#L395-L399

vladyan18:
Maybe it would be convenient to have a low-level virtual function in the style:
```solidity
function _setApprovalForToken(address approval, uint256 tokenId) internal virtual {
  _tokenApprovals[tokenId] = approval;
}
```

I think using such a function with zero address as input is no different from "delete" except in a less explicit syntax.

Amxx:
We don't propose these kind of functions because calling them would set the state without emitting the corresponding event.

frangio:
We can do the same we did for ERC20:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/token/ERC20/ERC20.sol#L346-L348

vladyan18:
> We can do the same we did for ERC20:

It looks like this will allow to avoid unnecessary SLOAD/SSTORE and replace it with custom logic at the cost of additional logical branching. On the other hand, this allows, if desired, to bypass the emit of the `Approval` event completely

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -74,33 +74,23 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
...
-        address from,
-        address to,
-        uint256 firstTokenId,
-        uint256 batchSize
-    ) internal virtual override {
-        super._beforeTokenTransfer(from, to, firstTokenId, batchSize);
-
-        if (batchSize > 1) {
-            // Will only trigger during construction. Batch transferring (minting) is not available afterwards.
-            revert ERC721EnumerableForbiddenBatchMint();
```
selarun13, line 89: 0x6F99E90Dd385c23826b416528e325214671FeBf6

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -74,33 +74,23 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
...
-        }
-
-        uint256 tokenId = firstTokenId;
+    function _update(address to, uint256 tokenId, address auth) internal virtual override returns (address) {
+        address previousOwner = super._update(to, tokenId, auth);
 
-        if (from == address(0)) {
+        if (previousOwner == address(0)) {
```
selarun13, line 82: 0x6F99E90Dd385c23826b416528e325214671FeBf6

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -74,33 +74,23 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
...
             _addTokenToAllTokensEnumeration(tokenId);
-        } else if (from != to) {
-            _removeTokenFromOwnerEnumeration(from, tokenId);
```
selarun13, line 97: 0x6F99E90Dd385c23826b416528e325214671FeBf6

contracts/token/ERC721/extensions/ERC721Enumerable.sol
```diff
@@ -74,33 +74,23 @@ abstract contract ERC721Enumerable is ERC721, IERC721Enumerable {
...
+            _removeTokenFromOwnerEnumeration(previousOwner, tokenId);
         }
         if (to == address(0)) {
             _removeTokenFromAllTokensEnumeration(tokenId);
-        } else if (to != from) {
+        } else if (previousOwner != to) {
             _addTokenToOwnerEnumeration(to, tokenId);
         }
+
+        return previousOwner;
```
selarun13, line 93: 0x6F99E90Dd385c23826b416528e325214671FeBf6

contracts/token/ERC721/extensions/ERC721Enumerable.sol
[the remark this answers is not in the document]
selarun13: 0x6F99E90Dd385c23826b416528e325214671FeBf6

selarun13 (commented):
0x6F99E90Dd385c23826b416528e325214671FeBf6

contracts/token/ERC721/ERC721.sol
```diff
@@ -152,109 +143,135 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+        if (from != address(0)) {
+            delete _tokenApprovals[tokenId];
+            unchecked {
+                _balances[from] -= 1;
+            }
+        }
+
+        if (to != address(0)) {
+            unchecked {
+                _balances[to] += 1;
```
vladyan18, line 266: Why not `increaseBalance(…)` btw? It may be logical again to use a function that can be overridden (then a similar point with a decrease). But of course it can cost extra gas if it doesn't inline

contracts/token/ERC721/ERC721.sol
```diff
@@ -152,109 +143,135 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+     *
+     * Emits a {Transfer} event.
+     *
+     * NOTE: If overriding this function in a way that tracks balances, see also {_increaseBalance}.
      */
-    function _safeMint(address to, uint256 tokenId, bytes memory data) internal virtual {
-        _mint(to, tokenId);
-        if (!_checkOnERC721Received(address(0), to, tokenId, data)) {
-            revert ERC721InvalidReceiver(to);
+    function _update(address to, uint256 tokenId, address auth) internal virtual returns (address) {
```
vladyan18, line 248: Isn’t `_update` too “generic” name for this function? Why not something like `_updateOwner`, `_changeOwner`, or even `_transfer` (and current non-virtual `_transfer` -> `_transferChecked` or smth)

contracts/token/ERC721/ERC721.sol
frangio: If we use `_increaseBalance` here, there is another risk regarding overrides, which is that an extension overriding both `_update` and `_increaseBalance` may end up executing its logic twice. See `ERC721Votes` for example, which overrides both in order to use `_transferVotingUnits`.

vladyan18: Maybe then give the function a name that more specifically reflects its purpose?

vladyan18: > Which one do you mean? `_update`?

`_increaseBalance` 
Maybe like `_increaseBalanceUnsafe`, `_increaseBalanceDirectly`, `_increaseBalanceManually` or smth

PR #4378, opened 2023-06-22
Add Governor signature nonces

ernestognw:
Fixes LIB-895

This PR adds `nonce` and `voter` arguments to governor vote casting signatures with the following objectives:

- The nonce is added so that signatures can be invalidated.
- The voter address guarantees that a signature can't be forged for any random address.

#### PR Checklist

- [x] Tests
- [ ] Documentation


frangio:
To remove the initcode size warning that is causing the upgradeable tests to fail, we have to update the `hardhat-ignore-warnings` package to the latest version and add in the config:

```diff
    'contracts-exposed/**/*': {
      'code-size': 'off',
+     'initcode-size': 'off',
    },
```

ernestognw:
@SocketSecurity ignore hardhat-ignore-warnings@0.2.9

ernestognw:
CI is failing because the storage layout became incompatible by the addition of `_nonces`, but that's expected

contracts/governance/Governor.sol
```diff
@@ -25,12 +26,15 @@ import "./IGovernor.sol";
  *
  * _Available since v4.3._
  */
-abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receiver, IERC1155Receiver {
+abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC721Receiver, IERC1155Receiver {
     using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
 
-    bytes32 public constant BALLOT_TYPEHASH = keccak256("Ballot(uint256 proposalId,uint8 support)");
+    bytes32 public constant BALLOT_TYPEHASH =
+        keccak256("Ballot(uint256 proposalId,uint8 support,address voter,uint256 nonce)");
```
Amxx, line 33: Do we ave any rational for ordering of these ? My intuition would be to put voter before support. I can't explain why, and it honestly doesn't change much ... but still I think we may want to set some guidelines.

contracts/governance/Governor.sol
```diff
@@ -519,17 +523,25 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
         uint8 v,
         bytes32 r,
         bytes32 s
     ) public virtual override returns (uint256) {
-        address voter = ECDSA.recover(
-            _hashTypedDataV4(keccak256(abi.encode(BALLOT_TYPEHASH, proposalId, support))),
+        uint256 currentNonce = _useNonce(voter);
+
+        address signer = ECDSA.recover(
+            _hashTypedDataV4(keccak256(abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, currentNonce))),
```
Amxx, line 534: How do you feel about:
```suggestion
        address signer = ECDSA.recover(
            _hashTypedDataV4(keccak256(abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, _useNonce(voter)))),
```
?

contracts/governance/Governor.sol
```diff
@@ -538,19 +550,24 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
...
+
+        address signer = ECDSA.recover(
             _hashTypedDataV4(
                 keccak256(
                     abi.encode(
                         EXTENDED_BALLOT_TYPEHASH,
                         proposalId,
                         support,
+                        voter,
+                        currentNonce,
```
Amxx, line 570: same here, I'd do

```suggestion
        address signer = ECDSA.recover(
            _hashTypedDataV4(
                keccak256(
                    abi.encode(
                        EXTENDED_BALLOT_TYPEHASH,
                        proposalId,
                        support,
                        voter,
                        _useNonce(voter),
```

contracts/governance/Governor.sol
ernestognw: I see value in setting a guideline for this, but the main reason for this order is that it's the same order in which I got the parameters of `castVoteWithReasonAndParamsBySig` to work. Other orderings don't compile due to `Stack too deep`.

I actually might consider using a struct memory but wanted to avoid that discussion for now :sweat_smile:

test/governance/Governor.test.js
```diff
@@ -189,13 +191,19 @@ contract('Governor', function (accounts) {
...
+
         // Run proposal
         await this.helper.propose();
         await this.helper.waitForSnapshot();
-        expectEvent(await this.helper.vote({ support: Enums.VoteType.For, signature }), 'VoteCast', {
-          voter: voterBySigAddress,
-          support: Enums.VoteType.For,
-        });
+        expectEvent(
+          await this.helper.vote({ support: Enums.VoteType.For, voter: voterBySigAddress, nonce: nonce, signature }),
```
Amxx, line 200:
```suggestion
          await this.helper.vote({ support: Enums.VoteType.For, voter: voterBySigAddress, nonce, signature }),
```

test/governance/extensions/GovernorWithParams.test.js
```diff
@@ -119,56 +120,102 @@ contract('GovernorWithParams', function (accounts) {
...
-        // Run proposal
-        await this.helper.propose();
-        await this.helper.waitForSnapshot();
+          const nonce = await this.mock.nonces(this.voterBySig.address);
 
-        const weight = web3.utils.toBN(web3.utils.toWei('7')).sub(rawParams.uintParam);
+          const tx = await this.helper.vote({
+            support: Enums.VoteType.For,
+            voter: this.voterBySig.address,
+            nonce: nonce,
```
Amxx, line 164:
```suggestion
            nonce,
```

Amxx (commented):
Very small comments about formating. 

What I'm worried about is not the code, its the "breaking the voteBySig" without params.
The params part I don't really worry about, but the the without params might be breaking tally. Do we have info on that ?

contracts/governance/Governor.sol
Amxx: wow ok. That is both a terrible reason, and a very practical one :P

ernestognw: It's sort of a deadline in a sense 🧠 

By the way, I removed the `currentNonce` variable as suggested so maybe there's now room for reordering these but I still have no criteria. I'd invoke @frangio again for an opinion (if any)

imagen(5)

ernestognw:
> Very small comments about formating.
> 
> What I'm worried about is not the code, its the "breaking the voteBySig" without params. The params part I don't really worry about, but the the without params might be breaking tally. Do we have info on that ?

We asked to Tally yesterday and they don't seem to have any major concerns 👍🏻

contracts/governance/Governor.sol
frangio: I don't think ordering is important so I would leave as is.

contracts/governance/Governor.sol
```diff
@@ -519,17 +523,23 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
     function castVoteBySig(
         uint256 proposalId,
         uint8 support,
+        address voter,
```
frangio, line 526: This is changing the ERC-165 ids, i.e. removing the previously supported interfaces (correctly) and adding new ones. However, we've discussed before that ERC-165 is not a good fit for this contract so we should really think whether we want to continue using it and particularly adding new interfaces.

No need to act on this PR, but it is an unsolved question we need to address.

.changeset/sixty-numbers-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`Governor`: Add a `voter` parameter to the `castVoteBySig` and `castVoteWithReasonAndParamsBySig` functions to avoid forging signatures for random addresses and allow the use of an internal nonce for invalidating signatures.
```
frangio, line 5: Mostly highlighting what I think is most important about this change:
```suggestion
`Governor`: Add `voter` and `nonce` parameters in signed ballots, to avoid forging signatures for random addresses, prevent signature replay, and allow invalidating signatures. Add `voter` as a new parameter in the `castVoteBySig` and `castVoteWithReasonAndParamsBySig` functions.
```

test/governance/Governor.test.js
```diff
@@ -297,6 +306,73 @@ contract('Governor', function (accounts) {
...
+                voter: this.voterBySig.address,
+                nonce,
+                signature: (...params) => {
+                  const sig = this.signature(...params);
+                  sig[69] ^= 0xff;
+                  return sig;
+                },
+              }),
+              'GovernorInvalidSigner',
+              [],
```
frangio, line 355: Doesn't this error have arguments?

contracts/governance/Governor.sol
```diff
@@ -538,19 +548,22 @@ abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receive
     function castVoteWithReasonAndParamsBySig(
         uint256 proposalId,
         uint8 support,
+        address voter,
         string calldata reason,
         bytes memory params,
         uint8 v,
         bytes32 r,
         bytes32 s
```
frangio, line 556: Given that we are changing the signature of the `castVoteBySig` functions, what do you think about changing these arguments to `bytes signature` and using the opportunity to add EIP-1271 compatibility?

I don't think we should do this here so it's not a blocker for this PR.

frangio (commented):
Minor comments and side notes.

contracts/governance/Governor.sol
frangio: This also aligns really well with the addition of the `voter` argument, which is needed for EIP-1271.

Amxx (dismissed):
LGTM,

+1 to frangio idea of changing the signature type and enabling ERC1271 in a followup PR.

test/governance/Governor.test.js
ernestognw: I realized this is an error and the tests were not working properly because they're missing an `await`. I'll mark as draft before I fix that

ernestognw: Fixed

contracts/governance/Governor.sol
ernestognw: According to what we discussed, I'll add the signature changes and 1271 compatibility in a follow up PR

ernestognw:
Upgradeable tests failed because of storage layout again, which is fine.

I'll continue with the signature refactor along with 1271 in another PR

frangio (approved):
LGTM

Amxx (approved):

PR #4379, opened 2023-06-22
Updating Ownable.sol NatSpec with latest comments

### LINKED ISSUE #4368: Ownable's NatSpec incorrectly describes the mechanism for setting the initial owner
I noticed in the most recent version (v4.9.0) that it is now necessary to explicitly specify the contract owner as a constructor argument, rather than having the owner automatically set.

I also saw that the statement "By default, the owner account will be the one that deploys the contract" is no longer accurate.

I suggest it should now say, "The owner account is provided as a constructor argument during deployment."

Hrishikesh-Thakkar:
Fixes #4368 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -10,8 +10,7 @@ import "../utils/Context.sol";
  * there is an account (an owner) that can be granted exclusive access to
  * specific functions.
  *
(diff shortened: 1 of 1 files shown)


Amxx:
Duplicate of #4369

PR #4382, opened 2023-06-23
Make `TransparentUpgradeableProxy` deploy its `ProxyAdmin` and optimize proxy interfaces

ernestognw:
- Makes `TransparentUpgradeableProxy` deploy its own `ProxyAdmin`
- Removes `upgrade`/`upgradeTo` in favor of `upgradeAndCall`/`upgradeToAndCall` from `ProxyAdmin` and `UUPSUpgradeable`, now the `data` argument is ignored if empty.
  - Adds an `UPGRADE_INTERFACE_VERSION` getter that can be used to decide which functions to use in an upgrade to address https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4372#issuecomment-1597884769.

Fixes LIB-947

#### PR Checklist

- [x] Tests
- [x] Documentation


frangio:
Will fix tests now but the current configuration seems to be the best one: `TransparentUpgradeableProxy` having just `upgradeTo`, and `ProxyAdmin` having just `upgradeToAndCall`. `TransparentUpgradeableProxy` deploys an instance of `ProxyAdmin` in the constructor, but `ProxyAdmin` is not strictly 1:1 because `upgradeToAndCall` receives the proxy address as an argument.

|                            | ProxyAdmin | TransparentUpgradeableProxy | total   |
|----------------------------|------------|-----------------------------|---------|
| baseline (4.9)             | 443289     | 630553                      | 1073842 |
| baseline (`master`)        | 321406     | 445272                      | 766678  |
| reduced interface          | 339872     | 298171                      | 638043  |
| embedded                   | 339896     | 618300                      | 618300  |
| embedded + 1:1 (storage)   | 363803     | 641995                      | 641995  |
| embedded + 1:1 (immutable) | 351271     | 629492                      | 629492  |

"Reduced interface" means proxy and proxy admin have just one of `upgradeTo`/`upgradeToAndCall`.

contracts/proxy/transparent/ProxyAdmin.sol
```diff
@@ -12,6 +12,8 @@ import {Address} from "../../utils/Address.sol";
  * explanation of why you would want to use this see the documentation for {TransparentUpgradeableProxy}.
  */
 contract ProxyAdmin is Ownable {
+    uint256 public immutable PROXY_ADMIN_VERSION = 2;
```
Amxx, line 15: What about doing a string "5.0.0". It would be less than 32 bytes, so it would fit into a single slot ... so it might not be more expensive than the uint256 approach

contracts/proxy/transparent/ProxyAdmin.sol
```diff
@@ -29,32 +31,7 @@ contract ProxyAdmin is Ownable {
...
-    }
-
-    /**
-     * @dev Upgrades `proxy` to `implementation`. See {TransparentUpgradeableProxy-_dispatchUpgradeTo}.
-     *
-     * Requirements:
-     *
-     * - This contract must be the admin of `proxy`.
-     */
-    function upgrade(ITransparentUpgradeableProxy proxy, address implementation) public virtual {
```
Amxx: Are we not keeping `upgradeTo` after all? I'm ok with that, but that was not what we agreed initially because of the plugin possibly using it. I guess that is what `PROXY_ADMIN_VERSION` is for.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -106,11 +106,11 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
     /**
      * @dev Upgrade the implementation of the proxy.
      */
-    function _dispatchUpgradeTo() private returns (bytes memory) {
-        _requireZeroValue();
+    function _dispatchUpgradeToAndCall() private returns (bytes memory) {
+        bool forceCall = msg.value > 0;
```
Amxx, line 110: At that point, I would remove forceCall altogether. That would slightly reduce the size of the contract (and so the deployment cost)

I don't think anyone would ever do `upgradeToAndCall(newImplementation, "")` with some value, and expect the value to be passed to the receive function to for migrating.

If you are really worried about value being stuck:
- the contract is upgradeable, so there is always a path to recovery
- I'd be ok to doing `if (data.length == 0 && msg.value > 0) revert StupidUpdate();`

Amxx (commented):
Minor comments. We should merge that before the audit

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -55,7 +54,7 @@ interface ITransparentUpgradeableProxy is IERC1967 {
  * WARNING: It is not recommended to extend this contract to add additional external functions. If you do so, the compiler
  * will not check that there are no selector conflicts, due to the note above. A selector clash between any new function
  * and the functions declared in {ITransparentUpgradeableProxy} will be resolved in favor of the new one. This could
- * render the admin operations inaccessible, which could prevent upgradeability. Transparency may also be compromised.
+ * render the `upgradeTo` function inaccessible, preventing upgradeability and compromising transparency.
  */
 contract TransparentUpgradeableProxy is ERC1967Proxy {
     // An immutable address for the admin avoid unnecessary SLOADs before each call
```
ericglau, line 60: Not part of this PR but somewhat related.
```suggestion
    // An immutable address for the admin to avoid unnecessary SLOADs before each call
```

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
Amxx, line 1: moving all of that code into `TransparentUpgradeableProxy.test.js` really affect the review process. Was it really necessary ?

test/proxy/ERC1967/ERC1967Proxy.test.js
```diff
@@ -9,5 +9,5 @@ contract('ERC1967Proxy', function (accounts) {
     return ERC1967Proxy.new(implementation, initData, opts);
```
Amxx, line 9: The calls to `createProxy` in `Proxy.behavior.js` no longer include the proxyAdminAddress. Shouldn't the arguments of this function reflect that ?

contracts/proxy/transparent/ProxyAdmin.sol
ernestognw: I would be in favor of this idea.

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
frangio: I think it's necessary eventually, the behavior is only used there, so the separation makes no sense.

But I agree it's very disruptive for this review so I'll reset it back to where it was and we can remove it later

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
frangio: > I don't think anyone would ever do `upgradeToAndCall(newImplementation, "")` with some value, and expect the value to be passed to the receive function to for migrating.

Why not? People do crazy stuff, there is nothing necessarily stupid about this.

Anyway, I added the revert that you suggested and set `forceCall = false`, though I'm not sure if that is what you meant by "remove forceCall altogether".

test/proxy/ERC1967/ERC1967Proxy.test.js
frangio: Yes. This is fixed now. For the record, the reason I removed the admin/initialOwner from the arguments of this function was that the base Proxy behavior shouldn't need to be aware of the initial owner.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
Amxx: You are saying that using the receive function with value is a valid (re)initialisation path ?

I know understand where all our misunderstandings come from. We are not targeting the same users.

Amxx: > though I'm not sure if that is what you meant by "remove forceCall altogether".

Right, I was thinking of removing it from the library.

contracts/proxy/transparent/ProxyAdmin.sol
```diff
@@ -3,33 +3,26 @@
...
 import {Ownable} from "../../access/Ownable.sol";
+import {ShortStrings, ShortString} from "../../utils/ShortStrings.sol";
 
 /**
  * @dev This is an auxiliary contract meant to be assigned as the admin of a {TransparentUpgradeableProxy}. For an
  * explanation of why you would want to use this see the documentation for {TransparentUpgradeableProxy}.
  */
 contract ProxyAdmin is Ownable {
+    // solhint-disable-next-line var-name-mixedcase
+    ShortString public immutable PROXY_ADMIN_VERSION = ShortStrings.toShortString(string("5.0.0"));
```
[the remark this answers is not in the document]
Amxx: Also, using an immutable will require some processing during the constructor that a constant may not need. Since we are trying to optimise for construction cost (and not access cost) I'm dubious the short string is the way to go.

test/proxy/transparent/ProxyAdmin.test.js
```diff
@@ -19,12 +20,12 @@ contract('ProxyAdmin', function (accounts) {
...
-    this.proxyAdmin = await ProxyAdmin.new(proxyAdminOwner);
-    const proxy = await TransparentUpgradeableProxy.new(
-      this.implementationV1.address,
-      this.proxyAdmin.address,
-      initializeData,
-    );
+    const proxy = await TransparentUpgradeableProxy.new(this.implementationV1.address, proxyAdminOwner, initializeData);
+
+    const proxyNonce = await web3.eth.getTransactionCount(proxy.address);
+    const proxyAdminAddress = computeCreateAddress(proxy.address, proxyNonce - 1); // Nonce already used
```
Amxx, line 26: you could get the proxyAdminAddress by reading the ERC1967 slot

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
```diff
@@ -17,8 +19,21 @@ const InitializableMock = artifacts.require('InitializableMock');
...
-module.exports = function shouldBehaveLikeTransparentUpgradeableProxy(createProxy, accounts) {
-  const [proxyAdminAddress, proxyAdminOwner, anotherAccount] = accounts;
+module.exports = function shouldBehaveLikeTransparentUpgradeableProxy(createProxy, initialOwner, accounts) {
+  const [proxyCreator, anotherAccount] = accounts;
+
+  async function createProxyWithImpersonatedProxyAdmin(logic, initData, opts) {
+    const proxy = await createProxy(logic, initData, { from: proxyCreator, ...opts });
+
+    // Expect proxy admin to be the first and only contract created by the proxy
+    const proxyAdminAddress = computeCreateAddress(proxy.address, 1);
```
Amxx, line 29: same here, we could get the value from the ERC1967 slot

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
```diff
@@ -17,8 +19,21 @@ const InitializableMock = artifacts.require('InitializableMock');
 const DummyImplementation = artifacts.require('DummyImplementation');
 const ClashingImplementation = artifacts.require('ClashingImplementation');
 
-module.exports = function shouldBehaveLikeTransparentUpgradeableProxy(createProxy, accounts) {
-  const [proxyAdminAddress, proxyAdminOwner, anotherAccount] = accounts;
+module.exports = function shouldBehaveLikeTransparentUpgradeableProxy(createProxy, initialOwner, accounts) {
+  const [proxyCreator, anotherAccount] = accounts;
+
+  async function createProxyWithImpersonatedProxyAdmin(logic, initData, opts) {
+    const proxy = await createProxy(logic, initData, { from: proxyCreator, ...opts });
```
Amxx, line 26: the "from" part is not necessary, is it ?
```suggestion
    const proxy = await createProxy(logic, initData, opts);
```

contracts/proxy/transparent/ProxyAdmin.sol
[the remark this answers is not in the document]
frangio: I agree. The getter for this will probably return a `bytes32`. ShortStrings are useful with immutables that are set in the constructor. There is no reason to use an immutable here, a constant can do just as well. So it would be better to use `string public constant PROXY_ADMIN_VERSION = "5.0.0"`.

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
frangio: Replying in the other thread.

test/proxy/transparent/ProxyAdmin.test.js
frangio: I prefer this way because we are testing the assumption that the ProxyAdmin is deployed by the proxy.

Amxx: We need to test both things:
- that the ProxyAdmin is deployed by the proxy
- that the ERC1967 slot is set correctly

So we'll have to get both values and compare them. I would have read the slot here, and add an `it` that check it matched the create prediction ... but doing the opposite also works.

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -70,37 +68,30 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
     error ProxyDeniedAdminAccess();
 
     /**
-     * @dev msg.value is not 0.
+     * @dev The proxy rejected value included with `upgradeToAndCall` that would not execute the `receive` function.
      */
-    error ProxyNonPayableFunction();
+    error ProxyRejectedValue();
```
Amxx, line 73: Is this needed? I don't see where it's used

frangio: Left from an earlier version. Removed now.

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
ernestognw: It's necessary because passing an empty object will throw as "unexpected 3 arguments, 2 are required"

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
```diff
@@ -49,64 +66,31 @@ module.exports = function shouldBehaveLikeTransparentUpgradeableProxy(createProx
...
-        newAdmin: proxyAdminAddress,
+        newAdmin: this.proxyAdminAddress,
       });
     });
 
-    it('sets the admin in the storage', async function () {
-      expect(await getAddressInSlot(this.proxy, AdminSlot)).to.be.equal(proxyAdminAddress);
+    it('sets the proxy admin in storage', async function () {
+      expect(await getAddressInSlot(this.proxy, AdminSlot)).to.be.equal(this.proxyAdminAddress);
     });
```
ernestognw, line 77: We need to check that the proxy admin was deployed with the correct owner

test/proxy/transparent/ProxyAdmin.test.js
ernestognw: We're already testing both cases in the following test:

```js
it('sets the proxy admin in storage with the correct initial owner', async function () {
  expect(await getAddressInSlot(this.proxy, AdminSlot)).to.be.equal(this.proxyAdminAddress);
  const proxyAdmin = await Ownable.at(this.proxyAdminAddress);
  expect(await proxyAdmin.owner()).to.be.equal(initialOwner);
});
```

ernestognw:
I can't LGTM my own PR but consider this comment and LGTM

contracts/proxy/utils/UUPSUpgradeable.sol
```diff
@@ -20,6 +20,17 @@ abstract contract UUPSUpgradeable is IERC1822Proxiable {
...
+     * @dev The version of the upgrade interface of the contract. If this getter is missing, both `upgradeTo(address)`
+     * and `upgradeToAndCall(address,bytes)` are present, and `upgradeTo` must be used if no function should be called,
+     * while `upgradeToAndCall` will invoke the `receive` function if the second argument is the empty byte string.
+     * If the getter returns `"5.0.0"`, only `upgradeToAndCall(address,bytes)` is present, and the second argument must
+     * be the empty byte string if no function should be called, being impossible to invoke the `receive` function
+     * during an upgrade.
+     */
+    // solhint-disable-next-line private-vars-leading-underscore
+    string internal constant UPGRADE_INTERFACE_VERSION = "5.0.0";
+
```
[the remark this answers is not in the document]
frangio: I considered something like this but this variable should continue to say 5.0.0 until the next time we change the upgrade interface, and this may be different for different features.

Amxx:
> It's necessary because passing an empty object will throw as "unexpected 3 arguments, 2 are required"

The issue is that if you call a function with an "empty" opts (`{}`) then web3 will not understand that as an opts, but as an argument. This is strange. The fix is to not forward the opts if its "null/undefined" ... and not assign it a default value.

This allows use to remove a big number of `{from:...}` that were not needed for anything but to make web3 understand that opts is an option object.

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
Amxx: > It's necessary because passing an empty object will throw as "unexpected 3 arguments, 2 are required"

The issue is that if you call a function with an "empty" opts ({}) then web3 will not understand that as an opts, but as an argument. This is strange. 

The fix is to not forward the opts if its "null/undefined" ... and not assign it a default value.

This allows use to remove a big number of {from:...} that were not needed for anything but to make web3 understand that opts is an option object.

(see my commit [a22e182](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4382/commits/a22e182dadc491f00611493f764439045ef42c78))

contracts/proxy/transparent/ProxyAdmin.sol
```diff
@@ -12,24 +12,24 @@ import {Ownable} from "../../access/Ownable.sol";
...
+     * @dev The version of the upgrade interface of the contract. If this getter is missing, both `upgrade(address)`
+     * and `upgradeAndCall(address,bytes)` are present, and `upgradeTo` must be used if no function should be called,
+     * while `upgradeAndCall` will invoke the `receive` function if the second argument is the empty byte string.
+     * If the getter returns `"5.0.0"`, only `upgradeAndCall(address,bytes)` is present, and the second argument must
+     * be the empty byte string if no function should be called, being impossible to invoke the `receive` function
+     * during an upgrade.
      */
-    constructor(address initialOwner) Ownable(initialOwner) {}
+    // solhint-disable-next-line private-vars-leading-underscore
+    string internal constant UPGRADE_INTERFACE_VERSION = "5.0.0";
```
frangio, line 23: Needs testing.

contracts/proxy/ERC1967/ERC1967Utils.sol
```diff
@@ -70,24 +75,18 @@ library ERC1967Utils {
     }
 
     /**
-     * @dev Perform implementation upgrade
+     * @dev Performs implementation upgrade with additional setup call if data is nonempty.
```
frangio, line 78: Explain checkNonPayable.

test/proxy/transparent/TransparentUpgradeableProxy.behaviour.js
ernestognw: Yes, it was very strange when I first found it. I decided to ignore it focusing on make the tests work, thanks for pushing a fix, it also looks good to me!

contracts/proxy/transparent/ProxyAdmin.sol
```diff
@@ -12,28 +12,29 @@ import {Ownable} from "../../access/Ownable.sol";
...
-     * @dev Sets the initial owner who can perform upgrades.
+     * @dev The version of the upgrade interface of the contract. If this getter is missing, both `upgrade(address)`
+     * and `upgradeAndCall(address,bytes)` are present, and `upgradeTo` must be used if no function should be called,
+     * while `upgradeAndCall` will invoke the `receive` function if the second argument is the empty byte string.
+     * If the getter returns `"5.0.0"`, only `upgradeAndCall(address,bytes)` is present, and the second argument must
+     * be the empty byte string if no function should be called, being impossible to invoke the `receive` function
+     * during an upgrade.
      */
-    constructor(address initialOwner) Ownable(initialOwner) {}
+    // solhint-disable-next-line private-vars-leading-underscore
```
Amxx, line 22: value is public, so we shouldn't need this
```suggestion
```

contracts/proxy/utils/UUPSUpgradeable.sol
```diff
@@ -20,6 +20,17 @@ abstract contract UUPSUpgradeable is IERC1822Proxiable {
...
 
+    /**
+     * @dev The version of the upgrade interface of the contract. If this getter is missing, both `upgradeTo(address)`
+     * and `upgradeToAndCall(address,bytes)` are present, and `upgradeTo` must be used if no function should be called,
+     * while `upgradeToAndCall` will invoke the `receive` function if the second argument is the empty byte string.
+     * If the getter returns `"5.0.0"`, only `upgradeToAndCall(address,bytes)` is present, and the second argument must
+     * be the empty byte string if no function should be called, being impossible to invoke the `receive` function
+     * during an upgrade.
+     */
+    // solhint-disable-next-line private-vars-leading-underscore
```
Amxx, line 31:
```suggestion
```

Amxx (dismissed):
minor comment about unnecessary natspec comments. Otherwize LGTM.

frangio:
Agree with the above, otherwise looks good!

ernestognw:
Nice catches @Amxx, it's ready for review now

Amxx (dismissed):

ernestognw:
Ready for review now, please check the new ERC1967Utils tests before re approving.

test/proxy/ERC1967/ERC1967Utils.test.js
```diff
@@ -0,0 +1,186 @@
...
+        );
+      });
+
+      it("reverts when beacon's implementation does not contain code", async function () {
+        const newBeacon = await UpgradeableBeacon.new(this.v2.address, admin);
+
+        // Corrupt beacon's implementation
+        // Note that the beacon itself is not ERC1967 compliant, so implementation is at slot 1
+        // after the owner in slot 0.
+        await setSlot(newBeacon, 1, anotherAccount);
```
frangio, line 149: Let's add `expect(await newBeacon().implementation()).to.equal(zero)` here to validate the assumption that the slot is correct.

test/proxy/ERC1967/ERC1967Utils.test.js
```diff
@@ -0,0 +1,186 @@
...
+  BeaconSlot,
+} = require('../../helpers/erc1967');
+
+const { ZERO_ADDRESS } = constants;
+
+const ERC1967Utils = artifacts.require('$ERC1967Utils');
+
+const V1 = artifacts.require('DummyImplementation');
+const V2 = artifacts.require('DummyImplementationV2');
+const UpgradeableBeacon = artifacts.require('UpgradeableBeacon');
```
frangio, line 18: This is causing the upgradeable tests to fail because we exclude most proxy contracts: 

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/84db204a4155b6a058a6e51ea2e053c4d29d29c8/scripts/upgradeable/transpile.sh#L29

If we want to do this we would need a new mock contract. Makes sense though given that what we want is to set it to zero and this contract doesn't allow that.

test/proxy/ERC1967/ERC1967Utils.test.js
ernestognw: I think you meant something like this:

```js
expect(await newBeacon.implementation()).to.equal(await getAddressInSlot(newBeacon.address, 1));
```

Isn't it?

frangio: Closing this, see my other comment.

frangio: Yeah that's what I meant though.

ernestognw: Even this is solved I would consider adding it

frangio: It shouldn't be necessary if we use a purpose-built mock as in my other comment but I'll let you make the decision.

test/proxy/ERC1967/ERC1967Utils.test.js
ernestognw: Added, thanks!

test/proxy/ERC1967/ERC1967Utils.test.js
```diff
@@ -0,0 +1,181 @@
...
+          const valueSlot = 0;
+          const newVal = web3.utils.hexToNumber('0xdeabeef');
+
+          // Initialize by calling migrate and modifying the slot 0 `value`
+          const initializeData = this.v2.contract.methods.migrate(newVal).encodeABI();
+          await this.utils.$upgradeToAndCall(this.v2.address, initializeData);
+
+          // Because of the delegatecall to `migrate`, the ERC1967Utils will hold the new value at slot 0
+          const slotValue = await getSlot(this.utils.address, valueSlot);
+          expect(web3.utils.hexToNumber(slotValue)).to.equal(newVal);
```
frangio, line 80: This test and the other similar one are failing in the upgradeable checks, because the transpiler adds `Initializable` with its own storage and this pushes the storage layout. We shouldn't make assumptions about storage positions in tests.

It might be better to test this by with an initialization function that emits an event.

frangio (approved):
Nice thanks all!

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
```diff
@@ -70,73 +68,40 @@ contract TransparentUpgradeableProxy is ERC1967Proxy {
...
-    /**
-     * @dev Initializes an upgradeable proxy managed by `_admin`, backed by the implementation at `_logic`, and
-     * optionally initialized with `_data` as explained in {ERC1967Proxy-constructor}.
+     * @dev Initializes an upgradeable proxy managed by an instance of a {ProxyAdmin} with an `initialOwner`,
+     * backed by the implementation at `_logic`, and optionally initialized with `_data` as explained in
+     * {ERC1967Proxy-constructor}.
      */
-    constructor(address _logic, address admin_, bytes memory _data) payable ERC1967Proxy(_logic, _data) {
-        _admin = admin_;
+    constructor(address _logic, address initialOwner, bytes memory _data) payable ERC1967Proxy(_logic, _data) {
```
sakulstra, line 75: Hey i just stumbled over this change when reviewing v5 and was wondering if it wouldn't make sense to have an overload with `ProxyAdmin _admin` instead of `address initialOwner`.

This way one could easily reuse a single ProxyAdmin for multiple contracts.

Amxx: > This way one could easily reuse a single ProxyAdmin for multiple contracts.

This used ti be the case in v4. We decided to change that.

In v4, the admin was stored in storage, which means that any call going trough the proxy would have to load that value, costing 2100 gas.

```solidity
function _fallback() internal virtual override {
        if (msg.sender == _admin) {
            ...
```

We wanted to reduce that gas cost, and so we decided to make the admin immutable. That means that it is not possible for a proxy to change which admin it uses. You could change the value stored in the ERC-1967 slot, but that would have no effect.

That means that if multiple proxy use the same ProxyAdmin, then you can separate the ownership. You may be able to change who owns the ProxyAdmin, but every proxy that point to it will share the same owner, forever.

That is why we decided to make the `TransparentUpgradeableProxy <> ProxyAdmin` a 1-to-1 relationship.

ernestognw: Agree with @Amxx.

We agreed that it's better to reduce the SLOAD cost in the proxy than allowing to reuse ProxyAdmins. The tradeoff is fine because this not only increases deployment costs but reduces every proxy interaction cost. Consider regular proxy interactions are more frequent (potentially millions of times) than admin interactions.

danhper: I just stumbled upon this while deploying a proxy using v5. Being used to the v4 behavior, I wrongly passed in a proxy admin making the deployment unusable. As far as I could tell, this change is not documented in any changelog. I think that it would be good to add this to the v5 changelog.

ernestognw: Thanks for sharing @danhper! The change is documented on both [OZ upgrades plugins](https://github.com/OpenZeppelin/openzeppelin-upgrades/releases/tag/%40openzeppelin%2Fhardhat-upgrades%403.0.0) and [OZ Contracts](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/CHANGELOG.md#proxy) (see Proxy section). Though I know it's counterintuitive, we strongly advice users to be careful about [major releases](https://docs.openzeppelin.com/contracts/5.x/backwards-compatibility#major_releases).

We'd be happy to hear suggestions on how to improve the way we communicate these changes.

PR #4386, opened 2023-06-23
Fix visibility in ERC721._checkOnERC721Received documentation

Hicham010:
This pull request converts the `_checkOnERC721Received` from a private function to an override-able internal function. This is according to the documentation on the function.

- [ ] Tests
- [ ] Documentation


frangio:
Thanks @Hicham010. I think it's the documentation that is wrong here. Are you suggesting this change because you noticed the discrepancy or because you have a use case that requires overriding this function?

Hicham010:
> Thanks @Hicham010. I think it's the documentation that is wrong here. Are you suggesting this change because you noticed the discrepancy or because you have a use case that requires overriding this function?

I noticed the discrepancy while reading the contract. I currently have no use case.

ernestognw:
I agree the documentation is what's not appropriate here. I don't see why someone would change this function and doing so would make the contract non compliant so I'd keep it private unless there's a concrete use case.

Can we instead update the docs removing the "Internal" mention @Hicham010?

Hicham010:
> I agree the documentation is what's not appropriate here. I don't see why someone would change this function and doing so would make the contract non compliant so I'd keep it private unless there's a concrete use case.
> 
> Can we instead update the docs removing the "Internal" mention @Hicham010?

Updated the docs.

frangio (approved):

Amxx (approved):

PR #4396, opened 2023-06-28
Add `ERC2771Forwarder` fuzz tests for avoiding loss of unused ETH

ernestognw:
Complementary to #4346

#### PR Checklist

- [x] Tests
- [ ] Documentation


ernestognw:
Marking as draft until https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4346 is merged

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -0,0 +1,163 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.19;
+
+import "forge-std/Test.sol";
+import "contracts/metatx/ERC2771Forwarder.sol";
+import "contracts/mocks/CallReceiverMock.sol";
```
Amxx, line 7: I guess we want to use explicit imports from now.

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -0,0 +1,163 @@
...
+    uint256 gas;
+    uint256 nonce;
+    uint48 deadline;
+    bytes data;
+}
+
+contract ERC2771ForwarderMock is ERC2771Forwarder {
+    constructor(string memory name) ERC2771Forwarder(name) {}
+
+    bytes32 private constant _FORWARD_REQUEST_TYPEHASH =
```
Amxx, line 22: This is already declared in the `ERC2771Forwarder`.

I wonder if constant and immutable variable should be internal instead of private. There is no risk of writting to it.

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -0,0 +1,163 @@
...
+        vm.deal(address(_erc2771Forwarder), initialBalance);
+        uint256 nonce = _erc2771Forwarder.nonces(_signer);
+
+        ERC2771Forwarder.ForwardRequestData[] memory batchRequestDatas = new ERC2771Forwarder.ForwardRequestData[](
+            batchSize
+        );
+
+        uint256 expectedRefund;
+
+        for (uint i; i < batchSize; ++i) {
```
Amxx, line 135: I now its not strictly necessary, but I find it clearer to explicitly initialise.

```suggestion
        for (uint256 i = 0; i < batchSize; ++i) {
```

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -0,0 +1,163 @@
...
+            bytes memory data;
+            bool succeed = i % 2 == 0; // Half of the requests will succeed
```
Amxx, line 137: Here almost exactly half of the request will fail, and all the request will have the same value. That limits the fuzzing quite a lot. IMO we should probably use a pseudo random sequence for succeed. For example, using value as a seed:
```suggestion
            bool succeed = uint256(keccak256(abi.encodePacked(value, i))) % 2 == 0;
```

Amxx (commented):
Left some minor comments, otherwise looks good.

test/metatx/ERC2771Forwarder.t.sol
ernestognw: I'd be in favor of making the `_FORWARD_REQUEST_TYPEHASH` internal, but I'm also in favor of having an internal `_structHash`, what do you think about both? 

CC @frangio because it's a design decision.

frangio: Internal seems fine.

test/metatx/ERC2771Forwarder.t.sol
ernestognw: Good idea, I wasn't feeling very comfortable with half of them.

ERC2771Forwarder.t.js
Amxx, line 1: I'm not really getting this file. If you want to procedurally generate solidity, I would argue it should be node using the existing framework. That is adding a "template" in `scripts/generate/templates/.` and listing it in `scripts/generate/run.js`. The value of reusing the same workflow it that it will (among other thing) check the consistency between the template file and the indexed solidity version is CLI.

Also, In this case I'm not even sure that deserves procedural generation. We usually do it to avoid code duplication when we want the same structure to be declared multiple times with multiple types. Here its about the maxSize, and it feels like you only instantiate it once.

Did you git add this file by mistake ?

ernestognw: Yes, it was a mistake after messing up with my local stash. I agree it should be in the templates folder and consumed by the `generate/run.js` script. I was using this for benchmarking the memory leak in the batch execution. Just removed

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -0,0 +1,158 @@
...
+            nonce: nonce,
+            deadline: deadline,
+            data: data
+        });
+
+        bytes32 digest = _erc2771Forwarder.structHash(request);
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(_signerPrivateKey, digest);
+        bytes memory signature = abi.encodePacked(r, s, v);
+
+        _requestData = ERC2771Forwarder.ForwardRequestData({
```
Amxx, line 84: Any reason to favor this syntax over 
```solidity
return ERC2771Forwarder.ForwardRequestData({ ... });
```
?

I'm always thinking "ok that is an identifier starting with an underscore, so it must be some private storage variable used for state" ... and its only when I see its not declared in the contract that I check for named return variable.

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -0,0 +1,158 @@
...
+            _forgeRequestData({
+                value: value,
+                nonce: nonce,
+                deadline: uint48(block.timestamp + 1),
+                data: abi.encodeCall(CallReceiverMock.mockFunction, ())
+            })
+        );
+
+        assertEq(address(_erc2771Forwarder).balance, initialBalance);
+    }
```
Amxx, line 114: This test is fuzzing the initialBalance and the value, but not the "paths" in the `execute` function. It only runs a function that we know is ok, with a deadline we know is ok, and a nonce we know is correct, with a fixed amount of gas

Ideally, we would want to consider many data, to check that whatever the outcome of the subcall, balances are preserved.

Maybe you could have a uint8 argument that is used to select one of "many" function. `mockFunction` will only be one of them. It should include: non payable functions, function that revert, function that panic, ...

Amxx: or just do 50/50 between `mockFunction` and `mockFunctionRevertsNoReason`

test/metatx/ERC2771Forwarder.t.sol
ernestognw: tbh I just didn't think about it, but you're right, we've historically preferred the explicit return. I updated with an explicit return so I'm solving.

test/metatx/ERC2771Forwarder.t.sol
ernestognw: You're right, ideally, we would fuzz multiple functions. However, it becomes difficult to conditionally expect a revert in the test, so I did the 50/50 between `mockFunction` and `mockFunctionRevertNoReason` you proposed to cover both paths.

The `deadline`, `nonce` and `gas` are fixed because the contract is about making sure there's no value left in the contract and I don't think there's a way the `deadline` or `nonce` might affect that. However, `gas` may affect it because of the `_checkForwardedGas` but it's already proven in the comments.

I think the suggestion you proposed for this test is enough for its purpose (not leaving ETH stuck). I also tried to fuzz the `gas` very early in #4346 but it becomes messier because we have to conditionally expect reverts as well, and it's difficult to predict. Would appreciate if you have any idea how it may work in case you think it's a valuable addition

Amxx (approved):

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        _erc2771Forwarder.execute{value: value}(requestData);
+        assertEq(address(_erc2771Forwarder).balance, initialBalance);
+    }
+
+    function testExecuteBatchAvoidsETHStuck(uint256 initialBalance, uint256 batchSize, uint256 value) public {
+        batchSize = bound(batchSize, 1, 10);
+        vm.assume(initialBalance < _MAX_ETHER);
+        vm.assume(value < _MAX_ETHER);
```
frangio, line 126: Is there a reason for not using bound for `initialBalance` and `value`? I'd think we should use bound in all cases.

frangio (commented):
Looks good but I have one comment.

test/metatx/ERC2771Forwarder.t.sol
ernestognw: I used assume because the bound is between 0 and `_MAX_ETHER`, so I think it has no difference with bound. Is there a significant difference? No problem changing it, tho.

frangio: From the [docs](https://book.getfoundry.sh/cheatcodes/assume):

>The assume cheatcode should mainly be used for very narrow checks. Broad checks will slow down tests as it will take a while to find valid values, and the test may fail if you hit the max number of rejects.

I think this qualifies as a "broad check".  (Narrow checks would be `vm.assume(a != 0)`.)

The other reason to use `bound` over `vm.assume` is to properly hit edge cases, but that doesn't apply here.

Amxx: My understanding is that `assume` removes the invalid branchs from the tree that is being explored, where `bound` doesn't need to. 

Basically if you `assume(x < 10)` and the system tries with `x = 11`, the test will be discarded and you will get no property check. On the other if you `bound` x to [0, 10], and `x = 11` then the bound will use modulo to get back in the scope, and you will get properties check from that one run.

So for arithmetic intervals, you get more out of your fuzzing when doing bound.

frangio (approved):

PR #4398, opened 2023-06-28
Implement recommendations from 5.0 audit Phase 1A

### LINKED ISSUE #4344: Resolve discussion about the overflow in ERC1155Supply.
See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4261/files#r1224323887

ernestognw:
Full audit report will be out later.

Fixes #4344 
Fixes LIB-917
Fixes LIB-931


ernestognw:
As a result of bd59f1c9dd426d38f808419683dce9b94e5da5cd, I decided to rename the `DelegateVotesChanged` arguments to voting units instead because it's more accurate. However, it's open to discussion and I will also add a changeset if y'all agree @frangio @Amxx

ernestognw:
Slither is failing but the fix will get into master with https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4346

contracts/mocks/ERC20VotesQuadratic.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.19;
+
+import "../token/ERC20/extensions/ERC20Votes.sol";
```
Amxx, line 5:
```suggestion
import {ERC20Votes} "../token/ERC20/extensions/ERC20Votes.sol";
```

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -408,20 +448,28 @@ abstract contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI, IER
...
             mstore(array1, 1)
+            // Store the single element at the next word after the length (where content starts)
             mstore(add(array1, 0x20), element1)
 
+            // Repeat for next array locating it right after the first array
             array2 := add(array1, 0x40)
             mstore(array2, 1)
             mstore(add(array2, 0x20), element2)
 
+            // Clear the free memory pointer by pointing after the second array
```
Amxx, line 472:
```suggestion
            // Update the free memory pointer by pointing after the second array
```

contracts/token/ERC20/extensions/ERC20FlashMint.sol
```diff
@@ -38,6 +42,10 @@ abstract contract ERC20FlashMint is ERC20, IERC3156FlashLender {
      * @dev Returns the maximum amount of tokens available for loan.
      * @param token The address of the token that is requested.
      * @return The amount of token that can be loaned.
+     *
+     * NOTE: This function does not consider any form of supply cap, so in case
```
Amxx, line 46: I'm not sure this is correct. This function uses `_mint` to create the tokens. `_mint` will call `update` with "from = 0". If present, the ERC20Capped module will check that the cap is not exceeded.

contracts/token/ERC20/extensions/ERC20FlashMint.sol
```diff
@@ -14,6 +14,10 @@ import {ERC20} from "../ERC20.sol";
  * Adds the {flashLoan} method, which provides flash loan support at the token
  * level. By default there is no fee, but this can be changed by overriding {flashFee}.
  *
+ * NOTE: When this extension is used along with the {ERC20Capped} extension, its cap
+ * won't be checked in {flashLoan}. Although it can be limited by also overriding {maxFlashLoan}
+ * to also cap the maximum amount of tokens that can be flash-minted.
+ *
```
Amxx, line 20: I'm not sure about the formulation here. 

In `flashLoan`, the mint operation has to follow the rules of `_update`, which is what `ERC20Capped` overrides. So the cap fill be respected when minting the "flash" tokens. It is correct to say that `maxFlashLoan` doesn't reflect that by default, but its IMO incorrect to say that "its cap won't be checked in {flashLoan}"

frangio:
We're not addressing the inconsistency of "value" vs "amount" in `IERC1155Receiver`. I've also mentioned it is similar to https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3864 and that we should probably address both.

contracts/token/ERC20/extensions/ERC20Wrapper.sol
```diff
@@ -79,4 +85,15 @@ abstract contract ERC20Wrapper is ERC20 {
...
     }
+
+    /**
+     * @dev Determines whether an account can receive either wrapped or underlying tokens by operating
+     * in this contract.
+     *
+     * By default, it disallows this contract to receive tokens so they don't get stuck, however,
+     * another account can still send tokens to it.
+     */
+    function _canReceiveToken(address account) internal virtual returns (bool) {
```
Amxx, line 96: To we have any plan for people to override this function ? IMO it would be simpler/cleaner to do
```solidity
if (account == address(this)) {
  revert ERC20InvalidReceiver(account);
}
```
in both functions and not add this extra piece of code.

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -590,4 +592,44 @@ contract('ERC20Votes', function (accounts) {
...
+        previousUnits: votingUnits.toString(),
+        newUnits: '0',
+      });
+      expectEvent(receipt, 'DelegateVotesChanged', {
+        delegate: charlie,
+        previousUnits: '0',
+        newUnits: votingUnits.toString(),
+      });
+    });
+  });
```
Amxx, line 634: This will not work if the transferred amount is not the entire user balance. IMO we should remove it, and add warning so that user don't try to do exactly that.

contracts/token/ERC20/ERC20.sol
```diff
@@ -198,6 +198,9 @@ abstract contract ERC20 is Context, IERC20, IERC20Metadata, IERC20Errors {
      * - `spender` cannot be the zero address.
      * - `spender` must have allowance for the caller of at least
      * `requestedDecrease`.
+     *
+     * NOTE: Although this function is designed to avoid double spending with {approval},
+     * it can still be frontruned, preventing any attempt of allowance reduction.
```
frangio, line 203:
```suggestion
     * it can still be frontrunned, preventing any attempt of allowance reduction.
```

test/token/ERC20/extensions/ERC20Votes.test.js
frangio: What are you proposing to remove? The tests, the ERC20VotesQuadratic mock, or the _toVotingUnits function?

Amxx: At the very least the quadratic mock and tests.

Amxx: Since the only think we support are linear functions. We could have a "rate" that would be 1:1 by default and that we could increase (for example 5 voting unit per token). 

If we do that we would have to be carefull of rounding errors. Rate should probably be an integer. It would have limited value for ERC20 but maybe it would be nice for some erc721 ?

frangio: >we would have to be carefull of rounding errors

Yeah this is another thing implementors of `_toVotingUnits` functions would have to keep in mind... Quite error prone. I don't think we should add the rate mechanism either, it's unrelated to the comments in the audit.

For ERC-721 we've talked about this but we were concerned that if the rate is a function people might implement it in a way that isn't constant and would cause errors. I guess if it's a pure function with no arguments it would need to be constant, but not if the contract is upgradeable.

contracts/governance/utils/IVotes.sol
```diff
@@ -19,9 +19,9 @@ interface IVotes {
     event DelegateChanged(address indexed delegator, address indexed fromDelegate, address indexed toDelegate);
 
     /**
-     * @dev Emitted when a token transfer or delegate change results in changes to a delegate's number of votes.
+     * @dev Emitted when a token transfer or delegate change results in changes to a delegate's number of voting units.
      */
-    event DelegateVotesChanged(address indexed delegate, uint256 previousBalance, uint256 newBalance);
+    event DelegateVotesChanged(address indexed delegate, uint256 previousUnits, uint256 newUnits);
```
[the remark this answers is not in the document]
Amxx: Note we'll need to update https://github.com/OpenZeppelin/openzeppelin-subgraphs at some point.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -183,12 +188,35 @@ abstract contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI, IER
...
+    /**
+     * @dev Version of {_update} that performs the token acceptance check by calling {IERC1155Receiver-onERC1155Received}
+     * or {IERC1155Receiver-onERC1155BatchReceived} on the receiver address if it contains code (eg. is a smart contract
+     * at the moment of execution).
+     *
+     * IMPORTANT: Overriding this function is discouraged because it poses a reentrancy risk from the receiver. So any
+     * update to the contract state after this function would break the check-effect-interaction pattern. Consider
+     * overriding {_update} instead.
+     */
+    function _updateChecked(
```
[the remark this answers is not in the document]
Amxx: `_updateWithAcceptanceCheck` ?

test/token/ERC20/extensions/ERC20Votes.test.js
Amxx: >It might be better to remove _toVotingUnits altogether and improve the documentation to make sure users use caution.

👍🏻

contracts/token/ERC20/extensions/ERC20FlashMint.sol
```diff
@@ -14,6 +14,10 @@ import {ERC20} from "../ERC20.sol";
  * Adds the {flashLoan} method, which provides flash loan support at the token
  * level. By default there is no fee, but this can be changed by overriding {flashFee}.
  *
+ * NOTE: When this extension is used along with the {ERC20Capped} extension, its cap
+ * won't be checked in {flashLoan}. Although it can be limited by also overriding {maxFlashLoan}
+ * to also cap the maximum amount of tokens that can be flash-minted.
```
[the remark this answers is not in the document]
ernestognw: Suggestion added, just changed `refect` -> `reflect`

contracts/governance/utils/IVotes.sol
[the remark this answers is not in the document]
ernestognw: Make sense to me, just added!
And yes, at this points the subgraphs are somewhat outdated

contracts/token/ERC20/extensions/ERC20FlashMint.sol
ernestognw: I went with @frangio's [suggestion](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4398#discussion_r1247963082), I think this is clearer and removes the "its cap won't be checked in {flashLoan}"

contracts/token/ERC1155/ERC1155.sol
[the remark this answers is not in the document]
ernestognw: I find `_updateChecked` better than `_updateWithAcceptanceCheck`, but I'd suggest `_acceptedUpdate`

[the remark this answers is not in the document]
frangio: Out of the two-word options I prefer `_updateChecked`. 

`_updateWithAcceptanceCheck` is good too but I agree shorter is better.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -146,13 +144,19 @@ abstract contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI, IER
...
+     * IMPORTANT: This function relies on the invariant of `amount <= fromBalance <= totalSupply` and also in the
+     * Solidity built-in arithmetic checks. Make sure to not break these checks when accessing this function.
      */
     function _update(
         address from,
         address to,
         uint256[] memory ids,
         uint256[] memory amounts,
-        bytes memory data
+        bytes memory /* data */
```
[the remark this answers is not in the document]
ernestognw: Removed!

contracts/token/ERC20/ERC20.sol
```diff
@@ -237,6 +240,10 @@ abstract contract ERC20 is Context, IERC20, IERC20Metadata, IERC20Errors {
      * the zero address. All customizations to transfers, mints, and burns should be done by overriding this function.
      *
      * Emits a {Transfer} event.
+     *
+     * IMPORTANT: This function relies on the invariant of
+     * `amount <= fromBalance <= totalSupply && toBalance + amount <= totalSupply` and also in the
+     * Solidity built-in arithmetic checks. Make sure to not break these checks when accessing this function.
      */
     function _update(address from, address to, uint256 amount) internal virtual {
         if (from == address(0)) {
```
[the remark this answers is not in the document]
ernestognw: We agreed to remove the note from the docstrings and instead keep this comment. The rationale is that the invariants are guaranteed by the contract anyway and the comment should be more targeted for future maintainers

test/token/ERC20/extensions/ERC20Votes.test.js
ernestognw: I didn't realize this doesn't fully work because of partial transfers but the example from @Amxx was clear enough. I agree with removing the whole tests for quadratic voting and I'll add a note in that regard.

contracts/token/ERC1155/ERC1155.sol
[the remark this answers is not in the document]
Amxx: I'd value clarity over shortness (it's not that long when compared to "_doBatchTransferAcceptanceCheck"). But either is fine by me.

If you both prefer "_updateChecked" let's just do that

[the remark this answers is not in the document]
Amxx: Is there a risk that "Checked" is interpreted by some devs as the opposite of "unchecked {}" ?

contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -72,4 +72,11 @@ abstract contract ERC20Votes is ERC20, Votes {
     function checkpoints(address account, uint32 pos) public view virtual returns (Checkpoints.Checkpoint224 memory) {
         return _checkpoints(account, pos);
     }
+
+    /**
+     * @dev Converts an `amount` of tokens into voting units.
+     */
+    function _toVotingUnits(uint256 amount) internal view virtual returns (uint256) {
```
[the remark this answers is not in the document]
ernestognw: Solving since we're removing `_toVotingUnits`

test/token/ERC20/extensions/ERC20Votes.test.js
ernestognw: Note added and changes rolled back

contracts/token/ERC1155/ERC1155.sol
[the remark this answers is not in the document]
ernestognw: Fair point! I think there's a reasonable risk and I'd favor clarity in that case. Still, `_updateWithAcceptanceCheck` doesn't fully convince me, what about `_updateAcceptanceChecked`?

contracts/token/ERC20/extensions/ERC20Wrapper.sol
ernestognw: The initial suggestion from Fran was to use a more general `_canReceiveToken` for every ERC20 but we haven't yet reached consensus with that so I'm implementing this suggestion instead.

Solving

contracts/mocks/ERC20VotesQuadratic.sol
```diff
@@ -0,0 +1,22 @@
...
+        uint256 balance = balanceOf(account);
+        return balance * balance;
+    }
+
+    /**
+     * @dev Converts an `amount` of tokens into voting units.
+     */
+    function _toVotingUnits(uint256 amount) internal view virtual override returns (uint256) {
+        return amount * amount;
+    }
```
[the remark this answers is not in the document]
ernestognw: Thanks for the examples, it's clear it won't work. I removed the `_toVotingUnits` and enhanced the documentation of `_getVotingUnits` to warn people about non linear voting behavior.

contracts/token/ERC1155/ERC1155.sol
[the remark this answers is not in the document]
ernestognw: Ok after reconsidering I think `_updateWithAcceptanceCheck` didn't convince me because there's a connector. But it's the same case for `_upgradeAndCall`, so let's go with `_updateWithAcceptanceCheck`.

ernestognw:
> We're not addressing the inconsistency of "value" vs "amount" in `IERC1155Receiver`. I've also mentioned it is similar to #3864 and that we should probably address both.

Addressed by replacing `amount` with `value` in both ERC20 and ERC1155 including extensions, although I kept the word "amount" where I considered appropriated

frangio (dismissed):

contracts/token/ERC1155/extensions/ERC1155Pausable.sol
```diff
@@ -17,7 +17,7 @@ import {Pausable} from "../../../security/Pausable.sol";
  * addition to inheriting this contract, you must define both functions, invoking the
  * {Pausable-_pause} and {Pausable-_unpause} internal functions, with appropriate
  * access control, e.g. using {AccessControl} or {Ownable}. Not doing so will
- * make the contract unpausable.
+ * make the contract not to have pausable capabilities at all.
```
Amxx, line 20: Minor suggestion:

```suggestion
 * make the pause mechanism of the contract unreachable, and thus unusable.
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -48,10 +48,10 @@ contract('ERC20Votes', function (accounts) {
       });
 
       it('minting restriction', async function () {
-        const amount = new BN('2').pow(new BN('224'));
-        await expectRevertCustomError(this.token.$_mint(holder, amount), 'ERC20ExceededSafeSupply', [
-          amount,
-          amount.subn(1),
+        const value = new BN('2').pow(new BN('224'));
```
Amxx, line 51: minor suggestion 

```suggestion
        const value = web3.utils.toBN(1).shln(224);
```

Amxx (approved):
LGTM. I made 2 very small suggestions.

contracts/token/ERC1155/extensions/ERC1155Pausable.sol
ernestognw: I also realized this changed had to be made on `ERC20Pausable` and `ERC721Pausable` so I implemented the same suggestion there.

frangio (approved):

PR #4399, opened 2023-06-28
Use explicit imports

frangio:
Note that because of Solidity override resolution (`override(A, B)`) sometimes it's necessary to import an identifier even if it's only used for that.

When both an interface and implementation are required, I've reused the same file for the import statement, as in:

```solidity
import {IGovernor, Governor} from "../../governance/Governor.sol";
```


frangio:
Do we want a changeset for this?

ernestognw:
> Do we want a changeset for this?

I think it may unexpectedly remove imported values that users might rely on. I'd say yes just for making sure it's mentioned somewhere.

contracts/token/ERC1155/ERC1155.sol
```diff
@@ -3,13 +3,13 @@
...
-import "./extensions/IERC1155MetadataURI.sol";
-import "../../utils/Context.sol";
-import "../../utils/introspection/ERC165.sol";
-import "../../utils/Arrays.sol";
-import "../../interfaces/draft-IERC6093.sol";
+import {IERC1155} from "./IERC1155.sol";
+import {IERC1155Receiver} from "./IERC1155Receiver.sol";
+import {IERC1155MetadataURI} from "./extensions/IERC1155MetadataURI.sol";
+import {Context} from "../../utils/Context.sol";
+import {IERC165, ERC165} from "../../utils/introspection/ERC165.sol";
```
Amxx, line 10: In these case, do we want to duplicate the import to fetch from the original file ?

```suggestion
import {IERC165} from "../../utils/introspection/IERC165.sol";
import {ERC165} from "../../utils/introspection/ERC165.sol";
```

Amxx (approved):

contracts/token/ERC1155/ERC1155.sol
frangio: I don't have a strong opinion, initially I thought there would be a lot more of these cases and that it would be better to have fewer import statements. But I think it's just a few places and we could do either.

ernestognw (approved):
LGTM, thank you for tackling this @frangio!

I just updated the docs references to also use explicit imports and added a linter rule to avoid global imports

ernestognw (approved):

docs/modules/ROOT/pages/governance.adoc
```diff
@@ -200,11 +200,12 @@ The Governor will automatically detect the clock mode used by the token and adap
 // SPDX-License-Identifier: MIT
```
frangio, line 200: Note: We have to move this example code to a separate file like the others in this article.

Issue #4402, opened 2023-06-29
Improve Integer Square Root


chgorman:
The integer square root function merged by #3242 is not optimal. A more efficient method would be preferred along with a proof showing the algorithm always produces the correct value.

ernestognw:
Hey @chgorman, sorry for the delayed response, we think the proposal you're making requires a bit of study from our side and we're focused on 5.0 at the moment.

We'll definitely come back to check it out after discussing it, probably after 5.0 since this doesn't change the interface and wouldn't be breaking if the function does exactly the same. It'll be a candidate for 5.1.

chgorman:
Feel free to reach out if there are any questions. A [previous document](https://github.com/alicenet/.github/blob/main/docs/efficient_isqrt.pdf) discusses validity proofs. Another [repo](https://github.com/chgorman/isqrt-gas) has a more extensive analysis.

james-toussaint:
Resolved with https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4403.

PR #4403, opened 2023-06-29
Improved integer square root.

chgorman:
A more efficient implementation of integer square root within Solidity; improves upon #3242. Solves Issue #4402.

Proof of validity discussed [here](https://github.com/alicenet/.github/blob/main/docs/efficient_isqrt.pdf).


frangio:
Thanks for this proposal.

Unfortunately our gas benchmarks are only able to measure non-view functions, so we can't see any improvement in the [gas report](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/5417465710#summary-14671376848) for this PR.

Can you share how much of an improvement you are observing? Please share a few inputs that would allow us to confirm that too.

chgorman:
The previously [linked article](https://github.com/alicenet/.github/blob/main/docs/efficient_isqrt.pdf) discusses integer square root computation within Solidity (in addition to giving proofs). Section 5 discusses gas costs; OZ had a mean gas cost of 948, while the new version had a mean gas cost of 767.

I am working on having a public repo where the gas analysis was performed, but it is not ready yet.

contracts/utils/math/Math.sol
```diff
@@ -245,7 +245,10 @@ library Math {
             result = (result + a / result) >> 1;
             result = (result + a / result) >> 1;
             result = (result + a / result) >> 1;
-            return result * result <= a ? result : (result - 1);
+            if (result * result <= x) {
```
Amxx, line 248: `x` is not defined here

Amxx:
The main change of this proposition is that it inlines initial estimation, instead of calling `log2`. That might indeed have some effect on the gas costs, but I would also argue it reduces the readability of the function.

contracts/utils/math/Math.sol
chgorman: Fixed.

contracts/utils/math/Math.sol
```diff
@@ -220,35 +220,29 @@ library Math {
...
+            if (aAux >= (1 << 8  )) { aAux >>= 8;   result <<= 4;  }
+            if (aAux >= (1 << 4  )) { aAux >>= 4;   result <<= 2;  }
+            if (aAux >= (1 << 2  )) {               result <<= 1;  }
+            result += (result >> 1);
             result = (result + a / result) >> 1;
             result = (result + a / result) >> 1;
             result = (result + a / result) >> 1;
             result = (result + a / result) >> 1;
             result = (result + a / result) >> 1;
             result = (result + a / result) >> 1;
```
Amxx, line 241: I only count 7 iterations, yet the PDF you linked shows that in some cases, 8 are required. I'm worried about that part. Do you have an example of value that needs 8 iterations that we could run both version of the code with ?

chgorman: First, **Newton iteration** refers to `result = (result + a/result) >> 1`. This is the Newton iteration step when computing square roots via Newton's method; see [here](https://en.wikipedia.org/wiki/Integer_square_root#Algorithm_using_Newton's_method).

So, `result += (result >> 1);` **is not** a Newton iteration but rather part of the initialization. Yes, you could replace it another Newton iteration, but that is more expensive.

As to your comments about "8 cases are required", you are misunderstanding the comment and figure. If **only** Newton iterations are performed **and the same initialization is used that is currently used by OZ, **then** it sometimes takes 8 Newton iterations to reach the correct value. Note that this **does not apply** either in the current OZ code nor the proposed change. That figure was primarily meant to show **why** a good initialization value is required for efficient computation.

chgorman:
"The "only" math difference that you PR makes is the `result += result >> 1` part." This is what allows the reduction of 7 to 6 Newton iterations. I could not find a similar initialization elsewhere. The other math difference is the return logic; with it, it is possible to prove that the algorithm is correct. For the proof, see [Appendix B](https://github.com/alicenet/.github/blob/main/docs/efficient_isqrt.pdf).

"Where does this line come from?" (referring to `result += result >> 1`): this is so that $result == 2^{e-1} + 2^{e-2}$ if $2^{e-1} \le sqrt(x) < 2^{e}$

"I'm not 100% but I was under the impression that the math.min includes an interation. I'd love to check the effect of it": please provide a proof that "math.min includes an iteration".

" I think result += (result >> 1); is supposed to be an iteration. I'm not sure if that is correct of not. For the sake of readability I'd keep the same syntax as the other ones.": As mentioned above, it is not a Newton iteration, but rather part of initialization.

Amxx:
> "I'm not 100% but I was under the impression that the math.min includes an interation. I'd love to check the effect of it": please provide a proof that "math.min includes an iteration".

The iteration is doign an arithmetic average between `result` and `a / result`. Also we know that at some point, the two values (`result` and `a/result`) will be (`ceil(sqrt(a))` and `floor(sqrt(a))`). So at some point, when the precision is down to the last digit, `result` and `a/result` are two good candidates, and the smallest one is the expected value (the other one is +1).
That is the logic behind the min operation. Now can it replace the last iterration, I'm not so sure, and I have not tested it.

Amxx:
@PaulRBerg You are the "math done in solidity" expert. WDYT of that initialization change ?

contracts/utils/math/Math.sol
```diff
@@ -216,39 +216,51 @@ library Math {
...
-
-        // At this point `result` is an estimation with one bit of precision. We know the true value is a uint128,
-        // since it is the square root of a uint256. Newton's method converges quadratically (precision doubles at
-        // every iteration). We thus need at most 7 iteration to turn our partial result with one bit of precision
-        // into the expected uint128 result.
         unchecked {
+            // Take care of easy edge cases
+            if (a <= 1) { return a; }
+            // This check ensures no overflow
+            if (a >= ((1 << 128) - 1)**2) { return (1 << 128) - 1; }
```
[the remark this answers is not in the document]
Amxx: interresting.

I'd rewrite that as:
```suggestion
            if (a >= type(uint128).max**2) { return type(uint128).max; }
```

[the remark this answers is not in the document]
Amxx: I wonder if the cost of that test outweight the cost of the `min` at the end.

chgorman:
> > "I'm not 100% but I was under the impression that the math.min includes an interation. I'd love to check the effect of it": please provide a proof that "math.min includes an iteration".
> 
> The iteration is doign an arithmetic average between `result` and `a / result`.

I really think it is better to of the iteration not as the average of `result` and `a / result`, but rather what it is: a Newton iteration for computing square roots. This allows for using the error bounds for Newton iterations.

> Also we know that at some point, the two values (`result` and `a/result`) will be (`ceil(sqrt(a))` and `floor(sqrt(a))`). 

Yes, it is true **at some point** that `result == floor(sqrt(a))` (that is, the Integer Square Root) when Newton iterations continue to be performed. But that is not your stopping criterion. Your stopping criterion is: after doing 7 Newton iterations with a particular initialization, return `min(result, a/result)`. What are the possible values for `result` after 7 Newton iterations, and does this ensure that `min(result, a/result)` always, in fact, returns `floor(sqrt(a))` (Integer Square Root)?

This specific return logic (using `min(result, a/result)` is discussed in [Appendix B.5](https://github.com/alicenet/.github/blob/main/docs/efficient_isqrt.pdf). Either point out the logical flaw or provide a stronger analysis as to why those results don't apply. For the record, I have not found an instance where the original OZ sqrt function returns an incorrect value, but that is not a proof.

contracts/utils/math/Math.sol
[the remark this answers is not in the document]
chgorman: > interresting. I guess this is not necessary if you do `min(result, a/result)` so the current version is good in that regard.

The current version has no proof of correctness. This proposed change does. See [Appendix B.4.3](https://github.com/alicenet/.github/blob/main/docs/efficient_isqrt.pdf). Have you looked at that report?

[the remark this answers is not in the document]
chgorman: Code was updated with suggestion.

chgorman:
A repo with additional analysis may be found [here](https://github.com/chgorman/isqrt-gas).

contracts/utils/math/Math.sol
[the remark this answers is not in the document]
chgorman: For reference, the overflow check (`if (a >= uint256(type(uint128).max)**2) { return type(uint128).max; }`) costs 23 gas.

Amxx:

chgorman:
Yes, I understand needing to take the time for gas optimization and not having time now.

With that said, I would like to add that this is not just gas optimization but also replacing an algorithm (which has not been proven correct) with a provably-correct algorithm. If there are questions about the initialization choice or the math, please reach out.

chgorman:
I made a couple of small improvements to the algorithm. Additionally, please see the extended analysis [here](https://github.com/chgorman/isqrt-gas).

ernestognw:
Just caught up with the convo, and I think this is a very interesting proposal.

I see the Unrolled3 version (the one used here) has the better results, and considering the data points are intended to be uniformly distributed in the uint256 space, I consider the optimization worth it.

However, I share the readability concerns with @Amxx, I think it's worth linking to the proof or rather writing it down (succinctly). The challenge imo is that if we come back we need to read the paper again to remember what we were doing. The comments on top of the initialization make sense but I'd try to expand a bit more on the documentation, right now it doesn't mention that's based on Newton's algorithm at all.

It's looking good, but I think it's worth measuring gas again especially because we've changed some things, including the log2 algorithm, so perhaps the gas impact is less now.

chgorman:
> I see the Unrolled3 version (the one used here) has the better results, and considering the data points are intended to be uniformly distributed in the uint256 space, I consider the optimization worth it.

Great. I agree.

> 
> However, I share the readability concerns with @Amxx, I think it's worth linking to the proof or rather writing it down (succinctly). The challenge imo is that if we come back we need to read the paper again to remember what we were doing. The comments on top of the initialization make sense but I'd try to expand a bit more on the documentation,

This is a valid point. I will rework the discussion to be more verbose; then if it is too verbose, we can work on trimming it down. I also think it is important for a proof to be included (more on proofs in another comment).

> right now it doesn't mention that's based on Newton's algorithm at all.

This is true only in the most stringent sense of not having the words "Newton's method" or "Newton's algorithm" in the comments. In [Line 258](https://github.com/chgorman/openzeppelin-contracts/blob/77db52d5cedd9b9038afc0976b9634366ad5eb3f/contracts/utils/math/Math.sol#L258), the comment is `// Perform the 6 required Newton iterations`. The first result returned when searching for "newton iteration" on Google returns the link to [Newton's method Wikipedia page](https://en.wikipedia.org/wiki/Newton%27s_method). But sure, I'll make sure to include a reference that this method is using Newton's method for computing square roots.

> Given this impl is similar to the original `sqrt`, I'd reuse the initialization comments [from there.](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3ac4add548178708f5401c26280b952beb244c1e/contracts/utils/math/Math.sol#L163).

I will try to use those comments as a guide.

> It's looking good, but I think it's worth measuring gas again especially because we've changed some things, including the log2 algorithm, so perhaps the gas impact is less now.

I would be willing to bet the improvements don't cause a significant efficiency increase; in particular, I think the changes will save at most 150 gas per call. But sure, I will include the updated logic as another algorithm for [my analysis](https://github.com/chgorman/isqrt-gas). Doing this will take a bit of time. Using that assumed gas savings of 150 (which effectively amounts to a mean gas cost of 800 on the tests), the Unrolled3 gas mean of 742 still provides a 9% savings.

TL;DR: I will work on being more verbose in my comments to provide better documentation (including adding a proof).

chgorman:
This comment is devoted to **proofs**. Tagging @Amxx , @PaulRBerg , and @ernestognw because they have either commented on this PR or have been referenced. Please feel free to have anyone else join this conversation as needed.

> I think it's worth linking to the proof or rather writing it down (succinctly).

As noted, [my analysis](https://github.com/chgorman/isqrt-gas) includes a proof of validity that the proposed method results a valid result. Does anyone dispute this assertion? That is, does anyone say my proof is invalid? If not, great; otherwise, where is the invalid logic?

I now ask the same **of you**: where is **your proof** that the current method (that is, the current method used by OZ) is valid? In particular, where is **your proof** that the return logic ensures that the correct value is returned?

I have no qualms about the current method in general; the logic is sound until the last step (the return logic). The current algorithm is this:

1. Initialize `result` to $2^{e}$, where $2^{e}\le \sqrt{a} < 2^{e+1}$.
2. Perform 7 Newton iterations, overwriting `result` each time.
3. Return `min(result, a/result)`.

Given your choice of initialization, after performing 7 Newton iterations, it is true that either `result = q` or `result = q+1`, where $q = \lfloor\sqrt{a}\rfloor$ (the true value of the integer square root of $a$).

As I noted in [my report](https://github.com/chgorman/isqrt-gas/blob/d42e99984e15338173996a935ded987a9bc83cb2/report/isqrt_analysis.pdf) (Appendix D.3 Final Check in this version), if we have $a = q^{2} + \ell$ for $\ell\in[0, 1, \cdots, q-1]$, then we also have $a = (q-1)(q+1) + (1 + \ell)$. Because $1+\ell \le q$, $\lfloor a/(q+1) \rfloor = q-1$.

So, we suppose that $a = q^{2} + \ell$ for $\ell\in[0, 1, \cdots, q-1]$ and that `result = q+1`. Then the current return logic will be `return min(q+1, q/(q+1)) = return min(q+1, q-1) = return q-1`. This means that the current algorithm will return an incorrect result.

So, my question is this: if my above analysis is incorrect, where **precisely** is my logic incorrect? Otherwise, where is **your proof** that this cannot happen? The specific form of $a$ above means that there are approximately $2^{255}$ values where this could occur (my count being $2^{255} - 2^{128} - 2^{127} - 2$); I'm not saying **this does happen**, but rather that **it could happen**. So, either provide a better analysis over generic Newton iteration error bounds, or let me know when you are finished checking the potential $2^{255}$ possible counterexamples.

**TL; DR**: With you telling me

> I think it's worth linking to the proof or rather writing it down (succinctly).

I ask you the same: where is **your proof** that the current OZ `sqrt` algorithm is correct? I want to see it.

chgorman:
I've updated the PR to include increased comments and added the associated proof. It is possibly too verbose at this point, but let me know your thoughts.

ernestognw:
> I would be willing to bet the improvements don't cause a significant efficiency increase; in particular, I think the changes will save at most 150 gas per call. But sure, I will include the updated logic as another algorithm for [my analysis](https://github.com/chgorman/isqrt-gas). Doing this will take a bit of time.

No need to bet. Agree it takes time.
Also, please be mindful this also takes time to read and understand. The reason I'm bringing back the gas measurements is because the PR is getting complex to follow, and the improvements are ~200 gas units. If you think the changes in log2 will save at most 150 gas per call, I think that's enough reason to recompare.

> As noted, [my analysis](https://github.com/chgorman/isqrt-gas) includes a proof of validity that the proposed method results a valid result. Does anyone dispute this assertion? That is, does anyone say my proof is invalid? If not, great; otherwise, where is the invalid logic?

We're not saying a proof **is needed in the code**, I say a proof would be great for developers to follow the code, and not everyone will follow a full paper. I haven't approved the PR because I need to sit down and validate the proof while also checking that the implementation is well commented. I'm pretty sure a couple of sentences are enough for the code to make sense when reading.

> I now ask the same of you: where is your proof that the current method (that is, the current method used by OZ) is valid? In particular, where is your proof that the return logic ensures that the correct value is returned?

It's been in the library for almost two years and it's been audited. That's enough for us. We're not asking you for a proof, we're trying to iterate the implementation before approving, that's the way we work and we generally come to better results this way.

Again, please be mindful this takes time from us as well. We leave the PR open for a bit (because we have other priorities than ~200 gas savings) and when we're back there's another full page of explanation that we need to follow before replying. I appreciate you're sharing the results, but the paper is enough.

I'll make some updates to the PR, and try to leave it in good shape comment-wise.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,74 @@ library Math {
...
+                aAux >>= 8;
+                result <<= 4;
+            }
+            if (aAux >= (1 << 4)) {
+                aAux >>= 4;
+                result <<= 2;
+            }
+            if (aAux >= (1 << 2)) {
+                result <<= 1;
+            }
```
ernestognw, line 389: Prettier forces this layout although honestly I liked the previous one.
I'm approving but we need another approve from @Amxx. I'd appreciate if you have suggestions for keeping the previous layout:

```solidity
if (aAux >= (1 << 128)) { aAux >>= 128; result = 1 << 64; }
if (aAux >= (1 << 64 )) { aAux >>= 64;  result <<= 32; }
if (aAux >= (1 << 32 )) { aAux >>= 32;  result <<= 16; }
if (aAux >= (1 << 16 )) { aAux >>= 16;  result <<= 8;  }
if (aAux >= (1 << 8  )) { aAux >>= 8;   result <<= 4;  }
if (aAux >= (1 << 4  )) { aAux >>= 4;   result <<= 2;  }
if (aAux >= (1 << 2  )) {               result <<= 1;  }
```

ernestognw (approved):

chgorman:
Using an `if` statement rather than the ternary operator saves 3 gas. For that reason, I think it's best to leave it.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,77 @@ library Math {
...
+            if (a <= 1) {
+                return a;
+            }
+            // This check ensures no overflow at return; see below.
+            if (a >= uint256(type(uint128).max) ** 2) {
+                return type(uint128).max;
+            }
+
+            uint256 aAux = a;
+            uint256 result = 1; // Initial approximate bit length.
```
chgorman, line 357: `result` stores the approximation (and eventual result). It's not "initial approximate bit length".

chgorman:
I appreciate the suggestions, but I will need to look over it some more.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,77 @@ library Math {
...
+
+            // For the first guess of `result` (e), we get the biggest power of 2 which is smaller than sqrt(a)
+            // (i.e. 2^e <= sqrt(a)). We know that e is at most 127 given (2^128)^2 overflows an uint256.
+            // Thus, we approximate e by iterating 2^{i/2} where i starts at 128, and applying the exponent
+            // to e if the result is still smaller than a (up to e == 127).
+            if (aAux >= (1 << 128)) {
+                aAux >>= 128;
+                result = 1 << 64;
```
Amxx, line 365: Result is already 1 (from line 357).
For consistency I would do
```suggestion
                result <<= 64;
```

Also, result is a value in the stack already, so wost should be similar. Its not like if it was storage and the read was expensive.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,77 @@ library Math {
...
+                aAux >>= 8;
+                result <<= 4;
+            }
+            if (aAux >= (1 << 4)) {
+                aAux >>= 4;
+                result <<= 2;
+            }
+            if (aAux >= (1 << 2)) {
+                result <<= 1;
+            }
```
Amxx, line 389: Here we should use the branchless approach from the log functions

```suggestion
            uint256 isGt;

            uint256 isGt = SafeCast.toUint(aAux > (1 << 128) - 1);
            aAux >>= isGt * 128;
            result <<= isGt * 64;

            uint256 isGt = SafeCast.toUint(aAux > (1 << 64) - 1);
            aAux >>= isGt * 64;
            result <<= isGt * 32;

            uint256 isGt = SafeCast.toUint(aAux > (1 << 32) - 1);
            aAux >>= isGt * 32;
            result <<= isGt * 16;

            uint256 isGt = SafeCast.toUint(aAux > (1 << 16) - 1);
            aAux >>= isGt * 16;
            result <<= isGt * 8;

            uint256 isGt = SafeCast.toUint(aAux > (1 << 8) - 1);
            aAux >>= isGt * 8;
            result <<= isGt * 4;

            uint256 isGt = SafeCast.toUint(aAux > (1 << 4) - 1);
            aAux >>= isGt * 4;
            result <<= isGt * 2;

            uint256 isGt = SafeCast.toUint(aAux > (1 << 2) - 1);
            result <<= isGt * 1;
```

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,77 @@ library Math {
...
+            result = (result + a / result) >> 1; // err := result - sqrt(a)<= 2^{e-72}
+            result = (result + a / result) >> 1; // err := result - sqrt(a) <= 2^{e-144}
+
+            // After 6 iterations, no more precision can be obtained since the max result is 127.
+            // Squaring result could overflow if a >= type(uint128).max^2, case discarded at the start.
+            // result is either sqrt(a) or sqrt(a) + 1.
+            if (result * result <= a) {
+                return result;
+            }
+            return result-1;
```
Amxx, line 412: Banchless option:
```suggestion
            return result - SafeCast.toUint(a < result * result);
```

Amxx: I'm wondering if we could do something like 
```
return result - SafeCast.toUint(result > a / result);
```
with no risk of overflow ... which would allow us to get rid of lines 351 to 354

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
-            result = (result + a / result) >> 1;
-            result = (result + a / result) >> 1;
-            result = (result + a / result) >> 1;
-            result = (result + a / result) >> 1;
-            result = (result + a / result) >> 1;
-            return min(result, a / result);
+            // Take care of easy edge cases when a == 0 or a == 1
+            if (a <= 1) {
+                return a;
+            }
```
Amxx, line 350: `a <= 1` should be more expensive to compute than `a == 0`. The first one requires a push and a comparison when the second can use the `isZero` opcode. 

I was thinking that de-optimizing the function is `a == 1` would be compensated by the saving for all other values of a ... but I was not able to measure that. Its strange.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
+                aAux >>= 8;
+                result <<= 4;
+            }
+            if (aAux >= (1 << 4)) {
+                aAux >>= 4;
+                result <<= 2;
+            }
+            if (aAux >= (1 << 2)) {
+                result <<= 1;
+            }
```
Amxx, line 385: This is very simmilar to the code we have in log2. In fact, its actually `1 << (log2(a) >> 1)`. In log2, we changed the logic to branchless, and that saved some gas. Surprisingly the same approach doesn't appear to work here.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
-            return min(result, a / result);
+            // Take care of easy edge cases when a == 0 or a == 1
+            if (a <= 1) {
+                return a;
+            }
+
+            uint256 aAux = a;
+            uint256 result = 1;
+
+            // For the first guess of `result` (e), we get the biggest power of 2 which is smaller than sqrt(a)
```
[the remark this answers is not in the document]
Amxx: also, `^` can be confused with xor. We should favor `**` for exponents

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
+            // (i.e. 2^e <= sqrt(a)). We know that e is at most 127 given (2^128)^2 overflows an uint256.
```
Amxx, line 356:
```suggestion
            // For our first guess, we get the biggest power of 2 which is smaller than the square root of the target.
            // (i.e. result = 2**n <= sqrt(a) < 2**(n+1)). We know that e is at most 127 because (2**128)**2 = 2**256
            // is bigger than any uint256.
```

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
+            // Thus, we approximate e by iterating 2^{i/2} where i starts at 128, and applying the exponent
+            // to e if the result is still smaller than a (up to e == 127).
```
Amxx, line 358: I don't think this is clear. If I did no understand what the algorithm is doing already, this would not help me. I think we should rewrite that part.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
+                aAux >>= 4;
+                result <<= 2;
+            }
+            if (aAux >= (1 << 2)) {
+                result <<= 1;
+            }
+
+            // We can use the fact that 2^e <= sqrt(a) to improve the estimation
+            // by computing the arithmetic mean between the current estimation and
+            // the next one (result * 2), ensuring that result - sqrt(a) <= 2^{e-2}.
```
Amxx, line 389: Again, I don't think that is clear.

Amxx: My understanding is that we know `result <= sqrt(a) < 2 * result` ... so right now `║ result - sqrt(a) ║ < result`
However, if we take the middle of that interval, then the new confidence is `║ 3*result/2 - sqrt(a) ║ < result / 2`

That still only gives us `result/2 = 2**(e-1)` ... where does `2**(e-2)` come from?

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
+
+            // Each Newton iteration will have f(x) = (x + a / x) / 2.
+            // Given the error (ε) is defined by x - sqrt(a), then we know that
+            // ε + 1 == ε^2 / 2x <= ε^2 / 2 * sqrt(a).
+            result = (result + a / result) >> 1; // err := result - sqrt(a) <= 2^{e-4.5}
+            result = (result + a / result) >> 1; // err := result - sqrt(a) <= 2^{e-9}
+            result = (result + a / result) >> 1; // err := result - sqrt(a) <= 2^{e-18}
+            result = (result + a / result) >> 1; // err := result - sqrt(a) <= 2^{e-36}
+            result = (result + a / result) >> 1; // err := result - sqrt(a) <= 2^{e-72}
+            result = (result + a / result) >> 1; // err := result - sqrt(a) <= 2^{e-144}
```
Amxx, line 400: Is that correct?
```suggestion
            result = (result + a / result) >> 1; // ε = ║ result - sqrt(a) ║ < 2**(e - 4.5)
            result = (result + a / result) >> 1; // ε = ║ result - sqrt(a) ║ < 2**(e - 9)
            result = (result + a / result) >> 1; // ε = ║ result - sqrt(a) ║ < 2**(e - 18)
            result = (result + a / result) >> 1; // ε = ║ result - sqrt(a) ║ < 2**(e - 36)
            result = (result + a / result) >> 1; // ε = ║ result - sqrt(a) ║ < 2**(e - 72)
            result = (result + a / result) >> 1; // ε = ║ result - sqrt(a) ║ < 2**(e - 144)
```
Where does the initial 4.5 come from ? Its not documented anywhere.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
+
+            // After 6 iterations, no more precision can be obtained since the max result is 127.
```
Amxx, line 402:
```suggestion
            // After 6 iterations, `ε < 2**(e - 144) < 1` (because e < 128). No more precision can be obtained given 
            // that we are working with integers.
```

chgorman:
For the record, here has been my design philosophy:

1. **Provably correct** Integer Square Root algorithm (followed closely by)
2. Efficient as possible

**Some Background:** I first started investigating Integer Square Root algorithms because it was needed for an algebraic sigmoid bonding curve. I needed an integer square root function, so I looked at algorithms online. The current OZ alg was one I found online, and there were others which were very similar. I **attempted to prove** the algorithm's correctness, yet I ran into trouble with the return logic: the return logic (`return min(result, a/result)`) did not make sense and it is not clear that it ensures the correct result is returned (this was discussed more above). Does anyone know where that return logic (`return min(result, a/result)`) comes from?

Saying

> It's been in the library for almost two years and it's been audited. That's enough for us.

is both not a proof and a weak argument. The Integer Square Root algorithm is a fundamental operation, and the proof of validity should be clear. Yes, the proof may require some knowledge of mathematics, but it's not particularly advanced. As noted above, I asked for a proof because **if there was one**, I wanted to see it because **I** could not construct one (as noted above).

> We're not asking you for a proof

Yes, I understand. I wanted to be verbose and then wanted to iterate the **discussion** to make it concise yet understandable.

As to

> we're trying to iterate the implementation before approving, that's the way we work and we generally come to better results this way.

Great. For the record, I iterated a lot on this algorithm as well, so if you are going to make changes to the code, would you please include Max, Mean, Median, Standard Deviation over 2048 samples? This allows for a systematic investigation of updates to the code and allows for an easy comparison. You could, for example, use the 2048 samples from [this repo](https://github.com/chgorman/isqrt-gas). The entire repo is devoted to one topic: provably correct efficient integer square root algorithms in Solidity. Please also ensure that all changes don't negate the proof of convergence.

Iterating **comments** are completely different. Yes, I understand not everyone "speaks math", but the comments should paint a clear picture of the algorithm and different parts.

**Algorithm Overview:** The algorithm has 3 parts:

1. For `2**(2e-2) <= a < 2**(2e)` (equivalently, `2**(e-1) <= sqrt(a) < 2**e)`), set `result == 2**(e-1) + 2**(e-2)`. With this initialization, we have `abs(result - sqrt(a)) <= 2**(e-2)`.
2. Perform Newton iterations. Because of the specified initialization, only 6 are required until Newton error bounds ensure that the error is less than or equal to 1. This means `result == Isqrt(a)` or `result == Isqrt(a) + 1`.
3. Return `result` if `result**2 <= a`; otherwise, return `result - 1`.

contracts/utils/math/Math.sol
ernestognw: > But that still only gives us result/2 = 2**(e-1) ... where does 2**(e-2) come from?

Right. But consider we're looking for `e` such that `2**e + c == a` and the worst case for c is `2**e - 1`. If we're using `e`, we would be above the square root. So you should note this as `result/2 = 2**(e-2)`.

contracts/utils/math/Math.sol
ernestognw: Yes it's correct. Feel free to update it

ernestognw: > Where does the initial 4.5 come from ? Its not documented anywhere.

Added a reference, but comes from using the error formula from Walter Rudin. Principles of Mathematical Analysis. 3rd ed. McGraw-Hill New York, 1976. Exercise 3.16 (b)

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,69 @@ library Math {
...
+            }
+
+            // We can use the fact that 2^e <= sqrt(a) to improve the estimation
+            // by computing the arithmetic mean between the current estimation and
+            // the next one (result * 2), ensuring that result - sqrt(a) <= 2^{e-2}.
+            result = (3 * result) >> 1;
+
+            // Each Newton iteration will have f(x) = (x + a / x) / 2.
+            // Given the error (ε) is defined by x - sqrt(a), then we know that
+            // ε + 1 == ε^2 / 2x <= ε^2 / 2 * sqrt(a).
```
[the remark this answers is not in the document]
ernestognw: I think the intention is to express `e_{+1}`. Would you be ok with that syntax?

contracts/utils/math/Math.sol
ernestognw: I reworded considering your comments, I think it should be clearer.

ernestognw:
> to iterate the discussion

We'd appreciate short and straight to the point discussions.

> Iterating comments are completely different. Yes, I understand not everyone "speaks math", but the comments should paint a clear picture of the algorithm and different parts.

I think we've asked you to help with that and yet you kept expanding the explanation.

> Great. For the record, I iterated a lot on this algorithm as well, so if you are going to make changes to the code, would you please include Max, Mean, Median, Standard Deviation over 2048 samples?

We'll keep iterating this. I'm sorry but this is an open source project and not an extension of your research. We'll dig into potential optimizations but I don't think we have time to give you a report with the samples.

contracts/utils/math/Math.sol
ernestognw: I'm also getting that the current implementation is better:

```
FOUNDRY_FUZZ_RUNS=10000 forge snapshot --via-ir --optimizer-runs 1000000000 --use 0.8.24 --match-test testSqrt -vvvv

## Current
MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4823, ~: 4986)

## New
MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4883, ~: 5012)
```

```
FOUNDRY_FUZZ_RUNS=10000 forge snapshot --use 0.8.24 --match-test testSqrt -vvvv

## Current
MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 5143, ~: 5302)

## New
MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 5225, ~: 5455)
```

ernestognw:
I tested the original version by @chgorman vs the version @Amxx is proposing:

```
FOUNDRY_FUZZ_RUNS=10000 forge snapshot --via-ir --optimizer-runs 1000000000 --use 0.8.24 --match-test testSqrt -vvvv

// chgorman
MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4680, ~: 4857)

// Amxx
MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4883, ~: 5012)

// master
MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4925, ~: 5054)
```

Both @chgorman and @Amxx versions seem like an improvement over the current implementation. However, we discussed how much the following block makes readability more difficult:

```solidity
if (a >= uint256(type(uint128).max) ** 2) {
    return type(uint128).max;
}
```

And we concluded that we won't merge something we're not comfortable with even if it's an improvement in gas. In this case, the gas improvement is around ~245 gas units using the original proposal (~200 vs @Amxx's). I still think the PR is good, but I'll leave the decision of whether to keep the original proposal or use the current version to @Amxx

ernestognw (approved):
LGTM again. Feel free to iterate in comments.

contracts/utils/math/Math.sol
Amxx: We have e such that "2**e <= sqrt(a) < 2**(e+1)"

Looks like there is a confusion between a and sqrt(a) here.

Amxx:
> I tested the original version by @chgorman vs the version @Amxx is proposing:
> 
> ```
> FOUNDRY_FUZZ_RUNS=10000 forge snapshot --via-ir --optimizer-runs 1000000000 --use 0.8.24 --match-test testSqrt -vvvv
> 
> // chgorman
> MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4680, ~: 4857)
> 
> // Amxx
> MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4883, ~: 5012)
> 
> // master
> MathTest:testSqrt(uint256,uint8) (runs: 10000, μ: 4925, ~: 5054)
> ```
> 
> Both @chgorman and @Amxx versions seem like an improvement over the current implementation. However, we discussed how much the following block makes readability more difficult:
> 
> ```solidity
> if (a >= uint256(type(uint128).max) ** 2) {
>     return type(uint128).max;
> }
> ```
> 
> And we concluded that we won't merge something we're not comfortable with even if it's an improvement in gas. In this case, the gas improvement is around ~245 gas units using the original proposal (~200 vs @Amxx's). I still think the PR is good, but I'll leave the decision of whether to keep the original proposal or use the current version to @Amxx

This estimation is with fuzzing numbers in which range? If the inputs are uint128 (don't go all the way to uint256) I don't get the same differences, because the number I put in are never taking the "shortcut" in @chgorman version. I think that realistic usage will probably not be using that shortcut either, as its very rare to see such large number manipulated for math (for cryptography we don't use this function anyway), and I think we should optimize "smaller" values

Amxx:
> Feel free to iterate in comments.

I will. I honestly think the current comments are very difficult to read. For those that read french, [all the proofs, including convergence, fit into <1 page](https://fr.wikipedia.org/wiki/M%C3%A9thode_de_H%C3%A9ron#Principe). We should aim for that level of clarity and simplicity.

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,75 @@ library Math {
...
+                result <<= 1;
+            }
+
+            // We know that result <= sqrt(a) < 2 * result. We can use the fact that
+            // 2**e <= sqrt(a) to improve the estimation by computing the arithmetic
+            // mean between the current estimation and the next one (e = 1).
+            result = (3 * result) >> 1;
+
+            // We define the error as ε = result - sqrt(a). Then we know that
+            // result = 2**e−1 + 2**e−2, and therefore ε0 = 2**e−1 + 2**e−2 - sqrt(n),
```
Amxx, line 395: > we know that result = 2**e−1 + 2**e−2

AFAIK, this is not correct

With the notation we have from the start:
- first "step": `result = 2**e <= sqrt(a) < 2**(e+1))`
- second step: `result = 2**e+2**(e-1)`
and the error is `<2**(e-1)`

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,75 @@ library Math {
...
+            // 3rd ed. McGraw-Hill New York, 1976. Exercise 3.16 (b)
+
+            result = (result + a / result) >> 1; // ε_1 := result - sqrt(a) <= 2**(e-4.5)
+            result = (result + a / result) >> 1; // ε_2 := result - sqrt(a) <= 2**(e-9)
+            result = (result + a / result) >> 1; // ε_3 := result - sqrt(a) <= 2**(e-18)
+            result = (result + a / result) >> 1; // ε_4 := result - sqrt(a) <= 2**(e-36)
+            result = (result + a / result) >> 1; // ε_5 := result - sqrt(a) <= 2**(e-72)
+            result = (result + a / result) >> 1; // ε_6 := result - sqrt(a) <= 2**(e-144)
+
+            // After 6 iterations, the precision of e is already above 128 (i.e. 144). Meaning that
```
Amxx, line 407: > precision of e

We computed e at step 0 (using all the ifs). Here we are dealing with the precision of x_n

contracts/utils/math/Math.sol
```diff
@@ -339,38 +339,75 @@ library Math {
...
+            // mean between the current estimation and the next one (e = 1).
+            result = (3 * result) >> 1;
+
+            // We define the error as ε = result - sqrt(a). Then we know that
+            // result = 2**e−1 + 2**e−2, and therefore ε0 = 2**e−1 + 2**e−2 - sqrt(n),
+            // leaving ε_0 <= 2**e−2. We also see ε_{+1} == ε**2 / 2x <= ε**2 / 2 * sqrt(a)
+            // as shown in Walter Rudin. Principles of Mathematical Analysis.
+            // 3rd ed. McGraw-Hill New York, 1976. Exercise 3.16 (b)
+
+            result = (result + a / result) >> 1; // ε_1 := result - sqrt(a) <= 2**(e-4.5)
```
Amxx, line 400: if `ε_0 <= 2**e−2`
and if `ε_1 = ε_0**2 / 2x <= ε_0**2 / 2 * sqrt(a)`

Then we have 
```
ε_1 <= (2**(e−2))**2 / 2 * sqrt(a) 
    <= 2**(2*e−4) / 2 * sqrt(a) 
    <= 2**(2*e−4) / 2 
    <= 2**(2*e−5)
```
How do we get rid of the extra "e" to get to `2**(e-5)` or `2**(e-4.5)`

Amxx, line 400: In our case here:
```x_0 = 2**(e-1) + 2**(e-2) = 3 * 2**(e-2)```
and
```ε_0 ≤ 2**(e-2)```

So 
```
ε_1 = ε_n² / (2 * x_n)
    = (2**(e-2))² / (2 * 3 * 2**(e-2))
    = 2**(2e-4) / (3 * 2**e-1)
    = 2**(e-3) / 3
    = 0.6666... * 2**(e-4)
```

This is not `2**(e-4.5)`

Amxx:
I've updated the comments in #4902. They are longer, but IMO explain everything clearly without eluding any point or refering to any external ressource. It includes minor variable renaming (for clarity) but no logic change. I'd be willing to merge that version.

If @ernestognw and @chgorman can review it, we can then forward this PR to include the commits in 4902, and merge here.

contracts/utils/math/Math.sol
Amxx: Actually, the confusion is in the definition of e. Comments in the code and PDF define it differently

contracts/utils/math/Math.sol
Amxx: Rather than adding a reference (to something that I couldn't easily find ... I mean I found a 352 pages long scanned document, with 3.16 being an unrelated definition ... but that doesn't go in the right direction) I'd rather add a proof. 

AFAIK, its faster to reasd the proof than find the reference :P

Amxx:
> Iterating **comments** are completely different. Yes, I understand not everyone "speaks math", but the comments should paint a clear picture of the algorithm and different parts.

Actually, that is my goal. Past and current versions of the comments don't really paint a clear picture.

chgorman:
I'm not going to be able to devote time to look at this for at least 36 hours.

contracts/utils/math/Math.sol
ernestognw: Note 3.16 is not the exercise but a section. I think you're looking for this:

Captura de pantalla 2024-02-15 a la(s) 12 45 15 p m

And the solution:

Captura de pantalla 2024-02-15 a la(s) 12 45 29 p m

ernestognw:
Just reviewed and left my comments in #4902, I think we can move those comments here and merge.

Amxx (approved):

PR #4405, opened 2023-06-30
Exclude address(0) in ERC721 approvals

frangio:
This PR implements a few changes to make the treatment of `address(0)` in ERC721 more uniform and consistent with the rest of the library:

- `_isApprovedOrOwner`: Ensure we return false if `spender` is `address(0)`. Don't revert if the token is not minted.
- `setApprovalForAll`: Revert if `operator` is `address(0)`.
- `approve`: Revert if `to` is `address(0)`.

The reasoning to reject `address(0)` is that it can show up unexpectedly by reading uninitialized values from storage, or any uninitialized variable, or as the return value of ecrecover, etc., and in the case of `_isApprovedOrOwner` if the spender is `address(0)` it could cause the function to return true in combination with `getApproved`, since that function returns 0 to indicate that no account is approved.

Proposing first before writing tests.

Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4136.
Closes https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4169 (superceded).
Closes https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4144 (superceded).

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -114,9 +114,6 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      */
     function approve(address to, uint256 tokenId) public virtual {
         address owner = ownerOf(tokenId);
-        if (to == owner) {
-            revert ERC721InvalidOperator(owner);
-        }
 
         if (_msgSender() != owner && !isApprovedForAll(owner, _msgSender())) {
             revert ERC721InvalidApprover(_msgSender());
@@ -131,7 +128,7 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
     function getApproved(uint256 tokenId) public view virtual returns (address) {
         _requireMinted(tokenId);
 
-        return _tokenApprovals[tokenId];
+        return _getApproved(tokenId);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -220,16 +217,22 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-     * - `tokenId` must exist.
      */
     function _isApprovedOrOwner(address spender, uint256 tokenId) internal view virtual returns (bool) {
-        address owner = ownerOf(tokenId);
-        return (spender == owner || isApprovedForAll(owner, spender) || getApproved(tokenId) == spender);
+        address owner = _ownerOf(tokenId);
+        return (owner != address(0) &&
+            (spender == owner ||
+                isApprovedForAll(owner, spender) ||
+                (_getApproved(tokenId) == spender && spender != address(0))));
```
Amxx, line 235: No actionnable here, but for the record I'm not a big fan of the linter's style.

Amxx (approved):
Waiting for the tests to run (curious about the gas changes).

If tests are good this LGTM.

contracts/token/ERC721/ERC721.sol
```diff
@@ -386,8 +389,12 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * Emits an {Approval} event.
      */
     function _approve(address to, uint256 tokenId) internal virtual {
+        address owner = ownerOf(tokenId);
```
Amxx, line 392: We should note that ownerOf is already called in the public `approve`.

If we want to minimize duplicated `sload`s, we could use a construction similar to `_update`'s condition. This function is almost never used though, so optimising it is not as critical as transferFrom.

PR #4406, opened 2023-06-30
Add security considerations to ERC2771Forwarder

frangio:


contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -20,6 +20,20 @@ import {Address} from "../utils/Address.sol";
...
+ * specified in the request. This contract does not implement any kind of retribution for this gas,
+ * and it is assumed that there is an out of band incentive for relayers to pay for execution on
+ * behalf of signers. Often, the relayer is operated by a project that will consider it a user
+ * acquisition cost.
+ *
+ * By offering to pay for gas, relayers are at risk of having that gas used by an attacker toward
+ * some other purpose that is not aligned with the expected out of band incentives. If you operate a
+ * relayer, consider whitelisting target contracts and function selectors. When relaying ERC-721 or
+ * ERC-1155 transfers specifically, consider rejecting the use of the `data` field, since it can be
+ * used to execute arbitrary code.
```
ernestognw, line 36: I'd like to use the opportunity to mention the scalability constrains in case someone is running the forwarder from a single relayer. It'd be something like:

```suggestion
 *
 * === Security Considerations
 *
 * If a relayer submits a forward request, it should be willing to pay up to 100% of the gas amount
 * specified in the request. This contract does not implement any kind of retribution for this gas,
 * and it is assumed that there is an out of band incentive for relayers to pay for execution on
 * behalf of signers. Often, the relayer is operated by a project that will consider it a user
 * acquisition cost.
 *
 * By offering to pay for gas, relayers are at risk of having that gas used by an attacker toward
 * some other purpose that is not aligned with the expected out of band incentives. If you operate a
 * relayer, consider whitelisting target contracts and function selectors. When relaying ERC-721 or
 * ERC-1155 transfers specifically, consider rejecting the use of the `data` field, since it can be
 * used to execute arbitrary code.
 *
 * === Scalability Considerations
 *
 * A relayer who subsidizes the execution of arbitrary requests may face throughput limitations by
 * the chain it's running on if there's a built-in maximum amount of non-executable transactions in
 * a node. For example, Geth and its forks use a configurable 
 * https://geth.ethereum.org/docs/fundamentals/command-line-options[`accountqueue`] flag that
 * limits the number of pending transactions for a single address, causing a long queue of relayed
 * requests to be dropped by the node operators.
 *
 * In case a relayer finds this limitation, the recommendation is to distribute the load among multiple
 * relayer accounts.
 *
```

Linking tho [Geth](https://github.com/ethereum/go-ethereum/blob/8bbb16b70ef4a9e732745aac104525a6f16306d6/cmd/utils/flags.go#L375) for reference

ernestognw (approved):

ernestognw (dismissed):
Sweet, thanks!

ernestognw (approved):
imagen

pcaversaccio:
Thanks for this @frangio!

PR #4411, opened 2023-07-01
Separate typehash into multiple lines

Hicham010:
This PR splits the type hash domains into multiple lines. This approach is more beneficial because it makes reading and updating the type hash easier and less error prone.

This is possible because the Solidity compiler allows splitting of string literals: https://docs.soliditylang.org/en/latest/types.html#string-literals-and-types

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -30,10 +30,22 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
     using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
 
     bytes32 public constant BALLOT_TYPEHASH =
-        keccak256("Ballot(uint256 proposalId,uint8 support,address voter,uint256 nonce)");
+        keccak256(
+            "Ballot("
+            "uint256 proposalId,"
+            "uint8 support,"
+            "address voter,"
+            "uint256 nonce)"
+        );
     bytes32 public constant EXTENDED_BALLOT_TYPEHASH =
         keccak256(
-            "ExtendedBallot(uint256 proposalId,uint8 support,address voter,uint256 nonce,string reason,bytes params)"
+            "ExtendedBallot("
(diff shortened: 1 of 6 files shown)


contracts/governance/Governor.sol
```diff
@@ -28,9 +28,20 @@ import {IGovernor, IERC6372} from "./IGovernor.sol";
 abstract contract Governor is Context, ERC165, EIP712, IGovernor, IERC721Receiver, IERC1155Receiver {
     using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
 
-    bytes32 public constant BALLOT_TYPEHASH = keccak256("Ballot(uint256 proposalId,uint8 support)");
+    bytes32 public constant BALLOT_TYPEHASH =
+        keccak256(
+            "Ballot("
+            "uint256 proposalId,"
+            "uint8 support)"
```
frangio, line 35: I'm not sure if I like this or if it would be better to put the parenthesis on its own line...

```solidity
            "Ballot("
            "uint256 proposalId,"
            "uint8 support"
            ")"
```

What do you think?

frangio (commented):
We've discussed this before and preferred the current style: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2854#issuecomment-989306136

I don't know if my opinion has changed... I'm not convinced this is less error prone than what we have currently. Let's see what others think. @Amxx @ernestognw

contracts/governance/Governor.sol
Hicham010: I found it more compact while not giving up on readability

Amxx:
> I'm not convinced this is less error prone than what we have currently

IMO its more error prone. Easy to forget a comma or add a space.

frangio:
Yeah personally I think it's slightly more error prone as well.

To make this less error prone I think we need Solidity support as in https://github.com/ethereum/solidity/issues/14157, https://github.com/ethereum/solidity/issues/14208.

Thanks @Hicham010 regardless.

PR #4415, opened 2023-07-03
Add override guidelines

frangio:
The idea is to establish guidelines to avoid common errors around multiple inheritance. These guidelines are quite strict, and we have to review if we're following them, and implement a rule to check it automatically.

Fixes LIB-897

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -137,7 +137,7 @@ In addition to the official Solidity Style Guide we have a number of other conve
 
   * Custom error names should not be declared twice along the library to avoid duplicated identifier declarations when inheriting from multiple contracts.
 
-* It is important to follow a few rules around the use of Solidity overrides to avoid introducing unintended consequences due to the interaction with multiple inheritance:
+* Solidity function overrides should follow the rules listed below in order to avoid introducing unintended consequences due to the interaction with multiple inheritance:
 
     1. When overriding a function `foo`, always invoke `super.foo`, and pass the same arguments that were received.
     2. Never use `super` outside of an override, or for a function other than the one being overridden.
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -136,3 +136,8 @@ In addition to the official Solidity Style Guide we have a number of other conve
     4. Declare the error in an extension if the error only happens in such extension or child contracts.
 
   * Custom error names should not be declared twice along the library to avoid duplicated identifier declarations when inheriting from multiple contracts.
+
+* It is important to follow a few rules around the use of Solidity overrides to avoid introducing unintended consequences due to the interaction with multiple inheritance:
+
+    1. When overriding a function `foo`, always invoke `super.foo`, and pass the same arguments that were received.
+    2. Never use `super` for a function other than the one being overridden.
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -140,4 +140,4 @@ In addition to the official Solidity Style Guide we have a number of other conve
 * It is important to follow a few rules around the use of Solidity overrides to avoid introducing unintended consequences due to the interaction with multiple inheritance:
 
     1. When overriding a function `foo`, always invoke `super.foo`, and pass the same arguments that were received.
-    2. Never use `super` for a function other than the one being overridden.
+    2. Never use `super` outside of an override, or for a function other than the one being overridden.


frangio:
Logical operators because of short-circuiting may lead to violating the first rule.

```solidity
function supportsInterface(bytes4 interfaceId) public view virtual override(ERC721, IERC165) returns (bool) {
    return interfaceId == bytes4(0x49064906) || super.supportsInterface(interfaceId);
}
```

Could this ever be an issue? Should there be an exception for this kind of pattern?

frangio:
In some cases we may want to change the arguments that get passed to `super` too.

For example, to implement https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3850 with backwards compatibility for Permit and Votes, we may write:

```solidity
function operationNonces(bytes32 typeHash, address signer) public view virtual returns (uint256) {
    return super.operationNonces(COMMON, signer);
}
```

ernestognw:
> ...
> 
> Could this ever be an issue? Should there be an exception for this kind of pattern?

I don't think it's an issue because the functionality is essentially this:

```solidity
function supportsInterface(bytes4 interfaceId) public view virtual override(ERC721, IERC165) returns (bool) {
    bool supported = super.supportsInterface(interfaceId);
    return interfaceId == bytes4(0x49064906) || supported;
}
```

I'm sure there's an embedded rule in this but I don't know how to describe it properly. I think the first rule can be rewritten to something like:

> When overriding a function `foo`, always invoke `super.foo` and pass the same arguments that were received **unless there's another condition superseding the `super.foo` call** 

In this case, it's proven that `interfaceId == bytes4(0x49064906)` supersedes `super.supportsInterface` because the OR condition only requires one of them to be true, and that's why the short-circuit happens in the first place, so I don't think the rule is violated.

ernestognw:
<details>
<summary>@frangio's on `super` arguments</summary>

> In some cases we may want to change the arguments that get passed to `super` too.
> 
> For example, to implement #3850 with backwards compatibility for Permit and Votes, we may write:
> 
> ```solidity
> function operationNonces(bytes32 typeHash, address signer) public view virtual override returns (uint256) {
>     if (typeHash == PERMIT_TYPEHASH) {
>         return super.operationNonces(COMMON, signer);
>     } else {
>         return super.operationNonces(typeHash, signer);
>     }
> }
> ```
> 
> Perhaps in that case we should write:
> 
> ```diff
>  function operationNonces(bytes32 typeHash, address signer) public view virtual override returns (uint256) {
>      if (typeHash == PERMIT_TYPEHASH) {
> -        return super.operationNonces(COMMON, signer);
> +        return operationNonces(COMMON, signer);
>      } else {
>          return super.operationNonces(typeHash, signer);
>      }
>  }
> ```
> 
> But this is now breaking the rule that super should always be called, and it's interrupting a chain of super calls and restarting it from the top, which could cause other issues. For example imagine an override that does `super.operationNonces(typeHash, signer) + 1`, it would now result in `+ 2`. This is definitely worse, the first option seems better.

</details>

In this case I don't think the problem is the rule, but the way the condition is presented. If what we want to change is the argument, we may want to do something like this instead:

```solidity
function operationNonces(bytes32 typeHash, address signer) public view virtual override returns (uint256) {
    return super.operationNonces(typeHash == PERMIT_TYPEHASH ? COMMON : typeHash, signer);
}
```

I'd argue the original snippet you presented still fist into the first rule because we're indeed **always invoking** super. There's no single `typeHash` value passed as an argument that causes the `super.operationNonces` not to be called.

frangio:
> ```solidity
> function operationNonces(bytes32 typeHash, address signer) public view virtual override returns (uint256) {
>     return super.operationNonces(typeHash == PERMIT_TYPEHASH ? COMMON : typeHash, signer);
> }
> ```

This is the same, what's broken is the part in the guidelines where it says "and pass the same arguments that were received".

GUIDELINES.md
```diff
@@ -136,3 +136,8 @@ In addition to the official Solidity Style Guide we have a number of other conve
     4. Declare the error in an extension if the error only happens in such extension or child contracts.
 
   * Custom error names should not be declared twice along the library to avoid duplicated identifier declarations when inheriting from multiple contracts.
+
+* It is important to follow a few rules around the use of Solidity overrides to avoid introducing unintended consequences due to the interaction with multiple inheritance:
```
ernestognw, line 140: I think it's important is inferred because it's on the guidelines, and I'd rather write the guideline as a straight-forward rule:

```suggestion
* Solidity function overrides should follow the rules listed below in order to avoid introducing unintended consequences due to the interaction with multiple inheritance:
```

frangio:
Regarding `supportsInterface`, the rule (as it is currently written) is violated because due to short circuiting there is a path through which `super.supportsInterface` is not invoked. The rule says it should always be invoked. It may be ambiguous whether it refers to syntactically or at runtime, but I meant that the code for super should run.

The issue this may cause is if an override "higher" in the chain does something like `require(...)` then this require will never execute.

ernestognw:
The `require` example was very clear, thanks.

Then I agree that the rule is violated because it's not invoked and it may contain a `require` that's being ignored and **should** be called. The reason I see here for not calling the `super.` is to avoid unnecessary code execution (gas costs), so I'd say the rule is correct, it just happens to be in Solidity where execution costs money. I'd suggest mentioning that short-circuits should be allowed only for non-reverting functions.

frangio:
Closing as stale. Feel free to reopen and merge if this is wanted.

PR #4416, opened 2023-07-03
Access Manager 

Amxx:
Fixes LIB-798

#### PR Checklist

- [x] Tests
- [ ] Documentation


test/access/manager/AccessManager.test.js
Amxx, line 1: This is an old version that doesn't match the current code

test/governance/extensions/GovernorTimelockManaged.test.js
Amxx, line 1: This is an old version that doesn't match the current code

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,498 @@
...
+     */
+    function canCall(address caller, address target, bytes4 selector) public view virtual returns (bool, uint32) {
+        AccessMode mode = getContractMode(target);
+        if (mode == AccessMode.Open) {
+            return (true, 0);
+        } else if (mode == AccessMode.Closed) {
+            return (false, Time.MAX_DURATION.get()); // use 0 ?
+        } else if (caller == address(this)) {
+            // Caller is AccessManager => call was relayed. In that case the relay already checked permissions.
+            return (true, 0);
```
frangio, line 114: This is concerning. I think we should have a flag in storage (transient storage, eventually) during a relay, and only return true if the flag is set, possibly we should only return true for the specific target/selector combination that was relayed.

contracts/access/manager/IAccessManager.sol
frangio, line 1: I still think that group identifiers need to be integers, and not `bytes32`. Again, the reasoning is that sequential integers (low numbers) are a lot easier to inspect/audit in a multisig operation, governance proposal, etc., than long strings of hexadecimals, especially if they are hashes.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,498 @@
...
+import "../../utils/Context.sol";
+import "../../utils/types/Time.sol";
+
+
+
+
+
+
+
+contract DelayedActions is Context {
```
frangio, line 17: `withDelay` and `_executeCheck` are not used. Overall I would merge this into `AccessManager`.

contracts/access/manager/IAuthority.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+interface IAuthority {
+    function canCall(address caller, address target, bytes4 selector) external view returns (bool allowed, uint32 delay);
```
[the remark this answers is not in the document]
Amxx: When the governor plans a call on some contract, and the contract has a manager, we are going to call `canCall` on that manager to check the permissions. This will be decoded as `(bool, uint32)` because we need the delay.

Are you suggesting that we do the decoding manually depending on the return length ? and that if the return length is more than a give value we assume what is beyond the interface's `bool` is a `uint32`. I find that quite dangerous.

contracts/governance/extensions/GovernorTimelock.sol
```diff
@@ -0,0 +1,191 @@
...
+
+        uint32 setback = 0;
+
+        ExecutionDetail[] memory details = proposalExecutionDetails(proposalId);
+        for (uint256 i = 0; i < targets.length; ++i) {
+            if (details[i].manager != address(0)) {
+                IAccessManager(details[i].manager).schedule(targets[i], calldatas[i]);
+            }
+            setback = uint32(Math.max(setback, details[i].delay)); // cast is safe, both parameters are uint32
+        }
```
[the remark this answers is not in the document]
Amxx: I see what you mean. I planned to use memory instead of storage when fetching `_executionDetails[proposalId]` because everything will be accessed anyway (and I wanted to avoid double sloads). I did no see the length in the heap as a cost issue, but you're right that be don't need everything to be available in memory at the same time, so we can extend less.

contracts/utils/types/Time.sol
```diff
@@ -0,0 +1,176 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.19;
+
+library Time {
```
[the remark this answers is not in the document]
Amxx: Yes, UDVT adds verbosity. and it is a downside. (honestly its not that bad, its not worst than pre 0.8.0 math).
On the other hand I think the type safety is a real plus. If prevents you from adding two timestamps.

The main issue we had with Timers (IMO) was the fact they were deeply connected to block.timestamp, and we could not use them for block numbers. Here, its mostly about adding timepoints and duration. The presence of `block.timestamp` is really reduced to a minimum. Apart from `isPast` everything is block.number compatible. We could have a `isBefore(Timepoint t1, Timepoint t2)` for that.

---

TLDR:
- Yes there are some Timers vibe, but I designed that so that issues we add with Timers would not apply here
- Its compatible with block.number (with a single exception that can easily be replaced)
- Type checking is awesome
- Verbosity is not that bad IMO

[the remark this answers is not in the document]
Amxx: In particular the `type Delay is uint112;` is an abstraction that we built in a non reusable way for ACDAR. This is something that we really need accessible with an interface that abstracts the complexity. That is not something you replace with `schedule + setback < block.timestamp`

contracts/access/manager/AccessManager.sol
Amxx: For the record, `_executeCheck` is used in `AccessManager.relay`

contracts/utils/types/Time.sol
```diff
@@ -0,0 +1,130 @@
...
+        return pack(self.get(), newValue, effect);
+    }
+
+    /**
+     * @dev Update a Delay object so that it takes a new duration after at a timepoint that is automatically computed
+     * to enforce the old delay at the moment of the update.
+     */
+    function update(Delay self, uint32 newValue) internal view returns (Delay) {
+        uint32 value = self.get();
+        uint32 setback = value > newValue ? value - newValue : 0; // todo: 0 means immediate update. ACDAR does something more opinionated
```
frangio, line 108: Perhaps there should be a `minEffectDelay` argument to replace 0.

contracts/utils/types/Time.sol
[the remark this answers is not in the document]
ernestognw: Catching up here, I agree it's verbose and difficult to follow but I think this particular abstraction makes sense, especially for preserving the guarantee of a delay when it's reduced.

My understanding of that guarantee is that we have a strong opinion of how to deal with an update, and thus such opinion should be reflected as a recommendation in the library as a whole (the repository, not this `library` itself).

> Instead of timepoint.add(setback).isPast(), it's a lot more familiar and intuitive to see schedule + setback < block.timestamp.

Yes because we've seen it through multiple reviews. I won't expect a regular developer to be this familiar with the delay's delay. Also, note that the subtle difference between `<` and `<=` is error-prone and potentially dangerous.

I'd try to find a way to abstract our "recommendation" for delay updates but I'm still not sure if it's worth making a `Time` library at the moment. Perhaps UDVT with structs would reduce verbosity and make things clearer but it's not yet supported 😞

contracts/access/manager/IAccessManager.sol
```diff
@@ -0,0 +1,77 @@
...
+        Open
+    }
+
+    // Structure that stores the details for a group/account pair. This structures fit into a single slot.
+    struct Access {
+        // Timepoint at which the user gets the permission. If this is either 0, or in the future, the group permission
+        // are not available. Should be checked using {Time-isSetAndPast}
+        uint48 since;
+        // delay for execution. Only applies to restricted() / relay() calls. This does not restrict access to
+        // functions that use the `onlyRole` modifier.
```
ernestognw, line 22: Not reviewing the docs yet but I think this should be `onlyGroup` instead of `onlyRole`, right?

contracts/access/manager/IManaged.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+interface IManaged {
```
ernestognw, line 5: Why not keeping the `AccessManagedUnauthorized` error and the `AuthorityUpdate` event in this interface?

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,558 @@
...
+contract AccessManager is Context, IAccessManager {
+    using Time for *;
+
+    uint256 public constant ADMIN_GROUP = type(uint256).min; // 0
+    uint256 public constant PUBLIC_GROUP = type(uint256).max; // 2**256-1
+
+    mapping(address target => AccessMode mode) private _contractMode;
+    mapping(address target => mapping(bytes4 selector => uint256 group)) private _allowedGroups;
+    mapping(uint256 group => Group) private _groups;
+    mapping(bytes32 => uint48) private _schedules;
```
ernestognw, line 20:
```suggestion
    mapping(bytes32 operationId => uint48 schedule) private _schedules;
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,558 @@
...
+
+pragma solidity ^0.8.20;
+
+import {IAccessManager} from "./IAccessManager.sol";
+import {IManaged} from "./IManaged.sol";
+import {Address} from "../../utils/Address.sol";
+import {Context} from "../../utils/Context.sol";
+import {Time} from "../../utils/types/Time.sol";
+
+contract AccessManager is Context, IAccessManager {
```
ernestognw, line 11: It's been difficult to follow the naming because I think we have a few different concepts here, such as `delay`, `operation`, `action`, `access`, `group`, etc. I'd like to suggest the following:

- Use `groupId` instead of `group` for `uint256` types identifying a group (Group is already a struct)
- Use `operation` uniformly instead of `action`

ernestognw (commented):
I still need to go through the AccessManager itself, I added just a few comments on the naming and I'll continue tomorrow since I'm finding it challenging to follow the logic.

contracts/access/manager/AccessManaged.sol
```diff
@@ -0,0 +1,45 @@
...
+        _checkCanCall(msg.sender, msg.sig);
+        _;
+    }
+
+    function authority() public view returns (address) {
+        return address(_authority);
+    }
+
+    function updateAuthority(address newAuthority) public {
+        if (msg.sender != _authority) {
```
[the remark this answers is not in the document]
Amxx: `_authority` should always be an AccessManager contract. IMO we don't anyone to be able to override that. 

This function is triggered through `AccessManager.updateAuthority()` which is `onlyGroup(ADMIN_GROUP)`. 
```solidity
    modifier onlyGroup(uint256 group) {
        address msgsender = _msgSender();
        if (!hasGroup(group, msgsender)) {
            revert AccessControlUnauthorizedAccount(msgsender, group);
        }
        _;
    }
```

IMO the "act by signature" is covered there. That doesn't mean we cannot use context here, I just think this is one of the cases where the risk-benefit is not clearly in favor of using Context

contracts/utils/types/Time.sol
```diff
@@ -0,0 +1,130 @@
...
+     * In particular, the {update} function guarantees that is the delay is reduced, the old delay still applies for
+     * some time. For example if the delay is currently 7 days to do an upgrade, the admin should not be able to set
+     * the delay to 0 and upgrade immediatly. If the admin wants to reduce the delay, the old delay (7 days) should
+     * still apply for some time.
+     *
+     *
+     * The `Delay` type is 128 bits long, and packs the following:
+     * [000:031] uint32 for the current value (duration)
+     * [032:063] uint32 for the pending value (duration)
+     * [064:111] uint48 for the effect date (timepoint)
```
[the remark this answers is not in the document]
Amxx: The `split` and `pack` functions are:
```solidity
function split(Delay self) internal pure returns (uint32, uint32, uint48) {
    uint112 raw = Delay.unwrap(self);
    return (
        uint32(raw), // oldValue
        uint32(raw >> 32), // newValue
        uint48(raw >> 64) // effect
    );
}

function pack(uint32 oldValue, uint32 newValue, uint48 effect) internal pure returns (Delay) {
    return Delay.wrap(uint112(oldValue) | (uint112(newValue) << 32) | (uint112(effect) << 64));
}
```
The old (current) value is at the begining, and that then comes the new (pending) value with an offset of 32, and last comes the effect timepoint with an offset of 64.

I tried on remix
```solidity
contract Test {
    function pack(uint32 oldValue, uint32 newValue, uint48 effect) public pure returns (uint112) {
        return uint112(oldValue) | (uint112(newValue) << 32) | (uint112(effect) << 64);
    }
}
```
And calling `pack(42, 0, 0)` returns `42`. `pack(0,0,1)` on the other hand returns `18446744073709551616`.

The question is how enumerate your bits (weither you start with the low weight or the high weight) ... and we are getting into endianness again ... I'm always thinking little endian, but that may be wrong.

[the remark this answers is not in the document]
Amxx: To me it just makes more sens that the smaller bits are the ones that come first, particularly in the case were you cast to a different integer type. 

Things like this 
```solidity
    function test(uint8 v) public pure returns (uint256 r) {
        assembly {
            mstore(0, v)
            r := mload(0)
        }
    }
    ```
behave like idendity.

---

Apparently, the evm is bigendian, but values are all padded right, so the accurate description (in memory) would be
```
     * [000:143] right padding
     * [144:191]  uint48 for the effect date (timepoint)
     * [192:223]  uint32 for the current value (duration) 
     * [224:255] uint32 for the pending value (duration)
```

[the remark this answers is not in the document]
Comment (Amxx) [reply contracts/utils/types/Time.sol:57]: @frangio what do you think is the clearer?

File: contracts/access/manager/AccessManager.sol (hunk 10 above)
Comment (Amxx) [reply contracts/access/manager/AccessManager.sol:11]: Done, 
I also renamed `id` to `operationId` where relevant

File: contracts/access/manager/IManaged.sol (hunk 8 above)
Comment (Amxx) [reply contracts/access/manager/IManaged.sol:5]: moved

File: contracts/utils/types/Time.sol (hunk 5 above)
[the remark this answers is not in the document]
Comment (Amxx) [reply contracts/utils/types/Time.sol:18]: @ernestognw for the record, [this is what it used to look like with `Timepoint` and a `Duration` UDVT ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/308df371443910e15678435027f59a5fdd7e13c1/contracts/utils/types/Time.sol)

File: contracts/utils/types/Time.sol (hunk 6 above)
Comment (Amxx) [reply contracts/utils/types/Time.sol:108]: Perhaps. That would definitelly be on the "opinionated" side. Particularly in a context where this lib could be used with any clock. What should the value be ?

File: contracts/governance/extensions/GovernorTimelock.sol [hunk 13]
```diff
@@ -0,0 +1,201 @@
...
+     * - calling canCall on the target's manager returns a 0 delay
+     * - calling canCall on the target's manager reverts
+     * Otherwise (calling canCall on the target's manager returns a non 0 delay), return the address of the
+     * AccessManager to use, and the delay for this call.
+     */
+    function _detectExecutionDetails(address target, bytes4 selector) private view returns (ExecutionDetail memory) {
+        // If target is not a contract, skip
+        if (target.code.length > 0) {
+            // Try to fetch authority. If revert, skip
+            try IManaged(target).authority() returns (address authority) {
```
frangio, line 186: `authority()` or a clashing signature may be defined but not return enough data, and this would cause the whole thing to revert. I don't think we can use `try` here, unless we do the decoding manually something like:

```solidity
            try IManaged(target).authority() {
                uint256 returnDataSize;
                address authority;
                assembly {
                    returnDataSize := returndatasize()
                    returndatacopy(0, 0, 32)
                    authority := mload(0)
                }
```

contracts/governance/extensions/GovernorTimelock.sol
```diff
@@ -0,0 +1,201 @@
...
+                // Check can call. If revert, skip
+                try IAccessManager(authority).canCall(address(this), target, selector) returns (
+                    bool /*allowed*/,
+                    uint32 delay
+                ) {
```
frangio, line 191: Here we still have the issue that the historical `canCall` present in other libraries doesn't return a `delay`, so this decoding may fail. We can handle it in a similar way as my previous comment. What do you think?

contracts/governance/extensions/GovernorTimelock.sol
```diff
@@ -0,0 +1,201 @@
...
+
+        delete _executionDetails[proposalId];
+
+        return proposalId;
+    }
+
+    /**
+     * @dev Default delay to apply to function calls that are not (scheduled and) executed through an AccessManager.
+     */
+    function _defaultDelay() internal view virtual returns (uint32) {
```
frangio, line 166: Delays are always measured in timestamps/seconds regardless of the governor clock, right? We should make this more explicit, I would consider naming this function `_defaultDelaySeconds` to re-enforce that it dosn't follow the clock.

contracts/governance/extensions/GovernorTimelock.sol
```diff
@@ -0,0 +1,201 @@
...
+import "../Governor.sol";
+import "../../access/manager/IManaged.sol";
+import {IAccessManager} from "../../access/manager/IAccessManager.sol";
+
+/**
+ * @dev TODO
+ *
+ * _Available since v5.0._
+ */
+abstract contract GovernorTimelockCompound is IGovernorTimelock, Governor {
```
frangio, line 15: This needs a name. Tough one. Does it even qualify as a "timelock" type of extension?

contracts/governance/extensions/GovernorTimelock.sol
frangio, line 1: Does the `onlyGovernance` modifier still work correctly with this contract? I think this is fine but I want to make sure.

The execution strategy for `onlyGovernance` function calls will always be "direct", because the governor doesn't have an authority, and the executor is the governor, so this means that the calls will be allowed and they will not go through the governance call deque. Additionally the only way to make calls out of the governor is via proposals, either directly or indirectly via `governor.relay()`.

contracts/access/manager/IAuthority.sol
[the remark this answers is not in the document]
frangio: Yes, that's correct. What do you think is dangerous about it? 

> if the return length is more than a give value we assume what is beyond the interface's `bool` is a `uint32`

This assumption is basically the same that would be happening if we define the interface the way it's defined now, except that it reverts if the return length is less.

Is there something else?

contracts/utils/types/Time.sol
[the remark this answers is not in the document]
frangio: Yes, this comment was about a previous version of the library a lot of which was cut out. Resolving.

contracts/utils/types/Time.sol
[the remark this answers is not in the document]
frangio: Byte 0 should be the least significant, as in the [`BYTE`](https://www.evm.codes/#1a?fork=shanghai) opcode.

It is true that when laid in memory the smaller address is the most significant byte (big endianness). But here we are describing a word as found in the stack IMO, not in memory.

contracts/access/manager/IAuthority.sol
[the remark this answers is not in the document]
Amxx: implemented

contracts/governance/extensions/GovernorTimelock.sol
Amxx: done.

contracts/governance/extensions/GovernorTimelock.sol
Amxx: should be resolved

contracts/governance/extensions/GovernorTimelock.sol
Amxx: Rename + added docs

contracts/governance/extensions/GovernorTimelock.sol
Amxx: I would say it does: any call for which there is no enforced delay (by an AccessManager) is subject to the `_defaultDelaySeconds` that dev can customize. Also, operations need to be queued before they can be executed.

contracts/governance/extensions/GovernorTimelock.sol
frangio: We could make it explicit by special-casing `if (target == this)` in the execution details extraction.

frangio: But that would mess with actually having `restricted` functions in the Governor. Which wouldn't be so weird.

frangio:
We need to add a function like `proposalMayNeedQueueing(uint256 proposalId) returns (bool)`. Name TBD.

frangio:
We want to add a `label(proposalId, string)` function.

contracts/utils/types/Time.sol
frangio: My proposal was to define this function as `update(Delay self, uint32 newValue, uint32 minEffectDelay)` -- perhaps `minSetback` is a better name given the naming here. It would be used with `Math.min` against the current setback expression. It seems strictly less opinionated than the current code which hardcodes 0 as the minimum.

We wouldn't choose the value, not in this libary.

contracts/governance/extensions/GovernorTimelock.sol
```diff
@@ -0,0 +1,215 @@
...
+        uint256 proposalId,
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 /*descriptionHash*/
+    ) internal virtual override {
+        ExecutionDetail[] storage details = _executionDetails[proposalId];
+        ExecutionDetail memory detail;
+
+        // TODO: enforce ETA (might include some _defaultDelaySeconds that are not enforced by any authority)
```
frangio, line 127: Seems important.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+    /**
+     * @dev Cancel a scheduled (delayed) operation.
+     *
+     * Requirements:
+     *
+     * - the caller must be the proposer, or a guardian of the targeted function
+     *
+     * Emits a {Canceled} event.
+     */
+    function cancel(address caller, address target, bytes calldata data) public virtual {
```
frangio, line 527: One possible issue we have is that a guardian could cancel an operation that is part of a governance proposal. It seems like that would cause the whole batch to revert, we don't have many other options.

contracts/governance/extensions/GovernorTimelock.sol
frangio: What do we do about this?

contracts/utils/types/Time.sol
frangio: My proposal was to define this function as `update(Delay self, uint32 newValue, uint32 minEffectDelay)` -- perhaps `minSetback` is a better name given the naming here. It would be used with `Math.min` against the current setback expression. It seems strictly less opinionated than the current code which hardcodes 0 as the minimum.

We wouldn't choose the value, not in this libary.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+
+        emit GroupExecutionDelayUpdate(groupId, account, newDuration, effectPoint);
+    }
+
+    /**
+     * @dev Internal version of {setGroupAdmin} without access control.
+     *
+     * Emit a {GroupAdminChanged} event
+     */
+    function _setGroupAdmin(uint256 groupId, uint256 admin) internal virtual {
```
frangio, line 315: The ADMIN_GROUP should not be allowed as an argument here.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+
+        emit GroupGranted(groupId, account, since, executionDelay);
+    }
+
+    /**
+     * @dev Internal version of {revokeGroup} without access control. This logic is also used by {renounceGroup}.
+     *
+     * Emit a {GroupRevoked} event
+     */
+    function _revokeGroup(uint256 groupId, address account) internal virtual {
```
frangio, line 266: PUBLIC_GROUP should not be allowed as an argument in many of these functions.

contracts/access/manager/utils/AccessManagedAdapter.sol
```diff
@@ -0,0 +1,44 @@
...
+     * @dev Initializes an adapter connected to an AccessManager instance.
+     */
+    constructor(address initialAuthority) AccessManaged(initialAuthority) {}
+
+    /**
+     * @dev Relays a function call to the target contract. The call will be relayed if the AccessManager allows the
+     * caller access to this function in the target contract, i.e. if the caller is a member of the group that is
+     * allowed for the function.
+     */
+    function relay(address target, bytes calldata data) external payable {
```
ernestognw, line 30: I think we need to warn users to not authorize any group to call `relay`. Imagine the following scenario:

- There's a `ProxyAdmin` owned by the `AccessManagedAdapter`
- There are two participants in the system: Alice and Bob
- Alice can call `upgradeAndCall` in the `ProxyAdmin` but Bob can't
- Additionally, this adapter itself is allowed to call `upgradeAndCall`
- Bob, on the other hand, is allowed to call `relay`

Because of its ability to call `relay`, Bob can reenter this function by calling relay so that `_msgSender()` is the address of the adapter who's also allowed to call `upgradeAndCall`.

Perhaps the scenario is a bit far fetched but I'd expect some cases. We may want to check at least that `_msgSender() != address(this)`

```suggestion
    function relay(address target, bytes calldata data) external payable {
        if(_msgSender() == address(this)) revert AccessManagedAdapterReentrantCall();
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+    }
+
+    /**
+     * @dev Schedule a delayed operation, and return the operation identifier.
+     *
+     * Emits a {Scheduled} event.
+     */
+    function schedule(address target, bytes calldata data) public virtual returns (bytes32) {
+        address caller = _msgSender();
+        bytes4 selector = bytes4(data[0:4]);
```
ernestognw, line 452: All of these `data[0:4]` should be:

```suggestion
        bytes4 selector = bytes4(msgData()[0:4]);
```

Otherwise we'd have a mismatch with the sender if this contract is using the ERC2771Forwarder

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+
+    /**
+     * @dev Execute a function that is delay restricted, provided it was properly scheduled beforeend, or the
+     * execution delay is 0.
+     *
+     * Emits a {Executed} event if the call was scheduled. Unscheduled call (with no delay) do not emit that event.
+     */
+    function relay(address target, bytes calldata data) public payable virtual {
+        address caller = _msgSender();
+        bytes4 selector = bytes4(data[0:4]);
```
ernestognw, line 481:
```suggestion
        bytes4 selector = bytes4(msgData()[0:4]);
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+     *
+     * Requirements:
+     *
+     * - the caller must be the proposer, or a guardian of the targeted function
+     *
+     * Emits a {Canceled} event.
+     */
+    function cancel(address caller, address target, bytes calldata data) public virtual {
+        address msgsender = _msgSender();
+        bytes4 selector = bytes4(data[0:4]);
```
ernestognw, line 529:
```suggestion
        bytes4 selector = bytes4(msgData()[0:4]);
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+    function setGrantDelay(uint256 groupId, uint32 newDelay) public virtual onlyGroup(ADMIN_GROUP) {
+        _setGrantDelay(groupId, newDelay, false);
+    }
+
+    /**
+     * @dev Internal version of {grantGroup} without access control.
+     *
+     * Emit a {GroupGranted} event
+     */
+    function _grantGroup(uint256 groupId, address account, uint32 grantDelay, uint32 executionDelay) internal virtual {
```
ernestognw, line 250: I've been thinking of a few scenarios where I think it makes sense to disallow `address(this)`. For example, if someone decides to make the manager itself its initial address.

Another thing to consider is that any user with permissions to call the `relay` function of the manager inherits its permission. We can just warn users since it's low likely imo but I think it's easier to:

```suggestion
        function _grantGroup(uint256 groupId, address account, uint32 grantDelay, uint32 executionDelay) internal virtual {
        if (account == address(this)) {
            revert AccessManagerInvalidAdmin(address(this));
        }
```

ernestognw (commented):
Didn't finish reviewing yet, I'll come back at this tomorrow

contracts/access/manager/AccessManager.sol
ernestognw: Ok I realized this is wrong completely.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+    // ==================================================== OTHERS ====================================================
+    /**
+     * @dev Change the AccessManager instance used by a contract that correctly uses this instance.
+     *
+     * Requirements:
+     *
+     * - the caller must be a global admin
+     */
+    function updateAuthority(IManaged target, address newAuthority) public virtual onlyGroup(ADMIN_GROUP) {
+        // todo set delay or document risks
```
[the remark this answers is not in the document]
Amxx: by making this "managed" by address(this)

```solidity
if (!canCall(_msgSender(), address(this), this.updateAuthority.selector) {
    revert Unauthorized();
}
```

+ set permission to some group, and add a delay to it ...

It involves some setup that could possibly be modified/broken by the admin.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+     * Emits a {AccessModeUpdated} event.
+     */
+    function _setContractMode(address target, AccessMode mode) internal virtual {
+        _contractMode[target] = mode;
+        emit AccessModeUpdated(target, mode);
+    }
+
+    // ============================================== DELAYED OPERATIONS ==============================================
+    /**
+     * @dev Return the timepoint at which an operation was scheduled.
```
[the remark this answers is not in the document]
Amxx: 💯 

How should be name this timepoint ?

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+     *
+     * Emit a {GroupExecutionDelayUpdate} event
+     */
+    function _setExecuteDelay(uint256 groupId, address account, uint32 newDuration, bool immediate) internal virtual {
+        if (_groups[groupId].members[account].since == 0) {
+            revert AccessManagerAcountNotInGroup(groupId, account);
+        }
+
+        // Here, we cannot use the "normal" `update` workflow because the delay is checked at execution and not
+        // enforced when scheduling.
```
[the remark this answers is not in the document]
Amxx: The assumption is that delay is checked at the begining (when sheduling) and not at the end.

> Why don't we enforce the delay at scheduling?

That is something inherited from the DelayedAction construction. I'm not 100% sure if it is still necessary. Will try to redesign that part and see if womething breaks.

contracts/governance/extensions/GovernorTimelock.sol
Amxx: That is why there is a TODO. The Governor part is still WIP

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+     * This function is usually called by the targeted contract to control immediate execution of restricted functions.
+     * Therefore we only return true is the call can be performed without any delay. If the call is subject to a delay,
+     * then the function should return false, and the caller should schedule the operation for future execution.
+     *
+     * We may be able to hash the operation, and check if the call was scheduled, but we would not be able to cleanup
+     * the schedule, leaving the possibility of multiple executions. Maybe this function should not be view?
+     *
+     * NOTE: The IAuthority interface does not include the `uint32` delay. This is an extension of that interface that
+     * is backward compatible. Some contract may thus ignore the second return argument. In that case they will fail
+     * to identify the indirect workflow, and will consider call that require a delay to be forbidden.
```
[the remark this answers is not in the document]
Amxx: > Is this because the contract started being used so we need to guarantee backward compatibility?

No

@frangio feels like the interface for `canCall` should have a simple type, that is just a boolean. Our implementation of that interface requires an additional value (for the delay) that we implemented in a backward compatbiel way.

contracts/access/manager/utils/AccessManagedAdapter.sol
Amxx: I'm not sure I follow.

- The `ProxyAdmin` is owner by the `AccessManagedAdapter`. So the only way to perform an upgrade is through the adapter's `relay` function.
- When calling the `relay` function, the adapter checks that the `_msgSender()`  is authorize to call the targeted function. In that case that means being allowed to call the `upgradeAndCall` on the `ProxyAdmin`

You are worried that someone may set the system such that:
- the `upgradeAndCall` on the `ProxyAdmin` is set so that you need the `UPGRADE` group to call it.
- the adapter is in the `UPGRADE` group (with Alice)
- the `relay` function on the adapter is set so that you need `SOME_GROUP` to call it (even though the function is not restricted and that is not enforced)
- Bob is added to `SOME_GROUP`

So that bob can call `relay` directly, with `target = adapter` and data that is a `relay` call, that way the adapter would allow the operation, since Bob is authorized to call relay ... and that would reenter, and execute with `msgSender = adapter` ... which is authorized to do an upgrade?

I find that very unlikelly to happen by mistake. You have to screw up twice: 
- give the right to upgrade to the adapter: that error could honestly happen
- setup call permissions for the relay function in the access manager: that is very unlikelly.

If we thing that is a serious issue, then why do 
```
if (_msgSender() == address(this)) revert AccessManagedAdapterReentrantCall();
```
and not
```
if (target == address(this)) revert AccessManagedAdapterReentrantCall();
```
?

@frangio WDYT? Are you concerned by this?

contracts/governance/extensions/GovernorTimelock.sol
Amxx: I'm not sure. Honestly I'm more focus on the AccessManager than on the Governor module right now

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+    function getGroupGrantDelay(uint256 groupId) public view virtual returns (uint32) {
+        return _groups[groupId].delay.get();
+    }
+
+    /**
+     * @dev Get the access details for a given account in a given group. These details include the timepoint at which
+     * membership becomes active, and the delay applied to all operation by this user that require this permission
+     * level.
+     */
+    function getAccess(uint256 groupId, address account) public view virtual returns (Access memory) {
```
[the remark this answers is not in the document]
frangio: I meant the Time value specifically, which is inside Access.

contracts/access/manager/AccessManager.sol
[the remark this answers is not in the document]
frangio: > Is this because the contract started being used so we need to guarantee backward compatibility?

I think the answer to this is yes? We are using the same interface that is defined elsewhere, e.g.:

https://github.com/dapphub/ds-auth/blob/a51e7caa1f03eeffed66d67f3b38605e0f1f881d/src/auth.sol#L17-L21

If we want to be compatible with this we need to define `IAuthority` in this way.

contracts/access/manager/IAuthority.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.19;
+
+interface IAuthority {
+    function canCall(address caller, address target, bytes4 selector) external view returns (bool allowed);
```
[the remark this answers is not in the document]
frangio: > a `canCall` call will fail for anyone using this interface and calling the version returning a tuple.

It's the other way around. If we define it here to return `(bool allowed)`, calling `AccessManager.canCall` will succeed, the extra return value is ignored.

On the other hand, if we define it here to return `(bool allowed, uint32 delay)`, calling this function on an Authority that only returns `allowed` will revert, because the return data is not large enough to decode two values.

contracts/access/manager/utils/AccessManagedAdapter.sol
frangio: `if (target == address(this)) revert AccessManagedAdapterInvalidTarget();` makes sense IMO.

contracts/access/manager/AccessManaged.sol
```diff
@@ -0,0 +1,46 @@
...
+    modifier restricted() {
+        _checkCanCall(_msgSender(), address(this), msg.sig);
+        _;
+    }
+
+    function authority() public view virtual returns (address) {
+        return _authority;
+    }
+
+    function updateAuthority(address newAuthority) public virtual {
```
frangio, line 67: The name should be [`setAuthority`](https://github.com/dapphub/ds-auth/blob/master/src/auth.sol#L45) to match the existing interface.

frangio:
I copied over docstrings from the `feat/access-manager` branch, mostly unchanged, I slightly adapted the contract docs for `AccessManager`. The contract is very different now but it mostly applies.

contracts/access/manager/AccessManager.sol
Amxx: Calling the relay function is not obvious though:
- the target and selector must have their permission set
- the caller (`_msgSender`) must have the corresponding permission
- if there is a setback, the operation must have been scheduled.

That means that you can only relay to targets that you are explicitelly allowed to relay. So the admin would have to give the user to call relay through the relay, by setting some specific permission to the relay function ... and then the manager should have some permission (which you want to forbid).

An equally reasonnable option would be to prevent `target == address(this)` in `_setFunctionAllowedGroup`, (and also to prevent changing the mode of the manager in `_setContractMode`)

contracts/access/manager/utils/AccessManagedAdapter.sol
Amxx: `target == address(this)` is possibly something we will need if we want to use the schedule+relay to enforce delays on the admin function

contracts/access/manager/AccessManager.sol
Amxx: we have two conversations in parallele. see: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4416#discussion_r1280113464

contracts/access/manager/utils/AccessManagedAdapter.sol
Amxx: Something that we may want to do (I wanted to add that to a followup PR, but if needed lets do it here)

In schedule
```solidity
if (target == address(this)) {
   // check that caller is an admin
   // use custom delay for admin functions
} else {
  // check that caller can call
  // use delay for normal indirect operations
}
// schedule normaly
```

In relay
```solidity
if (target == address(this)) {
   revert SomeError();
}
// relay normaly
```

And add a modifier that replaces `onlyGroup(ADMIN_GROUP)` that enforce that the call:
- comes from an admin
- apply the relay logic.

That way you can only shedule a call to the manager is you are an admin, and you cannot use relay to target the manager.

Would that be good ? Should we do it in this PR?

contracts/access/manager/AccessManager.sol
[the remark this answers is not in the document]
Amxx: See other discussions about relay to self and delays

contracts/access/manager/AccessManager.sol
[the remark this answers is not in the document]
Amxx: I changed the behavior so that it now refers to the "timepoint at which an operation will be executable"

contracts/access/manager/AccessManager.sol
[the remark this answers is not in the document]
Amxx: Fixed with the change of behavior.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+        _groups[groupId].guardian = guardian;
+
+        emit GroupGuardianChanged(groupId, guardian);
+    }
+
+    /**
+     * @dev Internal version of {setGrantDelay} without access control.
+     *
+     * The `immediate` flag can be used to bypass the delay protection and force the new delay to take effect
+     * immediately.
```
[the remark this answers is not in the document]
Amxx: I think its interresting that someone filling to modify/customize the AccessManager is able to force things.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+    function getSchedule(bytes32 id) public view virtual returns (uint48) {
+        return _schedules[id];
+    }
+
+    /**
+     * @dev Schedule a delayed operation, and return the operation identifier.
+     *
+     * Emits a {Scheduled} event.
+     */
+    function schedule(address target, bytes calldata data) public virtual returns (bytes32) {
```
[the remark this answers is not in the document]
Amxx: That will cause issues with the schedule to self :/

contracts/access/manager/utils/AccessManagedAdapter.sol
Amxx: One part that we missed is the current design of the `canCall` function

```soldity
    function canCall(address caller, address target, bytes4 selector) public view virtual returns (bool, uint32) {
        AccessMode mode = getContractMode(target);
        if (mode == AccessMode.Open) {
            return (true, 0);
        } else if (mode == AccessMode.Closed) {
            return (false, 0);
        } else if (caller == address(this)) {
            // Caller is AccessManager => call was relayed. In that case the relay already checked permissions. We
            // verify that the call "identifier", which is set during the relay call, is correct.
            return (_relayIdentifier == keccak256(abi.encodePacked(target, selector)), 0);
        } else {
            // ...
        }
    }
```
if the caller is the manager (the assumption is that is the case only through relay), we don't check any group, we only check that the call is the result of a relay that was registered in the _relayIdentifier.

This means that granting permissions (group) to `address(this)` has absolutelly no effect on the outcome of `canCall`. We can block that in `_grantGroup`, so that there is a verbose error explaining that fact, but as fact as security goes, that is not a concern

frangio: Note: The above comments are about AccessManager but this thread was about AccessManagerAdapter.

contracts/access/manager/AccessManaged.sol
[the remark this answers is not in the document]
frangio: >specially for incident response

I don't think ERC-2771 should be used for incident response.

contracts/access/manager/utils/AccessManagedAdapter.sol
Amxx: es, I got confused in so many convesations :/

frangio:
@Amxx I added `relayViaAdapter` to address the issue of delays through the adapter. The only thing I don't like about this is that it's making an external call to a potentially untrusted contract from the AccessManager, because adapter is user input to the function.

contracts/access/manager/AccessManager.sol
```diff
@@ -509,20 +507,12 @@ contract AccessManager is Context, Multicall, IAccessManager {
     }
 
     /**
-     * @dev Execute a function that is delay restricted, provided it was properly scheduled beforehand, or the
+     * @dev Execute a function that is delay restricted, provided it was properly scheduled beforeend, or the
```
Amxx, line 510: lets re-add this change

contracts/access/manager/IAuthority.sol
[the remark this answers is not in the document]
ernestognw: Ok, then I think we need to document this trick somehow, maybe for another PR (perhaps opening an issue).

I think it's not immediately evident the mechanics behind the `canCall` decoding. What do you think?

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+        address msgsender = _msgSender();
+        if (!hasGroup(groupId, msgsender)) {
+            revert AccessControlUnauthorizedAccount(msgsender, groupId);
+        }
+        _;
+    }
+
+    /**
+     * @dev Check that the caller is an admin and that the top-level function currently executing has been scheduled
+     * sufficiently ahead of time, if necessary according to admin delays.
```
ernestognw, line 69: What about mentioning this as well?

```suggestion
     * sufficiently ahead of time, if necessary according to admin delays.
     *
     * IMPORTANT: This modifier relies on `msg.sig`. It'll share permissions with the `0x00000000`
     * selector if used in the `receive` or `fallback` function.
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+    function canCall(address caller, address target, bytes4 selector) public view virtual returns (bool, uint32) {
+        AccessMode mode = getContractMode(target);
+        if (mode == AccessMode.Open) {
+            return (true, 0);
+        } else if (mode == AccessMode.Closed) {
+            return (false, 0);
+        } else if (caller == address(this)) {
+            // Caller is AccessManager => call was relayed. In that case the relay already checked permissions. We
+            // verify that the call "identifier", which is set during the relay call, is correct.
+            return (_relayIdentifier == keccak256(abi.encodePacked(target, selector)), 0);
```
ernestognw, line 113: I'd suggest putting `keccak256(abi.encodePacked(target, selector))` in a private function like:

```solidity
function relayIdentifier(address target, bytes4 selector) private pure returns (bytes32) {
  return keccak256(abi.encodePacked(target, selector));
}
```

The reason is that I think it's better if the definition of the identifier is encapsulated.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     *
+     * Emits a {GroupLabel} event.
+     */
+    function labelGroup(uint256 groupId, string calldata label) public virtual onlyGroup(ADMIN_GROUP) {
+        emit GroupLabel(groupId, label);
+    }
+
+    /**
+     * @dev Give permission to an account to execute function restricted to a group. Optionally, a delay can be
+     * enforced for any function call, byt this user, that require this level of permission. This call is only
```
ernestognw, line 225: We need to rephrase this.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     * - the caller must be a global admin
+     *
+     * Emits a {GroupGuardianChanged} event
+     */
+    function setGroupGuardian(uint256 groupId, uint256 guardian) public virtual onlyGroup(ADMIN_GROUP) {
+        _setGroupGuardian(groupId, guardian);
+    }
+
+    /**
+     * @dev Update the .
```
ernestognw, line 318:
```suggestion
     * @dev Update the delay for granting a `groupId`.
```

contracts/access/manager/AccessManager.sol
```diff
@@ -63,6 +64,21 @@ contract AccessManager is Context, Multicall, IAccessManager {
         _;
     }
 
+    /**
+     * @dev Check that the caller is an admin and that the top-level function currently executing has been scheduled
+     * sufficiently ahead of time, if necessary according to admin delays.
+     */
+    modifier onlyDelayedAdmin() {
```
Amxx, line 71: The name implies that the operation is necessarily delayed. How would you feel about just calling that
```solidity
modifier onlyAdmin() {
```

contracts/access/manager/utils/AccessManagedAdapter.sol
ernestognw: My point is that we should document better the capabilities of the `relay` function. Because if someone is allowed to use it, they can reenter so that any permissions given to the manager are given to the `relay` caller.

Although I agree it's not likely, this is a valid issue that may come from the audit, I just want to make sure we're already aware of that.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+        }
+    }
+
+    // =============================================== GROUP MANAGEMENT ===============================================
+    /**
+     * @dev Give a label to a group, for improved group discoverabily by UIs.
+     *
+     * Emits a {GroupLabel} event.
+     */
+    function labelGroup(uint256 groupId, string calldata label) public virtual onlyGroup(ADMIN_GROUP) {
```
Amxx, line 219: I feel like we should use a single modifier for all the admin-restricted functions
```suggestion
    function labelGroup(uint256 groupId, string calldata label) public virtual onlyDelayedAdmin() {
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,564 @@
...
+    /**
+     * @dev Change admin group for a given group.
+     *
+     * Requirements:
+     *
+     * - the caller must be a global admin
+     *
+     * Emit a {GroupAdminChanged} event
+     */
+    function setGroupAdmin(uint256 groupId, uint256 admin) public virtual onlyGroup(ADMIN_GROUP) {
```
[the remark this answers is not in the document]
Amxx:
```suggestion
    function setGroupAdmin(uint256 groupId, uint256 admin) public virtual onlyDelayedAdmin() {
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+    /**
+     * @dev Change guardian group for a given group.
+     *
+     * Requirements:
+     *
+     * - the caller must be a global admin
+     *
+     * Emits a {GroupGuardianChanged} event
+     */
+    function setGroupGuardian(uint256 groupId, uint256 guardian) public virtual onlyGroup(ADMIN_GROUP) {
```
Amxx, line 313:
```suggestion
    function setGroupGuardian(uint256 groupId, uint256 guardian) public virtual onlyDelayedAdmin() {
```

contracts/access/manager/AccessManager.sol
Amxx: should be resolved

contracts/access/manager/AccessManager.sol
Amxx: should be resolved

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+        emit GroupGrantDelayChanged(groupId, newDelay, effect);
+    }
+
+    /**
+     * @dev Set the execution delay for an admin function of the AccessManager. This update is not
+     * immediate and follows the same rules of {setExecuteDelay}.
+     *
+     * Emits an {AdminDelayUpdated} event.
+     */
+    function setAdminFunctionDelay(bytes4 selector, uint32 newDelay) public virtual onlyGroup(ADMIN_GROUP) {
```
Amxx, line 437:
```suggestion
    function setAdminFunctionDelay(bytes4 selector, uint32 newDelay) public virtual onlyDelayedAdmin() {
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     * @dev Set the operating mode of a contract to Custom. This enables the group mechanism for per-function access
+     * restriction and delay enforcement.
+     *
+     * Requirements:
+     *
+     * - the caller must be a global admin
+     *
+     * Emits a {AccessModeUpdated} event.
+     */
+    function setContractModeCustom(address target) public virtual onlyGroup(ADMIN_GROUP) {
```
Amxx, line 496:
```suggestion
    function setContractModeCustom(address target) public virtual onlyDelayedAdmin() {
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     * @dev Set the operating mode of a contract to Close. This prevents anyone from calling any `restricted()`
+     * function.
+     *
+     * Requirements:
+     *
+     * - the caller must be a global admin
+     *
+     * Emits a {AccessModeUpdated} event.
+     */
+    function setContractModeClosed(address target) public virtual onlyGroup(ADMIN_GROUP) {
```
Amxx, line 526:
```suggestion
    function setContractModeClosed(address target) public virtual onlyDelayedAdmin() {
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+
+        // Perform call
+        Address.functionCallWithValue(target, data, msg.value);
+
+        // Reset relay identifier
+        _relayIdentifier = relayIdentifierBefore;
+    }
+
+    /**
+     * @dev Consume a scheduled operation targetting the caller. If such an operation exists, mark it as consumed
```
Amxx, line 624:
```suggestion
     * @dev Consume a scheduled operation targeting the caller. If such an operation exists, mark it as consumed
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     * (emit an {Executed} event and clean the state). Otherwize, throw an error.
```
Amxx, line 625:
```suggestion
     * (emit an {Executed} event and clean the state). Otherwise, throw an error.
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     *
+     * This is usefull for contract that want to enforce that calls targetting them were scheduled on the manager,
```
Amxx, line 627:
```suggestion
     * This is useful for contract that want to enforce that calls targeting them were scheduled on the manager,
```

contracts/access/manager/AccessManager.sol
ernestognw: I agree we should consider preventing `target == address(this)` in `_setFunctionAllowedGroup` but I'm not convinced yet. I agree the likelihood of this is very granting access to the `relay` function and also having the manager with permissions is low likelihood but I think we're recommending the `relay` function as an adapter replacement, aren't we?

In such case the recommendation we're making is to give the manager permissions so that it can relay access managed functions.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     * Emits a {GroupExecutionDelayUpdated} event
+     */
+    function _setExecuteDelay(uint256 groupId, address account, uint32 newDuration) internal virtual {
+        if (groupId == PUBLIC_GROUP) {
+            revert AccessManagerLockedGroup(groupId);
+        } else if (_groups[groupId].members[account].since == 0) {
+            revert AccessManagerAcountNotInGroup(groupId, account);
+        }
+
+        Time.Delay newDelay = _groups[groupId].members[account].delay.update(newDuration, 0); // TODO: minsetback ?
```
ernestognw, line 377: Seems important

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+
+        delete _groups[groupId].members[account];
+
+        emit GroupRevoked(groupId, account);
+    }
+
+    /**
+     * @dev Internal version of {setExecuteDelay} without access control.
+     *
+     * Emits a {GroupExecutionDelayUpdated} event
```
ernestognw, line 368:
```suggestion
     * Emits a {GroupExecutionDelayUpdated} event.
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+     */
+    function _setExecuteDelay(uint256 groupId, address account, uint32 newDuration) internal virtual {
+        if (groupId == PUBLIC_GROUP) {
+            revert AccessManagerLockedGroup(groupId);
+        } else if (_groups[groupId].members[account].since == 0) {
```
ernestognw, line 373:
```suggestion
        } else if (_groups[groupId].members[account].isSet()) {
```

contracts/utils/types/Time.sol
```diff
@@ -0,0 +1,131 @@
...
+        return SafeCast.toUint48(block.number);
+    }
+
+    /**
+     * @dev Check if a timepoint is set, and in the past
+     */
+    function isSetAndPast(uint48 timepoint, uint48 ref) internal pure returns (bool) {
+        return timepoint != 0 && timepoint <= ref;
+    }
+
```
ernestognw, line 42: I've seen multiple comparisons with `0` and perhaps this abstraction makes sense:

```solidity
function isSet(uint48 timepoint) internal pure returns (bool) {
  return timepoint != 0;
}
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+
+        emit Scheduled(operationId, caller, target, data);
+        return operationId;
+    }
+
+    /**
+     * @dev Execute a function that is delay restricted, provided it was properly scheduled beforehand, or the
+     * execution delay is 0.
+     *
+     * Emits a {Executed} event if the call was scheduled. Unscheduled call (with no delay) do not emit that event.
```
ernestognw, line 588:
```suggestion
     * Emits an {Executed} event if the call was scheduled. Unscheduled calls (with no delay) do not emit that event.
```

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+        // If caller is not authorised, revert
+        if (!allowed && (setback == 0 || when != 0 && when < minWhen)) {
+            revert AccessManagerUnauthorizedCall(caller, target, selector);
+        }
+
+        // If caller is authorised, schedule operation
+        bytes32 operationId = _hashOperation(caller, target, data);
+        uint48 timepoint = _schedules[operationId];
+        // Cannot reschedule unless the operation has expired
+        if (timepoint != 0 && timepoint + expiration() > Time.timestamp()) {
```
ernestognw, line 575:
```suggestion
        if (timepoint != 0 && timepoint + expiration() >= Time.timestamp()) {
```

ernestognw (commented):
Will review the tests later

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+        address caller = _msgSender();
+        bytes4 selector = bytes4(data[0:4]);
+
+        // Fetch restriction to that apply to the caller on the targeted function
+        (bool allowed, uint32 setback) = canCall(caller, target, selector);
+
+        uint48 minWhen = Time.timestamp() + setback;
+
+        // If caller is not authorised, revert
+        if (!allowed && (setback == 0 || when != 0 && when < minWhen)) {
```
Amxx, line 567: we could do
```suggestion
        if (!allowed && (setback == 0 || when.isSetAndPast(minWhen - 1))) {
```

Amxx: But I don't think this is the right error to trigger. There is a difference between trying to execute something you are not allowed to do at all, and scheduling it too early.

contracts/access/manager/AccessManager.sol
```diff
@@ -0,0 +1,703 @@
...
+            revert AccessManagerUnauthorizedCall(caller, target, selector);
+        }
```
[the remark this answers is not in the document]
ernestognw: I think we don't want to offer error interface stability. I know @frangio has a preference for fewer errors and I'd second that because I see value in trying to minimize the error offering.

contracts/utils/types/Time.sol
Amxx: We used to have it. It was removed during the library simplification. I'm ok to re-add it.

contracts/access/manager/AccessManaged.sol
```diff
@@ -0,0 +1,100 @@
...
+        if (newAuthority.code.length == 0) {
+            revert AccessManagedInvalidAuthority(newAuthority);
+        }
+        _setAuthority(newAuthority);
+    }
+
+    /**
+     * @dev Transfers control to a new authority. Internal function with no access restriction.
+     */
+    function _setAuthority(address newAuthority) internal virtual {
```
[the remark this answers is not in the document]
Amxx: We are missing natspecs for this file. This should indeed be documented.

contracts/access/manager/AccessManager.sol
Amxx: There are two things:
- the _adminDelay is shared by msg.sig, so if we put a delay on function `0x00000000` it will also apply to the receive function (if for any strange reason someone adds that modifier to receive or fallback).
- the canCall group/permission system does not distinguish between function `0x00000000` and `receive`. This is affecting all `restricted()` function, not just the onlyAdmin here.

The first point should be documented in `setAdminFunctionDelay`

The second point should be documented somewhere where its visible. Possibly in both `setFunctionAllowedGroup` and `AccessManaged`

contracts/access/manager/AccessManager.sol
Amxx: wdyt about the new formulation ?

contracts/access/manager/AccessManager.sol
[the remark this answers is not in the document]
Amxx: OK, but is scheduling something too early realy a "you are not authorized to call" error ?

contracts/access/manager/AccessManager.sol
ernestognw: I don't understand what it's trying to communicate, particularly this part:

> a delay can be enforced for any function call, byt this user, that require this level of permission.

I think it's confusing

contracts/utils/types/Time.sol
ernestognw: I'd be in favor of re-adding it. I think the comparison with `0` is arbitrary and an extension may want to add extra states (eg. `_DONE_TIMESTAMP` being 0 in Timelock), so an overrideable place for it would be great.

contracts/access/manager/AccessManager.sol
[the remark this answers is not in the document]
ernestognw: I'd say the authorization comes from the schedule as well, so any action taken by an allowed user but out of time should be considered "Unauthorized" imo

contracts/access/manager/AccessManager.sol
ernestognw: I agree with both, let's document them. However, I am not sure about the locations. My proposal would be to mention in the contract's NatSpec that any operation involving a selector may consider `0x00000000` collisions. This way it's more succinct than adding the note in multiple places.

Currently, a similar note is in the `restricted` function but not in `canCall`. I'd be in favor of your suggestions, but just don't want to bloat the documentation with this warning everywhere

contracts/access/manager/AccessManager.sol
Amxx: you are looking at the old version (which was bad indeed).

Current version is:
```
     * @dev Add `account` to `groupId`. This gives him the authorization to call any function that is restricted to
     * this group. An optional execution delay (in seconds) can be set. If that delay is non 0, the user is required
     * to schedule any operation that is restricted to members this group. The user will only be able to execute the
     * operation after the delay expires. During this delay, admin and guardians can cancel the operation (see
     * {cancel}).
```

ernestognw: My bad, sorry for the confusion, than one looks better. Solving

contracts/utils/types/Time.sol
frangio: FWIW, `_DONE_TIMESTAMP` was a bad idea in TimelockController.

frangio:
Merging into `feat/access-manager`.

@ernestognw Aware of your review comments will address these in a follow up PR.

contracts/utils/types/Time.sol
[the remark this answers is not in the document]
frangio: Used a variation of @ernestognw's drawing, without the padding.

contracts/access/manager/IAuthority.sol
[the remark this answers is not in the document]
frangio: I think this might be sufficiently documented now in the AuthorityUtils library.

contracts/access/manager/utils/AccessManagedAdapter.sol
frangio: This is fixed in the latest code because:
- Not possible to assign a family to AccessManager, thus not possible to assign groups to functions
- Not possible to "close" the AccessManager contract
- Relay and Schedule implement special logic for the AccessManager

contracts/access/manager/AccessManager.sol
frangio: Addressed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4520/commits/a6ba3a6a62d3f6892bc9659157fcd84ec5f9b147

contracts/access/manager/AccessManaged.sol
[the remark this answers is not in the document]
frangio: I don't see how this is related to `virtual`. If I understood this comment correctly, I added a comment in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4520/commits/3dc4de568f4f6d0911882ebb35e69857030b2294 that should address it.

contracts/access/manager/IAuthority.sol
```diff
@@ -0,0 +1,37 @@
...
+     */
+    function canCall(address caller, address target, bytes4 selector) external view returns (bool allowed);
+}
+
+/**
+ * @dev Since `AccessManager` implements an extended IAuthority interface, invoking `canCall` with backwards compatibility
+ * for the preexisting `IAuthority` interface requires special care to avoid reverting on insufficient return data.
+ * This helper function takes care of invoking `canCall` in a backwards compatible way without reverting.
+ */
+function safeCanCall(
```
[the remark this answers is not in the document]
frangio: Renamed to `canCallWithDelay`.

contracts/access/manager/AccessManager.sol
frangio: There is already a note about the receive function in AccessManaged.

For AccessManager, we should detect if msg.data.length < 4.

PR #4417, opened 2023-07-03
Optimize Clones bytecode

ZumZoom:
This PR applies optimizations suggested by 0age in [The More-Minimal Proxy](https://medium.com/coinmonks/the-more-minimal-proxy-5756ae08ee48) post.

### Backwards compatibility notes

This PR changes the bytecode of the minimal proxy which leads to different behaviour of `predictDeterministicAddress`. Upgradable contracts that are using current version of the `predictDeterministicAddress` will be affected by this breaking change in case of upgrading to a new version.

### Effective changes

- clone bytecode size is reduced by 1 byte
- deployment gas cost is reduced by 206 gas
- clone runtime cost is reduced by 4 gas

### Gas before

```
·-------------------------------------------------------------------|---------------------------|-------------|-----------------------------·
|                       Solc version: 0.8.20                        ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 10000000 gas  │
····································································|···························|·············|······························
|  Methods                                                                                                                                  │
·························|··········································|·············|·············|·············|···············|··············
|  Contract              ·  Method                                  ·  Min        ·  Max        ·  Avg        ·  # calls      ·  usd (avg)  │
·························|··········································|·············|·············|·············|···············|··············
|  $Clones               ·  $clone(address)                         ·          -  ·          -  ·      64073  ·           14  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $Clones               ·  $cloneDeterministic(address,bytes32)    ·          -  ·          -  ·      64650  ·           16  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializeNonPayable()                  ·          -  ·          -  ·      46008  ·            4  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializeNonPayableWithValue(uint256)  ·          -  ·          -  ·      46226  ·            4  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializePayable()                     ·          -  ·          -  ·      46005  ·            8  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializePayableWithValue(uint256)     ·          -  ·          -  ·      46268  ·            8  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  Deployments                                                      ·                                         ·  % of limit   ·             │
····································································|·············|·············|·············|···············|··············
|  $Clones                                                          ·          -  ·          -  ·     266824  ·        2.7 %  ·          -  │
····································································|·············|·············|·············|···············|··············
|  DummyImplementation                                              ·          -  ·          -  ·     451677  ·        4.5 %  ·          -  │
·-------------------------------------------------------------------|-------------|-------------|-------------|---------------|-------------·
```

### Gas after

```
·-------------------------------------------------------------------|---------------------------|-------------|-----------------------------·
|                       Solc version: 0.8.20                        ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 10000000 gas  │
····································································|···························|·············|······························
|  Methods                                                                                                                                  │
·························|··········································|·············|·············|·············|···············|··············
|  Contract              ·  Method                                  ·  Min        ·  Max        ·  Avg        ·  # calls      ·  usd (avg)  │
·························|··········································|·············|·············|·············|···············|··············
|  $Clones               ·  $clone(address)                         ·          -  ·          -  ·      63867  ·           14  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $Clones               ·  $cloneDeterministic(address,bytes32)    ·          -  ·          -  ·      64444  ·           16  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializeNonPayable()                  ·          -  ·          -  ·      46004  ·            4  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializeNonPayableWithValue(uint256)  ·          -  ·          -  ·      46222  ·            4  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializePayable()                     ·          -  ·          -  ·      46001  ·            8  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  $DummyImplementation  ·  initializePayableWithValue(uint256)     ·          -  ·          -  ·      46264  ·            8  ·          -  │
·························|··········································|·············|·············|·············|···············|··············
|  Deployments                                                      ·                                         ·  % of limit   ·             │
····································································|·············|·············|·············|···············|··············
|  $Clones                                                          ·          -  ·          -  ·     264484  ·        2.6 %  ·          -  │
····································································|·············|·············|·············|···············|··············
|  DummyImplementation                                              ·          -  ·          -  ·     451677  ·        4.5 %  ·          -  │
·-------------------------------------------------------------------|-------------|-------------|-------------|---------------|-------------·
```

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/seven-wolves-report.md
+++ b/.changeset/seven-wolves-report.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`Clones`: Reduced clone bytecode size by 1 byte. ([#4417](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4417))
--- a/.changeset/seven-wolves-report.md
+++ b/.changeset/seven-wolves-report.md
@@ -1,5 +1,5 @@
 ---
-'openzeppelin-solidity': patch
+'openzeppelin-solidity': major
 ---
 
 `Clones`: Reduced clone bytecode size by 1 byte. ([#4417](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4417))
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -16,7 +16,7 @@ pragma solidity ^0.8.19;
  *
  * Bytecode implementation is a bit different from EIP 1167, due to optimizations suggested by 0age in
  * https://medium.com/coinmonks/the-more-minimal-proxy-5756ae08ee48[The More-Minimal Proxy].
- * 
+ *
  * _Available since v3.4._
  */
 library Clones {
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -14,6 +14,9 @@ pragma solidity ^0.8.19;
  * (salted deterministic deployment). It also includes functions to predict the addresses of clones deployed using the
  * deterministic method.
  *
+ * Bytecode implementation is a bit different from EIP 1167, due to optimizations suggested by 0age in
+ * https://medium.com/coinmonks/the-more-minimal-proxy-5756ae08ee48[The More-Minimal Proxy].
+ * 
  * _Available since v3.4._
  */
 library Clones {
@@ -30,12 +33,11 @@ library Clones {
     function clone(address implementation) internal returns (address instance) {
         /// @solidity memory-safe-assembly
         assembly {
-            // Cleans the upper 96 bits of the `implementation` word, then packs the first 3 bytes
-            // of the `implementation` address with the bytecode before the address.
-            mstore(0x00, or(shr(0xe8, shl(0x60, implementation)), 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000))
-            // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
-            mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3))
-            instance := create(0, 0x09, 0x37)
+            // Packs the first 1 byte of the `implementation` address with the bytecode before the address.
+            mstore(0x00, or(byte(12, implementation), 0x3d602c80600a3d3981f33d3d3d3d363d3d37363d7300))
+            // Packs the remaining 19 bytes of `implementation` with the bytecode after the address.
+            mstore(0x20, or(shl(0x68, implementation), 0x5af43d3d93803e602a57fd5bf3))
+            instance := create(0, 0x0a, 0x36)
         }
         if (instance == address(0)) {
             revert ERC1167FailedCreateClone();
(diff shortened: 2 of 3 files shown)


Amxx:
Hello @ZumZoom 

Something similar was proposed and rejected in #2704. Maybe we can re-evaluate the relevance of such a change in 5.0, but I still have concerns about the consequences this could have on upgradeable contracts.

ZumZoom:
I've missed the previous PR and discussion about the upgradability compatibility. Updated the description and fixed the changeset majority.

frangio:
@Amxx Are you concerned that a contract might upgrade from v4 to v5, or that a contract using the v5 library may use `predictDeterministicAddress` for a `deployer` contract using v4?

How about adding a `predictDeterministicAddressV4` getter preserving the old bytecode? It won't be included in the contract unless it's used. We could also name the new one `predictDeterministicAddressV5`, but I think that would be confusing.

ernestognw:
I agree this improvement couldn't be made when #2704 was proposed due to backwards compatibility, but this is the time to make this change. 

So far, iiuc, the main issue is that we're worried about the consequences of replacing the clone bytecode in `predictDeterministicAddress`, we're not worried about upgradeability because of the major increase to 5.0, so I'd agree with having a `predictDeterministicAddressV4` and change the semantics of `predictDeterministicAddress` to use this more minimal proxy.

> Beside our upgradeable contract, people may build contract of their own, that are upgradeable, and that use our (vanilla) libs. For them it might be an issue. Note that we are already them when we changed the storage layout of enumerable set/map.

Can you elaborate more on this concern @Amxx? I'm not sure how the storage layout of enumerable set/map affects here.

frangio:
We are explicitly saying that v5 is incompatible so if someone upgrades their contract from v4 to v5 I would assume they are aware it carries risk. We need to document things properly in the changelog.

I would be fine with making this change and keeping `predictDeterministicAddressV4` for compatibility. We may want to use different versioning and go with V0, since this was also available in V3.

Amxx:
Yes, Keeping a version of `predictDeterministicAddressV4` would be good.

frangio:
It's important to note that this is no longer an ERC-1167 minimal proxy. If we want to make this change we would also need to update the documentation to make that clear. But I am having second thoughts now that I realize this is the case. For example, are these proxies widel supported by block explorers? The lack of standardization makes it harder to be sure...

Also note that https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2689 is an optimization that removes 1 byte from creation code while keeping the runtime code standard ERC-1167. In theory we could combine these two optimizations, but if we're not comfortable dropping ERC-1167 we could use just https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2689 and get some gains without losing the standard aspect of it.

What do you think?

ernestognw:
Very good point @frngion, I did a quick test on Goerli and indeed the proxy is not recognized. I'm also feeling wary of supporting a non-standard Minimal Proxy.

This is a decision between the gas improvements and generalized explorers' support. To put it in perspective, saving 206 units of gas per deployment will require 4,855 clones to save 0.1 ETH at a 100 gwei gas price (`(0.1 / (206 * 100e9 / 1e18)) == 4854.3`). Similarly, 4 units of gas in runtime will require 250,000 transactions to save 0.1 ETH (`(0.1 / (4 * 100e9 / 1e18)) == 250000`)

My gut is that Etherscan will eventually support this version but at the moment I'm not sure how to evaluate the lack of support in explorers vs. the savings.

frangio:
We should go with #2689. There is a PR for that at #2704 but it's old and Clones has changed since then so we'd need a new one.

PR #4419, opened 2023-07-04
Refactor ERC721 to use _update pattern

frangio:
A simpler alternative to https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377 using:

```solidity
function _update(address from, address to, uint256 tokenId) internal;
```

`_burn` now has a `from` argument.

Fixes LIB-628

### DIFF
--- a/contracts/interfaces/draft-IERC6093.sol
+++ b/contracts/interfaces/draft-IERC6093.sol
@@ -69,11 +69,11 @@ interface IERC721Errors {
 
     /**
      * @dev Indicates an error related to the ownership over a particular token. Used in transfers.
-     * @param sender Address whose tokens are being transferred.
+     * @param claimedOwner Address whose tokens are being transferred.
      * @param tokenId Identifier number of a token.
-     * @param owner Address of the current owner of a token.
+     * @param actualOwner Address of the current owner of a token.
      */
-    error ERC721IncorrectOwner(address sender, uint256 tokenId, address owner);
+    error ERC721IncorrectOwner(address claimedOwner, uint256 tokenId, address actualOwner);
 
     /**
      * @dev Indicates a failure with the token `sender`. Used in transfers.
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -198,7 +198,7 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * @dev Returns the owner of the `tokenId`. Returns `address(0)` if token doesn't exist.
      *
      * IMPORTANT: Any overrides to this function that add ownership of tokens not tracked by the
-     * core ERC721 logic MUST be matched with the use of {_updateBalance} to keep balances
+     * core ERC721 logic MUST be matched with the use of {_increaseBalance} to keep balances
      * consistent with ownership. The invariant to preserve is that for any address `a` the value returned by
      * `balanceOf(a)` must be equal to the number of tokens such that `_ownerOf(tokenId)` is `a`.
      */
@@ -370,9 +370,10 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
     }
 
     /**
-     * @dev Unsafe write access to the balances, used by extensions that "mint" tokens using an {ownerOf} override.
+     * @dev Increases the balance of `account` by `value`.
+     * Write access to the balances, used by extensions that "mint" tokens using an {ownerOf} override.
      */
-    function _updateBalance(address account, uint128 value) internal virtual {
+    function _increaseBalance(address account, uint128 value) internal virtual {
         unchecked {
             _balances[account] += value;
         }
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -151,28 +151,23 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
     /**
      * @dev See {IERC721-transferFrom}.
      */
-    function transferFrom(address from, address to, uint256 tokenId) public virtual {
-        if (!_isApprovedOrOwner(_msgSender(), tokenId)) {
-            revert ERC721InsufficientApproval(_msgSender(), tokenId);
-        }
-
+    function transferFrom(address from, address to, uint256 tokenId) public {
+        _checkApproved(from, _msgSender(), tokenId);
         _transfer(from, to, tokenId);
     }
 
     /**
      * @dev See {IERC721-safeTransferFrom}.
      */
-    function safeTransferFrom(address from, address to, uint256 tokenId) public virtual {
+    function safeTransferFrom(address from, address to, uint256 tokenId) public {
         safeTransferFrom(from, to, tokenId, "");
     }
 
     /**
      * @dev See {IERC721-safeTransferFrom}.
      */
-    function safeTransferFrom(address from, address to, uint256 tokenId, bytes memory data) public virtual {
-        if (!_isApprovedOrOwner(_msgSender(), tokenId)) {
-            revert ERC721InsufficientApproval(_msgSender(), tokenId);
-        }
+    function safeTransferFrom(address from, address to, uint256 tokenId, bytes memory data) public {
+        _checkApproved(from, _msgSender(), tokenId);
         _safeTransfer(from, to, tokenId, data);
     }
 
@@ -194,15 +189,18 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      *
      * Emits a {Transfer} event.
      */
-    function _safeTransfer(address from, address to, uint256 tokenId, bytes memory data) internal virtual {
+    function _safeTransfer(address from, address to, uint256 tokenId, bytes memory data) internal {
         _transfer(from, to, tokenId);
-        if (!_checkOnERC721Received(from, to, tokenId, data)) {
-            revert ERC721InvalidReceiver(to);
-        }
+        _checkOnERC721Received(from, to, tokenId, data);
     }
 
     /**
-     * @dev Returns the owner of the `tokenId`. Does NOT revert if token doesn't exist
+     * @dev Returns the owner of the `tokenId`. Returns `address(0)` if token doesn't exist.
+     *
+     * IMPORTANT: Any overrides to this function that add ownership of tokens not tracked by the
+     * core ERC721 logic MUST be matched with the use of {_updateBalance} to keep balances
+     * consistent with ownership. The invariant being that for any address `a` the value returned by
+     * `balanceOf(a)` must be equal to the number of tokens such that `_ownerOf(tokenId)` is `a`.
      */
     function _ownerOf(uint256 tokenId) internal view virtual returns (address) {
         return _owners[tokenId];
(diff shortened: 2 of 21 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -221,15 +219,26 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+
+    function _checkApproved(address owner, address spender, uint256 tokenId) internal view virtual {
+        if (!_isApproved(owner, spender, tokenId)) {
+            address actualOwner = _ownerOf(tokenId);
+            if (owner == actualOwner) {
+                revert ERC721InsufficientApproval(spender, tokenId);
+            } else {
+                revert ERC721IncorrectOwner(owner, tokenId, actualOwner);
+            }
+        }
```
Amxx, line 241: This function checks the approval assuming owner is correct, and if that returns false, it gets the actual owner to throw the error. So the argument is considered correct in one case, and not correct in the other ?

I'm not really confortable with that.

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,41 +312,51 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+        _update(from, to, tokenId);
+    }
 
-        _beforeTokenTransfer(from, to, tokenId, 1);
-
-        // Check that tokenId was not transferred by `_beforeTokenTransfer` hook
-        owner = ownerOf(tokenId);
-        if (owner != from) {
-            revert ERC721IncorrectOwner(from, tokenId, owner);
+    function _update(address from, address to, uint256 tokenId) internal virtual {
```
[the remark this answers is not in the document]
Amxx: In particular, that is true of burn:

When you want to burn a token, you now have to get the owner using one sload, and then verify that you got the correct one by duplicating this sload.

contracts/token/ERC721/ERC721.sol
frangio: I mainly did this to get the current tests to pass, I don't know if this is necessary.

contracts/token/ERC721/ERC721.sol
```diff
@@ -221,15 +219,26 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
      * - `tokenId` must exist.
      */
-    function _isApprovedOrOwner(address spender, uint256 tokenId) internal view virtual returns (bool) {
-        address owner = ownerOf(tokenId);
-        return (spender == owner || isApprovedForAll(owner, spender) || getApproved(tokenId) == spender);
+    function _isApproved(address owner, address spender, uint256 tokenId) internal view virtual returns (bool) {
+        return owner == spender || isApprovedForAll(owner, spender) || getApproved(tokenId) == spender;
+    }
+
+    function _checkApproved(address owner, address spender, uint256 tokenId) internal view virtual {
```
ernestognw, line 233: NatSpec required here.

contracts/token/ERC721/ERC721.sol
ernestognw, line 325: NatSpec is missing

contracts/token/ERC721/ERC721.sol
```diff
@@ -194,15 +189,18 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
     }
 
     /**
-     * @dev Returns the owner of the `tokenId`. Does NOT revert if token doesn't exist
+     * @dev Returns the owner of the `tokenId`. Returns `address(0)` if token doesn't exist.
+     *
+     * IMPORTANT: Any overrides to this function that add ownership of tokens not tracked by the
+     * core ERC721 logic MUST be matched with the use of {_updateBalance} to keep balances
+     * consistent with ownership. The invariant being that for any address `a` the value returned by
+     * `balanceOf(a)` must be equal to the number of tokens such that `_ownerOf(tokenId)` is `a`.
```
ernestognw, line 203: There's no prior mention of the invariant described at the end. I think this connects better but feel free to propose something different.

```suggestion
     * IMPORTANT: Any overrides to this function that add ownership of tokens not tracked by the
     * core ERC721 logic MUST be matched with the use of {_updateBalance} to keep balances
     * consistent with ownership. The invariant to preserve is that for any address `a` the value returned by
     * `balanceOf(a)` must be equal to the number of tokens such that `_ownerOf(tokenId)` is `a`.
```

contracts/token/ERC721/extensions/ERC721Pausable.sol
```diff
@@ -27,14 +27,12 @@ abstract contract ERC721Pausable is ERC721, Pausable {
...
+    function _update(
         address from,
         address to,
-        uint256 firstTokenId,
-        uint256 batchSize
+        uint256 tokenId
     ) internal virtual override {
-        super._beforeTokenTransfer(from, to, firstTokenId, batchSize);
-
         _requireNotPaused();
```
ernestognw, line 35: I think it's a good moment to use `whenNotPaused` instead as suggested [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4261#discussion_r1212762822)

contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -64,15 +64,17 @@ abstract contract ERC721URIStorage is IERC4906, ERC721 {
...
+    function _update(address from, address to, uint256 tokenId) internal virtual override {
+        super._update(from, to, tokenId);
 
-        if (bytes(_tokenURIs[tokenId]).length != 0) {
-            delete _tokenURIs[tokenId];
+        if (to == address(0)) {
+            if (bytes(_tokenURIs[tokenId]).length != 0) {
+                delete _tokenURIs[tokenId];
+            }
         }
```
ernestognw, line 78:
```suggestion
        if (to == address(0) && bytes(_tokenURIs[tokenId]).length != 0) {
            delete _tokenURIs[tokenId];
        }
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,41 +312,51 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
         _owners[tokenId] = to;
 
         emit Transfer(from, to, tokenId);
+    }
 
-        _afterTokenTransfer(from, to, tokenId, 1);
+    /**
+     * @dev Unsafe write access to the balances, used by extensions that "mint" tokens using an {ownerOf} override.
+     */
+    function _updateBalance(address account, uint128 value) internal virtual {
```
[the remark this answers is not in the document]
frangio: The natspec is outdated. This function should not be described as unsafe. It will never lead to overflow

contracts/token/ERC721/ERC721.sol
```diff
@@ -343,41 +312,51 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
-        // Clear approvals from the previous owner
-        delete _tokenApprovals[tokenId];
-
-        // Decrease balance with checked arithmetic, because an `ownerOf` override may
-        // invalidate the assumption that `_balances[from] >= 1`.
-        _balances[from] -= 1;
+        if (from != address(0)) {
+            delete _tokenApprovals[tokenId];
+            unchecked {
+                _balances[from] -= 1;
```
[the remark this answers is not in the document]
frangio: This is a matter of preference but I am firmly in the `-= 1` camp... I don't know about you.

contracts/token/ERC721/extensions/ERC721Pausable.sol
frangio: Personally I prefer the function call. They are functionally equivalent but the function call is visible inside the function, I just find it more explicit. I don't know if everyone feels the same though.

ernestognw: I prefer `whenNotPaused` precisely because it's less explicit (== less pervasive?).

I'm fine with the function call if we agree to keep consistency between this and [ERC1155Pausable](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/bb64458928d08759e1f36cc0c11a8e76c976de0c/contracts/token/ERC1155/extensions/ERC1155Pausable.sol) where we use the modifier instead.

contracts/interfaces/draft-IERC6093.sol
```diff
@@ -69,11 +69,11 @@ interface IERC721Errors {
...
     /**
      * @dev Indicates an error related to the ownership over a particular token. Used in transfers.
-     * @param sender Address whose tokens are being transferred.
+     * @param claimedOwner Address whose tokens are being transferred.
      * @param tokenId Identifier number of a token.
-     * @param owner Address of the current owner of a token.
+     * @param actualOwner Address of the current owner of a token.
      */
-    error ERC721IncorrectOwner(address sender, uint256 tokenId, address owner);
+    error ERC721IncorrectOwner(address claimedOwner, uint256 tokenId, address actualOwner);
```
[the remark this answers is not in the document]
frangio: Possibly. I made this change because I found the existing names confusing.

Amxx:
This is an internal change to the data structure that does not affect the library interface. Does it need a changeset ?

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: Let's keep `-= 1` then

contracts/token/ERC721/ERC721.sol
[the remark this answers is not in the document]
ernestognw: > I don't understand why updateBalance would sound like it's a getter. 

My bad, I meant it sounds like a **setter** instead, not a value increase.

> increaseBalance is good too, but I thought the parallel with _update made sense.

Makes sense as long as the NatSpec correctly reflects what it does. I changed it for `_increaseBalance` because I'm strongly in favor of that name. We can discuss it.

contracts/interfaces/draft-IERC6093.sol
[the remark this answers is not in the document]
ernestognw: I see how the existing names are confusing, but I'd look for a way in which the naming makes sense for all errors. Would it be clear if we rename `sender` completely to `from`?

contracts/token/ERC721/ERC721.sol
```diff
@@ -151,28 +152,23 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
     /**
      * @dev See {IERC721-transferFrom}.
      */
-    function transferFrom(address from, address to, uint256 tokenId) public virtual {
-        if (!_isApprovedOrOwner(_msgSender(), tokenId)) {
-            revert ERC721InsufficientApproval(_msgSender(), tokenId);
-        }
-
+    function transferFrom(address from, address to, uint256 tokenId) public {
```
Amxx, line 155: Why are we removing virtual from these function ?

frangio: Same reason for `_transfer` but I suppose the public functions have other considerations. Do you think they should remain virtual?

Amxx: yes.

Our guidelines are to make (almost) everything virtual. I'm ok to making `_transfer`, `_mint` and `_burn` non virtual, because they can be bypassed by calling `_update` directly. But apart from these 3, everything else should remain virtual IMO.

frangio:
Closing in favor of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377.

PR #4420, opened 2023-07-04
Add support for typehash and beneficiary specific nonces

### LINKED ISSUE #3927: New abstractions for better nonce system
Lets discuss

- #3850
- [Out-of-order execution through "timelines"](https://github.com/amxx/permit#out-of-order-execution)

Amxx:
Fixes #3927

Proposal for typehash nonces / operation ids. This enables "per application" parallel nonces, meaning that the parallelism is explicitly controlled by the verifier [unlike what I proposed here](https://github.com/amxx/permit#out-of-order-execution). 

Still, I believe it would still be possible to build by-design out-of-order nonce mechanism on top of this PR.
```solidity
    function _verifyAndConsumeNonce(address owner, uint256 nonce) internal virtual {
        _useCheckedNonce(DEFAULT_TYPEHASH, owner, nonce >> 128, nonce % (1 << 128));
    }
```

- The default nonce is an alias to `DEFAULT_TYPEHASH = bytes32(0)` (this value is open to discussion). ERC20Permit and ERC20Votes will hook into that, preserving the current behavior.
- I made the opinionated choice that the operation nounces are operation ids with beneficiary = 0

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Nonces.sol
+++ b/contracts/utils/Nonces.sol
@@ -1,22 +1,74 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.19;
 
+import {Context} from "./Context.sol";
+
 /**
  * @dev Provides tracking nonces for addresses. Nonces will only increment.
  */
-abstract contract Nonces {
+ abstract contract Nonces is Context {
+    bytes32 internal constant DEFAULT_TYPEHASH = 0;
+
     /**
      * @dev The nonce used for an `account` is not the expected current nonce.
      */
     error InvalidAccountNonce(address account, uint256 currentNonce);
 
-    mapping(address => uint256) private _nonces;
+    error InvalidFastForward(address account, uint256 currentNonce);
+
+    // signer → typehash → beneficiary (track) → nonce
+    mapping(address => mapping(bytes32 => mapping(bytes32 => uint256))) private _nonces;
 
     /**
-     * @dev Returns an the next unused nonce for an address.
+     * @dev Returns the next unused nonce for an address.
      */
     function nonces(address owner) public view virtual returns (uint256) {
-        return _nonces[owner];
+        return operationIds(DEFAULT_TYPEHASH, owner, 0);
+    }
+
+    /**
+     * @dev Returns the next unused nonce for an address and a typehash.
+     */
+    function operationNonces(bytes32 typehash, address signer) public view virtual returns (uint256) {
+        return operationIds(typehash, signer, 0);
+    }
+
+    /**
+     * @dev Returns the next unused nonce for an address, a typehash and a beneficiary.
+     */
+    function operationIds(bytes32 typehash, address signer, bytes32 beneficiary) public view virtual returns (uint256) {
+        return _nonces[signer][typehash][beneficiary];
+    }
+
+    /**
+     * @dev Invalidate a chunk of nonces, up to `last` for the calling account and the specified typehash. After this
+     * call is executed, the next unused nonce for that account and that typehash will be `last + 1`.
+     *
+     * Requirements:
+     * - last must be at least the current nonce.
+     * - last must not invalidate more than 5000 nonces at once.
+     */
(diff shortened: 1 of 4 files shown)


contracts/utils/cryptography/EIP712.sol
```diff
@@ -32,7 +33,7 @@ import {IERC5267} from "../../interfaces/IERC5267.sol";
  *
  * @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
  */
-abstract contract EIP712 is IERC5267 {
```
Amxx, line 35: This is required for contract linearization.

contracts/utils/Nonces.sol
```diff
@@ -1,22 +1,74 @@
...
      */
     function nonces(address owner) public view virtual returns (uint256) {
-        return _nonces[owner];
+        return operationIds(DEFAULT_TYPEHASH, owner, 0);
+    }
+
+    /**
+     * @dev Returns the next unused nonce for an address and a typehash.
+     */
+    function operationNonces(bytes32 typehash, address signer) public view virtual returns (uint256) {
```
frangio, line 32: `nonces(owner)` comes from EIP-2616 but is a really bad name.

This one should be something like `getUnusedOperationNonce`?

contracts/utils/Nonces.sol
```diff
@@ -1,22 +1,74 @@
...
+
+    /**
+     * @dev Invalidate a chunk of nonces, up to `last` for the calling account and the specified typehash. After this
+     * call is executed, the next unused nonce for that account and that typehash will be `last + 1`.
+     *
+     * Requirements:
+     * - last must be at least the current nonce.
+     * - last must not invalidate more than 5000 nonces at once.
+     */
+    function useOperationNonce(bytes32 typehash, uint256 last) public virtual {
```
frangio, line 51: I think I would remove this one, `useOperationIds` should be enough. Having many aliases/variations pollutes the interface.

contracts/utils/Nonces.sol
frangio, line 1: Hm, we have to consider that this will add a lot of new functions to the token interfaces. Do we want to do that for ERC20Permit when it's not necessary? Starting to think that this should just be a separate feature.

contracts/utils/Nonces.sol
```diff
@@ -1,22 +1,74 @@
...
+     * @dev Returns the next unused nonce for an address and a typehash.
+     */
+    function operationNonces(bytes32 typehash, address signer) public view virtual returns (uint256) {
+        return operationIds(typehash, signer, 0);
+    }
+
+    /**
+     * @dev Returns the next unused nonce for an address, a typehash and a beneficiary.
+     */
+    function operationIds(bytes32 typehash, address signer, bytes32 beneficiary) public view virtual returns (uint256) {
```
frangio, line 39: "Beneficiary" is not the right term, we should probably use "track"?

contracts/utils/Nonces.sol
Amxx: operationNonces (and other) are names defined in the 6077 proposal. We should probably take some time to discuss all the naming.

contracts/utils/Nonces.sol
Amxx: Again, name comes from 6077. I like "track"

contracts/utils/Nonces.sol
frangio: > Would `Nonce` contain the internal getters and setters, and the "extension" the public ones ?

Yes this makes sense.

frangio: > Would `Nonce` contain the internal getters and setters, and the "extension" the public ones ?

Yes this makes sense.

frangio:
Related new issue: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4425

PR #4421, opened 2023-07-04
Add back IGovernor to docs

frangio:
In #4358 this was removed from the docs, but it should remain given that all over the Govenor docs it says "See IGovernor.xyz".

Perhaps we can take this opportunity to review how we want the documentation to read though, because this is not a nice reading experience. It feels like the docs should be there when you read the Governor interface, without needing to go through indirections. Should we change things in some way?

### DIFF
--- a/contracts/governance/README.adoc
+++ b/contracts/governance/README.adoc
@@ -52,6 +52,8 @@ NOTE: Functions of the `Governor` contract do not include access control. If you
 
 === Core
 
+{{IGovernor}}
+
 {{Governor}}
 
 === Modules


ernestognw:
Can you elaborate on the objectives of having a "better reading experience"? We have multiple ways of reading the docs, particularly the "See {IGovernor-something}" is for the docs website. If that's the case we can use `@inheritdoc` so the docs are there without the indirection.

frangio:
My comment was mainly about the website. We should also care about the reading experience when looking at the code, but there's not much we can do there, if there is an interface we will want the docs to be in the interface. In some cases where the separation between interface and contract may be unnecessary, such as IAccessControl and AccessControl, we could improve the experience by removing IAccessControl and putting docs directly in AccessControl.

Yes we could use inheritdoc, in fact there's a PR but it had an issue: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3505

ernestognw:
> My comment was mainly about the website. We should also care about the reading experience when looking at the code, but there's not much we can do there, if there is an interface we will want the docs to be in the interface. In some cases where the separation between interface and contract may be unnecessary, such as IAccessControl and AccessControl, we could improve the experience by removing IAccessControl and putting docs directly in AccessControl.
> 
> Yes we could use inheritdoc, in fact there's a PR but it had an issue: #3505

I see the linked PR didn't show the docs, but I also don't understand what was the intention because I can't see any @inheritdoc added within the PR.

Remember we already have `@inheritdoc`s in `AccessControlDefaultAdminRules` and `ERC2981` and they work fine afaik.

frangio:
>I can't see any @inheritdoc added within the PR.

True, the thing is if there is not NatSpec the `@inheritdoc` is implicit.

Amxx (approved):

ernestognw (approved):
Ok, I see there's an issue for the `@inheritdoc` discussion [here](https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3502), so I'm approving just to move on.

PR #4423, opened 2023-07-05
Remove retyped and renamed storage layout annotations

frangio:
Removes all `@custom:oz-retyped-from` and `@custom:oz-renamed-from` annotations for storage layout checks from OpenZeppelin Upgrades Plugins.

Also removes the `@custom:storage-size` annotations for the transpiler, all of them found in `upgradeable.patch`.

The changes to storage layout need to be documented in the changelog but the storage-size change is the least of changes given the upcoming change to namespaced storage.

Fixes LIB-932

### DIFF
--- a/scripts/upgradeable/upgradeable.patch
+++ b/scripts/upgradeable/upgradeable.patch
@@ -105,32 +105,6 @@ index 27627f43..e42a66e8 100644
      }
  }
  ```
-diff --git a/contracts/finance/VestingWallet.sol b/contracts/finance/VestingWallet.sol
-index f776a7ca..d7a7e0b0 100644
---- a/contracts/finance/VestingWallet.sol
-+++ b/contracts/finance/VestingWallet.sol
-@@ -19,6 +19,8 @@ import {Context} from "../utils/Context.sol";
-  *
-  * By setting the duration to 0, one can configure this contract to behave like an asset timelock that hold tokens for
-  * a beneficiary until a specified time.
-+ *
-+ * @custom:storage-size 52
-  */
- contract VestingWallet is Context {
-     event EtherReleased(uint256 amount);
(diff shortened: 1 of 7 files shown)


scripts/upgradeable/upgradeable.patch
```diff
@@ -202,12 +137,12 @@ index d94e956a..86bb5713 100644
...
  
  /**
-@@ -30,27 +29,19 @@ import {IERC5267} from "../../interfaces/IERC5267.sol";
+@@ -29,28 +28,18 @@ import {IERC5267} from "../../interfaces/IERC5267.sol";
+  * separator from the immutable values, which is cheaper than accessing a cached version in cold storage.
   *
   * _Available since v3.4._
-  *
+- *
 - * @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
```
Amxx, line 145: The patch removes this from the vanilla version ... but if we don't want it in the upgradeable version, why is it in the vanilla version in the first place ?

ernestognw: I think this is a mistake, we indeed want the annotation as per [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3969#issuecomment-1420145285). We may have just not ported it to upgradeable contracts

frangio: This comment is duplicate, discussing in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4424#discussion_r1253074417

frangio:
Note: The concern about EIP712 was resolved in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4424#discussion_r1253074417

ernestognw (approved):
LGTM

PR #4424, opened 2023-07-05
Remove "available since" comments

frangio:
The documentation includes comments saying "Available since vX.Y" meant to indicate that a feature was introduced in a minor, because the docs are only versioned by major. We don't need this since we're moving to 5.0.

~Blocked by https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4423 because the changes to `upgradeable.patch` conflict.~

Fixes LIB-948


contracts/token/ERC1155/utils/ERC1155Receiver.sol
```diff
@@ -7,7 +7,9 @@ import {IERC1155Receiver} from "../IERC1155Receiver.sol";
 import {IERC165, ERC165} from "../../../utils/introspection/ERC165.sol";
 
 /**
- * @dev _Available since v3.1._
+ * @dev Basic contract implementing the ERC-165 interface for {IERC1155Receiver}.
+ *
+ * NOTE: This contract does not suffice to receive tokens. See {ERC1155Holder}.
  */
 abstract contract ERC1155Receiver is ERC165, IERC1155Receiver {
```
frangio, line 14: I don't really understand why we have separate ERC1155Receiver and ERC1155Holder contracts. I think we should merge them, and only keep ERC1155Holder.

scripts/upgradeable/upgradeable.patch
```diff
@@ -202,12 +137,12 @@ index d94e956a..86bb5713 100644
...
-@@ -30,27 +29,19 @@ import {IERC5267} from "../../interfaces/IERC5267.sol";
-  *
-  * _Available since v3.4._
-  *
+@@ -27,28 +26,18 @@ import {IERC5267} from "../../interfaces/IERC5267.sol";
+  * NOTE: In the upgradeable version of this contract, the cached values will correspond to the address, and the domain
+  * separator of the implementation contract. This will cause the `_domainSeparatorV4` function to always rebuild the
+  * separator from the immutable values, which is cheaper than accessing a cached version in cold storage.
+- *
 - * @custom:oz-upgrades-unsafe-allow state-variable-immutable state-variable-assignment
```
Amxx, line 145: This part of the patch is removing some natspec comments from the vanilla version.
It looks to me that these natspec comments are only usefull in an upgradeable context. So why do we have them in the vanilla version in the first place ?

frangio: The vanilla contract is upgradeable-safe even though it uses immutable variables. If we keep these annotations, the transpiler will keep the immutable variables instead of converting them. That said I think the patch removes all immutable variables anyway.

As a side note, I'm realizing that using `unsafe-allow state-variable-assignment` in this way is bad because we only want it to apply to state-variable-immutable and this will allow assignments to any storage variable.

Amxx: > The vanilla contract is upgradeable-safe.

Is that something we want the upgrade plugin to detect ? Using the vanila EIP712 as an implementation for proxy is dangerous.

frangio: >Using the vanilla EIP712 as an implementation for proxy is dangerous.

Is it? I think the annotation is there because it is not dangerous.

Amxx: You're right. I forgot about the `address(this) == _cachedThis` check. Its indeed safe.

ernestognw (approved):
There's a missing import to remove from `scripts/generate/templates/SafeCast.js` that might require another commit but LGTM, let me know if you need reapproving

ernestognw (approved):

Issue #4318, opened 2023-06-06
`EIP712`: `_TYPE_HASH` including `salt`


pcaversaccio:
Since `EIP-5267` got merged I wanted to raise a question. In [`EIP-712`](https://eips.ethereum.org/EIPS/eip-712)'s definition of the `domainSeparator` a so-called salt is also listed:

> `bytes32 salt` an disambiguating salt for the protocol. This can be used as a domain separator of last resort.

`EIP-5267` also supports the `salt` parameters for retrieval via the `fields` parameter. Now, my question: shouldn't we adjust the [`EIP712`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol) to incorporate this `salt` parameter?

I.e.:

```diff
- bytes32 private constant _TYPE_HASH = keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
+ - bytes32 private constant _TYPE_HASH = keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract,bytes32 salt)");
```

and

```solidity
function _buildDomainSeparator(bytes32 salt) private view returns (bytes32) {
    return keccak256(abi.encode(_TYPE_HASH, _hashedName, _hashedVersion, block.chainid, address(this), salt));
}
```

Amxx:
Hello @pcaversaccio 

IMO most cases don't need the salt. In fact, the only use of salt I've seen in production is in some polygon contract that used chainId as the salt (and did not include chainId).

I wouldn't want to force that onto any user, particularly without a good documentation explaining when/why use the salt.

On the other hand, it would be better if our implementation could allow someone that want to include a salt to do so.

pcaversaccio:
> On the other hand, it would be better if our implementation could allow someone that want to include a salt to do so.

I think OpenZeppelin is optimising for security, correctness and completeness (to a certain extend). Thus, I definitely believe OZ as the most common Solidity smart contract library should allow for such a customisation if someone would like to include the salt.

Amxx:
We could have a version that does

```solidity
    function _buildDomainSeparator() private view returns (bytes32) {
        return _salt == bytes32(0)
            ? keccak256(abi.encode(_TYPE_HASH,           _hashedName, _hashedVersion, block.chainid, address(this)))
[...]

PR #4430, opened 2023-07-05
Move `ECDSA` message hash methods to its own `MessageHashUtils` library

ernestognw:
As part of #4318, we agreed that the `ECDSA` library can just focus on `recovery` methods, whereas a new library for ECDSA message hashes would be a suitable place for adding support to new message schemes.

It includes some updates to the documentation.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/hot-dingos-kiss.md
+++ b/.changeset/hot-dingos-kiss.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`MessageHashUtils`: Add a new library for creating message digest to be used along with ECDSA signing or recovery. These functions are moved from the `ECDSA` library.
--- a/.changeset/hot-dingos-kiss.md
+++ b/.changeset/hot-dingos-kiss.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': major
 ---
 
-`MessageHashUtils`: Add a new library for creating message digest to be used along with ECDSA signing or recovery. These functions are moved from the `ECDSA` library.
+`MessageHashUtils`: Add a new library for creating message digest to be used along with signing or recovery such as  ECDSA or ERC-1271. These functions are moved from the `ECDSA` library.
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -3,8 +3,6 @@
 
 pragma solidity ^0.8.19;
 
-import {Strings} from "../Strings.sol";
-
 /**
  * @dev Elliptic Curve Digital Signature Algorithm (ECDSA) operations.
  *
@@ -34,18 +32,6 @@ library ECDSA {
      */
     error ECDSAInvalidSignatureS(bytes32 s);
 
-    function _throwError(RecoverError error, bytes32 errorArg) private pure {
-        if (error == RecoverError.NoError) {
-            return; // no error: do nothing
-        } else if (error == RecoverError.InvalidSignature) {
-            revert ECDSAInvalidSignature();
-        } else if (error == RecoverError.InvalidSignatureLength) {
-            revert ECDSAInvalidSignatureLength(uint256(errorArg));
-        } else if (error == RecoverError.InvalidSignatureS) {
-            revert ECDSAInvalidSignatureS(errorArg);
-        }
-    }
-
     /**
      * @dev Returns the address that signed a hashed message (`hash`) with
      * `signature` or error string. This address can then be used for verification purposes.
@@ -58,7 +44,7 @@ library ECDSA {
      * verification to be secure: it is possible to craft signatures that
      * recover to arbitrary addresses for non-hashed data. A safe way to ensure
      * this is by receiving a hash of the original message (which may otherwise
-     * be too long), and then calling {toEthSignedMessageHash} on it.
+     * be too long), and then calling {MessageHashUtils-toEthSignedMessageHash} on it.
      *
      * Documentation for signature generation:
      * - with https://web3js.readthedocs.io/en/v1.3.4/web3-eth-accounts.html#sign[Web3.js]
@@ -97,7 +83,7 @@ library ECDSA {
      * verification to be secure: it is possible to craft signatures that
      * recover to arbitrary addresses for non-hashed data. A safe way to ensure
      * this is by receiving a hash of the original message (which may otherwise
-     * be too long), and then calling {toEthSignedMessageHash} on it.
+     * be too long), and then calling {MessageHashUtils-toEthSignedMessageHash} on it.
      */
     function recover(bytes32 hash, bytes memory signature) internal pure returns (address) {
         (address recovered, RecoverError error, bytes32 errorArg) = tryRecover(hash, signature);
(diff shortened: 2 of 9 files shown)


ernestognw:
Note that the `MessageHashUtils` name is not yet agreed but one of the options.

Some other options could be:

- `library DigestUtils {}`
- `library SignatureDigestUtils {}`
- `library MessageHashUtils {}`
- `library SignatureHashUtils {}`
- `library MessageDigestUtils {}`

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -177,63 +163,17 @@ library ECDSA {
...
-     */
-    function toTypedDataHash(bytes32 domainSeparator, bytes32 structHash) internal pure returns (bytes32 data) {
-        /// @solidity memory-safe-assembly
-        assembly {
-            let ptr := mload(0x40)
-            mstore(ptr, hex"19_01")
-            mstore(add(ptr, 0x02), domainSeparator)
-            mstore(add(ptr, 0x22), structHash)
-            data := keccak256(ptr, 0x42)
+    function _throwError(RecoverError error, bytes32 errorArg) private pure {
```
ernestognw, line 168: I moved this function at the end because it's the first thing one reads when opening the ECDSA library, this is a personal opinion but I think it's reasonable to keep at the end because it's private anyway and it's name it's very self-descriptive

scripts/upgradeable/upgradeable.patch
Amxx, line 1: Files to this file doesn't feel right. Can you double check ?

test/utils/cryptography/MessageHashUtils.test.js
```diff
@@ -0,0 +1,57 @@
...
+      expect(await this.messageHashUtils.methods['$toEthSignedMessageHash(bytes32)'](this.messageHash)).to.equal(
+        toEthSignedMessageHash(this.messageHash),
+      );
+    });
+
+    it('prefixes dynamic length data correctly', async function () {
+      const message = '0x' + Buffer.from('abcd').toString('hex');
+      expect(await this.messageHashUtils.methods['$toEthSignedMessageHash(bytes)'](message)).to.equal(
+        toEthSignedMessageHash(message),
+      );
```
Amxx, line 30: Message is already declared in the `beforeEach`
```suggestion
      expect(await this.messageHashUtils.methods['$toEthSignedMessageHash(bytes)'](this.message)).to.equal(
        toEthSignedMessageHash(this.message),
      );
```

test/utils/cryptography/MessageHashUtils.test.js
```diff
@@ -0,0 +1,57 @@
...
+      ).to.equal(toDataWithIntendedValidatorHash(this.verifyingAddress, this.message));
+    });
+  });
+
+  context('toTypedDataHash', function () {
+    it('returns the digest correctly', async function () {
+      const domain = {
+        name: 'Test',
+        version: 1,
+        chainId: await getChainId(),
```
Amxx, line 47: Here it doesn't have to be the actual chainId, since it won't be part of the reconstruction on chain (we pass the domain separator directly). We could hardcode any value here (or use a random value)

Amxx (commented):
- Minor comments on the test
- The patch file doesn't feel right

otherwize its good

test/utils/cryptography/MessageHashUtils.test.js
ernestognw: Good observation, thanks. I like the idea of a random chainId but I don't like randomness in tests because it makes the tests non-deterministic. Although I don't expect it to fail here based on that randomness, I hardcoded it to 1 anyway.

Let me know if you have a strong opinion. I'm solving for now because I think it's addressed.

scripts/upgradeable/upgradeable.patch
ernestognw: I merge master with:

```bash
git merge master -X theirs
```

Then applied the patch with `bash scripts/upgradeable/patch-apply.sh`, solved conflicts (`ECDSA` vs `MessageHashUtils` import in EIP-712), and saved the patch with `bash scripts/upgradeable/patch-save.sh`.

Should be fine but I did the same conflict resolution before, let me know if I'm missing something.

Amxx (dismissed):

.changeset/hot-dingos-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`MessageHashUtils`: Add a new library for creating message digest to be used along with ECDSA signing or recovery. These functions are moved from the `ECDSA` library.
```
frangio, line 5:
```suggestion
`MessageHashUtils`: Add a new library for creating message digest to be used along with signing or recovery such as  ECDSA or ERC-1271. These functions are moved from the `ECDSA` library.
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -22,7 +23,7 @@ function _verify(bytes32 data, bytes memory signature, address account) internal
 }
 ----
 
-WARNING: Getting signature verification right is not trivial: make sure you fully read and understand xref:api:utils.adoc#ECDSA[`ECDSA`]'s documentation.
+WARNING: Getting signature verification right is not trivial: make sure you fully read and understand xref:api:utils.adoc#MessageHashUtils[`MessageHashUtils`]'s and xref:api:utils.adoc#ECDSA[`ECDSA`]'s documentations.
```
frangio, line 26:
```suggestion
WARNING: Getting signature verification right is not trivial: make sure you fully read and understand xref:api:utils.adoc#MessageHashUtils[`MessageHashUtils`]'s and xref:api:utils.adoc#ECDSA[`ECDSA`]'s documentation.
```

scripts/upgradeable/upgradeable.patch
frangio, line 1: The patch looks good to me. The changes I see are to commit hashes and line numbers, as well as context that has changed.

frangio (approved):

Amxx (approved):

contracts/utils/cryptography/MessageHashUtils.sol
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @dev Returns the keccak256 digest of an EIP-191 signed data with version
+     * `0x45` (`personal_sign` messages).
+     *
+     * The digest is calculated by prefixing a bytes32 `messageHash` with
+     * `"\x19Ethereum Signed Message:\n32"` and hashing the result. It corresponds with the
+     * hash signed when using the https://eth.wiki/json-rpc/API#eth_sign[`eth_sign`] JSON-RPC method.
+     *
+     * NOTE: The `hash` parameter is intended to be the result of hashing a raw message with
+     * keccak256, althoguh any bytes32 value can be safely used because the final digest will
```
pcaversaccio, line 24: "althoguh" -> typo; I think for such a fix no need for a PR overhead and one of you maintainers can quickly fix it in `main` ;-): cc: @ernestognw, @frangio, @Amxx

pcaversaccio: Fixed via https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4462 (thx @Amxx)

PR #4288, opened 2023-05-29
Bump minimum pragma version to `0.8.19`

ernestognw:
#### PR Checklist

Bump the minimum pragma version to 0.8.19

- [x] Tests
- [x] Documentation
- [ ] Changeset entry (run `npx changeset add`)

PR #4433, opened 2023-07-06
Use named arguments in mapping types

### LINKED ISSUE #4343: Implement named parameters in mapping types
**🧐 Motivation**

Solidity version [`0.8.18`](https://github.com/ethereum/solidity/releases/tag/v0.8.18) introduced the possibility to use named parameters in mapping types. Since the minimum pragma version was bumped to Solidity version `0.8.19` via https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4288, I think it would make sense to implement this new feature to increase the overall code readability.

**📝 Details**

OZ contracts make extensive use of mappings, and this could be definitely a benefit:

```solidity
mapping(address owner => mapping(address spender => uint256 amount)) private _allowances;
```

ernestognw:
Fixes #4343 

After bumping pragma to 0.8.19 in #4288, we're now able to add named arguments to mapping types. This PR solves that.

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/access/AccessControl.sol
```diff
@@ -49,11 +49,11 @@ import {ERC165} from "../utils/introspection/ERC165.sol";
  */
 abstract contract AccessControl is Context, IAccessControl, ERC165 {
     struct RoleData {
-        mapping(address => bool) members;
+        mapping(address account => bool member) members;
```
Amxx, line 52:
```suggestion
        mapping(address account => bool isMember) members;
```

Amxx: or "membership" ?

contracts/mocks/compound/CompTimelock.sol
```diff
@@ -63,7 +63,7 @@ contract CompTimelock {
     address public pendingAdmin;
     uint256 public delay;
 
-    mapping(bytes32 => bool) public queuedTransactions;
+    mapping(bytes32 txHash => bool queued) public queuedTransactions;
```
Amxx, line 66: That is a mock, so I don't really care, bu I think we should not use txHash for anything that is not an actual txHash

contracts/token/ERC1155/extensions/ERC1155URIStorage.sol
```diff
@@ -17,7 +17,7 @@ abstract contract ERC1155URIStorage is ERC1155 {
     string private _baseURI = "";
 
     // Optional mapping for token URIs
-    mapping(uint256 => string) private _tokenURIs;
+    mapping(uint256 tokenId => string uri) private _tokenURIs;
```
Amxx, line 20: Its not really the uri, its the "uriExtension" that get concatenated with the `_baseURI`. Not sure if we want to reflect that.

contracts/token/ERC721/ERC721.sol
```diff
@@ -25,17 +25,13 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
     // Token symbol
     string private _symbol;
 
-    // Mapping from token ID to owner address
-    mapping(uint256 => address) private _owners;
+    mapping(uint256 tokenId => address owner) private _owners;
 
-    // Mapping owner address to token count
-    mapping(address => uint256) private _balances;
+    mapping(address owner => uint256 tokenCount) private _balances;
```
Amxx, line 30: why not simply balance ?

contracts/token/ERC721/extensions/ERC721URIStorage.sol
```diff
@@ -15,7 +15,7 @@ abstract contract ERC721URIStorage is IERC4906, ERC721 {
     using Strings for uint256;
 
     // Optional mapping for token URIs
-    mapping(uint256 => string) private _tokenURIs;
+    mapping(uint256 tokenId => string uri) private _tokenURIs;
```
Amxx, line 18: Same here, that is an extension. 

I'm also nothing that the ERC1155 implementation is cleaner, maybe we want to reproduce it here.

contracts/access/AccessControl.sol
ernestognw: Let's go with `isMember`

contracts/mocks/compound/CompTimelock.sol
ernestognw: I renamed this to `operationHash` instead because I think it reflects better, but let me know if we need to rollback it or you have another proposal.

contracts/token/ERC721/ERC721.sol
ernestognw: I just didn't want to be redundant, I changed it to balance.

contracts/token/ERC1155/extensions/ERC1155URIStorage.sol
ernestognw: I'd leave it as it is right now because I see we're also calling `tokenUri` to the variable itself and even when we concatenate with the `baseUri`. I agree with renaming them all but will prefer doing in another PR.

contracts/token/ERC721/extensions/ERC721URIStorage.sol
ernestognw: Same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4433#discussion_r1256692650

contracts/mocks/compound/CompTimelock.sol
```diff
@@ -31,23 +31,23 @@ contract CompTimelock {
     event NewPendingAdmin(address indexed newPendingAdmin);
     event NewDelay(uint256 indexed newDelay);
     event CancelTransaction(
-        bytes32 indexed txHash,
+        bytes32 indexed operationHash,
```
Amxx, line 34: Sorry for comming back to it, but actuelly this is a contract we vendor from Coumpound. Its in the mocks because we only use it for tests, but it should probably be in `contracts/vendor/compound`. 

I'm not sure we should touch these vendored contract at all.

Amxx: The difference between this and `contracts/vendor/compound` is that this doesn't end up in the npm package.

Amxx: @frangio WDYT ?

frangio: Agree the vendored contracts shouldn't be changed. They don't need to follow our conventions. In fact we could ignore them from linter rules for example.

ernestognw: I made the change just for consistency without further context on this contract and I initially considered it better to be consistent with the named mapping arguments but I agree this is not required.

I reverted back to the original version. Solving.

Amxx (dismissed):

RenanSouza2:
just a reminder to link this PR to issue #4343

ernestognw:
> just a reminder to link this PR to issue #4343

Thanks for the heada up @RenanSouza2, just linked

Amxx (approved):

contracts/access/AccessControl.sol
```diff
@@ -48,11 +48,11 @@ import {ERC165} from "../utils/introspection/ERC165.sol";
  */
 abstract contract AccessControl is Context, IAccessControl, ERC165 {
     struct RoleData {
-        mapping(address => bool) members;
+        mapping(address account => bool isMember) members;
         bytes32 adminRole;
     }
 
-    mapping(bytes32 => RoleData) private _roles;
+    mapping(bytes32 role => RoleData data) private _roles;
```
frangio, line 55: Here I don't think the `data` name is necessary because it's implicit from the `RoleData` struct name.

contracts/governance/Governor.sol
```diff
@@ -47,7 +47,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
     bytes32 private constant _ALL_PROPOSAL_STATES_BITMAP = bytes32((2 ** (uint8(type(ProposalState).max) + 1)) - 1);
     string private _name;
 
-    mapping(uint256 => ProposalCore) private _proposals;
+    mapping(uint256 proposalId => ProposalCore proposal) private _proposals;
```
frangio, line 50: Same here.

contracts/governance/compatibility/GovernorCompatibilityBravo.sol
```diff
@@ -31,11 +31,11 @@ abstract contract GovernorCompatibilityBravo is IGovernorTimelock, IGovernorComp
         uint256 forVotes;
         uint256 againstVotes;
         uint256 abstainVotes;
-        mapping(address => Receipt) receipts;
+        mapping(address voter => Receipt receipt) receipts;
```
frangio, line 34: Here, etc. Please review the others too.

contracts/access/AccessControl.sol
RenanSouza2: If I may leave my opinion here, I don't think any of the last names are needed because they are a duplicate of the name of the whole mapping

frangio: Yeah I think that is probably generally true.

ernestognw: I'm ok with removing the `data` from `RoleData` but not sure about generalizing. Look at these cases:

```
// ERC20
mapping(address account => mapping(address spender => uint256 value)) private _allowances;

// AccessControl
```solidity
mapping(address account => bool isMember) members;
```

The ERC20 case is one where the mapping is nested. Because there's no name for the nested `address => value` mapping, it's no obvious.

In the AccessControl, the mapping name is not explicitly representative of what it is, so removing `isMember` would be detrimental imo

ernestognw: Just finishing reviewing and I think most cases are like @RenanSouza2 is describing, so I'd agree with also removing those names if we agree. Aside from cases like the ones I shared, the guideline would be to avoid naming the `value` side of the mapping if its name is equal or very similar to the mapping's own name.

frangio: I agree with removing.

ernestognw: Just removed, I renamed a few of them where I see it makes sense. Regarding my counterexamples, this is what I did:

```solidity
// Previously
// mapping(address account => mapping(address spender => uint256 value)) private _allowances;
mapping(address account => mapping(address spender => uint256)) private _allowances;

// Previously
// mapping(address account => bool isMember) members;
```

frangio (approved):
Thanks!

frangio:
Storage layout check failing because of renamed variables or struct members, as expected, but ok for 5.0.

PR #4434, opened 2023-07-07
Remove superfluous receive() function from Proxy.sol

Amxx:
[Following this discussion in the forum](https://forum.openzeppelin.com/t/proxy-sol-fallback/36951/7)

In the past, when declaring a `fallback()`, solidity would trigger a warning if no `receive()` is implemented. That is apparently no longer the case.

In the case of Proxy, the `receive` function is not necessary. Any call, with or without data, with or without value, will be catched by `fallback() fallback`.

#### PR Checklist

- [x] Tests (nothing changes)

### DIFF
--- a/.changeset/eight-peaches-guess.md
+++ b/.changeset/eight-peaches-guess.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Proxy`: Removed redundant `receive` function.
--- a/contracts/proxy/Proxy.sol
+++ b/contracts/proxy/Proxy.sol
@@ -69,14 +69,6 @@ abstract contract Proxy {
     }
 
     /**
-     * @dev Fallback function that delegates calls to the address returned by `_implementation()`. Will run if call data
-     * is empty.
-     */
-    receive() external payable virtual {
-        _fallback();
-    }
-
-    /**
      * @dev Hook that is called before falling back to the implementation. Can happen as part of a manual `_fallback`
      * call, or as part of the Solidity `fallback` or `receive` functions.
      *


ernestognw (approved):
I'm approving but, can we try a bit more to reproduce the supposed warning with no `receive` function? I'm afraid this won't trigger for us but maybe will do for others importing the library. If we can make sure the warning was removed after a certain Solidity version would be better.

Amxx:
Note: tests show that is saved 1500~1900 gas per proxy deployment (`ERC1967Proxy`, `TransparentUpgradeableProxy` & `BeaconProxy`)

frangio:
The warning wasn't removed. It's just really hard to reproduce. I couldn't figure out the conditions until I found the warning in the Solidity repo[^1][^2].

[^1]: https://github.com/ethereum/solidity/blob/v0.8.20/test/libsolidity/syntaxTests/fallback/payable_fallback_without_receive_nonempty.sol
[^2]: https://github.com/ethereum/solidity/blob/v0.8.20/test/libsolidity/syntaxTests/fallback/payable_fallback_without_receive_nonempty_by_inheritance.sol


```solidity
contract C {
    fallback() external payable { }
    function f() public pure { }
}
```

```solidity
contract A {
    function f() external pure {}
}
contract C is A {
    fallback() external payable { }
}
```

The warning in both cases is:
>This contract has a payable fallback function, but no receive ether function. Consider adding a receive ether function.

This probably used to happen in the transparent proxy because we had a fallback and additional functions (`upgradeTo`, etc.), but this is no longer the case and should never be the case for us or users, for the reasons we do the manual dispatch and recommend the same.

frangio (approved):

ernestognw (approved):
Ulalá, references.

Thanks @frangio!

PR #4435, opened 2023-07-07
Use immutable beacon address in BeaconProxy

### LINKED ISSUE #2538: Use immutable variable to store beacon address in BeaconProxy
**🧐 Motivation**
Unlike the transparent proxy, the `_BEACON_SLOT` that is part of the BeaconProxy design doesn't need to be updated to update the proxy. Its rather the opposite:
- Upgrades are to be performed in the beacon's storage space
- Changed to the `_BEACON_SLOT` could, if performed incorrectly by the implementation, break the proxy.

In some cases the ability to update the `_BEACON_SLOT` might be a welcomed feature... but in other cases, it might be a cost (an sload is required) and an eventual security issue.

**📝 Details**
For cases where upgradeability of the `_BEACON_SLOT` is not requiered, the beacon address could be stored using an immutable variable. This would provide a new `ImmutableBeaconProxy` implementation

```
contract ImmutableBeaconProxy is Proxy {
    IBeacon immutable private _beacon;

    constructor(address beacon, bytes memory data) payable {
        require(
            Address.isContract(beacon),
            "BeaconProxy: beacon is not a contract"
        );
        require(
            Address.isContract(IBeacon(beacon).implementation()),
            "BeaconProxy: beacon implementation is not a contract"
        );

        _beacon = IBeacon(beacon);

        if (data.length > 0) {
            // cannot call `_implementation()` because immutable variables cannot be read during construction.
            Address.functionDelegateCall(IBeacon(beacon).implementation(), data, "BeaconProxy: function call failed");
        }
    }

    function _implementation() internal view override returns (address) {
        return _beacon.implementation();
    }
}
```

ericglau:
Fixes #2538

Uses an immutable variable to store the beacon's address in BeaconProxy, in addition to the beacon storage slot.  This avoids an extra SLOAD when retrieving the beacon address before each delegate call.

The immutable variable is used in `_implementation()` to read the beacon address before each delegate call.
The beacon slot is still written during construction, so that the beacon address can be read externally according to ERC-1967.  This beacon slot is required because the immutable variable must be private to avoid generating a getter function that could clash with a function from the implementation.

Rationale:
- Although there is an `upgradeBeaconToAndCall` internal function in ERC1967Utils which sets the beacon slot, that function is not exposed in BeaconProxy but is only called on construction.  Effectively, this means the beacon address is already not modifiable after the BeaconProxy is deployed.
- Therefore, there is no need to continue reading the beacon slot before every delegate call during the proxy's fallback function. We can use an immutable variable instead.

#### PR Checklist

- [x] Tests (not applicable as this only changes the proxy's internal implementation)
- [x] Documentation (updated comments)

### DIFF
--- a/.changeset/proud-seals-complain.md
+++ b/.changeset/proud-seals-complain.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Use immutable beacon address in BeaconProxy
--- a/.changeset/proud-seals-complain.md
+++ b/.changeset/proud-seals-complain.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-Use immutable beacon address in BeaconProxy
+`BeaconProxy`: Use an immutable variable to store the address of the beacon. It is no longer possible for a `BeaconProxy` to upgrade by changing to another beacon.
--- a/contracts/proxy/ERC1967/ERC1967Utils.sol
+++ b/contracts/proxy/ERC1967/ERC1967Utils.sol
@@ -161,10 +161,13 @@ library ERC1967Utils {
     }
 
     /**
-     * @dev Perform beacon upgrade with additional setup call. Note: This upgrades the address of the beacon, it does
-     * not upgrade the implementation contained in the beacon (see {UpgradeableBeacon-_setImplementation} for that).
+     * @dev Change the beacon and trigger a setup call.
      *
      * Emits an {IERC1967-BeaconUpgraded} event.
+     *
+     * CAUTION: Invoking this function has no effect on an instance of {BeaconProxy} since v5, since
+     * it uses an immutable beacon without looking at the value of the ERC-1967 beacon slot for
+     * efficiency.
      */
     function upgradeBeaconToAndCall(address newBeacon, bytes memory data, bool forceCall) internal {
         _setBeacon(newBeacon);
--- a/contracts/proxy/ERC1967/ERC1967Utils.sol
+++ b/contracts/proxy/ERC1967/ERC1967Utils.sol
@@ -138,6 +138,13 @@ library ERC1967Utils {
     bytes32 internal constant BEACON_SLOT = 0xa3f0ad74e5423aebfd80d3ef4346578335a9a72aeaee59ff6cb3582b35133d50;
 
     /**
+     * @dev Returns the current beacon.
+     */
+    function getBeacon() internal view returns (address) {
+        return StorageSlot.getAddressSlot(BEACON_SLOT).value;
+    }
+
+    /**
      * @dev Stores a new beacon in the EIP1967 beacon slot.
      */
     function _setBeacon(address newBeacon) private {
--- a/contracts/proxy/ERC1967/ERC1967Utils.sol
+++ b/contracts/proxy/ERC1967/ERC1967Utils.sol
@@ -138,13 +138,6 @@ library ERC1967Utils {
     bytes32 internal constant BEACON_SLOT = 0xa3f0ad74e5423aebfd80d3ef4346578335a9a72aeaee59ff6cb3582b35133d50;
 
     /**
-     * @dev Returns the current beacon.
-     */
-    function getBeacon() internal view returns (address) {
-        return StorageSlot.getAddressSlot(BEACON_SLOT).value;
-    }
-
-    /**
      * @dev Stores a new beacon in the EIP1967 beacon slot.
      */
     function _setBeacon(address newBeacon) private {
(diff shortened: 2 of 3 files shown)


contracts/proxy/ERC1967/ERC1967Utils.sol
```diff
@@ -137,13 +137,6 @@ library ERC1967Utils {
     // solhint-disable-next-line private-vars-leading-underscore
     bytes32 internal constant BEACON_SLOT = 0xa3f0ad74e5423aebfd80d3ef4346578335a9a72aeaee59ff6cb3582b35133d50;
 
-    /**
-     * @dev Returns the current beacon.
-     */
-    function getBeacon() internal view returns (address) {
-        return StorageSlot.getAddressSlot(BEACON_SLOT).value;
-    }
-
```
Amxx, line 146: I'd keep that as a primitive if anyone wants to implement a storage based beacon.

contracts/proxy/beacon/BeaconProxy.sol
```diff
@@ -12,8 +12,15 @@ import {ERC1967Utils} from "../ERC1967/ERC1967Utils.sol";
  *
  * The beacon address is stored in storage slot `uint256(keccak256('eip1967.proxy.beacon')) - 1`, so that it doesn't
  * conflict with the storage layout of the implementation behind the proxy.
+ *
+ * CAUTION: The beacon address can only be set once during construction, and cannot be changed afterwards.
+ * You must ensure that you either control the beacon, or trust the beacon to not upgrade the implementation maliciously.
  */
 contract BeaconProxy is Proxy {
+    // An immutable address for the beacon to avoid unnecessary SLOADs before each delegate call.
+    // This must be private to avoid generating a getter function that could clash with a function from the implementation.
```
Amxx, line 21: it could be internal. should it ?

frangio: If we are keeping `getBeacon` as an internal function, the variable should be private so it is accessed through the getter.

Amxx: Which getter ?

frangio: `getBeacon`

frangio: I was confused about your other comment in this PR, I thought getBeacon was a getter in the BeaconProxy.

Amxx: Its two different thing:
- In the `ERC1967Utils` library there is a getter to read the storage slot, that I think we should keep is someone ever needs to read that (for example to re-implement a storage based `BeaconProxy`)
- In the `BeaconProxy` (with immutable storage) I think it would be nice to have a possible access to the immutable value, which can be done in a safe way by making the immutable variable internal.

frangio: If we defined an internal function instead of making the variable internal it would be possible to do the storage variant by overriding, is that something we're interested in?

Amxx: Why not!

The parent's constructor will still write the value to immutable storage, but I guess its ok. Maybe if the immutable storage is not used anywhere that is optimized !

.changeset/proud-seals-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Use immutable beacon address in BeaconProxy
```
Amxx, line 5:
```suggestion
`BeaconProxy`: Use an immutable variable to store the address of the beacon. It is no longer possible for a `BeaconProxy` to upgrade by changing to another beacon.
```

Amxx (dismissed):

.changeset/proud-seals-complain.md
ericglau: > It is no longer possible for a `BeaconProxy` to upgrade by changing to another beacon.

From what I understand, this was not possible in the first place, unless users extended the BeaconProxy to allow this. We never exposed a way to modify the beacon storage slot in the BeaconProxy.

Amxx: You could do it if there is a function in the implementation that sets the ERC1967 slot for beacon

contracts/proxy/beacon/BeaconProxy.sol
frangio: Resolving as there is already an internal function `_getBeacon` that is being used, this means we don't need to make the variable internal.

frangio (approved):
I've updated docs to describe the potential issue with `upgradeToAndCall`.

Amxx (approved):

PR #4436, opened 2023-07-07
Remove unused imports

ernestognw:
Remove unused imports after changes for 5.0

### DIFF
--- a/contracts/access/AccessControl.sol
+++ b/contracts/access/AccessControl.sol
@@ -5,7 +5,6 @@ pragma solidity ^0.8.19;
 
 import {IAccessControl} from "./IAccessControl.sol";
 import {Context} from "../utils/Context.sol";
-import {Strings} from "../utils/Strings.sol";
 import {ERC165} from "../utils/introspection/ERC165.sol";
 
 /**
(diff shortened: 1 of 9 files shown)


frangio:
Not sure why this caused tests to fail but it may be related to the custom error heuristics. See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4437.

Amxx (dismissed):

ernestognw:
Seems like after transpilation, the tests are interpreting the errors as returndata. An example

```
AssertionError: Could not parse as custom error. VM Exception while processing transaction: reverted with an unrecognized custom error (return data: 0xd7e6bcf8)
```

Where `0xd7e6bcf8` is also `bytes4(keccak256('NotInitializing()'))`, confirming @frangio's theory with Hardhat's heuristics. I'm still confused why this happens after transpilation and not in the vanilla repo.

ernestognw:
Seems like importing `Address.sol` in `Initializable` just makes it work.

frangio:
Update: The issue was caused by https://github.com/NomicFoundation/hardhat/issues/4123! Cleaning vanilla artifacts after transpilation fixed it.

Amxx (approved):

PR #4437, opened 2023-07-08
Improve custom error helper when there is no match

frangio:
When the `expectRevertCustomError` helper doesn't find a custom error it fails by attempting to destructure the `null` return value of the `match` function and results in "TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))" which is really disconcerting when it shows up. This PR makes the original error text show up.

Side note: I've also slightly changed the matching logic to look for `custom error` in the message to make it more specific.


Amxx (approved):

ernestognw (approved):
LFG

PR #4440, opened 2023-07-09
Add suggested remappings in readme

### LINKED ISSUE #4439: Can we update Remapping details while working with Foundry
**🧐 Motivation**
While working with foundry, if we copy the below code from ReadMe, it will say file not found
```
import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
```

**📝 Details**
Since the file is not found, it may be confusing for some developers.
So, can we give the  remappings example in Readme? 

Note: Only while working with Foundry
```
remappings = ["@openzeppelin/=lib/openzeppelin-contracts/"]
```

Then, 
```
import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
```

If this will help few devs, then I would like to make this changes.

worksofallwin:
Added directions to remappings while working with foundry in README.md.

Fixes #4439 

Added below lines to give directions about remappings

-----------
Note: Only while working with Foundry

Head to foundry.toml and add below line
```
remappings = ["@openzeppelin/=lib/openzeppelin-contracts/"]
```
--------------

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,12 +40,8 @@ OpenZeppelin Contracts features a [stable API](https://docs.openzeppelin.com/con
 ```
 $ forge install OpenZeppelin/openzeppelin-contracts
 ```
-Note: Only while working with Foundry  
 
-Head to foundry.toml and add the below line
-```
-remappings = ["@openzeppelin/=lib/openzeppelin-contracts/"]
-```
+Add `@openzeppelin/=lib/openzeppelin-contracts/` in `remappings.txt.` 
 
 ### Usage
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,12 @@ OpenZeppelin Contracts features a [stable API](https://docs.openzeppelin.com/con
 ```
 $ forge install OpenZeppelin/openzeppelin-contracts
 ```
+Note: Only while working with Foundry  
+
+Head to foundry.toml and add the below line
+```
+remappings = ["@openzeppelin/=lib/openzeppelin-contracts/"]
+```
 
 ### Usage
 


README.md
```diff
@@ -40,6 +40,12 @@ OpenZeppelin Contracts features a [stable API](https://docs.openzeppelin.com/con
 ```
 $ forge install OpenZeppelin/openzeppelin-contracts
 ```
+Note: Only while working with Foundry  
+
+Head to foundry.toml and add the below line
+```
+remappings = ["@openzeppelin/=lib/openzeppelin-contracts/"]
+```
```
frangio, line 48: I thought `remappings.txt` was more common. I would write:

```suggestion

Add `@openzeppelin/=lib/openzeppelin-contracts/` in `remappings.txt.` 
```

frangio (approved):

pcaversaccio:
Just a quick point for discussion: I introduced as part of my PR https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4134, the file [`remappings.txt`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/remappings.txt):

```txt
openzeppelin/=contracts/
```

In this PR you start sticking to the `@` convention:

```txt
@openzeppelin/=lib/openzeppelin-contracts/
```

So my recommendation is to either change `remappings.txt` to:

```diff
- openzeppelin/=contracts/
+ @openzeppelin/=contracts/
```

and adjust the remappings in the test files where used, or state in the `README`:

```diff
- Add `@openzeppelin/=lib/openzeppelin-contracts/` in `remappings.txt.`
+ Add `openzeppelin/=lib/openzeppelin-contracts/` in `remappings.txt.`
```

It's a consistency question.

worksofallwin:
Hey @pcaversaccio 

Inside the package, I don't think `@` is required.

What do you think?

pcaversaccio:
> Inside the package, I don't think `@` is required.

Actually, I never use `@` in remappings; an example from my repo [here](https://github.com/pcaversaccio/snekmate/blob/main/remappings.txt):

```txt
utils/=lib/utils/
murky/=lib/murky/src/
solady/=lib/solady/src/
solmate/=lib/solmate/src/
prb/test/=lib/prb-test/src/
forge-std/=lib/forge-std/src/
erc4626-tests/=lib/erc4626-tests/
create-util/=lib/create-util/contracts/
openzeppelin/=lib/openzeppelin-contracts/contracts/
solidity-bytes-utils/=lib/solidity-bytes-utils/contracts/
```

Also, you might wanna consider `openzeppelin/=lib/openzeppelin-contracts/contracts/` instead of `openzeppelin/=lib/openzeppelin-contracts/`.

worksofallwin:
Sure, I will use `openzeppelin/=lib/openzeppelin-contracts/contracts/` instead of `openzeppelin/=lib/openzeppelin-contracts/`

```
openzeppelin/=lib/openzeppelin-contracts/contracts/
```

After adding this above line in my remappings.txt without `@`
If we use the below line, compiler will say file not found
```
import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
```

to make our remappings to work, then we have to update import without `@` as well

```
import {ERC721} from "openzeppelin/contracts/token/ERC721/ERC721.sol";
```

If we update our import statement we don't have to use `@`
But is it ok to remove `@` from import, because It represents as scoped package
what do you think?

pcaversaccio:
Actually, if you use my remapping (including `contracts/`), it's enough to write:

```solidity
import {ERC721} from "openzeppelin/token/ERC721/ERC721.sol";
```

Submodules are not like scoped packages since you can't target tags (branches are possible though). Thus, I don't think `@` is necessary. @frangio what's your opinion on this?

frangio:
Our documentation uses `@openzeppelin/contracts`  in import statements following the npm package, and we would want them to work in Foundry with the recommended remappings.

(Also note that OpenZeppelin builds things outside the Contracts library and simply `openzeppelin/` doesn't reflect that.)

PR #4442, opened 2023-07-10
Remove duplicated SLOAD in Arrays.findUpperBound

MollyHe5:
Reading a variable in storage costs more gas than reading a variable in memory. 
Replacing load of ```array.length``` with load of local variable ```high``` can save gas.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx (commented):
Its actually not a mload, because high will be in the stack and not in memory.

frangio (approved):

PR #4456, opened 2023-07-13
Add IERC721MultiMetadata extension

0xGh:
Adding a blueprint for the IERC721MultiMetadata (EIP-7160) - an optional EIP-721 multi-metadata extension.

See https://eips.ethereum.org/EIPS/eip-7160

#### PR Checklist

- [x] Tests (not needed)
- [x] Documentation (not needed - included link to EIP)


contracts/interfaces/IERC7160.sol
```diff
@@ -0,0 +1,60 @@
...
+import {IERC165} from "./IERC165.sol";
+
+/**
+ * @title IERC721MultiMetadata (EIP-7160) optional EIP-721 multi-metadata extension
+ * @author 0xG
+ * @author mpeyfuss
+ * @dev See https://eips.ethereum.org/EIPS/eip-7160 and https://eips.ethereum.org/EIPS/eip-721
+ *      Note: the ERC-165 identifier for this interface is 0x06e1bc5b.
+ */
+interface IERC7160 is IERC165 {
```
0xGh, line 15: Called the interface after the EIP but would it be possible to rename it to IERC721MultiMetadata? What's is the best practice?

Amxx:
@ernestognw 

I'm not a fan of this ERC. Particularly because it includes public setters (IMO that kind of standard should only include getters and events). Also There has been barelly any discussion in the eth magician forum (13 messages, 7 of which are by the author), and its unclear to me who needs/uses that.

So I would not go for a full implementation unless we get multiple independant requests by potential users.

However this PR is just about adding the interface in `contracts/interfaces`. IMO that is ok given the EIP is final. Worst case nobody ever uses that, but at least its not something we will have to activelly maintain and consider for security.

Amxx:
We need a (simple) changelog entry that says the interface was added.

0xGh:
@Amxx converted to draft since it needs an update – I will get to it soon.

Should I drop the dependency on `IERC165`?

I'd also like to take this opportunity to address your feedback/remarks on the ERC itself (happy to chat on discord, twitter dm or a chat service if you want).

> Particularly because it includes public setters (IMO that kind of standard should only include getters and events).

I disagree: in practice I think it is best to standardize the public interface for pinning and unpinning URIs since frontends will likely want to support pinning / unpinning. Therefore a standardized interface is much needed. 

> There has been barelly any discussion in the eth magician forum, and its unclear to me who needs/uses that.

To get a sense of who is using it you can check out this https://twitter.com/search?q=ERC-7160&src=typed_query&f=live The standard is young but artists are starting to adopt it.

> So I would not go for a full implementation unless we get multiple independant requests by potential users.

I agree, for now let's just add the interface. 

The standard leaves room for opinionated implementations around how/when one can pin/unpin and add uris. For now you can refer to the [Transient Labs implementation](https://github.com/Transient-Labs/tl-creator-contracts/blob/main/src/erc-721/multi-metadata/ERC7160TL.sol) (which is again opinionated) (I am not related to TL).

0xGh:
@Amxx this is ready for review. Added the changelog entry but please fix/relocate it if is wrong

PR #4458, opened 2023-07-14
Avoid `returndatacopy` in ERC2771Forwarder by calling via assembly

ernestognw:
As reported in https://github.com/ethereum/solidity/issues/12306, the `returndatacopy` is allocating memory after every call although we're purposely ignoring the `returndata` in forwarded requests. Therefore, calling with assembly will allow to avoid this unnecessary `returndatacopy` and reduce gas costs.

Moreover, the tests were improved to check for correctly checking gas bubbling up and a relayer griefing attack.

Also, a note was added about the availability of ERC2771 when trying to call the `receive` function.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -251,11 +251,19 @@ contract ERC2771Forwarder is EIP712, Nonces {
             // Nonce should be used before the call to prevent reusing by reentrancy
             uint256 currentNonce = _useNonce(signer);
 
-            (success, ) = request.to.call{gas: request.gas, value: request.value}(
-                abi.encodePacked(request.data, request.from)
-            );
+            uint256 reqGas = request.gas;
+            address to = request.to;
+            uint256 value = request.value;
+            bytes memory data = abi.encodePacked(request.data, request.from);
 
-            _checkForwardedGas(gasleft(), request);
+            uint256 gasLeft;
+
+            assembly {
+                success := call(reqGas, to, value, add(data, 0x20), mload(data), 0, 0)
+                gasLeft := gas()
+            }
+
+            _checkForwardedGas(gasLeft, request);
 
             emit ExecutedForwardRequest(signer, currentNonce, success);
         }
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -26,9 +26,11 @@ import {Address} from "../utils/Address.sol";
  * transactions in the mempool. In these cases the recommendation is to distribute the load among
  * multiple accounts.
  *
- * NOTE: The `receive()` function is not available with ERC2771 because it requires `msg.data` to
- * always have the `from` address at the end of the calldata, whereas the `receive()` function only
- * works when data is empty.
+ * WARNING: Any forwarded call to a contract relying on a specific calldata length will be affected
+ * this forwarder since the ERC2771 specification requires `msg.data` to always be suffixed with the
+ * `from` address, making the minimum calldata length received within the subcall is 20 for the address
+ * size in bytes. An example of this is the `receive` function, which can be accessed only when the
+ * calldata is nonempty.
  *
  * ==== Security Considerations
  *
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -27,10 +27,10 @@ import {Address} from "../utils/Address.sol";
  * multiple accounts.
  *
  * WARNING: Any forwarded call to a contract relying on a specific calldata length will be affected
- * this forwarder since the ERC2771 specification requires `msg.data` to always be suffixed with the
- * `from` address, making the minimum calldata length received within the subcall is 20 for the address
- * size in bytes. An example of this is the `receive` function, which can be accessed only when the
- * calldata is nonempty.
+ * by this forwarder since the ERC2771 specification requires `msg.data` to always be suffixed with the
+ * `from` address, adding the address size in bytes (20) to the calldata size. An example of an unexpected
+ * behavior could be an unintended fallback (or another function) invocation while trying to invoke the `receive`
+ * function, which can be accessed only if `msg.data.length == 0`.
  *
  * ==== Security Considerations
  *
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -26,6 +26,10 @@ import {Address} from "../utils/Address.sol";
  * transactions in the mempool. In these cases the recommendation is to distribute the load among
  * multiple accounts.
  *
+ * NOTE: The `receive()` function is not available with ERC2771 because it requires `msg.data` to
+ * always have the `from` address at the end of the calldata, whereas the `receive()` function only
+ * works when data is empty.
+ *
  * ==== Security Considerations
  *
  * If a relayer submits a forward request, it should be willing to pay up to 100% of the gas amount
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -26,12 +26,6 @@ import {Address} from "../utils/Address.sol";
  * transactions in the mempool. In these cases the recommendation is to distribute the load among
  * multiple accounts.
  *
- * WARNING: Any forwarded call to a contract relying on a specific calldata length will be affected
- * by this forwarder since the ERC2771 specification requires `msg.data` to always be suffixed with the
- * `from` address, adding the address size in bytes (20) to the calldata size. An example of an unexpected
- * behavior could be an unintended fallback (or another function) invocation while trying to invoke the `receive`
- * function, which can be accessed only if `msg.data.length == 0`.
- *
  * ==== Security Considerations
  *
  * If a relayer submits a forward request, it should be willing to pay up to 100% of the gas amount
--- a/test/metatx/ERC2771Forwarder.test.js
+++ b/test/metatx/ERC2771Forwarder.test.js
@@ -41,7 +41,7 @@ contract('ERC2771Forwarder', function (accounts) {
     this.alice.address = web3.utils.toChecksumAddress(this.alice.getAddressString());
 
     this.timestamp = await time.latest();
-    this.receiver = await CallReceiverMock.new(this.forwarder.address, { value: 1e18 });
+    this.receiver = await CallReceiverMock.new();
     this.request = {
       from: this.alice.address,
       to: this.receiver.address,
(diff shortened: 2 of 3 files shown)


test/metatx/ERC2771Forwarder.test.js
```diff
@@ -208,21 +216,30 @@ contract('ERC2771Forwarder', function (accounts) {
...
       const { transactions } = await web3.eth.getBlock('latest');
       const { gasUsed } = await web3.eth.getTransactionReceipt(transactions[0]);
 
       expect(gasUsed).to.be.equal(gasAvailable);
     });
+
+    it('bubbles out of gas forced by the relayer', async function () {
+      // A malicious relayer would provide a gas limit that is too low to execute the request.
+      const estimate = await this.estimateRequest(this.request);
+      const gasAvailable = estimate + 2_000; // Add some gas to let the first-level transaction pass
```
ernestognw, line 235: I'm worried this value (2_000) and its equivalent on the batch version (10_000) is arbitrary because if there's not enough gas, the transaction won't be even sent because it will revert with a "not enough gas" error before sending.

I don't have an idea of how to justify this value better.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -251,11 +255,19 @@ contract ERC2771Forwarder is EIP712, Nonces {
             // Nonce should be used before the call to prevent reusing by reentrancy
             uint256 currentNonce = _useNonce(signer);
 
-            (success, ) = request.to.call{gas: request.gas, value: request.value}(
-                abi.encodePacked(request.data, request.from)
-            );
+            uint256 reqGas = request.gas;
+            address to = request.to;
+            uint256 value = request.value;
+            bytes memory data = abi.encodePacked(request.data, request.from);
```
ernestognw, line 261: We can remove this memory allocation by modifying the `request.data` value in place using assembly. I think it's not a priority rn but we may open an issue.

ernestognw:
The [gas benchmarks reported by the CI](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/5556599580/attempts/1#summary-15052446486) are quite confusing because we also changed the tests. Fortunately, we have also Foundry to make comparisons. I fixed the batch fuzz test to a fixed `batchSize` of 10 for a more fair comparison and here are the results:

### Gas Reports

```diff
diff --git a/.gas-report b/.gas-report
index b499635e..7f1b46cf 100644
--- a/.gas-report
+++ b/.gas-report
@@ -1,9 +1,9 @@
 | test/metatx/ERC2771Forwarder.t.sol:ERC2771ForwarderMock contract |                 |        |        |        |         |
 |------------------------------------------------------------------|-----------------|--------|--------|--------|---------|
 | Deployment Cost                                                  | Deployment Size |        |        |        |         |
-| 835837                                                           | 5438            |        |        |        |         |
+| 818218                                                           | 5350            |        |        |        |         |
 | Function Name                                                    | min             | avg    | median | max    | # calls |
-| execute                                                          | 40746           | 40746  | 40746  | 40746  | 1       |
-| executeBatch                                                     | 138060          | 138060 | 138060 | 138060 | 1       |
+| execute                                                          | 40400           | 40400  | 40400  | 40400  | 1       |
+| executeBatch                                                     | 130033          | 130033 | 130033 | 130033 | 1       |
 | nonces                                                           | 2531            | 2531   | 2531   | 2531   | 2       |
 | structHash                                                       | 1909            | 1909   | 1909   | 1909   | 11      |
```

### Gas Snapshots

```diff
diff --git a/.gas-snapshot b/.gas-snapshot
index 8b9ee139..e2618d5d 100644
--- a/.gas-snapshot
+++ b/.gas-snapshot
@@ -1,2 +1,2 @@
-ERC2771ForwarderTest:testExecuteAvoidsETHStuck(uint256,uint256,bool) (runs: 10000, μ: 74319, ~: 74827)
-ERC2771ForwarderTest:testExecuteBatchAvoidsETHStuck(uint256,uint256,uint256) (runs: 10000, μ: 327458, ~: 331123)
+ERC2771ForwarderTest:testExecuteAvoidsETHStuck(uint256,uint256,bool) (runs: 10000, μ: 73998, ~: 74481)
+ERC2771ForwarderTest:testExecuteBatchAvoidsETHStuck(uint256,uint256,uint256) (runs: 10000, μ: 324413, ~: 328190)
```

contracts/metatx/ERC2771Forwarder.sol
ernestognw: This idea is wrong, `request` is in `calldata` so there's no way to reuse memory.

Also, we discussed a bit and we think it's a bit kamikaze to do this memory manipulation.

Solving.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -26,6 +26,10 @@ import {Address} from "../utils/Address.sol";
  * transactions in the mempool. In these cases the recommendation is to distribute the load among
  * multiple accounts.
  *
+ * NOTE: The `receive()` function is not available with ERC2771 because it requires `msg.data` to
+ * always have the `from` address at the end of the calldata, whereas the `receive()` function only
+ * works when data is empty.
+ *
```
Amxx, line 32: receive is can still be implemented and called if the sender is not the forwarder. 

Is in interresting to note that there is indeed an issue if the forwarder relays an empty "data" field, because in that case the caller will be interpreted as a selector. Maybe the Forwarder should require that `request.data.length >= 4`

ernestognw: I very much agree that calldata size dependency might be an issue, amazing catch!

I'll improve the note and solve this but let me know if you think it can be improved further

frangio: Reopening to make this explicit and make sure we agree:

>Maybe the Forwarder should require that `request.data.length >= 4`

I initially agreed with this idea, but I discussed it with @ernestognw and he made an interesting point that it would be very much Solidity and "Solidity ABI encoding" specific, by caring about 4-byte selectors and receive functions, whereas the rest of the contract is agnostic to Solidity specifics.

I think it's a good point so I would not add this check. It can be implemented at the request-generation and/or forwarder level.

test/metatx/ERC2771Forwarder.test.js
frangio: Could we perhaps estimate the simplest request possible and use that value as the offset?

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -26,6 +26,12 @@ import {Address} from "../utils/Address.sol";
  * transactions in the mempool. In these cases the recommendation is to distribute the load among
  * multiple accounts.
  *
+ * WARNING: Any forwarded call to a contract relying on a specific calldata length will be affected
```
frangio, line 29: I feel that this warning belongs in ERC2771Context rather than here, because it is a contract that inherits ERC2771Context that will be affected.

frangio (dismissed):
Looks good to me, minor comment regarding the docs update though:

test/metatx/ERC2771Forwarder.test.js
```diff
@@ -208,21 +216,30 @@ contract('ERC2771Forwarder', function (accounts) {
...
+      await expectRevert.outOfGas(this.forwarder.execute(this.requestData, { gas: gasAvailable }));
```
Amxx, line 236: My understanding of the test is
- the request is the normal one (`mockFunction()` with `100_000` gas)
- you estimate how much is actually needed (not 100_000)
- you give slightly more (estimate + 2000)
  - the forwarder use some gas for "pre-processing"
  - the forwarder passes 63/64 which you expect to be less than `estimate`
  - the inner call should revert (out of gas)
  - the forwarder call should detect that and consume all the gas (even though some is left)
    - that is what is checked by `expect(gasUsed).to.be.equal(gasAvailable);`
    
    
I took me a while to understand all that. Am I correct ? IMO we should add some documentation to explain how the test works.

contracts/metatx/ERC2771Forwarder.sol
ernestognw: I agree, I avoided adding anything to `Context` because it still requires doc improvements and I was thinking in this contract standalone. Solving

test/metatx/ERC2771Forwarder.test.js
ernestognw: The current `this.request` is a very simple request already, its estimation even adding the 21000 buffer that won't go in the subcall but covers initiating a relayed request:

```js
this.estimateRequest = request =>
      web3.eth.estimateGas({
        from: this.forwarder.address,
        to: request.to,
        data: web3.utils.encodePacked({ value: request.data, type: 'bytes' }, { value: request.from, type: 'address' }),
        value: request.value,
        gas: request.gas,
      });
```

In order to provide an accurate buffer we have to calculate how much gas is consumed from starting the transaction to the request's `CALL` opcode and use it as buffer. Also, doing an arbitrarily high buffer will make the underlying transaction not to revert.

We agreed internally to keep both so I'm solving

test/metatx/ERC2771Forwarder.test.js
ernestognw: Yes, all of that is correct, and now that you mention it I agree it may be difficult for others to parse. I added comments for both tests (single and batch).

contracts/metatx/ERC2771Forwarder.sol
Amxx: > Something interesting could be to add a reverting receive() implementation in ERC2771Context. It's a bit radical because it would cause receives to revert also when the sender is not the forwarder. This just makes sure that someone using ERC2771Context doesn't also implement a receive function expecting it to work with forwarded requests. What do you think?

I guess that receive would be virtual so that devs that override it with some checks. It's not a bad idea, but I'm worried it will be difficult to understand for devs.

ernestognw: Although radical I'd say it's more common to write positive test cases rather than negatives. For example, instead of writing a test proving a contract can't receive funds, developers will more usually write a spec for `should receive funds when ...`, which will fail.

It's definitely less likely to make a mistake imo. Though, I agree with @Amxx on overriding. However, a messy override would require mixing the forwarder with more than 2 inherited contracts with a `receive` function, and I don't think anyone would get a complex inheritance in a forwarder.

test/metatx/ERC2771Forwarder.test.js
```diff
@@ -208,21 +216,44 @@ contract('ERC2771Forwarder', function (accounts) {
...
+      // the forwarder by executing requests and providing a top-level call gas limit that
+      // is too low to successfully finish the request after the 63/64 rule.
+
+      // We set the baseline to the gas limit consumed by a successful request if it was executed
+      // normally. Note this includes the 21000 buffer that also the relayer will be charged to
+      // start a request execution.
+      const estimate = await this.estimateRequest(this.request);
+
+      // Because the relayer call consumes gas until the `CALL` opcode, the gas left after failing
+      // the subcall is not be enough to finish the top level call (after testing), so we add a
```
ernestognw, line 243:
```suggestion
      // the subcall won't enough to finish the top level call (after testing), so we add a
```

frangio (approved):
Thanks for adding the test comments. Looks good to me.

PR #4460, opened 2023-07-14
Allow Initializable versions greater than 256

### LINKED ISSUE #3924: Allow reinitializer versions greater than 256
To maintain storage compatibility with the original Initializable, we made reinitializer versions of type `uint8`. We could use a larger type here once we break compatibility. It still makes sense to pack the value with the `initializing` boolean, so it can't be more than uint248, but even that sounds like an exaggeration and we might want to make it a smaller value like uint64 for the possibility of adding some other values in the same storage slot in the future.

Suggested in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3870.

ernestognw:
Fixes LIB-828
Fixes #3924 

A greater initialized was proposed for #3924, which is implemented within this PR. Also, the EIP-7201 for Namespaced storage is used here to reduce the risk of storage collisions in upgradeable contracts that are also initializable. In this way, the Initializable variables are placed at a storage location that's guaranteed not to clash.

#### PR Checklist

- [x] Tests
- [ ] Documentation


ernestognw:
Bumping pragma to 0.8.20 on Initializable to support NatSpec in structs.

Also, keeping it as a draft at the moment because I suspect we may require a test for storage collisions on storage slot 0.

contracts/proxy/utils/Initializable.sol
```diff
@@ -89,17 +102,20 @@ abstract contract Initializable {
      * Emits an {Initialized} event.
      */
     modifier initializer() {
-        bool isTopLevelCall = !_initializing;
-        if (!(isTopLevelCall && _initialized < 1) && !(address(this).code.length == 0 && _initialized == 1)) {
+        // solhint-disable-next-line var-name-mixedcase
```
ernestognw, line 105: There are several of these for the `$` storage variable. We either keep this rule or wait on https://github.com/protofire/solhint/pull/453 to be merged if that's the case.

frangio:
I like this but before reviewing I'll let @Amxx comment whether he thinks this is a good idea.

contracts/proxy/utils/Initializable.sol
```diff
@@ -153,26 +151,74 @@ abstract contract Initializable {
      * Emits an {Initialized} event the first time it is successfully executed.
      */
     function _disableInitializers() internal virtual {
-        if (_initializing) {
+        // solhint-disable-next-line var-name-mixedcase
+        InitializableStorage storage $ = _getInitializableStorage();
+
+        if ($._initializing) {
             revert AlreadyInitialized();
```
Amxx, line 158: That is not changed by this PR, but I'm wondering if that is the right error to trigger.

contracts/proxy/utils/Initializable.sol
```diff
@@ -153,26 +151,74 @@ abstract contract Initializable {
...
      */
     function _isInitializing() internal view returns (bool) {
-        return _initializing;
+        return _getInitializableStorage()._initializing;
+    }
+
+    /**
+     * @dev Returns a pointer to the storage namespace.
+     */
+    // solhint-disable-next-line var-name-mixedcase
```
Amxx, line 183: Is `_getInitializableStorage` not a correct name already?

ernestognw: It's because of the `$` 😢

contracts/proxy/utils/Initializable.sol
```diff
@@ -153,26 +151,74 @@ abstract contract Initializable {
...
+    function _getInitializableStorage() private pure returns (InitializableStorage storage $) {
+        assembly {
+            $.slot := _INITIALIZABLE_STORAGE
+        }
+    }
+
+    function _setInitializedVersion(uint64 version) private returns (bool) {
+        // solhint-disable-next-line var-name-mixedcase
```
Amxx, line 191: I don't understand why this is necessary

ernestognw: Yes, I opened [a PR](https://github.com/protofire/solhint/pull/453) but requires a bit more discussion because they may allow `$` in some other rules.

frangio: We should simply get rid of these casing rules and write our own.

frangio:
We discussed offline that we should not include https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3950 in this PR because it breaks a pattern that we use for testing contracts with initializers, as described in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3344.

frangio (changes requested):
We discussed offline that we should not include https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3950 in this PR because it breaks a pattern that we use for testing contracts with initializers, as described in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3344.

ernestognw:
The storage layout checks are failing as expected

contracts/proxy/utils/Initializable.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v4.9.0) (proxy/utils/Initializable.sol)
 
-pragma solidity ^0.8.19;
+pragma solidity ^0.8.20;
```
Amxx, line 4: Why do we go to 0.8.20 here ?

contracts/proxy/utils/Initializable.sol
```diff
@@ -55,14 +55,27 @@ pragma solidity ^0.8.19;
...
+        bool _initializing;
+    }
 
-    /**
-     * @dev Indicates that the contract is in the process of being initialized.
-     */
-    bool private _initializing;
+    // keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.Initializable")) - 1))
+    bytes32 private constant _INITIALIZABLE_STORAGE =
+        0xf0c57e16840df040f15088dc2f81fe391c3923bec73e23a9662efc9c229c6a0e;
```
Amxx, line 78: We may want to do the computation differently.

test/proxy/utils/Initializable.test.js
```diff
@@ -213,7 +213,7 @@ contract('Initializable', function () {
     it('old and new patterns in good sequence', async function () {
       const ok = await DisableOk.new();
       await expectEvent.inConstruction(ok, 'Initialized', { version: '1' });
-      await expectEvent.inConstruction(ok, 'Initialized', { version: '255' });
+      await expectEvent.inConstruction(ok, 'Initialized', { version: '18446744073709551615' }); // MAX_UINT64
```
Amxx, line 216: Would the test support an hex syntax ? IMO `0xFFFFFFFFFFFFFFFF` would be more readable

contracts/proxy/utils/Initializable.sol
frangio: `@custom` in structs is only available in 0.8.20. We need to do this in the whole library.

contracts/proxy/utils/Initializable.sol
ernestognw: Ah, amazing catch! Note we haven't updated EIP-7201 to reflect the impact of Verkle Trees neither.

Are we ok with?:

```suggestion
    // keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.Initializable")) - 256)) & ~bytes32(uint256(0xFF))
    bytes32 private constant _INITIALIZABLE_STORAGE =
        0x2cb0f341752e459a3615b447e07b189024fd0be4129b407c58ddd277d84b6800;
```

test/proxy/utils/Initializable.test.js
ernestognw: Unfortunately not, but we can do `BigInt('0xFFFFFFFFFFFFFFFF').toString()` which is similar imo. What do you think?

contracts/proxy/utils/Initializable.sol
frangio: Let's not change this until we change the EIP.

test/proxy/utils/Initializable.test.js
frangio: `(2n ** 64n - 1n).toString()` might be slightly better.

ernestognw: I like this one better. Updated.

frangio:
Closed accidentally.

ernestognw:
> Closed accidentally.

`reinitialize(2)`

test/proxy/utils/Initializable.test.js
```diff
@@ -213,7 +213,7 @@ contract('Initializable', function () {
     it('old and new patterns in good sequence', async function () {
       const ok = await DisableOk.new();
       await expectEvent.inConstruction(ok, 'Initialized', { version: '1' });
-      await expectEvent.inConstruction(ok, 'Initialized', { version: '255' });
+      await expectEvent.inConstruction(ok, 'Initialized', { version: (2n ** 64n - 1n).toString() }); // MAX_UINT64
```
Amxx, line 216: we should consider generalizing this syntax.

Amxx:
LGTM. Do we wait on the linter rule or do we merge like this and come back to it later?

contracts/proxy/utils/Initializable.sol
ernestognw: I agree with you but we may decide on this once we decide to tackle #3950 as well

test/proxy/utils/Initializable.test.js
ernestognw: I used `web3.utils.toBN(1).shln(64).subn(1).toString()` instead because I think we've previously agreed on a similar one in the ERC2771Forwarder (but for `uint48`), does this work?

ernestognw:
> LGTM. Do we wait on the linter rule or do we merge like this and come back to it later?

I'd rather not wait on them, I'm sure we'll revisit after starting with the transpiler work

.changeset/thick-pumpkins-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`Initializable`: Allow versions greater than 256.
```
frangio, line 5:
```suggestion
`Initializable`: Use the namespaced storage pattern to avoid putting critical variables in slot 0. Allow reinitializer versions greater than 256.
```

frangio (approved):

PR #4469, opened 2023-07-18
Add ERC165 support for IERC5313 in AccessControlDefaultAdminRules

### LINKED ISSUE #4464: AccessControlDefaultAdminRules implements IERC5313 but does not handle it in supportsInterface()
The contract AccessControlDefaultAdminRules implements the interface IERC5313. One would expect that if we call supportsInterface() it would signal true with the interfaceId of IERC5313. However, it currently returns false.

**💻 Environment**

Used with hardhat and OpenZeppelin Version 4.9

**📝 Details**

The current implementation of supportsInterface only handles IAccessControlDefaultAdminRules and IAccessControl but does not handle IERC5313 (even though it implements the interface).

```
        return interfaceId == type(IAccessControlDefaultAdminRules).interfaceId || interfaceId == type(IERC5313).interfaceId || super.supportsInterface(interfaceId);
    }
```

**🔢 Code to reproduce bug**

Call of supportsInterface('0x8da5cb5b') returns false (0x8da5cb5b is the interfaceId of IERC5313)

ernestognw:
Fixes #4464

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/silly-bottles-swim.md
+++ b/.changeset/silly-bottles-swim.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`AccessControlDefaultAdminRules`: Add ERC165 support for the IERC5313 interface.
--- a/contracts/access/extensions/AccessControlDefaultAdminRules.sol
+++ b/contracts/access/extensions/AccessControlDefaultAdminRules.sol
@@ -63,7 +63,10 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @dev See {IERC165-supportsInterface}.
      */
     function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
-        return interfaceId == type(IAccessControlDefaultAdminRules).interfaceId || super.supportsInterface(interfaceId);
+        return
+            interfaceId == type(IAccessControlDefaultAdminRules).interfaceId ||
+            interfaceId == type(IERC5313).interfaceId ||
+            super.supportsInterface(interfaceId);
     }
 
     /**
--- a/test/access/AccessControl.behavior.js
+++ b/test/access/AccessControl.behavior.js
@@ -224,7 +224,7 @@ function shouldBehaveLikeAccessControlEnumerable(admin, authorized, other, other
 }
 
 function shouldBehaveLikeAccessControlDefaultAdminRules(delay, defaultAdmin, newDefaultAdmin, other) {
-  shouldSupportInterfaces(['AccessControlDefaultAdminRules']);
+  shouldSupportInterfaces(['AccessControlDefaultAdminRules', 'ERC5313']);
 
   for (const getter of ['owner', 'defaultAdmin']) {
     describe(`${getter}()`, function () {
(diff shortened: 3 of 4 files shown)


Amxx (approved):

frangio:
[ERC-5315](https://eips.ethereum.org/EIPS/eip-5313) doesn't require (or even recommend?) publishing the interface via ERC-165, so I would not add this.

ernestognw:
To be fair, I don't exactly remember why we decided to add EIP-5315 to ACDAR, but the only benefit of being compliant imo is actually exposing the interface. The EIP states the following in regard to EIP-165

> Because this specification does not extend EIP-165, calling this EIP’s owner function cannot result in complete certainty that the result is indeed the owner.

I'm not comfortable with the tradeoff of assuming there's an `owner()` function if the contract contains the selector vs. not exposing the interface with EIP-165. It's a security consideration and we don't loose much by adding EIP-165 here.

frangio:
> the only benefit of being compliant imo is actually exposing the interface

The ERC doesn't say that the interface should be published via ERC-165. It actually says the opposite, that the ERC-165 interface should not be expected.

This would be a non-standard addition to ERC-5313. Which is why I don't think we should merge this.

ernestognw:
I agree it's an opinionated additional change, but I think that's also enough reason for removing ERC 5313 altogether, what other could be the benefit?

frangio:
>removing ERC 5313 altogether

Does this mean removing `owner()`?

My impression was that if there is an account that is effectively an owner, it makes sense to publish it using ERC-5313 `owner()`. But I am open to removing it if we don't think it makes sense.

ernestognw:
I meant just removing the IERC5313 from the inheritance. I think it can be more confusing than helping. If i remember correctly we added it because there's some demand for this standard in NFT marketplaces, I recall @Amxx mentioned something related.

If that's the case, how are these benefits supposed to be communicated to other contracts whatsoever?

ernestognw:
We discussed this internally and we decided not to support the IERC5313 interface in the `supportsInterface` function.

The idea of having ERC5313 compliance is because the EIP rationale mentions "Backwards compatibility with existing contracts" but doesn't require EIP-165 for the same reason. It makes sense to do an `eth_call` for the `owner()` interface and use the value returned because it's very unlikely to accidentally hit a contract that doesn't have an `owner()` but an equivalent selector and no parameters.

Similarly, we think there's no reason for a contract to maliciously forge a selector that's ERC5313 compliant but means something different.

We consider the main advantages of having ERC5313 to be:

- A documented ERC compliance for code readers
- A compiler check for including the `owner()` function (important when making changes to the codebase)

And also we agreed that adding it to EIP-165 sets a precedent about contracts using ERC5313 so that developers may expect it to be reported in the `supportsInterface` which is encouraging non-compliance with 5313. Therefore we also recommend not adding EIP-165 to 5313.

RobertOttBS:
Ok, thanks a lot. However with the explanation I do not understand then why you included supportsInterface() for IAccessControl and IAccessControlDefaultAdminRules() which also do not extend EIP-165. Currently, IRC5313 is the only interfaces in the contract that is implemented but returns false when asking for it with supportsInterface(). Anyway, I don't see it as a big issue. Developers can programatically implement workarounds.

frangio:
ERC-165 has never been guaranteed for all interfaces. IERC20 for example is not published via ERC-165.

Interfaces are only published in that way if:
1) The ERC spec requires it. Example: ERC721
2) We define the interface (not an ERC) and we need a way to provide feature detection. Example: AccessControl, Governor

Some interfaces don't fall in either of these categories. And when the ERC does not mention ERC-165 we avoid including it because it's a non-standard behavior that people may start to rely on but other implementations will not necessarily follow.

RobertOttBS:
Thats fine with us. We have implemented our expected behaviour in the sub-contracts of AccessControlDefaultAdminRules so as said, developers can implement such workarounds easily and still have other contracts figuring out whether or not 'owner()' is implemented. But thanks a lot for the explanations!

PR #4473, opened 2023-07-21
Explicitly document EIP712._domainSeparatorV4 optimization

SteMak:
Proposing to add the notice to the `EIP712._domainSeparatorV4` function because it may be not intuitive why the optimization is reasonable and `immutable` variables are selected to store cached data.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/cryptography/EIP712.sol
+++ b/contracts/utils/cryptography/EIP712.sol
@@ -78,7 +78,10 @@ abstract contract EIP712 is IERC5267 {
      * @dev Returns the domain separator for the current chain.
      *
      * NOTE: Caching the domain separator saves 6+ times more Gas than is wasted if the contract address or
-     * `block.chainid` mismatches with the cached ones compared to the direct `_buildDomainSeparator` call.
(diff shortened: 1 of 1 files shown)


contracts/utils/cryptography/EIP712.sol
```diff
@@ -76,6 +76,12 @@ abstract contract EIP712 is IERC5267 {
 
     /**
      * @dev Returns the domain separator for the current chain.
+     *
+     * NOTE: Caching the domain separator saves 6+ times more Gas than is wasted if the contract address or
+     * `block.chainid` mismatches with the cached ones, compared to the direct `_buildDomainSeparator` call.
+     *
+     * NOTE: Usage of `immutable` variables saves 40+ times more Gas than is wasted if the contract address or
+     * `block.chainid` mismatches with the cached ones, compared to the usage of state variables.
```
Amxx, line 84: I'd comment on why the cache is used for (in which case we expect the values to change) more than on the savings. IMO, stating `saves x+ times` is really not clear. If we want to give number (I don't think we should) it should be a "raw" number. Note that this will change depending on the compiler version and settings anyway.

PR #4476, opened 2023-07-24
Improve ERC4626 fees example

ernestognw:
This PR improves the documentation of the ERC4626Fees example in the docs with better documentation and better specifying the units used in the contract.

See this [forum thread](https://forum.openzeppelin.com/t/please-explain-erc4626fees-math/37141/3).

### DIFF
--- a/contracts/mocks/docs/ERC4626Fees.sol
+++ b/contracts/mocks/docs/ERC4626Fees.sol
@@ -72,19 +72,19 @@ abstract contract ERC4626Fees is ERC4626 {
     // === Fee configuration === ///
 
     function _entryFeeBasePoint() internal view virtual returns (uint256) {
-        return 0; // i.e. 1_000 for 1%
+        return 0; // replace with e.g. 1_000 for 1%
     }
 
     function _exitFeeBasePoint() internal view virtual returns (uint256) {
-        return 0; // i.e. 1_000 for 1%
+        return 0; // replace with e.g. 1_000 for 1%
     }
 
     function _entryFeeRecipient() internal view virtual returns (address) {
-        return address(0); // i.e. treasury
+        return address(0); // replace with e.g. a treasury address
     }
 
     function _exitFeeRecipient() internal view virtual returns (address) {
-        return address(0); // i.e. treasury
+        return address(0); // replace with e.g. a treasury address
     }
 
     // === Fee operations === ///
--- a/contracts/mocks/docs/ERC4626Fees.sol
+++ b/contracts/mocks/docs/ERC4626Fees.sol
@@ -89,12 +89,12 @@ abstract contract ERC4626Fees is ERC4626 {
 
     // === Fee operations === ///
 
-    /// @dev Calculates fee on `assets` not including fees. Used in {IERC4626-mint} and {IERC4626-withdraw} operations.
+    /// @dev Calculates the fees that should be added to an amount `assets` that does not already include fees. Used in {IERC4626-mint} and {IERC4626-withdraw} operations.
     function _feeOnRaw(uint256 assets, uint256 feePCM) private pure returns (uint256) {
         return assets.mulDiv(feePCM, _PCM_SCALE, Math.Rounding.Ceil);
     }
 
-    /// @dev Calculates fee on `assets` already including fees. Used in {IERC4626-deposit} and {IERC4626-redeem} operations.
+    /// @dev Calculates the fee part of an amount `assets` that already includes fees. Used in {IERC4626-deposit} and {IERC4626-redeem} operations.
     function _feeOnTotal(uint256 assets, uint256 feePCM) private pure returns (uint256) {
         return assets.mulDiv(feePCM, feePCM + _PCM_SCALE, Math.Rounding.Ceil);
     }
--- a/contracts/mocks/docs/ERC4626Fees.sol
+++ b/contracts/mocks/docs/ERC4626Fees.sol
@@ -91,19 +91,11 @@ abstract contract ERC4626Fees is ERC4626 {
 
     /// @dev Calculates fee on `assets` not including fees. Used in {IERC4626-mint} and {IERC4626-withdraw} operations.
     function _feeOnRaw(uint256 assets, uint256 feeBasePoint) private pure returns (uint256) {
-<<<<<<< Updated upstream
-        return assets.mulDiv(feeBasePoint, 1e5, Math.Rounding.Ceil);
-=======
-        return assets.mulDiv(feeBasePoint, _PCM_SCALE, Math.Rounding.Up);
->>>>>>> Stashed changes
+        return assets.mulDiv(feeBasePoint, _PCM_SCALE, Math.Rounding.Ceil);
     }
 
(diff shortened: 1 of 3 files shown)


contracts/mocks/docs/ERC4626Fees.sol
```diff
@@ -64,27 +69,41 @@ abstract contract ERC4626Fees is ERC4626 {
...
+>>>>>>> Stashed changes
     }
 
+    /// @dev Calculates fee on `assets` already including fees. Used in {IERC4626-deposit} and {IERC4626-redeem} operations.
     function _feeOnTotal(uint256 assets, uint256 feeBasePoint) private pure returns (uint256) {
+<<<<<<< Updated upstream
         return assets.mulDiv(feeBasePoint, feeBasePoint + 1e5, Math.Rounding.Ceil);
+=======
+        return assets.mulDiv(feeBasePoint, feeBasePoint + _PCM_SCALE, Math.Rounding.Up);
+>>>>>>> Stashed changes
```
Amxx, line 107:
```suggestion
    function _feeOnRaw(uint256 assets, uint256 feeBasePoint) private pure returns (uint256) {
        return assets.mulDiv(feeBasePoint, _PCM_SCALE, Math.Rounding.Ceil);
    }

    /// @dev Calculates fee on `assets` already including fees. Used in {IERC4626-deposit} and {IERC4626-redeem} operations.
    function _feeOnTotal(uint256 assets, uint256 feeBasePoint) private pure returns (uint256) {
        return assets.mulDiv(feeBasePoint, feeBasePoint + _PCM_SCALE, Math.Rounding.Ceil);
```

Amxx:
This file might be in docs, there is an actual mock and unit tests for it. The mock require an update.

contracts/mocks/docs/ERC4626Fees.sol
ernestognw: Issues of pushing just before running into a call. Thanks!

ernestognw:
I didn't notice it, thanks for flagging!

contracts/mocks/docs/ERC4626Fees.sol
```diff
@@ -64,27 +69,33 @@ abstract contract ERC4626Fees is ERC4626 {
...
     function _exitFeeRecipient() internal view virtual returns (address) {
-        return address(0);
+        return address(0); // i.e. treasury
     }
 
-    function _feeOnRaw(uint256 assets, uint256 feeBasePoint) private pure returns (uint256) {
-        return assets.mulDiv(feeBasePoint, 1e5, Math.Rounding.Ceil);
+    // === Fee operations === ///
+
+    /// @dev Calculates fee on `assets` not including fees. Used in {IERC4626-mint} and {IERC4626-withdraw} operations.
```
Amxx, line 92:
```suggestion
    /// @dev Calculates fee to apply on an amount `assets` that does already include fees. Used in {IERC4626-mint} and {IERC4626-withdraw} operations.
```

contracts/mocks/docs/ERC4626Fees.sol
```diff
@@ -64,27 +69,33 @@ abstract contract ERC4626Fees is ERC4626 {
...
+    function _feeOnRaw(uint256 assets, uint256 feePCM) private pure returns (uint256) {
+        return assets.mulDiv(feePCM, _PCM_SCALE, Math.Rounding.Ceil);
     }
 
-    function _feeOnTotal(uint256 assets, uint256 feeBasePoint) private pure returns (uint256) {
-        return assets.mulDiv(feeBasePoint, feeBasePoint + 1e5, Math.Rounding.Ceil);
+    /// @dev Calculates fee on `assets` already including fees. Used in {IERC4626-deposit} and {IERC4626-redeem} operations.
```
Amxx, line 97:
```suggestion
    /// @dev Calculates the fee part of an amount `assets` that already includes fees. Used in {IERC4626-deposit} and {IERC4626-redeem} operations.
```

contracts/mocks/docs/ERC4626Fees.sol
ernestognw: I consciously decided to be short here because at this point I'd use 2 lines instead, and because these are the docs I'd prefer it to be the less bloated possible.

Same applies to the other one.

Thoughts @frangio?

frangio: 2 lines is good (we can still use `///`). "fee on `assets` not including fees" sounds very confusing.

I believe there is an error in @Amxx's suggestion, should say "does _not_ already include".

contracts/mocks/docs/ERC4626Fees.sol
```diff
@@ -64,27 +69,33 @@ abstract contract ERC4626Fees is ERC4626 {
         }
     }
 
-    function _entryFeeBasePoint() internal view virtual returns (uint256) {
-        return 0;
+    // === Fee configuration === ///
+
+    function _entryFeePCM() internal view virtual returns (uint256) {
+        return 0; // i.e. 1_000 for 1%
```
frangio, line 75: The use of `i.e.` in these comments is not right. `0 // i.e. 1000` means "0, equivalently 1000".

We should write something like `// replace with e.g. 1_000 for 1%`.

contracts/mocks/docs/ERC4626Fees.sol
```diff
@@ -7,36 +7,41 @@ import {ERC4626} from "../../token/ERC20/extensions/ERC4626.sol";
 import {SafeERC20} from "../../token/ERC20/utils/SafeERC20.sol";
 import {Math} from "../../utils/math/Math.sol";
 
+/// @dev ERC4626 vault with entry/exit fees expressed in https://en.wikipedia.org/wiki/Per_cent_mille[PCM].
```
frangio, line 10: I expect bps to be more common in finance than PCM. Are we sure we don't want to fix the scale constant to be actual bps?

Amxx: We can go to BPS. The actual cahnges to the code will be smaller, but the have to change the unit tests instead.

frangio: Good data points shared here by @Amxx showing that bps might be more used: https://forum.openzeppelin.com/t/please-explain-erc4626fees-math/37141/5

ernestognw: Thanks for checking it out, I completely missed taking a look.
I reverted back to base point instead.

contracts/mocks/docs/ERC4626Fees.sol
ernestognw: TIL `i.e.` is not `in example`

Solved.

frangio (dismissed):

frangio (approved):

frangio:
Note: The term is "basis point" not "base point". Fixed in the last 2 commits.

ernestognw:
> Note: The term is "basis point" not "base point". Fixed in the last 2 commits.

I hesitated about it and kept it as before, agreeing with the change.

PR #4480, opened 2023-07-25
Add custom errors to docs

### LINKED ISSUE #4477: Include custom errors in documentation
### Description

After OpenZeppelin Contracts 5.0 and with the addition of custom errors, there are some use cases of listing the custom errors in the documentation. We should consider adding them to the docs in a (maybe) similar way to how we show events docs.

ernestognw:
This PR adds custom error documentation to the docs templates. Also, it adds EIP-6093 to the interfaces documentation given the amount of references to it. Most of the error links would be broken otherwise.

Fixes #4477 

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/contracts/interfaces/draft-IERC6093.sol
+++ b/contracts/interfaces/draft-IERC6093.sol
@@ -3,8 +3,7 @@ pragma solidity ^0.8.19;
 
 /**
  * @dev Standard ERC20 Errors
- * Interface of the ERC6093 custom errors for ERC20 tokens
- * as defined in https://eips.ethereum.org/EIPS/eip-6093[the EIP]
+ * Interface of the https://eips.ethereum.org/EIPS/eip-6093[ERC-6093] custom errors for ERC20 tokens.
  */
 interface IERC20Errors {
     /**
@@ -50,8 +49,7 @@ interface IERC20Errors {
 
 /**
  * @dev Standard ERC721 Errors
- * Interface of the ERC6093 custom errors for ERC721 tokens
- * as defined in https://eips.ethereum.org/EIPS/eip-6093[the EIP]
+ * Interface of the https://eips.ethereum.org/EIPS/eip-6093[ERC-6093] custom errors for ERC721 tokens.
  */
 interface IERC721Errors {
     /**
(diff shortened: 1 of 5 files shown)


contracts/interfaces/draft-IERC6093.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.19;
 /**
  * @dev Standard ERC20 Errors
  * Interface of the ERC6093 custom errors for ERC20 tokens
- * as defined in https://eips.ethereum.org/EIPS/eip-6093
+ * as defined in https://eips.ethereum.org/EIPS/eip-6093[the EIP]
```
Amxx, line 7: I'd write it as 
```
* Interface of the https://eips.ethereum.org/EIPS/eip-6093[ERC-6093] custom errors for ERC20 tokens
```

ernestognw: I agree, extending it was too repetitive, and didn't notice. Updated

Amxx:
This looks good to me if we want to add the errors in the interface page, next to the other interface functions... but maybe we should have a specific, separate, section. WDYT ?

ernestognw:
> This looks good to me if we want to add the errors in the interface page, next to the other interface functions... but maybe we should have a specific, separate, section. WDYT ?

Ideally, I'd like to give the same importance to all interfaces listed and this alternative seems a bit unbalanced:

Captura de pantalla 2023-07-25 a la(s) 15 04 03


What's your idea for having a specific place for them?

frangio:
In `AccessControl` we should mention the error that is emitted, as discussed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4475. Can we add it in this PR?

frangio (approved):
Looks nice. I'm fine with the current structure of the interfaces page. We could probably do better than this but I don't have ideas atm and I wouldn't block merging on that.

ernestognw:
> In `AccessControl` we should mention the error that is emitted, as discussed in #4475. Can we add it in this PR?

Of course!

frangio (approved):

PR #4481, opened 2023-07-25
Make ERC2771Context return original sender address if `msg.data.length <= 20`

ernestognw:
Fixes an audit finding. Report will be published soon.

The current implementation of the ERC2771Forwarder doesn't check that the `msg.data` has a length of at least the ERC2771 suffixed address, nullifying the result and producing the `address(0)` whenever accessing an out-of-bounds `calldata` location (i.e. calldata with size 19).

This PR checks for the correct length and falls back to the original sender.

### DIFF
--- a/.changeset/unlucky-beans-obey.md
+++ b/.changeset/unlucky-beans-obey.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-`ERC2771Context`: Return the forwarder address whenever `msg.data.length` is not appended with the request signer address (i.e. length is less than 20 bytes).
+`ERC2771Context`: Return the forwarder address whenever the `msg.data` or calls sent by a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes).
--- a/.changeset/unlucky-beans-obey.md
+++ b/.changeset/unlucky-beans-obey.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-`ERC2771Context`: Return the forwarder address whenever the `msg.data` of a call originating from a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes).
+`ERC2771Context`: Return the forwarder address whenever the `msg.data` of a call originating from a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes), as specified by ERC-2771.
--- a/.changeset/unlucky-beans-obey.md
+++ b/.changeset/unlucky-beans-obey.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-`ERC2771Context`: Return the forwarder address whenever the `msg.data` or calls sent by a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes).
+`ERC2771Context`: Return the forwarder address whenever the `msg.data` of a calls originating from a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes).
--- a/.changeset/unlucky-beans-obey.md
+++ b/.changeset/unlucky-beans-obey.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/unlucky-beans-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`ERC2771Context`: Return the forwarder address whenever `msg.data.length` is not appended with the request signer address (i.e. length is less than 20 bytes).
```
Amxx, line 5: I'm not sure what "`msg.data.length` is not appended with" means. The address is appended to te msg.data, and the msg.data.length increases by 20.

```suggestion
`ERC2771Context`: Return the forwarder address whenever the `msg.data` or calls sent by a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes).
```

test/metatx/ERC2771Context.test.js
```diff
@@ -79,6 +81,15 @@ contract('ERC2771Context', function (accounts) {
...
         await expectEvent.inTransaction(tx, ERC2771ContextMock, 'Sender', { sender: this.sender });
       });
+
+      it('returns the original sender when calldata length is less than 20 bytes (address length)', async function () {
+        // The forwarder doesn't produce calls with calldata length less than 20 bytes
+        const recipient = await ERC2771ContextMock.new(anotherAccount);
+
+        const { tx } = await recipient.msgSender({ from: anotherAccount });
+
+        await expectEvent.inTransaction(tx, ERC2771ContextMock, 'Sender', { sender: anotherAccount });
```
Amxx, line 91: In that case the recipient is called directly, so you can do
```suggestion
        const { receipt } = await recipient.msgSender({ from: anotherAccount });

        await expectEvent(receipt, 'Sender', { sender: anotherAccount });
```
(which is faster)

Amxx (dismissed):

ernestognw:
I thought accepting a suggestion doesn't dismiss reviews, but it does.
Can you reapprove @Amxx?

.changeset/unlucky-beans-obey.md
frangio: The current description is a bit misleading because it sounds like we can know whether the request signer address is appended, but the only thing we can know is if the length is greater than 20 bytes.

```suggestion
`ERC2771Context`: Return the forwarder address whenever `msg.data` is smaller than 20 bytes as specified by ERC-2771, since the request signer address can't be extracted.
```

frangio: Reviewing the new changeset...

ernestognw: Note that the suggestion you're making is missing that the change only applies when called via the forwarder. Now that I think about it, it seems that the forwarder address is always returned if the calldata length is less than 20 bytes.

.changeset/unlucky-beans-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`ERC2771Context`: Return the forwarder address whenever the `msg.data` of a call originating from a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes).
```
frangio, line 5:
```suggestion
`ERC2771Context`: Return the forwarder address whenever the `msg.data` of a call originating from a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes), as specified by ERC-2771.
```

ernestognw: Thoughts on?

```suggestion
`ERC2771Context`: Return the forwarder address whenever the `msg.data` of a call originated from a trusted forwarder is not long enough to contain the request signer address (i.e. `msg.data.length` is less than 20 bytes), as specified by ERC-2771
```

Amxx: Either is good for me. When you reach concensus you can merge without my formal approval.

ernestognw: No issues 👌

frangio (approved):

Amxx (approved):

frangio:
For future observers, the title of this PR says `<= 20` but it should have been `< 20`. The code does the latter (correctly).

PR #4487, opened 2023-07-29
Fix issues in the ERC4646Fee documentation

Amxx:
#4476 introduced some example comments that are not correct.


frangio (approved):

ernestognw (commented):
Late to the party but yes, this is correct.

PR #4489, opened 2023-07-31
Update pragma to 0.8.20

Amxx:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4460#discussion_r1278425684

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v4.9.0) (utils/Arrays.sol)
 
-pragma solidity ^0.8.19;
+pragma solidity ^0.8.20;
 
 import {StorageSlot} from "./StorageSlot.sol";
 import {Math} from "./math/Math.sol";
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -57,7 +57,7 @@ library Arrays {
     function unsafeAccess(address[] storage arr, uint256 pos) internal pure returns (StorageSlot.AddressSlot storage) {
         bytes32 slot;
         // We use assembly to calculate the storage slot of the element at index `pos` of the dynamic array `arr`
-        // following https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
+        // following https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
 
         /// @solidity memory-safe-assembly
         assembly {
@@ -75,7 +75,7 @@ library Arrays {
     function unsafeAccess(bytes32[] storage arr, uint256 pos) internal pure returns (StorageSlot.Bytes32Slot storage) {
         bytes32 slot;
         // We use assembly to calculate the storage slot of the element at index `pos` of the dynamic array `arr`
-        // following https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
+        // following https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
 
         /// @solidity memory-safe-assembly
         assembly {
@@ -93,7 +93,7 @@ library Arrays {
     function unsafeAccess(uint256[] storage arr, uint256 pos) internal pure returns (StorageSlot.Uint256Slot storage) {
         bytes32 slot;
         // We use assembly to calculate the storage slot of the element at index `pos` of the dynamic array `arr`
-        // following https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
+        // following https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
 
         /// @solidity memory-safe-assembly
         assembly {
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -57,7 +57,7 @@ library Arrays {
     function unsafeAccess(address[] storage arr, uint256 pos) internal pure returns (StorageSlot.AddressSlot storage) {
         bytes32 slot;
         // We use assembly to calculate the storage slot of the element at index `pos` of the dynamic array `arr`
-        // following https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
+        // following https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
 
         /// @solidity memory-safe-assembly
         assembly {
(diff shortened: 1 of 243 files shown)


ernestognw (dismissed):
LGTM 🚢

contracts/utils/Arrays.sol
```diff
@@ -57,7 +57,7 @@ library Arrays {
     function unsafeAccess(address[] storage arr, uint256 pos) internal pure returns (StorageSlot.AddressSlot storage) {
         bytes32 slot;
         // We use assembly to calculate the storage slot of the element at index `pos` of the dynamic array `arr`
-        // following https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
+        // following https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
```
Amxx, line 60: I think we don't use latest here in case the docs ever change in a way that would make the link invalid. For example the tag could change. I'd be ok with changing that to `v0.8.20`.

frangio: Yeah I prefer using a fixed version.

frangio (approved):

contracts/utils/Arrays.sol
ernestognw: We have a few others using `latest`. I think we had this conversation in another PR and we decided to do `latest` iirc. No problem for now, just flagging in case it's worth discussing a better strategy.

frangio: I would vote for using a fixed version in all links to Solidity docs.

ernestognw: I'd vote for consistency, and I'd push for using `0.8.x` or something that doesn't fixes the docs to an specific patch.

pcaversaccio:
Hey guys, could you quickly clarify the need for `0.8.20`? My recommendation I always give is to only use Solidity versions `>=0.8.20` for EVM networks that support the opcode `PUSH0` since solc defaults to `shanghai` for these versions. And the OZ library can't enforce the `paris` EVM version anyways, so I'm afraid that people will use it and deploy code on non-`PUSH0` supporting EVM chains that either will fail at deployment time already or later on runtime.

Amxx:
We want to benefit from some of the 0.8.20 (and beyond) features. In particular, better support for natspec comments.

Release of our code, in particular compiled bytecode for the upgrade plugin is an issue with PUSH0. IMO we should stick with paris for the artefacts we provide (at least for a while).

If the issue is about devs depending on our lib, using a recent version of the compiler, comilling with default settings, and deploying to chains that don't support PUSH0 ... I'd say its on them for not testing in a proper test network (all chains and L2 have a test duplicate that you should use before you go live). This is going to happen anyway, regardless of the pragma we use.

pcaversaccio:
>  This is going to happen anyway, regardless of the pragma we use.

Not if you use version `0.8.19`. But in any case, please add warnings in the docs as well as in the OpenZeppelin wizard about the `PUSH0`.

Amxx:
Right, are you proposing that we prevent ourself and our knowledgable users from using new, more efficient, version of the compiler, in order to protect the less knowledgable users?

pcaversaccio:
> Right, are you proposing that we prevent ourself and our knowledgable users from using new, more efficient, version of the compiler, in order to protect the less knowledgable users?

No, I only propose to make the default compilation of the OpenZeppelin library "noob-safe" until `shanghai` is mainstream. Optimsim e.g. doesn't even support `PUSH0` yet. Do you really believe that the majority of devs using OZ contracts are the knowledgeable ones? I understand that this is not a released version yet, I'm just raising potential issues if that version is released soon.

JorgeAtPaladin:
Documenting this sounds like a good middle ground

Amxx:
> Do you really believe that the majority of devs using OZ contracts are the knowledgeable ones?

That is something that causes many internal dissagrements :P

I'm personally expecting people that target polygon, to test their entier app (contracts + indexer + front +...) on mumbai first. Is that too much to expect?

pcaversaccio:
> I'm personally expecting people that target polygon, to test their entier app (contracts + indexer + front +...) on mumbai first. Is that too much to expect?

Well, yeah that's how it should work, but expectations usually don't meet reality. I just want to remind you that there was just recently an issue on zkSync where a project deployed a contract that uses `transfer` instead of `call` to transfer native tokens out, but since zkSync uses AA with additional logic for which you need gas, every withdraw call reverted with OOG. They unit tested their code, they even fork-tested on mainnet, but didn't fork test on zkSync.

Anyways, I personally think we should highlight this potentially breaking issue everywhere where devs interact in some sense with the OZ library, e.g. GitHub, Docs, NPM side, and the OZ Wizard.

frangio:
Do we need to document this issue if the pragma remains ^0.8.19?

I'm interested in helping prevent these issues but I'm not even confident that people will see these warnings. And the issue is much larger than OpenZeppelin unfortunately.

Amxx:
> but didn't fork test on zkSync

When you fork test, do you get the behavior or the forked chain, or just the state? Like if polygon doesn't have PUSH0, and I fork polygon using foundry or hardhat, does my local fork detect that the forked chain is still doesn't have PUSH0, or does it run à Shanghai node with the state of the remote chain?

pcaversaccio:
> Do we need to document this issue if the pragma remains ^0.8.19?

In any case, I would document it properly. Ultimately, it's all about education.

> I'm interested in helping prevent these issues but I'm not even confident that people will see these warnings. And the issue is much larger than OpenZeppelin unfortunately.

Agreed, and I don't know how to solve it really.

> When you fork test, do you get the behavior of the forked chain, or just the state? Like if polygon doesn't have PUSH0, and I fork polygon using foundry or hardhat, does my local fork detect that the forked chain doesn't have PUSH0, or does it run a Shanghai node with the state of the remote chain?

My understanding (and I can be wrong) is that fork tests are local simulations with accounts & states and storage slot information fetched from the RPC. I.e. you don't get the full chain behaviour and thus can't detect that specific issue.

I think the desired behaviour would look like the following:
- Default compilation of any OZ contract defaults to the evm version `paris`
- If dev wants to change it, he needs to adjust the config

E.g. Foundry does [default](https://github.com/foundry-rs/foundry/blob/master/config/src/lib.rs#L1732) to `paris` and not `shanghai`.

Amxx:
> The question is whether OZ fully delegates that responsibility to the tooling side at the end of the day.

We, and other libraries like PRB math, already do that. We don't control the users development environment. We don't control the users compiler settings. If we were providing a `hardat.config.js` for reuse we would indeed have a responsability, but that is not the case.

We do provide compiled artefacts in the npm package, which are used by our upgrades plugin. That part we control, and IMO it should be targetting paris (see #4503).

pcaversaccio:
> We do provide compiled artefacts in the npm package, which are used by our upgrades plugin. That part we control, and IMO it should be targetting paris (see https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4503).

That's great. But you don't think any further documentation is necessary at all?

frangio:
I've opened an issue to set the Hardhat default to paris:

- https://github.com/NomicFoundation/hardhat/issues/4232

With this change I would feel comfortable keeping the pragma at 0.8.20 and I think there would be less of a need for documentation.

pcaversaccio:
I feel we got a good compromise.

Amxx:
> But you don't think any further documentation is necessary at all?

We could for sure use more documentation. I'm not sure where we would put that though. #4503 is a clear actionable. Documentation would also be nice, its just not that to clear me what the actionnable is.

pcaversaccio:
In the UI of the OpenZeppelin Wizard and maybe in the [Security Center](https://contracts.openzeppelin.com/security) somewhere?

PR #4492, opened 2023-07-31
update Governor mappings from private to internal

mattstam:
### Description 

Changes the `_proposals` and `_governanceCall` mappings to internal from private.

### Rationale

For many of the useful overrides that you'd want to make on the Governor, accessing this state is essential. For example, it is impractical to override `propose()` if you do not have access to the storage for proposals.

This leads to inefficient and dangerous workarounds that could otherwise be avoided by making this storage internal.

#### PR Checklist

- [X] Tests
- [X] Documentation

### DIFF
--- a/.changeset/young-bats-eat.md
+++ b/.changeset/young-bats-eat.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+update Governor mappings from private to internal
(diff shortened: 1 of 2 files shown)


Amxx:
Hello @mattstam 

[Has per our guidelines, all storage should be private](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/GUIDELINES.md#solidity-conventions).

PR #4494, opened 2023-07-31
Optimize muldiv

vladyan18:
This pull request contains two small changes to the mulDiv function to slightly optimize gas.

#### Reduce stack operations in mulDiv
The product of x and y to obtain the least significant bits of the result is moved above, to the place where the variable prod0 is declared.

https://github.com/vladyan18/openzeppelin-contracts/blob/e54956d8453f04f59491511df8d353dbaab98202/contracts/utils/math/Math.sol#L127
```solidity
uint256 prod0 = x * y; // Least significant 256 bits of the product
uint256 prod1; // Most significant 256 bits of the product
assembly {
  let mm := mulmod(x, y, not(0))
  prod1 := sub(sub(mm, prod0), lt(mm, prod0))
}
```

**Rationale**: with compiler optimization enabled, the old implementation led to unnecessary initialization of the variable by zero. As a result, this initialization also added unnecessary stack operations below in the code. Since this code is in an unchecked block, calculating the product in assembly does not provide gas savings by itself.

Gas tests, small inputs. Optimizer 200 runs, 0.8.20
**Gas before**: 198
**Gas after**: 190 (-4%)

#### Simplify twos calculation in mulDiv

An easier method was used to factor powers of two out of denominator due to which the number of necessary operations is reduced by one.

https://github.com/vladyan18/openzeppelin-contracts/blob/e54956d8453f04f59491511df8d353dbaab98202/contracts/utils/math/Math.sol#L165
```solidity
uint256 twos = denominator & (0 - denominator);
```
**Rationale**: the previous implementation uses a unary negation operation and ends up needing more operations to get the same result. The proposed line gives the same result inside an unchecked block with fewer operations.

Gas tests, big inputs. Optimizer 200 runs, 0.8.20
**Gas before**: 497
**Gas after**: 494 (-0.6%)

#### PR Checklist

- [x] Tests
- [x] Documentation

_Probably new tests and documentation are not required for this pull request._

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -162,8 +162,7 @@ library Math {
             // Factor powers of two out of denominator and compute largest power of two divisor of denominator. Always >= 1.
             // See https://cs.stackexchange.com/q/138556/92363.
 
-            // Does not overflow because the denominator cannot be zero at this stage in the function.
-            uint256 twos = denominator & (~denominator + 1);
+            uint256 twos = denominator & (0 - denominator);
             assembly {
                 // Divide denominator by twos.
                 denominator := div(denominator, twos)
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -124,11 +124,10 @@ library Math {
             // 512-bit multiply [prod1 prod0] = x * y. Compute the product mod 2^256 and mod 2^256 - 1, then use
             // use the Chinese Remainder Theorem to reconstruct the 512 bit result. The result is stored in two 256
             // variables such that product = prod1 * 2^256 + prod0.
-            uint256 prod0; // Least significant 256 bits of the product
+            uint256 prod0 = x * y; // Least significant 256 bits of the product
             uint256 prod1; // Most significant 256 bits of the product
             assembly {
                 let mm := mulmod(x, y, not(0))
-                prod0 := mul(x, y)
                 prod1 := sub(sub(mm, prod0), lt(mm, prod0))
             }
 
--- a/.changeset/pink-suns-mix.md
+++ b/.changeset/pink-suns-mix.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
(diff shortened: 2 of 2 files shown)


contracts/utils/math/Math.sol
```diff
@@ -124,11 +124,10 @@ library Math {
             // 512-bit multiply [prod1 prod0] = x * y. Compute the product mod 2^256 and mod 2^256 - 1, then use
             // use the Chinese Remainder Theorem to reconstruct the 512 bit result. The result is stored in two 256
             // variables such that product = prod1 * 2^256 + prod0.
-            uint256 prod0; // Least significant 256 bits of the product
+            uint256 prod0 = x * y; // Least significant 256 bits of the product
```
Amxx, line 127: does that change anything ?

contracts/utils/math/Math.sol
```diff
@@ -163,8 +162,7 @@ library Math {
             // Factor powers of two out of denominator and compute largest power of two divisor of denominator. Always >= 1.
             // See https://cs.stackexchange.com/q/138556/92363.
 
-            // Does not overflow because the denominator cannot be zero at this stage in the function.
-            uint256 twos = denominator & (~denominator + 1);
+            uint256 twos = denominator & (0 - denominator);
```
Amxx, line 165: That is indeed a good solution !

Amxx (approved):
These are great findings !

contracts/utils/math/Math.sol
Amxx: For the record, before solidity v8, the code used to be [`uint256 twos = -denominator & denominator;`](https://github.com/Uniswap/v3-core/blob/412d9b236a1e75a98568d49b1aeb21e3a1430544/contracts/libraries/FullMath.sol#L64) but solidity 0.8 forbid the `-x` syntax for unsigned integers. Doing `(0-x)` indeed fixes it.

frangio (approved):

ernestognw (approved):
Thanks!

PR #4495, opened 2023-07-31
Refactor Governor proposal struct for efficient access

vladyan18:
This PR prevents the entire ProposalCore structure from being cached into memory. 

https://github.com/vladyan18/openzeppelin-contracts/blob/8866e573f50c7a1cad23a5e47c297ee8cb47b1dc/contracts/governance/Governor.sol#L158-L160
```solidity
// ProposalCore is just one slot. We can load it from storage to stack with a single sload
(bool proposalExecuted, bool proposalCanceled) = (proposal.executed, proposal.canceled);
```

The previous implementation loaded the entire struct into memory for caching, with only the two fields of the struct being used later on. Although only two fields are required further, writing the structure to memory requires allocating a full word for each field of the structure, resulting in suboptimal gas consumption.

According to the gas-report, in most tests using this method, the change results in savings of about 200 gas.

#### PR Checklist

- [X] Tests
- [X] Documentation


Amxx:
The ProposalCore struct is only one slot, so it can be cached with a single sload. Your proposal sets a storage pointer then access it twice, which costs 2 sload.

vladyan18:
> The ProposalCore struct is only one slot, so it can be cached with a single sload.

Yes, the structure takes up one slot and requires one SLOAD. However, the memory does not pack the structure fields, so the memory allocates a full word for each field of the structure.

> Your proposal sets a storage pointer then access it twice, which costs 2 sload.

Actually no, in my version the compiler reduces it to one SLOAD. I left a pointer to a storage to explicitly avoid recalculating the corresponding memory slot (hashing). In fact, you can also leave this to the compiler and remove this line.

.changeset/popular-rice-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Optimized the use of the ProposalCore structure in the state getter
```
Amxx, line 5: I'm not sure we want a changelog for that, considering this is already new 5.0 code.

If we want a changelog entry, it would have to be more in line with the format we ususally use

ernestognw: Yeah, I'd agree not to have a changelog entry in this case.

ernestognw (dismissed):
The savings come from not allocating memory, this makes sense to me because the stack assignments can be already performed by the storage pointer.

Thanks!

Amxx:
> The savings come from not allocating memory, this makes sense to me because the stack assignments can be already performed by the storage pointer.

And not "depacking" from sotrage to memory. That is something we (at least I) forget ... I'm wondering if we do that kind of things elsewhere.

Amxx (dismissed):

frangio:
This is a great observation.

I was doing some experiments (with simple code) and it doesn't seem necessary to write the assignment in one line using tuples. Something like:

```solidity
bool proposalExecuted = proposal.executed;
bool proposalCanceled = proposal.canceled;
```

appears to be similarly optimized.

vladyan18:
> and it doesn't seem necessary to write the assignment in one line using tuples. 

Yeah, absolutely. It's just that I personally prefer to use this syntax in simple cases to emphasize that this is one SLOAD. But in case you prefer on different lines, it is in most cases exactly the same.

frangio:
Note: We just merged https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4360 into `master` where among other changes we had refactored the Proposal struct to optimize access assuming that it was loaded to memory as a single word. We have to follow up with this PR (after updating it as necessary) to simplify it back since that assumption was invalid.

vladyan18:
I think that here is similar situation:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/21716722ad78ee3e9bf21a5b47ff420323c8fe27/contracts/governance/extensions/GovernorVotesQuorumFraction.sol#L52-L55

Amxx:
After my merge + struct refactor, the gas number don't look good. I'm wondering if that may be because it checked against an older snapshot of the gas costs. Just in case, I'm re-running the tests.

If that is not it, we may have to revert te struct merge part

contracts/governance/Governor.sol
```diff
@@ -311,19 +305,20 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
 
-        _proposals[proposalId].core = ProposalCore({
+        _proposals[proposalId] = ProposalCore({
             proposer: proposer,
             voteStart: SafeCast.toUint48(snapshot),
             voteDuration: SafeCast.toUint32(duration),
             executed: false,
-            canceled: false
+            canceled: false,
+            eta: 0
```
frangio, line 321: It's probably because this is now writing 2 slots. We should try writing this as separate assignments for each member.

Amxx: Like 
```solidity
proposal.proposer = proposer;
proposal.voteStart = SafeCast.toUint48(snapshot);
proposal.voteDuration = SafeCast.toUint32(duration);
```
?

I thought that was the one thing not to do.

Amxx: I did the commit, it works. 

I feel like I'm no longer understand the compiler. All the things we used to do for gas optimisation now turn out to be less efficient than the stuff we moved away from.

Amxx (approved):

contracts/governance/Governor.sol
vladyan18: > I guess the compiler just got better

Seem like `eta` is in separate slot in storage. So plain struct initialization writes “0” to this slot, since you need to initialize `eta` field. In your new approach compiler does not touch this slot at all

frangio (approved):
Thanks!

Upgradeable tests are failing as expected due to storage layout change, ok for 5.0.

PR #4497, opened 2023-07-31
Refactor abs without logical branching

vladyan18:
The calculation of the absolute value of a signed number has been replaced with a version without logical branching. 
Based on [**open-source** formula from "Bit Twiddling Hacks" by By Sean Eron Anderson](https://graphics.stanford.edu/~seander/bithacks.html#IntegerAbs).

```solidity
/**
 * @dev Returns the absolute unsigned value of a signed value.
 */
function abs(int256 n) internal pure returns (uint256) {
  unchecked {
    // Formula from the "Bit Twiddling Hacks" by Sean Eron Anderson
    int256 mask = n >> 255;
    return uint256((n + mask) ^ mask);
  }
}
```

**Rationale:** logical branching requires the use of additional rather expensive opcodes. The proposed version is cheaper. In addition, it seems that a small pure function without logical branching can often be placed by the compiler optimizer directly into the place it is used, which will further reduce the overhead.

#### Gas checks

Optimizer 200 runs, solidity 0.8.20

The results are highly dependent on compilation settings and surrounding code. The most conservative difference obtained is:

Old version, positive number (100): 70
Old version, negative number (-100): 85

New version, positive number (100): 48
New version, negative number (-100): 48

#### PR Checklist

- [X] Tests
- [X] Documentation

### DIFF
--- a/.changeset/popular-carpets-search.md
+++ b/.changeset/popular-carpets-search.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`SignedMath`: Refactor `abs` without logical branching
--- a/contracts/utils/math/SignedMath.sol
+++ b/contracts/utils/math/SignedMath.sol
@@ -36,8 +36,9 @@ library SignedMath {
      */
     function abs(int256 n) internal pure returns (uint256) {
         unchecked {
-            // must be unchecked in order to support `n = type(int256).min`
-            return uint256(n >= 0 ? n : -n);
+            // Formula from the "Bit Twiddling Hacks" by Sean Eron Anderson
+            int256 mask = n >> 255;
+            return uint256((n + mask) ^ mask);
         }
     }
 }


frangio:
Thanks! This is cool. Would you mind adding a property test in `Math.t.sol`, or better yet a new file `SignedMath.t.sol`?

Amxx:
@vladyan18 

In the ressource you linked, there is another interresting entry

> Compute the minimum (min) or maximum (max) of two integers without branching

Have you measured the possible savings for this ?

vladyan18:
> @vladyan18
> 
> In the ressource you linked, there is another interresting entry
> 
> > Compute the minimum (min) or maximum (max) of two integers without branching
> 
> Have you measured the possible savings for this ?

I tried a little to explore this point in my project. So far I've come to the conclusion that this only makes sense if the compiler will inline the function.
But maybe it's worth looking into more.

Amxx:
I had a quick look, the typing is preventing a lot of the necessary tricks. We would have to make it assembly.

vladyan18:
> I had a quick look, the typing is preventing a lot of the necessary tricks. We would have to make it assembly.

I experimented like this:
https://github.com/cryptoalgebra/Algebra/blob/c071b615a28be0a9a18ff2b839025757b3632103/src/core/contracts/libraries/PriceMovementMath.sol#L17

Amxx:
> > I had a quick look, the typing is preventing a lot of the necessary tricks. We would have to make it assembly.
> 
> I experimented like this: https://github.com/cryptoalgebra/Algebra/blob/c071b615a28be0a9a18ff2b839025757b3632103/src/core/contracts/libraries/PriceMovementMath.sol#L17

According to remix estimations, this saves 33 gas: 319→ 285 gas

The assembly readability is not great. But it would be encapsulated/isolated, and we would do fuzzing tests for it.

@frangio, @ernestognw : how do you feel about this ?

Amxx:
> **UPD: my version isn't working with` type(int256).min`**

That is a big no IMO.

Amxx:
I'd focus on the abs that works (we need to make sure it does → through fuzzing)

min an max we can discuss in another PR later.

frangio:
I edited the average test because it was underspecifying the rounding behavior. I also added a simpler average test to complement it but it even though it doesn't test the full range.

Amxx:

Amxx:
Gas report shows how that reduces the size of the code produced. It fails however at estimating execution cost.

test/utils/math/SignedMath.t.sol
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+            assertTrue(result == b);
+        }
+    }
+
+    // AVERAGE
+    // 1. simple test, not full int256 range
+    function testAverage1(int256 a, int256 b) public {
+        a = bound(a, 0, type(int256).max / 2);
+        b = bound(b, 0, type(int256).max / 2);
```
Amxx, line 36: I don't see a reason to only consider positive values.
```suggestion
        a = bound(a, type(int256).min / 2, type(int256).max / 2);
        b = bound(b, type(int256).min / 2, type(int256).max / 2);
```

Amxx (dismissed):

ernestognw:
Also testing with foundry confirms savings:

```diff
diff --git a/.gas-snapshot b/.gas-snapshot
index 08d0ca59d..5fa058674 100644
--- a/.gas-snapshot
+++ b/.gas-snapshot
@@ -1,5 +1,5 @@
-SignedMathTest:testAbs(int256) (runs: 5000, μ: 449, ~: 440)
+SignedMathTest:testAbs(int256) (runs: 5000, μ: 397, ~: 395)
```

ernestognw (approved):
Just added comments to the method. Functionality remains the same.

Amxx (approved):

PR #4499, opened 2023-08-01
optimize function setTo in library BitMaps

gzliudan:
The current code of function `setTo` in library `BitMaps` is as below:

```solidity
    /**
     * @dev Sets the bit at `index` to the boolean `value`.
     */
    function setTo(BitMap storage bitmap, uint256 index, bool value) internal {
        if (value) {
            set(bitmap, index);
        } else {
            unset(bitmap, index);
        }
    }
```

When the parameter `value` is 
- `true`: call function `set` to set value to 1.
- `false`: call function `unset` to set value to 0.

If the new value equals the old value, this behavior will waste gas. This PR compares the old and new values first, if they are same then do nothing.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


.changeset/silly-students-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+optimize function setTo in library BitMaps
```
Amxx, line 5:
```suggestion
`BitMaps`: optimize `setTo` if value does not change.
```

contracts/utils/structs/BitMaps.sol
```diff
@@ -33,10 +33,12 @@ library BitMaps {
...
-        if (value) {
-            set(bitmap, index);
-        } else {
-            unset(bitmap, index);
+        uint256 bucket = index >> 8;
+        uint256 mask = 1 << (index & 0xff);
+        bool old = bitmap._data[bucket] & mask != 0;
+
+        if (old != value) {
+            bitmap._data[bucket] ^= mask;
```
Amxx, line 41: We should avoid duplicating sload.
```suggestion
        uint256 data = bitmap._data[bucket];
        bool oldValue = data & mask != 0;

        if (oldValue != value) {
            bitmap._data[bucket] = data ^ mask;
        }
```

gzliudan: Yes, I will resubmit this PR today later.

.changeset/silly-students-agree.md
gzliudan: Updated.

contracts/utils/structs/BitMaps.sol
gzliudan: Updated !

vladyan18:
Maybe it's better to move this check here (and to `unset`)?
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5ae630684a0f57de400ef69499addab4c32ac8fb/contracts/utils/structs/BitMaps.sol#L46-L50

Something like:
```solidity
uint256 block = bitmap._data[bucket];
if (block & mask != mask) {
  bitmap._data[bucket] = block | mask;
}
```

Then in `setTo `it will be possible to leave the old syntax

Amxx:
That would be fine, and it would also reduce the cost of set/unset no-ops.

contracts/utils/structs/BitMaps.sol
```diff
@@ -46,7 +46,11 @@ library BitMaps {
     function set(BitMap storage bitmap, uint256 index) internal {
         uint256 bucket = index >> 8;
         uint256 mask = 1 << (index & 0xff);
-        bitmap._data[bucket] |= mask;
+        uint256 chunk = bitmap._data[bucket];
+
+        if (chunk & mask != mask) {
```
Amxx, line 51: This would also work, and may be cheaper ?
```suggestion
        if (chunk & mask == 0) {
```

Amxx:

Amxx:
Gas report show that this changes actually increass both the contract size and the execution cost:

Capture d’écran du 2024-01-18 16-07-23

I think swe should close that PR. @ernestognw WDYT?

ernestognw:
> I think swe should close that PR. @ernestognw WDYT?

I agree given the gas report, but I'm not sure why. In theory we should be saving gas but the gas increase seems a bit too high to what it was before

gzliudan:
Maybe the data is related to test cases.

gzliudan:
Perhaps the test results are related to the use cases. If the new and old data are the same, it will reduce gas. If the new and old data are different, it will increase gas. But overall, if data is normal, it should save gas.

gzliudan:
@Amxx I added some test cases. Could you please create a gas report again?

Amxx:
My guess is the increase comes from the cost of the if branching

Amxx:
> @Amxx I added some test cases. Could you please create a gas report again?

Report is produced automatically by the CI, but I'm now realizing they are not public. Here is the latest one (with the extra tests you added)

Capture d’écran du 2024-01-19 11-47-24

gzliudan:
I update test cases again. If the gas cost is not improved, then we can close this PR.

Amxx:
file:///home/amxx/Images/Captures%20d%E2%80%99%C3%A9cran/Capture%20d%E2%80%99%C3%A9cran%20du%202024-01-19%2013-59-23.png

PR #4502, opened 2023-08-02
Implement recommendations from 5.0 audit Phase 1B 

ernestognw:
Full audit report will be out later.

Fixes LIB-985
Fixes LIB-987
Fixes LIB-988
Fixes LIB-989
Fixes LIB-990
Fixes LIB-994
Fixes LIB-995
Fixes LIB-1007
Fixes LIB-1006
Fixes LIB-1005
Fixes LIB-1004
Fixes LIB-1001
Fixes LIB-1003
Fixes LIB-1010


ernestognw:
Force pushes are to make it easier to navigate each fix commit.

contracts/proxy/utils/UUPSUpgradeable.sol
```diff
@@ -126,7 +126,9 @@ abstract contract UUPSUpgradeable is IERC1822Proxiable {
     function _authorizeUpgrade(address newImplementation) internal virtual;
 
     /**
-     * @dev Perform implementation upgrade with security checks for UUPS proxies, and additional setup call.
+     * @dev Performs ab implementation upgrade with a security check for UUPS proxies, and additional setup call.
```
Amxx, line 129:
```suggestion
     * @dev Performs an implementation upgrade with a security check for UUPS proxies, and additional setup call.
```

contracts/finance/VestingWallet.sol
```diff
@@ -154,10 +175,19 @@ contract VestingWallet is Context {
...
         }
     }
+
+    /**
+     * @dev Reverts if the sender is not the {releaser} address.
+     */
+    function _checkReleaser() internal view virtual {
+        if (_msgSender() != releaser()) {
+            revert VestingWalletUnauthorizedReleaser(_msgSender());
+        }
```
Amxx, line 191: That is a major behavior change that I don't remember we agreed on. 

IMO, if we want that (I'm not sold) we need stronger documentation about that change. We also need to provide an easy path to re-implementing the current behavior (that doesn't involve overriding `_checkReleaser` without a super call).

Maybe something like "if releaser is 0, than anyone can release"
```solidity

```

test/finance/VestingWallet.behavior.js
```diff
@@ -22,35 +24,51 @@ function shouldBehaveLikeVesting(beneficiary) {
...
+    it('releases a linearly vested schedule', async function () {
+      let released = web3.utils.toBN(0);
+      const before = await this.getBalance(beneficiary);
+      const releaser = await this.mock.releaser();
 
-    {
-      const receipt = await this.mock.methods[release](...args);
+      {
+        // Allows gas price to be 0 so no ETH is spent in the transaction.
+        await setNextBlockBaseFeePerGas(0);
```
Amxx, line 39: I guess you added that so that the gas cost don't interfer with measuring the balances movement ... but we don't check the balances movements.

In [test/finance/VestingWallet.test.js](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/files#diff-3632407a8b9c87a3131e1d83a32f0397eeba2af61927b061825f67bdba757d1d), we may want to add somthing to the `this.checkRelease` for Eth Vesting

contracts/finance/VestingWallet.sol
Amxx: Before anyone proposes that, I really not a fan of advertizing

```solidity
    function releaser() public view virtual override returns (address) {
        return _msgSender();
    }
```

Amxx: (we need optionals values)

.changeset/fifty-owls-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`Address`: Removed the functions overloaded with a customized error message in favor of a custom error indicating the call has failed. This only applies if the underlying revert reason cannot be bubbled up.
```
frangio, line 5:
```suggestion
`Address`: Removed the ability to customize error messages, a common custom error is always used if the underlying revert reason cannot be bubbled up.
```

contracts/metatx/ERC2771Context.sol
```diff
@@ -18,27 +18,54 @@ abstract contract ERC2771Context is Context {
...
 
-    function _msgSender() internal view virtual override returns (address sender) {
+    /**
+     * @dev Override for `msg.data`. Defaults to the original `msg.sender` whenever
+     * a call is not performed by the trusted forwarder or the calldata length is less than
+     * 20 bytes (an address length).
+     */
+    function _msgSender() internal view virtual override returns (address) {
         if (isTrustedForwarder(msg.sender) && msg.data.length >= 20) {
+            address sender;
```
frangio, line 52: What is the reason to remove the named return value? It makes the code nicer. We tend to use this pattern when the return value is set in assembly.

contracts/metatx/ERC2771Context.sol
```diff
@@ -18,27 +18,54 @@ abstract contract ERC2771Context is Context {
     /// @custom:oz-upgrades-unsafe-allow state-variable-immutable
     address private immutable _trustedForwarder;
 
+    /**
+     * @dev Sets an immutable trusted forwarder.
```
frangio, line 22:
```suggestion
     * @dev Initializes the contract with a trusted forwarder, which will be able to invoke functions on
     * this contract on behalf of other accounts.
```

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -186,31 +196,36 @@ contract ERC2771Forwarder is EIP712, Nonces {
      */
     function _validate(
         ForwardRequestData calldata request
-    ) internal view virtual returns (bool alive, bool signerMatch, address signer) {
-        signer = _recoverForwardRequestSigner(request);
-        return (request.deadline >= block.timestamp, signer == request.from, signer);
+    ) internal view virtual returns (bool unexpired, bool signerMatch, address signer) {
+        (address recovered, bool isValid) = _recoverForwardRequestSigner(request);
+        return (isValid && request.deadline >= block.timestamp, recovered == request.from, recovered);
```
frangio, line 201: Signature validity should be reflected in the `signerMatch` return value.

```suggestion
        return (request.deadline >= block.timestamp, isValid && recovered == request.from, recovered);
```

contracts/metatx/ERC2771Forwarder.sol
frangio, line 1: During the audit we mentioned that it would be a good idea to add `require(target.isTrustedForwarder(this))` in the forward functions. It could also be a part of `_validate`.

I think we need to do this for safety. It also mostly neutralizes the issue of token approvals.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -97,8 +107,8 @@ contract ERC2771Forwarder is EIP712, Nonces {
      * receiver is provided.
      */
     function verify(ForwardRequestData calldata request) public view virtual returns (bool) {
-        (bool alive, bool signerMatch, ) = _validate(request);
-        return alive && signerMatch;
+        (bool unexpired, bool signerMatch, ) = _validate(request);
```
frangio, line 110: Personally I would go with "active" instead of "unexpired".

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -186,31 +196,36 @@ contract ERC2771Forwarder is EIP712, Nonces {
...
+                    request.gas,
+                    nonces(request.from),
+                    request.deadline,
+                    keccak256(request.data)
                 )
-            ).recover(request.signature);
+            )
+        ).tryRecover(request.signature);
+
+        return (recovered, err == ECDSA.RecoverError.NoError);
```
frangio, line 228: Can we put the boolean return value first? It's more common to see it that way.

contracts/proxy/transparent/ProxyAdmin.sol
```diff
@@ -9,14 +9,16 @@ import {Ownable} from "../../access/Ownable.sol";
 /**
  * @dev This is an auxiliary contract meant to be assigned as the admin of a {TransparentUpgradeableProxy}. For an
  * explanation of why you would want to use this see the documentation for {TransparentUpgradeableProxy}.
+ *
+ * NOTE: This contract does not make use of {Context} deliberately.
```
frangio, line 13: I think this comment can be removed from this contract. There is no use of `msg.sender` (except through Ownable, which does use Context).

contracts/proxy/utils/UUPSUpgradeable.sol
```diff
@@ -25,7 +25,7 @@ abstract contract UUPSUpgradeable is IERC1822Proxiable {
      * and `upgradeToAndCall(address,bytes)` are present, and `upgradeTo` must be used if no function should be called,
      * while `upgradeToAndCall` will invoke the `receive` function if the second argument is the empty byte string.
      * If the getter returns `"5.0.0"`, only `upgradeToAndCall(address,bytes)` is present, and the second argument must
-     * be the empty byte string if no function should be called, being impossible to invoke the `receive` function
+     * be the empty byte string if no function should be called, making impossible to invoke the `receive` function
```
frangio, line 28:
```suggestion
     * be the empty byte string if no function should be called, making it impossible to invoke the `receive` function
```

contracts/proxy/utils/UUPSUpgradeable.sol
```diff
@@ -126,7 +126,9 @@ abstract contract UUPSUpgradeable is IERC1822Proxiable {
     function _authorizeUpgrade(address newImplementation) internal virtual;
 
     /**
-     * @dev Perform implementation upgrade with security checks for UUPS proxies, and additional setup call.
+     * @dev Performs an implementation upgrade with a security check for UUPS proxies, and additional setup call.
+     *
+     * The security check calling {proxiableUUID} expects the id to be the ERC1967's implementation.
```
frangio, line 131:
```suggestion
     * As a security check, {proxiableUUID} is invoked in the new implementation, and the return value
     * is expected to be the implementation slot in ERC1967.
```

contracts/utils/Address.sol
```diff
@@ -54,8 +54,10 @@ library Address {
      * plain `call` is an unsafe replacement for a function call: use this
      * function instead.
      *
-     * If `target` reverts with a revert reason, it is bubbled up by this
-     * function (like regular Solidity function calls).
+     * If `target` reverts with a revert reason or custom error, it is bubbled
+     * up by this function (like regular Solidity function calls). However, if
+     * the call reverted with no returned reason, it default revert with an
+     * {FailedInnerCall} error.
```
frangio, line 60:
```suggestion
     * the call reverted with no returned reason, this function reverts with a
     * {FailedInnerCall} error.
```

test/utils/Address.test.js
```diff
@@ -355,28 +329,5 @@ contract('Address', function (accounts) {
...
-        [],
-      );
-    });
-  });
-
-  describe('verifyCallResult', function () {
-    it('returns returndata on success', async function () {
-      const returndata = '0x123abc';
-      expect(await this.mockFnPointer.verifyCallResult(true, returndata)).to.equal(returndata);
-    });
```
frangio, line 372: I think we still have to keep this.

contracts/finance/VestingWallet.sol
frangio, line 1: Let's revert changes to VestingWallet in this PR.

contracts/metatx/ERC2771Context.sol
ernestognw: No concrete reason is provided for the issue other than what the title describes ("Unused Named Return Variables"). I'd agree with reverting this change.

frangio: Yeah let's revert. This variable is used.

contracts/finance/VestingWallet.sol
ernestognw: Removed. Also changed the linear ticket to "in progress"

test/utils/Address.test.js
ernestognw: Re added in [3d09549](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/commits/3d09549bd3cd5110f78979e217d7e2d4e9cf2434)

contracts/utils/Address.sol
ernestognw: Fixed in [3d09549](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/commits/3d09549bd3cd5110f78979e217d7e2d4e9cf2434)

contracts/proxy/utils/UUPSUpgradeable.sol
ernestognw: Fixed in [68963df](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

contracts/proxy/utils/UUPSUpgradeable.sol
ernestognw: Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/commit/68963df79ea364f3ddf8bcd9e70aab119dac45ab

contracts/proxy/transparent/ProxyAdmin.sol
ernestognw: I added it by accident. Removed

contracts/metatx/ERC2771Forwarder.sol
ernestognw: Yes, fixed in [53b4feb](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/commits/53b4febe4267abcf11dfcbfa69c942cd477c79d4)

contracts/metatx/ERC2771Forwarder.sol
ernestognw: Fixed in [7f6da22](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

contracts/metatx/ERC2771Forwarder.sol
ernestognw: I don't think this is necessary. In the case of batching we'd need to do this multiple times, which also increases the risk of reentrancy issues.

contracts/metatx/ERC2771Forwarder.sol
ernestognw: Fixed in 53b4feb

contracts/metatx/ERC2771Context.sol
ernestognw: Fixed in [8309315](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)

contracts/metatx/ERC2771Context.sol
ernestognw: Sure thing, removed in [58546a1](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/commits/58546a17c4e78aedf55aa3ae49e81d421bdca4a6)

.changeset/fifty-owls-retire.md
ernestognw: Fixed along with 316c30c

contracts/finance/VestingWallet.sol
ernestognw: I removed the commit with the `onlyReleaser` since we agreed on discussing further with the research team, let's see what alternative we can agree on :D

test/finance/VestingWallet.behavior.js
ernestognw: This no longer applies because the `onlyReleaser` commit was reverted.

contracts/finance/VestingWallet.sol
frangio, line 1: Let's add a changeset for this?

contracts/finance/VestingWallet.sol
```diff
@@ -154,7 +154,7 @@ contract VestingWallet is Context {
     function _vestingSchedule(uint256 totalAllocation, uint64 timestamp) internal view virtual returns (uint256) {
         if (timestamp < start()) {
             return 0;
-        } else if (timestamp > end()) {
+        } else if (timestamp >= end()) {
```
frangio, line 157: Let's add a changeset for this?

contracts/metatx/ERC2771Forwarder.sol
frangio: I feel pretty strongly that we should add this line. The additional cost should be in the low hundreds of gas per call. It restricts the forwarder to work with targets who accept it as a forwarder, and nullifies the issue of token approvals.

It's a staticcall, what is the reentrancy risk?

contracts/metatx/ERC2771Context.sol
Amxx, line 1: Just for the record, some people did ask for supporting multiple forwarders at the same time, using a `mapping(address => bool)` mapping.

Our answer was that this is storage intensive (sloads) and that we don't want to provide it by default, but that they can easily implement that by overriding `isTrustedForwarder(address)`. With this new design, they can still do that override, but it is unclear how the `trustedForwarder()` getter should be dealt with.

contracts/metatx/ERC2771Forwarder.sol
ernestognw: We discussed this internally and we agreed to make a `require(target.isTrustedForwarder(this))` check to prevent misuse of the forwarder. It was implemented in [this commit](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/commits/42e068ad0783a48664e00336adb4555441061ea3)

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -269,6 +300,56 @@ contract ERC2771Forwarder is EIP712, Nonces {
...
+
+            // Avoid strange returndatasizes (eg. an EOA)
+            if eq(returndatasize(), 0x20) {
+                // Copy from memory and clean the `returnValue` as a boolean.
+                returnValue := shr(255, shl(255, mload(0)))
+            }
+        }
+
+        return success && returnValue;
+    }
```
Amxx, line 351: Do we really need assembly. Whenever possible, I tink we should stick with assembly. If solidity doiesn't provide the guarantees we need, then we should raise an issue.

Amxx: I guess the issue with 
```solidity
(bool success, bytes memory returndata) = target.staticcall(abi.encodeCall(ERC2771Context.isTrustedForwarder, (address(this))));
return success && returndata.length >= 0x20 && abi.decode(returndata, (bool));
```
is the returndata bomb

And the issue with 
```solidity
        if (target.code.length == 0) {
            return false;
        }
        try ERC2771Context(target).isTrustedForwarder(address(this)) returns (bool trusted) {
            return trusted;
        } catch {
            return false;
        }
```
if that target may return nothing ...

Amxx: https://github.com/ethereum/solidity/issues/14467

contracts/proxy/transparent/ProxyAdmin.sol
```diff
@@ -4,19 +4,19 @@
...
 import {ITransparentUpgradeableProxy} from "./TransparentUpgradeableProxy.sol";
-import {Ownable} from "../../access/Ownable.sol";
+import {Ownable2Step, Ownable} from "../../access/Ownable2Step.sol";
 
 /**
  * @dev This is an auxiliary contract meant to be assigned as the admin of a {TransparentUpgradeableProxy}. For an
  * explanation of why you would want to use this see the documentation for {TransparentUpgradeableProxy}.
  */
-contract ProxyAdmin is Ownable {
+contract ProxyAdmin is Ownable2Step {
```
Amxx, line 13: Same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/files#r1284069930

contracts/utils/Address.sol
```diff
@@ -180,47 +127,25 @@ library Address {
                 }
             }
```
Amxx, line 128: Can we clean that and merge the two ifs ?

```solidity
if (returndata.length == 0 && target.code.length == 0) {
    revert AddressEmptyCode(target);
}
```

contracts/utils/Address.sol
```diff
@@ -95,83 +81,44 @@ library Address {
...
         address target,
         bool success,
-        bytes memory returndata,
-        function() internal view customRevert
+        bytes memory returndata
     ) internal view returns (bytes memory) {
-        if (success) {
+        if (!success) {
+            _revert(returndata);
+        } else {
```
Amxx, line 121: In `verifyCallResult` we do

```solidity
if (success) { 
    // case success
} else {
    // case revert
}
```

and here we do

```solidity
if (!success) { 
    // case revert 
} else {
    // case success
}
```

Any reason to not use the same order in both functions ?

contracts/metatx/ERC2771Forwarder.sol
ernestognw: Yeah, is a combination of those but also to avoid the memory leak we've seen in `Address` with the custom `string memory errorMsg` I think. @frangio mentioned it but I never found the issue (never knew if it was in Github)

contracts/proxy/beacon/UpgradeableBeacon.sol
```diff
@@ -4,15 +4,15 @@
...
+import {Ownable2Step, Ownable} from "../../access/Ownable2Step.sol";
 
 /**
  * @dev This contract is used in conjunction with one or more instances of {BeaconProxy} to determine their
  * implementation contract, which is where they will delegate all function calls.
  *
  * An owner is able to change the implementation the beacon points to, thus upgrading the proxies that use this beacon.
  */
-contract UpgradeableBeacon is IBeacon, Ownable {
+contract UpgradeableBeacon is IBeacon, Ownable2Step {
```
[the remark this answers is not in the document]
ernestognw: I'm also wary about this but the research team has strong opinions on this, I'll be in favor of holding back with these favoring a Ownable+2Steps merged design that's leaner as a middle point.

contracts/proxy/transparent/ProxyAdmin.sol
ernestognw: Oh, this is an amazing reason!

contracts/proxy/beacon/UpgradeableBeacon.sol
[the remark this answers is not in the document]
Amxx: With all due respect to the research team, we should have the final world here. If no one in the contract team wants to force the use of Ownable2Step, then we should not use it.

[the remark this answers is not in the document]
frangio: Yes sorry if this was not clear before I do not think we should use Ownable2Step by default anywhere.

contracts/metatx/ERC2771Forwarder.sol
frangio: We do not need assembly for encoding the data though.

frangio: @ernestognw @Amxx Please see https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502/commits/889c11271f704e25b6d5685e9b0b8c36e4695fd3 and let me know what you think.

Amxx: IMO the less assembly the better !

frangio (approved):

Amxx (approved):

ernestognw (commented):
Can't approve myself but LGTM

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -82,6 +93,11 @@ contract ERC2771Forwarder is EIP712, Nonces {
      */
     error ERC2771ForwarderExpiredRequest(uint48 deadline);
 
+    /**
+     * @dev The request targert doesn't trust the `forwarder`.
```
ernestognw, line 97:
```suggestion
     * @dev The request target doesn't trust the `forwarder`.
```

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -269,6 +300,36 @@ contract ERC2771Forwarder is EIP712, Nonces {
...
+    function _isTrustedByTarget(address target) private view returns (bool) {
+        bytes memory encodedParams = abi.encodeCall(ERC2771Context.isTrustedForwarder, (address(this)));
+
+        bool success;
+        uint256 returnSize;
+        uint256 returnValue;
+        /// @solidity memory-safe-assembly
+        assembly {
+            // Because the return value is a boolean (requires 1 byte copied to memory) and the
+            // calldata length is 24 bytes, we can safely reuse the scratch space in memory (0x00 - 0x3F).
```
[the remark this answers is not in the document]
frangio: Ah yes I kept this because I read the first part and though it referred to the returnvalue.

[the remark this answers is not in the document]
ernestognw: Yeah, the `staticcall` also mentions the scratch space but in that case it's correct.

PR #4504, opened 2023-08-02
Use bytes.concat in MessageHashUtils

frangio:
We can use bytes.concat instead of abi.encodePacked here which is clearer.

Updating an existing changeset to mention this PR too.

In `toDataWithIntendedValidatorHash` we keep `abi.encodePacked` because one of the arguments is an address.

### DIFF
--- a/.changeset/flat-bottles-wonder.md
+++ b/.changeset/flat-bottles-wonder.md
@@ -2,4 +2,6 @@
 'openzeppelin-solidity': minor
 ---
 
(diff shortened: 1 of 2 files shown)


Amxx (approved):
FYI, I considered that change in #4296, but was not sure because of the casting.

both
```solidity
keccak256(bytes.concat("\x19Ethereum Signed Message:\n", bytes(Strings.toString(message.length)), message));
```
and
```solidity
keccak256(string.concat("\x19Ethereum Signed Message:\n", Strings.toString(message.length), string(message)));
```

should work. Any reason to favor the first one ?

PR #4508, opened 2023-08-04
Use `Ownable` in `VestingWallet` instead of an immutable beneficiary

ernestognw:
Fixes LIB-996
Fixes LIB-986

Currently, the beneficiary of the VestingWallet is immutable without an option for transferability. This is solved by making the VestingWallet inherit from `Ownable2Step`, also allowing to restrict the `release()` function to only the interested party.

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/finance/VestingWallet.sol
```diff
@@ -31,18 +36,19 @@ contract VestingWallet is Context {
...
     uint64 private immutable _start;
     uint64 private immutable _duration;
 
     /**
-     * @dev Set the beneficiary, start timestamp and vesting duration of the vesting wallet.
+     * @dev Sets the sender as the initial owner, the beneficiary as the pending owner, the start timestamp and the
+     * vesting duration of the vesting wallet.
      */
-    constructor(address beneficiaryAddress, uint64 startTimestamp, uint64 durationSeconds) payable {
+    constructor(address beneficiaryAddress, uint64 startTimestamp, uint64 durationSeconds) payable Ownable(msg.sender) {
```
ernestognw, line 46: I'd like to discuss doing this:

```suggestion
    constructor(address benefactorAddress, address beneficiaryAddress, uint64 startTimestamp, uint64 durationSeconds) payable Ownable(benefactorAddress) {
```

I think it has a few extra implications we need to consider, but I think this is definitely the way to go because a `VestingWallet` is the kind of contract that might be deployed with immutable clones. In such cases the deployer won't be able to claim the tokens back in case they're never claimed.

I guess the best solution to this is to allow setting a benefactor address (perhaps with a different name)

Amxx (commented):
That is an interresting approach, but a very opinionated one. I don't feel confortable doing that change for our main `VestingWallet` contract.

Basically, this feels like a preset. Something interresting that is built on top of our building blocks. Something that we may want to provide as a "ready to use to address interresting usecases" but not as "the defaut vesting wallet building block"

What I would do:
- `VestingWallet` is abstract. It does not have any beneficiary in storage. It has an abstract `function beneficiary() public view return (address);` unimplemented function. It also has no protection on the release functions.
- A `VestingWalletClaimable is VestingWallet, Ownable2Step`

```solidity
constructor VestingWalletClaimable(address beneficiaryAddress, address recoveryAddress, uint64 startTimestamp, uint64 durationSeconds) 
    payable 
    VestingWallet(startTimestamp, durationSeconds) 
    Ownable(recoveryAddress) 
{
    transferOwnership(beneficiaryAddress);
}

function beneficiary() public view virtual override returns (address) {
    return owner();
}

function release() public virtual override onlyOwner {
    super.release();
}

function release(address token) public virtual override onlyOwner {
    super.release(token);
}
```

contracts/finance/VestingWallet.sol
Amxx: you can present it as a "recovery address" in case the beneficiary doesn't claim ownership of the vesting

```suggestion
    constructor(address beneficiaryAddress, address recoveryAddress, uint64 startTimestamp, uint64 durationSeconds) payable Ownable(recoveryAddress) {
```

Amxx: is "claiming owership of the vesting a taxable event"?

contracts/finance/VestingWallet.sol
```diff
@@ -43,11 +42,10 @@ contract VestingWallet is Context, Ownable2Step {
      * @dev Sets the sender as the initial owner, the beneficiary as the pending owner, the start timestamp and the
      * vesting duration of the vesting wallet.
      */
-    constructor(address beneficiaryAddress, uint64 startTimestamp, uint64 durationSeconds) payable Ownable(msg.sender) {
-        if (beneficiaryAddress == address(0)) {
+    constructor(address beneficiary, uint64 startTimestamp, uint64 durationSeconds) payable Ownable(beneficiary) {
+        if (beneficiary == address(0)) {
```
Amxx, line 46: maybe not in the PR, but we should remove that check here and move it to `Ownable`'s constructor

.changeset/healthy-gorillas-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+`VestingWallet`: Use Ownable2Step instead of an immutable `beneficiary`. The initial owner is set to the benefactor (`msg.sender`) but transferred to the beneficiary address so that unclaimed tokens can be recovered by the benefactor.
```
Amxx, line 5: this is no longer accurate.

test/utils/Create2.test.js
```diff
@@ -59,7 +59,12 @@ contract('Create2', function (accounts) {
         addr: offChainComputed,
       });
 
-      expect(await VestingWallet.at(offChainComputed).then(instance => instance.beneficiary())).to.be.equal(other);
+      const instance = await VestingWallet.at(offChainComputed);
+
+      // Needs to be accepted
+      await instance.acceptOwnership({ from: other });
```
Amxx, line 65: That is no longer relevant, right?

ernestognw:
Yes, I'm also not in favor of this change. We prepared it as a response for the audit so we can evaluate it.

My opinion is also to not make this change because it's very easy to implement by inheritance and it's also forcing users into a behavior.

Playing devil's advocate, I really like the design of a pending owner that needs to claim the VestingWallet. The concept of recovering may be interesting to explore, but doesn't justify adding Ownable2Step imo.

contracts/finance/VestingWallet.sol
ernestognw: I agree we should reconsider this. I opened an issue
https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4511

contracts/finance/VestingWallet.sol
ernestognw: I don't think it is, because a user claiming the ownership doesn't have access to transfer the tokens yet.

ernestognw (commented):
Looks good to me again

Amxx (approved):

PR #4512, opened 2023-08-05
Refactor access to Checkpoint struct without using memory

vladyan18:
Fixes mentioned in:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4495#issuecomment-1664394485

#### PR Checklist

- [X] Tests
- [X] Documentation

### DIFF
--- a/contracts/governance/extensions/GovernorVotesQuorumFraction.sol
+++ b/contracts/governance/extensions/GovernorVotesQuorumFraction.sol
@@ -49,9 +49,11 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
         uint256 length = _quorumNumeratorHistory._checkpoints.length;
 
         // Optimistic search, check the latest checkpoint
-        Checkpoints.Checkpoint224 memory latest = _quorumNumeratorHistory._checkpoints[length - 1];
-        if (latest._key <= timepoint) {
-            return latest._value;
+        Checkpoints.Checkpoint224 storage latest = _quorumNumeratorHistory._checkpoints[length - 1];
+        uint32 latestKey = latest._key;
+        uint224 latestValue = latest._value;
+        if (latestKey <= timepoint) {
+            return latestValue;
         }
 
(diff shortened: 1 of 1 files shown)


vladyan18:
I would also like in this PR to slightly change some lines in:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/Checkpoints.sol

Since in this contract the structure (-s) is repeatedly written to the memory. Also, I suspect that [such structure reading](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/54a235f8959ecffb1916cf5693ec9bbd695cbf71/contracts/utils/structs/Checkpoints.sol#L98) can create an overhead (since `self` is read multiple times). 

However, this may require changing relatively many lines of code, and I'm not sure that such changes are desired for version 5.0. So I would like to hear the opinion of the maintainers on this.

frangio (approved):
Looks good, gas is slightly impproved.

frangio:
> I would also like in this PR to slightly change some lines in: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/Checkpoints.sol
> 
> Since in this contract the structure (-s) is repeatedly written to the memory. Also, I suspect that [such structure reading](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/54a235f8959ecffb1916cf5693ec9bbd695cbf71/contracts/utils/structs/Checkpoints.sol#L98) can create an overhead (since `self` is read multiple times).
> 
> However, this may require changing relatively many lines of code, and I'm not sure that such changes are desired for version 5.0. So I would like to hear the opinion of the maintainers on this. I'm not sure if it's worth creating a separate issue for this, but if necessary, I can create one.

Open to this experiment if you want to give it a shot, but agree with doing it for 5.1. Feel free to open a PR though we may just keep it open until post-5.0.

Amxx (approved):

PR #4516, opened 2023-08-07
Some fixes and improvements to AccessManager

Amxx:
- In canCall, return `(false, 0)` instead of `(false, delay)` if the caller is in the middle of the grant delay, with execution delay set
- Enforce execution delay when operating on a family operation
- Early exit in `_canCallExtended` (to avoid some sloads)

TODO: `_checkFamilyDelay` should check the execution delay of the caller, but that requires merging this modifier with the ADMIN_GROUP check

#### PR Checklist

- [ ] Tests
- [ ] Documentation


frangio:
This was superceded by https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4518

PR #4519, opened 2023-08-08
Remove outdated warning from ERC2771Forwarder

frangio:
Since we added `_isTrustedByTarget` we no longer need this warning.

### DIFF
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -27,10 +27,6 @@ import {Address} from "../utils/Address.sol";
  * transactions in the mempool. In these cases the recommendation is to distribute the load among
  * multiple accounts.
  *
- * WARNING: Do not approve this contract to spend tokens. Anyone can use this forwarder
- * to execute calls with an arbitrary calldata to any address. Any form of approval may
- * result in a loss of funds for the approving party.
- *
  * NOTE: Batching requests includes an optional refund for unused `msg.value` that is achieved by
  * performing a call with empty calldata. While this is within the bounds of ERC-2771 compliance,
  * if the refund receiver happens to consider the forwarder a trusted forwarder, it MUST properly


ernestognw:
I disagree with removing it. Perhaps needs updating but I think there's a non-negligible risk of approving a forwarder (or an upgradeable contract that may become a forwarder) trusted by some tokens out there. Consider we actually keep recommending using `ERC2771Context` so the probability of users interacting with assets configured to trust a forwarder is not low imo.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -27,10 +27,6 @@ import {Address} from "../utils/Address.sol";
  * transactions in the mempool. In these cases the recommendation is to distribute the load among
  * multiple accounts.
  *
- * WARNING: Do not approve this contract to spend tokens. Anyone can use this forwarder
- * to execute calls with an arbitrary calldata to any address. Any form of approval may
- * result in a loss of funds for the approving party.
- *
  * NOTE: Batching requests includes an optional refund for unused `msg.value` that is achieved by
```
ernestognw, line 30: I'd go with this

```suggestion
* WARNING: Do not approve this contract to spend tokens. Anyone can use this forwarder
* to execute calls with an arbitrary calldata to any trustee contract (i.e. contracts trusting
* this forwarder). Any form of approval may result in a loss of funds for the approving party.
```

ernestognw (changes requested):

frangio:
If the forwarder is trusted I expect that the contract will use `ERC2771Context._msgSender`, and in this case allowance given to the forwarder would not be at risk. In the case of `transferFrom`, the spender would be the request signer and not the forwarder itself.

ernestognw:
Ok you're right, I think I focused on too much assembly I forgot the initial intention 😅

ernestognw (approved):

PR #4522, opened 2023-08-09
Change _approve argument name in ERC721 

### LINKED ISSUE #4457: Change _approve argument name in ERC721
In the functions approve and _approve in the ERC20 the areguments are (address to, uint tokenId). I suggest to change the name of the arguments to (address spender, uint tokenId) to be more intuitive.

PS: the name used in the EIP is _approved but I find this confusing because approved could be the agent approved to transfer or the token that is being approved

RenanSouza2:
Fixes #4457

Uses the 'operator' as argument to stay consistent with other functions

#### PR Checklist

- [X] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/loud-coins-love.md
+++ b/.changeset/loud-coins-love.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Uses operator as name argument to stay consistent with other functions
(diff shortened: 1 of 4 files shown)


RenanSouza2:
I made some slightly bigger changes to the tests, let me know what you think of them

frangio:
Thank you @RenanSouza2, but my interpretation of the ERC is that "operator" is reserved for "approved for all" accounts, and not for tokenId-specific approvals. For example this is the description of the behavior of `safeTransferFrom`:

>Throws unless `msg.sender` is the current owner, an authorized operator, or the approved address for this NFT.

A distinction is drawn between the "operator" and the "approved address for this NFT".

PR #4523, opened 2023-08-10
Add Governor module connecting with AccessManager

frangio:
This module allows a Governor instance to use contracts restricted by an AccessManager.


contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+     */
+    function propose(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        string memory description
+    ) public virtual override returns (uint256) {
+        uint256 proposalId = super.propose(targets, values, calldatas, description);
+
+        uint32 maxDelay = baseDelaySeconds();
```
ernestognw, line 82: If I'm interpreting correctly this should be the min delay

```suggestion
        uint32 minDelay = baseDelaySeconds();
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+
+        ExecutionPlan storage plan = _executionPlan[proposalId];
+
+        for (uint256 i = 0; i < targets.length; ++i) {
+            (address manager, uint32 delay) = _detectExecutionRequirements(targets[i], bytes4(calldatas[i]));
+            plan.managers.push(IAccessManager(manager));
+            // downcast is safe because both arguments are uint32
+            maxDelay = uint32(Math.max(delay, maxDelay));
```
ernestognw, line 90: Because here we're getting the `max`, the lowest it can go is `maxDelay` so:

```suggestion
            minDelay = uint32(Math.max(delay, minDelay));
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+        delete _executionPlan[proposalId];
+
+        return proposalId;
+    }
+
+    /**
+     * @dev Check if the execution of a call needs to be performed through an AccessManager and what delay should be
+     * applied to this call.
+     *
+     * Returns { manager: address(0), delay: _baseDelaySeconds() } if:
```
ernestognw, line 181: We're returning 0 here because `baseDelaySeconds` is checked anyway in `propose`

```suggestion
     * Returns { manager: address(0), delay: 0 } if:
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+    ) internal virtual override returns (uint256) {
+        uint256 proposalId = super._cancel(targets, values, calldatas, descriptionHash);
+
+        ExecutionPlan storage plan = _executionPlan[proposalId];
+
+        for (uint256 i = 0; i < targets.length; ++i) {
+            IAccessManager manager = plan.managers[i];
+            if (address(manager) != address(0)) {
+                // Attempt to cancel, but allow to revert if a guardian cancelled part of the proposal
+                try manager.cancel(address(this), targets[i], calldatas[i]) {} catch {}
```
ernestognw, line 168: I'm not sure if this is the best idea because we're deleting the execution line a few lines later. I'd suggest checking `getSchedule` in the catch block:

```suggestion
                try manager.cancel(address(this), targets[i], calldatas[i]) {} catch {
                  require(manager.schedule(targets[i]) == 0);
                }
```

However, I see how this can block the canceling. We should at least document that we assume the operation is considered canceled in the manager.

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        if (authority != address(0)) {
+            // Check if governor can call according to authority, and try to detect a delay
+            (bool authorized, uint32 delay) = AuthorityUtils.canCallWithDelay(authority, address(this), target, selector);
+
+            // If direct call is not authorized, and delayed call is possible
+            if (!authorized && delay > 0) {
+                return (authority, delay);
+            }
```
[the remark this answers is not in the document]
ernestognw: Ok I see the `authority` is not returned here so the call is made directly by the governor in `_executeOperations`. The worst that can happen in the scenario I propose is that the call goes through the `relay` function

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+ * @dev TODO
+ */
+abstract contract GovernorTimelockAccess is Governor {
+    struct ExecutionPlan {
+        uint32 delay;
+        IAccessManager[] managers;
+    }
+
+    mapping(uint256 => ExecutionPlan) private _executionPlan;
+    mapping(address target => address) private _authorityOverride;
```
ernestognw, line 25: I don't see currently a way of writing this variable, flagging but I'm assuming it's WIP

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+        if (block.timestamp < eta) {
+            revert GovernorUnmetDelay(proposalId, eta);
+        }
+
+        ExecutionPlan storage plan = _executionPlan[proposalId];
+
+        for (uint256 i = 0; i < targets.length; ++i) {
+            IAccessManager manager = plan.managers[i];
+            if (address(manager) != address(0)) {
+                manager.relay{value: values[i]}(targets[i], calldatas[i]);
```
ernestognw, line 141: The proposal will get stuck if one of the operations was canceled in the manager blocks the execution of others so it's a DoS vector.

Let's:

```suggestion
                try manager.relay{value: values[i]}(targets[i], calldatas[i]) {} catch {
                  manager.getSchedule(<operation id>);
                }
```

I'm realizing there's no way of getting the manager's operation id, is it?

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+     * Returns { manager: address(0), delay: _baseDelaySeconds() } if:
+     * - target does not have code
+     * - target does not implement IAccessManaged
+     * - calling canCall on the target's manager returns a 0 delay
+     * - calling canCall on the target's manager reverts
+     * Otherwise (calling canCall on the target's manager returns a non 0 delay), return the address of the
+     * AccessManager to use, and the delay for this call.
+     */
+    function _detectExecutionRequirements(address target, bytes4 selector) private view returns (address, uint32) {
+        address authority = _authorityOverride[target];
```
ernestognw, line 190: This will be zero at the current state. Solve after adding a way of updating the `authorityOverride`.

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: `schedule` returns the operation id.

> The proposal will get stuck if one of the operations was canceled in the manager blocks the execution of others so it's a DoS vector.

Yes, I am not convinced that your proposal is the right thing to do. Proposals have been atomic so far and this changes that.

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: What's the issue this suggestion is addressing?

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+    mapping(address target => address) private _authorityOverride;
+
+    uint32 _baseDelay;
+
+    error GovernorUnmetDelay(uint256 proposalId, uint256 neededTimestamp);
+
+    event BaseDelaySet(uint32 oldBaseDelaySeconds, uint32 newBaseDelaySeconds);
+
+    constructor(uint32 initialBaseDelay) {
+        _baseDelay = initialBaseDelay;
```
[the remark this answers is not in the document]
frangio: Very much so. The right thing is to use `_setBaseDelay` though.

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+     */
+    function setBaseDelaySeconds(uint32 newBaseDelay) public virtual onlyGovernance {
+        _setBaseDelaySeconds(newBaseDelay);
+    }
+
+    /**
+     * @dev Change the value of {baseDelaySeconds}. Internal function without access control.
+     */
+    function _setBaseDelaySeconds(uint32 newBaseDelay) internal virtual {
+        emit BaseDelaySet(_baseDelay, newBaseDelay);
```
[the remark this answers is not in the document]
frangio: It simplifies the code because we need to include the old delay in the event. If we emit after setting, we need to store the previous value in a temporary variable.

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    function _executeOperations(
+        uint256 proposalId,
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 /* descriptionHash */
+    ) internal virtual override {
+        uint256 eta = proposalEta(proposalId);
```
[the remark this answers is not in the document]
frangio: `proposalEta` docs say that it _doesn't_ use the clock:
>Unlike {proposalSnapshot} and {proposalDeadline}, this doesn't use the governor clock, ...

There are no plans to add `clock()` support because AccessManager is timestamp-based.

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: But we are using `Math.max` below. What makes you think it should be min?

frangio: Oh I see, your reasoning is the same as mine, but you read the name as "the minimum delay we need to wait", whereas I've been reading it as "the maximum delay required for any operation".

I think I prefer `maxDelay` because it is consistent with `Math.max`. What do you think?

contracts/governance/extensions/GovernorTimelockAccess.sol
ernestognw: The issue is that by ignoring the return/revert of the `manager.cancel`, we're taking the operation as cancelled, when in reality something else may have happened

frangio: Right. There is an assumption here that `manager.cancel` only reverts if the operation has already been cancelled. It's a pretty strong assumption that it would make sense to validate in the code.

contracts/governance/extensions/GovernorTimelockAccess.sol
ernestognw: Actually, my rule of thumb is to name the fixed variable as the opposite of the operation.
Let's imagine X is constant and Y is variable

- `Math.max(X, Y)` goes for the max value, which can't be *less* than X
- `Math.min(X, Y)` goes for the min value, which can't be *more* than X

>  I've been reading it as "the maximum delay required for any operation"

Can you elaborate on this? I can't follow why is the maximum delay required, it'd mean that less delay is possible, is that the case?

frangio: Less delay may be possible for a subset of the proposal operations. But the whole package, atomically, can only be executed after all operations' delays are satisfied. That means we have to wait the max. But I think we agree on this because you haven't questioned the use of `Math.max`.

frangio: I don't think it's a good idea to rename the variable to `minDelay` because `minDelay = Math.max(minDelay, otherDelay)` seems wrong to me

I am open to renaming to `neededDelay` though.

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: Here's how I've decided to address this risk: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4523/commits/3820637fb66a00ae0f6bfa03cac3a3c22199b410

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: Here's how I've decided to address this risk: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4523/commits/3820637fb66a00ae0f6bfa03cac3a3c22199b410

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: Resolving in favor of the other similar comment.

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,216 @@
...
+
+        // Get authority from target contract if there is no override
+        if (authority == address(0)) {
+            bool success;
+            bytes memory returndata;
+
+            (success, returndata) = target.staticcall(abi.encodeCall(IAccessManaged.authority, ()));
+            if (success && returndata.length >= 0x20) {
+                authority = abi.decode(returndata, (address));
+            }
```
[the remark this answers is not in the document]
frangio: >any addition outside of the if statements can be executed

And why would that be an issue necessarily?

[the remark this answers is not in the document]
ernestognw: A line changed only show +- 3 lines in Github. It's easy to miss that the path where the target is not `IAccessManaged`.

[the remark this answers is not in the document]
frangio: I can make the same argument if there is a change to the final return statement that should affect all branches, but the diff is only showing the last return statement so the reviewer doesn't realize that there are more return statements that should be changed.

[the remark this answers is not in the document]
ernestognw: Yes. I mentioned there were a few alternatives. I was thinking on using naming the return values and using them to have a single consistent return but it didn't go as I expected.

I agree this is the best arrangement now.

contracts/access/manager/AccessManager.sol
```diff
@@ -568,44 +575,57 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
-        uint48 prevTimepoint = _schedules[operationId];
+        uint48 prevTimepoint = _schedules[operationId].timepoint;
         if (prevTimepoint != 0 && !_isExpired(prevTimepoint)) {
             revert AccessManagerAlreadyScheduled(operationId);
         }
 
-        uint48 timepoint = when == 0 ? minWhen : when;
-        _schedules[operationId] = timepoint;
-        emit OperationScheduled(operationId, timepoint, caller, target, data);
+        nonce = _schedules[operationId].nonce + 1;
```
ernestognw, line 627: What do you think?

```suggestion
        unchecked {
          nonce = _schedules[operationId].nonce + 1;
        }
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -17,16 +17,18 @@ import {Time} from "../../utils/types/Time.sol";
...
     struct ExecutionPlan {
+        uint16 length;
         uint32 delay;
-        OperationDetails[] operations;
+        // We use mappings instead of arrays because it allows us to pack values in storage more tightly without storing
+        // the length redundantly.
+        // We pack 8 operations' data in each bucket. Each uint32 value is set to 1 upon proposal creation if it has to
+        // be scheduled and relayed through the manager. Upon queuing, the value is set to nonce + 1, where the nonce is
+        // that which we get back from the manager when scheduling the operation.
+        mapping (uint256 bucket => uint32[8]) managerData;
```
ernestognw, line 27: I'd agree with `bucket` if this was a generic library, but it can be contextualized:

```suggestion
        mapping (uint256 operationId => uint32[8]) managerData;
```

Thoughts?

contracts/access/manager/AccessManager.sol
frangio: I'm not sure. It's 32 bits, it could overflow in theory.

contracts/governance/extensions/GovernorTimelockAccess.sol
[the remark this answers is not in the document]
ernestognw: My bad I think I read that pretty quick but you're actually right. Thanks for confirming

contracts/access/manager/AccessManager.sol
ernestognw: Yes, but you need `(1 << 32) - 1` operationId collisions, which are difficult to mine and I don't see anyone overflowing this. Even doing it intentionally doesn't seem problematic for me. Even with a 1 day full-governance cycle (propose+vote+queue), it'll take ~11 million years.

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,264 @@
...
+        plan.delay = neededDelay;
+
+        return proposalId;
+    }
+
+    /**
+     * @dev Function to queue a proposal to the timelock.
+     *
+     * NOTE: execution delay is estimated based on the delay information retrieved in {proposal}. This value may be
+     * off if the delay were updated during the vote.
```
ernestognw, line 133:
```suggestion
     * off if the delay was updated during the vote.
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -0,0 +1,264 @@
...
+            (bool delayed, uint32 nonce) = _getManagerData(plan, i);
+            if (delayed) {
+                uint32 relayedNonce = _manager.relay{value: values[i]}(targets[i], calldatas[i]);
+                if (relayedNonce != nonce) {
+                    // TODO: custom error
+                    revert();
+                }
+            } else {
+                (bool success, bytes memory returndata) = targets[i].call{value: values[i]}(calldatas[i]);
+                Address.verifyCallResult(success, returndata);
```
ernestognw, line 180: We need to thoroughly document that this may get stuck if one of the operations fail. We should do this before the audit

ernestognw (commented):
Looks really good. Of course we still need tests and some docs but I think it's ready solidity-wise

contracts/access/manager/AccessManager.sol
frangio: Note that this is in AccessManager so it doesn't need to go through governance cycle. If it was a 10 minute delay (the smallest that I think might be reasonable), you could schedule once every 10 minutes, how long would it take to overflow?

ernestognw: True, this is not related to governance. In the case of an execution every 10 minutes it'll take ~81,000 years (`2**32*10/60/24/365`), it'll be even ~8,000 years if it's every minute or ~1634 years if done every block of 12 seconds (`2**32*12/60/60/24/365`)

frangio: Adding a comment to explain use of unchecked following our guidelines:

```suggestion
        unchecked {
            // Overflowing the nonce would take over 1000 years
            nonce = _schedules[operationId].nonce + 1;
        }
```

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: Resolving in favor of your newest comment.

contracts/governance/Governor.sol
```diff
@@ -208,6 +208,13 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         return _proposals[proposalId].eta;
     }
 
+    /**
+     * @dev See {IGovernor-proposalEta}.
+     */
```
ernestognw, line 213:
```suggestion
    /**
     * @dev See {IGovernor-proposalNeedsQueuing}.
     */
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -96,6 +96,14 @@ abstract contract GovernorTimelockAccess is Governor {
         return (delay, indirect);
     }
 
+    /**
+     * @dev See {IGovernor-proposalEta}.
+     */
```
ernestognw, line 101:
```suggestion
    /**
     * @dev See {IGovernor-proposalNeedsQueuing}.
     */
```

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: Added in contract-level docs.

PR #4527, opened 2023-08-15
Migrate FV specs to CVL2

ernestognw:
Fixes LIB-1018

### Description

With the new version of Certora's Verification Language (CVL2), we're updating the specs we already have.
Some of the most important changes are mainly casting related. There are now 3 ways of casting into a type:

- `assert_uint256`: Reports as a counterexample everything not fitting into a `uint256`
- `require_uint256`:  Ignores everything not fitting into a `uint256`
- `to_mathint`: Only apply to `mathint`, `bytes32` and other safe casts

The previous behavior we had in CVL1 was mostly that of the `require_<type>`, so now there are some changes replacing with `assert_<type>` including sanity checks.

There are some improvements to be made but the purpose of this PR is to complete the migration to CVL2

### FV Checklist

- [x] AccessControl
- [x] AccessControlDefaultAdminRules
- [x] ERC20
- [x] Ownable
- [x] Initializable
- [x] Ownable2Steps
- [x] Pausable
 - [x] TImelockController
- [x] ERC721
- [x] ERC20Wrapper
- [x] ERC20Flashmint
- [x] EnumerableSet
- [x] EnumerableMap

#### PR Checklist

- [ ] Tests
- [ ] Documentation


certora/harnesses/ERC20WrapperHarness.sol
```diff
@@ -2,10 +2,15 @@
 
 pragma solidity ^0.8.20;
 
-import "../patched/token/ERC20/extensions/ERC20Wrapper.sol";
+import {ERC20Permit} from "../patched/token/ERC20/extensions/ERC20Permit.sol";
```
Amxx, line 5: Do we need to add ERC20Permit in this harness ?

certora/specs/ERC20FlashMint.spec
```diff
@@ -1,29 +1,29 @@
...
-    flashFeeReceiver() returns (address) envfree
+    function flashFeeReceiver() external returns (address) envfree;
 
     // function summaries below
-    _mint(address account, uint256 amount)              => specMint(account, amount)
-    _burn(address account, uint256 amount)              => specBurn(account, amount)
-    _transfer(address from, address to, uint256 amount) => specTransfer(from, to, amount)
+    function _._mint(address account, uint256 amount)              internal => specMint(account, amount)        expect void ALL;
+    function _._burn(address account, uint256 amount)              internal => specBurn(account, amount)        expect void ALL;
+    function _._transfer(address from, address to, uint256 amount) internal => specTransfer(from, to, amount)   expect void ALL;
```
Amxx, line 12: How is that affected by `_update` ?

certora/specs/EnumerableMap.spec
```diff
@@ -48,7 +48,7 @@ invariant indexedContained(uint256 index)
     {
         preserved {
             requireInvariant consistencyIndex(index);
-            requireInvariant consistencyIndex(to_uint256(length() - 1));
+            requireInvariant consistencyIndex(require_uint256(length() - 1));
```
Amxx, line 51: I'd do `assert_uint256` just to be sure. `length()` should never return anything bigger than 2**129-1.

certora/specs/ERC20FlashMint.spec
frangio: It looks like we should be "overriding" `_update` instead of these 3 functions.

One thing I don't understand is, is this completely replacing the logic of these three functions? It sounds like we would want to add ghosts on top of the default functionalit but if I understand correctly we're replacing the internal logic completely.

Amxx:
ERC721: done

certora/harnesses/ERC20WrapperHarness.sol
ernestognw: It's needed because the `ERC20Wrapper.spec` imports the `ERC20.spec`, which imports `IERC2612.spec`. Although that spec is not run for the wrapper, the functions declared for `IERC2612` are expected there. The easiest workaround I found was to add permit to the harness.

Is there any alternative you can think of?

Amxx: My understanding was that it was ok to import an interface that the target contract does not implement. Any rule that includes a call to these function would be skipped. In this case that means that running the ERC20.spec on the ERC20Wrapper would not check all of the properties, `rule permit` and `rule onlyHolderOfSpenderCanChangeAllowance` would be skipped.

Did that change in CVL2?

ernestognw: Yes, it changed and now the alternative is to add an `optional` annotation. More [here](https://docs.certora.com/en/latest/docs/cvl/cvl2/changes.html?highlight=bytes32#optional-methods-block-entries).

The issue is not because the rules may include calls to functions that aren't implemented, but because the `envfreeFuncsStaticCheck`. Previously in CVL1 any unknown method was just skipped but since CVL2 they're indeed checked.

Adding the `optional` annotation is an alternative but in that case we may accidentally skip it in the `ERC20.spec` as well.

certora/specs/ERC20FlashMint.spec
ernestognw: > is this completely replacing the logic of these three functions?

Yes.

I replaced these 3 summaries with only one `_._update` summary and it worked. Also tried removing it but the counter-example is similar to what we've seen before where the contract starts with arbitrary storage values. I have a hunch that we can remove this summarization by using a few `require` statements, but as far as I'm concerned, the summarization is fine because `_update` is already proven in the ERC20 spec.

certora/specs/ERC20.spec
```diff
@@ -1,28 +1,32 @@
...
 │ Ghost & hooks: sum of all balances                                                                                  │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-ghost sumOfBalances() returns uint256 {
-  init_state axiom sumOfBalances() == 0;
+ghost mathint sumOfBalances {
+    init_state axiom sumOfBalances == 0;
+}
+
+hook Sload uint256 balance _balances[KEY address addr] STORAGE {
```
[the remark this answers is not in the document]
ernestognw: I added a comment explaining why it's needed. However, I don't think removing it should be ideal. 

The reason why we need it now seems to be because the casting in CVL2 works differently and previously (iiuc) it avoided any operation overflowing the `uint256` as if we were doing a `require_uint256()` cast. Something I discussed with @frangio is that this change may have uncovered an undesired behavior in this spec.

I couldn't think of an alternative but I'd be happy to simplify this.

certora/harnesses/ERC20WrapperHarness.sol
Amxx: Good to know !

So its between that and `optional`. Either is fine I guess.

The `optional` would possibly be better if we think users will use our specs to validate their code (that may not be Permit). But unless that is something we consider I don't think we need to change the spec.

certora/specs/ERC20.spec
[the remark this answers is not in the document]
Amxx: I'm not sure I understand. You are saying that we used to work with `uint256` (instead of `mathint`) ... and that discarded cases where an overflow would happen.

If that is what we are trying to replicate, why are  we not doing
```require balance <= max_uint256;```
or something like that.

Here the comparison is between `balance` and `sumOfBalance`. Its a way stronger assumption than "no overflow"

File: certora/specs/ERC20.spec [hunk 5]
```diff
@@ -21,6 +21,12 @@ ghost mathint sumOfBalances {
     init_state axiom sumOfBalances == 0;
 }
 
+// Because `balance` has a uint256 type, any balance addition in CVL1 behaved as a `require_uint256()` casting,
+// leaving out the possibility of overflow. This is not the case in CVL2 where casting became more explicit.
+// A counterexample in CVL2 is having an initial state where Alice initial balance is larger than totalSupply, which
```
Amxx, line 26: Counterexample to what ? to a rule ? an invariant ?

We should have an invariant that we can require and that would say that Alice's balance cannot be bigger than the totalSupply.

Said otherwise, that is a property we want to prove, not assume.

certora/specs/ERC20.spec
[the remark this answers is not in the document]
ernestognw: Here's [a run](https://prover.certora.com/output/6876/316c194b9cd448b4a56a8e5238434087?anonymousKey=13b47c009f652fe97457fa50466df049f7807fea) without the new Sload hook added.

Many rules fail. I'll take the example of the `transfer` run. The initial state is the following:

Captura de pantalla 2023-09-08 a la(s) 9 27 59

And the example is doing a `ERC20PermitHarness.transfer(to=0x2712, value=15)` where sender is `0x4c8`.
Because `0x2712` has `0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff6` balance, when adding `15`, the result is a balance of 5 because the uint256 was overflown.

However, the rule is checking that:

```
assert to_mathint(balanceOf(recipient)) == (recipientBalanceBefore + (holder == recipient ? 0 : amount));
```

where `recipientBalanceBefore` is `0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff6`.

Consider that since CVL2, [mathematical operations return a `mathint`](https://docs.certora.com/en/latest/docs/cvl/cvl2/changes.html?highlight=bytes32#mathematical-operations-return-mathint), therefore the `recipientBalanceBefore + amount` will result in `0x10000000000000000000000000000000000000000000000000000000000000005`.

When comparing, it turns out that `0x10000000000000000000000000000000000000000000000000000000000000005 != 5`.

The counterexample seems valid, but I couldn't find why it was not happening in CVL1

> You are saying that we used to work with uint256 (instead of mathint) ... and that discarded cases where an overflow would happen.

This is currently my strongest hypothesis but not 100% sure tbh. The reason I think so is because the casting changes added in CVL2 although they're not explicit about the previous behavior.

Also, any attempt to add an invariant that avoid this will result in a similar result. I tried a few alternatives with invariants but didn't work. Counterexamples were similar to the run I just provided

certora/specs/ERC20.spec
ernestognw: > Counterexample to what ? to a rule ? an invariant ?

To any rule/invariant checking for balance consistency.

I mostly agree with you. I recall trying to define this `require` as an invariant but it was again failing because the invariant can't be proved for a run similar to the one I described in [this convo](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4527#discussion_r1320028525).

Defining an invariant such that `balance[<some address>] <= sumOfBalances` may work for only one address but you'd need to apply it to every address with a `forall` quantifier, but I couldn't manage to make it work.

certora/specs/ERC20.spec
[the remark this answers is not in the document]
Amxx: Clearly, this counter example is the result of the initial state missing some invariant. 

We (devs) would expect the initial state to follow

- `_balances[0x4c8] <= _totalSupply`
- `_balances[0x2712] <= _totalSupply`
- `_balances[0x4c8] + _balances[0x2712] <= _totalSupply` (that last part is more problematic ...)

because there are invariants that we have in mind.

My point is, if these invariants are necessary for the rules to pass, we need to
1. express them as invariant
2. prove them
3. do a `requireInvariant` in the rules that need that assumption (and yes, there will be many).

frangio (approved):

PR #4531, opened 2023-08-17
Move beneficiary zero address check to Ownable

### LINKED ISSUE #4511: Move `VestingWallet`'s beneficiary zero address check to `Ownable`
# Description

Currently, `VestingWallet` makes a check to validate that the owner is not the zero address. However, we consider this check should be performed in the `Ownable` contract instead.

> maybe not in the PR, but we should remove that check here and move it to `Ownable`'s constructor

_Originally posted by @Amxx in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4508#discussion_r1284792005_

NishantKoyalwar:
Fixes #4511 
Fixes LIB-1034

moved beneficiary zero address check to `Ownable`

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/clever-bats-kick.md
+++ b/.changeset/clever-bats-kick.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-`Ownable`: prevent using address(0) as the initial owner.
+`Ownable`: Prevent using address(0) as the initial owner.
--- a/.changeset/clever-bats-kick.md
+++ b/.changeset/clever-bats-kick.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+moved beneficiay zero address check to Ownable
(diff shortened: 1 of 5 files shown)


frangio:
@NishantKoyalwar Can you add a changeset?

---


@Amxx Does this make sense to you? I think it looks good.

Amxx:
The check changes the behavior of Ownable but I think its acceptable. LGTM. 

Just missing tests/changeset

.changeset/clever-bats-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+moved beneficiay zero address check to Ownable
```
Amxx, line 5:
```suggestion
`Ownable`: prevent using address(0) as the initial owner.
```

frangio (approved):

frangio (approved):

PR #4534, opened 2023-08-19
Use namespaced storage for upgradeable contracts

### LINKED ISSUE #2964: Use diamond storage instead of gaps in upgradeable contracts
**🧐 Motivation**
Gaps are problematic and this: https://eips.ethereum.org/EIPS/eip-2535#facets-state-variables-and-diamond-storage seems to be a great idea.

**📝 Details**
The proposed changes look like: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/pull/134 
I think this will improve the usability of upgradeable contracts and avoid a lot of common mistakes that are hard to fix.

frangio:
This PR configures the transpiler to use namespaces instead of gaps in the upgradeable code.

There are two changes to the code that I found necessary. In one case, the transpiled code resulted in a stack too deep compilation error, I assume because the `$` storage pointer takes up a place in the stack. In another case, I had to change a `public immutable` variable into a private variable, because immutables without an override get converted to storage variables and the namespace transformation enforces storage variables to be private. This implies that we can't use public immutables in the codebase in general.

The transpiler code can be found in https://github.com/OpenZeppelin/openzeppelin-transpiler/pull/127.

Fixes #2964

### DIFF
--- a/.changeset/wet-bears-heal.md
+++ b/.changeset/wet-bears-heal.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': major
+---
+
+Upgradeable contracts now use namespaced storage (EIP-7201).
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -5,7 +5,7 @@ runs:
   steps:
     - uses: actions/setup-node@v3
       with:
-        node-version: 14.x
+        node-version: 16.x
     - uses: actions/cache@v3
       id: cache
       with:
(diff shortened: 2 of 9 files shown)


frangio:
I uploaded the result of this transpilation to a branch in openzeppelin-contracts-upgradeable so we can review the resulting changes: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/commit/d16ec788fd0d1d0c544ff701262eee05e84258c3

frangio:
The latest results are in https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/commit/3b8d9b141994dba3f548e8646566b6bf1e5f1282.

Amxx (approved):

mudgen:
With this change OpenZeppelin upgradeable contracts should now be compatible with EIP-2535 Diamonds.   Would be great to see an implementation of it in OpenZeppelin.

PR #4535, opened 2023-08-21
caching result of reading storage variable to save gas

MollyHe5:
Caching the result of reading storage variable $._initialized to stack could reduce one sload, thus saving gas.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


vladyan18:
Hello! Have you checked the difference in gas? It seems to me that the compiler is capable of optimizing this case on its own.

frangio:
There is some improvement according to our gas measurements: https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/5930652468

Close to 100 gas, so it does sound like the compiler is not optimizing the redundant SLOAD.

frangio (approved):

Amxx (approved):

PR #4538, opened 2023-08-24
Remove redundant memory usage in ERC2981 royaltyInfo

0xVolosnikov:
Same change as in #4495 

There is no actual need to use memory for this structure:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/token/common/ERC2981.sol#L62

- [X] Tests
- [X] Documentation

### DIFF
--- a/.changeset/perfect-coins-accept.md
+++ b/.changeset/perfect-coins-accept.md
@@ -1,5 +1,0 @@
----
-'openzeppelin-solidity': minor
----
-
-`ERC2981`: Optimized memory usage in `royaltyInfo`
--- a/.changeset/perfect-coins-accept.md
+++ b/.changeset/perfect-coins-accept.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`ERC2981`: Optimized memory usage in `royaltyInfo`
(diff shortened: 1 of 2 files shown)


Amxx:
Apparently, this change does not affect the runtime cost of the contract.
Capture d’écran du 2024-01-18 15-27-49

It does however reduces the size / deploying cost of the contract. This is nice I guess. 

@ernestognw WDYT?

ernestognw:
I made a quick Foundry Test and applied to both versions:

```
pragma solidity ^0.8.13;

import "forge-std/Test.sol";
import {ERC2981} from "../src/ERC2981.sol";

contract ERC2981Mock is ERC2981 {}

contract ERC2981Test is Test {
    ERC2981 public erc2981;

    function setUp() public {
        erc2981 = new ERC2981Mock();
    }

    function testRoyaltyInfo() public {
        (address receiver, uint256 royaltyAmount) = erc2981.royaltyInfo(0, 100);
        assertEq(receiver, address(0));
        assertEq(royaltyAmount, 0);
    }
}
```

The snapshots results with `forge snapshot --via-ir --optimizer-runs 1000000000 --use 0.8.24 -vvvv` were:

```
ERC2981ModifiedTest:testRoyaltyInfo() (gas: 10060)
ERC2981Test:testRoyaltyInfo() (gas: 10240)
```

So it seems it's saving around ~180 gas units, which is good along with the bytecode size reduction in my opinion

ernestognw (approved):
LGTM. Any reason not to merge @Amxx?

Amxx (approved):

PR #4539, opened 2023-08-25
Use Trace208 in Votes to support ERC6372 clocks

Amxx:
ERC-6372 clocks are 48 bits but our Votes contracts use 32 bit keys. This changes the Votes library to match clocks and use 48 bit keys, to avoid fallible conversions.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/governance/extensions/GovernorVotesQuorumFraction.sol
+++ b/contracts/governance/extensions/GovernorVotesQuorumFraction.sol
@@ -12,9 +12,9 @@ import {Checkpoints} from "../../utils/structs/Checkpoints.sol";
  * fraction of the total supply.
  */
 abstract contract GovernorVotesQuorumFraction is GovernorVotes {
-    using Checkpoints for Checkpoints.Trace224;
+    using Checkpoints for Checkpoints.Trace208;
 
-    Checkpoints.Trace224 private _quorumNumeratorHistory;
+    Checkpoints.Trace208 private _quorumNumeratorHistory;
 
     event QuorumNumeratorUpdated(uint256 oldQuorumNumerator, uint256 newQuorumNumerator);
 
@@ -49,13 +49,13 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
         uint256 length = _quorumNumeratorHistory._checkpoints.length;
 
         // Optimistic search, check the latest checkpoint
-        Checkpoints.Checkpoint224 memory latest = _quorumNumeratorHistory._checkpoints[length - 1];
+        Checkpoints.Checkpoint208 memory latest = _quorumNumeratorHistory._checkpoints[length - 1];
         if (latest._key <= timepoint) {
             return latest._value;
         }
 
         // Otherwise, do the binary search
-        return _quorumNumeratorHistory.upperLookupRecent(SafeCast.toUint32(timepoint));
+        return _quorumNumeratorHistory.upperLookupRecent(SafeCast.toUint48(timepoint));
     }
 
     /**
@@ -102,7 +102,7 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
         }
 
         uint256 oldQuorumNumerator = quorumNumerator();
-        _quorumNumeratorHistory.push(SafeCast.toUint32(clock()), SafeCast.toUint224(newQuorumNumerator));
+        _quorumNumeratorHistory.push(clock(), SafeCast.toUint208(newQuorumNumerator));
 
         emit QuorumNumeratorUpdated(oldQuorumNumerator, newQuorumNumerator);
     }
--- a/contracts/token/ERC20/extensions/ERC20Votes.sol
+++ b/contracts/token/ERC20/extensions/ERC20Votes.sol
@@ -30,10 +30,10 @@ abstract contract ERC20Votes is ERC20, Votes {
      * @dev Maximum token supply. Defaults to `type(uint208).max` (2^208^ - 1).
      *
      * This maximum is enforced in {_update}. It limits the total supply of the token, which is otherwise a uint256,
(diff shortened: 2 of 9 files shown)


contracts/token/ERC20/extensions/ERC20Votes.sol
```diff
@@ -27,10 +27,10 @@ abstract contract ERC20Votes is ERC20, Votes {
     error ERC20ExceededSafeSupply(uint256 increasedSupply, uint256 cap);
 
     /**
-     * @dev Maximum token supply. Defaults to `type(uint224).max` (2^224^ - 1).
+     * @dev Maximum token supply. Defaults to `type(uint208).max` (2^208^ - 1).
      */
-    function _maxSupply() internal view virtual returns (uint224) {
-        return type(uint224).max;
+    function _maxSupply() internal view virtual returns (uint256) {
```
frangio, line 32: Why change this to `uint256`? Using the smaller type prevents returning a number that is too large for the data structure to support it.

```suggestion
    function _maxSupply() internal view virtual returns (uint208) {
```

frangio: `uint208` makes it a type error to override with a function that returns `type(uint256).max`, for example, or any literal that is too large to fit uint208.

Amxx: What is multiple module implement that function with different values (votes limit to 208, but another module would limit to 192 or 128 for some other reason) how would we resolve that ?

If the return types are different it's a mess. If they are both uint256 there may be some hope

frangio: This is not a concrete problem we have currently so I would go with the option that is safer.

Amxx: If that issue happens during the 5.x cycle, would we be able to change it without a major change ?

I don't think overriding this with a bigger function is something anyone would realistically do.

Amxx: Also the only place we use it, we do an implicit upcast to uint256

frangio: If something like this happens in a minor release I think we would just add an internal function with a name other than `_maxSupply` so they wouldn't clash, and `ERC20Votes` could implement that internal function in terms of the existing one.

frangio: If you feel strongly about this I can accept it. The risk if >uint208 is returned is simply that some transfers might revert but it's a recoverable situation.

frangio:
In https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4539/commits/74b8d264a6cee03ada659b0e62e724841f0a13fc I added `continue-on-error` if the PR is labelled as "breaking change". But note that this doesn't re-run the checks when the label is applied. To do that we would need to add these lines in checks.yml but I feared they might cause tests to reset too often:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b2e7bab92044031a495a53b4ccc31c84137f4319/.github/workflows/changeset.yml#L7-L11

Amxx:
> In [74b8d26](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/74b8d264a6cee03ada659b0e62e724841f0a13fc) I added `continue-on-error` for the storage layout checks if the PR is labelled as "breaking change". But note that this doesn't re-run the checks when the label is applied. To do that we would need to add these lines in checks.yml but I feared they might cause tests to restart too often:
> 
> https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b2e7bab92044031a495a53b4ccc31c84137f4319/.github/workflows/changeset.yml#L7-L11
> 
> Also, we may want to use a separate label specific to storage breakage, so we don't accidentally miss storage changes when we're doing a less serious breaking change in a minor release. @Amxx what do you think?

A tag is nice, but I think we should also consider the changeset. This could be done in two ways:
- Forget the tag, and automatically disable the storage check if there is a "major" changeset.
- Keep the tag, but add a check that if this tag is turned on, then we must have a "major" changeset.

I don't have a strong opinion for the tag name. I'm ok with "breaking change", but if you want to make it more storage specific, then "storage change" would be ok I guess.

The question is, do we want all "breaking change" tags to require a major changeset, or do we accept such tags in patch releases?

frangio:
Is your idea that the layout check would be disabled if there is _any_ major changeset in the PR branch, or if the current PR is _adding_ a new major changeset?

I went with the label because I thought you meant the second option and it seemed a lot more complicated to implement.

Amxx:
if the current PR is adding a new major changeset

frangio (approved):
I would prefer to make `_maxSupply` return `uint208` but not a blocker to merge. We should probably document the consequence of returning a number that is larger than `type(uint208).max` though.

Amxx:
I added some documentation.

frangio (approved):

contracts/governance/extensions/GovernorVotesQuorumFraction.sol
```diff
@@ -49,15 +49,15 @@ abstract contract GovernorVotesQuorumFraction is GovernorVotes {
         uint256 length = _quorumNumeratorHistory._checkpoints.length;
 
         // Optimistic search, check the latest checkpoint
-        Checkpoints.Checkpoint224 storage latest = _quorumNumeratorHistory._checkpoints[length - 1];
-        uint32 latestKey = latest._key;
-        uint224 latestValue = latest._value;
+        Checkpoints.Checkpoint208 memory latest = _quorumNumeratorHistory._checkpoints[length - 1];
```
vladyan18, line 52: Why memory?

frangio: The merge with `master` wasn't done correctly. Thanks for raising.

PR #4540, opened 2023-08-25
Optimize checkpoints

vladyan18:
Some optimizations, as i wrote in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4512#issuecomment-1666441216

- removed redundant memory usage
- used unchecked blocks for safe arithmetics

- [X] Tests
- [X] Documentation


vladyan18:
@Amxx (I apologize for the tagging, this is originally your code, judging by the Git Blame)

I'm very interested in why the library relies on **non-strictly** increasing order everywhere, like here:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b2e7bab92044031a495a53b4ccc31c84137f4319/contracts/utils/structs/Checkpoints.sol#L164

Based on the descriptions of the functions, it looks like they were created with the expectation that there may be elements with the same keys:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b2e7bab92044031a495a53b4ccc31c84137f4319/contracts/utils/structs/Checkpoints.sol#L151


At the same time, the insertion implementation does not allow adding multiple elements with the same key. This, of course, can be done manually by forcibly adding an element to the array. But you can also add an incorrect element (with a decreasing key) in such way. So it doesn't look like there should be multiple elements with the same key in an array in normal use of the library.

If the array is strictly ordered, then it is strange that the search does not handle the case when it hits the desired key (the case of equality). In the worst case, if the array is N long and the target is exactly in the middle, the current implementation will do something like `log2(n) - 1` excessive reads of cold storage slots (~2100 gas each).

So with simple example like (keys) `[2, 3, 5, 7, 11, 12, 13, 14]` i tried two implementations of lowerLookup: current and with separate handling of equalityю Gas estimations for `lowerLookup(11)`:
**current version**: 31712
**modified version**: 26762


I'm interested if there's a reason not to make such changes.

Amxx:
> I'm very interested in why the library relies on **non-strictly** increasing order everywhere

That is an interresting question. We always try to be on the safe(r) side. Given that the simplest version of the code (the one that doesn't handle equality) is the safest (it works with more structure) it felt like the better option.

Now, it may reasonable to assume that this will only be used with structures produced using `_insert`, and in that case we can indeed optimise some runtime cost, at the expense of more complexe code (and some more deployment cost)

@frangio WDYT?

vladyan18:
force-pushed: removed unchecked blocks for safe arithmetic, since this is a standalone change and would probably be better done in another pull request

vladyan18:
Created a separate pull request for [this question](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4540#issuecomment-1693954128): 
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4544

Amxx (approved):

Amxx:
@frangio Do we merge that for 5.0, or is 5.0 frozen and that goes into 5.1?

frangio:
I would put this in the bin for 5.1 with other optimizations we haven't merged yet.

.changeset/brown-cooks-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Checkpoints`: removed redundant memory usage
```
ernestognw, line 5:
```suggestion
---
'openzeppelin-solidity': patch
---

`Checkpoints`: Optimized checkpoint access by removing redundant memory usage.
```

Amxx (approved):

PR #4545, opened 2023-08-28
Optimize nested mapping access in ERC721Enumerable

0xVolosnikov:
Optimized access to nested mapping: avoiding the need to calculate the index of a slot in the storage several times.

- [X] Tests
- [X] Documentation


Amxx (approved):
I'm not a big fan of this syntax, which I find less readable ... but it saves 76gas on some run.

@ernestognw WDYT?

ernestognw:
> I'm not a big fan of this syntax, which I find less readable ... but it saves 76gas on some run.
> 
> @ernestognw WDYT?

Also not a fan of the syntax, but I think it's pretty much what we intended here and I'd rather keep the gas savings since I don't see how this might be dangerous or something.

ernestognw (approved):

PR #4546, opened 2023-08-28
Add getRoleMembers method to return all accounts that have role

vlad-khramov:
`AccessControlEnumerable` uses `AddressSet` for storing accounts that have certain role. `AddressSet` has very convenient method `values` to return all elements of the set, but it wan't used in `AccessControlEnumerable`.

Since `values` is recommened to use only in view calls, original warning was also added to documentation of `getRoleMembers` method

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx (approved):

ernestognw (approved):

ernestognw (approved):
Thanks!

Amxx (approved):

PR #4551, opened 2023-08-30
 Move security directory contents to utils 

frangio:
Since security is a project-wide focus we are removing the security directory and relocating the contents to utils. Another option that was considered was a "response" directory for incident response utilities, we may come back to this idea in the future if we grow the offering for incident response.


Amxx (approved):

Werewolves0493:
hi, how are you?
Thank you very much for your great service of OpenZeppelin

I installed
```
"@openzeppelin/contracts": "^5.0.0",
"@openzeppelin/contracts-upgradeable": "^5.0.0",
```
The problem is
```
$ npx hardhat compile
Error HH404: File @openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol, imported from contracts/AccountManager.sol, not found.
```

So I found this but where can I import `ReentrancyGuardUpgradeable `from?
Thank you

Issue #4547, opened 2023-08-29
The _update method in ERC721 contains an explicit delete of the approval


vladyan18:
I mentioned it in a comment: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#issuecomment-1692045182
Now I'm opening an issue so that the discussion is not lost

**📝 Details**

I find this line causing some inconvenience:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/token/ERC721/ERC721.sol#L254

Taking into account the fact that both the process of [approving ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/token/ERC721/ERC721.sol#L399) and [getting an approved address](_getApproved) are placed in separate virtual functions and can be overriden. In this case, if the internal mechanism is somehow replaced, the above line will remain (or you need to redefine the entire `_update `function), creating unnecessary gas waste.

At the same time, there are projects that really redefine the approval logic (and storage), for example:
https://github.com/Uniswap/v3-periphery/blob/6cce88e63e176af1ddb6cc56e029110289622317/contracts/NonfungiblePositionManager.sol#L395-L399

As mentioned in the discussion, this implementation (explicit delete in `_update`) arose because [OZ used to do _approve(address(0), tokenId);](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.7/contracts/token/ERC721/ERC721.sol#L341) but this was changed to avoid emiting the Approval event (https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#issuecomment-1692053310).

In fact, such an unnecessary storage slot touch can indeed be avoided if the `_update` function is completely overriden in its implementation. This solves the problem, albeit in a somewhat uncomfortable way in my opinion.

On the other hand, ERC20 takes a different approach in similar circumstances (https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#issuecomment-1692131099) : https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/token/ERC20/ERC20.sol#L338-L349C6

vladyan18:
UPD: it looks like it will be impossible to override `_update` in your own contract without headache - balances and other state variables are private.

[...]

PR #4552, opened 2023-08-30
Follow _approve overrides in ERC721._update

frangio:
`ERC721._update` is currently using `delete _tokenApprovals[tokenId]`, bypassing potential `_approve` overrides. Originally reported in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4377#issuecomment-1692045182, along with an [example of an `_approve` override](https://github.com/Uniswap/v3-periphery/blob/6cce88e63e176af1ddb6cc56e029110289622317/contracts/NonfungiblePositionManager.sol#L395-L399) that would have resulted in an issue.

This PR fixes this in a similar way to how we've set up the `_approve` function for `ERC20`, by adding a `bool emitEvent` argument to the function that defaults to `true`. I've made the 3-argument variants in both contracts non-virtual because they are aliases of the 4-argument `_approve`.

No changeset because this is part of the ERC721 refactor.

Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4547

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -251,7 +251,9 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
 
         // Execute the update
         if (from != address(0)) {
-            delete _tokenApprovals[tokenId];
+            // Clear approval. No need to re-authorize or emit the Approval event
+            _approve(address(0), tokenId, address(0), false);
+
             unchecked {
                 _balances[from] -= 1;
             }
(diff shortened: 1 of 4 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -395,19 +397,33 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
...
+    /**
+     * @dev Variant of `_approve` with an optional flag to enable or disable the {Approval} event. The event is not
+     * emitted in the context of transfers.
      */
-    function _approve(address to, uint256 tokenId, address auth) internal virtual returns (address) {
-        address owner = ownerOf(tokenId);
+    function _approve(address to, uint256 tokenId, address auth, bool emitEvent) internal virtual {
+        // Avoid reading the owner unless necessary
+        if (emitEvent || auth != address(0)) {
+            address owner = ownerOf(tokenId);
```
vladyan18, line 414: imo it's a bit obscure that `ownerOf` implicitly checks existance of `tokenId`

Amxx (approved):

PR #4555, opened 2023-08-31
Refactor Time library to use valueBefore/valueAfter

Amxx:
- Restructure `Time.Delay` 
- Delete `withUpdateAt`
- Add unit tests

Fixes LIB-1031
Fixes LIB-1047

#### PR Checklist

- [x] Tests
- [ ] Documentation


test/utils/types/Time.test.js
```diff
@@ -0,0 +1,150 @@
...
+
+  it('isSetAndPast', async function () {
+    for (const timepoint of [0n, 1n, 2n, 17n, 42n, MAX_UINT48])
+      for (const ref of [0n, 1n, 2n, 17n, 42n, MAX_UINT48]) {
+        expect(await this.mock.$isSetAndPast(timepoint, ref)).to.be.equal(timepoint != 0 && timepoint <= ref);
+      }
+  });
+
+  describe('Delay', function () {
+    describe('helpers', function () {
```
frangio, line 47: Is the goal of this section to test the JS helpers or the Solidity code?

```suggestion
    describe('packing and unpacking', function () {
```

test/utils/types/Time.test.js
```diff
@@ -0,0 +1,150 @@
...
+        expect(delay).to.be.deep.equal({ valueBefore: 0n, valueAfter: value, effect: 0n });
+      }
+    });
+
+    it('getAt & getFullAt', async function () {
+      const valueBefore = 24194n;
+      const valueAfter = 4214143n;
+
+      for (const effect of [0n, 1n, 15n, 16n, 17n, 42n, MAX_UINT48])
+        for (const timepoint of [0n, 1n, 15n, 16n, 17n, 42n, MAX_UINT48]) {
```
frangio, line 84: We use the same array everywhere, could we define it as a constant? The only different is the max but we could handle that in some way.

test/utils/types/Time.test.js
```diff
@@ -0,0 +1,150 @@
...
+          );
+
+          const getFullAt = await this.mock.$getFullAt(delay, timepoint);
+          expect(getFullAt[0]).to.be.bignumber.equal(String(isPast ? valueAfter : valueBefore));
+          expect(getFullAt[1]).to.be.bignumber.equal(String(isPast ? 0n : valueAfter));
+          expect(getFullAt[2]).to.be.bignumber.equal(String(isPast ? 0n : effect));
+        }
+    });
+
+    it('get & getFull', async function () {
```
frangio, line 100: This test relies on the timepoint chosen by the node and may only be testing one of the paths. We should set the chain clock to test what we're interested in, no?

test/utils/types/Time.test.js
```diff
@@ -0,0 +1,150 @@
...
+  });
+
+  describe('clocks', function () {
+    it('timestamp', async function () {
+      expect(await this.mock.$timestamp()).to.be.bignumber.equal(web3.utils.toBN(await clock.timestamp()));
+    });
+
+    it('block number', async function () {
+      expect(await this.mock.$blockNumber()).to.be.bignumber.equal(web3.utils.toBN(await clock.blocknumber()));
+    });
```
frangio, line 36: Are we using the `blockNumber` function anywhere? I can't find it. If not it might be a better idea to remove it from the Solidity code?

contracts/utils/types/Time.sol
```diff
@@ -130,16 +123,16 @@ library Time {
     function unpack(Delay self) internal pure returns (uint32, uint32, uint48) {
         uint112 raw = Delay.unwrap(self);
         return (
-            uint32(raw), // oldValue
-            uint32(raw >> 32), // newValue
+            uint32(raw >> 32), // valueBefore
+            uint32(raw), // valueAfter
             uint48(raw >> 64) // effect
```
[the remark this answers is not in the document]
Amxx: I added intermediary values that are constructed in increasing shifts. Is that good for you ? Reordering arguments created quite some confusion IMO.

test/utils/types/Time.test.js
```diff
@@ -0,0 +1,150 @@
...
+const unpackDelay = delay => ({
+  valueBefore: (BigInt(delay) >> 32n) % (1n << 32n),
+  valueAfter: (BigInt(delay) >> 0n) % (1n << 32n),
+  effect: (BigInt(delay) >> 64n) % (1n << 48n),
+});
+
+const packDelay = ({ valueBefore, valueAfter = 0n, effect = 0n }) =>
+  (BigInt(valueAfter) % (1n << 32n) << 0n) +
+  (BigInt(valueBefore) % (1n << 32n) << 32n) +
+  (BigInt(effect) % (1n << 48n) << 64n);
```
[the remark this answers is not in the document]
Amxx: I added a `asUint` function. Let me know what you think.

test/utils/types/Time.test.js
```diff
@@ -0,0 +1,150 @@
...
+        expect(await this.mock.$isSetAndPast(timepoint, ref)).to.be.equal(timepoint != 0 && timepoint <= ref);
+      }
+  });
+
+  describe('Delay', function () {
+    describe('helpers', function () {
+      const valueBefore = 17n;
+      const valueAfter = 42n;
+      const effect = 69n;
+      const delay = 1272825341158973505578n;
```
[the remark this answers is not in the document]
Amxx: One of the check we have is that `packDelay` is correct. If a regression was introduced in the helper test and the solidity at the same time, this would catch it.

test/utils/types/Time.test.js
Amxx: We are not using it. My idea was that this helper was providing a uint48 (timepoint format) version of blocknumber. We could use it in `Votes.sol`'s clock.

test/utils/types/Time.test.js
Amxx: The issue with the chain clock is we can never set it in the past, only increase it. That could makes the testing possibly unstable ?

test/utils/types/Time.test.js
frangio: Since we are not using it currently I would vote to remove it, it can be added later.

frangio: Ok I see it's being used now.

Amxx: Are you ok with how its being used ?

frangio: Yes.

My only issue with the block number in this library now is that the functions that aren't `*At` in the library automatically work with timestamps. A potential issue that I imagine is if someone uses `Delay` with block number based times, but uses `get` instead of `getAt` somewhere.

Amxx: Should we add a additional argument to the `get` that specify which mode they are operating on.

Maybe we could have a `Time.clock(<this enum>) returns (uint48)` function.

test/utils/types/Time.test.js
Amxx: I implemented that using a `effectForTimepoint` helper. WDYT ?

test/utils/types/Time.test.js
frangio: At that point it feels overengineered really.

Amxx: so do we keep it like this ?

frangio (approved):

PR #4542, opened 2023-08-26
Use leading underscore solhint rule for private constants

vladyan18:
Partially closes the comment https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4132#discussion_r1254828352

I limited this rule to only **private** constants so as not to create a breaking change.

#### PR Checklist

- [X] Tests
- [X] Documentation
- [X] Changeset entry (run `npx changeset add`)

PR #4560, opened 2023-09-01
Define ERC-4906 interfaceId in a private constant

Amxx:
Fixes LIB-1036


frangio:
We don't need underscores for constants anymore. This is now enforced in `master` after https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4542.

PR #4566, opened 2023-09-02
Refactor ERC721 `_requireMinted` and `ownerOf`

frangio:
There was duplication between `ownerOf` and `_requireMinted`. There was also a source of confusion in functions that used `ownerOf` to get the owner and revert if there wasn't an owner. Fixed this by changing `_requireMinted` into `_requireOwned` with the owner as a return value.

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -65,11 +65,7 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
      * @dev See {IERC721-ownerOf}.
      */
     function ownerOf(uint256 tokenId) public view virtual returns (address) {
-        address owner = _ownerOf(tokenId);
-        if (owner == address(0)) {
-            revert ERC721NonexistentToken(tokenId);
-        }
-        return owner;
+        return _requireOwned(tokenId);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/token/ERC721/ERC721.sol
```diff
@@ -427,12 +423,15 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
     }
 
     /**
-     * @dev Reverts if the `tokenId` has not been minted yet.
+     * @dev Reverts if the `tokenId` doesn't have a current owner (it hasn't been minted, or it has been burned).
+     * Returns the owner.
      */
-    function _requireMinted(uint256 tokenId) internal view virtual {
-        if (_ownerOf(tokenId) == address(0)) {
+    function _requireOwned(uint256 tokenId) internal view virtual returns (address) {
```
Amxx, line 429: Should this be virtual ?

frangio: I don't see why not.

Amxx: I'm wondering what would happen if someone overrides this to change the returned value instead of overriding `_ownerOf`.

frangio: Oh I see what you mean. Alright, I'm removing virtual.

Amxx (approved):

PR #4568, opened 2023-09-04
Better context on _spendAllowance NatSpec

0xneves:
Changed "Might emit" to "Won't emit" in `_spendAllowance` function.

`_approve` will only handle events when the fourth argument (boolean) is `true`. 

Such a condition will never trigger as `_spendAllowance` will always call `_approve` passing a hardcoded `false` input.

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -356,7 +356,7 @@ abstract contract ERC20 is Context, IERC20, IERC20Metadata, IERC20Errors {
      * Does not update the allowance value in case of infinite allowance.
      * Revert if not enough allowance is available.
      *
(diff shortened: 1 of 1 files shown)


contracts/token/ERC20/ERC20.sol
```diff
@@ -356,7 +356,7 @@ abstract contract ERC20 is Context, IERC20, IERC20Metadata, IERC20Errors {
      * Does not update the allowance value in case of infinite allowance.
      * Revert if not enough allowance is available.
      *
-     * Might emit an {Approval} event.
+     * Won't emit an {Approval} event.
```
Amxx, line 359:
```suggestion
     * Does not emit an {Approval} event.
```

frangio (approved):
Thanks

PR #4569, opened 2023-09-04
Add a boolean to AccessManager.GrantGroup

Amxx:
This will help distinguish granting membership (where this since is the effect at which the membership and execution delay both takes effect) from updates to the execution delay.

Fixes LIB-1051

#### PR Checklist

- [x] Tests

### DIFF
--- a/contracts/access/manager/IAccessManager.sol
+++ b/contracts/access/manager/IAccessManager.sol
@@ -29,7 +29,7 @@ interface IAccessManager {
     event OperationCanceled(bytes32 indexed operationId, uint32 indexed nonce);
 
     event GroupLabel(uint64 indexed groupId, string label);
-    event GroupGranted(uint64 indexed groupId, address indexed account, uint32 delay, uint48 since, bool isUpdate);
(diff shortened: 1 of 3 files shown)


contracts/access/manager/IAccessManager.sol
```diff
@@ -29,7 +29,7 @@ interface IAccessManager {
     event OperationCanceled(bytes32 indexed operationId, uint32 indexed nonce);
 
     event GroupLabel(uint64 indexed groupId, string label);
-    event GroupGranted(uint64 indexed groupId, address indexed account, uint32 delay, uint48 since);
+    event GroupGranted(uint64 indexed groupId, address indexed account, uint32 delay, uint48 since, bool isUpdate);
```
frangio, line 32: I had originally thought we would emit the opposite of this. Something like `bool newMember` that is true when the grant is an actual grant. What do you think?

Amxx: Its symetric, so either is good to me.

Amxx: changed

frangio (approved):

PR #4570, opened 2023-09-04
Prevent setting address(0) as the initialAdmin in AccessManager

Amxx:
Similarly to the changes proposed in #4531, we want to prevent construction with address(0) as the admin.

Fixes LIB-1057

#### PR Checklist

- [x] Tests


contracts/access/manager/IAccessManager.sol
```diff
@@ -52,6 +52,7 @@ interface IAccessManager {
     error AccessManagerUnauthorizedAccount(address msgsender, uint64 groupId);
     error AccessManagerUnauthorizedCall(address caller, address target, bytes4 selector);
     error AccessManagerCannotCancel(address msgsender, address caller, address target, bytes4 selector);
+    error AccessManagerInvalidInitialAdmin(address);
```
frangio, line 55:
```suggestion
    error AccessManagerInvalidInitialAdmin(address initialAdmin);
```

frangio (approved):

PR #4573, opened 2023-09-04
Remove unused named return variables

Amxx:
Fixes LIB-1033

For consistency with the rest of the codebase, removing unused named variables from the methods' signatures

### DIFF
--- a/contracts/access/extensions/AccessControlDefaultAdminRules.sol
+++ b/contracts/access/extensions/AccessControlDefaultAdminRules.sol
@@ -190,7 +190,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      * @inheritdoc IAccessControlDefaultAdminRules
      */
     function pendingDefaultAdminDelay() public view virtual returns (uint48, uint48) {
-        schedule = _pendingDelaySchedule;
+        uint48 schedule = _pendingDelaySchedule;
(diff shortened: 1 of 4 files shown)


frangio (approved):

contracts/access/extensions/AccessControlDefaultAdminRules.sol
```diff
@@ -189,7 +189,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
     /**
      * @inheritdoc IAccessControlDefaultAdminRules
      */
-    function pendingDefaultAdminDelay() public view virtual returns (uint48 newDelay, uint48 schedule) {
+    function pendingDefaultAdminDelay() public view virtual returns (uint48, uint48) {
         schedule = _pendingDelaySchedule;
```
frangio, line 193:
```suggestion
        uint48 schedule = _pendingDelaySchedule;
```

frangio (changes requested):

frangio:
We are probably using the named return values in tests.

Amxx:
> We are probably using the named return values in tests.

Yes, we do ```const { newDelay, schedule } = await this.accessControl.pendingDefaultAdminDelay();```

frangio (approved):
I had added a new guideline:
> Functions with multiple return values should use named return values for clarity, even if the function uses an explicit `return` statement.

But reverted it because I don't think we should add the guideline without reviewing the rest of the codebase.

This is our intention though.

PR #4574, opened 2023-09-04
Fix accuracy of docs for ERC20._burn

0xneves:
Current descriptions:
The description mentions the destination of the tokens to the 0 address, but this is the behavior of an old version of OpenZeppelin ERC20.

Expected description: 
Tokens aren't sent anymore to the `address(0)` when using `_burn`. Instead, the total supply is lowered and the burnt amount is removed from the account balance.


frangio (approved):
Yeah I don't think the previous text was accurate at any point.

Thanks!

PR #4575, opened 2023-09-05
Make isConsumingScheduleOp return bytes4 to mitigate clashes

Amxx:
1. Return bytes4 out of caution in case the selector for `isConsumingScheduleOp` clashes with another function in the contract (which may be `Ownable` and not have that getter from `AccessManaged`).
2. Use a custom error.

Fixes LIB-1056

#### PR Checklist

- [ ] Tests
- [ ] Documentation


frangio (approved):
@Amxx I renamed the new custom error. I was thinking we could also rename `CannotCancel` to `UnauthorizedCancel` for consistency. What do you think?

PR #4576, opened 2023-09-05
Improve Initializable readability using intermediate variables

Amxx:
Fixes LIB-1041

#### PR Checklist

### DIFF
--- a/contracts/proxy/utils/Initializable.sol
+++ b/contracts/proxy/utils/Initializable.sol
@@ -104,9 +104,19 @@ abstract contract Initializable {
         // solhint-disable-next-line var-name-mixedcase
         InitializableStorage storage $ = _getInitializableStorage();
 
+        // Cache values to avoid duplicated sloads
         bool isTopLevelCall = !$._initializing;
         uint64 initialized = $._initialized;
-        if (!(isTopLevelCall && initialized < 1) && !(address(this).code.length == 0 && initialized == 1)) {
+
+        // Are allowed:
+        // - "initialSetup" calls: the contract is not in the initializing state and no previous version was
+        //                         initialized
+        // - "construction" calls: the contract is already initialized at step 1 (no reininitialization) and the
+        //                         current contract is not yet constructed.
+        bool initialSetup = initialized == 0 && isTopLevelCall;
+        bool construction = initialized == 1 && address(this).code.length == 0;
+
+        if (!firstCall && !construction) {
             revert AlreadyInitialized();
         }
         $._initialized = 1;
--- a/contracts/proxy/utils/Initializable.sol
+++ b/contracts/proxy/utils/Initializable.sol
@@ -116,7 +116,7 @@ abstract contract Initializable {
         bool initialSetup = initialized == 0 && isTopLevelCall;
         bool construction = initialized == 1 && address(this).code.length == 0;
 
-        if (!firstCall && !construction) {
+        if (!initialSetup && !construction) {
             revert AlreadyInitialized();
         }
         $._initialized = 1;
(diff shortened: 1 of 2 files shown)


contracts/proxy/utils/Initializable.sol
```diff
@@ -104,9 +104,19 @@ abstract contract Initializable {
...
+        // Cache values to avoid duplicated sloads
         bool isTopLevelCall = !$._initializing;
         uint64 initialized = $._initialized;
-        if (!(isTopLevelCall && initialized < 1) && !(address(this).code.length == 0 && initialized == 1)) {
+
+        // Are allowed:
+        // - "initialSetup" calls: the contract is not in the initializing state and no previous version was
+        //                         initialized
+        // - "construction" calls: the contract is already initialized at step 1 (no reininitialization) and the
+        //                         current contract is not yet constructed.
```
frangio, line 115:
```suggestion
        // Allowed calls:
        // - initialSetup: the contract is not in the initializing state and no previous version was
        //                 initialized
        // - construction: the contract is initialized at version 1 (no reininitialization) and the
        //                 current contract is just being deployed
```

?

contracts/proxy/utils/Initializable.sol
```diff
@@ -104,9 +104,19 @@ abstract contract Initializable {
...
+        //                 initialized
+        // - construction: the contract is initialized at version 1 (no reininitialization) and the
+        //                 current contract is just being deployed
+        bool initialSetup = initialized == 0 && isTopLevelCall;
+        bool construction = initialized == 1 && address(this).code.length == 0;
+
+        if (!initialSetup && !construction) {
             revert AlreadyInitialized();
         }
         $._initialized = 1;
```
vladyan18, line 122:
```solhint
$._initialized = 1;
$._initializing = isTopLevelCall;
```

May be cheaper here, removing redundant SLOAD operation in case of `!isTopLevelCall`

Amxx: I'm not sure I understand your proposal. Before doing `_;`, we want to make sure the initialising is true, this would possibly invert it to false.

Also I'm not sure where the duplicated sload is.

vladyan18: > Also I'm not sure where the duplicated sload is.

Since `$` is a [struct with two fields](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/175b1e9e09b19f90d96b472a0e0b4ebea9a47762/contracts/proxy/utils/Initializable.sol#L65-L74) packed in one storage slot, compiler need to load this slot if you change only one field. If you write all the fields of the structure at once, the compiler does not need to read this slot before SSTORE

vladyan18: So it could (in theory) save around 100 gas here

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/175b1e9e09b19f90d96b472a0e0b4ebea9a47762/contracts/proxy/utils/Initializable.sol#L122-L125

If do it like

```solidity
```

vladyan18: > this would possibly invert it to false

Upd: yes, in this case my proposal should be rather:

```solidity
  $._initialized = 1;
  $._initializing = true;
```

frangio: Writing it that way requires proving that the `_;` part doesn't change the value of `$._initialized`.

I think it is true, because `$._initializing = true` prevents that value from changing in the rest of the contract, but it's more difficult to see.

I think I prefer to keep the current code at this stage (audit fixes) and come back to this for a future release. @vladyan18 can you open an issue?

frangio (approved):

frangio (approved):

PR #4581, opened 2023-09-06
Improve natspec documentation and comments

Amxx:
Fixes LIB-1060
Fixes LIB-1038

TODO: documentation of interfaces `IAccessManaged` and `IAccessManager`

#### PR Checklist

- [x] Documentation

### DIFF
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -480,7 +480,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
     }
 
     /**
-     * @dev Set the delay for management operations on a given class of contract.
+     * @dev Set the delay for management operations on a given target contract.
      *
      * Requirements:
      *
@@ -493,7 +493,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
     }
 
     /**
-     * @dev Internal version of {setClassAdminDelay} without access control.
+     * @dev Internal version of {setTargetAdminDelay} without access control.
      *
      * Emits a {TargetAdminDelayUpdated} event
      */
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -813,7 +813,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * @dev An extended version of {canCall} for internal use that considers restrictions for admin functions.
      *
      * Returns:
-     * - bool immediate: weither the operation can be executed immediatly (with no delay)
+     * - bool immediate: whether the operation can be executed immediately (with no delay)
      * - uint32 delay: the execution delay
      *
      * If immediate is true, the delay can be disregarded and the operation can be immediatly executed.
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -816,7 +816,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * - bool immediate: whether the operation can be executed immediately (with no delay)
      * - uint32 delay: the execution delay
      *
-     * If immediate is true, the delay can be disregarded and the operation can be immediatly executed.
+     * If immediate is true, the delay can be disregarded and the operation can be immediately executed.
      * If immediate is false, the operation can be executed if and only if delay is greater than 0.
      */
     function _canCallExtended(address caller, address target, bytes calldata data) private view returns (bool, uint32) {
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -51,6 +51,7 @@ import {Time} from "../../utils/types/Time.sol";
 contract AccessManager is Context, Multicall, IAccessManager {
     using Time for *;
 
+    // Structure that stores the details for a target contract.
     struct TargetConfig {
         mapping(bytes4 selector => uint64 groupId) allowedGroups;
         Time.Delay adminDelay;
@@ -78,6 +79,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
         Time.Delay grantDelay;
     }
 
+    // Structure that stores the details for a scheduled operation. This structure fits into a single slot.
     struct Schedule {
         uint48 timepoint;
         uint32 nonce;
@@ -455,7 +457,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      *
      * - the caller must be a global admin
      *
-     * Emits a {FunctionAllowedGroupUpdated} event per selector
+     * Emits a {TargetFunctionGroupUpdated} event per selector
      */
     function setTargetFunctionGroup(
         address target,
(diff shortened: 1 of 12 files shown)


contracts/access/manager/AccessManager.sol
```diff
@@ -810,6 +811,13 @@ contract AccessManager is Context, Multicall, IAccessManager {
     // =================================================== HELPERS ====================================================
     /**
      * @dev An extended version of {canCall} for internal use that considers restrictions for admin functions.
+     *
+     * Returns:
+     * - bool immediate: weither the operation can be executed immediatly (with no delay)
```
vladyan18, line 816: “weither” -> “whether”

Amxx, line 816:
```suggestion
     * - bool immediate: whether the operation can be executed immediately (with no delay)
```

contracts/token/ERC721/ERC721.sol
```diff
@@ -195,10 +195,11 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
 
     /**
      * @dev Checks if `spender` can operate on `tokenId`, assuming the provided `owner` is the actual owner.
-     * Reverts if `spender` has not approval for all assets of the provided `owner` nor the actual owner approved the `spender` for the specific `tokenId`.
+     * Reverts if `spender` has not approval for all assets of the provided `owner` nor the actual owner approved
```
vladyan18, line 198: “has not approval” -> “does not have approval”?

contracts/access/manager/AccessManager.sol
```diff
@@ -810,6 +811,13 @@ contract AccessManager is Context, Multicall, IAccessManager {
     // =================================================== HELPERS ====================================================
     /**
      * @dev An extended version of {canCall} for internal use that considers restrictions for admin functions.
+     *
+     * Returns:
+     * - bool immediate: whether the operation can be executed immediately (with no delay)
+     * - uint32 delay: the execution delay
+     *
+     * If immediate is true, the delay can be disregarded and the operation can be immediatly executed.
```
Amxx, line 819:
```suggestion
     * If immediate is true, the delay can be disregarded and the operation can be immediately executed.
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -33,8 +33,11 @@ library ECDSA {
     error ECDSAInvalidSignatureS(bytes32 s);
 
     /**
-     * @dev Returns the address that signed a hashed message (`hash`) with
-     * `signature` or error string. This address can then be used for verification purposes.
+     * @dev Returns the address that signed a hashed message (`hash`) with `signature` or an error. This will not
+     * return address(0) without also returning an error description. Errors are documented using an enum (error type)
+     * and a bytes32 providing additional information about the error.
+     *
+     * If no error is returned, then the address can then be used for verification purposes.
```
vladyan18, line 40: “then the address can then be” -> “then the address can be”?

contracts/token/ERC721/ERC721.sol
Amxx:
```suggestion
     * Reverts if `spender` does not have approval from the provided `owner` for the given token or for all its assets
```
?

contracts/utils/cryptography/ECDSA.sol
Amxx:
```suggestion
     * If no error is returned, then the address can be used for verification purposes.
```

scripts/generate/templates/EnumerableMap.js
```diff
@@ -57,14 +57,10 @@ import {EnumerableSet} from "./EnumerableSet.sol";
...
-// repetition as possible, we write it in terms of a generic Map type with
-// bytes32 keys and values.
-// The Map implementation uses private functions, and user-facing
-// implementations (such as Uint256ToAddressMap) are just wrappers around
-// the underlying Map.
-// This means that we can only create new EnumerableMaps for types that fit
-// in bytes32.
+// To implement this library for multiple types with as little code repetition as possible, we write it in
+// terms of a generic Map type with bytes32 keys and values. The Map implementation uses private functions,
+// and user-facing implementations (such as UintToAddressMap) are just wrappers around the underlying Map.
```
vladyan18, line 62: “are just wrappers” -> “that are just wrappers”?

Amxx: > [...], and user-facing implementations (...) are just wrappers around the underlying Map.

This (current version) feels ok to me. @frangio ? Maybe we can remove the "(such as UintToAddressMap)" part.

vladyan18:
"if it as already been queued" doesn't sound right

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9ef69c03d13230aeff24d91cb54c9d24c4de7c8b/contracts/governance/extensions/GovernorTimelockControl.sol#L107C104-L107C117

contracts/token/ERC721/extensions/ERC721Consecutive.sol
```diff
@@ -19,12 +19,12 @@ import {Checkpoints} from "../../../utils/structs/Checkpoints.sol";
...
- * when relying on hooks.
+ * IMPORTANT: This extension does not call the {_update} function for tokens minted in batch. Any logic added to this
+ * function through overrides will not be triggered when token are minted in batch. You may want to also override
+ * {_increaseBalance} or {_mintConsecutive} to account for these mints.
  *
- * IMPORTANT: When overriding {_afterTokenTransfer}, be careful about call ordering. {ownerOf} may return invalid
- * values during the {_afterTokenTransfer} execution if the super call is not called first. To be safe, execute the
+ * IMPORTANT: When overriding {_mintConsecutive}, be careful about call ordering. {ownerOf} may return invalid
+ * values during the {_mintConsecutive} execution if the super call is not called first. To be safe, execute the
  * super call before your custom logic.
```
frangio, line 28: This comment referred to `_afterTokenTransfer` because of how burning was implemented. I think it should now refer to `_update`.

```suggestion
 * IMPORTANT: When overriding {_update}, be mindful about call ordering. {ownerOf} may return invalid
 * values during {_update} execution if `ERC721Consecutive`'s code is not executed first. To be safe, make sure
 * `ERC721Consecutive._update` fully executes before your custom logic, and test for the expected behavior.
```

scripts/generate/templates/EnumerableMap.js
frangio: Yeah the current version is okay but I removed parentheses to make it less confusing.

frangio:
>"if it as already been queued" doesn't sound right

@vladyan18 Fixed.

contracts/token/ERC721/extensions/ERC721Consecutive.sol
Amxx: I was really meaning `_mintConsecutive`. I guess this is true of both functions, buth since we this comment is in `ERC721Consecutive`, I think we should focus on the consecutive minting part.

Maybe we want another similar warning in `ERC721`.

In both cases I think the reasonning is the same: if you put logic before calling super, you'll be executing that logic in a context where the change have not been applied. Tokens are still owned by the previous owner, and balances have not been updated. It feels obvious, maybe we don't need a warning at all.

frangio:
> TODO: documentation of interfaces IAccessManaged and IAccessManager

Let's tackle this in another PR.

frangio (approved):

PR #4582, opened 2023-09-07
Remove SafeERC20.safePermit

Amxx:
+ Added documentation about the danger of ERC20Permit as an authentication method.

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/token/ERC20/extensions/IERC20Permit.sol
frangio, line 1: It looks like `IERC20Permit` is not listed in the docs site. (https://docs.openzeppelin.com/contracts/4.x/api/token/erc20)

I added it in this PR, let's see if it looks nice.

contracts/token/ERC20/extensions/IERC20Permit.sol
```diff
@@ -10,6 +10,26 @@ pragma solidity ^0.8.20;
  * Adds the {permit} method, which can be used to change an account's ERC20 allowance (see {IERC20-allowance}) by
  * presenting a message signed by the account. By not relying on {IERC20-approve}, the token holder account doesn't
  * need to send a transaction, and thus is not required to hold Ether at all.
+ *
+ * ==== Security Considerations
```
frangio, line 14: @Amxx I don't think the utilities page was the right location for these docs. I moved it here and made some changes to it.

contracts/token/ERC20/extensions/IERC20Permit.sol
```diff
@@ -10,6 +10,26 @@ pragma solidity ^0.8.20;
...
+ * There are two important considerations concerning the use of `permit`. The first is that a valid permit signature
+ * expresses an allowance, and it should not be assumed to convey additional meaning. In particular, it should not be
+ * considered as an intention to spend the approval in any specific way. The second is that because permits have
+ * built-in replay protection and can be submitted by anyone, they can be frontrun. A protocol that uses permits should
+ * take this into consideration and allow a `permit` call to fail. Combining these two aspects, a good pattern that can
+ * be generally recommended is:
+ *
+ * ```
+ * try token.permit(msg.sender, spender, value, deadline, v, r, s) {} catch {}
+ * ```
```
frangio, line 25: We could expand this into a more complete example like this too:

```solidity
function doThingWithPermit(..., uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s) {
      try token.permit(msg.sender, address(this), value, deadline, v, r, s) {} catch {}
      doThing(...); // will use token.transferFrom
}
```

contracts/token/ERC20/extensions/IERC20Permit.sol
```diff
@@ -10,6 +10,26 @@ pragma solidity ^0.8.20;
...
  * need to send a transaction, and thus is not required to hold Ether at all.
+ *
+ * ==== Security Considerations
+ *
+ * There are two important considerations concerning the use of `permit`. The first is that a valid permit signature
+ * expresses an allowance, and it should not be assumed to convey additional meaning. In particular, it should not be
+ * considered as an intention to spend the approval in any specific way. The second is that because permits have
+ * built-in replay protection and can be submitted by anyone, they can be frontrun. A protocol that uses permits should
+ * take this into consideration and allow a `permit` call to fail. Combining these two aspects, a good pattern that can
+ * be generally recommended is:
```
Amxx, line 21: We should not recommend the pattern without explciitelly saying what we recommend it for !!!

People see its recommended, use it, and then don't understand that it doesn't do what they expect it to do. IMO we should say: "if your usecase is X, and you understand that Y, then we recommend using the following pattern"

contracts/token/ERC20/extensions/IERC20Permit.sol
```diff
@@ -32,6 +52,9 @@ interface IERC20Permit {
      * For more information on the signature format, see the
      * https://eips.ethereum.org/EIPS/eip-2612#specification[relevant EIP
      * section].
+     *
+     * WARNING: relying on {permit} for user authentication is a bad practice.
+     * See xref:utilities.adoc#authentication-using-permit[this].
```
Amxx, line 57: link is no longer good.

contracts/token/ERC20/extensions/IERC20Permit.sol
frangio: Do you say that because it says "_generally_ recommended"? Can you think of any scenarios where this would not be recommended?

I think the recommendation should apply always, "unless you know what you're doing".

An example of something different might be a protocol like CowSwap, where a solver may submits your permit along with the trade, so it will not be `msg.sender`. But this is exactly the kind of thing where you need to be very careful about how you design it so I put it in the "you better know what you're doing" bin.

Amxx: > Can you think of any scenarios where this would not be recommended?

It really demands what you are trying to achieve as a dev. For example if you try to do a "one step deposit" that set the approval through permit, then consume it ... you may be tempted to do a try catch on the permit, than consume all the approval you have (if any). I think that is a bad design, that your recommandation does address. People might think its a good idea because you recommand something that allows that.

frangio: >that your recommandation does address

Do you mean "does not address"?

Does this example shown in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4582/files#r1319293034 helpe?

Amxx: I did mean "does address".

Amxx: I think that the current wording is ok.

PR #4584, opened 2023-09-07
Add guidelines about naming returns in function that return multiple values

Amxx:
Do we want a CI check for that ?

Fixes LIB-1058

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -147,4 +147,4 @@ In addition to the official Solidity Style Guide we have a number of other conve
 
   * Custom error names should not be declared twice along the library to avoid duplicated identifier declarations when inheriting from multiple contracts.
 
-* Functions with multiple return values should use named return values for clarity, even if the function uses an explicit `return` statement.
+* Functions with multiple return values should use named return values for clarity, even if the function uses an explicit `return` statement. Although named variables can be used, the `return` statement is preferred for clarity.
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -146,3 +146,5 @@ In addition to the official Solidity Style Guide we have a number of other conve
     4. Declare the error in an extension if the error only happens in such extension or child contracts.
 
   * Custom error names should not be declared twice along the library to avoid duplicated identifier declarations when inheriting from multiple contracts.
+
+* Functions with multiple return values should use named return values for clarity, even if the function uses an explicit `return` statement.
--- a/contracts/governance/extensions/GovernorStorage.sol
+++ b/contracts/governance/extensions/GovernorStorage.sol
@@ -86,7 +86,12 @@ abstract contract GovernorStorage is Governor {
      */
     function proposalDetails(
         uint256 proposalId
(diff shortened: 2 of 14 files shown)


contracts/governance/extensions/GovernorStorage.sol
```diff
@@ -100,14 +100,19 @@ abstract contract GovernorStorage is Governor {
...
             address[] memory targets,
             uint256[] memory values,
             bytes[] memory calldatas,
             bytes32 descriptionHash
-        ) = proposalDetails(proposalId);
-        return (proposalId, targets, values, calldatas, descriptionHash);
+        )
+    {
+        proposalId = _proposalIds[index];
+        (targets, values, calldatas, descriptionHash) = proposalDetails(proposalId);
```
frangio, line 116: The guidelines are about putting names on the return values, but not necessarily about using those variables. What are our guidelines for actually using those variables instead of explicit return statements?

Amxx: Naming the return variable implies declaring them. This means the name is used and cannot be redeclared. The example here is a good one: do we want to redeclare everything with a different name, and then do an explicit return ? Sometimes maybe, but here it felt too much.

contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -217,8 +217,8 @@ contract ERC2771Forwarder is EIP712, Nonces {
      */
     function _recoverForwardRequestSigner(
         ForwardRequestData calldata request
-    ) internal view virtual returns (bool, address) {
-        (address recovered, ECDSA.RecoverError err, ) = _hashTypedDataV4(
+    ) internal view virtual returns (bool isValid, address recovered) {
+        (address signer, ECDSA.RecoverError err, ) = _hashTypedDataV4(
```
ernestognw, line 221: I hate that we can't do:

```solidity
(recovered, ECDSA.RecoverError err, ) = _hashTypedDataV4(
```

GUIDELINES.md
```diff
@@ -146,3 +146,5 @@ In addition to the official Solidity Style Guide we have a number of other conve
     4. Declare the error in an extension if the error only happens in such extension or child contracts.
 
   * Custom error names should not be declared twice along the library to avoid duplicated identifier declarations when inheriting from multiple contracts.
+
+* Functions with multiple return values should use named return values for clarity, even if the function uses an explicit `return` statement.
```
ernestognw, line 150: I'd mention the return statement is generally preferred for clarity. I don't want the repo to abuse of named return values.

```suggestion
* Functions with multiple return values should use named return values for clarity, even if the function uses an explicit `return` statement. Although named variables can be used, the `return` statement is preferred for clarity.
```

contracts/metatx/ERC2771Forwarder.sol
Amxx: We can do
```
ECDSA.RecoverError err;
(recovered, err, ) = _hashTypedDataV4(
```

PR #4585, opened 2023-09-07
Remove non-standard increaseAllowance and decreaseAllowance from ERC20

### LINKED ISSUE #4583: Discussion to remove `increaseAllowance` and `decreaseAllowance` from `ERC20`
> If there is another issue that discusses the same topic, feel free to close this one.

I wanted to quickly get your opinion on whether it would make sense to remove the functions `increaseAllowance` and `decreaseAllowance` from the [`ERC20`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol) contract and move it to an extension contract instead. My arguments are the following:

- These functions are _not_ part of the [EIP-20](https://eips.ethereum.org/EIPS/eip-20) specs.
- These functions may allow for further phishing possibilities (instead of the common `approve` or `permit` ones; see e.g. just 12 hours ago someone lost $24m since he got tricked into signing a malicious `increaseAllowance` payload  https://etherscan.io/tx/0xcbe7b32e62c7d931a28f747bba3a0afa7da95169fcf380ac2f7d54f3a2f77913).
- The security concerns that fix `increaseAllowance` and `decreaseAllowance` are not critical nor high in the wild (and `decreaseAllowance` can be frontrunned also) and thus I think the responsibility can be delegated to the devs to decide whether to include it or not.
- If such a change is implemented, the upcoming breaking version `5.0.0` would be suitable.

Amxx:
(See discussion in issue)

Fixes #4583

#### PR Checklist

- [x] Tests
- [x] Documentation


frangio:
Should we add any documentation about the existence of `increase/decreaseAllowance` in previous versions?

frangio:
Regarding documentation, my thinking was that we'd want to document for wallets that there is an alternative way to change allowance and any protections to `approve` should also be used with `increaseAllowance`. We don't think our documentation is the best place for this warning, and we're considering publishing a blog post about it.

ernestognw:
> > Should we add any documentation about the existence of `increase/decreaseAllowance` in previous versions?
> 
> We definitely shoud. I came here because the official open zeppelin docs say that the functions exist: image
> 
> Yet, I couldn't find them in the source code, hence posting here!

You're looking at v4 sir. Please change the documentation version to 5.x 😄

PR #4586, opened 2023-09-08
Improve AccessManager docs

frangio:
In `revokeRole`, `renounceRole`, and `_revokeRole` the event is not always emitted.

Fixes LIB-1091

### DIFF
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -168,13 +168,15 @@ contract AccessManager is Context, Multicall, IAccessManager {
     }
 
     /**
-     * @dev Get the permission level (role) required to call a function. This only applies for contract that are
-     * operating under the `Custom` mode.
+     * @dev Get the role required to call a function.
      */
     function getTargetFunctionRole(address target, bytes4 selector) public view virtual returns (uint64) {
         return _targets[target].allowedRoles[selector];
     }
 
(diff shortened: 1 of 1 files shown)


contracts/access/manager/AccessManager.sol
```diff
@@ -277,14 +277,14 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
-     * effect.
+     * @dev Remove an account from a role, with immediate effect. If the account is not in the role, this call
+     * has no effect.
      *
      * Requirements:
      *
      * - the caller must be in the role's admins
      *
-     * Emits a {RoleRevoked} event
+     * Emits a {RoleRevoked} event if the account was in the role.
```
Amxx, line 287: `was in the role` or `had the role`.

To me an element is in a group, but a member has a role. Somehow a group is something you are in while a role is something attached to you. Does that make sens ?

frangio: I see. That makes sense.

frangio: I followed the wording that was in the rest of the docs but that makes sense.

frangio: I fixed this, and changed a few other things I noticed missing in the docs.

Amxx (approved):

PR #4591, opened 2023-09-08
Improve GovernorTimelockAccess

Amxx:
1. Ensure that `_cancel` works correctly even when an operation has already been cancelled at the manager level, and even when it has been rescheduled after that.
2. Ensure that calls are executed through the AccessManager even if it authorizes the call without a delay. This is to make sure that we support operations on legacy contracts (e.g. Ownable) that are under control of (e.g. owned by) the `AccessManager`.
3. Add a mechanism to ignore restrictions claimed  by the AccessManager to mitigate a denial of service on functions that should be called directly.

Fixes LIB-1066
Fixes LIB-1065 (part of)

### DIFF
--- a/contracts/governance/extensions/GovernorTimelockAccess.sol
+++ b/contracts/governance/extensions/GovernorTimelockAccess.sol
@@ -234,7 +234,7 @@ abstract contract GovernorTimelockAccess is Governor {
             for (uint256 i = 0; i < targets.length; ++i) {
                 (, bool withDelay, uint32 nonce) = _getManagerData(plan, i);
                 if (withDelay) {
-                    bytes32 operationId = keccak256(abi.encode(address(this), targets[i], calldatas[i]));
+                    bytes32 operationId = _manager.hashOperation(address(this), targets[i], calldatas[i]);
                     if (nonce == _manager.getNonce(operationId)) {
                         // Attempt to cancel considering the operation could have been cancelled and rescheduled already
                         try _manager.cancel(address(this), targets[i], calldatas[i]) returns (uint32) {} catch {}
--- a/contracts/governance/extensions/GovernorTimelockAccess.sol
+++ b/contracts/governance/extensions/GovernorTimelockAccess.sol
@@ -129,8 +129,7 @@ abstract contract GovernorTimelockAccess is Governor {
         bool ignored
     ) public virtual onlyGovernance {
         for (uint256 i = 0; i < selectors.length; ++i) {
-            bytes4 selector = selectors[i];
-            _setAccessManagerIgnored(target, selector, ignored);
+            _setAccessManagerIgnored(target, selectors[i], ignored);
         }
     }
 
--- a/contracts/governance/extensions/GovernorTimelockAccess.sol
+++ b/contracts/governance/extensions/GovernorTimelockAccess.sol
@@ -97,7 +97,9 @@ abstract contract GovernorTimelockAccess is Governor {
      * delayed since queuing, and an array indicating which of the proposal actions will be executed indirectly through
      * the associated {AccessManager}.
      */
-    function proposalExecutionPlan(uint256 proposalId) public view returns (uint32 delay, bool[] memory indirect, bool[] memory withDelay) {
+    function proposalExecutionPlan(
+        uint256 proposalId
+    ) public view returns (uint32 delay, bool[] memory indirect, bool[] memory withDelay) {
         ExecutionPlan storage plan = _executionPlan[proposalId];
 
         uint32 length = plan.length;
@@ -231,12 +233,11 @@ abstract contract GovernorTimelockAccess is Governor {
             for (uint256 i = 0; i < targets.length; ++i) {
                 (, bool withDelay, uint32 nonce) = _getManagerData(plan, i);
                 if (withDelay) {
-                    // Attempt to cancel considering the operation could have been cancelled and rescheduled already
-                    try _manager.cancel(address(this), targets[i], calldatas[i]) returns (uint32 canceledNonce) {
-                        if (canceledNonce != nonce) {
-                            revert GovernorMismatchedNonce(proposalId, nonce, canceledNonce);
-                        }
-                    } catch {}
+                    bytes32 operationId = keccak256(abi.encode(address(this), targets[i], calldatas[i]));
+                    if (nonce == _manager.getNonce(operationId)) {
+                        // Attempt to cancel considering the operation could have been cancelled and rescheduled already
+                        try _manager.cancel(address(this), targets[i], calldatas[i]) returns (uint32) {} catch {}
+                    }
                 }
             }
         }
--- a/contracts/governance/extensions/GovernorTimelockAccess.sol
+++ b/contracts/governance/extensions/GovernorTimelockAccess.sol
@@ -197,7 +197,7 @@ abstract contract GovernorTimelockAccess is Governor {
             (bool immediate, uint32 delay) = AuthorityUtils.canCallWithDelay(
                 address(_manager),
                 address(this),
-                targets[i],
+                target,
                 selector
             );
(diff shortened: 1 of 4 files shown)


contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -97,17 +97,20 @@ abstract contract GovernorTimelockAccess is Governor {
      * delayed since queuing, and an array indicating which of the proposal actions will be executed indirectly through
      * the associated {AccessManager}.
```
frangio, line 98:
```suggestion
     * delayed since queuing, an array indicating which of the proposal actions will be executed indirectly through
     * the associated {AccessManager}, and another indicating which will be scheduled during {queue}. Note that
     * those that must be scheduled are cancellable by `AccessManager` guardians.
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -97,17 +97,20 @@ abstract contract GovernorTimelockAccess is Governor {
...
+        delay = plan.delay;
+        indirect = new bool[](length);
+        withDelay = new bool[](length);
         for (uint256 i = 0; i < length; ++i) {
-            (indirect[i], ) = _getManagerData(plan, i);
+            (indirect[i], withDelay[i], ) = _getManagerData(plan, i);
         }
 
-        return (delay, indirect);
+        return (delay, indirect, withDelay);
```
frangio, line 113: I'm really struggling with this new convention... it's completely redundant to have this statement given the named return variables. I feel like I would rather remove it, or not have the named return variables.

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -223,12 +231,12 @@ abstract contract GovernorTimelockAccess is Governor {
...
-                    // Attempt to cancel considering the operation could have been cancelled and rescheduled already
-                    uint32 canceledNonce = _manager.cancel(address(this), targets[i], calldatas[i]);
-                    if (canceledNonce != nonce) {
-                        revert GovernorMismatchedNonce(proposalId, nonce, canceledNonce);
+                (, bool withDelay, uint32 nonce) = _getManagerData(plan, i);
+                if (withDelay) {
+                    bytes32 operationId = keccak256(abi.encode(address(this), targets[i], calldatas[i]));
+                    if (nonce == _manager.getNonce(operationId)) {
+                        // Attempt to cancel considering the operation could have been cancelled and rescheduled already
+                        try _manager.cancel(address(this), targets[i], calldatas[i]) returns (uint32) {} catch {}
```
vladyan18, line 239: Isn't the try-catch too "soft" here? The goal is to process exactly the case where it has already been cancelled, and not any possible reasons for the revert

contracts/governance/extensions/GovernorTimelockAccess.sol
Amxx: IMO, the named return variables are really usefull.

We have multiple option:
- redeclare the variable with a different name inside the body and do an explicit return. I'm not a fan of that because we need a different name for the same variable ...
- using the nameds return variable, and do an explicit return (like here). Its as explicit as one can be, but it can also be confusing. We actually don't need that last part
- remove the explicit return ... would that be more error prone ???

I tink we need to come up with guidelines for then entire repo.

contracts/governance/extensions/GovernorTimelockAccess.sol
Amxx: We want to make sure that the `cancel` function on the manager possibly failling for one of the calls does not prevent us moving forward (and attempting to cancel the other calls).

Do you have any idea for alternative implementations?

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -223,12 +232,12 @@ abstract contract GovernorTimelockAccess is Governor {
...
             for (uint256 i = 0; i < targets.length; ++i) {
-                (bool delayed, uint32 nonce) = _getManagerData(plan, i);
-                if (delayed) {
-                    // Attempt to cancel considering the operation could have been cancelled and rescheduled already
-                    uint32 canceledNonce = _manager.cancel(address(this), targets[i], calldatas[i]);
-                    if (canceledNonce != nonce) {
-                        revert GovernorMismatchedNonce(proposalId, nonce, canceledNonce);
+                (, bool withDelay, uint32 nonce) = _getManagerData(plan, i);
+                if (withDelay) {
+                    bytes32 operationId = keccak256(abi.encode(address(this), targets[i], calldatas[i]));
```
frangio, line 237: Let's make `hashOperation` a public function in the AccessManager and use it here.

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio, line 1: We need to add the override mechanism.

contracts/governance/extensions/GovernorTimelockAccess.sol
Amxx: It wouldn't be that ugly. What I want to understand is the benefit/drawbacks of doing that.

- What are the case where that would cause the cancel to fail?
- Are they cases where we want to fail?
- Are they cases where we dont want to fail?

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -92,22 +109,63 @@ abstract contract GovernorTimelockAccess is Governor {
...
+     * See Security Considerations above.
+     */
+    function setAccessManagerIgnored(
+        address target,
+        bytes4[] calldata selectors,
+        bool ignored
+    ) public virtual onlyGovernance {
+        for (uint256 i = 0; i < selectors.length; ++i) {
+            bytes4 selector = selectors[i];
+            _setAccessManagerIgnored(target, selector, ignored);
```
Amxx, line 133: If we are not reusing the selector, do we get anything from caching it?
```suggestion
            _setAccessManagerIgnored(target, selectors[i], ignored);
```

contracts/governance/extensions/GovernorTimelockAccess.sol
```diff
@@ -220,15 +285,19 @@ abstract contract GovernorTimelockAccess is Governor {
...
-                    // Attempt to cancel considering the operation could have been cancelled and rescheduled already
-                    uint32 canceledNonce = _manager.cancel(address(this), targets[i], calldatas[i]);
-                    if (canceledNonce != nonce) {
-                        revert GovernorMismatchedNonce(proposalId, nonce, canceledNonce);
+                (, bool withDelay, uint32 nonce) = _getManagerData(plan, i);
+                // Only attempt to cancel if the execution plan included a delay
+                if (withDelay) {
+                    bytes32 operationId = _manager.hashOperation(address(this), targets[i], calldatas[i]);
+                    // Check first if the current operation nonce is the one that we observed previously. It could
+                    // aready have been cancelled and rescheduled. We don't want to cancel unless it is exactly the
```
Amxx, line 296:
```suggestion
                    // already have been cancelled and rescheduled. We don't want to cancel unless it is exactly the
```

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: No, this was leftover from previous code.

contracts/governance/extensions/GovernorTimelockAccess.sol
frangio: Added comments that should explain a bit more.

frangio:
@Amxx This PR should be ready now.

Amxx:
> @Amxx This PR should be ready now.

LGTM. 

Its my PR, so I cannot formally approve: consider this message my approval. Fell free to add you approval and merge.

PR #4597, opened 2023-09-13
Fix function documentation in Nonces

0xneves:
The description of the function nonce(address) had a description not matching its true behavior. 

Current behavior: 
Return the next nonce of an address

Expect behavior:
Return the current nonce of an address

### DIFF
--- a/contracts/utils/Nonces.sol
+++ b/contracts/utils/Nonces.sol
@@ -13,7 +13,7 @@ abstract contract Nonces {
     mapping(address account => uint256) private _nonces;
 
     /**
-     * @dev Returns an the next unused nonce for an address.
+     * @dev Returns the current nonce for an address.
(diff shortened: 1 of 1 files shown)


contracts/utils/Nonces.sol
```diff
@@ -13,7 +13,7 @@ abstract contract Nonces {
     mapping(address account => uint256) private _nonces;
 
     /**
-     * @dev Returns an the next unused nonce for an address.
+     * @dev Returns the current nonce for an address.
```
frangio, line 16: The existing description is accurate, but has a grammar error. This function returns the "next unused nonce", i.e. the nonce that will work when passed to `_useCheckedNonce`.

```suggestion
     * @dev Returns the next unused nonce for an address.
```

PR #4599, opened 2023-09-14
Rename ProposalCore.eta to etaSeconds

frangio:
The other members in the struct use the Governor clock, whereas the `eta` member specifically uses timestamps, so we rename the variable to reduce ambiguity.

Fixes LIB-1080

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -40,7 +40,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         uint32 voteDuration;
         bool executed;
         bool canceled;
-        uint48 eta;
+        uint48 etaSeconds;
     }
 
     bytes32 private constant ALL_PROPOSAL_STATES_BITMAP = bytes32((2 ** (uint8(type(ProposalState).max) + 1)) - 1);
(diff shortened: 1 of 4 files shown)


ernestognw (commented):
If I'm not mistaken, change is also needed in:

- GovernorTimelockAccess
- IGovernor
- GovernorTimelockCompound

frangio:
@ernestognw I edited the event parameter. Did you also mean the `eta` local variables? What do you think about just using `eta` there? It's clear from context in most cases that it's timestamps.

ernestognw:
> @ernestognw I edited the event parameter. Did you also mean the `eta` local variables? What do you think about just using `eta` there? It's clear from context in most cases that it's timestamps.

Yeah, for IGovernor I was referring to the event. For other contracts, I don't see why is it clearer on the other two contracts. The issue is that someone may make a mistake by misinterpreting the `eta` so I'd say it's a similar case regardless.

frangio:
I was concerned that it might make the code less clear by being more verbose. But if we consider the ambiguity dangerous it's hard to argue why local variables would be different. My idea was that it's not ambiguous when it looks like `eta = Time.timestamp()`, but it's not always like that.

ernestognw:
Yeah although they're local I agree it's dangerous even for us. More than once I've had this hunch while reading the code that the eta is incompatible until I realize we've documented it doesn't follow the clock.

Naming is an important indicator here imo

ernestognw (approved):

Amxx (approved):

PR #4600, opened 2023-09-14
Wrap docstrings to 120 chars

frangio:
For readability in certain environments.

Fixes LIB-1077

### DIFF
--- a/contracts/utils/structs/Checkpoints.sol
+++ b/contracts/utils/structs/Checkpoints.sol
@@ -33,14 +33,16 @@ library Checkpoints {
      *
      * Returns previous value and new value.
      *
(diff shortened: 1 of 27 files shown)


contracts/utils/structs/Checkpoints.sol
```diff
@@ -33,14 +33,16 @@ library Checkpoints {
      *
      * Returns previous value and new value.
      *
-     * IMPORTANT: Never accept `key` as a user input, since an arbitrary `type(uint32).max` key set will disable the library.
+     * IMPORTANT: Never accept `key` as a user input, since an arbitrary `type(uint32).max`
```
Amxx, line 36: This doesn't look like 120 chars.

PR #4606, opened 2023-09-15
Remove Time.Delay *At functions

Amxx:
The `Time.Delay` library is partially generic for different clock types, but the important `withUpdate` function is hardcoded to work with timestamps and it's not as simple as the rest of the functions to make it work with block numbers. This PR removes the `*At` variants of functions to leave a timestamp-only interface for the time being.

Fixes LIB-1047

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/types/Time.sol
+++ b/contracts/utils/types/Time.sol
@@ -103,10 +103,15 @@ library Time {
      * enforce the old delay at the moment of the update. Returns the updated Delay object and the timestamp when the
      * new delay becomes effective.
      */
-    function withUpdate(Delay self, uint32 newValue, uint32 minSetback) internal view returns (Delay, uint48) {
-        uint32 value = self.get();
+    function withUpdate(
+        Delay self,
+        uint48 clock,
+        uint32 newValue,
+        uint32 minSetback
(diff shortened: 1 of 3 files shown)


contracts/utils/types/Time.sol
```diff
@@ -103,10 +103,15 @@ library Time {
...
      * new delay becomes effective.
      */
-    function withUpdate(Delay self, uint32 newValue, uint32 minSetback) internal view returns (Delay, uint48) {
-        uint32 value = self.get();
+    function withUpdate(
+        Delay self,
+        uint48 clock,
+        uint32 newValue,
+        uint32 minSetback
+    ) internal pure returns (Delay, uint48) {
```
frangio, line 111: The issue with this function is that `clock` is expected to always be the "current" timepoint, and the semantics of passing some other timepoint are quite unclear: if there is some existing pending delay update, `withUpdate` does not result in a delay object with two future updates, it may just jump ahead of time to the existing future update.

frangio (approved):

ernestognw:
LGTM lmao

PR #4612, opened 2023-09-18
Fix `AccessManager._checkAuthorized` in `execute` context

ernestognw:
Admin functions in the AccessManager restricted with the `onlyAuthorized` modifier were not callable via `execute`. This was because `_canCallExtended` ignored the case where `caller == this`. This is particularly important for `GovernorTimelockAccess`, which assumes that anything with restricted access can be invoked through `execute` and always does so.

`_canCallExtended` is also refactored slightly to separate the logic for `target == this`.

This PR does not include tests for the fix because we're cherry picking fixes done in the middle of a test refactor, but they are included in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4613.

### DIFF
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -845,6 +845,8 @@ contract AccessManager is Context, Multicall, IAccessManager {
      */
     function _canCallSelf(address caller, bytes calldata data) private view returns (bool immediate, uint32 delay) {
         if (caller == address(this)) {
+            // Caller is AccessManager, this means the call was sent through {execute} and it already checked
+            // permissions. We verify that the call "identifier", which is set during {execute}, is correct.
             return (_isExecuting(address(this), bytes4(data)), 0);
         }
 
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -841,7 +841,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
     }
 
     /**
-     * @dev A version of {canCall} that checks for admin restructions in this contract.
+     * @dev A version of {canCall} that checks for admin restrictions in this contract.
      */
     function _canCallSelf(address caller, bytes calldata data) private view returns (bool immediate, uint32 delay) {
         if (caller == address(this)) {
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -760,7 +760,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      */
     function _checkAuthorized() private {
         address caller = _msgSender();
-        (bool immediate, uint32 delay) = _canCallExtended(caller, address(this), _msgData());
+        (bool immediate, uint32 delay) = _canCallSelf(caller, _msgData());
         if (!immediate) {
             if (delay == 0) {
                 (, uint64 requiredRole, ) = _getAdminRestrictions(_msgData());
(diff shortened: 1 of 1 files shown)


frangio:
There is a lot in this PR, can we reduce it to the `_canCallExtended` refactor and the fix for the `execute` issue?

frangio:
Can we make this PR focused exclusively on the refactor of `_canCallExtended` and the fix to the `onlyAuthorized` issue?

It will be better to review it on its own, and we don't need to include the other changes in the first release candidate.

contracts/access/manager/AccessManager.sol
```diff
@@ -13,23 +13,32 @@ import {Time} from "../../utils/types/Time.sol";
...
  *
- * There is a special role defined by default named "public" which all accounts automatically have.
+ * * The target's {AccessManaged-authority} via {updateAuthority}.
+ * * Close or open a target via {setTargetClosed} keeping the permissions intact.
+ * * Each role's permissions through {setTargetFunctionRole}.
  *
- * In addition to the access rules defined by each target's functions being assigned to roles, then entire target can
- * be "closed". This "closed" mode is set/unset by the admin using {setTargetClosed} and can be used to lock a contract
- * while permissions are being (re-)configured.
+ * By default every address is member of the `PUBLIC_ROLE` and every contract allows public access until configured otherwise.
```
Amxx, line 31: I'm really not sure that "every contract allows public access". If that is the case, it sounds like a huge risk.

contracts/access/manager/AccessManager.sol
```diff
@@ -406,7 +443,10 @@ contract AccessManager is Context, Multicall, IAccessManager {
     /**
      * @dev Internal version of {setRoleAdmin} without access control.
      *
-     * Emits a {RoleAdminChanged} event
+     * Emits a {RoleAdminChanged} event.
+     *
+     * NOTE: Setting the admin role as the `PUBLIC_ROLE` is allowed, but it will effectively allow
```
Amxx, line 448: Do we want to prevent that ? (same question for `_setRoleGuardian`)

ernestognw: I'd personally prevent both because I don't see why anyone would do that but we need to analyze a bit better.
We can do it in the tests PR #4613

contracts/access/manager/AccessManager.sol
ernestognw: Let's keep the docs discussion in #4613

ernestognw:
> Can we make this PR focused exclusively on the refactor of `_canCallExtended` and the fix to the `onlyAuthorized` issue?
> 
> It will be better to review it on its own, and we don't need to include the other changes in the first release candidate.

Sure thing, just updated and also submitted #4613 with the tests WIP

contracts/access/manager/AccessManager.sol
```diff
@@ -833,25 +833,43 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
-            return (delay == 0, delay);
+            return _canCallSelf(caller, data);
         } else {
             bytes4 selector = bytes4(data);
             return canCall(caller, target, selector);
         }
     }
 
+    /**
+     * @dev A version of {canCall} that checks for admin restructions in this contract.
```
frangio, line 844:
```suggestion
     * @dev A version of {canCall} that checks for admin restrictions in this contract.
```

contracts/access/manager/AccessManager.sol
```diff
@@ -833,25 +833,43 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
+     */
+    function _canCallSelf(address caller, bytes calldata data) private view returns (bool immediate, uint32 delay) {
+        if (caller == address(this)) {
```
frangio, line 847: Copying the text from `canCall`:

```suggestion
        if (caller == address(this)) {
            // Caller is AccessManager, this means the call was sent through {execute} and it already checked
            // permissions. We verify that the call "identifier", which is set during {execute}, is correct.
```

frangio (commented):
Looks good. Minor comments:

Amxx (approved):
LGTM

PR #4613, opened 2023-09-18
Improve AccessManager tests

ernestognw:
Reaches 100% coverage in AccessManager

Every test that either makes use of the `_canCallExtended`, `_canCallSelf` or `canCall` is split in two sections:

- The restrictions tests: These only check the possible execution paths depending on the role and delays. They don't make any assertion other than expected errors.
- The state tests: These are setup with enough permissions so the call succeeds and check for state changes or event emissions.

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/access/manager/AccessManager.sol
```diff
@@ -120,39 +132,51 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
-     * We may be able to hash the operation, and check if the call was scheduled, but we would not be able to cleanup
-     * the schedule, leaving the possibility of multiple executions. Maybe this function should not be view?
+     * If `immediate` is true, the delay can be disregarded and the operation can be immediately executed, otherwise
+     * the operation can be executed if and only if delay is greater than 0.
      *
      * NOTE: The IAuthority interface does not include the `uint32` delay. This is an extension of that interface that
      * is backward compatible. Some contracts may thus ignore the second return argument. In that case they will fail
      * to identify the indirect workflow, and will consider calls that require a delay to be forbidden.
+     *
+     * NOTE: This function doesn't not report the permissions of this manager itself. These are defined by the
```
vladyan18, line 146: "doesn't not"

contracts/access/manager/AccessManager.sol
```diff
@@ -120,17 +132,25 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
```
ernestognw, line 146: Thanks @vladyan18!

```suggestion
     * NOTE: This function does not report the permissions of this manager itself. These are defined by the
```

contracts/access/manager/AccessManager.sol
```diff
@@ -13,23 +13,32 @@ import {Time} from "../../utils/types/Time.sol";
...
- * them defines a role, and may confer access to some of the restricted functions in the system, as configured by admins
- * through the use of {setFunctionAllowedRoles}.
+ * The restriction rules for such functions are defined in terms of "roles" identified by an `uint64` and scoped
+ * by target (`address`) and function selectors (`bytes4`). These roles are stored in this contract and can be
+ * configured by each target's admins (`ADMIN_ROLE` members) after a delay (see {getTargetAdminDelay}).
  *
- * Note that a function in a target contract may become permissioned in this way only when: 1) said contract is
- * {AccessManaged} and is connected to this contract as its manager, and 2) said function is decorated with the
- * `restricted` modifier.
+ * A target admin can configure the following:
```
Amxx, line 25:
```suggestion
 * For each target contract, admins can configure the following:
```

contracts/access/manager/AccessManager.sol
```diff
@@ -13,23 +13,32 @@ import {Time} from "../../utils/types/Time.sol";
...
  *
- * There is a special role defined by default named "public" which all accounts automatically have.
+ * * The target's {AccessManaged-authority} via {updateAuthority}.
+ * * Close or open a target via {setTargetClosed} keeping the permissions intact.
+ * * Each role's permissions through {setTargetFunctionRole}.
```
Amxx, line 29:
```suggestion
 * * The role allowed / not allowed to call a given function (identified by its selector) through {setTargetFunctionRole}.
```

contracts/access/manager/AccessManager.sol
```diff
@@ -13,23 +13,32 @@ import {Time} from "../../utils/types/Time.sol";
...
  *
- * In addition to the access rules defined by each target's functions being assigned to roles, then entire target can
- * be "closed". This "closed" mode is set/unset by the admin using {setTargetClosed} and can be used to lock a contract
- * while permissions are being (re-)configured.
+ * By default every address is member of the `PUBLIC_ROLE` and every contract allows public access until configured otherwise.
```
Amxx, line 31: By default: `_targets[<target>].allowedRoles[<selector>]` is `0`, meaning that only the ADMIN can call it. Its not public access by default.

```suggestion
 * By default every address is member of the `PUBLIC_ROLE` and every target function is restricted to the `ADMIN_ROLE` until configured otherwise.
```

contracts/access/manager/IAccessManager.sol
```diff
@@ -29,6 +29,13 @@ interface IAccessManager {
     event OperationCanceled(bytes32 indexed operationId, uint32 indexed nonce);
 
     event RoleLabel(uint64 indexed roleId, string label);
+    /**
+     * @dev Emitted when `account` is granted `roleId`.
+     *
+     * NOTE: The meaning of the `since` argument depends on the `newMember` argument.
+     * If the role is granted to a new member, the `since` argument indicates when the delay goes into effect,
+     * otherwise it indicates the time when the roleId comes into effect.
```
Amxx, line 37:
```suggestion
     * If the role is granted to a new member, the `since` argument indicates when the account gets the role,
     * otherwise it indicates the execution delay for this account and roleId is updated.
```

contracts/mocks/AccessManagedTarget.sol
```diff
@@ -15,4 +16,13 @@ abstract contract AccessManagedTarget is AccessManaged {
     function fnUnrestricted() public {
         emit CalledUnrestricted(msg.sender);
     }
+
+    function setIsConsumingScheduledOp(bool isConsuming) external {
+        // Memory layout is 0x....<_consumingSchedule (boolean)><authority (address)>
+        _consumingSchedule().value |= bytes32(uint256(isConsuming ? 1 : 0)) << 161;
```
Amxx, line 22: if bool `isConsuming` is false, and if the bit is already set, this function will not reset it.
```suggestion
        bytes32 mask = 1 << 161;
        if (isConsuming) {
            _consumingSchedule().value |= mask;
        } else {
            _consumingSchedule().value &= ~mask;
        }
```

contracts/utils/types/Time.sol
```diff
@@ -96,22 +96,22 @@ library Time {
      * enforce the old delay at the moment of the update. Returns the updated Delay object and the timestamp when the
      * new delay becomes effective.
      */
-    function withUpdate(Delay self, uint32 newValue, uint32 minSetback) internal view returns (Delay, uint48) {
+    function withUpdate(Delay self, uint32 newValue, uint32 minSetback) internal view returns (Delay, uint48 effect) {
```
Amxx, line 99:
```suggestion
    function withUpdate(Delay self, uint32 newValue, uint32 minSetback) internal view returns (Delay updatedDelay, uint48 effect) {
```

test/helpers/access-manager.js
```diff
@@ -0,0 +1,53 @@
...
+    SOME: {
+      id: web3.utils.toBN(42),
+    },
+    PUBLIC: {
+      id: MAX_UINT64,
+    },
+  };
+
+  // Names
+  Object.assign(roles, Object.fromEntries(Object.entries(roles).map(([k, v]) => [k, { ...v, name: k }])));
```
Amxx, line 24:
```suggestion
  Object.entries(roles).forEach(([ name, role ]) => role.name = name);
```

test/access/manager/AccessManager.test.js
```diff
@@ -1,1028 +1,2653 @@
...
+    it('initializes setup roles correctly', async function () {
+      for (const { id: roleId, admin, guardian, members } of Object.values(BASE_ROLES)) {
+        expect(await this.manager.getRoleAdmin(roleId)).to.be.bignumber.equal(admin.id);
+        expect(await this.manager.getRoleGuardian(roleId)).to.be.bignumber.equal(guardian.id);
+
+        const isMember = Object.fromEntries(members.map(member => [member, true]));
+
+        for (const user of this.roles.PUBLIC.members) {
+          expect(await this.manager.hasRole(roleId, user).then(formatAccess)).to.be.deep.equal([!!isMember[user], '0']);
+        }
```
Amxx, line 107: Would that work ?

```suggestion
        for (const user of this.roles.PUBLIC.members) {
          expect(await this.manager.hasRole(roleId, user).then(formatAccess)).to.be.deep.equal(members.includes(user), '0']);
        }
```

contracts/access/manager/AccessManager.sol
ernestognw: Accepting, just changed it a bit.

test/access/manager/AccessManager.test.js
ernestognw: Yes it worked, I was for some reason planning to cache the members

contracts/mocks/AccessManagedTarget.sol
ernestognw: I checked it out and yes, you're right. I'm updating with another bug I found from this.

contracts/access/manager/AccessManager.sol
```diff
@@ -13,23 +13,32 @@ import {Time} from "../../utils/types/Time.sol";
...
+ * {AccessManaged} contract, be connected to this contract as its manager and implement the {AccessManaged-restricted}
+ * modifier on a set of functions selected to be permissioned. Note that any function without this setup won't be
+ * effectively restricted.
  *
- * An AccessManager instance will define a set of roles. Accounts can be added into any number of these roles. Each of
- * them defines a role, and may confer access to some of the restricted functions in the system, as configured by admins
- * through the use of {setFunctionAllowedRoles}.
+ * The restriction rules for such functions are defined in terms of "roles" identified by an `uint64` and scoped
+ * by target (`address`) and function selectors (`bytes4`). These roles are stored in this contract and can be
+ * configured by each target's admins (`ADMIN_ROLE` members) after a delay (see {getTargetAdminDelay}).
```
ernestognw, line 23:
```suggestion
 * configured by admins (`ADMIN_ROLE` members) after a delay (see {getTargetAdminDelay}).
```

contracts/access/manager/AccessManager.sol
```diff
@@ -13,23 +13,32 @@ import {Time} from "../../utils/types/Time.sol";
...
  *
- * Note that a function in a target contract may become permissioned in this way only when: 1) said contract is
- * {AccessManaged} and is connected to this contract as its manager, and 2) said function is decorated with the
- * `restricted` modifier.
+ * For each target contract, admins can configure the following:
```
ernestognw, line 25:
```suggestion
 * For each target contract, admins can configure the following without any delay:
```

test/access/manager/AccessManager.test.js
```diff
@@ -1,1028 +1,2658 @@
...
+            const args = [984910, time.duration.days(2)];
+            this.calldata = this.manager.contract.methods[method](...args).encodeABI();
           });
 
-          await time.increase(grantDelay);
+          shouldBehaveLikeDelayedAdminOperation();
+        });
 
-          expect(await this.manager.hasRole(ROLES.SOME, user).then(formatAccess)).to.be.deep.equal([true, '0']);
+        it.only('reverts setting grant delay for the PUBLIC_ROLE', async function () {
```
frangio, line 959:
```suggestion
        it('reverts setting grant delay for the PUBLIC_ROLE', async function () {
```

ernestognw: Oh gosh, sorry.

Amxx (dismissed):
Solidity looks good. 
Don't have bandwidth to review tests right now. Will take a look later (that is not a merge blocker)

test/access/manager/AccessManager.test.js
```diff
@@ -1,1030 +1,2590 @@
...
-  PUBLIC: MAX_UINT64,
-};
-Object.assign(ROLES, Object.fromEntries(Object.entries(ROLES).map(([key, value]) => [value, key])));
-
-const executeDelay = web3.utils.toBN(10);
-const grantDelay = web3.utils.toBN(10);
-const MINSETBACK = time.duration.days(5);
-
-const formatAccess = access => [access[0], access[1].toString()];
+const BASE_ROLES = buildBaseRoles();
```
frangio, line 49: Please move this inside `contract`.

test/access/manager/AccessManager.test.js
```diff
@@ -1,1030 +1,2590 @@
...
-    expect(await this.manager.hasRole(ROLES.SOME_ADMIN, user).then(formatAccess)).to.be.deep.equal([false, '0']);
-    expect(await this.manager.hasRole(ROLES.SOME, admin).then(formatAccess)).to.be.deep.equal([false, '0']);
-    expect(await this.manager.hasRole(ROLES.SOME, manager).then(formatAccess)).to.be.deep.equal([false, '0']);
-    expect(await this.manager.hasRole(ROLES.SOME, member).then(formatAccess)).to.be.deep.equal([true, '0']);
-    expect(await this.manager.hasRole(ROLES.SOME, user).then(formatAccess)).to.be.deep.equal([false, '0']);
-    expect(await this.manager.hasRole(ROLES.PUBLIC, admin).then(formatAccess)).to.be.deep.equal([true, '0']);
-    expect(await this.manager.hasRole(ROLES.PUBLIC, manager).then(formatAccess)).to.be.deep.equal([true, '0']);
-    expect(await this.manager.hasRole(ROLES.PUBLIC, member).then(formatAccess)).to.be.deep.equal([true, '0']);
-    expect(await this.manager.hasRole(ROLES.PUBLIC, user).then(formatAccess)).to.be.deep.equal([true, '0']);
+  describe('during construction', function () {
```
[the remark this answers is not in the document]
ernestognw: Not sure if I follow the idea. The suite contains tests that don't share a common `beforeEach` and requiring instantiating an AccessManager each one.

Maybe we can do something like:

```js
it('defaults to ... during construction', ...)
it('reverts with zero address owner during construction', ...)
```

test/access/manager/AccessManager.test.js
```diff
@@ -1,1030 +1,2590 @@
...
-  });
+      // Set admin role avoiding default
+      if (admin.id !== this.roles.ADMIN.id) {
+        await this.manager.$_setRoleAdmin(roleId, admin.id);
+      }
+
+      // Set guardian role avoiding default
+      if (guardian.id !== this.roles.ADMIN.id) {
+        await this.manager.$_setRoleGuardian(roleId, guardian.id);
+      }
```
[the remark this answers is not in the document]
ernestognw: Of course we have:

- [Admin](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/033337b8566c406ca5abf49300497074862588db/test/access/manager/AccessManager.test.js#L435)
- [Guardian](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/033337b8566c406ca5abf49300497074862588db/test/access/manager/AccessManager.test.js#L451)

test/access/manager/AccessManager.test.js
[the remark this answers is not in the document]
ernestognw: Done!

test/access/manager/AccessManager.test.js
```diff
@@ -1,1030 +1,2590 @@
...
+      beforeEach('set calldata', function () {
+        this.calldata = '0x12345678';
+        this.role = { id: web3.utils.toBN(379204) };
+      });
+
+      shouldBehaveLikeCanCall({
+        closed: function () {
+          it('should return false and no delay', async function () {
+            const { immediate, delay } = await this.manager.canCall(
+              Wallet.generate().getChecksumAddressString(),
```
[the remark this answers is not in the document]
ernestognw: I moved the random address generatiorsn outside the test suite.

> Separately, why don't we use this.caller here?

The rationale is that `this.caller` is very likely to have some sort of permissions so I use random addresses wherever I think there could be an accidental override with a privileged account.

scripts/upgradeable/transpile.sh
```diff
@@ -6,7 +6,7 @@ VERSION="$(jq -r .version contracts/package.json)"
 DIRNAME="$(dirname -- "${BASH_SOURCE[0]}")"
 
 bash "$DIRNAME/patch-apply.sh"
-sed -i "s/<package-version>/$VERSION/g" contracts/package.json
+sed -i'' -e "s/<package-version>/$VERSION/g" "contracts/package.json"
```
Amxx, line 9: can you explain this change ?

test/helpers/access-manager.js
```diff
@@ -0,0 +1,69 @@
...
+  artifacts._getArtifactPathSync('AccessManagerUpgradeable');
+  artifacts._getArtifactPathSync('AccessManagedUpgradeable');
+
+  // ERC-7201 namespace location for AccessManager
+  EXECUTION_ID_STORAGE_SLOT += 0x40c6c8c28789853c7efd823ab20824bbd71718a8a5915e855f6f288c9a26ad00n;
+  // ERC-7201 namespace location for AccessManaged
+  CONSUMING_SCHEDULE_STORAGE_SLOT += 0xf3177357ab46d8af007ab3fdb9af81da189e1068fefdc0073dca88a2cab40a00n;
+} catch (_) {
+  // eslint-disable-next-line no-empty
+}
```
Amxx, line 60: Non blocker, but I would suggest we do:
```suggestion
function getSlotWithOptionalNamespaceOffset(base, contractName) {
    try {
        artifacts._getArtifactPathSync(`${contractName}Upgradeable`);
        return base + erc7201Location(`openzeppelin.storage.${contractName}`);
    } catch (_) {
        return base;
    }
}

let EXECUTION_ID_STORAGE_SLOT = getSlotWithOptionalNamespaceOffset(3n, 'AccessManagerUpgradeable);
let CONSUMING_SCHEDULE_STORAGE_SLOT = getSlotWithOptionalNamespaceOffset(0n, 'AccessManagedUpgradeable');
```
with `erc7201Location` in a dedicated helper (function is already implmented in `openzeppelin-transpiler/src/utils/erc7201.ts`

scripts/upgradeable/transpile.sh
ernestognw: Yes, it's a version that works with both Linux and Mac's `sed`.

More [here](https://stackoverflow.com/a/4247319).

test/helpers/access-manager.js
ernestognw: I really like the idea. Let's do it after merging this in #4632, [this file is repeated there](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4632/files#diff-14bce8fea9388f45cc5a3661e6fd010ae6f2b2871876904db78abe3204cead44) and by changing it here their histories will diverge.

scripts/upgradeable/transpile.sh
Amxx: I learned something today ... and it reinforces my believe that true devs are not/shoud not be using mac

ernestognw: I'll be happy not being a true dev, thanks!

Amxx: Your using a mac ? All my condolences (particularly on being stuck with a non standard keyboard)

ernestognw: Not a problem on my side. I honestly don't care what you're using as long as we keep a workspace that works for both

Amxx (approved):
Everything LGTM. 

Again, did no finish reviewing all the tests. The part I reviewed looks good. Contracts loog good.

I raised some comments that should be addressed in a separate PR (they are about the testing part, so don't affect the 5.0 release)

Issue #4617, opened 2023-09-19
Add memory array library to utils


marjon-call:
**🧐 Motivation**
Add the attached library to the utils folder. This library allows developers to manipulate memory arrays. This means that developers are no longer bound to only working with static arrays in memory.

**📝 Details**
Below is the link to the repo:
https://github.com/marjon-call/memLibrary

I suggest adding this to the utils folder, so that any developer can access dynamic arrays in memory.

PR #4620, opened 2023-09-20
Added contracts/utils/MemoryArray.sol

akash-kumar-saw:
Fixes Issue #4617

Added library that allows developers to manipulate memory arrays efficiently.

File Added : contracts/utils/MemoryArray.sol


akash-kumar-saw:
Hi @marjon-call,

You can review this PR and can approve it.

Thanks for your precious time : )

Amxx (commented):
The code proposed is funamentally flawed. It "work" (with high gas cost) when the array being manipulated is the last one allocated, but other cases break. If we want to provide these features (I'm not sure we want that) we need a completelly new implementation.

PR #4624, opened 2023-09-25
Add named return parameters and `_checkSelector` function to AccessManager

ernestognw:
Improvements after #4613

The `_checkSelector` function is added to keep consistency of the way the calldata selector is consumed.

#### PR Checklist

- [x] Tests in #4613 
- [ ] Documentation

### DIFF
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -130,7 +130,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * is backward compatible. Some contracts may thus ignore the second return argument. In that case they will fail
      * to identify the indirect workflow, and will consider calls that require a delay to be forbidden.
      */
-    function canCall(address caller, address target, bytes4 selector) public view virtual returns (bool, uint32) {
+    function canCall(address caller, address target, bytes4 selector) public view virtual returns (bool immediate, uint32 delay) {
         if (isTargetClosed(target)) {
             return (false, 0);
         } else if (caller == address(this)) {
@@ -139,8 +139,9 @@ contract AccessManager is Context, Multicall, IAccessManager {
             return (_isExecuting(target, selector), 0);
         } else {
             uint64 roleId = getTargetFunctionRole(target, selector);
-            (bool isMember, uint32 currentDelay) = hasRole(roleId, caller);
-            return isMember ? (currentDelay == 0, currentDelay) : (false, 0);
+            bool inRole;
+            (inRole, delay) = hasRole(roleId, caller);
+            return inRole ? (delay == 0, delay) : (false, 0);
         }
     }
 
@@ -220,11 +221,11 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * [2] Pending execution delay for the account.
      * [3] Timestamp at which the pending execution delay will become active. 0 means no delay update is scheduled.
      */
-    function getAccess(uint64 roleId, address account) public view virtual returns (uint48, uint32, uint32, uint48) {
+    function getAccess(uint64 roleId, address account) public view virtual returns (uint48 since, uint32 currentDelay, uint32 pendingDelay, uint48 effect) {
         Access storage access = _roles[roleId].members[account];
 
-        uint48 since = access.since;
-        (uint32 currentDelay, uint32 pendingDelay, uint48 effect) = access.delay.getFull();
+         since = access.since;
+        (currentDelay, pendingDelay, effect) = access.delay.getFull();
 
         return (since, currentDelay, pendingDelay, effect);
     }
@@ -233,7 +234,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * @dev Check if a given account currently had the permission level corresponding to a given role. Note that this
      * permission might be associated with a delay. {getAccess} can provide more details.
      */
-    function hasRole(uint64 roleId, address account) public view virtual returns (bool, uint32) {
+    function hasRole(uint64 roleId, address account) public view virtual returns (bool isMember, uint32 executionDelay) {
         if (roleId == PUBLIC_ROLE) {
             return (true, 0);
         } else {
@@ -439,10 +440,6 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * Emits a {RoleGrantDelayChanged} event
      */
     function _setGrantDelay(uint64 roleId, uint32 newDelay) internal virtual {
-        if (roleId == PUBLIC_ROLE) {
-            revert AccessManagerLockedRole(roleId);
-        }
-
         uint48 effect;
         (_roles[roleId].grantDelay, effect) = _roles[roleId].grantDelay.withUpdate(newDelay, minSetback());
 
@@ -578,7 +575,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
 
         // if call is not authorized, or if requested timing is too soon
         if ((!immediate && setback == 0) || (when > 0 && when < minWhen)) {
-            revert AccessManagerUnauthorizedCall(caller, target, bytes4(data[0:4]));
+            revert AccessManagerUnauthorizedCall(caller, target, _checkSelector(data));
         }
 
         // Reuse variable due to stack too deep
@@ -631,7 +628,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
 
         // If caller is not authorised, revert
         if (!immediate && setback == 0) {
-            revert AccessManagerUnauthorizedCall(caller, target, bytes4(data));
+            revert AccessManagerUnauthorizedCall(caller, target, _checkSelector(data));
         }
 
         // If caller is authorised, check operation was scheduled early enough
@@ -644,7 +641,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
 
         // Mark the target and selector as authorised
         bytes32 executionIdBefore = _executionId;
-        _executionId = _hashExecutionId(target, bytes4(data));
+        _executionId = _hashExecutionId(target, _checkSelector(data));
 
         // Perform call
         Address.functionCallWithValue(target, data, msg.value);
@@ -707,7 +704,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      */
     function cancel(address caller, address target, bytes calldata data) public virtual returns (uint32) {
         address msgsender = _msgSender();
-        bytes4 selector = bytes4(data[0:4]);
+        bytes4 selector = _checkSelector(data);
 
         bytes32 operationId = hashOperation(caller, target, data);
         if (_schedules[operationId].timepoint == 0) {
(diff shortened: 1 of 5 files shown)


contracts/access/manager/AccessManager.sol
```diff
@@ -439,10 +450,6 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * Emits a {RoleGrantDelayChanged} event
      */
     function _setGrantDelay(uint64 roleId, uint32 newDelay) internal virtual {
-        if (roleId == PUBLIC_ROLE) {
-            revert AccessManagerLockedRole(roleId);
-        }
```
frangio, line 444: Note to self and others: this is removed because the public role is not grantable according to the logic of `_grantRole`.

contracts/access/manager/AccessManager.sol
```diff
@@ -878,4 +890,11 @@ contract AccessManager is Context, Multicall, IAccessManager {
     function _isExpired(uint48 timepoint) private view returns (bool) {
         return timepoint + expiration() <= Time.timestamp();
     }
+
+    /**
+     * @dev Extracts the selector from calldata. Panics if data is not at least 4 bytes
+     */
+    function _checkSelector(bytes calldata data) private pure returns (bytes4) {
```
frangio, line 897: Why not `_getSelector`?

ernestognw: We've been using consistently a convention for reverting functions prefixing them with `_check*`.

I think that sort of makes visually clearer that it reverts (panics really).

frangio: I guess my question is because those functions usually don't return anything.

It would make sense for me if `_getSelector` reverts when it's not possible to get a selector. Curious about @Amxx's thoughts.

ernestognw: My thought was the opposite because in this case I think is more important to make the panic more evident. Even if we don't guarantee an error interface yet, we should try to be as clear as possible.

I'd also like to hear @Amxx's opinion

contracts/access/manager/AccessManager.sol
Amxx: Given that the public role is indeed not grantable, does it make sens to set the grant delay ? IMO it send the wrong signal that you can set a delay, and emit the corresponding event, when you know the delay will never be enforced.

contracts/access/manager/AccessManager.sol
Amxx: I don't like this function, because it will panic (revert) in some cases where IMO it should just says no. This will cause some "getter" functions to revert, which imo is bad.

IMO the only place where its unclear is in cancel. 

Please see [this diff](https://github.com/OpenZeppelin/openzeppelin-contracts/compare/master...Amxx:openzeppelin-contracts:fv/AccessManager#diff-d33289f3eb8b3043d73cb1803745e64ae1338991933016a3230321df1045d38b) (subset of the FV PR)

contracts/access/manager/AccessManager.sol
ernestognw: Perhaps documentation is fine for the case of granting a delay to the PUBLIC_ROLE, I wouldn't add an extra check for a case that is low likely to happen

contracts/access/manager/AccessManager.sol
ernestognw: > This will cause some "getter" functions to revert, which imo is bad.

Can you give a concrete example? I couldn't find one getter function reverting because of that.

> IMO the only place where its unclear is in cancel.

Yeah `cancel` is definitely a place, but I'd consider inconsistent the current `schedule` function. This is because even when it's supposed to revert with an `AccessManagerUnauthorizedCall` it panics.

Also, using only `bytes4` pads the result, which can be misleading (eg. `AccessManagerUnauthorizedCall` reports a different selector to what it was actually used).

ernestognw: > I guess my question is because those functions usually don't return anything.

@frangio, it's a similar case with `_useAndCheckNonce`. We can call it `_getAndCheckSelector`.

contracts/access/manager/AccessManager.sol
```diff
@@ -779,8 +786,10 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
      */
-    function _getAdminRestrictions(bytes calldata data) private view returns (bool, uint64, uint32) {
-        bytes4 selector = bytes4(data);
+    function _getAdminRestrictions(
+        bytes calldata data
+    ) private view returns (bool restricted, uint64 roleAdminId, uint32 executionDelay) {
+        bytes4 selector = _checkSelector(data);
 
         if (data.length < 4) {
             return (false, 0, 0);
```
frangio, line 795: We need to swap these two so that this function doesn't revert.

contracts/access/manager/AccessManager.sol
```diff
@@ -831,11 +839,15 @@ contract AccessManager is Context, Multicall, IAccessManager {
...
+        address caller,
+        address target,
+        bytes calldata data
+    ) private view returns (bool immediate, uint32 delay) {
         if (target == address(this)) {
             return _canCallSelf(caller, data);
         } else {
-            bytes4 selector = bytes4(data);
+            bytes4 selector = _checkSelector(data);
             return canCall(caller, target, selector);
```
Amxx, line 851:
```suggestion
            return data.length < 4 ? (false, 0) : canCall(caller, target, _checkSelector(data));
```

contracts/access/manager/AccessManager.sol
```diff
@@ -779,8 +786,10 @@ contract AccessManager is Context, Multicall, IAccessManager {
      * - uint64: which role is this operation restricted to
      * - uint32: minimum delay to enforce for that operation (on top of the admin's execution delay)
      */
-    function _getAdminRestrictions(bytes calldata data) private view returns (bool, uint64, uint32) {
-        bytes4 selector = bytes4(data);
+    function _getAdminRestrictions(
+        bytes calldata data
+    ) private view returns (bool restricted, uint64 roleAdminId, uint32 executionDelay) {
+        bytes4 selector = _checkSelector(data);
```
Amxx, line 792: Move this a fex lines bellow, after:
```
if (data.length < 4) {
    return (false, 0, 0);
}
````

File: contracts/access/manager/AccessManager.sol [hunk 6]
```diff
@@ -847,7 +859,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
         if (caller == address(this)) {
```
Amxx, line 859: Here, do

```
if (data.length < 4) {
    return (false, 0);
} else if (caller == address(this)) {
```

contracts/access/manager/AccessManager.sol
ernestognw: We decided to keep `_checkSelector` after a meeting.

contracts/access/manager/AccessManager.sol
frangio: >IMO it send the wrong signal that you can set a delay, and emit the corresponding event, when you know the delay will never be enforced

This is an interesting point, but this is the grant delay for a role that every account automatically has. Observing `RoleGrantDelayChanged` for such a role should provide no information.

Amxx: > I can't think of any incorrect conclusion someone could draw from this event.

I can imagine someone thinking that new user (if that means anything) are in the grant role with a delay. For example that when a create a new seedphrase/private key (or when I create a new account smart contract) I'll get in the public group after a delay.

Sure that makes no sens, but I'm pretty sure some people out these will believe that.

Amxx (approved):
LGTM, with a minor comment.

contracts/access/manager/AccessManager.sol
Amxx: Note that in `_setRoleAdmin` and `_setRoleGuardian` we don't allow setting the admin/guardian for the `PUBLIC_ROLE`. I'm not sure why that would be forbiden, but setting the grantDelay would be authorized.

contracts/access/manager/AccessManager.sol
```diff
@@ -139,8 +143,9 @@ contract AccessManager is Context, Multicall, IAccessManager {
             return (_isExecuting(target, selector), 0);
         } else {
             uint64 roleId = getTargetFunctionRole(target, selector);
-            (bool isMember, uint32 currentDelay) = hasRole(roleId, caller);
-            return isMember ? (currentDelay == 0, currentDelay) : (false, 0);
+            bool inRole;
+            (inRole, delay) = hasRole(roleId, caller);
+            return inRole ? (delay == 0, delay) : (false, 0);
```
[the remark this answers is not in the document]
ernestognw: I agree with your point, I see it risky to have an assignment to the return value here, specially because multiple things around the contract depend on the delay returned being 0 if the user doesn't have the role. I'm updating

Amxx (approved):

frangio:
What do we do with:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/2472e51e80e581a90370cfd59d6bba6a12bfdf52/contracts/governance/extensions/GovernorTimelockAccess.sol#L195

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/2472e51e80e581a90370cfd59d6bba6a12bfdf52/contracts/access/manager/AccessManaged.sol#L109

Amxx:
In `AccessManaged`, we could do
```
function _checkCanCall(address caller, bytes calldata data) internal virtual {
  if (data.length < 4) {
    revert AccessManagedUnauthorized(caller);
  }
  (bool immediate, uint32 delay) = AuthorityUtils.canCallWithDelay(
  ...
}
```

Amxx:
In the governor whe should definitelly not revert. We don't know where the proposed call is going. Could be calling a fallback function, or a custom contract that doesn't use 4 bytes selectors

frangio:
My opinion is:

- `AccessManaged`: Panic. This is dev error, they used the modifier on the fallback function or something like that
- `GovernorTimelockAccess`: Add `if (calldatas[i].length < 4) continue;`. Anything else is arguably a bug?

Amxx:
> GovernorTimelockAccess: Add if (calldatas[i].length < 4) continue;. Not doing this is arguably a bug?

yes

ernestognw:
In `AccessManaged` there's no way of getting less than 4 bytes calldata, so I agree with Fran.
For the case of `GovernorTimelockAccess` that's definitely a bug!

frangio:
The operation will not be ignored, as far as I can tell. We would just never ask the AccessManager about that operation, and it would be executed directly on the target.

ernestognw:
Tested short calldata in the `restricted` function of the `AccessManaged` in [this commit](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4632/commits/f63793c0d4068cb3478bdb47906709f636c40b8c)

frangio (approved):
LGTM

Amxx (approved):

frangio:
I didn't realize but should've added a changeset. Needed to trigger the release automation.

PR #4627, opened 2023-09-26
Improve documentation about backwards compatibility

frangio:
It's important to remind people of storage layout incompatibilities of 5.0 with respect to 4.9. The rest of the documentation about backwards compatibility needed a refresh too.


docs/modules/ROOT/pages/backwards-compatibility.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+== Storage Layout
+
+Minor and patch updates always preserve storage layout compatibility. This means that a live contract can be upgraded from one minor to another without corrupting the storage layout. In some cases it may be necessary to initialize new state variables when upgrading, although we expect this to be infrequent.
+
+We recommend using xref:upgrades-plugins::index.adoc[OpenZeppelin Upgrades Plugins or CLI] to ensure storage layout safety of upgrades.
+
+== Solidity Version
+
+The minimum Solidity version required to compile the contracts will remain unchanged un minor and patch updates. New contracts in minor releases may make use of newer Solidity features and require a more recent version of the compiler.
```
Amxx, line 48:
```suggestion
The minimum Solidity version required to compile the contracts will remain unchanged in minor and patch updates. New contracts in minor releases may make use of newer Solidity features and require a more recent version of the compiler.
```

docs/modules/ROOT/pages/backwards-compatibility.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+Infrequently a patch or minor update will remove or change an API in a breaking way, but only if the previous API is considered insecure. These breaking changes will be noted in the changelog and release notes, and published along with a security advisory.
+
+=== Draft or Pre-Final ERCs
+
+ERCs that are not Final can change in incompatible ways. For this reason, we avoid shipping implementations of ERCs before they are Final. Some exceptions are made for ERCs that have been published for a long time and that seem unlikely to change. Breaking changes to the ERC specification are still technically possible in those cases, so these implementations are published in files named `draft-*.sol` to make that condition explicit.
+
+=== Virtual & Overrides
+
+All functions in this library are virtual, but this does not mean that overrides are encouraged. There is a subset of functions that are designed to be overridden. By defining overrides outside of this subset you are potentially relying on internal implementation details. We make efforts to preserve backwards compatibility even in these cases but it is extremely difficult and easy to accidentally break. Caution is advised.
```
ernestognw, line 20:
```suggestion
Almost all functions in this library are virtual with some exceptions, but this does not mean that overrides are encouraged. There is a subset of functions that are designed to be overridden. By defining overrides outside of this subset you are potentially relying on internal implementation details. We make efforts to preserve backwards compatibility even in these cases but it is extremely difficult and easy to accidentally break. Caution is advised.
```

Amxx (approved):

Issue #3928, opened 2021-11-09
Don't rename interfaces IWhatever -> IWhateverUpgradeable


gnarvaja:
Also reported some time ago in https://forum.openzeppelin.com/t/why-contracts-upgradeable-has-ierc721upgradeable-and-i-whatever-upgradeable/14793 

**💻 Environment**

openzeppelin-contracts == 4.3.2
openzeppelin-contracts-upgradeable == 4.3.2

**📝 Details**

The package openzeppelin-contracts-upgradeable renames the interfaces when they are exactly the same as non-upgradeable (vanilla) versions. This isn't necessary and brings a lot of issues like this one

**🔢 Code to reproduce bug**

```solidity
interface IMyContract is IERC721 {
...
}

contract MyContract is IMyContract, ERC721Upgradeable {
...
}
```

Then when I compile I get errors like these:

```
11 | contract PolicyNFT is IERC721,
   | ^ (Relevant source part starts here and spans across multiple lines).
Note: Definition in "ERC721Upgradeable": 
   --> @openzeppelin/contracts-upgradeable/token/ERC721/ERC721Upgradeable.sol:184:5:
    |
184 |     function safeTransferFrom(
    |     ^ (Relevant source part starts here and spans across multiple lines).
Note: Definition in "IERC721": 
   --> @openzeppelin/contracts/token/ERC721/IERC721.sol:136:5:
    |
136 |     function safeTransferFrom(
    |     ^ (Relevant source part starts here and spans across multiple lines).

TypeError: Derived contract must override function "safeTransferFrom". Two or more base classes define function with same name and parameter types.
```

I think XXXUpgradeable contracts shouldn't define new interfaces but instead implement the standard ones. Interfaces are the same between upgradeable and non-upgradeable contracts.

If name clashes are an issue when importing files from `contracts` and `contracts-upgradeable`, those can be easily avoided using explicit imports

```solidity
import {IERC20Metadata} from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
import {ERC20Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
```

frangio:
This needs a few changes in the transpiler to get fixed. See https://github.com/OpenZeppelin/openzeppelin-transpiler/issues/108.

PR #4628, opened 2023-09-27
Enable partial transpilation for upgradeable package

frangio:
Things that don't need to be transpiled for upgradeability, such as interfaces, libraries, and some stateless contracts, are now imported into `@openzeppelin/contracts-upgradeable` from a peer dependency on `@openzeppelin/contracts`.

Uses https://github.com/OpenZeppelin/openzeppelin-transpiler/issues/130.

Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3928.


scripts/upgradeable/transpile.sh
```diff
@@ -36,7 +37,8 @@ npx @openzeppelin/upgrade-safe-transpiler@latest -D \
   -x '!contracts/proxy/beacon/IBeacon.sol' \
   -p 'contracts/**/presets/**/*' \
   -n \
-  -N 'contracts/mocks/**/*'
+  -N 'contracts/mocks/**/*' \
+  -q '@openzeppelin/'
```
frangio, line 41: @Amxx Does this look right to you? We have to do this because the contract source paths already have a `contracts/` prefix, but it's not really that `@openzeppelin/` is the peer dependency. Ideally I would write something like `@openzeppelin/contracts/=contracts/`, like a Foundry remapping, but I don't want to complicate the transpiler more...

Amxx: I used to do `-q '@openzeppelin/contracts/..` for my tests. What you wrote works just fine.

We can change to a remaping / remapings in a further version of the transpiler

.github/workflows/checks.yml
```diff
@@ -58,6 +58,14 @@ jobs:
...
       - name: Transpile to upgradeable
         run: bash scripts/upgradeable/transpile.sh
+      - name: Check out copy of vanilla
+        uses: actions/checkout@v4
+        with:
+          path: .vanilla
+      - name: Install vanilla as dependency
+        run: |
+          mkdir -p node_modules/@openzeppelin
+          ln -sr ".vanilla/contracts" node_modules/@openzeppelin/contracts
```
Amxx, line 68: Would that work ?

```suggestion
      - name: Prepare folder for for vanilla copy
        run:
          mkdir -p node_modules/@openzeppelin
      - name: Check out copy of vanilla
        uses: actions/checkout@v4
        with:
          path: node_modules/@openzeppelin
```

Amxx (approved):

.github/workflows/checks.yml
Amxx, line 68: Given that the contracts are already checked out, would it be simpler/faster to do 
```suggestion
      - name: Copy non-upgradeable contracts as dependency
        run:
          cp -r contracts node_modules/@openzeppelin/contracts
```

scripts/prepack.sh
```diff
@@ -4,9 +4,24 @@ set -euo pipefail
...
+mkdirp build/contracts
 cp artifacts/contracts/**/*.json build/contracts
 rm build/contracts/*.dbg.json
 
 node scripts/remove-ignored-artifacts.js
+
+cp README.md contracts/
+
+mkdirp contracts/build/contracts
+cp -r build/contracts/*.json contracts/build/contracts
```
Amxx, line 27: Here, the artifacts are copied twice, to `build/contracts` and `contracts/build/contracts`.

It looks like this is a leftover from when we had 2 packages. Maybe we should simplify that

frangio: Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4628/commits/f63cf2da50589f776a2f0eabf045c2b679d6a3ca

Amxx (approved):
I think this is good.

package.json
```diff
@@ -2,9 +2,9 @@
   "name": "openzeppelin-solidity",
   "description": "Secure Smart Contract library for Solidity",
   "version": "4.9.2",
+  "private": true,
   "files": [
     "/contracts/**/*.sol",
-    "/build/contracts/*.json",
```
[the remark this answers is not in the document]
Amxx: actually, the entier "files" section of this package.json is probably useless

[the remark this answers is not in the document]
frangio: My reasoning is that there may be some workflow that relies on `npm pack` at the root, I don't know.

PR #4553, opened 2023-08-31
Optimized gas costs in `ceilDiv`

carter-ya:
Fixes #4548  

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changeset entry (run `npx changeset add`)

PR #4630, opened 2023-09-27
Optimize ceilDiv() unchecked math

k06a:
Code is equivalent in terms of math overflow safety

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -110,8 +110,10 @@ library Math {
             return a / b;
         }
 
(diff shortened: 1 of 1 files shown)


Amxx (approved):
LGTM for 5.1

k06a:
@Amxx what about `4.9.4`? BTW, it's 176 gas cheaper on latest compiler with a default settings.

Amxx:
We are not planing any `4.9.4`.

Amxx:
This was actually fixed in #4553

Issue #3736, opened 2022-09-28
EIP-1363 support


TrejGun:
**🧐 Motivation**

I was using ERC223 for some time with ERC998 but it seems to be obsolete and superseded by EIP-1363. It would be good to have the out-of-the-box implementation of a notification mechanism for receiving ERC20 tokens, the same as you have for ERC721 and ERC1155. We would like to use this standard for ERC998 tokens and for OZ VestingWallet contract.

**📝 Details**

TrejGun:
related discussion 
https://github.com/OpenZeppelin/openzeppelin-contracts/issues/3575

frangio:
There is advanced work to do this in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3017. It's been delayed due to issues around interpretation of the spec. We'll get back to it as soon as priorities allow.

Dexaran:
@ernestognw 

I did my research on [EIP-1363](https://eips.ethereum.org/EIPS/eip-1363) and I'd like to provide my feedback.

## Response to the creator of the issue #3736

>I was using ERC223 for some time with ERC998 but it seems to be obsolete and superseded by EIP-1363. 

It is not true.

[ERC-223](https://eips.ethereum.org/EIPS/eip-223) is not superseded by EIP-1363 - those are two different standards that define different approach of implementing tokens.

The main idea of ERC-223 is to create a token that will work exactly as Ether (native currency of Ethereum chain) works. There can be only one standard that behaves identical to Ether because Ether is Ether and it has only one logic.

The problem of ERC-20 standard is now evident:

- The standard contains known security vulnerabilities that remain in wontfix state
- There are three sub-types of ERC-20 tokens: (1) tokens that return `false` instead of reverting() a transaction on error during transfer like DAO token, (2) tokens that return `true` on successful transfer and revert() on error like it is described in the EIP-20 specification, (3) tokens that return nothing and revert() on error like USDT or BNB token. In fact that means that the most popular tokens (USDT and BNB) directly violate ERC-20 specification and these are not compatible with ERC-20 standard.
[...]

PR #4631, opened 2023-09-27
Add ERC1363 implementation

vittominacori:
This PR adds the ERC1363 implementation following the v5 guidelines.

It uses inheritdoc whenever is possible and custom errors following the ERC-6093 rationale.

Would like to replace #3525 and #3017 as they are outdated.

Fixes: #3736 
Fixes LIB-1181

#### PR Checklist

- [x] Tests
- [x] Documentation


zicoz18:
NOT DIRECTLY RELATED TO THE PR:

I have been looking at some standards such as ERC777 and ERC1363. ERC777 used to be included in OpenZeppelin contract and it got removed. I was wondering why such standards that extend ERC20 to be able to add a call to transfers or approvals is not highly  adopted. Since they are backwards compatible, why not use them on new tokens? I am considering to use them but want to learn the reason why others are not using it. Maybe related to complexity, extra work for integrations, security reasons (possible reentrancy in on `tokensReceived` for ERC777 and `onTransferReceived` for ERC1363) or maybe UX reasons? Is there any sources that discuss the trade-offs?

Amxx:
> I was wondering why such standards that extend ERC20 to be able to add a call to transfers or approvals is not highly adopted

For ERC777:
Partly because its a pain to work with (smart contracts can't receive them without first registring in the 1822 registry) and because they are notoriously dangerous. ERC777 opens a lot of reentrancy paths that are dangerous if not properly considered.

Amxx:
> Is there any sources that discuss the trade-offs?

This one obviously comes to mind: #2620
There are probably many others

zicoz18:
> This one obviously comes to mind: #2620 There are probably many others

Thanks a lot for the answers 🙏

ernestognw:
Thanks for putting this together @vittominacori, we appreciate the efforts of keeping an ERC1363 PR up to date. We didn't include this in 5.0 because we couldn't get consensus among maintainers and we can introduce this at some point in 5.x, so we didn't rush it.

I provided some extra thoughts in #3736, let's continue the discussion there.

TrejGun:
@vittominacori thanks for the hard work

your implementation of ERC1363Holder is different from this one
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3017/files#diff-c58adac93b31ae1df7b3146f591519a7aea8027210de859ff7fe85fa5b81ad3b

can you please explain this additional checks and why they are not included

vittominacori:
> ERC1363Holder

The file you linked is not an ERC1363Holder but it is a mock file for testing purpose.
I also added [ERC1363ReceiverMock](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4631/files#diff-c58adac93b31ae1df7b3146f591519a7aea8027210de859ff7fe85fa5b81ad3b) and [ERC1363SpenderMock](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4631/files#diff-d84127bb7f913a92ccf54affa6994f0ca6117229ed9816f17fd6835d67294997).

In addition here it is the [ERC1363Holder](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4631/files#diff-366ef0ae600665ad0c25360da5acffa3a68e0ca79ab4c1c33ac1d334ae73b17e) that is a utils file that will allow a contract to receive ERC1363 token transfers or approval. It has been added for sample purpose as the ERC721Holder or ERC1155Holder.

TrejGun:
ok I got it is just to test different types of errors that could be potentially produced by the contract

test/utils/introspection/SupportsInterface.behavior.js
```diff
@@ -30,6 +30,16 @@ const INTERFACES = {
...
+  ERC1363: [
+    'transferAndCall(address,uint256)',
+    'transferAndCall(address,uint256,bytes)',
+    'transferFromAndCall(address,address,uint256)',
+    'transferFromAndCall(address,address,uint256,bytes)',
+    'approveAndCall(address,uint256)',
+    'approveAndCall(address,uint256,bytes)',
+  ],
+  ERC1363Receiver: ['onTransferReceived(address,address,uint256,bytes)'],
+  ERC1363Spender: ['onApprovalReceived(address,uint256,bytes)'],
```
Amxx, line 42:
```suggestion
```
These interfaces should not appear here. This helper is only for ERC165 checks, which is not part of the receiver and spender interfaces

test/token/ERC20/utils/ERC1363Holder.test.js
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(async function () {
+    this.token = await ERC1363.new(name, symbol);
+    this.receiver = await ERC1363Holder.new();
+
+    await this.token.$_mint(owner, balance);
+  });
+
+  describe('receives ERC1363 token transfers', function () {
+    it('via transferAndCall', async function () {
+      await this.token.methods['transferAndCall(address,uint256)'](this.receiver.address, balance, { from: owner });
```
Amxx, line 23: for completeness, we should check that the event is emitted

test/utils/introspection/SupportsInterface.behavior.js
vittominacori: Right. Removed.

test/token/ERC20/utils/ERC1363Holder.test.js
vittominacori: Added test for events

Amxx (commented):
We (I) wanted to merge ERC1363 into OZ contract for a while. This did not happen so far because of an ambiguity in the interpretation of the final ERC. We now think we found a solution that would satisfy everyone, and would want to implement it in that PR, targeting the next minor release.

About the ambiguity: different reader of the ERC have different interpretation of what should happen when calling a `xxxAndCall` function targeting an EOA. I personally think it would be preferable that the function does not revert. However, because the ambiguity exist, and because there are deployed instances that revert when targeting an EOA, callers will not be able to make any assumption about the bahavior of the contract, and will have to assume that the call can revert.

Therefore, we decided to implement ERC1363 which function that revert when calling EOAs, and add helper functions in `SafeERC20` that provide a safe way to call `xxxAndCall` if the target has code, and revert to the basic ERC20 function otherwise.

---

Regarding the PR itself:

- Custom errors for ERC1363 are not (yet) standardized in an ERC. For consistency sake, they should be declared in the ERC1363 contract itself (like its done for other extensions such as ERC20Permit)
- In an effort to help with the ongoing migration of our tests from truffle + web3.js to ethers, all new tests should be written using the new syntax. I took the liberty of doing that change.
- Since ERC1363 behavior in only implemented by ERC1363 (and not by other contracts), I don't think we need a `.behavior.js` file here.

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,151 @@
...
+     */
+    function approveAndCall(address spender, uint256 value, bytes memory data) public virtual returns (bool) {
+        approve(spender, value);
+        _checkOnApprovalReceived(spender, value, data);
+        return true;
+    }
+
+    /**
+     * @dev Private function to invoke `onTransferReceived` on a target address.
+     * This will revert if the target doesn't implement the `IERC1363Receiver` interface or
```
ernestognw, line 91:
```suggestion
     * This will revert if the target doesn't implement the {IERC1363Receiver} interface or
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,151 @@
...
+     * @inheritdoc IERC1363
+     */
+    function approveAndCall(address spender, uint256 value, bytes memory data) public virtual returns (bool) {
+        approve(spender, value);
+        _checkOnApprovalReceived(spender, value, data);
+        return true;
+    }
+
+    /**
+     * @dev Private function to invoke `onTransferReceived` on a target address.
```
ernestognw, line 90:
```suggestion
     * @dev Performs a call to {IERC1363-onTransferReceived} on a target address.
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,151 @@
...
+     * This will revert if the target doesn't implement the `IERC1363Receiver` interface or
+     * if the target doesn't accept the token transfer or
+     * if the target address is not a contract.
+     *
+     * @param from Address representing the previous owner of the given token amount.
+     * @param to Target address that will receive the tokens.
+     * @param value The amount of tokens to be transferred.
+     * @param data Optional data to send along with the call.
```
ernestognw, line 98: These don't show up in the docs and regardless, I don't think they're needed because the function is private.

I'd just remove them

```suggestion
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * @dev Private function to invoke `onApprovalReceived` on a target address.
+     * This will revert if the target doesn't implement the `IERC1363Spender` interface or
+     * if the target doesn't accept the token approval or
+     * if the target address is not a contract.
+     *
+     * @param spender The address which will spend the funds.
+     * @param value The amount of tokens to be spent.
+     * @param data Optional data to send along with the call.
```
ernestognw, line 129: Similar suggestion as in `_checkOnTransferReceived`

```suggestion
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,151 @@
...
+                /// @solidity memory-safe-assembly
+                assembly {
+                    revert(add(32, reason), mload(reason))
+                }
+            }
+        }
+    }
+
+    /**
+     * @dev Private function to invoke `onApprovalReceived` on a target address.
```
ernestognw, line 122:
```suggestion
     * @dev Private function to invoke {IERC1363-onApprovalReceived} on a target address.
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,151 @@
...
+     * This will revert if the target doesn't implement the `IERC1363Spender` interface or
```
ernestognw, line 123:
```suggestion
     * This will revert if the target doesn't implement the {IERC1363Spender} interface or
```

contracts/token/ERC20/utils/ERC1363Holder.sol
ernestognw, line 1: I see an issue with providing this helper with both ERC1363Receiver and ERC1363Spender. Not all contracts wanting to support receiving tokens may want to support approvals and vice versa.

For example, if an ERC20 implements the ERC1363Receiver interface, it will allow draining any stuck tokens on it by calling:

```solidity
address vaultAddress = ...
uint256 vaultBalance = IERC20(vaultAddress).balanceOf(vaultAddress);
address sender = address(this);

bytes memory drain = abi.encodeCall(
  IERC20.transferFrom,
  (vaultAddress, sender, vaultBalance)
);

bytes memory approveAndDrain = abi.encodeCall(
  IERC1363.approveAndCall,
  (sender, type(uint256).max, drain)
)

IERC1363(vaultAddress).transferAndCall(vaultAddress, 0, approveAndDrain)
```

Similarly, a contract if a contract wants to be a spender but it doesn't have a withdrawal mechanism, it'll be supporting the receiver as well.

Perhaps the second case is not as serious as the first one potentially is. But I'd prefer having both separated and document the risks of using them together.

ernestognw (commented):
Haven't reviewed tests

.changeset/friendly-nails-push.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`ERC1363`: add `ERC1363` implementation other than `IERC1363Errors`, `ERC1363Holder` and tests.
```
ernestognw, line 5: The tests are not relevant for the changelog imo.

```suggestion
`ERC1363`: Add implementation of the token payable standard allowing execution of contract code after transfers and approvals, along with its `ERC1363Receiver` and `ERC1363Spender` extensions. 
```

ernestognw: I'm assuming we'll provide 2 extensions instead of one

contracts/token/ERC20/utils/ERC1363Holder.sol
Amxx: I don't understand the issue. Lets discuss that on tomorrows call.

Amxx: Also, I'm not really confident with this helper.

We have similar helpers for ERC721 and ERC1155, because its needed for accepting "safeTransferFrom" ... and then its the responsability of the contract to deal with the assets received.
However @vittominacori insisted that ERC1363 `xxxAndCall` functions are NOT `safeTransfer`/`safeTransferFrom`. That is the reason for reverting on EOA. If we wanted to mimic the ERC721/ERC1155 behavior we would NOT revert on EOA.

Considering that, and knowing that most ERC20 movements are done using `transfer`/`transferFrom` ... I'm not sure why anyone would implement ERC1363Receiver/ERC1363Spender in a way that does not include any custom logic.

.changeset/friendly-nails-push.md
Amxx:
```suggestion
`ERC1363`: Add implementation of the token payable standard allowing execution of contract code after transfers and approvals.
```

IMO `ERC1363Receiver` and `ERC1363Spender` should just be the interfaces (that are alredy in the codebase) and the mock (that are not relevant here)

Amxx: > I'm assuming we'll provide 2 extensions instead of one

Why would we provide 2 extensions ?

contracts/token/ERC20/extensions/ERC1363.sol
Amxx:
```suggestion
     * @dev Performs a call to {IERC1363-onApprovalReceived} on a target address.
```

contracts/token/ERC20/utils/ERC1363Holder.sol
vittominacori: I've added that helper with both the receiver and spender interfaces, to be similar of what have been done for ERC721 or ERC1155.

If you think that could be confusing I would remove it instead of providing 2 different files.

ernestognw: > I don't understand the issue. Lets discuss that on tomorrows call.

To summarize, I see 2 issues:

- ERC1363 enables an ERC20 to make calls to any contract (event without balance using 0 `value`)
- An ERC20 inheriting ERC1363Spender or ERC1363Receiver can receive calls

I was thinking that an ERC20 could implement ERC1363Spender without ERC1363Receiver because the receiver interface enabled any user to transfer 0 value and then make the ERC20 approve the caller so that the caller can transfer tokens from the ERC20 contract. It's common for people to send tokens to the contract address.

I see my example is missing a fallback in the calling contract. Also, I'm realizing this is possible with an `approveAndCall(..., 0)` anyway. I think we should include in the docs that any ERC20 implementing ERC1363 must not have any kind of permissions (including permissions to approve its own tokens).

> Considering that, and knowing that most ERC20 movements are done using transfer/transferFrom ... I'm not sure why anyone would implement ERC1363Receiver/ERC1363Spender in a way that does not include any custom logic.

Agree

.changeset/friendly-nails-push.md
ernestognw: > Why would we provide 2 extensions ?

I thought it was better than 1 including both behaviors. Now I agree we only should include interfaces

contracts/token/ERC20/utils/ERC1363Holder.sol
Amxx: > ERC1363 enables an ERC20 to make calls to any contract (event without balance using 0 value)

Yes, but that is a very specific call, and it will revert if the return value is not the one expected. IMO that is not a security issue. If it is, then its also affecting ERC721 and ERC1155 that have similar patterns

Amxx: > An ERC20 inheriting ERC1363Spender or ERC1363Receiver can receive calls
I'm not sure what you mean here.

Amxx: > I was thinking that an ERC20 could implement ERC1363Spender without ERC1363Receiver because the receiver interface enabled any user to transfer 0 value and then make the ERC20 approve the caller so that the caller can transfer tokens from the ERC20 contract. It's common for people to send tokens to the contract address.


I don't understand

.changeset/friendly-nails-push.md
Amxx: You meant 2 "*Holder" contract?

I thing the ambiguity is that I would personnaly not call these extensions. To me `ERC721Pausable` is an extension of `ERC721` ... but `ERC721Holder` is not an extension.

ernestognw: Yes, that's why I meant.

contracts/token/ERC20/utils/ERC1363Holder.sol
ernestognw: > Yes, but that is a very specific call, and it will revert if the return value is not the one expected. IMO that is not a security issue. If it is, then its also affecting ERC721 and ERC1155 that have similar patterns

Right, I got confused thinking that `bytes memory data` was arbitrary data execution for a moment. I see how this is not an issue now.

contracts/interfaces/IERC1363Spender.sol
```diff
@@ -10,8 +10,8 @@ pragma solidity ^0.8.20;
  */
 interface IERC1363Spender {
     /**
-     * @dev Whenever an ERC1363 tokens `owner` approved this contract via `approveAndCall`
-     * to spent their tokens, this function is called.
+     * @dev Whenever an ERC-1363 token `owner` approves this contract via `approveAndCall`
```
Amxx, line 13:
```suggestion
     * @dev Whenever an ERC-1363 token `owner` approves this contract via `approveAndCall`
```

contracts/interfaces/IERC1363.sol
```diff
@@ -7,13 +7,14 @@ import {IERC20} from "./IERC20.sol";
...
- * @dev Interface of an ERC1363 compliant contract, as defined in the
- * https://eips.ethereum.org/EIPS/eip-1363[EIP].
+ * @title IERC1363
+ * @dev Interface of the ERC1363 standard as defined in the
+ * https://eips.ethereum.org/EIPS/eip-1363[EIP-1363].
  *
- * Defines a interface for ERC20 tokens that supports executing recipient
- * code after `transfer` or `transferFrom`, or spender code after `approve`.
+ * Defines an extension interface for ERC20 tokens that supports executing code on a recipient contract
+ * after `transfer` or `transferFrom`, or code on a spender contract after `approve`, in a single transaction.
```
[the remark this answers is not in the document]
Amxx: The formulation you propose could be interpreted as 

> If I call {ERC20-transfer} or {ERC20-transferFrom}, then code can be executed.

That would be incorrect, since the ERC20 functions are not modified, and don't trigger any callback. Also it doesn't say anything about where the code is executed (the important part of 1363 is that we call the receiver, we don't really care about the code that is executed in the contract)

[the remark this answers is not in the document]
Amxx: What about something like
```suggestion
 * Defines an extension interface for {ERC20} tokens which provides alternative versions of {ERC20-transfer}, {ERC20-transferFrom} and {ERC20-approve} that include callbacks. The callbacks allow the recipient of a transfer (or of an approval) to process the received tokens (or the approval) in the same transaction.
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,151 @@
...
+
+    /**
+     * @inheritdoc IERC165
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165, IERC165) returns (bool) {
+        return interfaceId == type(IERC1363).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @inheritdoc IERC1363
```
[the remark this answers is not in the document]
Amxx: Would it work if we do

```
/**
 * @inheritdoc IERC1363
 *
 * @notice Reverts if `to` is an account without code or the recipient contract does not implement {IERC1363-onTransferReceived}
 */
```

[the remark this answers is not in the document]
vittominacori: Should be `{IERC1363Receiver-onTransferReceived}`

vittominacori:
@Amxx @ernestognw is it stale again? is there something to change or to discuss?

Amxx:
We plan on putting that in 5.1

Just want to finish test migration before we merge any new feature.

Amxx (dismissed):
LGTM

Still needs second approval

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -82,6 +82,51 @@ library SafeERC20 {
...
+     * code. This can be used to implement an {ERC721}-like safe transfer that rely on {ERC1363} checks when
+     * targeting contracts.
+     */
+    function approveAndCallRelaxed(IERC1363 token, address to, uint256 value, bytes memory data) internal {
+        if (to.code.length == 0) {
+            forceApprove(token, to, value);
+        } else {
+            token.approveAndCall(to, value, data);
+        }
+    }
```
ernestognw, line 128: In theory no token supporting `approveAndCall` will require approval to be set to 0. But practically it is possible for a token behaving like that to upgrade and add the ERC1363 module. It implies that if there exists an UDST-like token supporting ERC1363, it will revert when calling a contract, but not with an EOA.

Thinking twice about this, I wonder if it makes sense to add mention this inconsistency in the docs:

```suggestion
     * @dev Perform an {ERC1363} approveAndCall, with a fallback to the simple {ERC20} approve if the target has no
     * code. This can be used to implement an {ERC721}-like safe transfer that rely on {ERC1363} checks when
     * targeting contracts.
     *
     * NOTE: If the `to` address has code (i.e. is a contract), this function doesn't deal with tokens requiring approval
     * to be set to zero first (like USDT) before updating the approval.
     */
    function approveAndCallRelaxed(IERC1363 token, address to, uint256 value, bytes memory data) internal {
        if (to.code.length == 0) {
            forceApprove(token, to, value);
        } else {
            token.approveAndCall(to, value, data);
        }
    }
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -82,6 +82,51 @@ library SafeERC20 {
...
+    /**
+     * @dev Perform an {ERC1363} transferAndCall, with a fallback to the simple {ERC20} transfer if the target has no
+     * code. This can be used to implement an {ERC721}-like safe transfer that rely on {ERC1363} checks when
+     * targeting contracts.
+     */
+    function transferAndCallRelaxed(IERC1363 token, address to, uint256 value, bytes memory data) internal {
+        if (to.code.length == 0) {
+            safeTransfer(token, to, value);
+        } else {
+            token.transferAndCall(to, value, data);
```
ernestognw, line 94: We should match the behavior of `safeTransfer`

```suggestion
            _callOptionalReturn(token, abi.encodeCall(token.transferAndCall, (to, value, data)));
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -82,6 +82,51 @@ library SafeERC20 {
...
+        IERC1363 token,
+        address from,
+        address to,
+        uint256 value,
+        bytes memory data
+    ) internal {
+        if (to.code.length == 0) {
+            safeTransferFrom(token, from, to, value);
+        } else {
+            token.transferFromAndCall(from, to, value, data);
```
ernestognw, line 113: Similar

```suggestion
            _callOptionalReturn(token, abi.encodeCall(token.transferFromAndCall, (from, to, value, data)));
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -82,6 +82,51 @@ library SafeERC20 {
...
+    /**
+     * @dev Perform an {ERC1363} approveAndCall, with a fallback to the simple {ERC20} approve if the target has no
+     * code. This can be used to implement an {ERC721}-like safe transfer that rely on {ERC1363} checks when
+     * targeting contracts.
+     */
+    function approveAndCallRelaxed(IERC1363 token, address to, uint256 value, bytes memory data) internal {
+        if (to.code.length == 0) {
+            forceApprove(token, to, value);
+        } else {
+            token.approveAndCall(to, value, data);
```
ernestognw, line 126: Similar.

```suggestion
            _callOptionalReturn(token, abi.encodeCall(token. approveAndCall, (to, value, data)));
```

ernestognw (commented):
Haven't reviewed the tests, I'll push some updates

test/token/ERC20/utils/SafeERC20.test.js
```diff
@@ -144,6 +151,102 @@ describe('SafeERC20', function () {
...
+        await this.token.$_approve(this.owner, this.other, ethers.MaxUint256);
+
+        await expect(
+          this.token
+            .connect(this.other)
+            .transferFromAndCall(this.owner, this.receiver, value, ethers.Typed.bytes(data)),
+        )
+          .to.be.revertedWithCustomError(this.token, 'ERC1363InvalidReceiver')
+          .withArgs(this.receiver);
+      });
```
ernestognw, line 205: I think we can remove this test since it's not testing SafeERC20, but testing the ERC1363 instead.

```suggestion
```

test/token/ERC20/utils/SafeERC20.test.js
```diff
@@ -144,6 +151,102 @@ describe('SafeERC20', function () {
...
+
+      it('can transferFromAndCall to an EOA using helper', async function () {
+        await this.token.$_mint(this.owner, value);
+        await this.token.$_approve(this.owner, this.mock, ethers.MaxUint256);
+
+        await expect(this.mock.$transferFromAndCallRelaxed(this.token, this.owner, this.receiver, value, data))
+          .to.emit(this.token, 'Transfer')
+          .withArgs(this.owner, this.receiver, value);
+      });
```
ernestognw, line 214: This is the happy path, and it's testing a token that returns true on every call. We need a case for a token that returns false and it should revert, and another when the token doesn't return any value and it should pass.

test/token/ERC20/utils/SafeERC20.test.js
```diff
@@ -144,6 +151,102 @@ describe('SafeERC20', function () {
...
+      it('can transferFromAndCall to an ERC1363Receiver using helper', async function () {
+        await this.token.$_mint(this.owner, value);
+        await this.token.$_approve(this.owner, this.mock, ethers.MaxUint256);
+
+        await expect(this.mock.$transferFromAndCallRelaxed(this.token, this.owner, this.erc1363Receiver, value, data))
+          .to.emit(this.token, 'Transfer')
+          .withArgs(this.owner, this.erc1363Receiver, value)
+          .to.emit(this.erc1363Receiver, 'Received')
+          .withArgs(this.mock, this.owner, value, data);
+      });
```
ernestognw, line 225: Similarly, we need to check for `false` and no return value.

test/token/ERC20/utils/SafeERC20.test.js
```diff
@@ -144,6 +151,102 @@ describe('SafeERC20', function () {
...
+
+      it('can transferAndCall to an ERC1363Receiver using helper', async function () {
+        await this.token.$_mint(this.mock, value);
+
+        await expect(this.mock.$transferAndCallRelaxed(this.token, this.erc1363Receiver, value, data))
+          .to.emit(this.token, 'Transfer')
+          .withArgs(this.mock, this.erc1363Receiver, value)
+          .to.emit(this.erc1363Receiver, 'Received')
+          .withArgs(this.mock, this.mock, value, data);
+      });
```
ernestognw, line 190: Also missing cases when the token returns false and no return value.

test/token/ERC20/utils/SafeERC20.test.js
```diff
@@ -144,6 +151,102 @@ describe('SafeERC20', function () {
...
+          .withArgs(this.mock, this.owner, value, data);
+      });
+    });
+
+    describe('approveAndCall', function () {
+      it('cannot approveAndCall to an EOA directly', async function () {
+        await expect(this.token.approveAndCall(this.receiver, value, ethers.Typed.bytes(data)))
+          .to.revertedWithCustomError(this.token, 'ERC1363InvalidSpender')
+          .withArgs(this.receiver);
+      });
```
ernestognw, line 233: We can remove this as well

```suggestion
```

test/token/ERC20/utils/SafeERC20.test.js
```diff
@@ -144,6 +151,102 @@ describe('SafeERC20', function () {
...
+          .withArgs(this.mock, this.receiver, value);
+      });
+
+      it('can approveAndCall to an ERC1363Spender using helper', async function () {
+        await expect(this.mock.$approveAndCallRelaxed(this.token, this.erc1363Spender, value, data))
+          .to.emit(this.token, 'Approval')
+          .withArgs(this.mock, this.erc1363Spender, value)
+          .to.emit(this.erc1363Spender, 'Approved')
+          .withArgs(this.mock, value, data);
+      });
```
ernestognw, line 247: These are also missing the cases for:

- False return value
- True return value
- When the token requires approval to be set to zero before a non-zero value.

ernestognw (commented):
Still pending to review ERC1363 tests.

contracts/token/ERC20/utils/SafeERC20.sol
Amxx: Well, I'm not so sure about that. ERC-20 didn't always have the boolean return, or at least some contract were deployed before it was implemented. That is a legacy we have to deal with. 
I don't think there is such a legacy for ERC-1363. If there are any contract that implement 1363 without returning a bool, I would consider them to not follow the standard and to be out of scope.

This change is not breaking anything, but its reducing the overall readability in order to support cases that should not exist, and that we should probably not support if they exist.

As to checking the return value. That is something we could do, but that the ERC doesn't really ask for, because the return of the function is described as 
```
   * @return true unless throwing
```

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * @notice Reverts if `to` is an account without code or the recipient contract does not implement
+     * {IERC1363Receiver-onTransferReceived}
+     */
+    function transferAndCall(address to, uint256 value) public virtual returns (bool) {
```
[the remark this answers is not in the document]
Amxx: fair.

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,162 @@
...
+        approve(spender, value);
+        _checkOnApprovalReceived(spender, value, data);
+        return true;
+    }
+
+    /**
+     * @dev Performs a call to {IERC1363Receiver-onTransferReceived} on a target address.
+     * This will revert if the target doesn't implement the {IERC1363Receiver} interface or
+     * if the target doesn't accept the token transfer or
+     * if the target address is not a contract.
```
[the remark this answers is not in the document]
Amxx: It should be made clear that its either one of there, and not all of these. Usually when someone writes 

> Requirement:
> - [a]
> - [b]
> - [c]

That means that [a], [b] AND [c] must be true. Not that [a], [b] OR [c] must be true

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,162 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * @dev Performs a call to {IERC1363Spender-onApprovalReceived} on a target address.
+     * This will revert if the target doesn't implement the {IERC1363Spender} interface or
+     * if the target doesn't accept the token approval or
+     * if the target address is not a contract.
```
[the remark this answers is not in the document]
Amxx: same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4631#discussion_r1458665909

test/token/ERC20/utils/SafeERC20.test.js
Amxx: We don't have ERC-1363 mocks that return false or that return no data when called. Both would not match the ERC requirements. I'm not sure we have to specifically write non-standard contracts, that are not known to exist already, to test against them.

The idea of this library was to help interract with standard and common tokens. 

For ERC-20:
- "standard" includes tokens that return false instead of reverting (that is allowed)
- "common" includes tokens without return data (that is not in the ERC, but some important contract do that ...)

For ERC-1363:
- "standard" → the ERC explicitelly says that the token should return true or revert. returning false, or not returning anything, is not ok
- "common" → I'm not sure there is any widelly used ERC-1363 that returns false, or returns no data.

So again, I don't think we should support (and test) interractions with broken contracts.

contracts/token/ERC20/extensions/ERC1363.sol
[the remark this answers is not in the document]
ernestognw: In this case it's `success = [a], [b] AND [c]`. The previous formulation was the inverse; `fail = ![a], ![b] OR ![c]`.

I think this is good right now.

contracts/token/ERC20/utils/SafeERC20.sol
ernestognw: Thanks for pointing out the ERC spec itself! I assumed it was just returning a boolean indicating the result.

You're right, and there shouldn't be legacy ERC1363 that justify this change. I also think those tokens would be non-standard and should be considered out of scope. However, the standard doesn't specify the return value meaning for `approveAndCall`.

I think we should check for the return value in `approveAndCall` given the ambiguity, what do you think?

contracts/token/ERC20/extensions/ERC1363.sol
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * @dev Moves a `value` amount of tokens from the caller's account to `to`
+     * and then calls {IERC1363Receiver-onTransferReceived} on `to`.
+     *
+     * NOTE: Reverts if `to` is an account without code or the recipient contract does not implement
+     * {IERC1363Receiver-onTransferReceived}
+     */
+    function transferAndCall(address to, uint256 value, bytes memory data) public virtual returns (bool) {
+        transfer(to, value);
```
ernestognw, line 57: A potential issue with using the ERC20 `transfer` function like this is that someone can override `transfer` and return `false`. It'd be within the ERC20 spec but it'll allow calls to arbitrary `to`. An ERC1363Receiver shouldn't blindly trusts calls from tokens since they may be triggered unconditionally. 

I'm thinking we shoud:

- Wrap the `transfer` function in an internal `function _callTransfer(address to, uint256 value) private returns (bool)` so that users can hook to this return value using `super` and decide what to do with it (see next point).
- Add a warning in the `*AndCall` functions indicating that the function doesn't not check the return value of the underlying `approve`/`transfer` and users MUST override `_callTransfer` to adapt for ERC20's returning false.

test/token/ERC20/utils/SafeERC20.test.js
ernestognw: Yes, I agree with your stand. But in my opinion the discussion point is around what "common tokens" mean.

Since the ERC20 standard includes tokens returning `false` instead of reverting, it's possible (and very likely imo) that an ERC1363 will end up behaving as in the ERC20 standard by overriding the `transfer` function to return false, either in the first deployment or an upgrade.

See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4631#discussion_r1459521680

contracts/token/ERC20/utils/SafeERC20.sol
Amxx: Would you be ok with the check being "high level"?

```solidity
    function transferAndCallRelaxed(IERC1363 token, address to, uint256 value, bytes memory data) internal {
        if (to.code.length == 0) {
            safeTransfer(token, to, value);
        } else {
            if (!token.transferAndCall(to, value, data)) {
                revert SafeERC20FailedOperation(token);
            }
        }
    }
```

contracts/token/ERC20/extensions/ERC1363.sol
Amxx: I would really avoid adding more internal / private functions. There are already a lot. 

We could do 
```solidity
    function transferAndCall(address to, uint256 value, bytes memory data) public virtual returns (bool) {
        if (!transfer(to, value)) {
            revert ERC1363TransferFailed(to, value);
        }
        _checkOnTransferReceived(_msgSender(), to, value, data);
        return true;
    }
    ```

Comment (ernestognw) [reply contracts/token/ERC20/extensions/ERC1363.sol]: Right, this makes sense to avoid wrapping `transfer` in an internal `_callTransfer`. I'm ok with it.

File: contracts/token/ERC20/utils/SafeERC20.sol (hunk 14 above)
Comment (Amxx) [reply contracts/token/ERC20/utils/SafeERC20.sol]: Note that `_callOptionalReturn` will pass if the call returns no data. IMO it should not.

Comment (ernestognw) [reply contracts/token/ERC20/utils/SafeERC20.sol]: Yeah, I agree since it's completely non-standard to not return a value.
It works for me.

**Review by ernestognw — DISMISSED** [review]
LGTM 🚀

**Review by ernestognw — DISMISSED** [review]

**Comment by vittominacori** [comment]
I missed why it is saying that `target MUST implement the interface` but only that `target SHOULD return interface id`.

It reverts if target returns something different than the interface id.

(also `transferFromAndCall` and `approveAndCall`)

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e86bb45477d335294086c333047626b7fd317e5a/contracts/token/ERC20/extensions/ERC1363.sol#L57-L66

**Comment by Amxx** [comment]
Actually, the target should not return the interface id, it should return the function selector.

Starting a new PR to fix that 🤦🏻

**Comment by vittominacori** [comment]
Yes, but why it says should instead of must? Is there any option that the target cannot return the selector and the function does not revert?

File: contracts/token/ERC20/extensions/ERC1363.sol (hunk 23 above)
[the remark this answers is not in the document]
Comment (vittominacori) [reply contracts/token/ERC20/extensions/ERC1363.sol]: What if someone wants to override that method with a default custom `data` instead of the empty data value?

**Comment by vittominacori** [comment]
@Amxx @ernestognw I'm not sure about checking ERC20 methods return values.

The ERC1363 does not explicitly say that the ERC20 methods MUST succeed.

It says:
> Defines a token interface for ERC-20 tokens that supports executing recipient code after `transfer` or `transferFrom`, or spender code after `approve`.

Below it says:
> #notice Transfer tokens from `msg.sender` to another address and then call `onTransferReceived` on receiver

Maybe logically ok to force checking that the ERC20 methods returned `true` as it says "Transfer tokens..." but IMO it is not required.

---

Also regarding ERC1363 returning false, the spec says that methods:
> #return true unless throwing

So it is unnecessary to test ERC1363 against returning false as it is out of the standard.

**Comment by ernestognw** [comment]
> @Amxx @ernestognw I'm not sure about checking ERC20 methods return values.
The ERC1363 does not explicitly say that the ERC20 methods MUST succeed.

Yeah but also it doesn't mention they have to be treated as a noop. We concluded that it's more dangerous to allow a noop because an attacker can force a transfer failure and still call the target. This is misleading and may compromise the assumption that the caller is trusted (or always is called after a successful transfer).

I also believed it shouldn't be required, but we strive for the most secure implementation out of the box, and I do think this is it.

> Also regarding ERC1363 returning false, the spec says that methods:
>
>> #return true unless throwing
>
> So it is unnecessary to test ERC1363 against returning false as it is out of the standard.

Yeah. However, that's not the case for `approveAndCall`.
I agree we shouldn't guarantee compatibility with non-standard ERC1363, but this is also a design decision based on the likelihood of making a mistake (by returning false) and the inconsistency between the return param definition.

I think I'd be open to changing this. It's unclear to me if there's a potential issue arising, but I think reverting is the safest. Compliant tokens will work perfectly using the `SafeERC20` functions

File: contracts/mocks/token/ERC1363ReceiverMock.sol [hunk 27]
```diff
@@ -24,9 +24,9 @@ contract ERC1363ReceiverMock is IERC1363Receiver {
         _error = RevertType.None;
     }
 
-    function setUp(bytes4 retval, RevertType err) public {
+    function setUp(bytes4 retval, RevertType error) public {
         _retval = retval;
-        _error = err;
+        _error = error;
```
ernestognw, line 29: @Amxx I changed this because `error` is a reserved word. Not a big deal but I think I prefer `err`

Amxx: If we want to rename this, then we shoudl also rename the other places (ERC1363SpenderMock for example) for consistency.

ernestognw: Ah, I didn't know it was on other mocks.

I changed it initially fearing that something (e.g. transpiler) may break due to the reserved keyword.

Solving

Morhef (commented):
Morhef

PR #4632, opened 2023-09-27
Improve `AccessManaged` and `AuthorityUtils` tests

ernestognw:
Reaches 100% coverage in `AccessManaged` and `AuthorityUtils` tests

#### PR Checklist

- [x] Tests
- [ ] Documentation


test/access/manager/AccessManaged.test.js
```diff
@@ -0,0 +1,133 @@
...
+    });
+
+    describe('when role is granted with execution delay', function () {
+      beforeEach(async function () {
+        const executionDelay = web3.utils.toBN(911);
+        await this.authority.$_grantRole(this.role, roleMember, 0, executionDelay);
+      });
+
+      it('reverts if the operation is not scheduled', async function () {
+        const calldata = await this.managed.contract.methods[method]().encodeABI();
```
Amxx, line 61: This should not require an await, does it ?
```suggestion
        const calldata = this.managed.contract.methods[method]().encodeABI();
```

test/access/manager/AuthorityUtils.test.js
```diff
@@ -0,0 +1,91 @@
...
+        const { immediate, delay } = await this.mock.$canCallWithDelay(
+          this.authority.address,
+          user,
+          other,
+          '0x12345678',
+        );
+        expect(immediate).to.equal(this.immediate);
+        expect(delay).to.be.bignumber.equal(this.delay);
+      });
+    });
```
Amxx, line 72: This is actually not something we expect to see. Its ok to test it, but then we should also test `false`, `delay`

Maybe we should have two for loops:
```
for (const immediate of [ true, false ]) {
  for (const delay of [ 0, 42 ].map(web3.utils.toBN)) {
    it(`returns (immediate=${immediate}, delay=${delay})`, async function () {
      const authority = await AuthorityDelayMock.new();
      await authority._setImmediate(this.immediate);
      await authority._setDelay(delay);
      const result = await this.mock.$canCallWithDelay(
          this.authority.address,
          user,
          other,
          '0x12345678',
        );
        expect(result.immediate).to.equal(immediate);
        expect(result.delay).to.be.bignumber.equal(delay);
    }
  }
}
```

Amxx (dismissed):
LGTM

2 minor non-blocking comment.

test/access/manager/AccessManaged.test.js
Amxx: there are other calls to `encodeABI()` that this may apply to.

frangio (approved):

PR #4639, opened 2023-09-29
Update remappings.txt for upgradeable contracts and set up submodule

frangio:
Following #4628, this PR sets up the vanilla contracts in a submodule and updates remappings.txt accordingly.

In order for Hardhat to respect remappings.txt, we need the `hardhat-foundry` plugin to be installed, which needs the `forge` executable in scope.

Find the result for this PR here: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/tree/partial-submodules

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -15,3 +15,7 @@ runs:
       run: npm ci
       shell: bash
       if: steps.cache.outputs.cache-hit != 'true'
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: nightly
--- a/.github/workflows/upgradeable.yml
+++ b/.github/workflows/upgradeable.yml
@@ -18,10 +18,12 @@ jobs:
           token: ${{ secrets.GH_TOKEN_UPGRADEABLE }}
       - name: Fetch current non-upgradeable branch
         run: |
-          git fetch "https://github.com/${{ github.repository }}.git" "$REF"
+          git fetch "$REMOTE" master # Fetch default branch first for patch to apply cleanly
+          git fetch "$REMOTE" "$REF"
           git checkout FETCH_HEAD
         env:
           REF: ${{ github.ref }}
+          REMOTE: https://github.com/${{ github.repository }}.git
       - name: Set up environment
         uses: ./.github/actions/setup
       - run: bash scripts/git-user-config.sh
--- a/.github/workflows/upgradeable.yml
+++ b/.github/workflows/upgradeable.yml
@@ -29,4 +29,6 @@ jobs:
       - run: bash scripts/git-user-config.sh
       - name: Transpile to upgradeable
         run: bash scripts/upgradeable/transpile-onto.sh ${{ github.ref_name }} origin/${{ github.ref_name }}
+        env:
+          REMOTE: https://github.com/${{ github.repository }}.git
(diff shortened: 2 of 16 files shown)


.github/workflows/upgradeable.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - release-v*
+      - partial-submodules
```
frangio, line 8: Should be removed before merging.

.github/workflows/upgradeable.yml
```diff
@@ -29,4 +30,6 @@ jobs:
       - run: bash scripts/git-user-config.sh
       - name: Transpile to upgradeable
         run: bash scripts/upgradeable/transpile-onto.sh ${{ github.ref_name }} origin/${{ github.ref_name }}
+        env:
+          REMOTE: https://github.com/${{ github.repository }}.git
```
Amxx, line 34: can you explain the need for this remote ? Is it in case the transpile job is running from another repo (the upgradeable repo maybe) ?

Amxx, line 34: I'm not sure REMOVE it the best name. It should reflect that this is a dependency/submodule/peer project

Amxx (dismissed):
Naming of the REMOTE envvar is a bit non-intuitive, but otherwize looks good.

Note: I added an argv params to force enable/disable the foundry package.

.github/actions/setup/action.yml
```diff
@@ -15,3 +15,7 @@ runs:
       run: npm ci
       shell: bash
       if: steps.cache.outputs.cache-hit != 'true'
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: nightly
```
ernestognw, line 21: Why is nightly needed?

hardhat.config.js
```diff
@@ -114,3 +121,7 @@ if (argv.coverage) {
   require('solidity-coverage');
   module.exports.networks.hardhat.initialBaseFeePerGas = 0;
 }
+
+function hasFoundry() {
```
ernestognw, line 125: This will fail if Foundry is not installed. Why not just adding `true` as default in `argv`?

.github/workflows/upgradeable.yml
ernestognw: I agree it's not clear in this `.yml` because it lacks of context. `SUBMODULE_REMOTE` should be better imo

hardhat.config.js
Amxx: Does it ? My understanding is it should return false. 
(its frangio's code)

Amxx: If I do 
```
require('child_process').spawnSync('eahzoth', { stdio: 'ignore' }).error === undefined
```
I get a clean `false`.

ernestognw: You're right, I thought for a second it'll fail after returning false because Foundry was required for compiling, but it's not the case.

> (its frangio's code)

very on point! haha

.github/actions/setup/action.yml
frangio: I don't know, I copied it from the foundry tests job.

.github/workflows/upgradeable.yml
frangio: Renamed.

ernestognw (approved):
LGTM

PR #4640, opened 2023-09-29
Fixed  compile, coverage, doc, docs:watch script

koyuncuoglum95:
This is my first time contributing to the OpenZeppelin. I fixed first four scripts such as compile, coverage, docs: docs:watch. As a result, they are running successfully in terminal. Here are the result and proof of the work. 

Compile Script Result
compilation

Compile Script Result2
compilation2

Coverage Result 1
runCoverage

Coverage  Result 2
runCoverage2

Coverage Result 3
coverageResult

docs Result
docs

Doc:watch
docs2


package.json
```diff
@@ -9,10 +9,10 @@
   ],
   "scripts": {
     "compile": "hardhat compile",
-    "coverage": "env COVERAGE=true hardhat coverage",
+    "coverage": "cross-env COVERAGE=true hardhat coverage",
```
Amxx, line 12: is env not working correctly ? I would not want to add a build dependency unless its absolutely necessary

package.json
```diff
@@ -9,10 +9,10 @@
...
     "docs": "npm run prepare-docs && oz-docs",
     "docs:watch": "oz-docs watch contracts docs/templates docs/config.js",
-    "prepare-docs": "scripts/prepare-docs.sh",
+    "prepare-docs": "bash scripts/prepare-docs.sh",
```
Amxx, line 15: This script starts with `#!/usr/bin/env bash`. Is it necessary to force bash here? I'm wondering if that may introduce issues on exotic systems ... such as windows machines

Amxx (commented):
Thank you @koyuncuoglum95 for this submission.

@frangio, @ernestognw  WDYT ?

package.json
frangio: This entire PR is about making the scripts work on Windows.

frangio:
Please fix the merge conflicts.

package.json
koyuncuoglum95: When I run this script, it was giving me an error and didn't work on my terminal. So, I added cross-env instead of env and it worked.

package.json
koyuncuoglum95: @Amxx Hello my friend, As @frangio  said, it is necessary to use bash on Windows OS.

koyuncuoglum95:
> Please fix the merge conflicts.

I will fix it next time and this was my first time contributing here.

koyuncuoglum95:
> If we'll support Windows as a development environment, I'd rather support all workflows and not just the docs and coverage. In any case, the CI runs these processes so anyone pushing a PR can look at the coverage and docs preview anyway.

Hi @ernestognw, thanks for your feedback and I will be more careful next time.

ernestognw:
No issues @koyuncuoglum95, we appreciate the PR, and I think it's fine if you keep these adjustments in your local environment so you can work on the repository. We're wary of accepting changes to the dependencies given its possible security implications and also because the attack surface of supporting Windows increases.

This is why the CI does anything you may need 😄

frangio:
> This is why the CI does anything you'd need for contributing

But it should be possible to run the scripts in one's local environment.

I think Windows support is a good thing and this PR is in a good direction. But we'd need the merge conflicts to be resolved before we can merge it.

PR #4642, opened 2023-09-29
Improve `GovernorTimelockAccess` tests

ernestognw:
Reaches 100% coverage in GovernorTimelockAccess

#### PR Checklist

- [x] Tests
- [ ] Documentation


test/governance/extensions/GovernorTimelockAccess.test.js
```diff
@@ -112,6 +114,81 @@ contract('GovernorTimelockAccess', function (accounts) {
...
+          'GovernorOnlyExecutor',
+          [voter1],
+        );
+
+        await impersonate(this.mock.address);
+        const { receipt } = await this.mock.setBaseDelaySeconds(baseDelay, { from: this.mock.address });
+        expectEvent(receipt, 'BaseDelaySet', {
+          oldBaseDelaySeconds: '0',
+          newBaseDelaySeconds: baseDelay,
+        });
```
Amxx, line 132: IMO it would be better to test that using an actual governance call then using an impersonate.

In prod, you won't be able to impersonate, so we should confirm that governance calls are able to do it. Also, there is way more moving pieces in a governance call (like potentially setting the governanceCall queue) that we want to make sure don't get in the way.

Amxx: This comment applies to basically everywhere "impersonate" is used.

test/governance/extensions/GovernorTimelockAccess.test.js
```diff
@@ -112,6 +114,81 @@ contract('GovernorTimelockAccess', function (accounts) {
...
+
+        expect(await this.mock.baseDelaySeconds()).to.be.bignumber.eq(baseDelay);
+      });
+
+      it('sets access manager ignored', async function () {
+        const target = Wallet.generate().getChecksumAddressString();
```
Amxx, line 138: Why do we need to generate a wallet. IMO we could just use one of the accounts. 

On line 25 we can do 
```
const [admin, voter1, voter2, voter3, voter4, target] = accounts;
```

test/governance/extensions/GovernorTimelockAccess.test.js
```diff
@@ -141,6 +223,107 @@ contract('GovernorTimelockAccess', function (accounts) {
...
+        expect(indirect).to.deep.eq([true]);
+        expect(withDelay).to.deep.eq([true]);
+
+        await this.helper.waitForSnapshot();
+        await this.helper.vote({ support: Enums.VoteType.For }, { from: voter1 });
+        await this.helper.waitForDeadline();
+        await this.helper.queue();
+        await this.manager.grantRole(roleId, this.mock.address, 0, { from: admin }); // Purposefully remove delay so the returned nonce differs
+        await this.helper.waitForEta();
+        await expectRevertCustomError(this.helper.execute(), 'GovernorMismatchedNonce', [this.proposal.id, 1, 0]);
```
Amxx, line 272: Is that part still relevant?

AFAIK, if:
- you don't have a delay
- operation was scheduled when you had a delay
- you execute

then `execute` does
```
if (setback != 0 || getSchedule(operationId) != 0) {
    nonce = _consumeScheduledOp(operationId);
}
```

therefore, the nonce is correctly returned.

Am I missing something ?

test/governance/extensions/GovernorTimelockAccess.test.js
```diff
@@ -141,6 +223,107 @@ contract('GovernorTimelockAccess', function (accounts) {
...
+        await this.helper.waitForEta();
+
+        // Set original proposal back for executing
+        this.proposal = await this.helper.setProposal(
+          [this.restricted.operation, this.unrestricted.operation],
+          'descr',
+        );
+
+        // Attempt to execute
+        await expectRevertCustomError(this.helper.execute(), 'GovernorMismatchedNonce', [this.proposal.id, 1, 2]);
```
Amxx, line 324:
```suggestion
        // Set proposals
        const proposal1 = new GovernorHelper(this.mock, mode);
        await helper1.setProposal([this.restricted.operation, this.unrestricted.operation], 'descr');

        const proposal2 = new GovernorHelper(this.mock, mode);
        await this.helper.setProposal([this.restricted.operation], 'descr');

        // Go through all the governance process
        await proposal1.propose();
        expect(await this.mock.proposalNeedsQueuing(proposal1.currentProposal.id)).to.be.eq(true);
        const { delay: planDelay, indirect, withDelay } = await this.mock.proposalExecutionPlan(proposal1.currentProposal.id);
        expect(planDelay).to.be.bignumber.eq(web3.utils.toBN(delay));
        expect(indirect).to.deep.eq([true, false]);
        expect(withDelay).to.deep.eq([true, false]);
        await proposal1.waitForSnapshot();
        await proposal1.vote({ support: Enums.VoteType.For }, { from: voter1 });
        await proposal1.waitForDeadline();
        await proposal1.queue();
        await proposal1.waitForEta();

        // Suddenly cancel one of the proposed operations in the manager
        await this.manager.cancel(this.mock.address, this.restricted.operation.target, this.restricted.operation.data, {
          from: admin,
        });

        // Reschedule the same operation in a different proposal to avoid "AccessManagerNotScheduled" error
        await proposal2.propose();
        await proposal2.waitForSnapshot();
        await proposal2.vote({ support: Enums.VoteType.For }, { from: voter1 });
        await proposal2.waitForDeadline();
        await proposal2.queue(); // This will schedule it again in the manager
        await proposal2.waitForEta();

        // Attempt to execute
        await expectRevertCustomError(proposal1.execute(), 'GovernorMismatchedNonce', [proposal1.currentProposal.id, 1, 2]);
```

test/governance/extensions/GovernorTimelockAccess.test.js
```diff
@@ -259,7 +460,112 @@ contract('GovernorTimelockAccess', function (accounts) {
...
+          const eta = await this.mock.proposalEta(this.proposal.id);
+          const txCancel = await this.helper.cancel('internal');
+          expectEvent(txCancel, 'ProposalCanceled', { proposalId: this.proposal.id });
+
+          await time.increase(eta); // waitForEta()
+          await expectRevertCustomError(this.helper.execute(), 'GovernorUnexpectedProposalState', [
+            this.proposal.id,
+            Enums.ProposalState.Canceled,
+            proposalStatesToBitMap([Enums.ProposalState.Succeeded, Enums.ProposalState.Queued]),
+          ]);
```
Amxx, line 511: same comment, I would use 2 helper objects

Amxx (commented):
This PR adds checks for the return values of `proposalNeedsQueuing` and `proposalExecutionPlan` for all proposals

I'd argue we don't need to check it that many times. Many just 3~5 times on interresting patterns. Having this check everywhere reduced the test readability.

test/governance/extensions/GovernorTimelockAccess.test.js
ernestognw: Removed all impersonate usage

test/governance/extensions/GovernorTimelockAccess.test.js
ernestognw: Initially put it there with the intention of not mixing with the accounts, but makes sense. I named it `other` because it's not actually a target

Amxx: `other` is great

test/governance/extensions/GovernorTimelockAccess.test.js
ernestognw: The tests were old and didn't account for recent simplifications in AccessManager. I removed the test because it's no longer relevant

test/governance/extensions/GovernorTimelockAccess.test.js
```diff
@@ -112,6 +112,139 @@ contract('GovernorTimelockAccess', function (accounts) {
...
+
+        expectEvent(receipt, 'BaseDelaySet', {
+          oldBaseDelaySeconds: '0',
+          newBaseDelaySeconds: baseDelay,
+        });
+
+        expect(await this.mock.baseDelaySeconds()).to.be.bignumber.eq(baseDelay);
+      });
+
+      it.only('sets access manager ignored', async function () {
```
Amxx, line 150:
```suggestion
      it('sets access manager ignored', async function () {
```

ernestognw:
> This PR adds checks for the return values of `proposalNeedsQueuing` and `proposalExecutionPlan` for all proposals
> 
> I'd argue we don't need to check it that many times. Many just 3~5 times on interresting patterns. Having this check everywhere reduced the test readability.

I agree, I'd like to make tests por those specific cases so I'll take it if I got time.

Amxx:
`test/metatx/ERC2771Forwarder.test.js` is failing ... that is strange

frangio:
I think it's caused by timing issues, the error message is about expiring forward requests.

frangio (approved):
This is reasonably good. Let's merge as is and follow up with improvements.

PR #4643, opened 2023-10-01
Update eth-gas-reporter

frangio:
Removes a bunch of deprecated dependencies with security advisories.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
         "graphlib": "^2.1.8",
         "hardhat": "^2.9.1",
         "hardhat-exposed": "^0.3.12-1",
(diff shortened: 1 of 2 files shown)


cgewecke:
Ah this gas-compare check looks really nice...didn't realize you had this. 

is there an example of it running? Or of the report getting generated?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/scripts/checks/compareGasReports.js

frangio:
It's supposed to show up in the checks sumamry but it's apparently broken, I had not realized until now.

Here's one in a previous PR: https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/5791533605#summary-15696494605

PR #4644, opened 2023-10-02
Make AccessManager.execute/schedule more conservative when delay is 0

frangio:
This PR makes the `schedule` function revert when the caller does not have an enforced delay. Note that this doesn't prevent a situation where a caller is able to schedule a call but later has their delay removed. To make that case more consistent with the normal schedule/execute case, this PR also makes `schedule` consume an available schedule even when the caller doesn't have a currently enforced delay, and correctly return the corresponding nonce which is important for `GovernorTimelockAccess`.

Cherry picked from https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4613, where tests were adjusted.


Amxx (dismissed):
LGTM

.changeset/thirty-drinks-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': major
```
ernestognw, line 2: Not sure if it's major, isn't a patch although we're in pre-release?

ernestognw (approved):
LGTM + minor comment

.changeset/thirty-drinks-happen.md
ernestognw: Ok make sense, also make sure it'll be picked up by the changeset versioning, which I think it will

ernestognw (approved):
Thanks!

frangio:
The skip wasn't correctly written.

PR #4646, opened 2023-10-02
Fix release tagging

frangio:
We're having an issue with automated releases where the tag that gets created in the process is not correctly assigned to the release commit (even though the rest of the release is all done correctly). We believe this is because `npx changeset tag; git push --tags` is either not creating the tag or not pushing it correctly, causing the subsequent `createRelease` API call in `github-release.js` to create the tag in the default branch. This PR removes the manual tagging and relies on `createRelease` by specifying the `target_commitish` option ([docs here](https://octokit.github.io/rest.js/v18#repos-create-release)).

### DIFF
--- a/.github/workflows/release-cycle.yml
+++ b/.github/workflows/release-cycle.yml
@@ -147,16 +147,12 @@ jobs:
         with:
           name: ${{ github.ref_name }}
           path: ${{ steps.pack.outputs.tarball }}
-      - name: Tag
-        run: npx changeset tag
       - name: Publish
         run: bash scripts/release/workflow/publish.sh
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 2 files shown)


Amxx (approved):

.github/workflows/release-cycle.yml
```diff
@@ -147,16 +147,12 @@ jobs:
         with:
           name: ${{ github.ref_name }}
           path: ${{ steps.pack.outputs.tarball }}
-      - name: Tag
-        run: npx changeset tag
```
ernestognw, line 151: Isn't this creating the actual changelog?

ernestognw (commented):
Not sure if this will work, I think `npx changeset tag` write the changelog. And anyway, it looks very suspicious that we're removing the changeset command

ernestognw (approved):
The changelog is done in the Changesets PR, so no issue

PR #4648, opened 2023-10-03
Fix coverage analysis

frangio:
It appears that `solidity-coverage` and `hardhat-foundry` are incompatible.


ernestognw (approved):
I wonder if there's another command that can conflict between Foundry and Hardhat, but I guess for the purposes of coverage analysis this is fine.

PR #4649, opened 2023-10-04
Ensure constant getters show in docs

frangio:
The main example of what was missing can be seen in `UPGRADE_INTERFACE_VERSION` in `UUPSUpgradeable`.

### DIFF
--- a/docs/templates/properties.js
+++ b/docs/templates/properties.js
@@ -1,4 +1,4 @@
-const { isNodeType } = require('solidity-ast/utils');
+const { isNodeType, findAll } = require('solidity-ast/utils');
 const { slug } = require('./helpers');
 
 module.exports.anchor = function anchor({ item, contract }) {
@@ -37,6 +37,21 @@ module.exports['has-events'] = function ({ item }) {
 
 module.exports['has-errors'] = function ({ item }) {
   return item.inheritance.some(c => c.errors.length > 0);
+};
+
+module.exports.functions = function ({ item }) {
+  return [...findAll(['VariableDeclaration', 'FunctionDefinition'], item)]
+    .filter(f =>
+      f.nodeType === 'VariableDeclaration' ? f.visibility === 'public' : f.visibility !== 'private'
+    );
+};
+
+module.exports.returns2 = function ({ item }) {
+  if (isNodeType('VariableDeclaration', item)) {
+    return [{ type: item.typeDescriptions.typeString }];
+  } else {
+    return item.returns;
+  }
 };
 
 module.exports['inherited-functions'] = function ({ item }) {
--- a/docs/templates/properties.js
+++ b/docs/templates/properties.js
@@ -40,10 +40,9 @@ module.exports['has-errors'] = function ({ item }) {
 };
 
 module.exports.functions = function ({ item }) {
-  return [...findAll(['VariableDeclaration', 'FunctionDefinition'], item)]
-    .filter(f =>
-      f.nodeType === 'VariableDeclaration' ? f.visibility === 'public' : f.visibility !== 'private'
-    );
+  return [...findAll(['VariableDeclaration', 'FunctionDefinition'], item)].filter(f =>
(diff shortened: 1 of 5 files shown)


docs/templates/properties.js
```diff
@@ -39,6 +39,20 @@ module.exports['has-errors'] = function ({ item }) {
   return item.inheritance.some(c => c.errors.length > 0);
 };
 
+module.exports.functions = function ({ item }) {
+  return [...findAll(['VariableDeclaration', 'FunctionDefinition'], item)].filter(f =>
+    f.nodeType === 'VariableDeclaration' ? f.visibility === 'public' : f.visibility !== 'private',
+  );
+};
+
+module.exports.returns2 = function ({ item }) {
```
ernestognw, line 48: Why not `returnsExtended` or similar? I don't think `returns2` is really expressive

docs/templates/properties.js
```diff
@@ -39,6 +39,20 @@ module.exports['has-errors'] = function ({ item }) {
   return item.inheritance.some(c => c.errors.length > 0);
 };
 
+module.exports.functions = function ({ item }) {
+  return [...findAll(['VariableDeclaration', 'FunctionDefinition'], item)].filter(f =>
+    f.nodeType === 'VariableDeclaration' ? f.visibility === 'public' : f.visibility !== 'private',
+  );
+};
```
ernestognw, line 46: IIUC this should be updated in [solidity-docgen](https://github.com/OpenZeppelin/solidity-docgen), is it?

ernestognw (approved):

docs/templates/properties.js
frangio: Yes, but it's more complicated because of backwards compatibility. `functions` currently returns `FunctionDefinition` nodes and not `VariableDeclaration` nodes, so it might break some templates.

docs/templates/properties.js
frangio: I prefer the short version, if you feel strongly about it we could use `returnsExtended`.

frangio:
I moved getters to the end of the list of functions because they were showing up before the constructor.

docs/templates/properties.js
```diff
@@ -39,6 +39,21 @@ module.exports['has-errors'] = function ({ item }) {
   return item.inheritance.some(c => c.errors.length > 0);
 };
 
+module.exports.functions = function ({ item }) {
+  return [
+    ...[...findAll('FunctionDefinition', item)].filter(f => f.visibility !== 'private'),
+    ...[...findAll('VariableDeclaration', item)].filter(f => f.visibility === 'public'),
+  ];
```
ernestognw, line 46: I think this works:

```suggestion
  return [
    ...findAll('FunctionDefinition', item).filter(f => f.visibility !== 'private'),
    ...findAll('VariableDeclaration', item).filter(f => f.visibility === 'public'),
  ];
```

docs/templates/properties.js
ernestognw: Let's keep `returns2`, don't want to block this

docs/templates/properties.js
ernestognw: Also, we can keep the constructor getters at first

```suggestion
  const functionDefs = findAll('FunctionDefinition', item)
  const constructor = functionDefs.find(f => f.kind == 'constructor');
  return [
    constructor,
    ...findAll('VariableDeclaration', item).filter(f => f.visibility === 'public'),
    ...functionDefs.filter(f => f.visibility !== 'private')
  ];
```

For cases like AccessManager and AccessControl I think it's valuable to have the `ADMIN_ROLE`/`DEFAULT_ADMIN_ROLE` definitions at first. 

It's true that `ADMIN_ROLE` and `PUBLIC_ROLE` don't have NatSpec anyways, so I wouldn't have problems keeping the current code.

ernestognw (approved):

docs/templates/properties.js
frangio: >I think this works

It doesn't because `findAll` is a generator function, it doesn't return an array.

PR #4651, opened 2023-10-04
Add a gas optimization in `Pausable` similar to the one used in the `ReentrancyGuard`

### LINKED ISSUE #4650: `Pausable` can be more gas efficient using a similar mechanism as in the `ReentrancyGuard`
I noticed that the Pausable contract can be optimized for gas usage by adopting a similar mechanism used in the ReentrancyGuard contract.

**💻 Environment**

OpenZeppelin Contracts 4.9.3

**📝 Details**

The `ReentrancyGuard` contract uses `uint256` constants to avoid the extra gas costs associated with boolean operations. This mechanism can be applied to the `Pausable` contract to make it more gas-efficient.

**🔢 Code to reproduce bug**

eugenioclrc:
Fixes #4650

I noticed that the `Pausable` contract can be optimized for gas usage by adopting a similar mechanism used in the `ReentrancyGuard` contract.
The `ReentrancyGuard` contract uses `uint256` constants to avoid the extra gas costs associated with boolean operations. This mechanism can be applied to the `Pausable` contract to make it more gas-efficient.

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx:
@coderabbitai review

.changeset/tame-rules-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
```
arr00:
```suggestion
'openzeppelin-solidity': minor
```

contracts/utils/Pausable.sol
```diff
@@ -15,7 +15,15 @@ import {Context} from "../utils/Context.sol";
...
  */
 abstract contract Pausable is Context {
-    bool private _paused;
+    // The values being non-zero value makes deployment a bit more expensive,
+    // but in exchange the refund on every call to Pausable will be lower in
+    // amount. Since refunds are capped to a percentage of the total
+    // transaction's gas, it is best to keep them low in cases like this one, to
+    // increase the likelihood of the full refund coming into effect.
+    uint256 private constant NOT_PAUSED = 1;
+    uint256 private constant PAUSED = 2;
```
Amxx:
```suggestion
    uint256 private constant PAUSED = 1;
    uint256 private constant NOT_PAUSED = 2;
```

contracts/utils/Pausable.sol
```diff
@@ -15,7 +15,15 @@ import {Context} from "../utils/Context.sol";
...
-    bool private _paused;
+    // The values being non-zero value makes deployment a bit more expensive,
+    // but in exchange the refund on every call to Pausable will be lower in
+    // amount. Since refunds are capped to a percentage of the total
+    // transaction's gas, it is best to keep them low in cases like this one, to
+    // increase the likelihood of the full refund coming into effect.
+    uint256 private constant PAUSED = 1;
+    uint256 private constant NOT_PAUSED = 2;
+
+    uint256 private _paused;
```
Amxx, line 26:
```suggestion
    /// @custom:oz-retyped-from bool
    uint256 private _paused;
```

Amxx:
Need to check impact on the upgradeable version (storage).

Amxx:
**Discussion from slack**

- @Amxx

> The PR is retyping from a bool to a uint256 (we could use a uint8 if we wanted). That is safe because its the contracts only variable, so its not packed with anything else. Its alone in its slot. I've check memory alignment and its all fine if we do `true = 1 = PAUSED` (and `false = 0 != PAUSED`). However, we still get errors from the upgrades plugin (as expected)
> 
> ```
> Storage layout incompatibility found in PausableUpgradeable:
> contracts/utils/PausableUpgradeable.sol:30: Upgraded `_paused` to an incompatible type
>  - Bad upgrade from bool to uint256
> ```
> In these cases, my memory tells me to use `/// @custom:oz-retyped-from bool` ... but AFAIK this is for state variable. Here we have a namespace structure that is modified. The location remains the same, but the struct goes from
> ```
> /// @custom:storage-location erc7201:openzeppelin.storage.Pausable
> struct PausableStorage {
>     bool _paused;
> }
> ```
> to
> ```
> /// @custom:storage-location erc7201:openzeppelin.storage.Pausable
> struct PausableStorage {
>     uint256 _paused;
> }
> ```
> Is there an easy way to fix that error ?

- @ericglau

> We can’t use that annotation inside structs because Solidity doesn’t expose NatSpec within structs.  Upgrades plugin issue: https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/802
(A workaround could be to use a parser on the source code itself, but that is a large effort)

- @Amxx 

> Thank you for the feedback. Considering there is no way the upgrades plugin will support that retyping in time for 5.6, I'm deprioritizing the PR

Amxx (commented):
Note: _paused could be a uint8. Any reason to store it as uint256 ?

contracts/utils/Pausable.sol
```diff
@@ -15,7 +15,16 @@ import {Context} from "../utils/Context.sol";
...
-    bool private _paused;
+    // The values being non-zero value makes deployment a bit more expensive,
+    // but in exchange the refund on every call to Pausable will be lower in
+    // amount. Since refunds are capped to a percentage of the total
+    // transaction's gas, it is best to keep them low in cases like this one, to
+    // increase the likelihood of the full refund coming into effect.
+    uint256 private constant PAUSED = 1;
+    uint256 private constant NOT_PAUSED = 2;
+
+    /// @custom:oz-retyped-from bool
```
Amxx, line 26: if we make that a 6.0 (breaking) feature, we don't need that retyped-from comment.

PR #4659, opened 2023-10-05
Update "Using with Upgrades" page for 5.0

ericglau:
Updates the "Using with Upgrades" page as follows:
- Changes the Hardhat example to use ethers v6
- Describe peer dependency on `@openzeppelin/contracts` and that it is used for interfaces and libraries
  - Includes this package in the npm install instructions.  Even though it is a peer dependency and would be installed automatically by recent versions of npm, it would be better practice to install it directly so that it is included in the user's `package.json`.
- Describe namespaced storage instead of gaps


Amxx (approved):
LGTM

ernestognw (approved):

ericglau:
Added a note that inheritance order can be changed, as long as all inherited contracts use namespaced storage.

Amxx (approved):

ernestognw (approved):

PR #4674, opened 2023-10-11
Update SupportsInterface.behavior.js

mvillere:
Add Interface for ERC1155MetadataURI, so ERC1155 contracts can be fully checked (OZ ERC1155 supportsInterface() call will return true for both ERC1155 and ERC1155MetadataURI)

Fixes #???? 

Added ERC1155MetadataURI interface definition, so that tests can be run using ERC1155MetadataURI key.
My company maintains its own version of this definition here: https://github.com/NFTCulture/nftc-contracts/blob/main/src/utils/introspection/OZ_Interfaces.ts

We use use that definition in our own typescript tests. We tested our own implementation of shouldSupportInterfaces (https://github.com/NFTCulture/nftc-contracts/blob/main/src/utils/introspection/test/behaviors/SupportsInterfaces.ts) and an ERC1155 contract derived from OZ ERC1155, and the tests pass as expected.

Since we do not use the javascript library and most of the implementation, we did not write JS tests in OZ library for this specific thing.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/token/ERC1155/ERC1155.behavior.js
+++ b/test/token/ERC1155/ERC1155.behavior.js
@@ -896,7 +896,7 @@ function shouldBehaveLikeERC1155([minter, firstTokenHolder, secondTokenHolder, m
       });
     });
 
-    shouldSupportInterfaces(['ERC165', 'ERC1155']);
+    shouldSupportInterfaces(['ERC165', 'ERC1155', 'ERC1155MetadataURI']);
   });
 }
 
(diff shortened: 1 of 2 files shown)


ernestognw:
Thanks for pointing out @mvillere! I'm also adding the new interfaceId to ERC1155 behaviors.

@Amxx We probably don't want to keep using ERC165 in some cases but imo this is already there and wasn't tested.

ernestognw (approved):

PR #4728, opened 2023-11-09
Migrate governance tests to ethers.js

RenanSouza2:
A proposal for migration to governance tests

#### PR Checklist

- [ ] Tests
- [ ] Documentation


RenanSouza2:
This is the draft yet but I think I've cracked a way to migrate this one that AFAIK should be one of the hardest to do

ernestognw:
> This is the draft yet but I think I've cracked a way to migrate this one that AFAIK should be one of the hardest to do

Yeah, this is probably from the hardest. I'd suggest focusing only on Governor and then moving to the extensions. Thanks for putting this together! We'll review once it's ready

test/helpers/time.js
```diff
@@ -23,7 +24,17 @@ module.exports = {
...
+  clockFromReceipt: {
+    blocknumber: txPromisse =>
+      Promise.resolve(txPromisse)
+        .then(tx => tx.blockNumber)
+        .then(BigInt),
+    timestamp: txPromisse =>
+      Promise.resolve(txPromisse)
+        .then(tx => ethers.provider.getBlock(tx.blockNumber))
+        .then(block => block.timestamp)
+        .then(BigInt),
```
ernestognw, line 36: These aren't needed since the transaction response already contains `blockNumber`. I'm removing this and replacing where it fails

test/helpers/enums.js
```diff
@@ -2,10 +2,20 @@ function Enum(...options) {
...
-module.exports = {
-  Enum,
-  ProposalState: Enum('Pending', 'Active', 'Canceled', 'Defeated', 'Succeeded', 'Queued', 'Expired', 'Executed'),
-  VoteType: Enum('Against', 'For', 'Abstain'),
-  Rounding: Enum('Floor', 'Ceil', 'Trunc', 'Expand'),
-  OperationState: Enum('Unset', 'Waiting', 'Ready', 'Done'),
-};
+function EnumBigInt(...options) {
+  return Object.fromEntries(options.map((key, i) => [key, i]));
+}
```
ernestognw, line 7: This is actually not BigInt and just regular number. Still, I think this is the way we should implement it and we can have a single version.

I'll update.

ernestognw (commented):
This is in the right direction, thanks @RenanSouza2 🚀 

Just added a few commits but I think we can improve a couple of things before we merge. 

After going down the rabbit hole and migrated `Votes.test.js` (I wanted to avoid having two versions of `EIP.6372.behavior.js`), I realized that so far we'd used the pattern of a `module.exports.bigint` variant, but I think there are more complex cases like `Votes.behavior` or `EIP6372.behavior.js`.

My opinion on those more complex cases is that we should create two different files since keeping a version compatible with both doesn't look feasible in my opinion, and I'm convinced we can get a better result without constraining the ethers.js variants. A good example is the GovernorHelper.

Let's migrate first the `governance.new.js` before merging this PR, I'll work on that.

test/governance/Governor.test.js
```diff
@@ -1,669 +1,666 @@
...
+  );
+  return {
+    token,
+    mock,
+    helper,
+    proposal: helper.currentProposal,
+  };
+}
+
+async function fixture() {
```
ernestognw, line 83: I decided to put the fixture outside, otherwise, I see easily that we end up with multiple fixtures, losing the benefit of having just one state to load.

This way we load every variable once and reset to that snapshot.

test/helpers/governance.new.js
```diff
@@ -0,0 +1,221 @@
...
+    // proposal id
+    const id = ethers.keccak256(
+      ethers.AbiCoder.defaultAbiCoder().encode(['address[]', 'uint256[]', 'bytes[]', 'bytes32'], shortProposal),
+    );
+
+    const currentProposal = {
+      id,
+      targets,
+      values,
+      signatures: [''],
```
ernestognw, line 204: I realized we don't use this field anywhere even in the Truffle tests. I didn't remove it from the old `governance.js` file because we'll remove it but just mentioning for visibility

ernestognw (approved):
Alright, this LGTM for a first stage of the Governor tests migration, let's wait on @Amxx review

ernestognw (approved):

Amxx:
Added ERC20Votes and ERC721Votes to the PR

test/helpers/time.js
```diff
@@ -21,8 +21,8 @@ module.exports = {
 
 // TODO: deprecate the old version in favor of this one
 module.exports.bigint = {
-  clock: mapObject(module.exports.clock, fn => () => fn().then(BigInt)),
-  clockFromReceipt: mapObject(module.exports.clockFromReceipt, fn => receipt => fn(receipt).then(BigInt)),
+  clock: mapObject(module.exports.clock, fn => () => fn().then(ethers.toBigInt)),
+  clockFromReceipt: mapObject(module.exports.clockFromReceipt, fn => receipt => fn(receipt).then(ethers.toBigInt)),
   forward: module.exports.forward,
-  duration: mapObject(module.exports.duration, fn => n => BigInt(fn(n))),
+  duration: mapObject(module.exports.duration, fn => n => ethers.toBigInt(fn(ethers.toNumber(n)))),
```
ernestognw: Is there any reason for preferring `ethers.toBigInt` `BigInt`? I see `fn` is returning is the same anyways so there shouldn't be a diff.

test/governance/extensions/GovernorTimelockAccess.test.js
```diff
@@ -1,383 +1,363 @@
...
 
-  const name = 'OZ-Governor';
-  const version = '1';
-  const tokenName = 'MockToken';
-  const tokenSymbol = 'MTKN';
-  const tokenSupply = web3.utils.toWei('100');
-  const votingDelay = web3.utils.toBN(4);
-  const votingPeriod = web3.utils.toBN(16);
-  const value = web3.utils.toWei('1');
+function prepareOperation({ sender, target, value = 0n, data = '0x' }) {
```
ernestognw: The default should be at least 4 bytes long, otherwise the `data.slice(0, 10)` for the selector would get a different value, isn't it?

```suggestion
function prepareOperation({ sender, target, value = 0n, data = '0x00000000' }) {
```

ernestognw (commented):
Half of the way, I'll continue reviewing

test/governance/utils/Votes.test.js
```diff
@@ -1,13 +1,15 @@
+require('array.prototype.at/auto');
```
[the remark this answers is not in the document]
Amxx:
```suggestion
```

test/governance/utils/Votes.test.js
```diff
@@ -1,13 +1,15 @@
...
+
```
Amxx, line 2:
```suggestion
```

test/helpers/time.js
Amxx: I tink ethers's version does more check that the value provided is correct. Also it clearly states that it covers a varriety of input types. 

`ethers.toBigInt(value: BigNumberish | Uint8Array)⇒bigint`

I also think the name is more explicitely saying that a conversion is hapening.

In our case, both may work just fine.

test/governance/extensions/GovernorTimelockAccess.test.js
Amxx: I think a better option is to keep `data = '0x'` by default but then do 
```
selector: data.slice(0, 10).padEnd(10,'0'),
```

ernestognw: Right, that works!

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,181 +1,197 @@
...
+          this.recipientVotes = 0n;
         });
 
         afterEach(async function () {
-          expect(await this.votes.getVotes(account1)).to.be.bignumber.equal(this.account1Votes);
-          expect(await this.votes.getVotes(account2)).to.be.bignumber.equal(this.account2Votes);
+          expect(await this.votes.getVotes(this.holder)).to.equal(this.holderVotes);
+          expect(await this.votes.getVotes(this.recipient)).to.equal(this.recipientVotes);
 
           // need to advance 2 blocks to see the effect of a transfer on "getPastVotes"
```
ernestognw, line 190: Is this comment still true though? I only see one `mine()` and no other non-view function

test/governance/Governor.test.js
```diff
@@ -83,577 +100,517 @@ contract('Governor', function (accounts) {
...
+        expect(await this.mock.proposalNeedsQueuing(this.proposal.id)).to.be.false;
 
         // Run proposal
-        const txPropose = await this.helper.propose({ from: proposer });
-
-        expectEvent(txPropose, 'ProposalCreated', {
-          proposalId: this.proposal.id,
-          proposer,
-          targets: this.proposal.targets,
-          // values: this.proposal.values,
```
ernestognw, line 113: Any reason why this was commented out before and now it's included?

test/governance/Governor.test.js
```diff
@@ -83,577 +100,517 @@ contract('Governor', function (accounts) {
...
+            this.proposal.id,
+            this.proposer.address,
+            this.proposal.targets,
+            this.proposal.values,
+            this.proposal.signatures,
+            this.proposal.data,
+            (await time.clockFromReceipt[mode](txPropose)) + votingDelay,
+            (await time.clockFromReceipt[mode](txPropose)) + votingDelay + votingPeriod,
+            this.proposal.description,
+          );
```
ernestognw, line 138:
```suggestion
        const timepoint = await time.clockFromReceipt[mode](txPropose);

        await expect(txPropose)
          .to.emit(this.mock, 'ProposalCreated')
          .withArgs(
            this.proposal.id,
            this.proposer.address,
            this.proposal.targets,
            this.proposal.values,
            this.proposal.signatures,
            this.proposal.data,
            timepoint + votingDelay,
            timepoint + votingDelay + votingPeriod,
            this.proposal.description,
          );
```

test/helpers/governance.js
```diff
@@ -134,61 +112,56 @@ class GovernorHelper {
...
 
   /**
    * Specify a proposal either as
-   * 1) an array of objects [{ target, value, data, signature? }]
-   * 2) an object of arrays { targets: [], values: [], data: [], signatures?: [] }
+   * 1) an array of objects [{ target, value, data }]
+   * 2) an object of arrays { targets: [], values: [], data: [] }
    */
   setProposal(actions, description) {
     let targets, values, signatures, data, useCompatibilityInterface;
```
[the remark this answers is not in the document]
ernestognw: Just removed `useCompatibilityInterface` and `signatures` in f2086aa. Let's see if the tests pass

test/governance/Governor.test.js
Amxx: We used to not be able the check array of BN. With the migration to bigint we can check it.

Also, ethers+chai requires all events members to be included. we would put `anyValue`, but since we have the value its better to check it.

test/helpers/governance.js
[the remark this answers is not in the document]
Amxx: fixing that in a commit

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
           });
         });
 
         describe('getPastVotes', function () {
           it('reverts if block number >= current block', async function () {
             const clock = await this.token.clock();
-            await expectRevertCustomError(this.token.getPastVotes(other1, 5e10), 'ERC5805FutureLookup', [5e10, clock]);
+            await expect(this.token.getPastVotes(this.other1, 50000000000n))
+              .to.be.revertedWithCustomError(this.token, 'ERC5805FutureLookup')
+              .withArgs(50000000000n, clock);
```
ernestognw, line 429:
```suggestion
              .withArgs(50_000_000_000n, clock);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
 
           it('returns the latest block if >= last checkpoint block', async function () {
-            const { receipt } = await this.token.delegate(other1, { from: holder });
-            const timepoint = await clockFromReceipt[mode](receipt);
-            await time.advanceBlock();
-            await time.advanceBlock();
+            const tx = await this.token.connect(this.holder).delegate(this.other1);
+            const timepoint = await time.clockFromReceipt[mode](tx);
+            await mine();
+            await mine();
```
ernestognw, line 440: Either we're missing a comment about why we need 2 `mine()` or we don't need them. I'll check it out

ernestognw: Ok, it's needed so there's no `ERC5805FutureLookup` error.
We can do:

```suggestion
            await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
-
-            expect(await this.token.getPastVotes(other1, timepoint - 1)).to.be.bignumber.equal('0');
-            expect(await this.token.getPastVotes(other1, timepoint + 1)).to.be.bignumber.equal(
-              '10000000000000000000000000',
-            );
+            await mine();
+            const tx = await this.token.connect(this.holder).delegate(this.other1);
+            const timepoint = await time.clockFromReceipt[mode](tx);
+            await mine();
+            await mine();
```
ernestognw, line 451:
```suggestion
            await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
-            );
-            expect(await this.token.getPastVotes(other1, t4.timepoint)).to.be.bignumber.equal(
-              '10000000000000000000000000',
-            );
-            expect(await this.token.getPastVotes(other1, t4.timepoint + 1)).to.be.bignumber.equal(
-              '10000000000000000000000000',
-            );
+            const t1 = await this.token.connect(this.holder).delegate(this.other1);
+            await mine();
+            await mine();
```
ernestognw, line 460:
```suggestion
            await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
+            const t2 = await this.token.connect(this.holder).transfer(this.other2, 10);
+            await mine();
+            await mine();
```
ernestognw, line 463:
```suggestion
            await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
+            const t3 = await this.token.connect(this.holder).transfer(this.other2, 10);
+            await mine();
+            await mine();
```
ernestognw, line 466:
```suggestion
            await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
+            const t4 = await this.token.connect(this.other2).transfer(this.holder, 20);
+            await mine();
+            await mine();
```
ernestognw, line 469:
```suggestion
            await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
 
         it('returns the latest block if >= last checkpoint block', async function () {
-          const { receipt } = await this.token.$_mint(holder, supply);
-          const timepoint = await clockFromReceipt[mode](receipt);
-          await time.advanceBlock();
-          await time.advanceBlock();
+          const tx = await this.token.$_mint(this.holder, supply);
+          const timepoint = await time.clockFromReceipt[mode](tx);
+          await mine();
+          await mine();
```
ernestognw, line 509:
```suggestion
          await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
-
-          expect(await this.token.getPastTotalSupply(timepoint - 1)).to.be.bignumber.equal('0');
-          expect(await this.token.getPastTotalSupply(timepoint + 1)).to.be.bignumber.equal(
-            '10000000000000000000000000',
-          );
+          await mine();
+          const tx = await this.token.$_mint(this.holder, supply);
+          const timepoint = await time.clockFromReceipt[mode](tx);
+          await mine();
+          await mine();
```
ernestognw, line 520:
```suggestion
          await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
-          expect(await this.token.getPastTotalSupply(t3.timepoint + 1)).to.be.bignumber.equal(
-            '9999999999999999999999980',
-          );
-          expect(await this.token.getPastTotalSupply(t4.timepoint)).to.be.bignumber.equal('10000000000000000000000000');
-          expect(await this.token.getPastTotalSupply(t4.timepoint + 1)).to.be.bignumber.equal(
-            '10000000000000000000000000',
-          );
+          const t1 = await this.token.$_mint(this.holder, supply);
+          await mine();
+          await mine();
```
ernestognw, line 529:
```suggestion
          await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
+          const t2 = await this.token.$_burn(this.holder, 10n);
+          await mine();
+          await mine();
```
ernestognw, line 532:
```suggestion
          await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
+          const t3 = await this.token.$_burn(this.holder, 10n);
+          await mine();
+          await mine();
```
ernestognw, line 535:
```suggestion
          await mine(2);
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
+          const t4 = await this.token.$_mint(this.holder, 20n);
+          await mine();
+          await mine();
```
ernestognw, line 538:
```suggestion
          await mine(2);
```

test/helpers/txpool.js
```diff
@@ -1,30 +1,20 @@
...
+    const responses = await Promise.all(txs.map(fn => fn()));
     // mine one block
-    await network.provider.send('evm_mine');
+    await mine();
     // fetch receipts
-    const receipts = await Promise.all(promises);
+    const receipts = await Promise.all(responses.map(response => response.wait()));
     // Sanity check, all tx should be in the same block
-    const minedBlocks = new Set(receipts.map(({ receipt }) => receipt.blockNumber));
+    const minedBlocks = new Set(receipts.map(receipt => receipt.blockNumber));
```
ernestognw, line 15: We have a `unique` function in `scripts/helpers.js` that we only use in the governance helpers. I feel we should decide between using the `new Set()` syntax vs a `unique` function.

Also, by making a quick search I also realized we have the same `unique` function in `scripts/helpers.js`.

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,181 +1,197 @@
...
+const TOKENS = [
+  { Token: '$ERC721Votes', mode: 'blocknumber' },
   // no timestamp mode for ERC721Votes yet
-};
+];
+
+const name = 'My Vote';
+const symbol = 'MTKN';
+const version = '1';
+const tokens = [10000000000000000000000000n, 10n, 20n, 30n];
```
ernestognw, line 17: This number is just too long

```suggestion
const tokens = [ethers.parseEther('10000000'), 10n, 20n, 30n];
```

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,181 +1,197 @@
...
 
-          const { receipt } = await this.votes.transferFrom(account1, account2, tokens[0], { from: account1 });
-          const timepoint = await clockFromReceipt[mode](receipt);
+          const tx = await this.votes.connect(this.holder).transferFrom(this.holder, this.recipient, tokens[0]);
+          const timepoint = await time.clockFromReceipt[mode](tx);
 
-          await time.advanceBlock();
-          await time.advanceBlock();
+          await mine();
+          await mine();
```
ernestognw, line 138:
```suggestion
          await mine(2);
```

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,181 +1,197 @@
...
-          this.account2Votes = '0';
+          await this.votes.$_mint(this.holder, tokens[1]);
+          await this.votes.$_mint(this.holder, tokens[2]);
+          await this.votes.$_mint(this.holder, tokens[3]);
+
+          const total = await this.votes.balanceOf(this.holder);
+
+          const t1 = await this.votes.connect(this.holder).delegate(this.other1);
+          await mine();
+          await mine();
```
ernestognw, line 156:
```suggestion
          await mine(2);
```

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,181 +1,197 @@
...
+          const t2 = await this.votes.connect(this.holder).transferFrom(this.holder, this.other2, tokens[0]);
+          await mine();
+          await mine();
```
ernestognw, line 159:
```suggestion
          await mine(2);
```

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,181 +1,197 @@
...
+          const t3 = await this.votes.connect(this.holder).transferFrom(this.holder, this.other2, tokens[2]);
+          await mine();
+          await mine();
```
ernestognw, line 162:
```suggestion
          await mine(2);
```

test/token/ERC721/extensions/ERC721Votes.test.js
```diff
@@ -1,181 +1,197 @@
...
+          const t4 = await this.votes.connect(this.other2).transferFrom(this.other2, this.holder, tokens[2]);
+          await mine();
+          await mine();
```
ernestognw, line 165:
```suggestion
          await mine(2);
```

test/governance/utils/Votes.behavior.js
```diff
@@ -1,348 +1,321 @@
...
-          await time.advanceBlock();
-          const { receipt } = await this.votes.delegate(accounts[2], { from: accounts[1] });
-          const timepoint = await clockFromReceipt[mode](receipt);
-          await time.advanceBlock();
-          await time.advanceBlock();
+          await mine();
+          const delegate = await this.votes.connect(this.alice).delegate(this.bob);
+          const timepoint = await time.clockFromReceipt[mode](delegate);
+          await mine();
+          await mine();
```
ernestognw, line 316:
```suggestion
          await mine(2);
```

test/governance/utils/Votes.behavior.js
```diff
@@ -1,348 +1,321 @@
...
-          await time.advanceBlock();
-          await time.advanceBlock();
-
-          const latest = await this.votes.getVotes(accounts[2]);
-          expect(await this.votes.getPastVotes(accounts[2], timepoint)).to.be.bignumber.equal(latest);
-          expect(await this.votes.getPastVotes(accounts[2], timepoint + 1)).to.be.bignumber.equal(latest);
+          const delegate = await this.votes.connect(this.alice).delegate(this.bob);
+          const timepoint = await time.clockFromReceipt[mode](delegate);
+          await mine();
+          await mine();
```
ernestognw, line 304:
```suggestion
          await mine(2);
```

test/governance/Governor.test.js
```diff
@@ -663,349 +621,375 @@ contract('Governor', function (accounts) {
...
           );
-          await expectRevertCustomError(this.helper.propose(), 'GovernorInvalidProposalLength', [1, 0, 1]);
+          await expect(this.helper.propose())
+            .to.be.revertedWithCustomError(this.mock, 'GovernorInvalidProposalLength')
+            .withArgs(1, 0, 1);
         });
       });
 
       describe('frontrun protection using description suffix', function () {
+        const shouldPropose = () => {
```
ernestognw, line 744: I'd keep this one as a regular function instead of an arrow function. The reason is that Mocha itself [discourages passing arrow functions](https://github.com/mochajs/mocha/blob/eca4fec9eee1332a0474b80aa1740822438955f3/docs/index.md#arrow-functions).

Here we don't need the `this` context outside of the `it` blocks, but I think we should follow for consistency.

```suggestion
        function shouldPropose() {
```

test/token/ERC20/extensions/ERC20Votes.test.js
```diff
@@ -1,585 +1,556 @@
...
+              .then(ethers.Signature.from);
+
+            const tx = await this.token.delegateBySig(this.delegatee, nonce, ethers.MaxUint256, v, r, s);
+
+            const { args } = await tx
+              .wait()
+              .then(receipt => receipt.logs.find(event => event.fragment.name == 'DelegateChanged'));
+            expect(args[0]).to.not.equal(this.holder.address);
+            expect(args[1]).to.equal(ethers.ZeroAddress);
+            expect(args[2]).to.equal(this.delegatee.address);
```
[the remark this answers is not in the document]
Amxx: Here there is one `not.equal`.

We can do `not.deep.equal` ... and that would pass because arg[0] is different ... but then that would not check that args[1] and args[2] are in fact equals

ernestognw (approved):
LGTM

test/helpers/txpool.js
Amxx: done

test/token/ERC20/extensions/ERC20Votes.test.js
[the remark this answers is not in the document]
ernestognw: Got it, it makes sense.

ernestognw (approved):

PR #4730, opened 2023-11-10
Use ERC-XXX syntax

RenanSouza2:
#### PR Checklist

- [ ] Tests
- [ ] Documentation


ernestognw:
Hey @RenanSouza2, this looks good to me but I'm wondering what's the reason for this change?

RenanSouza2:
Hey :)

it is also for consistency since 6372 is an ERC: (https://eips.ethereum.org/EIPS/eip-6372), 

It is almost meaningless and I can withdraw if you want to

ernestognw (approved):
Ah, I see our behavioral tests start with `ERC*`. I think then this makes sense.

However, the only case I'm worried about is the EIP-712, I'm thinking it should be ERC-712. But maybe is another conversation.

I'll wait for @Amxx confirmation because he may have thoughts. From now it LGTM

Amxx (commented):
This change makes sens to me, but its not complete.

IMO, we should also change the references to "EIP-6372" in  
- contracts/governance/extensions/GovernorVotes.sol
- contracts/governance/utils/Votes.sol
- contracts/governance/README.adoc
- contracts/governance/IGovernor.sol

Amxx:
About 712:

EIP-712 is part of the "interface" track, and not of the "ERC" track. That is why its refered as an EIP and not an ERC. However, it does depend on ERC-191, which is a very similar standard (signature format) and that is part of the "ERC" track.

I believe there is some inconsistency here. IMO an EIP should not depend on an ERC. I'm not sure how that should be resolved.

Amxx:
@ernestognw 

I just made this commit 4c1db78e8fdc2b0137d546809fdf0d87d859a522 that use the `ERC-XXX` syntax in many places. What do you think? I might me missing some.

ernestognw:
> About 712:
> 
> EIP-712 is part of the "interface" track, and not of the "ERC" track. That is why its refered as an EIP and not an ERC. However, it does depend on ERC-191, which is a very similar standard (signature format) and that is part of the "ERC" track.
> 
> I believe there is some inconsistency here. IMO an EIP should not depend on an ERC. I'm not sure how that should be resolved.

I agree with this inconsistency, but let's not block this for that. We can't change the name of the interface anyway since it's breaking.

> @ernestognw

I just made this commit https://github.com/OpenZeppelin/openzeppelin-contracts/commit/4c1db78e8fdc2b0137d546809fdf0d87d859a522 that use the ERC-XXX syntax in many places. What do you think? I might me missing some.

Added a couple and fixed the CI

ernestognw (approved):
LGTM

ernestognw (approved):
Fixed conflicts. LGTM

PR #4733, opened 2023-11-11
Implement revert tests for VestingWallet

RenanSouza2:
Implement revert tests for VestingWallet

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/finance/VestingWallet.test.js
+++ b/test/finance/VestingWallet.test.js
@@ -46,7 +46,6 @@ describe('VestingWallet', function () {
       beforeEach(async function () {
         await this.sender.sendTransaction({ to: this.mock, value: this.amount });
 
-        this.getBalance = ethers.provider.getBalance;
         this.checkRelease = async (tx, amount) => {
           await expect(tx).to.emit(this.mock, 'EtherReleased').withArgs(amount);
           await expect(tx).to.changeEtherBalances([this.mock, this.beneficiary], [-amount, amount]);
@@ -71,7 +70,6 @@ describe('VestingWallet', function () {
         this.token = await ethers.deployContract('$ERC20', ['Name', 'Symbol']);
         await this.token.$_mint(this.mock, this.amount);
 
-        this.getBalance = this.token.balanceOf;
         this.checkRelease = async (tx, amount) => {
           await expect(tx).to.emit(this.token, 'Transfer').withArgs(this.mock.target, this.beneficiary.address, amount);
           await expect(tx).to.changeTokenBalances(this.token, [this.mock, this.beneficiary], [-amount, amount]);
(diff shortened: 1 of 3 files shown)


Amxx (dismissed):
LGTM

Amxx (approved):

test/finance/VestingWallet.test.js
```diff
@@ -62,11 +71,19 @@ describe('VestingWallet', function () {
         this.token = await ethers.deployContract('$ERC20', ['Name', 'Symbol']);
         await this.token.$_mint(this.mock, this.amount);
 
-        this.getBalance = account => this.token.balanceOf(account);
```
ernestognw, line 65: We don't need this functions anymore now that we're using `changeTokenBalances` and `changeEtherBalances`

ernestognw (approved):

RenanSouza2:
@ernestognw I made a bigger cleanup, see if you approve it

ernestognw:
> @ernestognw I made a bigger cleanup, see if you approve it

Thanks, I changed a couple things, the main one is that I brought back the `releasedEvent` and `argsVerify`. Otherwise, we're not checking the `ERC20Released` event. Also, I think the vesting wallet behavior should check for a release event every time, regardless what other checks are contained in `checkRelease`.

ernestognw (approved):

RenanSouza2:
and what do you say abbout this proposal?

https://github.com/RenanSouza2/openzeppelin-contracts/tree/add-finance-tests-proposal/test/finance

ernestognw:
> and what do you say abbout this proposal?
> 
> 
> 
> https://github.com/RenanSouza2/openzeppelin-contracts/tree/add-finance-tests-proposal/test/finance
> 
> 
> 
> I've incorporated both events tests in 'checkRelease'

Not convinced. Right now it makes no difference because the behavioral tests are only used here, but having the event checks outside the behavioral test seems an antipattern to me. It shouldn't be possible to pass the behavioral test without an event.

RenanSouza2:
> Not convinced. Right now it makes no difference because the behavioral tests are only used here, but having the event checks outside the behavioral test seems an antipattern to me. It shouldn't be possible to pass the behavioral test without an event being emitted on releasing.

Makes sense, this is a way to enfor the verification of at least one test

ernestognw (approved):
Noted some small improvements. Re approving.

ernestognw:
> Makes sense, this is a way to enforce the verification of at least one test

Yeah, I think it's a small decision honestly, so I'm open to changing it if @Amxx sees something I'm ignoring.

The PR is in good shape for merging in my opinion but we can create another PR if anyone feels strongly about one of the last decisions. I'm merging

PR #4739, opened 2023-11-14
Migrate address to ethersjs

RenanSouza2:
Opening this PR separetedly because of unknown errors on coverage test

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/utils/Address.test.js
+++ b/test/utils/Address.test.js
@@ -1,6 +1,8 @@
 const { ethers } = require('hardhat');
 const { expect } = require('chai');
 const { loadFixture } = require('@nomicfoundation/hardhat-network-helpers');
+const { PANIC_CODES } = require("@nomicfoundation/hardhat-chai-matchers/panic");
+
 const coder = ethers.AbiCoder.defaultAbiCoder();
 
 async function fixture() {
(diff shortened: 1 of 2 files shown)


test/utils/Address.test.js
```diff
@@ -1,333 +1,287 @@
...
 
-        await expectRevert(this.mock.$functionCall(this.target.address, abiEncodedCall), 'CallReceiverMock: reverting');
+        await expect(this.mock.$functionCall(this.target, abiEncodedCall)).to.be.revertedWith(
+          'CallReceiverMock: reverting',
+        );
       });
 
       it('reverts when the called function runs out of gas', async function () {
-        const abiEncodedCall = this.target.contract.methods.mockFunctionOutOfGas().encodeABI();
+        if (process.env.COVERAGE) this.skip();
```
RenanSouza2, line 118: Not elegant but this test was not working in coverage

Amxx (dismissed):

test/utils/Address.test.js
Amxx: I think that is because you were not using `gasLimit` to set the transaction's amount of gas. This is a big difference between web3.js and ethers.js

Amxx: Also, when a tests needs to be disable on coverage, we name it
```javascript
it('[skip-on-coverage] ...', async function () {
   ...
});
```

(that prefix is specified in `.solcover.js`)

Amxx (approved):

test/utils/Address.test.js
RenanSouza2: Can't believe I missed this

> I think that is because you were not using `gasLimit` to set the transaction's amount of gas. This is a big difference between web3.js and ethers.js

Can't believe I missed this

ernestognw (approved):
Pretty neat, LGTM

Issue #4214, opened 2023-05-04
Governor shadows EIP-712's `_name`


ernestognw:
**📝 Details**

Governor initializes with the following constructor:

```solidity
constructor(string memory name_) EIP712(name_, version()) {
    _name = name_;
}
```

However, `_name` is also the name of the variable in EIP712:

```solidity
constructor(string memory name, string memory version) {
    _name = name.toShortStringWithFallback(_nameFallback);
    _version = version.toShortStringWithFallback(_versionFallback);
    _hashedName = keccak256(bytes(name));
    _hashedVersion = keccak256(bytes(version));

    _cachedChainId = block.chainid;
    _cachedDomainSeparator = _buildDomainSeparator();
    _cachedThis = address(this);
}
```

Apparently, there's no point in having the storage variable in the Governor when the name in EIP712 can be used.

frangio:
There is a point about shadowing, which IMO is not concerning, and a point about redundancy. A similar thing comes up in ERC20 with ERC20Permit, since it needs a name for EIP712 and ERC20 already has one.

balajipachai:
> **memo Details**
> 
> Governor initializes with the following constructor:
> 
> ```solidity
> constructor(string memory name_) EIP712(name_, version()) {
>     _name = name_;
> }
> ```
> 
> However, `_name` is also the name of the variable in EIP712:
> 
> ```solidity
> constructor(string memory name, string memory version) {
>     _name = name.toShortStringWithFallback(_nameFallback);
>     _version = version.toShortStringWithFallback(_versionFallback);
>     _hashedName = keccak256(bytes(name));
>     _hashedVersion = keccak256(bytes(version));
> 
>     _cachedChainId = block.chainid;
>     _cachedDomainSeparator = _buildDomainSeparator();
>     _cachedThis = address(this);
> }
> ```
> 
> Apparently, there's no point in having the storage variable in the Governor when the name in EIP712 can be used.

I agree since EIP712 already has

> There is a point about shadowing, which IMO is not concerning, and a point about redundancy. A similar thing comes up in ERC20 with ERC20Permit, since it needs a name for EIP712 and ERC20 already has one.

Talking about redundancy can't we just remove _name from Governor contract?

balajipachai:
@frangio @Amxx Any updates on this?

RenanSouza2:
[...]

PR #4741, opened 2023-11-15
Use the inherited EIP712 name in the Governor contract

BCJdevelopment:
The Governor contract includes a {_name} state variable, but the parent EIP712 contract already uses and stores the exact same value. This is unnecessary duplicate storage.

This PR removes the duplicate {_name} state variable from the Governor, and in the Governor's public "name()" function, returns the result of the internal EIP712 function for retrieving the {_name} value.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ernestognw:
This requires a bit more discussion because it may be a breaking change (it was in 4.x).
Here's the context: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4214

PR #4743, opened 2023-11-16
Migrate erc20 to ethersjs

RenanSouza2:
Migrate 'ERC20' tests to ethersjs

#### PR Checklist

- [ ] Tests
- [ ] Documentation


test/token/ERC20/ERC20.behavior.js
```diff
@@ -1,5 +1,7 @@
-const { BN, constants, expectEvent } = require('@openzeppelin/test-helpers');
+const { ethers } = require('hardhat');
 const { expect } = require('chai');
+
+const { BN, constants, expectEvent } = require('@openzeppelin/test-helpers');
```
Amxx, line 4: Test migration should remove this from the dependency.

test/token/ERC20/ERC20.test.js
```diff
@@ -1,34 +1,44 @@
...
+
+  return { initialHolder, recipient, anotherAccount, tokens };
+}
+
+describe('ERC20', function () {
+  beforeEach(async function () {
+    Object.assign(this, await loadFixture(fixture));
+  });
 
   for (const { Token, forcedApproval } of TOKENS) {
```
Amxx, line 35: I would move the `fixture` function inside this, so that only one contract is deployed.

Something like
```javascript
for (const { Token, forcedApproval } of TOKENS) {
  async function fixture() {
    // ...
    const token = await ethers.deployContract(Token, [name, symbol]);
    await token.$_mint(initialHolder, initialSupply);
  
    return { /* ... */, token };
  }

    Object.assign(this, await loadFixture(fixture));
  });
  
  shouldBehaveLikeERC20(initialSupply, { forcedApproval });
  
  // ...
}

File: test/token/ERC20/ERC20.test.js [hunk 3]
```diff
@@ -39,162 +49,169 @@ contract('ERC20', function (accounts) {
...
           beforeEach('minting', async function () {
-            this.receipt = await this.token.$_mint(recipient, value);
+            this.tx = await this.token.$_mint(this.recipient, value);
           });
 
           it('increments totalSupply', async function () {
-            const expectedSupply = initialSupply.add(value);
-            expect(await this.token.totalSupply()).to.be.bignumber.equal(expectedSupply);
+            const expectedSupply = initialSupply + value;
+            await expect(await this.token.totalSupply()).to.equal(expectedSupply);
```
Amxx, line 76:
```suggestion
            expect(await this.token.totalSupply()).to.equal(expectedSupply);
```

test/token/ERC20/ERC20.test.js
```diff
@@ -39,162 +49,169 @@ contract('ERC20', function (accounts) {
...
           });
 
           it('increments recipient balance', async function () {
-            expect(await this.token.balanceOf(recipient)).to.be.bignumber.equal(value);
+            expect(await this.token.balanceOf(this.recipient)).to.equal(value);
```
Amxx, line 80:
```suggestion
            expect(this.tx).to.changeTokenBalance(this.token, this.recipient, value);
```

test/token/ERC20/ERC20.test.js
```diff
@@ -39,162 +49,169 @@ contract('ERC20', function (accounts) {
...
-            'reverted with panic code 0x11 (Arithmetic operation underflowed or overflowed outside of an unchecked block)',
+          await expect(this.token.$_mint(this.recipient, ethers.MaxUint256)).to.be.revertedWithPanic(
+            PANIC_CODES.ARITHMETIC_UNDER_OR_OVERFLOW,
           );
         });
 
         describe('for a non zero account', function () {
           beforeEach('minting', async function () {
-            this.receipt = await this.token.$_mint(recipient, value);
+            this.tx = await this.token.$_mint(this.recipient, value);
```
Amxx, line 71:
```suggestion
            this.tx = this.token.$_mint(this.recipient, value);
```

test/token/ERC20/ERC20.test.js
Amxx, line 76:
```suggestion
            await this.tx;
            expect(await this.token.totalSupply()).to.equal(initialSupply + value);
```

test/token/ERC20/ERC20.test.js
```diff
@@ -39,162 +49,169 @@ contract('ERC20', function (accounts) {
...
+            await expect(this.token.$_burn(this.initialHolder, initialSupply + 1n))
+              .to.be.revertedWithCustomError(this.token, 'ERC20InsufficientBalance')
+              .withArgs(this.initialHolder.address, initialSupply, initialSupply + 1n);
           });
 
           const describeBurn = function (description, value) {
             describe(description, function () {
               beforeEach('burning', async function () {
-                this.receipt = await this.token.$_burn(initialHolder, value);
+                this.tx = await this.token.$_burn(this.initialHolder, value);
```
Amxx, line 108:
```suggestion
                this.tx = this.token.$_burn(this.initialHolder, value);
```

test/token/ERC20/ERC20.test.js
```diff
@@ -39,162 +49,169 @@ contract('ERC20', function (accounts) {
...
               it('decrements totalSupply', async function () {
-                const expectedSupply = initialSupply.sub(value);
-                expect(await this.token.totalSupply()).to.be.bignumber.equal(expectedSupply);
+                expect(await this.token.totalSupply()).to.equal(initialSupply - value);
               });
 
               it('decrements initialHolder balance', async function () {
-                const expectedBalance = initialSupply.sub(value);
-                expect(await this.token.balanceOf(initialHolder)).to.be.bignumber.equal(expectedBalance);
+                expect(await this.token.balanceOf(this.initialHolder)).to.changeTokenBalance(
```
Amxx, line 116:
```suggestion
                await expect(this.tx).to.changeTokenBalance(
```

test/token/ERC20/ERC20.test.js
RenanSouza2: What @ernestognw was to have only one fixture per file to increase the optimization

Amxx: Its between:
- one fixture, executed one time, that deploys X contracts
- X fixtures, executed one time each, that deploys 1 contracts each.

Number of deployment is the same... but the second option does more snapshots (each feature comes with its own snapshot).

I like that the second option sets `this.token` correctly without having to rely on an additional `beforeEach`. I also like the fact that it makes the fixture code simpler. I'm not sure what the performance difference are.

ernestognw: I think is generally better to have a single fixture for 2 reasons:

1. Seems like [fixtures are identified by comparing the function with a list of fixtures](https://github.com/NomicFoundation/hardhat/blob/c3aad2c556a4825d53667af40a065a350e5317be/packages/hardhat-network-helpers/src/loadFixture.ts#L38). My understanding of this is that multiple fixtures slow down execution if there are lots of them.
2. Usually fixtures differ in 1 or 2 contracts (i.e. deploy the same with exception of one or two contracts), my conclusion is that it's better to have 1 single fixture with 5 contracts deployed than 2 fixtures with 4 contracts deployed.

Number 1 is only a problem if we have multiple fixtures, but we only have 2, and Number 2 is just a small preference. No strong opinion on this one, both are fine.

test/token/ERC20/ERC20.behavior.js
RenanSouza2: I did it and removed unused libs

test/token/ERC20/ERC20.test.js
RenanSouza2: See if you approve the new version

test/token/ERC20/ERC20.behavior.js
```diff
@@ -1,335 +1,270 @@
...
+            beforeEach(async function () {
+              this.tx = await this.token
+                .connect(this.recipient)
+                .transferFrom(this.initialHolder, this.anotherAccount, value);
+            });
 
-              expect(await this.token.balanceOf(to)).to.be.bignumber.equal(value);
+            it('transfers the requested value', async function () {
+              expect(await this.token.balanceOf(this.initialHolder)).to.equal(0n);
+              expect(await this.token.balanceOf(this.anotherAccount)).to.equal(value);
```
Amxx, line 48: await expect(this.tx)..to.changeTokenBalances(this.token, [ this.initialHolder, this.anotherAccount], [ -value, value ]);

Amxx, line 48:
```suggestion
              await expect(this.tx)..to.changeTokenBalances(this.token, [ this.initialHolder, this.anotherAccount], [ -value, value ]);
```

test/token/ERC20/ERC20.behavior.js
```diff
@@ -1,335 +1,270 @@
...
-
-        expect(await this.token.balanceOf(from)).to.be.bignumber.equal('0');
+      beforeEach(async function () {
+        this.tx = await this.transfer(this.initialHolder, this.recipient, value);
+      });
 
-        expect(await this.token.balanceOf(to)).to.be.bignumber.equal(value);
+      it('transfers the requested value', async function () {
+        expect(await this.token.balanceOf(this.initialHolder)).to.equal(0n);
+        expect(await this.token.balanceOf(this.recipient)).to.equal(value);
```
Amxx, line 178:
```suggestion
        await expect(this.tx).to.changeTokenBalances(this.token, [ this.initialHolder, recipient], [ -value, value ]);
```

test/token/ERC20/ERC20.behavior.js
```diff
@@ -1,335 +1,270 @@
...
-
-        expect(await this.token.balanceOf(from)).to.be.bignumber.equal(balance);
+      beforeEach(async function () {
+        this.tx = await this.transfer(this.initialHolder, this.recipient, value);
+      });
 
-        expect(await this.token.balanceOf(to)).to.be.bignumber.equal('0');
+      it('transfers the requested value', async function () {
+        expect(await this.token.balanceOf(this.initialHolder)).to.equal(balance);
+        expect(await this.token.balanceOf(this.recipient)).to.equal(0n);
```
Amxx, line 197:
```suggestion
        await expect(this.tx).to.changeTokenBalances(this.token, [ this.initialHolder, this.recipient], [ -value, value ]);
```

test/token/ERC20/extensions/ERC20Wrapper.test.js
```diff
@@ -37,175 +40,169 @@ contract('ERC20Wrapper', function (accounts) {
...
-      const { tx } = await this.token.$_recover(receiver);
-      await expectEvent.inTransaction(tx, this.token, 'Transfer', {
-        from: ZERO_ADDRESS,
-        to: receiver,
-        value: '0',
-      });
+      await this.underlying.connect(this.initialHolder).approve(this.token, initialSupply);
+      await this.token.connect(this.initialHolder).depositFor(this.initialHolder, initialSupply);
+
+      const tx = this.token.$_recover(this.recipient);
```
Amxx, line 182: Sometimes we await these `const tx =` or `this.tx =`, and sometimes we don't. I think we should be consistent.

If its works without the await, I'd use that everywhere.

Amxx (commented):
Very nice ! I left some comments.

test/token/ERC20/extensions/ERC20Wrapper.test.js
RenanSouza2: I tested remove then and they don't aways work

if the event is tested first it works because the 'await expect(tx) triggers the execution. 

But if the balance is tested with no other tests prior it will fail because the execution is only triggered by the await

ernestognw (approved):
Flawless, thanks!

PR #4749, opened 2023-11-20
Migrate utils-cryptography to ethers

RenanSouza2:
#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx (dismissed):

Amxx (approved):

test/utils/cryptography/ECDSA.test.js
```diff
@@ -1,104 +1,101 @@
...
-  }
-  if (long[32] >> 7 === 1) {
-    throw new Error("invalid signature 's' value");
-  }
-  const short = long.slice(0, 64);
-  short[32] |= long[64] % 27 << 7; // set the first bit of the 32nd byte to the v parity bit
-  return web3.utils.bytesToHex(short);
+function toERC2098Format(signature) {
+  return ethers.Signature.from(signature).compactSerialized;
 }
```
ernestognw, line 11: I think this abstraction is no longer needed and perhaps we can replace with a simple:

```js
function toSignature(signature) {
  return ethers.Signature.from(signature);
}
```

This way we can access `r, s, v` more easily without the `split` function.

test/utils/cryptography/ECDSA.test.js
```diff
@@ -1,104 +1,101 @@
...
         raw[64], // v
-        web3.utils.bytesToHex(raw.slice(0, 32)), // r
-        web3.utils.bytesToHex(raw.slice(32, 64)), // s
+        ethers.hexlify(raw.slice(0, 32)), // r
+        ethers.hexlify(raw.slice(32, 64)), // s
       ];
     default:
       expect.fail('Invalid signature length, cannot split');
   }
 }
```
ernestognw, line 31: We can remove it almost everywhere using `toSignature` instead except for the `reverts with high-s value signature` tests since it's impossible to instantiate a high-s signature in ethers. We can get `r, s, v` values in the following way:

```js
const r = ethers.dataSlice(highSSignature, 0, 32);
const s = ethers.dataSlice(highSSignature, 32, 64);
const v = ethers.dataSlice(highSSignature, 64, 65);
```

ernestognw:
Alright, tests passing and additions are equivalent to previous approval. Merging.

Thanks both!

ernestognw (approved):

PR #4750, opened 2023-11-20
Migrate EIP712 to ethersjs

RenanSouza2:
#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx (approved):

Amxx (dismissed):

RenanSouza2:
This one needs to update the 'upgradeable.patch' file and tbh I have no Idea how to do it

Amxx (approved):

test/helpers/eip712-types.js
```diff
@@ -0,0 +1,16 @@
...
+    { name: 'salt', type: 'bytes32' },
+  ],
+  Permit: [
+    { name: 'owner', type: 'address' },
+    { name: 'spender', type: 'address' },
+    { name: 'value', type: 'uint256' },
+    { name: 'nonce', type: 'uint256' },
+    { name: 'deadline', type: 'uint256' },
+  ],
+};
```
ernestognw, line 16: I'm pretty sure we're missing a bunch of these. At least `ForwardRequest` from meta transactions is coming to my mind, I'll look for definitions and add to this file

test/helpers/eip712.js
ernestognw, line 1: I have the sense that all of these definitions can be replaced with ethers [TypedDataEncoder](https://docs.ethers.org/v6/api/hashing/#TypedDataEncoder). I'll give it a shot

ernestognw: Discarded since I see ethers doesn't provide some of the values we need. We may revisit.

test/helpers/eip712-types.js
ernestognw: Moved them

ernestognw:
Moved the fixture outside because it seems better to have a single fixture with a snapshot to go back instead of many.

ernestognw (approved):
LGTM 🚢

PR #4754, opened 2023-11-23
✅ excluding private keys interaction test case

mohamadhammoud:
**Description of Changes:**
In the existing ERC20 Permit tests, the focus was primarily on direct interactions with private keys. This pull request extends the test suite to cover scenarios where interactions are performed through a wallet interface, aligning more closely with real-world usage where private keys are not explicitly handled in the code.

The added tests ensure that the ERC20 permit functionality works as expected when users interact with it via a wallet interface, such as MetaMask, rather than directly managing private keys. This enhances the robustness of the tests by simulating more realistic user interactions.

Fixes #???? 
1. Extended the ERC20 Permit test suite to include wallet interface interactions.
2. Ensured compatibility with existing test patterns within OpenZeppelin's framework.
3. Maintained consistency with OpenZeppelin's coding standards and testing conventions.

#### PR Checklist

- [X] Tests
- [X] Documentation

### DIFF
--- a/.changeset/fluffy-jobs-brake.md
+++ b/.changeset/fluffy-jobs-brake.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Added Wallet Interface Interaction Tests to ERC20 Permit: Enhanced the ERC20 Permit test suite by introducing tests for wallet interface interactions. This update simulates more realistic user scenarios, ensuring the permit functionality works seamlessly when interacting through wallet interfaces like MetaMask, without direct handling of private keys.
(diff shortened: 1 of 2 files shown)


Amxx (changes requested):
This change should be part of the overall translation of ERC20 tests to ethers.js. This is in the pipeline already.

Also, tests migration don't need a changelog entry.

Amxx:
This change was implemented as part of the ERC20 test migration (#4773)

See: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/token/ERC20/extensions/ERC20Permit.test.js#L62-L71

PR #4755, opened 2023-11-24
Add note in ERC20Wrapper about rebasing tokens

ernestognw:
The ERC20Wrapper contract doesn't include any mention of rebasing tokens. Although there's no way of abstracting the fee mechanism to the wrapper, it can also be a use case to wrap fee-on-transfer tokens. Regardless, it makes sense to add a small note mentioning such cases should be considered.

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/contracts/token/ERC20/extensions/ERC20Wrapper.sol
+++ b/contracts/token/ERC20/extensions/ERC20Wrapper.sol
@@ -12,6 +12,9 @@ import {SafeERC20} from "../utils/SafeERC20.sol";
  * Users can deposit and withdraw "underlying tokens" and receive a matching number of "wrapped tokens". This is useful
  * in conjunction with other modules. For example, combining this wrapping mechanism with {ERC20Votes} will allow the
  * wrapping of an existing "basic" ERC-20 into a governance token.
+ *
+ * NOTE: Any mechanism in which the underlying token changes the {balanceOf} of an account without an explicit transfer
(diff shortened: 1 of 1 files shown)


contracts/token/ERC20/extensions/ERC20Wrapper.sol
```diff
@@ -12,6 +12,9 @@ import {SafeERC20} from "../utils/SafeERC20.sol";
  * Users can deposit and withdraw "underlying tokens" and receive a matching number of "wrapped tokens". This is useful
  * in conjunction with other modules. For example, combining this wrapping mechanism with {ERC20Votes} will allow the
  * wrapping of an existing "basic" ERC-20 into a governance token.
+ *
+ * NOTE: Any mechanism in which the underlying token changes the {balanceOf} of an account without an explicit transfer
+ * may desynchronize the wrapper's supply and its balance. See {_recover} for recovering value accrued to the wrapper.
```
Amxx, line 17:
```suggestion
 * NOTE: Any mechanism in which the underlying token changes the {balanceOf} of an account without an explicit transfer
 * may desynchronize the wrapper's supply and its balance. See {_recover} for recovering value accrued to the wrapper.
 *
 * WARNING: Wrapping a token that implement fees may lead to under-collateralization, making the wrapped asset unredeamable. Please exercise caution when wrapping such tokens.
```

Amxx (approved):

PR #4756, opened 2023-11-24
Migrate to Node 20

### LINKED ISSUE #4719: Migrate to Node 20
**🧐 Motivation**

We have a few dependency PR updates that require a higher node version:

- https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4716
- https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4702
- https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4781
- https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4780
- https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4770

The minimum node version between both is `18` but ideally we should upgrade to LTS (20)

**📝 Details**

We need to update [the CI](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/abba0d047a24c8feaadf35589ea81a433c86ec0a/.github/actions/setup/action.yml). If migrated correctly the CI should pass along with a PR solving this.

Note updating dependencies is considered highly sensitive so we need to be careful while upgrading.

ernestognw:
Fixes #4719 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -5,7 +5,7 @@ runs:
   steps:
     - uses: actions/setup-node@v3
       with:
-        node-version: 16.x
+        node-version: 20
     - uses: actions/cache@v3
       id: cache
       with:


.github/actions/setup/action.yml
```diff
@@ -5,7 +5,7 @@ runs:
   steps:
     - uses: actions/setup-node@v3
```
ericglau, line 6: Perhaps we should also try moving `setup-node` to v4 as well.  According to setup-node's [release notes](https://github.com/actions/setup-node/releases), v4 changes the action itself to use node 20.

ernestognw: Ha! Good prompt, there's a PR for that but it still uses 16.x in the setup file:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4702

I'll update here and the other PR should be closed automatically

.github/actions/setup/action.yml
```diff
@@ -3,9 +3,7 @@ name: Setup
 runs:
   using: composite
   steps:
-    - uses: actions/setup-node@v3
-      with:
-        node-version: 16.x
+    - uses: actions/setup-node@v4
```
ericglau, line 6: We should still specify node 20 to be more explicit.  setup-node's documentation recommends to "always specify Node.js version and don't rely on the system one" 
```suggestion
    - uses: actions/setup-node@v4
      with:
        node-version: 20.x
```

ernestognw:
To check differences, I replaced `.gitignore` to only ignore `node_modules` and then I ran a few scripts in the `package.json`. I staged the results with `git add .` and then re-ran the scripts using node v20 to get the output differences. The scripts ran and their results were:

#### `npm run compile`

The build info file changed. Output is too long to get a diff and the [rest diff is here](https://gist.github.com/ernestognw/0b2d9cd41c44af7dffa386e96cafcf31). 

The only relevant thing I can see is that there are missing functions in the `hardhat-exposed` output, but those don't affect the tests. Also, this block was removed from the `solidity-files-cache.json`:

```
"metadata": {
  "useLiteralContent": false
}
```

Do you think this may be a problem for upgrades plugins @ericglau?

#### `npm run coverage`

[Diff here](https://gist.github.com/ernestognw/1cb032e37717de63625926ba340d00b2). Only changes to the HTML output files.

#### `npm run docs`

[Diff here](https://gist.github.com/ernestognw/4665e4414e3510868a2e1bab23d04101). Very similar result to the compile script.

#### `npm run prepack`

[Diff here](https://gist.github.com/ernestognw/8caeec336a4393c02643a71da1985a90). Some doc files disappear but this is normal since `prepack` also runs `npm run clean`.

I think this is generally fine but would appreciate if you take a look @Amxx.

.github/actions/setup/action.yml
ernestognw: I used `code blame` to see if there was a related comment justifying an explicit version in the past. But forgot to check the docs 😅 

I agree, let's revert it back

ericglau:
> Also, this block was removed from the solidity-files-cache.json:
> ```
> "metadata": {
>   "useLiteralContent": false
> }
> ```
I think this part is fine since useLiteralContent is `false` by default anyways, according to Solidity docs.

Amxx:
> The only relevant thing I can see is that there are missing functions in the `hardhat-exposed` output, but those don't affect the tests.

That is the only thing that wories me. I'd think we should test the `hardhat-exposed` repo with node 20 and check that everything passes there.

Amxx (approved):

ernestognw:
For the record, we realized the diff in `hardhat-exposed` is normal.

Since I ran all the scripts with Node 16 first and then staged the changes, some functions generated during coverage were staged. After re-rerunning `npm run compile` with Node 20 those selectors were removed.

ericglau (approved):

PR #4766, opened 2023-11-28
Add Governor note for state changes between proposal creation and execution

ernestognw:
The Governor doesn't check that a proposal is executable at the moment of calling `propose` or `execute`. While this makes sense because the execution requirements can be modified since proposal creation until execution, it's worth noting that such changes may have effects on those proposals that may not be obvious.

A concrete example would be a proposal that transfers a total of 3 ETH among 3 targets. In this case, the proposal will depend on the Governor having 3 ETH on its balance, but a faster proposal may use the funds of the Governor, blocking the execution of the 3rd proposal.

Although, it's not an issue since the proposal can still be executed since the `execute` function is payable, we're adding a note to take in consideration any potential state change between the proposal creation and the proposal execution

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/contracts/governance/IGovernor.sol
+++ b/contracts/governance/IGovernor.sol
@@ -318,10 +318,10 @@ interface IGovernor is IERC165, IERC6372 {
      *
      * Emits a {ProposalCreated} event.
      *
-     * NOTE: The state of the Governor and `targets` may change since proposal creation until its execution. For example,
-     * the balance of this contract may change or its access control permissions may be modified, possibly compromising
-     * the proposal's ability to execute successfully (e.g. the governor doesn't have enough value to cover a proposal 
-     * with multiple transfers).
+     * NOTE: The state of the Governor and `targets` may change since the proposal creation until its execution.
+     * For example, the balance of this contract could be updated or its access control permissions may be modified,
+     * possibly compromising the proposal's ability to execute successfully (e.g. the governor doesn't have enough
+     * value to cover a proposal with multiple transfers).
      */
     function propose(
         address[] memory targets,
--- a/contracts/governance/IGovernor.sol
+++ b/contracts/governance/IGovernor.sol
@@ -317,6 +317,11 @@ interface IGovernor is IERC165, IERC6372 {
      * duration specified by {IGovernor-votingPeriod}.
      *
      * Emits a {ProposalCreated} event.
+     *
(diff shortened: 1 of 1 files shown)


ernestognw:
I added the note to the `propose` function because I think it's something to note when creating a proposal and not when it's executed. Just FYI @Amxx

contracts/governance/IGovernor.sol
```diff
@@ -317,6 +317,11 @@ interface IGovernor is IERC165, IERC6372 {
      * duration specified by {IGovernor-votingPeriod}.
      *
      * Emits a {ProposalCreated} event.
+     *
+     * NOTE: The state of the Governor and `targets` may change since the proposal creation until its execution.
+     * For example, the balance of this contract could be updated or its access control permissions may be modified,
+     * possibly compromising the proposal's ability to execute successfully (e.g. the governor doesn't have enough
+     * value to cover a proposal with multiple transfers).
```
Amxx:
```suggestion
     * NOTE: The state of the Governor and `targets` may change since the proposal creation until its execution.
     * This may be the result of third party actions on the targeted contract, or of other governor proposals.
     * For example, the balance of this contract could be updated or its access control permissions may be modified,
     * possibly compromising the proposal's ability to execute successfully (e.g. the governor doesn't have enough
     * value to cover a proposal with multiple transfers).
```

contracts/governance/IGovernor.sol
```diff
@@ -317,6 +317,11 @@ interface IGovernor is IERC165, IERC6372 {
      * duration specified by {IGovernor-votingPeriod}.
      *
      * Emits a {ProposalCreated} event.
+     *
+     * NOTE: The state of the Governor and `targets` may change between the proposal creation and its execution.
+     * For example, the balance of this contract could be updated or its access control permissions may be modified,
+     * possibly compromising the proposal's ability to execute successfully (e.g. the governor doesn't have enough
+     * value to cover a proposal with multiple transfers).
```
Amxx, line 324:
```suggestion
     * NOTE: The state of the Governor and `targets` may change between the proposal creation and its execution.
     * This may be the result of third party actions on the targeted contracts, or of other governor proposals.
     * For example, the balance of this contract could be updated or its access control permissions may be modified,
     * possibly compromising the proposal's ability to execute successfully (e.g. the governor doesn't have enough
     * value to cover a proposal with multiple transfers).
```

Amxx (approved):

PR #4775, opened 2023-11-30
Update Math `try*` operations return reference

### LINKED ISSUE #4753: The overflow flag in the math lib is inverted
In the math library, it looks like the return value of the overflow flag is inverted. 
It returns a `false` when there is an overflow and returns a `true` for a valid result. 
The impact might be minimal, if the contracts using the library don't track the overflow flag. 
But the logic would be incorrect if they indeed to track the overflow return param. 

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/932fddf69a699a9a80fd2396fd1a2ab91cdda123/contracts/utils/math/Math.sol#L25

v5.0.0
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.0.0/contracts/utils/math/Math.sol#L25

**💻 Environment**
On your github

**📝 Details**

The `overflow` flag in the math library is inverted. 
For instance below is the `tryAdd` function which returns a (bool,uint256). 
The `bool` is an overflow flag as shown in the NatSpec comment below. 
However, the return is a `false` when there is an overflow and the return is a `true` when the result is valid.
This also extends to the other functions.

    /**
     * @dev Returns the addition of two unsigned integers, with an overflow flag.
     */
    function tryAdd(uint256 a, uint256 b) internal pure returns (bool, uint256) {
        unchecked {
            uint256 c = a + b;
            if (c < a) return (false, 0); //@audit should be `return (true,0)` 
            return (true, c);  //@audit should be `return (false,c)`
        }
    }
**🔢 Code to reproduce bug**

ernestognw:
Fixes #4753 

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -20,9 +20,9 @@ library Math {
     }
 
     /**
-     * @dev Returns the addition of two unsigned integers, with an overflow flag.
+     * @dev Returns the addition of two unsigned integers, with an success flag (no overflow).
      */
-    function tryAdd(uint256 a, uint256 b) internal pure returns (bool, uint256) {
+    function tryAdd(uint256 a, uint256 b) internal pure returns (bool success, uint256 result) {
         unchecked {
             uint256 c = a + b;
             if (c < a) return (false, 0);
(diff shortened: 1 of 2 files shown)


ernestognw:
Added named return parameters since I think they're good for clarity, but I'm open to removing them if you think it's too big of a change @Amxx

Amxx (approved):

PR #4792, opened 2023-12-10
Remove Context from "Ownable" contract

BCJdevelopment:
By default, the "Ownable" and "Ownable2Step" contracts make use of "Context" to create the option for _msgSender() to be used for validating message senders.

However, for most "ownerOnly" types of functions, proper sender authorization is critical, and these should not be exposed by default to the potential risks or complications from an ERC2771 forwarder implementation.

This PR removes the Context from Ownable and Ownable2Step, and just uses msg.sender by default. Developers can override these functions as desired to use Context if they want, but it shouldn't be included by default.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/access/Ownable.sol
+++ b/contracts/access/Ownable.sol
@@ -3,8 +3,6 @@
 
 pragma solidity ^0.8.20;
 
-import {Context} from "../utils/Context.sol";
-
 /**
  * @dev Contract module which provides a basic access control mechanism, where
  * there is an account (an owner) that can be granted exclusive access to
@@ -17,7 +15,7 @@ import {Context} from "../utils/Context.sol";
  * `onlyOwner`, which can be applied to your functions to restrict their use to
  * the owner.
  */
-abstract contract Ownable is Context {
+abstract contract Ownable {
     address private _owner;
 
     /**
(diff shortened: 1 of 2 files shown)


Amxx:
> Developers can override these functions as desired to use Context

That is the whole point of Context. By default, `_msgSender()` returns `msg.sender`, and developpers can override it.

BCJdevelopment:
> > Developers can override these functions as desired to use Context
> 
> That is the whole point of Context. By default, `_msgSender()` returns `msg.sender`, and developpers can override it.

Yes, which is fine for developers to integrate into functions that they want to use meta-transactions for, but my point is that most "onlyOwner" modified functions are not anticipated to use the ERC2771 forwarder as well. In most cases, the owner is a DAO Executor, a multi-sig, or some other admin entity that does not need to use meta-transactions for these onlyOwner operations. 

But when the "Ownable" contracts use Context by default, and then developers override the context functions to use a context forwarder that has security vulnerabilities, suddenly the functions that are supposed to be admin-only could be exposed as well.

Take the recent address-spoofing attack, https://blog.openzeppelin.com/arbitrary-address-spoofing-vulnerability-erc2771context-multicall-public-disclosure

If their were also "ownerOnly" functions on the vulnerable contract, those would be exposed by default to the same attack vector. And I just don't think most "ownerOnly" functions are expected to be used in meta-transactions anyways, so it seems safer to not use Context in the Ownable contracts by default.

PR #4793, opened 2023-12-11
Migrate ERC721 tests

Amxx:
ERC721Votes is not covered. We should do it in a separate PR after Governance

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/test/token/ERC721/extensions/ERC721Consecutive.test.js
+++ b/test/token/ERC721/extensions/ERC721Consecutive.test.js
@@ -54,15 +54,7 @@ describe('ERC721Consecutive', function () {
                   batch.receiver.address /* toAddress */,
                 );
             } else {
-              // ".to.not.emit" doesn't check the parameters
-              // await expect(this.token.deploymentTransaction())
-              //   .to.not.emit(this.token, 'ConsecutiveTransfer')
-              //   .withArgs(
-              //     first, /* fromTokenId */
-              //     first + batch.amount - 1n, /* toTokenId */
-              //     ethers.ZeroAddress, /* fromAddress */
-              //     batch.receiver, /* toAddress */
-              //   );
+              // ".to.not.emit" only looks at event name, and doesn't check the parameters
             }
             first += batch.amount;
           }
--- a/test/token/ERC721/extensions/ERC721Consecutive.test.js
+++ b/test/token/ERC721/extensions/ERC721Consecutive.test.js
@@ -33,7 +33,7 @@ describe('ERC721Consecutive', function () {
           batches.map(({ amount }) => amount),
         ]);
 
-        return { accounts, alice, bruce, chris, receiver, batches, delegates, token };
+        return { alice, bruce, chris, receiver, batches, delegates, token };
       }
 
       beforeEach(async function () {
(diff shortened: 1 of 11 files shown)


test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -1,222 +1,244 @@
...
-              // expectEvent.notEmitted.inConstruction only looks at event name, and doesn't check the parameters
+              // ".to.not.emit" doesn't check the parameters
+              // await expect(this.token.deploymentTransaction())
+              //   .to.not.emit(this.token, 'ConsecutiveTransfer')
+              //   .withArgs(
+              //     first, /* fromTokenId */
+              //     first + batch.amount - 1n, /* toTokenId */
+              //     ethers.ZeroAddress, /* fromAddress */
+              //     batch.receiver, /* toAddress */
+              //   );
```
ernestognw, line 65: Is there any reason for keeping this? I guess we previously kept it hoping there would be a way to check for it

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -1,222 +1,244 @@
...
-              expect(await this.token.ownerOf(tokenId)).to.be.equal(owners[tokenId]);
+            if (owners[tokenId] != ethers.ZeroAddress) {
+              expect(await this.token.ownerOf(tokenId)).to.equal(owners[tokenId]);
             }
           }
         });
 
         it('balance & voting power are set', async function () {
-          for (const account of accounts) {
+          for (const account of this.accounts) {
```
ernestognw, line 85: This test relies on `this.accounts` being equal to `await ethers.getSigners()`. I think there's a chance we (or any contributor) forget and only pass a subset of the accounts to this test.

I don't have a clear suggestion, but wanted to point out. Maybe we add a comment in the fixture?

ernestognw (approved):
LGTM just minor nits

test/token/ERC721/extensions/ERC721Consecutive.test.js
Amxx: Yes, that is what I was thinking

test/token/ERC721/extensions/ERC721Consecutive.test.js
Amxx: this loop checks many accounts. This includes account with token, accounts without token, accounts with delegation and accounts without delegation.

If the scope of the loop was reduced, it would still pass, but coverage would possibly go down.

I see two alternatives:
- loop over an explicit set of signers 
```
for (const accounts of [this.alice, this.bruce, this.chris, this.receiver]) {
```
- loop over all signers explicitelly
```
for (const accounts of await ethers.getSigners()) {
```

WDYT?

test/token/ERC721/extensions/ERC721Consecutive.test.js
ernestognw: Ok, let's drop it

test/token/ERC721/extensions/ERC721Consecutive.test.js
ernestognw: Let's do the second one, I checked and the speed difference in the tests is almost none.

ernestognw:
Looks like `await ethers.getSigners()` is not returning the same addresses in coverage mode for some reason. I couldn't replicate so to avoid blocking I rolled it back since I still consider it good to go anyways.

PR #4803, opened 2023-12-20
Refactor time helper and remove custom error helper.

Amxx:
Replace the `forward` function in `test/helpers/time.js` with
- `increaseTo`: set the timestamp/blocknumber to a fix value (same as before)
- `increaseBy`: set the timestamp/blocknumber relative to the current value


ernestognw (approved):
LGTM, I made a small change in `AccessControl.behavio.js` where we could use `increaseBy` instead of `increaseTo`.

PR #4813, opened 2023-12-30
Erc721 cleanup

RenanSouza2:
#### PR Checklist

this is more of a suggestion than anything else

biggest takeaways:
 - removal of unused uri tests in default ERC721 behaviour
 - change of ordering to avoid 'describes'with only one 'it'
 - better use of 'changeTokenBalance'

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/token/ERC721/ERC721.behavior.js
+++ b/test/token/ERC721/ERC721.behavior.js
@@ -10,7 +10,6 @@ const firstTokenId = 5042n;
 const secondTokenId = 79217n;
 const nonExistentTokenId = 13n;
 const fourthTokenId = 4n;
-const baseURI = 'https://api.example.com/v1/';
 
 const RECEIVER_MAGIC_VALUE = '0x150b7a02';
 
@@ -935,31 +934,6 @@ function shouldBehaveLikeERC721Metadata(name, symbol) {
         await expect(this.token.tokenURI(nonExistentTokenId))
           .to.be.revertedWithCustomError(this.token, 'ERC721NonexistentToken')
           .withArgs(nonExistentTokenId);
-      });
-
-      describe('base URI', function () {
-        beforeEach(function () {
-          if (!this.token.interface.hasFunction('setBaseURI')) {
-            this.skip();
-          }
-        });
-
-        it('base URI can be set', async function () {
-          await this.token.setBaseURI(baseURI);
-          expect(await this.token.baseURI()).to.equal(baseURI);
-        });
-
-        it('base URI is added as a prefix to the token URI', async function () {
-          await this.token.setBaseURI(baseURI);
-          expect(await this.token.tokenURI(firstTokenId)).to.equal(baseURI + firstTokenId.toString());
-        });
-
-        it('token URI can be changed by changing the base URI', async function () {
-          await this.token.setBaseURI(baseURI);
-          const newBaseURI = 'https://api.example.com/v2/';
-          await this.token.setBaseURI(newBaseURI);
-          expect(await this.token.tokenURI(firstTokenId)).to.equal(newBaseURI + firstTokenId.toString());
-        });
       });
     });
   });
--- a/test/token/ERC721/ERC721.behavior.js
+++ b/test/token/ERC721/ERC721.behavior.js
@@ -13,7 +13,7 @@ const fourthTokenId = 4n;
 
 const RECEIVER_MAGIC_VALUE = '0x150b7a02';
 
-function shouldBehaveLikeERC721() {
+function shouldBehaveLikeERC721(extraTxTests) {
   shouldSupportInterfaces(['ERC165', 'ERC721']);
 
   describe('with minted tokens', function () {
@@ -79,12 +79,7 @@ function shouldBehaveLikeERC721() {
           await expect(this.tx).to.changeTokenBalance(this.token, this.owner, -1);
         });
 
-        it('adjusts owners tokens by index', async function () {
-          if (!this.token.tokenOfOwnerByIndex) this.skip();
-
-          expect(await this.token.tokenOfOwnerByIndex(this.to, 0n)).to.equal(tokenId);
-          expect(await this.token.tokenOfOwnerByIndex(this.owner, 0n)).to.not.equal(tokenId);
-        });
+        extraTxTests?.different(tokenId);
       };
 
       // opts:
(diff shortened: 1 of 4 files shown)


test/token/ERC721/ERC721.behavior.js
```diff
@@ -9,17 +9,10 @@ const { RevertType } = require('../../helpers/enums');
...
-const fourthTokenId = 4n;
-const baseURI = 'https://api.example.com/v1/';
 
 const RECEIVER_MAGIC_VALUE = '0x150b7a02';
 
-function shouldBehaveLikeERC721() {
-  beforeEach(async function () {
-    const [owner, newOwner, approved, operator, other] = this.accounts;
-    Object.assign(this, { owner, newOwner, approved, operator, other });
-  });
```
ernestognw, line 21: I'm not convinced about this change. There's a transitive dependency between each test suite using `shouldBehaveLikeERC721` and their fixtures. 

In other words, the fixture of (let's say) ERC721Wrapper will need to know the values needed in `this` context (owner, newOwner, etc.). I think it's easy to lose track of unnecessary variables in fixtures this way.

Let's stick to only passing `accounts` so that this function can figure its own variables.

test/token/ERC721/ERC721.behavior.js
```diff
@@ -30,44 +23,30 @@ function shouldBehaveLikeERC721() {
     });
 
     describe('balanceOf', function () {
-      describe('when the given address owns some tokens', function () {
-        it('returns the amount of tokens owned by the given address', async function () {
-          expect(await this.token.balanceOf(this.owner)).to.equal(2n);
-        });
+      it('when the given address owns some tokens', async function () {
```
ernestognw, line 26: I'd combine both the `describe` text and the `it` text so this is clear on its intention. Otherwise, we will only know the situation the test is describing but we won't know the intended result. Although the test is pretty straight forward (`expect(...).to.equal(...)`) I think we prefer to be explicit.

```suggestion
      it('returns the amount of tokens owned by the given address when the given address owns some tokens', async function () {
```

Same for all others

test/token/ERC721/ERC721.behavior.js
```diff
@@ -82,154 +61,132 @@ function shouldBehaveLikeERC721() {
...
+      //  - extra: extra arguments
+      //  - unrestricted: operator does not need to be allowed
       const shouldTransferTokensByUsers = function (fragment, opts = {}) {
+        const extra = opts.extra ?? [];
+
         describe('when called by the owner', function () {
           beforeEach(async function () {
-            this.tx = () =>
-              this.token.connect(this.owner)[fragment](this.owner, this.to, tokenId, ...(opts.extra ?? []));
+            this.tx = await this.token.connect(this.owner)[fragment](this.owner, this.to, tokenId, ...extra);
```
ernestognw, line 92: Any reason for this preference? I feel like it should be each `it` block's responsibility to run the function. For example, it feels weird that the `transferWasSuccessful` function has tests that may pass even without running a transaction before.

I'd prefer keeping the original definition (`this.tx = () => ...`) so that `transferWasSuccessful` has more explicit `it` blocks.

test/token/ERC721/ERC721.test.js
```diff
@@ -8,7 +8,6 @@ const symbol = 'NFT';
 
 async function fixture() {
   return {
-    accounts: await ethers.getSigners(),
```
Amxx, line 11: I'm not sure its ok to remove this. `this.accounts` is used in the fixtures.

test/token/ERC721/extensions/ERC721Enumerable.test.js
Amxx, line 1: `ERC721Enumerable` is defined in the the ERC document. Unlike the other extensions, this is not an OZ design. That is historically why it was not in the `extensions` folder.

I would not move the test file to the extensions folder unless we also move the solidity file to the corresponding extensions folder ... which would be a breaking change.

For now I would keep it were it is.

test/token/ERC721/extensions/ERC721Wrapper.test.js
```diff
@@ -10,32 +10,33 @@ const tokenId = 1n;
...
-  const [owner, approved, other] = accounts;
+  const [owner, approved, other] = await ethers.getSigners();
 
   const underlying = await ethers.deployContract('$ERC721', [name, symbol]);
   await underlying.$_safeMint(owner, tokenId);
   await underlying.$_safeMint(owner, otherTokenId);
   const token = await ethers.deployContract('$ERC721Wrapper', [`Wrapped ${name}`, `W${symbol}`, underlying]);
 
-  return { accounts, owner, approved, other, underlying, token };
+  return { owner, approved, other, underlying, token };
```
Amxx, line 20: account is needed for some behavior to run.

test/token/ERC721/ERC721.behavior.js
```diff
@@ -9,901 +9,837 @@ const { RevertType } = require('../../helpers/enums');
...
-    describe('transfers', function () {
-      const tokenId = firstTokenId;
-      const data = '0x42';
+    shouldSupportInterfaces(['ERC721']);
 
+    describe('with minted tokens', function () {
       beforeEach(async function () {
-        await this.token.connect(this.owner).approve(this.approved, tokenId);
-        await this.token.connect(this.owner).setApprovalForAll(this.operator, true);
+        const [owner, approved, operator, other] = await ethers.getSigners();
```
Amxx, line 26: this is conflicting with line 18

test/token/ERC721/ERC721.behavior.js
```diff
@@ -9,901 +9,837 @@ const { RevertType } = require('../../helpers/enums');
...
-  beforeEach(async function () {
-    const [owner, newOwner, approved, operator, other] = this.accounts;
-    Object.assign(this, { owner, newOwner, approved, operator, other });
-  });
-
-  shouldSupportInterfaces(['ERC721']);
-
-  describe('with minted tokens', function () {
+function shouldBehaveLikeERC721(extraTxTests) {
+  describe('ERC712 behavior', function () {
```
Amxx, line 16: if we want to add that (I'm not sure of that), then the name should be something like "behaves like an ERC-721" or "follows the ERC-721 standard".

Amxx (commented):
I'm not convinced by this PR. It changes (breaks) a lot of things, for a debatable upside. I'm not really sure what this is fixing or improving. It is breaking things though.

Said otherwize, I think this PR tries to do to much, and is difficult to read. IMO we should address the different improvement opportunities in independent PR, that are easy to review.

RenanSouza2:
> 'm not convinced by this PR. It changes (breaks) a lot of things, for a debatable upside. I'm not really sure what this is fixing or improving. It is breaking things though.

Yes, I started to do one thing and start doing other things as well, I can revert to only removing unused URI tests

RenanSouza2:
closing this because the changes became too convoluted

PR #4824, opened 2024-01-11
Fix gas report generation

Amxx:
Gas report generaton has been broken for a while.
It looks like the json file is no longer generated by default. Adding an envvar should fix that.

WIP: multiple issues have been identified and are being worked on.


hardhat.config.js
```diff
@@ -101,30 +106,23 @@ module.exports = {
...
     exclude: ['vendor/**/*'],
   },
-  docgen: require('./docs/config'),
-};
-
-if (argv.gas) {
-  require('hardhat-gas-reporter');
-  module.exports.gasReporter = {
+  gasReporter: {
+    enabled: argv.gas,
```
Amxx, line 118: this is supposed to be true by default, but removing it causes the gas reporter to not run.

Amxx:
This should work.
@ernestognw, how do you feel about `hre.getSigners` as a replacement for `hre.ethers.getSigners` ?

ernestognw:
> This should work.
@ernestognw, how do you feel about hre.getSigners as a replacement for hre.ethers.getSigners ?

I like the idea and looks very clean, but I don't like that it's implicit (e.g. is not obvious for a contributor). I need to review first to fully understand the changes. I think it's fine but not preferrable.

Amxx:
Removed `env-contracts.js` that was not working as intended anyway. Will create an issue to consider re-implementing this mechanism.

hardhat.config.js
```diff
@@ -55,10 +55,15 @@ const argv = require('yargs/yargs')()
     },
   }).argv;
 
-require('@nomicfoundation/hardhat-toolbox');
+// "implies" doesn't work correctly. we force gas if gasReport is set
+argv.gas ||= !!argv.gasReport;
```
ernestognw, line 59: Pretty weird it's not working as expected. I'll take a look to understand what's going on here. We shouldn't merge it if we're misinterpreting how it works.

Otherwise, we may remove the `implies`.

test/sanity.test.js
```diff
@@ -3,23 +3,14 @@ const { expect } = require('chai');
 const { loadFixture, mine } = require('@nomicfoundation/hardhat-network-helpers');
 
 async function fixture() {
-  const signers = await ethers.getSigners();
-  const addresses = await Promise.all(signers.map(s => s.getAddress()));
-  return { signers, addresses };
+  return {};
 }
```
ernestognw, line 7: Shall we remove the fixture whatsoever?

Amxx: If we do so, the sanity checkpointing check fails.

Amxx: Which is to be expected, because there is no native checkpoint system anymore, we rely on fixture for that.

So basically this file checks that the fixture actually resets the blockchain state.

ernestognw: The sanity checkpointing check?

I'm not sure what that means, but, shouln't it be fine if we also remove the `Object.assign(this,...)`?

hardhat.config.js
```diff
@@ -21,12 +21,12 @@ const argv = require('yargs/yargs')()
...
       type: 'boolean',
       default: false,
+      implies: 'gasReport',
     },
     gasReport: {
       alias: 'enableGasReportPath',
       type: 'string',
-      implies: 'gas',
-      default: undefined,
+      default: 'gasReporterOutput.json',
```
Amxx, line 29: you are confsing two things:
- the json report that is always written to `gasReporterOutput.json`
- the human readable table, that is either in the terminal, or in a file.

This the the later. If we set that to `gasReporterOutput.json`, then we risk override the json file with non machine readable data.

We need to keep that undefined by default

test/sanity.test.js
ernestognw: > So basically this file checks that the fixture actually resets the blockchain state

Ah, this makes sense. Let's keep it then.

hardhat.config.js
Amxx: let me propose something

Amxx: See: 50d7935

ernestognw: This looks waay better. It was kinda obvious sorry 😅 .

I wonder how many other things we're leaving around that could be simplified.

.github/workflows/checks.yml
```diff
@@ -29,6 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       FORCE_COLOR: 1
+      CI: true # needed for "eth-gas-reporter" to produce "gasReporterOutput.json"
```
ernestognw, line 32: Let's keep a explicit reference for this.

```suggestion
      # Needed for "eth-gas-reporter" to produce a "gasReporterOutput.json"
      # according to https://github.com/cgewecke/hardhat-gas-reporter/blob/ec80a50b14967290b290fc36192a14fa09f4d3c2/README.md?plain=1#L145
      CI: true
```

ernestognw (approved):
LGTM but I see the runnings are still not producing the gas comparison. Is it because it's not merged yet?

.github/workflows/checks.yml
Amxx: line 145 really is not what we want to point to.

Amxx: Is that ok ?

```suggestion
      # Needed for "eth-gas-reporter" to produce a "gasReporterOutput.json"
      # according to https://github.com/cgewecke/eth-gas-reporter/blob/master/docs/gasReporterOutput.md
      CI: true
```

Amxx:
> LGTM but I see the runnings are still not producing the gas comparison. Is it because it's not merged yet?

Yes, gas reports produced by PR are not pushed, because nothing is going to be compared to that.
We save the reports for commits on branches that will be references for further PR.

.github/workflows/checks.yml
ernestognw: Yeah, you're right, I got it from clicking the `Copy Permalink` option. I'd like to reference the current commit so it doesn't get lost.

Let's do:

```suggestion
      # Needed for "eth-gas-reporter" to produce a "gasReporterOutput.json"
      # according to https://github.com/cgewecke/hardhat-gas-reporter/blob/ec80a50b14967290b290fc36192a14fa09f4d3c2/README.md
      CI: true
```

.github/workflows/checks.yml
```diff
@@ -29,6 +29,9 @@ jobs:
     runs-on: ubuntu-latest
     env:
       FORCE_COLOR: 1
+      # Needed for "eth-gas-reporter" to produce a "gasReporterOutput.json" as documented in
+      # https://github.com/cgewecke/eth-gas-reporter/blob/master/docs/gasReporterOutput.md
+      CI: true
```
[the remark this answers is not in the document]
Amxx: I'm not sure why you want to point to this readme. It doesn't include any mention of `process.env.CI` being needed to generate the `gasReporterOutput.json`.

Only mentions of the `gasReporterOutput.json` in this readme are about parallelisation and merging the results, which we don't care about here.

Only proper documentation I could find (outside of the source code itseleft) is in the `cgewecke/eth-gas-reporter` repo. If you want to tag a particular commit, it should be of the corresponding doc.

```suggestion
      # Needed for "eth-gas-reporter" to produce a "gasReporterOutput.json" as documented in
      # https://github.com/cgewecke/eth-gas-reporter/blob/v0.2.27/docs/gasReporterOutput.md
      CI: true
```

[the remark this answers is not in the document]
ernestognw: I'm sorry, I've been carrying the incorrect URL so far.
Yes, that's the only mention I found as well. Let's keep it. Also agree with the version tag.

ernestognw (approved):

PR #4829, opened 2024-01-11
Add missing return value names in IAccessManager

bxmmm1:
Copied & pasted the values from AccessManager.sol so that they are in sync

#### PR Checklist

- [x] Tests
- [x] Documentation


ernestognw:
Just added a label to ignore the changeset since it's not required imo. Also deleted the empty changeset you added.

I'm fine adding this change but let's wait on @Amxx

ernestognw (approved):

Amxx (approved):

PR #4832, opened 2024-01-12
Remove `hardhat-foundry` and check harnesses compilation in CI

ernestognw:
Before merging #4781, I tried running the FV in master and it resulted in a compilation error after https://github.com/OpenZeppelin/openzeppelin-contracts/commit/25c416d01c8d9c7b9b9dd3bdd0994e10fb26daa3

I fixed the issues and renamed variables accordingly. I'm also adding a check for it in the CI to avoid happening again.

#### PR Checklist

- [x] Tests
- [ ] Documentation


certora/specs/EnumerableMap.spec
```diff
@@ -69,13 +69,13 @@ invariant atUniqueness(uint256 index1, uint256 index2)
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
 │ Invariant: index <> value relationship is consistent                                                                │
 │                                                                                                                     │
-│ Note that the two consistencyXxx invariants, put together, prove that at_ and _indexOf are inverse of one another.  │
+│ Note the two consistencyXxx invariants, put together, prove that at_ and _positionOf are inverse of one another.    │
```
Amxx, line 72: We should either do 
- "Note that ..." as in "Please note that ..."
- "Note: ..."

But the current formulation is not correct

certora/specs/EnumerableSet.spec
```diff
@@ -52,13 +52,13 @@ invariant atUniqueness(uint256 index1, uint256 index2)
 ┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
 │ Invariant: index <> key relationship is consistent                                                                  │
 │                                                                                                                     │
-│ Note that the two consistencyXxx invariants, put together, prove that at_ and _indexOf are inverse of one another.  │
+│ Note the two consistencyXxx invariants, put together, prove that at_ and _positionOf are inverse of one another.    │
```
Amxx, line 55: same as https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4832/files#r1451019922

package.json
```diff
@@ -9,6 +9,7 @@
   ],
   "scripts": {
     "compile": "hardhat compile",
+    "compile:harnesses": "env SRC=./certora/harnesses FOUNDRY=false npm run compile",
```
Amxx, line 12: Is `FOUNDRY=false` absolutelly necessary here ? Is the foundry package actually preventing compilation?

ernestognw: Yes, otherwise we get this error:

```shell
> openzeppelin-solidity@5.0.1 compile:harnesses
> env SRC=./certora/harnesses npm run compile


> openzeppelin-solidity@5.0.1 compile
> hardhat compile

Error in plugin hardhat-foundry: User-configured sources path (./certora/harnesses) doesn't match path configured in foundry (contracts)
```

It's just a path mismatch or something similar. The fix would be to make the `foundry.toml` to match I think, but that's more work.

hardhat.config.js
Amxx, line 1: We may want to setup the config so that any change to `src` automatically disables foundry

certora/specs/EnumerableMap.spec
```diff
@@ -66,13 +66,13 @@ invariant atUniqueness(uint256 index1, uint256 index2)
...
 
 /*
-┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
-│ Invariant: index <> value relationship is consistent                                                                │
-│                                                                                                                     │
-│ Note the two consistencyXxx invariants, put together, prove that at_ and _positionOf are inverse of one another.    │
-│ This proves that we have a bijection between indices (the enumerability part) and keys (the entries that are set    │
-│ and removed from the EnumerableMap).                                                                                │
-└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
```
Amxx, line 69: these lines are designed to be 119 char long so that they show correctly in a 120 column terminal or in a IDE that sets line wrapping at 120 chars. We should avoid making them longer.

120 is also what we have set in .prettierrc

hardhat.config.js
ernestognw: It should be something like this

```diff
diff --git a/hardhat.config.js b/hardhat.config.js
index 70d22cb23..2c57e77ca 100644
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -51,10 +51,12 @@ const argv = require('yargs/yargs')()
     src: {
       alias: 'source',
       type: 'string',
-      default: 'contracts',
     },
   }).argv;
 
+// If a source path is provided, it will conflict with the default path set in `foundry.toml`
+if (argv.src) argv.foundry = false;
+
 require('@nomicfoundation/hardhat-chai-matchers');
 require('@nomicfoundation/hardhat-ethers');
 require('hardhat-exposed');
```

However, I'm not convinced because it seems that we just can't use Foundry with another directory. Although we don't need foundry now, I think the best choice is [defining a profile](https://book.getfoundry.sh/reference/config/overview?highlight=profile) for harnesses in `foundry.toml` and compile with forge instead:

```
diff --git a/.github/workflows/checks.yml b/.github/workflows/checks.yml
index b4348e91e..00502fc34 100644
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -108,7 +108,7 @@ jobs:
       - name: Compile harnesses
         run: |
           make -C certora apply
-          npm run compile:harnesses
+          FOUNDRY_PROFILE=harnesses forge compile
 
   slither:
     runs-on: ubuntu-latest
diff --git a/foundry.toml b/foundry.toml
index 859f210cf..423871b5c 100644
--- a/foundry.toml
+++ b/foundry.toml
@@ -5,6 +5,9 @@ libs = ['node_modules', 'lib']
 test = 'test'
 cache_path  = 'cache_forge'
 
+[profile.harnesses]
+src = 'certora/harnesses'
+
 [fuzz]
 runs = 10000
 max_test_rejects = 150000
```

What do you think?

certora/specs/EnumerableMap.spec
ernestognw: lmao I got this so wrong 😂

hardhat.config.js
Amxx: > I think the best choice is [defining a profile](https://book.getfoundry.sh/reference/config/overview?highlight=profile) for harnesses in foundry.toml and compile with forge instead

We already have two compilation flows:
- one with hardhat 
- one that is included in certora
That is already one too many IMO.

My issue is that each flow needs to be configured separatelly. If/when we decide to update the version of solidity we use (toward 6.0 for example), we will have to change that in the `hardhat.config.js`. Ideally that is the only place where we'll have to update it.

Having a foundry compilation flow sounds like something we don't really need. I'm not sure what it will bring that our existing flow can't already do.

Amxx: Right now 

```diff
diff --git a/hardhat.config.js b/hardhat.config.js
index 70d22cb23..2c57e77ca 100644
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -51,10 +51,12 @@ const argv = require('yargs/yargs')()
     src: {
       alias: 'source',
       type: 'string',
-      default: 'contracts',
     },
   }).argv;
 
+// If a source path is provided, it will conflict with the default path set in `foundry.toml`
+if (argv.src) argv.foundry = false;
+
 require('@nomicfoundation/hardhat-chai-matchers');
 require('@nomicfoundation/hardhat-ethers');
 require('hardhat-exposed');
```

would be my favorite option.

Amxx: @ernestognw please check [6b1d59c](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4832/commits/6b1d59c888729ff034dfe92524e1cf2eb5d8a5f4) and let me know what you think.

ernestognw: Ok, I like the approach. I added a few commits because I think [this change](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4832/commits/7b4b82de94ffb3cb492df8c1cb89a38bbeb1d6b5#diff-cf4ef7c51dc9f81cad1d504da0d1c3a3437ac7b7d1374ee7127886cf1d1a5092R74) was not ok. But I may lack of context.

I'm looking at what's going on with the failed upgradeable tests

ernestognw: Looking good now

hardhat.config.js
```diff
@@ -115,7 +114,7 @@ module.exports = {
   },
   networks: {
     hardhat: {
-      allowUnlimitedContractSize,
+      allowUnlimitedContractSize: !withOptimizations,
```
Amxx: any reason for this rollback ? IMO, if the coverage is enabled, then optimizations should be on, but contract size should not be limitted. (coverage does increasse the size)

Amxx:
I first wanted to use conflicts, but then decided against it because it caused issues with SRC.

I review you proposal, but it doesn't work. If I do `FOUNDRY=true yarn test` or even `FOUNDRY=false yarn test`, then both `foundry` and `src` are defined (first one is a boolean, second one is a string) ... which causes conflict to fail:

```
Les arguments foundry et src sont mutuellement exclusifs
```

as nice as `yargs`'s "conflict" is, it doesn't work for us.

hardhat.config.js
Amxx: > But I may lack of context.

I think that is the context you are missing:  https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4832#issuecomment-1893446701

Amxx:
It appears that both `checks/tests`, and `checks/tests-foundry` work without the foundry hardhat package. I thus propose removing it, which removes the `hardhat.config.js` complexity

Amxx:
It appears that both checks/tests, and checks/tests-foundry work without the foundry hardhat package.

The package still has a usage though: it overrides the `TASK_COMPILE_GET_REMAPPINGS` hardhat task so the remappings.txt file is considered. This is used by the upgradeable tests.

I'm proposing we remove the hardhat-foundry, because of the complexity is adds to handling configs in hardhat.config.js ... and add a dedicated script in the `hardhat` folder to maintain the `remappings.txt` support.

ernestognw:
> It appears that both checks/tests, and checks/tests-foundry work without the foundry hardhat package.
> 
> The package still has a usage though: it overrides the `TASK_COMPILE_GET_REMAPPINGS` hardhat task so the remappings.txt file is considered. This is used by the upgradeable tests.
> 
> I'm proposing we remove the hardhat-foundry, because of the complexity is adds to handling configs in hardhat.config.js ... and add a dedicated script in the `hardhat` folder to maintain the `remappings.txt` support.

After reviewing it looks good. The downside of this is that remappings are inherited from other submodules and also include some defaults. Now we shouldn't have access to those but aren't required. In any case, we can always:

- Read remappings from `forge remappings` conditionally if `forge` is available
- Adapt our remappings.txt

LGTM

PR #4834, opened 2024-01-13
Run gas-compare action on push to master 

ernestognw:
After #4824, the #4822 PR was synced. However, it is still not producing the gas comparison because the report from `master` is not executed given `github.base_ref` is empty when a commit triggered the run. In that case, [the `master` report is just skipped.](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/7508535099/job/20445181704)

With this change, we'll be running the comparison only in commits to master and pull requests targeting to master

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -45,7 +45,7 @@ jobs:
         run: npm run test:generation
       - name: Compare gas costs
         uses: ./.github/actions/gas-compare
-        if: github.base_ref == 'master'
+        if: github.event_name == 'pull_request' && github.base_ref == 'master' || github.event_name == 'push' && github.ref_name == 'master'
         with:
           token: ${{ github.token }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         run: npm run test:generation
       - name: Compare gas costs
         uses: ./.github/actions/gas-compare
-        if: github.base_ref == 'master'
+        if: (github.event_name == 'pull_request' && github.base_ref == 'master') || (github.event_name == 'push' && github.ref_name == 'master')
```
Amxx, line 48: Do we want the gas comparison to only work for master ?

If we ever do a `v6-next` branch (or similar) I guess we will want the gas comparaison to run.

I'd just remove that line, and have the action always run.
```suggestion
```

Amxx: Note: this change (either removing the if or making it support push to master) should also be added to the "compare storage layout" action on line 74

ernestognw: For the record, we discussed in a meeting that removing the check will allow comparisons to other branches, but it's likely that comparisons after major changes may fail (e.g. comparing `v6-next` vs. `master`). Not a problem for now.

Amxx (approved):

PR #4835, opened 2024-01-13
Lower fuzz runs to 5000

ernestognw:
Right now the foundry tests are the job that takes the longest to run (~9 min). But, after the ethers test migration we made the unit tests way faster (~4 min). Now that the fuzz tests are taking longer than the unit tests, I think it's safe to lower the runs a bit.

I couldn't find any concrete reason to use 10_000 and I think we'd benefit from faster CI jobs.

Also, I made the 5_000 runs to only apply in the CI so our dev experience is better when running locally (with 256 runs)

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -86,8 +86,6 @@ jobs:
         uses: ./.github/actions/setup
       - name: Run tests
         run: forge test -vv
-        env:
-          FOUNDRY_PROFILE: CI
 
   coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -86,6 +86,8 @@ jobs:
         uses: ./.github/actions/setup
       - name: Run tests
         run: forge test -vv
+        env:
+          FOUNDRY_PROFILE: CI
 
   coverage:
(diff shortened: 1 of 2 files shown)


Amxx:
First I'd like to understand what the default value is. If the default value for foundry is 5000, then do we need to explicitly set the value. If not, what is the default value, is it 10000? It is unclear to me what value is used if we are not using the profile that you define.

Moreover, I'm really not a fan of having different tests configurations in local and in CI. This might affect the reproducibility of issues that may arise is CI.

I think we should do things way simpler:
- not have a `FOUNDRY_PROFILE` envvar
- keep a single `[fuzz]` profile for all fuzing tests (local and in the CI)
- remove the `runs =` completelly, to use the default value (if it is 5000), of keep it, with a comment that says that the default is `X` (only ok IMO is x < 5000).

Amxx:
> The spirit of the change is to keep the CI running with 5000 runs but use the default locally

This means using different settings in CI and locally, with I really don't like.

ernestognw:
Right, it's not an issue imo but there's no real strong reason to have a different CI config either.

Let's just lower the fuzz runs.

Amxx (approved):

PR #4837, opened 2024-01-17
Remove obsolete ERC4626 `mint` NatSpec

dglowinski:
Removes NatSpec for ERC4626 `mint` function, which says that minting is allowed when the price of a share is zero. With the introduction of virtual shares and assets mitigation mechanism for the inflation attack, this is no longer true, because a price of a share can no longer be zero. 

Zero price of a share in this context means that shares amount requested in `mint` would be converted to zero assets by `previewMint`. It would only be possible if  `totalSupply` was greater than zero, and `totalAssets` was zero in the previous version of the code (given that `previewMint` function rounds up):
```
function _convertToAssets(uint256 shares, Math.Rounding rounding) internal view virtual returns (uint256) {
        return
            (supply == 0) ? _initialConvertToAssets(shares, rounding) : shares.mulDiv(totalAssets(), supply, rounding);
}
```

The new version of the function unconditionally adds 1 virtual asset to the numerator in the conversion, making 'zero price of a share' impossible, given upward rounding.
```
function _convertToAssets(uint256 shares, Math.Rounding rounding) internal view virtual returns (uint256) {
     return shares.mulDiv(totalAssets() + 1, totalSupply() + 10 ** _decimalsOffset(), rounding);
}
```

The original comment was introduced in: [Improve some NatSpec and revert reasons #3809](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3809)

It was made obsolete in [ERC4626 inflation attack mitigation #3979](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3979)

Where the test of the original behaviour [was removed](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3979/files#diff-d8baad4431e5aa30ca1424007f3d81c10c0824d13f1843b195d2334c1e0b7048L244-L261)

#### PR Checklist

- [ ] Tests
- [x] Documentation


Amxx (approved):
LGTM. Needs @ernestognw approval.

ernestognw (approved):
Looks good, thanks for clearly documenting @dglowinski!

PR #4839, opened 2024-01-17
Add a Math.inv function that inverse a number in Z/nZ

Amxx:
This is a very common function use in finite fields such as the ones that power ECDSA curves.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/math/Math.sol
```diff
@@ -218,6 +218,33 @@ library Math {
         return result;
     }
 
+    /**
+     * @notice Calculate the inverse of a number in Z/nZ.
+     *
+     * If p is a prime, then we Z/nZ is a field commonly noted Fp. In that case all elements are inversible, expect 0.
+     * If p is not a prime, then Z/nZ is not a field, and some elements might not be inversible.
+     *
+     * When trying to inverse a value that is not invesible, 0 is returned.
```
Amxx, line 227:
```suggestion
     * When trying to inverse a value that is not invertible, 0 is returned.
```

Amxx:
Some ressources:
- [A solidity implementation](https://github.com/tdrerup/elliptic-curve-solidity/blob/master/contracts/curves/EllipticCurve.sol#L32-L55) that uses signed numbers instead of addmod and mulmod. Its solidity 0.5.0, and porting it nativelly in 0.8.0 causes overflows that showed up when doing fuzzing.
- [Noble's implementation in typescript](https://github.com/paulmillr/noble-curves/blob/main/src/abstract/modular.ts#L54-L76): they use bigint so they don't have to do the modulo at every step. The gcd = 1 check comes from there. 
- [Euclidean algorithm on Wikipedia](https://en.wikipedia.org/wiki/Euclidean_algorithm)

ernestognw:
This is my current understanding:

- `Z` is the set of natural numbers (i.e. `{… , ⁻2, ⁻1, 0, 1, 2, …}`)
- `Z/nZ` is the set of natural numbers numbers `mod n` (i.e. `{0, 1, . . . , n − 1}`)

What we're providing with this function is the [modular multiplicative inverse](https://en.wikipedia.org/wiki/Modular_multiplicative_inverse) of an `uint256 a` in a field defined by `n` (it's `uint256 p` in the function implementation currently). 

The multiplicative inverse of `a` is an integer `x` such that the product `ax` is congruent. This is that `ax ≡ 1 (mod n)`, and the way to calculate such thing is using the Euclidean algorithm (runs in `O(log2(n))`) to find the greatest common divisor (`gcd` for short), and, (if `gcd == 1`), then the multiplicative `x` can be found from the Bezout's identity that states:

```
ax + ny = gcd = 1

ax + ny = 1

ax - 1 = (-y)m

ax ≡ 1 (mod n) # We've found `x`
```

If the `gcd` of `a, n ` (i.e. `gcd(a,n)`) is 1, it means they're relative primes. So, for a field `n` to have multiplicative inverses for every value in the field, then `n` should be prime. This is because if `n` is prime, then the amount of relative primes `< n` is `n - 1` (it excludes only the 0).

contracts/utils/math/Math.sol
```diff
@@ -218,6 +218,31 @@ library Math {
...
 
+    /**
+     * @notice Calculate the inverse of a number in Z/nZ.
+     *
+     * If p is a prime, then we Z/nZ is a field commonly noted Fp. In that case all elements are inversible, expect 0.
+     * If p is not a prime, then Z/nZ is not a field, and some elements might not be inversible.
+     *
+     * If the input values is not inversible, 0 is returned.
+     */
+    function inv(uint256 a, uint256 p) internal pure returns (uint256) {
```
ernestognw, line 229: I think we should call this function as `invMod`. Using `inv` is too generic imo

contracts/utils/math/Math.sol
```diff
@@ -218,6 +218,31 @@ library Math {
         return result;
     }
 
+    /**
+     * @notice Calculate the inverse of a number in Z/nZ.
+     *
+     * If p is a prime, then we Z/nZ is a field commonly noted Fp. In that case all elements are inversible, expect 0.
+     * If p is not a prime, then Z/nZ is not a field, and some elements might not be inversible.
+     *
+     * If the input values is not inversible, 0 is returned.
```
ernestognw, line 227: I'd suggest using `n` all over:

```suggestion
     * If n is a prime, then Z/nZ is a field commonly noted Fn. In that case all elements are inversible, expect 0.
     * If n is not a prime, then Z/nZ is not a field, and some elements might not be inversible.
     *
     * If the input values is not inversible, 0 is returned.
```

Is there any reason for preferring `p`?

contracts/utils/math/Math.sol
```diff
@@ -218,6 +218,31 @@ library Math {
         return result;
     }
 
+    /**
+     * @notice Calculate the inverse of a number in Z/nZ.
+     *
+     * If p is a prime, then we Z/nZ is a field commonly noted Fp. In that case all elements are inversible, expect 0.
+     * If p is not a prime, then Z/nZ is not a field, and some elements might not be inversible.
```
ernestognw, line 225: If I understood correctly, this is a field, but it's just not a finite field. If I'm wrong, would you mind explaining the difference?

contracts/utils/math/Math.sol
Amxx: Usually, Z/nZ is the group structure defined by the equivalence classes resulting from the modulo operations (and the common addition and multiplication operation). We often use numbers between 0 and n-1 as the canonical elements of each equivalent class.

It turns that that if the characterictic number for this group structure is a prime, then the group is actually a field (all non-zero element are invertible). This is ussually noted as Fp.

So yes, Fp is Z/nZ when n is prime, I've never seen Fn written ... It would be confused with "Function". 
I've seen Z/pZ written, but writting it that way implied the restriction that p is prime.

Amxx: relevant link: https://en.wikipedia.org/wiki/Finite_field

contracts/utils/math/Math.sol
```diff
@@ -218,6 +218,62 @@ library Math {
         return result;
     }
 
+    /**
+     * @notice Calculate the modular multiplicative inverse of a number in Z/nZ.
```
ernestognw, line 222: These don't show up in the docs. I see there are many `@notice`s, so perhaps makes sense to change them all.

```suggestion
     * @dev Calculate the modular multiplicative inverse of a number in Z/nZ.
```

ernestognw (approved):
LGTM

contracts/utils/math/Math.sol
ernestognw: Ok, it's clear to me now.

The current version looks good, but feel free to add the "commonly noted as Fp" if you think it's useful.

PR #4842, opened 2024-01-18
Bound lookup in arrays with duplicate

### LINKED ISSUE #4744: Add a findLowerBound function to the Arrays library
Has request by @alanarvelo during DevConnect, we should add a lower lookup function to `Arrays.sol`. This is already available in `Checkpoints` but is missing from `Arrays`.

Amxx:
This PR does the following:
- Deprecate `Array.findUpperBound` which does not match the "upper bound" function in other standard libraries
- Add `Array.lowerBound` and `Array.upperBound` that perform lower and upper bound lookup in storage arrays (with support for duplicates).
- Add `Array.lowerBoundMemory` and `Array.upperBoundMemory` that perform lower and upper bound lookup in memory arrays (with support for duplicates).

Fixes #4744
Fixes LIB-1182
Fixes LIB-1188

#### PR Checklist

- [x] Tests
- [x] Documentation


.changeset/thick-pumpkins-report.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Arrays`: add new functions `lowerBound`, `upperBound`, lowerBoundMemory`and`upperBoundMemory` for lookups in sorted arrays with potential duplicates.
```
Amxx, line 5:
```suggestion
`Arrays`: add new functions `lowerBound`, `upperBound`, `lowerBoundMemory` and `upperBoundMemory` for lookups in sorted arrays with potential duplicates.
```

contracts/utils/Arrays.sol
```diff
@@ -20,6 +20,9 @@ library Arrays {
      *
      * `array` is expected to be sorted in ascending order, and to contain no
      * repeated elements.
+     *
+     * Deprecated in favor of `lowerBound` and `findUpperBound`. Note that this actually implements a lower bound
```
ernestognw, line 24: I can't make a suggestion, but we should change the NatSpec of this function to reflect its actual behavior. Right now it's saying it does one thing and it "notes" that it does another.

What about?

```
* @dev Searches a sorted `array` and returns the last index that contains
* a value greater or equal to `element`. If no such index exists (i.e. all
* values in the array are strictly less than `element`), the array length is
* returned. Time complexity O(log n).
```

And then this note:

```suggestion
     * IMPORTANT: Deprecated. This implementation behaves as {lowerBound}, which should be used instead.
```

test/utils/Arrays.test.js
```diff
@@ -4,22 +4,22 @@ const { loadFixture } = require('@nomicfoundation/hardhat-network-helpers');
 
 const { randomArray, generators } = require('../helpers/random');
 
-// See https://en.cppreference.com/w/cpp/algorithm/ranges/lower_bound
+// See https://en.cppreference.com/w/cpp/algorithm/lower_bound
 const lowerBound = (array, value) => {
   const i = array.findIndex(element => value <= element);
```
ernestognw, line 9: According to the C++ Reference:

> Returns an iterator pointing to the first element in the range [first, last) such that element < value (or comp(element, value)) is false, (i.e. that is greater than or equal to value), or last if no such element is found.

What I see in this function is that `findIndex` returns the first element that satisfies the `value <= element` condition, but according to C++'s, shouldn't it be `value > element`? Seems off to me.

```suggestion
  const i = array.findIndex(element => value > element);
```

contracts/utils/Arrays.sol
```diff
@@ -49,6 +52,150 @@ library Arrays {
...
 
+    /**
+     * @dev Searches a sorted `array` and returns the first index that contains
+     * a value greater or equal to `element`. If no such index exists (i.e. all
+     * values in the array are strictly less than `element`), the array length is
+     * returned. Time complexity O(log n).
+     *
+     * `array` is expected to be sorted in ascending order,
+     *
+     * See https://en.cppreference.com/w/cpp/algorithm/ranges/lower_bound
```
ernestognw, line 63:
```suggestion
     * @dev Searches an `array` sorted in ascending order and returns the first
     * index that contains a value greater or equal than `element`. If no such index
     * exists (i.e. all values in the array are strictly less than `element`), the array
     * length is returned. Time complexity O(log n).
     *
     * See C++'s https://en.cppreference.com/w/cpp/algorithm/lower_bound[lower_bound]
```

contracts/utils/Arrays.sol
```diff
@@ -49,6 +52,150 @@ library Arrays {
...
+
+    /**
+     * @dev Searches a sorted `array` and returns the first index that contains
+     * a value strictly greater to `element`. If no such index exists (i.e. all
+     * values in the array are strictly less than `element`), the array length is
+     * returned. Time complexity O(log n).
+     *
+     * `array` is expected to be sorted in ascending order,
+     *
+     * See https://en.cppreference.com/w/cpp/algorithm/upper_bound
```
ernestognw, line 99:
```suggestion
     * @dev Searches an `array` sorted in ascending order and returns the first
     * index that contains a value strictly greater than `element`. If no such index
     * exists (i.e. all values in the array are strictly less than `element`), the array
     * length is returned. Time complexity O(log n).
     *
     * See C++'s https://en.cppreference.com/w/cpp/algorithm/upper_bound[upper_bound]
```

ernestognw (commented):
Will continue reviewing

test/utils/Arrays.test.js
ernestognw: Solving, I see this is wrong because it's expecting the opposite of `value > element`

contracts/utils/Arrays.sol
```diff
@@ -49,6 +52,150 @@ library Arrays {
...
```
ernestognw, line 171: I think this would be simpler

```suggestion
     * @dev Same as {upperBound}, but with an array in memory.
```

contracts/utils/Arrays.sol
```diff
@@ -49,6 +52,150 @@ library Arrays {
...
+
+    /**
+     * @dev Searches a sorted `array` and returns the first index that contains
+     * a value greater or equal to `element`. If no such index exists (i.e. all
+     * values in the array are strictly less than `element`), the array length is
+     * returned. Time complexity O(log n).
+     *
+     * `array` is expected to be sorted in ascending order,
+     *
+     * See https://en.cppreference.com/w/cpp/algorithm/ranges/lower_bound
```
ernestognw, line 135:
```suggestion
     * @dev Same as {lowerBound}, but with an array in memory.
```

test/utils/Arrays.test.js
```diff
@@ -82,40 +82,52 @@ describe('Arrays', function () {
...
 
         for (const [name, input] of Object.entries(tests)) {
-          it(name, async function () {
-            // findUpperBound does not support duplicated
-            if (hasDuplicates(array)) this.skip();
-            expect(await this.mock.findUpperBound(input)).to.equal(lowerBound(array, input));
+          describe(name, function () {
+            it('[deprecated] findUpperBound', async function () {
+              // findUpperBound does not support duplicated
+              if (hasDuplicates(array)) this.skip();
```
ernestognw, line 88: We should test for duplicates for explicitness.

I'll try to do it.

ernestognw (approved):
LGTM

test/utils/Arrays.test.js
```diff
@@ -85,8 +85,11 @@ describe('Arrays', function () {
           describe(name, function () {
             it('[deprecated] findUpperBound', async function () {
               // findUpperBound does not support duplicated
-              if (hasDuplicates(array)) this.skip();
-              expect(await this.mock.findUpperBound(input)).to.be.equal(lowerBound(array, input));
+              if (hasDuplicates(array)) {
+                expect(await this.mock.findUpperBound(input)).to.be.equal(upperBound(array, input) - 1);
```
Amxx, line 89: If there is more than one duplicate, this will probably fail.

PR #4853, opened 2024-01-23
Fuzz Base64 and Base64URL

ernestognw:
Following #4692 

#### PR Checklist

- [x] Tests
- [ ] Documentation


test/utils/Base64.t.sol
```diff
@@ -6,27 +6,33 @@ import {Test} from "forge-std/Test.sol";
...
+    function _removePadding(string memory input) internal pure returns (string memory) {
+        bytes memory bytesInput = bytes(input);
+        uint256 length = bytesInput.length;
+        if (length == 0) return input;
+
+        uint256 padding = 0;
+
+        while (bytesInput[length - padding - 1] == 0x3d) {
+            padding++;
+        }
```
Amxx, line 29: The fact that the while does not overflow only depends on the string not boing only paddings.

Maybe we should use a for to explicit the end.
```suggestion
        for (padding=0; padding<length; ++padding) {
            if (bytesInput[length - padding - 1] != 0x3d) break;
        }
```

Amxx: That way we should be able to remove the `if (length == 0) return input;` check

Amxx (approved):

PR #4859, opened 2024-01-28
Procedurally generate Arrays.sol

RenanSouza2:
This PR creates scripts to generate the Arrays.sol file,

It also includes tests for unsafeMemoryAccess functions

#### PR Checklist

- [X] Tests
- [ ] Documentation


RenanSouza2:
Hey, 

Are you interested in adding a function for bytes32 memory array?

There are lots of simplifications that can be done if there are the same functions for both array and storage

Amxx:
Hello @RenanSouza2 

We have multiple PR targetting Array.sol. #4846, #4842. I'm not sure we need to generate this file procedurally. I'd delay that decision until after the above PR are processed.

ernestognw (commented):
I made a few adjustments after the other Arrays PRs got merged.
Right now it looks good to me but we might want to discuss how the generation script is ordered.

WDYT @Amxx ?

Amxx (approved):

ernestognw:
I agree with minimizing the diff. So LGTM as well.
However, I do think the previous ordering is better when reading the generation script. No need to update it now since I don't think the ordering is strictly important, but wanted to point it out.

PR #4863, opened 2024-01-30
Imported ForwardRequest struct correctly

iamgauravpant:
ForwardRequest struct resides in eip712-types.js file , fixed that import

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx:
Hello @iamgauravpant 

`ForwardRequest` is indeed declared in `test/helpers/eip712-types.js`, but it is re-exported in `test/helpers/eip712.js`. This is so that we can import them both at once, and limit the number of different imports.

PR #4868, opened 2024-02-01
Remove deprecated Truffle code

Amxx:
This was missed. It is no longer relevant since we transitionned away from Truffle.


ernestognw (approved):
I've seen it before but never opened a PR, LGTM

PR #4870, opened 2024-02-02
Add VestingWalletWithCliff

### LINKED ISSUE #4701: Lack of inclusion of a "Vesting Cliff" feature for VestingWallet
From the 5.0 release audit, this recommendation was made and considered for next versions.

> N-18 Lack of Inclusion of a "Vesting Cliff"
>
> A "vesting cliff" is a prevalent feature in vesting schedules. It defines a date before which no tokens can be released, despite the linear progression of the vesting schedule prior to that date. Even though projects can extend the contract to include this functionality, given its prevalence in vesting schedules, it would be beneficial to incorporate it into the default contract. This addition would help reduce potential implementation errors and enhance the utility of the default contract.
Consider adding a cliffTimestamp input to the constructor. This timestamp could be used in a cliff() view, whose value should not exceed the end() time. In the _vestingSchedule method, if the timestamp is earlier than the cliff(), the totalAllocation should be set to 0.
>
> Update: Acknowledged, will resolve. The OpenZeppelin Contracts team stated:
>> This variant will be considered for future development. We consider it an easy
implementation to be built on top of the VestingWallet.

Amxx:
... an extension of VestingWallet with an added cliff in the release schedule

Fixes #4701
Fixes LIB-1190

#### PR Checklist

- [x] Tests
- [ ] Documentation


Amxx:
Following discussion, moving to a model where the extension is an abstract contract that needs to be inherited/compiled.

contracts/finance/VestingWalletWithCliff.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @dev Virtual implementation of the vesting formula. This returns the amount vested, as a function of time, for
+     * an asset given its total historical allocation.
+     */
+    function _vestingSchedule(
+        uint256 totalAllocation,
+        uint64 timestamp
+    ) internal view virtual override returns (uint256) {
+        return timestamp < cliff() ? 0 : super._vestingSchedule(totalAllocation, timestamp);
```
ernestognw, line 44: This is not always calling `super`, and it's part of the discussion we've had before in regards to always call `super`.

In this case it harmless, so it doesn't make sense to call super every time since the ternary operator is extending the original branching in `_vestingWallet`, making it equivalent to:

```solidity
function _vestingSchedule(uint256 totalAllocation, uint64 timestamp) internal view virtual returns (uint256) {
    if (timestamp < cliff() || timestamp < start()) {
        return 0;
    } else if (timestamp >= end()) {
        return totalAllocation;
    } else {
        return (totalAllocation * (timestamp - start())) / duration();
    }
}
```

I think it doesn't make sense to call `super` every time here, but if `_vestingWallet` is overridden by an user and it includes a check or an state update, it might be an issue.

I don't have a strong opinion. But I'm leaning towards not calling `super`. What do you think?

contracts/finance/VestingWalletWithCliff.sol
```diff
@@ -0,0 +1,46 @@
...
+    uint64 private immutable _cliff;
+
+    error InvalidCliffDuration(uint64 cliffSeconds, uint64 durationSeconds);
+
+    /**
+     * @dev Sets the sender as the initial owner, the beneficiary as the pending owner, the start timestamp, the
+     * vesting duration and the duration of the cliff of the vesting wallet.
+     */
+    constructor(uint64 cliffSeconds) {
+        if (cliffSeconds > duration()) {
```
ernestognw, line 23: I don't think this makes a practical difference but imo it makes sense to exclude the case where `cliffSecodns == duration()`

```suggestion
        if (cliffSeconds >= duration()) {
```

contracts/finance/VestingWalletWithCliff.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    uint64 private immutable _cliff;
+
+    error InvalidCliffDuration(uint64 cliffSeconds, uint64 durationSeconds);
+
+    /**
+     * @dev Sets the sender as the initial owner, the beneficiary as the pending owner, the start timestamp, the
+     * vesting duration and the duration of the cliff of the vesting wallet.
+     */
+    constructor(uint64 cliffSeconds) {
```
ernestognw, line 22: I see `durationSeconds` is used in `VestingWallet`, but there it's straightforward because the parameter is the actual duration. In the case of `cliffSeconds`, it's added over the `start`, which may seem confusing.

We can:

- Rename `cliffSeconds` to `cliffDurationSeconds`
- Rename the `cliff()` function to `cliffTimestamp`

Leaning towards number 1:

```suggestion
    constructor(uint64 cliffDurationSeconds) {
```

contracts/finance/VestingWalletWithCliff.sol
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * @dev Extension of {VestingWallet} that adds a cliff to the vesting schedule.
+ */
+abstract contract VestingWalletWithCliff is VestingWallet {
+    using SafeCast for *;
+
+    uint64 private immutable _cliff;
+
+    error InvalidCliffDuration(uint64 cliffSeconds, uint64 durationSeconds);
```
ernestognw, line 16:
```suggestion
    /// @dev The specified cliff duration is larger than the vesting duration.
    error InvalidCliffDuration(uint64 cliffSeconds, uint64 durationSeconds);
```

test/finance/VestingWalletWithCliff.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+  return { mock, duration, start, beneficiary, cliff, schedule, vestingFn, env };
+}
+
+describe('VestingWallet', function () {
+  beforeEach(async function () {
+    Object.assign(this, await loadFixture(fixture));
+  });
+
+  it('rejects zero address for beneficiary', async function () {
```
ernestognw, line 74:
```suggestion
  it('rejects a larger cliff than vesting duration', async function () {
```

contracts/finance/VestingWalletWithCliff.sol
Amxx: I think not caller super at all is dangerous. Lets discuss that next week.

contracts/finance/VestingWalletWithCliff.sol
Amxx: IMO, if both are equal, that is ok. No real reason to reject.

There is a usecase for the "basic" VestingWallet where start = end (duration = 0) → that release everything at once at some point in the future.

If someone creates a factory for `VestingWalletCliff`, and someone what to deploy an instance that release everything at one, then we would have to do start = cliff = end (duration = 0, cliffSeconds = 0) ... and I think that is ok.

ernestognw (dismissed):
I've been thinking a lot about the `_vestingSchedule` not calling super, and even tried to think about implementing the cliff in terms of the `_vestingSchedule`, but also realized that allows an attacker to extend the cliff undefinitely if they have enough resources.

So far I think this is the best approach, and also, if the "cliff" also means ignoring the scheduling function at all, then it makes sense to return 0 directly.

I added a note about this to `_vestingWallet`, since in the end, this is a threat if they ever have to dissambiguate the inheritance graph, so ideally they should take a look at this and decide whether or not they're okay with the cliff implementation.

ernestognw (approved):

Amxx:
Its right here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/finance/VestingWalletCliff.sol

Amxx:
@urataps This PR was not released yet. It will be in 5.1

PR #4872, opened 2024-02-02
Replace custom errors with native panic codes in DoubleEndedQueue

### LINKED ISSUE #4867: Idea: use panic codes instead of custom errors
There are some cases where we use custom error, but Panic codes may be more relevant/standard.

Some example (not exaustive):
- In `Math.mulDiv`, if the `prod1 > denominator`, we should probably Panic with `0x11` (ARITHMETIC_UNDER_OR_OVERFLOW) if denominator > 0 or `0x12` (DIVISION_BY_ZERO) if denominator == 0 instead of using `MathOverflowedMulDiv`

- In DoubleEndedQueue, we should probably replace
  - `QueueEmpty` → panic 0x31 (POP_ON_EMPTY_ARRAY)
  - `QueueFull` → panic 0x32 (ARRAY_ACCESS_OUT_OF_BOUNDS)
  - `QueueOutOfBounds` → panic 0x41 (TOO_MUCH_MEMORY_ALLOCATED)

Amxx:
Fixes #4867

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/structs/DoubleEndedQueue.sol
+++ b/contracts/utils/structs/DoubleEndedQueue.sol
@@ -2,6 +2,8 @@
 // OpenZeppelin Contracts (last updated v5.0.0) (utils/structs/DoubleEndedQueue.sol)
 pragma solidity ^0.8.20;
 
+import { Panic } from "../Panic.sol";
+
 /**
  * @dev A sequence of items with the ability to efficiently push and pop items (i.e. insert and remove) on both ends of
  * the sequence (called front and back). Among other access patterns, it can be used to implement efficient LIFO and
@@ -15,21 +17,6 @@ pragma solidity ^0.8.20;
  * ```
  */
 library DoubleEndedQueue {
-    /**
-     * @dev An operation (e.g. {front}) couldn't be completed due to the queue being empty.
-     */
-    error QueueEmpty();
-
-    /**
-     * @dev A push operation couldn't be completed due to the queue being full.
-     */
-    error QueueFull();
-
-    /**
-     * @dev An operation (e.g. {at}) couldn't be completed due to an index being out of bounds.
-     */
-    error QueueOutOfBounds();
-
     /**
      * @dev Indices are 128 bits so begin and end are packed in a single storage slot for efficient access.
      *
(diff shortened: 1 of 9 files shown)


contracts/utils/structs/DoubleEndedQueue.sol
```diff
@@ -78,12 +65,12 @@ library DoubleEndedQueue {
...
      * @dev Inserts an item at the beginning of the queue.
      *
-     * Reverts with {QueueFull} if the queue is full.
+     * Reverts with {RESOURCE_ERROR} panic code if the queue is full.
      */
     function pushFront(Bytes32Deque storage deque, bytes32 value) internal {
         unchecked {
             uint128 frontIndex = deque._begin - 1;
-            if (frontIndex == deque._end) revert QueueFull();
+            if (frontIndex == deque._end) Panic.panic(Panic.RESOURCE_ERROR);
```
ernestognw, line 73: Not sure if this is the right fit because it indicates a resource error in memory. This is in storage.

I'd say using `Panic.ARRAY_OUT_OF_BOUNDS` would be a better fit, what do you think?

Amxx: This is not an access that falls beyond the limit of the array, this is a fundamental limitation. Ressources available are limited and this is trying to use more than is available.

Note that `RESOURCE_ERROR` is documented as
> resource error (too large allocation or too large array)

I believe "too large array" is pretty much the problem here.

Also, note that for this error to happen, someone would have to "push" more elements in the array than there are atoms in the observable universe. This is NEVER going to happen, so no need to worry to much.

ernestognw: > This is NEVER going to happen, so no need to worry to much.

lmao yes I'm worrying too much.

> This is not an access that falls beyond the limit of the array, this is a fundamental limitation of the structure.

I still think the line is blur. Imo this is because the array **is bounded** to the 2**256 elements.

I'm solving since I'm fine either since it shouldn't happen.

ernestognw:
I'll take some time to try updating the docs to show NatSpec for internal constants.

Otherwise we can't reference the listed codes. Also I noticed that the Panic library is not in the docs and also that we're interested in showing the panic codes in the docs too.

ernestognw (approved):

PR #4879, opened 2024-02-06
Update hardhat-exposed to remove `mock/_import.sol`

Amxx:
#3298 introduced a trick to expose some libraries. Without this trick, testing the upgradeable contract would fail because some vanila contract would not be exposed correctly.

`hardhat-exposed@0.3.14-0` fixes that by transitivelly exposing all the imported files. In out case, that would expose `Panic` through the `Xxx.sol → Math.sol → Panic.sol` path.

I still want to note that if we ever include a library that is not used by any contract (mock or otherwise) ... then we would need to reimplement the import trick.


ernestognw:
> I still want to note that if we ever include a library that is not used by any contract (mock or otherwise) ... then we would need to reimplement the import trick.

This is fine imo.

ernestognw (approved):

PR #345, opened 2017-08-01
Add Transfer event when token is minted

SylTi:
cf issue https://github.com/OpenZeppelin/zeppelin-solidity/issues/344

PR #177, opened 2017-04-01
Add an example of some possible improvements to the docs

maurelian:
Only creating this PR to answer [a question](https://openzeppelin.slack.com/archives/C229D43PG/p1490993555001618) from @federicobond in Slack.

This is an example of some fairly small improvements that I think would be very impactful.

PR #4881, opened 2024-02-07
Implement P256 verification via RIP-7212 precompile with Solidity fallback

Amxx:
Fixes LIB-1225

## Why do we care about secp256r1?

Most security application uses secp256r1 (also known as p256). This lead hardware manufacturers to implement it, and leave other “exotic” curves on the side. Today, billions of people around the world own devices with spetial security hardware that supports secp256r1. If that was the curve used by ethereum, all these people would basically already own a hardware wallet … but unfortunatelly that is not the case.

If we cannot easily modify the curves supported by major smartphones manufacturer, we can provide tools to verify secp256r1 curve onchain. This would allow control of ERC-4337 smart wallets (among others) through a device designed to handle security keys (something users are notoriously bad at).

## What @openzeppelin/contracts could provide

Existing wallets provide mechanisms to produce secp256k1 signature, both for transactions and messages. Solidity provides a precompile that, given a hash and a signature, will recover the address of the signer (using secp256k1). No such precompile exist for secp256r1.

There exist solidity implementations of the secp256r1 “verification” workflow. There is also a proposal to provide that verification through a precompile. Even if the precompile is implemented, it is likelly that many chains will not upgrade soon. A solidity implementation would remain usefull for users on these chains.

In some cases, users may want to follow the “recovery” flow that they are familiar with. There is also no proposal for a precompile that would do that operation. A solidity implementation would possibly be usefull to many users, and remain uncontested in the near future.

### Stack too depth
Current proposed implementation works well when turning optimization on. However, compilation fails with "stack to deep" if optimizations are NOT turned on. This PR does enable optimizations for all tests to circumvent this issue. Also, users will have to enable optimizations if they want to use this library, which they should definitelly do given the gast costs.

- [x] This is still an issue when running coverage :/
  - This was fixed by adding `details: { yul: true },` to the optimizer settings. This change in optimization setup may affect the accuracy of gas reporting in this PR (reference doesn't use the same settings)

### Benchmarking
[This repo](https://github.com/Amxx/secp256r1) provides benchmarking of this implementation against other existing ones.
Capture d’écran du 2024-02-07 11-33-29

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,316 @@
...
+
+    /**
+     * @dev Point doubling on the jacobian coordinates
+     * https://en.wikibooks.org/wiki/Cryptography/Prime_Curve/Jacobian_Coordinates
+     */
+    function _jDouble(uint256 x, uint256 y, uint256 z) private pure returns (uint256 x2, uint256 y2, uint256 z2) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            let p := P
+            let yy := mulmod(y, y, p)
```
ernestognw, line 196: Shouldn't we return the point at infinity (0,1,0) if `y == 0` at this point?

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,316 @@
...
+        /// @solidity memory-safe-assembly
+        assembly {
+            let p := P
+            let zz1 := mulmod(z1, z1, p) // zz1 = z1²
+            let zz2 := mulmod(z2, z2, p) // zz2 = z2²
+            let u1 := mulmod(x1, zz2, p) // u1 = x1*z2²
+            let u2 := mulmod(x2, zz1, p) // u2 = x2*z1²
+            let s1 := mulmod(y1, mulmod(zz2, z2, p), p) // s1 = y1*z2³
+            let s2 := mulmod(y2, mulmod(zz1, z1, p), p) // s2 = y2*z1³
+            let h := addmod(u2, sub(p, u1), p) // h = u2-u1
```
ernestognw, line 174: Similar to the `_jDouble` function, I see a couple of operations not followed from the reference here:

```
...
 if (U1 == U2)
   if (S1 != S2)
     return POINT_AT_INFINITY
   else 
     return POINT_DOUBLE(X1, Y1, Z1)
...
```

I want to make sure I understand why this is being ignored

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,316 @@
...
+        uint256 x2,
+        uint256 y2,
+        uint256 z2
+    ) private pure returns (uint256 x3, uint256 y3, uint256 z3) {
+        if (z1 == 0) {
+            return (x2, y2, z2);
+        }
+        if (z2 == 0) {
+            return (x1, y1, z1);
+        }
```
ernestognw, line 164: Is this because adding some coordinates to the point at infinity results in the same coordinates?

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,316 @@
...
+                u1 <<= 2;
+                u2 <<= 2;
+            }
+        }
+        return _affineFromJacobian(x, y, z);
+    }
+
+    /**
+     * @dev Precompute a matrice of useful jacobian points associated to a given P. This can be seen as a 4x4 matrix
+     * that contains combinaison of P and G (generator) up to 3 times each. See table below:
```
ernestognw, line 265:
```suggestion
     * @dev Precompute a matrice of useful jacobian points associated with a given P. This can be seen as a 4x4 matrix
     * that contains combination of P and G (generator) up to 3 times each. See the table below:
```

ernestognw (commented):
Left a couple of questions and notes while reviewing. I've been delaying this review for a while but at first sight it looks really well implemented, and I mostly need to familiarize and check the math is correct.

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,316 @@
...
+    /**
+     * @dev signature verification
+     * @param h - hashed message
+     * @param r - signature half R
+     * @param s - signature half S
+     * @param qx - public key coordinate X
+     * @param qy - public key coordinate Y
+     */
+    function verify(uint256 h, uint256 r, uint256 s, uint256 qx, uint256 qy) internal view returns (bool) {
+        if (r == 0 || r >= N || s == 0 || s >= N || !isOnCurve(qx, qy)) return false;
```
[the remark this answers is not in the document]
Amxx: I'm not exactly sure how that could be used maliciously. I think the point is more like: we know this cannot possibly be a valid input, so we should reject it. We should count on the caller doing the sanity check.

Can we prove that is r (or s) is 0, then the function will return false without reverting? Maybe.

- If `s = 0`, we get `w = 0`, which is not actually an inverse. That gives `u1 = 0`.
- If `r = 0`, we get `u2 = 0`.

In `_jMultShamir`, it is unclear to me that any of there value being 0 can be treated in a specific way. If both are 0, then the end point is (0, 0) and we might get "true".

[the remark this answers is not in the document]
Amxx: I tested that. If you remove the zero check, then this test passes:

```
    /// forge-config: default.fuzz.runs = 512
    function testVerifyZero(uint256 seed, bytes32 digest) public {
        uint256 privateKey = bound(uint256(keccak256(abi.encode(seed))), 1, P256.N - 1);

        (uint256 x, uint256 y) = P256.getPublicKey(privateKey);
        assertTrue(P256.verify(uint256(digest), 0, 0, x, y));
    }
```

Said otherwize: without this check `r=0, s=0` would be a valid signature that would be verified for any `h` and any `qx, qy`.

contracts/utils/cryptography/P256.sol
Amxx: This part should be discussed over a call, but here are my first thought

A jacobian point (x,y,z) represents the carthesian point (x/z², y/z³). Therefore, any jacobian point that has `z=0` is at infinity (because x/z² and y/z³ are x/0 and y/0, which is not defined).

For the case of `_jDouble`, if y=0 we have `z' = 2*y*z = 0`

So my understanding is that, if y=0 we know the result is a point at infinity, and we can potentially skip some computation is we already know one. However, if we don't skip, and we do the computation "normaly", we get a point at infinity, which is correct. We could have skipped the computation, but given that in our case `addmod` and `mulmod` are "cheap native operation" (which is not the case in an x86 machine), its probably ok.

We can try implementing the skip, and see if its actually saving gas, but AFAIK, this is an optimisation issue and not a correctness issue.

contracts/utils/cryptography/P256.sol
Amxx: I believe that this is a similar case to https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4881#discussion_r1579436787, but I'm not 100% sure, so lets explore that

Amxx: if `u1=u2`, then `h = u2-u1 = 0`, and so `h*z1*z2 = 0` ... so it appears we return a point at infinity

Amxx:
Relevant source for discussion: https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian.html

contracts/utils/cryptography/P256.sol
Amxx: For the record:

Here the check is so that (0,0,0) is a "neutral element" for the jacobian addition. We could make the check as 
```solidity
        if (x1 == 0 && y1 == 0 && z1 == 0) {
            return (x2, y2, z2);
        }
        if (x2 == 0 && y2 == 0 && z2 == 0) {
            return (x1, y1, z1);
        }
```
but that would be more expensive (and that function is called 140 times when you verify)

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,315 @@
...
+        uint256 z = 0;
+        unchecked {
+            for (uint256 i = 0; i < 128; ++i) {
+                if (z > 0) {
+                    (x, y, z) = _jDouble(x, y, z);
+                    (x, y, z) = _jDouble(x, y, z);
+                }
+                // Read 2 bits of u1, and 2 bits of u2. Combining the two give a lookup index in the table.
+                uint256 pos = ((u1 >> 252) & 0xc) | ((u2 >> 254) & 0x3);
+                (x, y, z) = _jAdd(x, y, z, points[pos].x, points[pos].y, points[pos].z);
```
Amxx, line 254:
```suggestion
                for (pos > 0) {
                    (x, y, z) = _jAdd(x, y, z, points[pos].x, points[pos].y, points[pos].z);
                }
```

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,317 @@
...
+            for (uint256 i = 0; i < 128; ++i) {
+                if (z > 0) {
+                    (x, y, z) = _jDouble(x, y, z);
+                    (x, y, z) = _jDouble(x, y, z);
+                }
+                // Read 2 bits of u1, and 2 bits of u2. Combining the two give a lookup index in the table.
+                uint256 pos = ((u1 >> 252) & 0xc) | ((u2 >> 254) & 0x3);
+                if (pos > 0) {
+                    (x, y, z) = _jAdd(x, y, z, points[pos].x, points[pos].y, points[pos].z);
+                }
```
Amxx, line 256: So here the if is optional.

If we remove the if, we are going to load `points[0]` which is `(0,0,0)` ... and the `_jAdd` will skip that as the "neutral element". The if here as a cost. 15/16 we pay it for no real reason (and we still pay the check in _jAdd). 1/16 the if avoids the overhead of a function call.

I'm going to benchmark which one is better and comment that so we don't go back and forward.

Amxx: I checked, skipping the mloads in 1/16 cases is a bigger gain than the loss of the if in the other 15/16 cases. Keeping the if is the more effective solution here

contracts/utils/cryptography/P256.sol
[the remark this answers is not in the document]
ernestognw: Perfect, thanks for the thorough explanation. I agree this check is required, is just that I didn't get why in the beginning.

Solving

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,337 @@
...
+     * the solidity implementation otherwise.
+     *
+     * @param h - hashed message
+     * @param r - signature half R
+     * @param s - signature half S
+     * @param qx - public key coordinate X
+     * @param qy - public key coordinate Y
+     */
+    function verify(uint256 h, uint256 r, uint256 s, uint256 qx, uint256 qy) internal view returns (bool) {
+        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
```
ernestognw, line 50: Address `0x100` is from RIP-7212 but EIP-7212 specifies `0x0b`. Seems like it'll be different for mainnet

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,337 @@
...
+            result := keccak256(0x00, 0x40)
+        }
+    }
+
+    /**
+     * @dev check if a point is on the curve.
+     */
+    function isOnCurve(uint256 x, uint256 y) internal pure returns (bool result) {
+        /// @solidity memory-safe-assembly
+        assembly {
```
ernestognw, line 140:
```suggestion
        assembly ("memory-safe") {
```

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,337 @@
...
+    function _jAdd(
+        uint256 x1,
+        uint256 y1,
+        uint256 z1,
+        uint256 x2,
+        uint256 y2,
+        uint256 z2
+    ) private pure returns (uint256 rx, uint256 ry, uint256 rz) {
+        /// @solidity memory-safe-assembly
+        assembly {
```
ernestognw, line 178:
```suggestion
        assembly ("memory-safe") {
```

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,337 @@
...
+        }
+    }
+
+    /**
+     * @dev Point doubling on the jacobian coordinates
+     * Reference: https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian.html#doubling-dbl-1998-cmo-2
+     */
+    function _jDouble(uint256 x, uint256 y, uint256 z) private pure returns (uint256 rx, uint256 ry, uint256 rz) {
+        /// @solidity memory-safe-assembly
+        assembly {
```
ernestognw, line 206:
```suggestion
        assembly ("memory-safe") {
```

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /**
+     * @dev Hash public key into an address
+     * @param qx - public key coordinate X
+     * @param qy - public key coordinate Y
+     */
+    function getAddress(uint256 qx, uint256 qy) internal pure returns (address result) {
+        /// @solidity memory-safe-assembly
+        assembly {
```
ernestognw, line 128:
```suggestion
        assembly ("memory-safe") {
```

contracts/utils/cryptography/P256.sol
Amxx: Are they kidding? What the actual fuck?

Amxx: https://x.com/ulerdogan/status/1802623656813646127

ernestognw: Cool, so I see it should be safe although I'm really concerned about how unclear this is.
I would even think twice before releasing the `0x100` as it may turn out at the end that they were going to use `0x0b` 😂

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * @dev derivate public key
+     * @param privateKey - private key
+     */
+    function getPublicKey(uint256 privateKey) internal view returns (uint256, uint256) {
+        (uint256 x, uint256 y, uint256 z) = _jMult(GX, GY, 1, privateKey);
+        return _affineFromJacobian(x, y, z);
+    }
```
ernestognw, line 161: Is this function needed? I don't see a real use case for it as it would require exposing the private key anyway

contracts/utils/cryptography/P256.sol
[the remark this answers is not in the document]
ernestognw: I see you already added tests for malleability, thanks!

And yes, I would be in favor of adding it to `verifySolidity` and `recovery`, I still think it's a footgun otherwise and I'm not even sure what's the use case for not restricting it.

I'm adding a warning for now, I think is the very least

contracts/utils/cryptography/P256.sol
ernestognw: Ok, I was missing why we're using these coordinates in the first place. I found [this article](https://hackmd.io/@cailynyongyong/HkuoMtz6o) that helped me understand better. 

So if I'm understanding correctly, we prefer operating in the jacobian space because it has advantages over the affine space. The transformation from affine coordinates to jacobian corresponds to what you described above, but the point with `z = 0` is just undefined given the representation (i.e. `x/0**2` or `y/z**3`)

In such case then it doesn't matter to return here because the returned point will also be undefined (`z' = 0`).
I'm solving now it makes sense, please point me out if I'm getting something wrong

contracts/utils/cryptography/P256.sol
ernestognw: > if u1=u2, then h = u2-u1 = 0, and so h*z1*z2 = 0 ... so it appears we return a point at infinity

Yes, now this is clear, thanks

> with r = s2-s1, which is 0 if s2=s1 (case where the ref says we should return double) and not 0 if s2 != s1 (case where the ref says we should return point at infinity, which we do)

Right, so basically where the reference says that we should return double, it'll be doubling a point where `Z1 == 0`, which would the point at infinity anyway.

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,353 @@
...
+     * @param h - hashed message
+     * @param r - signature half R
+     * @param s - signature half S
+     * @param qx - public key coordinate X
+     * @param qy - public key coordinate Y
+     *
+     * WARNING: Signatures are malleable, and this function does not check for malleability. Consider rejecting
+     * the upper half order of the curve (i.e. s > N/2)
+     */
+    function verify(uint256 h, uint256 r, uint256 s, uint256 qx, uint256 qy) internal view returns (bool) {
```
ernestognw, line 55: Any reason for representing the hash as `uint256`? I think we're generally more used to `bytes32`

Amxx: No real reason. All params could be either uint256 or bytes32.

I think that is since the verification is math heavy, we'll need to cast everything into uint256 anyway.

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,363 @@
...
+     * @param qx - public key coordinate X
+     * @param qy - public key coordinate Y
+     */
+    function getAddress(uint256 qx, uint256 qy) internal pure returns (address result) {
+        assembly ("memory-safe") {
+            mstore(0x00, qx)
+            mstore(0x20, qy)
+            result := keccak256(0x00, 0x40)
+        }
+    }
```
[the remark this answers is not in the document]
Amxx: > Using a bytes32 as a "public key identifier" feels more natural to me

That "preverts" us from using AccessControl to give roles to the keys ... but I see your point.

Lets remove them, and we can always re-add them latter if we want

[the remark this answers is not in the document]
ernestognw: > That "preverts" us from using AccessControl to give roles to the keys

Right, this is a good point. I guess it remains to be seen whether users would do this or assign the role to a controlled account using 1271 in the end.

Agree, let's solve it for now and we can discuss it better later.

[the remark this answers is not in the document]
Amxx: We already discussed that ...

If you use an ERC-1271 account you are just moving who does the verification. The ERC-1271 contract will get a hash and a signature ... it will recover a public key of the signer ... and then what ? It needs to check that this identity is an authorized signer. This is where the ERC-1271 contract could use AccessControl ... but only if it turns the public key it just recovered into an address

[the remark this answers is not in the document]
Amxx: At some point, someone must check "is that P256 key that I just verified an authorized identity". 

That is when it's useful to represent it as an address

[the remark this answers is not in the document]
ernestognw: > and then what ? It needs to check that this identity is an authorized signer. This is where the ERC-1271 contract could use AccessControl ... but only if it turns the public key it just recovered into an address

This sounds like we haven't discussed it enough.
Regardless, I don't think it makes sense to discuss right now since we don't have evidence that users would like to use ERC-1271 over non-accessible addresses.

> but only if it turns the public key it just recovered into an address

Not really. A contract can be just a signer defined by an internal public key:

```solidity
contract P256Identity is IERC1271 {
    using P256 for bytes32;

    // bytes4(keccak256("isValidSignature(bytes,bytes)")
    bytes4 internal constant EIP1271_MAGIC_VALUE = 0x20c13b0b;

    struct PublicKey {
        bytes32 qx;
        bytes32 qy;
    }

    PublicKey publicKey;

    constructor(bytes32 qx, bytes32 qy) {
        publicKey = PublicKey(qx, qy);
    }

    function isValidSignature(
        bytes32 digest,
        bytes memory signature
    ) external view returns (bytes4) {
        return
            _verifyP256Identity(digest, signature)
                ? EIP1271_MAGIC_VALUE
                : bytes4(0);
    }

    function _verifyP256Identity(
        bytes32 digest,
        bytes memory signature
        PublicKey memory pubKey = publicKey();

        (uint256 r, uint256 s) = _splitSignature(signature);

        if (normalize) {
            digest = sha256(abi.encode(digest));
        }

        return digest.verify(digest, r, s, pubKey.qx, pubKey.qy);
    }
    function _splitSignature(
        bytes memory signature
    ) private pure returns (uint256 r, uint256 s) {
        r = uint256(bytes32(signature[00:31]));
        s = uint256(bytes32(signature[32:63]));
    }
}
```

In this way, it's clear that the contract has an address, is controlled by a single private key, and it's already composable with the broader ecosystem. Also it doesn't allow to deposit ETH that might get lost.

contracts/utils/cryptography/P256.sol
ernestognw: My concern is that users may need to cast in the frontend using Javascript where the encoding can be a mess. Although it's trivial to cast using ethers or viem, I don't think it's straightforward to guess.

> I think that is since the verification is math heavy, we'll need to cast everything into uint256 anyway.

We can hide the interface and expose only `bytes32`

```solidity
function verifySolidity(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) internal view returns (bool) {. 
  return _verifySolidity(uint256(h), uint256(r), uint256(s), uint256(qx), uint256(qy));
}

function _verifySolidity(uint256 h, uint256 r, uint256 s, uint256 qx, uint256 qy) private view returns (bool) {
  // Current solidity verification
}
```

What do you think?

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,324 @@
...
+        uint256 u2 = mulmod(s, w, N);
+        (uint256 x, uint256 y) = _jMultShamir(points, u1, u2);
+        return (x, y);
+    }
+
+    /**
+     * @dev derivate public key
+     * @param privateKey - private key
+     */
+    function getPublicKey(uint256 privateKey) internal view returns (uint256, uint256) {
```
ernestognw, line 132: Would you mind if I move this function to the tests?
I see they’re relevant to derive the public key in Solidity but I’m expecting Foundry to eventually release it in Vm where the underlying implementation is in rust

Amxx: Ok

contracts/utils/cryptography/P256.sol
Amxx: > My concern is that users may need to cast in the frontend using Javascript

That really depend on the library you are using. In ethers.js, uint256 is BigNumberish which accepts many format, including hex string and buffers. It is actually less restictive than bytes32.

Amxx: I'm be ok with changing the types, but not through one more variant of the function (override). We have enough already. More function is more complex testing, more confusion.

ernestognw: Yes, agree on not adding more functions. Let's change only the types. I'll push a commit

test/utils/cryptography/P256.t.sol
```diff
@@ -0,0 +1,125 @@
...
+pragma solidity ^0.8.20;
+
+import {Test} from "forge-std/Test.sol";
+import {P256} from "@openzeppelin/contracts/utils/cryptography/P256.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+contract P256Test is Test {
+    /// forge-config: default.fuzz.runs = 512
+    function testVerify(uint256 seed, bytes32 digest) public {
+        uint256 privateKey = bound(uint256(keccak256(abi.encode(seed))), 1, P256.N - 1);
```
cairoeth, line 12: this is cheaper and faster:
```suggestion
    function testVerify(string calldata seed, bytes32 digest) public {
        (, uint256 privateKey) = makeAddrAndKey(seed);
```

also for `testRecover`

test/utils/cryptography/P256.t.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Test} from "forge-std/Test.sol";
+import {P256} from "@openzeppelin/contracts/utils/cryptography/P256.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+contract P256Test is Test {
+    /// forge-config: default.fuzz.runs = 512
```
cairoeth, line 10: any reason for 512 runs? is pretty low imo, especially as the workflow will use this value instead of 5000 from foundry.toml

test/utils/cryptography/P256.t.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Test} from "forge-std/Test.sol";
+import {P256} from "@openzeppelin/contracts/utils/cryptography/P256.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+contract P256Test is Test {
```
cairoeth, line 9: think we should have a unit test for out of bound public keys -- i can add one

cairoeth: it also doesn't hurt to have unit tests that import and use https://github.com/C2SP/wycheproof

cairoeth (commented):
some comments on testing

test/utils/cryptography/P256.t.sol
ernestognw: > it also doesn't hurt to have unit tests that import and use https://github.com/C2SP/wycheproof

Agree, I was thinking in loading the test vectors as we did in RSA (but for hardhat), wdyt?

test/utils/cryptography/P256.t.sol
ernestognw: Good catch, though the difference here is that the order of G (N) is way smaller for P256 than for secp256k1 (which I assume is what `makeAddrAndKey` uses). Perhaps it's enough with `privateKey = bound(privateKey, 1, P256.N - 1)`

cairoeth: yes good point, i think we can do `privateKey = bound(privateKey, 1, P256.N - 1)`

test/utils/cryptography/P256.t.sol
cairoeth: added with foundry here: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4881/commits/93629369e57325c87463989994415106066acdd1 

checked with hardhat and is pretty slow compared to this, but we can do it if we really want to

samjove:
Unsubscribe

test/utils/cryptography/P256.t.sol
Amxx: It gets very long very fast. 500 is already pretty good IMO

contracts/utils/cryptography/P256.sol
frangio: IMO should be `bytes32`. `uint` types have arithmetic operations which should not be done on these values.

frangio:
IMO the API should be `verify` and `verifyNative` (renamed from `verify7212`). `tryVerify7212` should be private, `verifyNative` assumes that the precompile is available. `verifySolidity` should be private too, though it may be needed internally for testing though...

The API functions should also include malleability checks. Optionally there could separate functions without malleability checks (one for `verify` and one for `verifyNative`?). This is what EIP-7212 recommends:

> Wrapper libraries SHOULD add a malleability check by default, with functions wrapping the raw precompile call (exact NIST FIPS 186-5 spec, without malleability check) clearly identified. For example, P256.verifySignature and P256.verifySignatureWithoutMalleabilityCheck. Adding the malleability check is straightforward and costs minimal gas.

cairoeth:
> IMO the API should be `verify` and `verifyNative` (renamed from `verify7212`). `tryVerify7212` should be private, `verifyNative` assumes that the precompile is available. `verifySolidity` should be private too, though it may be needed internally for testing though...
> 
> The API functions should also include malleability checks. Optionally there could separate functions without malleability checks (one for `verify` and one for `verifyNative`?). This is what EIP-7212 recommends:
> 
> > Wrapper libraries SHOULD add a malleability check by default, with functions wrapping the raw precompile call (exact NIST FIPS 186-5 spec, without malleability check) clearly identified. For example, P256.verifySignature and P256.verifySignatureWithoutMalleabilityCheck. Adding the malleability check is straightforward and costs minimal gas.

++ on all points -- adding malleability check in the "default" function also aligns with the other utils like ECDSA and RSA in terms of security checks

test/utils/cryptography/P256.t.sol
ernestognw: I see the tests are failing. Previously I shared these cases I identified where failing with @Amxx and I think these are those failing here too. FYI 

#### Case 1 (OZ = false, P256Verifier = true)
r:  0x000000000000000000000000000000004319055358e8617b0c46353d039cdaab
s:  0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc63254e
x:  0x0ad99500288d466940031d72a9f5445a4d43784640855bf0a69874d2de5fe103
> wycheproof/ecdsa_secp256r1_sha256_p1363_test.json EcdsaP1363Verify SHA-256 #113: k*G has a large x-coordinate
> wycheproof/ecdsa_webcrypto_test.json EcdsaP1363Verify SHA-256 #113: k*G has a large x-coordinate

#### Case 2 (OZ = false, P256Verifier = true)
r:  0x7cf27b188d034f7e8a52380304b51ac3c08969e277f21b35a60b48fc47669978
s:  0xb6db6db6249249254924924924924924625bd7a09bec4ca81bcdd9f8fd6b63cc
x:  0xc6a771527024227792170a6f8eee735bf32b7f98af669ead299802e32d7c3107
> wycheproof/ecdsa_secp256r1_sha256_test.json EcdsaVerify SHA-256 #345: extreme value for k and s^-1
> wycheproof/ecdsa_secp256r1_sha256_p1363_test.json EcdsaP1363Verify SHA-256 #177: extreme value for k and s^-1

#### Case 3 (OZ = false, P256Verifier = true)
r:  0x7cf27b188d034f7e8a52380304b51ac3c08969e277f21b35a60b48fc47669978
s:  0xb6db6db6249249254924924924924924625bd7a09bec4ca81bcdd9f8fd6b63cc
x:  0xc6a771527024227792170a6f8eee735bf32b7f98af669ead299802e32d7c3107
> wycheproof/ecdsa_secp256r1_sha256_test.json EcdsaVerify SHA-256 #345: extreme value for k and s^-1
> wycheproof/ecdsa_secp256r1_sha256_p1363_test.json EcdsaP1363Verify SHA-256 #177: extreme value for k and s^-1

#### Case 4 (OZ = false, P256Verifier = true)
r:  0x000000000000000000000000000000004319055358e8617b0c46353d039cdaab
s:  0xffffffff00000000ffffffffffffffffbce6faada7179e84f3b9cac2fc63254e
x:  0xd705d16f80987e2d9b1a6957d29ce22febf7d10fa515153182415c8361baaca4
> wycheproof/ecdsa_webcrypto_test.json EcdsaP1363Verify SHA-256 #113: k*G has a large x-coordinate
> wycheproof/ecdsa_secp256r1_sha256_p1363_test.json EcdsaP1363Verify SHA-256 #113: k*G has a large x-coordinate

ernestognw:
> verifySolidity should be private too, though it may be needed internally for testing though

I understand that ideally the Solidity version won't be used if EIP7212 is shipped to every network, but I guess is not the case and perhaps some networks may never get the precompile. Think about networks running in the OP Stack or something similar, perhaps it's easier for them to run their rollup with a Solidity implementation that waiting for a generalized release.

I'm not sure how upgrades to this chains work but the point goes along. I would be in favor of keeping it so I left it `internal` for now.

> Optionally there could separate functions without malleability checks (one for verify and one for verifyNative?).

Added them as private but I know @Amxx was strongly in favor of keeping the malleability by default because the tooling produces values on both sides of the curve. I updated the tests accordingly to ensure a lower-order `s` value, I think that the tooling produces both values is easily solvable by the Dapp developer.

Currently the interface is left with:

- verify
- verifyNative
- verifySolidity
- recovery
- isOnCurve

If we expose the malleable ones, I would be in favor only of `verifyMalleable` and also a `checkLowerS` function as we discussed @Amxx. I tried implementing it now but it's unclear whether the `checkLowerS` should revert or just return false. I didn't added it yet because I'm not sure if it's useful in the end.

I think fuzz tests are failing due to the [cases I shared](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4881#discussion_r1649544090) with @cairoeth. The Hardhat tests are failing because there's a case using `r` as `s`, which may be in the lower order.

frangio:
> Think about networks running in the OP Stack or something similar, perhaps it's easier for them to run their rollup with a Solidity implementation that waiting for a generalized release.

Right but I'm saying they should use `verify`, i.e. native with fallback to Solidity. I'm thinking using the Solidity version directly should be discouraged? The only reason I can think one might want to use the Solidity version directly is if the precompile is there but it's somehow not compliant.

It would be fine to make it internal, my main point is that it should be discouraged. For example, there can only be a raw variant without malleability check.

Amxx:
Comming back from my weekend, I see a lot of changes, so I have questions

- what is the point of having both `verify` and `_verifyMalleable` if the `_verifyMalleable` is private and only ever called by `verify` ?
- Same question for `verifyNative` and `_verifyNativeMalleable` ?
- Why is `verifySolidity` accepting maleable, but not named accordingly. Not that it is internal, so accessible.
- Do we expect all callers to implement `ensureLowerOrderS` ?

Last week we agreed to have the function malleable by default, and having an helper (that was opt-in) so devs can easily check the S value. Now something completelly different is implemented...

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,298 @@
...
+    ) private view returns (bool valid, bool supported) {
+        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
+        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
+    }
+
+    /**
+     * @dev Same as {verify}, but only the Solidity implementation is used.
+     */
+    function verifySolidity(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) internal view returns (bool) {
+        if (r == 0 || uint256(r) >= N || s == 0 || uint256(s) >= N || !isOnCurve(qx, qy)) {
```
Amxx, line 97:
```suggestion
        if (r == 0 || uint256(r) >= N || s == 0 || uint256(s) > HALF_N || !isOnCurve(qx, qy)) {
```

test/utils/cryptography/P256.t.sol
cairoeth: FYI^ the `s` parameters are above the malleability threshold so we have to make sure we get to the underlying issue not just skip in case we add malleability checks.

cairoeth: also want to make sure we are good for the following cases that occurred in other well known implementations:
- [verify signatures whose R,S > curve order](https://github.com/paritytech/libsecp256k1/commit/b525d5d318d9672a40250c1725fa1bb3156688b7)
- [all zero signatures](https://github.com/paulmillr/noble-secp256k1/commit/9082b405a3d1958b5b6a699bb408215ae30eea5b#diff-177d97aec07fd5b22f04049b34d74939b1d8364a7a83e8fb69ee3ee8695105fbR256-R270)
- [invalid r,s](https://github.com/paulmillr/noble-secp256k1/commit/c514f79ebdab1042dac214446f1ca32214a5edfd#diff-177d97aec07fd5b22f04049b34d74939b1d8364a7a83e8fb69ee3ee8695105fbR242-R255)
- [scalar multiplications](https://github.com/ziglang/zig/issues/15267)
- [message is curve order](https://github.com/paritytech/libsecp256k1/issues/62) -- verify instead of sign.

differential fuzzing with [cryptofuzz](https://github.com/guidovranken/cryptofuzz) would be amazing, will check it out.

test/utils/cryptography/P256.t.sol
cairoeth: added in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4881/commits/cbd2ff5b16a47c46381626c82cafc6221d98aa54

frangio:
@Amxx re: malleable by default

This is discouraged by EIP-7212:

>Wrapper libraries SHOULD add a malleability check by default, with functions wrapping the raw precompile call (exact NIST FIPS 186-5 spec, without malleability check) clearly identified. For example, P256.verifySignature and P256.verifySignatureWithoutMalleabilityCheck. Adding the malleability check is straightforward and costs minimal gas.

ernestognw:
> Comming back from my weekend, I see a lot of changes, so I have questions

I'll answer below, note that it wasn't meant to be final, I pushed what I thought was agreed from everyone's opinion but it's still open to discussion to consider yours 🙌🏻  

> * what is the point of having both `verify` and `_verifyMalleable` if the `_verifyMalleable` is private and only ever called by `verify` ?
> * Same question for `verifyNative` and `_verifyNativeMalleable` ?

To discuss whether it makes sense to inline it or not. Given that you favored keeping a malleable version, I thought it was easier to make it internal and remove the underscore if we agreed. Not a fan of that though.

> * Why is `verifySolidity` accepting maleable, but not named accordingly. Not that it is internal, so accessible.

Considering @frangio's opinion I got the sense it should be private but I waited to hear your opinion. Please don't consider it decided yet

> * Do we expect all signers/app to implement `ensureLowerOrderS` ? The ethereum wallets that produce secp256k1 signature all know to use a lower S value, but other cryptographic tools do not. Noble does not. I'm unsure, but I wouldn't be surprised if smartphone enclaves (our main target) do not either. Do we want to force apps into modifying the signatures produces by the signers?


I'm not surprised that Ethereum wallets started producing all signatures in the lower `S` value. When using signatures in a context that's not specific to smart contracts it may be fine to allow for malleability, it's just that we care about replayability and double spending in smart contracts.

I agree with you. Apps will be forced to modify the signatures produced by their signers but the workaround is fairly easy and perhaps is the most responsible way to design a wallet (a wallet that allows malleability would be risky!). I would argue this is probably why Ethereum wallets produce secp256k1 signatures in the lower S in the first place. 
 
> Last week we agreed to have the function malleable by default, and having an helper (that was opt-in) so devs can easily check the S value. Now something completelly different is implemented...

We can consider the helper, but I feel the overall consensus is to disallow malleability by default.

> Note, I'm in vavor of:
> 
> * having the least opinionated library
> * providing as much option to the devs
> * having as little exposed functions as possible.

So far I agree on providing flexibility to the devs, but the goal is to do it in the most secure way. If we don't find strong reasons to allow for malleability, then I wouldn't justify the risk of leaving a footgun in the library for the sake of tooling compatibility.

In my opinion, the tradeoff here is between broader compatibility and default security. When comparing the two I can see that:

1. The workaround is an easy function to invert the signature to the lower S order. We can even commit to contributing to these libraries to mitigate the concern.
2. The EIP-7212 already suggests wrapping the precompile.

My conclusion is that preserving broader compatibility puts our users at risk for a (yet unclear) use case, while disallowing malleability may be a pain for libraries dealing with P256 signatures but we can specify guidelines for adapting signatures produced in the higher order.

Amxx:
Let's keep it how it is right now then (inlined malleability protection in tryNative and solidity, everything else gets it indirectly) + fewer internal functions

test/utils/cryptography/P256.test.js
```diff
@@ -0,0 +1,114 @@
...
+    expect(await this.mock.$recovery(this.messageHash, this.recovery, ...this.signature)).to.deep.equal(this.publicKey);
+  });
+
+  it('reject signature with flipped public key coordinates ([x,y] >> [y,x])', async function () {
+    // flip public key
+    this.publicKey.reverse();
+
+    expect(await this.mock.$verify(this.messageHash, ...this.signature, ...this.publicKey)).to.be.false;
+    expect(await this.mock.$verifySolidity(this.messageHash, ...this.signature, ...this.publicKey)).to.be.false;
+    expect(this.mock.$verifyNative(this.messageHash, ...this.signature, ...this.publicKey)).to.eventually.be.false; // Flipped public key is not in the curve
```
Amxx, line 59:
```suggestion
    expect(await this.mock.$verifyNative(this.messageHash, ...this.signature, ...this.publicKey)).to.be.false; // Flipped public key is not in the curve
```

cairoeth:
in accordance to secp256r1 spec, fixed a bug with https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4881/commits/d8f4f7eb52468d60404627410980f96975ac9243 for when `n < p` and `x % n == r` but `x != r`, so some signatures that should be accepted were not.

ernestognw:
I didn't see the issue with the test vectors in Foundry @Amxx, any reasons to remove them?

Amxx:
> I didn't see the issue with the test vectors in Foundry @Amxx, any reasons to remove them?

Our guideline has alway been to do unit tests in hardhat, and fuzzing in foundry. So instead of using a .jsonl that was processed by someone else, and reading it in solidity, and doing that unit testing in foundry, ... I decided to get the original json file, unmodified, and processing it in javascript like we do for RSA.

test/utils/cryptography/P256.test.js
```diff
@@ -0,0 +1,156 @@
...
+    for (const { key, tests } of require('./ecdsa_secp256r1_sha256_p1363_test.json').testGroups) {
+      // parse public key
+      let [x, y] = [key.wx, key.wy].map(v => ethers.stripZerosLeft('0x' + v, 32));
+      if (x.length > 66 || y.length > 66) continue;
+      x = ethers.zeroPadValue(x, 32);
+      y = ethers.zeroPadValue(y, 32);
+
+      // run all tests for this key
+      for (const { tcId, comment, msg, sig, result } of tests) {
+        // only keep properly formated signatures
```
Amxx, line 134:
```suggestion
        // only keep properly formatted signatures
```

.changeset/odd-lobsters-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`P256`: Add a library for verification/recovery of Secp256r1 (Aka P256) signatures.
```
ernestognw, line 5:
```suggestion
`P256`: Library for verification and public key recovery of P256 (aka secp256r1) signatures.
```

test/utils/cryptography/P256.t.sol
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Test, console2} from "forge-std/Test.sol";
+import {stdJson} from "forge-std/StdJson.sol";
```
Amxx, line 6:
```suggestion
import {Test} from "forge-std/Test.sol";
```

test/utils/cryptography/P256.t.sol
ernestognw: > FYI^ the s parameters are above the malleability threshold so we have to make sure we get to the underlying issue not just skip in case we add malleability checks.

Yes, found out the hard way trying to debug a panic lol

> [verify signatures whose R,S > curve order](https://github.com/paritytech/libsecp256k1/commit/b525d5d318d9672a40250c1725fa1bb3156688b7)

Seems like these are already covered by the test vectors and fixed by bounding r to N

ernestognw (dismissed):
LGTM 🚀!

ernestognw (approved):
I managed to make the implementation compile and pass tests without requiring to enable `via-ir`. Although I agree users should be encouraged to use it given the gas costs, I think it's better not to force them.

LGTM again

Amxx:
I understand the logic for 9b240146a72383a551c2451c1854ed9c176246d5, but honestly that is an abomination. We surelly can do better.

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,297 @@
...
+        bytes32 s,
+        bytes32 qx,
+        bytes32 qy
+    ) private view returns (bool valid, bool supported) {
+        if (r == 0 || uint256(r) >= N || s == 0 || uint256(s) > HALF_N || !isOnCurve(qx, qy)) {
+            return (false, true); // signature is invalid, and its not because the precompile is missing
+        }
+
+        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
+        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
```
pcaversaccio, line 90: What if Ethereum puts another precompile at `0x100` in the future that also returns a `0x20` length return data? Might be an edge case, but EIP and RIP must not align in the long run! Or another example is, if an L2 already put a precompile on `0x100` which is different but also returns `0x20` length (haven't done research on this tbh).

Amxx: We were told that EIP-7212 is not up to date, that RIP-7212 is the one to follow, and that precompile address allocation would be organised at the RIP level to avoid conflicts.

If the community can't figure that out, wtf are we (library devs) expected to do ?

pcaversaccio: "would be organised at the RIP level to avoid conflicts." -> talk is cheap ;) so idk, I just don't trust this really. The reason why I haven't provided a native way to call the `P256` precompile in my snekmate (Vyper) library is that I leave this decision to the devs deploying it. I.e. they plan to deploy on 5 chains, 3 of them have the precompile and 2 not, so they write their own wrapper around it using e.g. `0x100` as target since it's ensured that it exists there.

Amxx: If you don't want to risk it, just use `verifySolidity`

pcaversaccio: you know how people are: "hey let's gas optimise everything and native looks sweet in our tests since we run it on a `P256`-supported fork." I think a good compromise is to put a warning in the docs about the current situation?

Amxx: Please feel free to propose some change with this additional doc :)

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,302 @@
...
+        uint256 y2,
+        uint256 z2
+    ) private pure returns (uint256 rx, uint256 ry, uint256 rz) {
+        assembly ("memory-safe") {
+            let p := P
+            let z1 := mload(add(p1, 0x40))
+            let s1 := mulmod(mload(add(p1, 0x20)), mulmod(mulmod(z2, z2, p), z2, p), p) // s1 = y1*z2³
+            let s2 := mulmod(y2, mulmod(mulmod(z1, z1, p), z1, p), p) // s2 = y2*z1³
+            let r := addmod(s2, sub(p, s1), p) // r = s2-s1
+            let u1 := mulmod(mload(add(p1, 0x00)), mulmod(z2, z2, p), p) // u1 = x1*z2²
```
Amxx, line 182:
```suggestion
            let u1 := mulmod(mload(p1), mulmod(z2, z2, p), p) // u1 = x1*z2²
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -30,6 +30,47 @@ function _verify(bytes32 data, bytes memory signature, address account) internal
...
+    bytes32 s,
+    bytes32 qx,
+    bytes32 qy
+) internal pure returns (bool) {
+    // Will only call the precompile at address(0x100)
+    return data.verifyNative(data, r, s, qx, qy);
+}
+----
+
+IMPORTANT: The P256 library only allows for `s` values in the lower order of the curve (i.e. `s < N`) to prevent malleability. In case your tooling produces signatures in both sides of the curve, consider flipping the `s` value to keep compatibility.
```
pcaversaccio, line 72:
```suggestion
IMPORTANT: The P256 library only allows for `s` values in the lower order of the curve (i.e. `s <= N/2`) to prevent malleability. In case your tooling produces signatures in both sides of the curve, consider flipping the `s` value to keep compatibility.
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -30,6 +30,47 @@ function _verify(bytes32 data, bytes memory signature, address account) internal
...
+    bytes32 r,
+    bytes32 s,
+    bytes32 qx,
+    bytes32 qy
+) internal pure returns (bool) {
+    return data.verify(data, r, s, qx, qy);
+}
+----
+
+By default, the `verify` function will try calling the (https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md)[RIP-7212] precompile at address `0x100` and will fallback to an implementation in Solidity if not available. We encourage you to use `verifyNative` if you know the precompile is available in the chain you're working on.
```
pcaversaccio, line 54:
```suggestion
By default, the `verify` function will try calling the (https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md)[RIP-7212] precompile at address `0x100` and will fallback to an implementation in Solidity if not available. We encourage you to use `verifyNative` if you know the precompile is available on the chain you're working on and on any other chain on which you intend to use the same bytecode in the future. In case of any doubts regarding the implementation roadmap of the native precompile `P256` of potential future target chains, please use `verifySolidity`.
```

contracts/utils/cryptography/P256.sol
frangio: Are there any concrete past examples or discussion about L2s deploying Ethereum precompiles on different addresses than L1?

pcaversaccio: Not that I'm personally aware of but there are examples where L1 precompiles are not deployed on L2. One example is ZKsync Era: https://docs.zksync.io/build/developer-reference/ethereum-differences/pre-compiles. All I'm trying to say is that we need to be prepared for more EVM fragmentation than any of us as library authors would like. Also, the past doesn't really imply the future, so not sure if this is a good indicator...

frangio: > L1 precompiles are not deployed on L2

Yes this definitely has to be considered, that's why the default interface in this library implements a fallback to a Solidity implementation.



> I'm trying to say is that we need to be prepared for more EVM fragmentation than any of us as library authors would like.

I agree with this, but not with the idea that we should stop providing features.

frangio: Well, actually this library is designed to work on every reasonable chain, because of the Solidity fallback. My point was that a wrapper for the precompile is valuable.

pcaversaccio: > but not with the idea that we should not provide features unless they work on every chain.

I never said this. My main point here is that the precompile check is _not_ future proof as the return data of length 32 bytes could also come from other precompiles in the future on that address. I'm not saying this will happen, but I have seen too much shit over the last years to be paranoid enough about this edge case.

frangio:
Re: Transaction reverted: trying to deploy a contract whose code is too large

I think `allowUnlimitedContractSize` can be safely set to true always.

If a non-exposed contract is too large it should trigger the compiler warning and fail CI anyway:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c3f8b760ad9d0431f33e8303658ccbdcc1610f24/hardhat.config.js#L91-L97

ernestognw:
> I understand the logic for https://github.com/OpenZeppelin/openzeppelin-contracts/commit/9b240146a72383a551c2451c1854ed9c176246d5, but honestly that is an abomination. We surelly can do better.

I'd like to hear why this is a problem if using scratch space is safe. Regardless of how hacky it is I do think it's worse that we didn't discuss the config and is again there. I would appreciate if you communicate better your decisions. It's pretty much impossible to follow your preferences here!

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,304 @@
...
+        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
+    }
+
+    /**
+     * @dev Same as {verify}, but only the Solidity implementation is used.
+     */
+    function verifySolidity(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) internal view returns (bool) {
+        if (r == 0 || uint256(r) >= N || s == 0 || uint256(s) > HALF_N || !isOnCurve(qx, qy)) {
+            return false;
+        }
```
cairoeth, line 99: for a public key to be valid, it should:
1. not be point of infinity
2. have coordinates $(x, y)$ satisfy $0 <= x, y <= p$, and
3. satisfy the equation $y^2=x3+ax+b$ modulo $p$

on property one: the public key is used in `_preComputeJacobianPoints` by `points[0x01] = JPoint(px, py, 1); // 1,0 (p)`, which would never be the point of infinity. `_affineFromJacobian` uses the convention that $(0,0)$ represents the point of infinity, so in this case, the conversion in `_preComputeJacobianPoints` (L278) would be incorrect, but given that the point of infinity is not valid anyways, it's not an issue as we instead reject the public key $(0,0)$ via `isOnCurve`.

on property two: x and y coordinates of the public key get reduced to modulo p in calculations, so missing a check for property two means that the verify function will check the signature for the public key with coordinates $(x \bmod p, y \bmod p)$. This would allow for public keys $(x,y)$, where $x<2^{256} - p$ or $y<2^{256} - p$, which there would be another pair $(x^\prime,y^\prime)$ — such as $(x+y,y)$ — that can be used as a public key and for which signatures made for $(x,y)$ would also verify.

on property three: we are checking that it lies on the curve via the Weierstrass equation (`isOnCurve`).

i think properties 1 and 3 are ok, but missing 2.

Amxx: So should we add `uint256(qx) >= P || uint256(qy) >= P` ? Also, is it P or N?

cairoeth: yeah that should be good. it's P because we are using that for the modulus operations on places like `isOnCurve`  
```suggestion
    function verifySolidity(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) internal view returns (bool) {
        if (r == 0 || uint256(r) >= N || s == 0 || uint256(s) > HALF_N || uint256(qx) >= P || uint256(qy) >= P || !isOnCurve(qx, qy)) {
            return false;
        }
```

tests pass with this change, but i think it's useful to add a unit test for this property specifically, i can add it

cairoeth: i think it also makes sense to have this in `_tryVerifyNative`

Amxx: I added the check directly in `isOnCurve`

Amxx: Note: the public key is unlikelly to be arbitrary user input. It will be recorded/whitelisted onchain, by some admin. So this attack vector is not really critical. But still its good to add the (cheap) check in `isOnCurve`

contracts/utils/cryptography/P256.sol
ernestognw: Thanks for pointing out @pcaversaccio. I agree with the concern and now I think more about it I also agree we should explain the situation at the very minimum. We raised [a similar concern for Math.modexp](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3298#discussion_r1475446317) that ended up in [an explicit documentation mention](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/4032b42694ff6599b17ffde65b2b64d7fc8a38f8/contracts/utils/math/Math.sol#L317), so I would consider adding a note here.

Regardless, both RIP-7212 and EIP-7212 agree that the precompile "returns 1 in 32 bytes format". The current implementation reverts if the returndata can't be decoded as a boolean. In my opinion, it's unlikely that a precompile at `address(0x100)` will randomly return `true` but I recognize the possibility. Would you agree the decoding constrains the possibility?

In regards to why someone would use

contracts/utils/cryptography/P256.sol
ernestognw: > This would allow for public keys $(x,y)$, where $x<2^{256} - p$ or $y<2^{256} - p$, which there would be another pair $(x^\prime,y^\prime)$ — such as $(x+p,y)$ — that can be used as a public key and for which signatures made for $(x,y)$ would also verify.'

Although this is true, it's unclear to me whether you can create a signature for public keys $(x,y)$ if it's not on the curve. I still think checking is valuable and pretty much agree with @Amxx's conclusion that mostly public keys won't be arbitrary and the check is cheap.

cairoeth: `if it's not on the curve`: the problem is that you can find $(x^\prime,y^\prime)$ (like $(x+p,y)$ ) that are on the curve based on the original $(x,y)$ which are on the curve too -- it's like a signature malleability attack vector.

ernestognw: Got it, so it's the other way around. Still it's relevant that public keys are unlikely to be arbitrary, but let's keep the check regardless.

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,313 @@
...
+     * IMPORTANT: This function disallows signatures where the `s` value is above `N/2` to prevent malleability.
+     * To flip the `s` value, compute `s = N - s`.
+     */
+    function verify(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) internal view returns (bool) {
+        (bool valid, bool supported) = _tryVerifyNative(h, r, s, qx, qy);
+        return supported ? valid : verifySolidity(h, r, s, qx, qy);
+    }
+
+    /**
+     * @dev Same as {verify}, but it will revert if the required precompile is not available.
```
ernestognw, line 64: I guess this would address general concerns about the availability of the precompile

```suggestion
     * @dev Same as {verify}, but it will revert if the required precompile is not available.
     *
     * IMPORTANT: The use of this function relies on the RIP-7212 precompile to be available at `address(0x100)`.
     * Make sure the specified address contains the expected precompile, otherwise the returned value may be
     * misinterpreted as a positive boolean.
```

contracts/utils/cryptography/P256.sol
pcaversaccio: > Would you agree the decoding constrains the possibility?

Overall yes, but I want to raise that this might result in an unexpected behaviour since you don't expect a revert from a precompile usually (but no return data to indicate an error).

ernestognw: I agree it's unexpected behavior, but it's a manageable risk. 

>  I think a good compromise is to put a warning in the docs about the current situation?

I hear you with this and think an "IMPORTANT" block should be enough

pcaversaccio: sgtm!

contracts/utils/cryptography/P256.sol
```diff
@@ -0,0 +1,313 @@
...
+        uint256 u1 = mulmod(N - (uint256(h) % N), w, N);
+        uint256 u2 = mulmod(uint256(s), w, N);
+        (uint256 x, uint256 y) = _jMultShamir(points, u1, u2);
+        return (bytes32(x), bytes32(y));
+    }
+
+    /**
+     * @dev Checks if (x, y) are valid coordinates of a point on the curve.
+     * In particular this function checks that x <= P and y <= P.
+     */
```
cairoeth, line 142: add note on why (x,y) are not valid coordinates
```suggestion
    /**
     * @dev Checks if (x, y) are valid coordinates of a point on the curve.
     * In particular this function checks that x <= P and y <= P. There is no need to 
     * explicitly check if both x and y are 0, because the presence of 'b' coefficient
     * in the Weierstrass equation.
     */
```

cairoeth (commented):
nit

contracts/utils/cryptography/P256.sol
Amxx: what about:

```
* Make sure the any logic (code or precompile) deployed at that address is the expected one, otherwise the returned value may be
* misinterpreted as a positive boolean.
```

contracts/utils/cryptography/P256.sol
Amxx: Its a property of the secp265r1 curve that x = y = 0 is not on it. Given that the documentation already states that `(x, y) are valid coordinates of a point on the curve`,  I'm not sure its necessary to explicitelly mention that.

Amxx: And if we really want to add this, I'd say 

> because the `b` coefficient of the Weierstrass equation is not 0 in secp256r1

contracts/utils/cryptography/P256.sol
ernestognw: Yes, I'd be ok with that 👍🏻

ernestognw:
```suggestion
     * Make sure any logic (code or precompile) deployed at that address is the expected one, 
     * otherwise the returned value may be misinterpreted as a positive boolean.
```

ernestognw:
It turns out that after #5098, `forge coverage` hits the `Stack too deep` error again although tests compile without `--via-ir`.
We can use `--ir-minimum` in scripts/checks/coverage.sh and should work. I'll try it out

ernestognw (approved):
LGTM, as I pointed out, I think it's fine that forge coverage runs with `--ir-minimum`

PR #4883, opened 2024-02-07
Add variants of Array.sort for address[] and bytes32[]

### LINKED ISSUE #4882: Make _quicksort work with raw memory addresses
**🧐 Motivation**
Go fast.

**📝 Details**
Function `sort` from `utils/Arrays.sol` uses a private function `_quickSort`, which works on the array pointer and two indexes. It makes multiple array accesses, and for each of them it needs to calculate the offset of the actual data (`+32`) and the position of the array item (`+32*idx`).

An alternative would be to make `_quickSort` accept 2 raw memory pointers, and work with them. `sort` would call `assembly { let ptr := array } _quicksort(ptr + 32, ptr + 32 + array.length * 32>`. Instead of performing `idx++` to switch to the next item, `_quickSort` would do identically priced `ptr+=32`. Accessing the memory would be much cheaper, because it would be plain `mload`, without the index to pointer conversion. This approach would probably have less logic in assembly than the current one, because it would be limited to simple `mload`s and `mstore`s, probably just 2 one-line helper functions. The code would be comparably unsafe and would require similar expertise to understand when compared the current implementation, which does unchecked memory array operations anyway.

WDYT?

Amxx:
This PR also refactors the `_quicksort` private function to use "memory pointers". This was shown to save ~16% gas when sorting a random array of length 100.

Fixes #4882

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -13,7 +13,7 @@ library Arrays {
     using StorageSlot for bytes32;
 
     /**
-     * @dev Sort an array of integers (in memory) in increasing order.
+     * @dev Sort an array of integers (in memory) following the provided comparator function.
      *
      * This function does the sorting "in place", meaning that it overrides the input. The object is returned for
      * convenience, but that returned value can be discarded safely if the caller has a memory pointer to the array.
@@ -23,71 +23,119 @@ library Arrays {
      * when executing this as part of a transaction. If the array being sorted is too large, the sort operation may
      * consume more gas than is available in a block, leading to potential DoS.
      */
-    function sort(uint256[] memory array) internal pure returns (uint256[] memory) {
-        _quickSort(array, 0, array.length);
+    function sort(
+        bytes32[] memory array,
+        function(bytes32, bytes32) pure returns (bool) comp
+    ) internal pure returns (bytes32[] memory) {
+        _quickSort(array, 0, array.length, comp);
         return array;
     }
 
     /**
-     * @dev Variant of {sort(uint256[] memory)} that operates on an array of addresses (in memory).
-     */
-    function sort(address[] memory array) internal pure returns (address[] memory) {
-        uint256[] memory casted;
-        assembly {
-            casted := array
-        }
-        sort(casted);
-        return array;
-    }
-
-    /**
-     * @dev Variant of {sort(uint256[] memory)} that operates on an array of bytes32 (in memory).
+     * @dev Variant of {sort} that sorts an array of bytes32 in increassing order.
      */
     function sort(bytes32[] memory array) internal pure returns (bytes32[] memory) {
-        uint256[] memory casted;
-        assembly {
-            casted := array
-        }
-        sort(casted);
+        _quickSort(array, 0, array.length, _compIncr);
         return array;
     }
 
     /**
-     * @dev Performs a quick sort on an array in memory. The array is sorted in increasing order.
+     * @dev Variant of {sort} that sorts an array of address following a provided comparator function.
+     */
+    function sort(
+        address[] memory array,
+        function(address, address) pure returns (bool) comp
+    ) internal pure returns (address[] memory) {
+        function(bytes32, bytes32) pure returns (bool) castedComp;
+        bytes32[] memory castedArray;
+        assembly {
+            castedComp := comp
+            castedArray := array
+        }
+        sort(castedArray, castedComp);
+        return array;
+    }
+
+    /**
+     * @dev Variant of {sort} that sorts an array of address in increassing order.
+     */
+    function sort(address[] memory array) internal pure returns (address[] memory) {
+        bytes32[] memory castedArray;
+        assembly {
+            castedArray := array
+        }
+        sort(castedArray);
+        return array;
+    }
+
+    /**
+     * @dev Variant of {sort} that sorts an array of uint256 following a provided comparator function.
+     */
+    function sort(
+        uint256[] memory array,
+        function(uint256, uint256) pure returns (bool) comp
+    ) internal pure returns (uint256[] memory) {
+        function(bytes32, bytes32) pure returns (bool) castedComp;
+        bytes32[] memory castedArray;
+        assembly {
+            castedComp := comp
+            castedArray := array
+        }
+        sort(castedArray, castedComp);
+        return array;
+    }
+
+    /**
+     * @dev Variant of {sort} that sorts an array of uint256 in increassing order.
+     */
+    function sort(uint256[] memory array) internal pure returns (uint256[] memory) {
+        bytes32[] memory castedArray;
+        assembly {
+            castedArray := array
+        }
+        sort(castedArray);
+        return array;
+    }
+
+    /**
+     * @dev Performs a quick sort on an array in memory. The array is sorted following the `comp` comparator.
      *
      * Invariant: `i <= j <= array.length`. This is the case when initially called by {sort} and is preserved in
      * subcalls.
      */
-    function _quickSort(uint256[] memory array, uint256 i, uint256 j) private pure {
+    function _quickSort(
+        bytes32[] memory array,
+        uint256 i,
+        uint256 j,
+        function(bytes32, bytes32) pure returns (bool) comp
+    ) private pure {
         unchecked {
             // Can't overflow given `i <= j`
             if (j - i < 2) return;
 
             // Use first element as pivot
-            uint256 pivot = unsafeMemoryAccess(array, i);
+            bytes32 pivot = unsafeMemoryAccess(array, i);
             // Position where the pivot should be at the end of the loop
             uint256 index = i;
 
             for (uint256 k = i + 1; k < j; ++k) {
                 // Unsafe access is safe given `k < j <= array.length`.
-                if (unsafeMemoryAccess(array, k) < pivot) {
+                if (comp(unsafeMemoryAccess(array, k), pivot)) {
                     // If array[k] is smaller than the pivot, we increment the index and move array[k] there.
                     _swap(array, ++index, k);
                 }
             }
 
-            // Swap pivot into place
-            _swap(array, i, index);
-
-            _quickSort(array, i, index); // Sort the left side of the pivot
-            _quickSort(array, index + 1, j); // Sort the right side of the pivot
+            _swap(array, i, index); // Swap pivot into place
+            _quickSort(array, i, index, comp); // Sort the left side of the pivot
+            _quickSort(array, index + 1, j, comp); // Sort the right side of the pivot
         }
     }
 
     /**
      * @dev Swaps the elements at positions `i` and `j` in the `arr` array.
      */
-    function _swap(uint256[] memory arr, uint256 i, uint256 j) private pure {
+    function _swap(bytes32[] memory arr, uint256 i, uint256 j) private pure {
         assembly {
             let start := add(arr, 0x20) // Pointer to the first element of the array
             let pos_i := add(start, mul(i, 0x20))
@@ -97,6 +145,13 @@ library Arrays {
             mstore(pos_i, val_j)
             mstore(pos_j, val_i)
         }
+    }
+
+    /**
+     * @dev Comparator for sorting arrays in increassing order.
+     */
+    function _compIncr(bytes32 a, bytes32 b) private pure returns (bool) {
+        return uint256(a) < uint256(b);
     }
 
     /**
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -35,7 +35,7 @@ library Arrays {
      * @dev Variant of {sort} that sorts an array of bytes32 in increassing order.
      */
     function sort(bytes32[] memory array) internal pure returns (bytes32[] memory) {
-        sort(array, _defaultComp);
+        _quickSort(array, 0, array.length, _defaultComp);
         return array;
     }
 
@@ -46,13 +46,7 @@ library Arrays {
         address[] memory array,
         function(address, address) pure returns (bool) comp
     ) internal pure returns (address[] memory) {
-        function(bytes32, bytes32) pure returns (bool) castedComp;
-        bytes32[] memory castedArray;
-        assembly {
-            castedComp := comp
-            castedArray := array
-        }
-        sort(castedArray, castedComp);
+        _quickSort(_castToBytes32Array(array), 0, array.length, _castToBytes32Comp(comp));
         return array;
     }
 
@@ -60,11 +54,7 @@ library Arrays {
      * @dev Variant of {sort} that sorts an array of address in increassing order.
      */
     function sort(address[] memory array) internal pure returns (address[] memory) {
-        bytes32[] memory castedArray;
-        assembly {
-            castedArray := array
-        }
-        sort(castedArray);
+        _quickSort(_castToBytes32Array(array), 0, array.length, _defaultComp);
         return array;
     }
 
@@ -75,13 +65,7 @@ library Arrays {
         uint256[] memory array,
         function(uint256, uint256) pure returns (bool) comp
     ) internal pure returns (uint256[] memory) {
-        function(bytes32, bytes32) pure returns (bool) castedComp;
-        bytes32[] memory castedArray;
-        assembly {
-            castedComp := comp
-            castedArray := array
-        }
-        sort(castedArray, castedComp);
+        _quickSort(_castToBytes32Array(array), 0, array.length, _castToBytes32Comp(comp));
         return array;
     }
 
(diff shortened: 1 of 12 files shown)


test/utils/structs/DoubleEndedQueue.test.js
```diff
@@ -8,13 +8,7 @@ async function fixture() {
...
   /** Rebuild the content of the deque as a JS array. */
   const getContent = () =>
-    mock.$length(0).then(length =>
-      Promise.all(
-        Array(Number(length))
-          .fill()
-          .map((_, i) => mock.$at(0, i)),
-      ),
-    );
+    mock.$length(0).then(length => Promise.all(Array.from({ length }, (_, i) => mock.$at(0, i))));
```
Amxx, line 11:
```suggestion
    mock.$length(0).then(length => Promise.all(Array.from({ length: Number(length) }, (_, i) => mock.$at(0, i))));
```

contracts/utils/Arrays.sol
```diff
@@ -75,6 +127,43 @@ library Arrays {
...
+    }
+
+    /// @dev Helper: low level cast address comp function to bytes32 comp function
+    function _castToBytes32Comp(
+        function(uint256, uint256) pure returns (bool) input
+    ) private pure returns (function(bytes32, bytes32) pure returns (bool) output) {
+        assembly {
+            output := input
+        }
+    }
```
ernestognw, line 165: Big 🧠

contracts/utils/Arrays.sol
```diff
@@ -75,6 +127,43 @@ library Arrays {
         }
     }
 
+    /// @dev Comparator for sorting arrays in increassing order.
+    function _defaultComp(bytes32 a, bytes32 b) private pure returns (bool) {
+        return uint256(a) < uint256(b);
+    }
```
ernestognw, line 133: I wonder if the default comp function should have `uint256, uint256` arguments. It's weird that we convert all comparison functions to `bytes32` and at the end we compare `uint256`.

The benefit is more semantical, I think it's easier to follow that a function compares numbers and not bytes.

Not sure if it's even possible though, there may be a casting limitation I'm missing

test/utils/Arrays.test.js
```diff
@@ -154,22 +116,79 @@ describe('Arrays', function () {
...
+              this.elements = Array.from({ length }, generators[type]);
+            });
+
+            it('sort array', async function () {
+              // nothing to do here, beforeEach and afterEach already take care of everything.
+            });
 
-          beforeEach(async function () {
-            Object.assign(this, await loadFixture(fixture));
+            if (length > 1) {
```
ernestognw, line 156: what happens if `length <= 1`? We should test that case too

ernestognw (commented):
Looking good

test/utils/Arrays.test.js
Amxx: if length is 0 or 1, then any random array is already: shuffled, sorted, and reverse sorted (simultaneously).
No need to check.

Amxx: If length is 0 or 1, we do the "normal" test ... and that is it ... not need for particaular cases

contracts/utils/Arrays.sol
Amxx: I wanted the default to be about bytes32[], because I feel that is the most "fundamental" type. This is possibly going to change if we do #4883 and operate on pointers directly.

test/utils/Arrays.test.js
ernestognw: Right. The logic in the `afterEach is a bit disconcertant`.

I'd suggest having it below the `beforeEach` for clarity. But not a blocker.

contracts/utils/Arrays.sol
Amxx: I changes the base type (returned by `_mload`, and used for the comparator) to uint256

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
+    /**
+     * @dev Variant of {sort} that sorts an array of bytes32 in increassing order.
+     */
+    function sort(bytes32[] memory array) internal pure returns (bytes32[] memory) {
+        sort(_castToUint256Array(array), _defaultComp);
+        return array;
+    }
+
+    /**
+     * @dev Performs a quick sort of a segment of memory. The segment sorted starts at `begin` (inclusive), and ends
```
Amxx, line 87:
```suggestion
     * @dev Performs a quick sort of a segment of memory. The segment sorted starts at `begin` (inclusive), and stops
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
-                // Unsafe access is safe given `k < j <= array.length`.
-                if (unsafeMemoryAccess(array, k) < pivot) {
-                    // If array[k] is smaller than the pivot, we increment the index and move array[k] there.
-                    _swap(array, ++index, k);
+            uint256 pos = begin;
+
+            for (uint256 it = begin + 0x20; it < end; it += 0x20) {
+                if (comp(_mload(it), pivot)) {
+                    // If array[k] is smaller than the pivot, we increment the position of the pivot and move array[k]
+                    // there.
```
Amxx, line 104:
```suggestion
                    // If value stored at the iterator's position is smaller than the pivot, we increment the position 
                    // of the pivot and move the value there.
```

contracts/utils/Arrays.sol
Amxx: resolved: it turns out `<` is available for `bytes32`

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
+    ) internal pure returns (bytes32[] memory) {
+        uint256 begin;
+        uint256 end;
+        /// @solidity memory-safe-assembly
+        assembly {
+            begin := add(array, 0x20)
+            end := add(begin, mul(mload(array), 0x20))
+        }
+        _quickSort(begin, end, comp);
+        return array;
```
Amxx, line 38: An alternative would be
```suggestion
        uint256 ptr;
        /// @solidity memory-safe-assembly
        assembly {
            ptr := add(array, 0x20)
        }
        unchecked {
            _quickSort(ptr, ptr + array.length * 0x20, comp);
        }
        return array;
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
-     * @dev Swaps the elements at positions `i` and `j` in the `arr` array.
+     * @dev Swaps the elements memory location `ptr1` and `ptr2`.
      */
-    function _swap(uint256[] memory arr, uint256 i, uint256 j) private pure {
+    function _swap(uint256 ptr1, uint256 ptr2) private pure {
+        assembly {
+            let val1 := mload(ptr1)
+            let val2 := mload(ptr2)
+            mstore(ptr1, val2)
+            mstore(ptr2, val1)
```
Amxx, line 133:
```suggestion
            let value1 := mload(ptr1)
            let value2 := mload(ptr2)
            mstore(ptr1, value2)
            mstore(ptr2, value1)
```

contracts/utils/Arrays.sol
ernestognw: I prefer the current state. If we're trying to preserver readability I think the original is best. Although I'd prefer a non-assembly alternative

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
+        assembly {
+            begin := add(array, 0x20)
+            end := add(begin, mul(mload(array), 0x20))
+        }
+        _quickSort(begin, end, comp);
+        return array;
+    }
+
+    /**
+     * @dev Variant of {sort} that sorts an array of bytes32 in increassing order.
```
ernestognw, line 42:
```suggestion
     * @dev Variant of {sort} that sorts an array of bytes32 in increasing order.
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
+        bytes32[] memory array,
+        function(bytes32, bytes32) pure returns (bool) comp
+    ) internal pure returns (bytes32[] memory) {
+        uint256 begin;
+        uint256 end;
+        /// @solidity memory-safe-assembly
+        assembly {
+            begin := add(array, 0x20)
+            end := add(begin, mul(mload(array), 0x20))
+        }
```
ernestognw, line 36: What about having these functions in Array.sol:

```solidity
function begin(bytes32[] memory array) internal pure returns (bytes32 ptr) {
  assembly { ptr := add(array, 0x20) }
} 

function end(bytes32[] memory array) internal pure returns (bytes32 ptr) {
  assembly { ptr := add(begin, mul(mload(array), 0x20)) }
} 
```

So we can do:

```solidity
_quickSort(array.start(), array.end());
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
+    function sort(
+        address[] memory array,
+        function(address, address) pure returns (bool) comp
+    ) internal pure returns (address[] memory) {
+        sort(_castToBytes32Array(array), _castToBytes32Comp(comp));
+        return array;
+    }
+
+    /**
+     * @dev Variant of {sort} that sorts an array of address in increassing order.
```
ernestognw, line 60:
```suggestion
     * @dev Variant of {sort} that sorts an array of address in increasing order.
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
+    function sort(
+        uint256[] memory array,
+        function(uint256, uint256) pure returns (bool) comp
+    ) internal pure returns (uint256[] memory) {
+        sort(_castToBytes32Array(array), _castToBytes32Comp(comp));
+        return array;
+    }
+
+    /**
+     * @dev Variant of {sort} that sorts an array of uint256 in increassing order.
```
ernestognw, line 79:
```suggestion
     * @dev Variant of {sort} that sorts an array of uint256 in increasing order.
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
     }
 
     /**
-     * @dev Performs a quick sort on an array in memory. The array is sorted in increasing order.
+     * @dev Performs a quick sort of a segment of memory. The segment sorted starts at `begin` (inclusive), and stops
+     * at end (exclusive). Sorting follows the `comp` comparator.
      *
-     * Invariant: `i <= j <= array.length`. This is the case when initially called by {sort} and is preserved in
-     * subcalls.
+     * Invariant: `begin <= end`. This is the case when initially called by {sort} and is preserved in subcalls.
```
ernestognw, line 90:
```suggestion
     * Invariant: `begin <= end <= array.length`. This is the case when initially called by {sort} and is preserved in subcalls.
```

I think it's important to keep `array.length` because we're using `unsafeMemoryAccess`. If we don't mention it at all, it's dangerous for someone to provide an arbitrary end. Actually, I'm leaning more towards a note:

```suggestion
     * Invariant: `begin <= end`. This is the case when initially called by {sort} and is preserved in subcalls.
     *
     * IMPORTANT: Memory locations between `begin` and `end` are not validated/zeroed. This function should
     * be used only if the limits are within a memory array.
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
-            for (uint256 k = i + 1; k < j; ++k) {
-                // Unsafe access is safe given `k < j <= array.length`.
-                if (unsafeMemoryAccess(array, k) < pivot) {
-                    // If array[k] is smaller than the pivot, we increment the index and move array[k] there.
-                    _swap(array, ++index, k);
+            uint256 pos = begin;
+
+            for (uint256 it = begin + 0x20; it < end; it += 0x20) {
+                if (comp(_mload(it), pivot)) {
+                    // If the value stored at the iterator's position is smaller than the pivot, we increment the
```
ernestognw, line 103:
```suggestion
                    // If the value stored at the iterator's position returns true at comparison, we increment the
```

contracts/utils/Arrays.sol
```diff
@@ -23,55 +23,151 @@ library Arrays {
...
+    function _swap(uint256 ptr1, uint256 ptr2) private pure {
+        assembly {
+            let value1 := mload(ptr1)
+            let value2 := mload(ptr2)
+            mstore(ptr1, value2)
+            mstore(ptr2, value1)
+        }
+    }
+
+    /// @dev Comparator for sorting arrays in increassing order.
```
ernestognw, line 137:
```suggestion
    /// @dev Comparator for sorting arrays in increasing order.
```

ernestognw (commented):
Currently we have 2 overrides for each function:

```solidity
function sort(<type>[] memory array) ... {}
function sort(<type>[] memory array, function(<type>, <type>) pure returns (bool) comp) ... {}
```

The default is in increasing order, (which I agree with), but I feel we should include the decreasing order too.

What do you think of defining an enum:

```solidity
enum Order {
  Ascending,
  Descending
}
```

So that it can be provided as an argument to the `sort` functions:

```solidity
function sort(<type>[] memory array, Order order) ... {}
```

This may require more discussion but I think is worth.

Amxx:
> What do you think of defining an enum:
> 
> ```solidity
> enum Order {
>   Ascending,
>   Descending
> }
> ```

I don't think we should restrict ordering to comparator that we know and implement. If someone wants to do "all the odd values, in increassing order, then all the even values, in decreassing order", so that the array looks like [1,3,5,7,9,10,8,6,4,2,0] ... then I think that should be possible. You may also want to sort array of bytes32 in increassing order of the right most part, discarding the first 16bytes for the ordering.

This is possible in many langages, [including Javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), and I think we should support that as well.

contracts/utils/Arrays.sol
Amxx: - `end` is a memory pointer, it is not smaller than the length.

I'm ok with the second proposal is ok to me (though I would rewrite "validated/zeroed" as "allocated".

contracts/utils/Arrays.sol
Amxx: What about 
```suggestion
                    // If the value stored at the iterator's position comes before the pivot, we increment the
```

ernestognw: Right, that sounds good

ernestognw:
> > What do you think of defining an enum:
> > ```solidity
> > enum Order {
> >   Ascending,
> >   Descending
> > }
> > ```
> 
> I don't think we should restrict ordering to comparator that we know and implement. If someone wants to do "all the odd values, in increassing order, then all the even values, in decreassing order", so that the array looks like [1,3,5,7,9,10,8,6,4,2,0] ... then I think that should be possible. You may also want to sort array of bytes32 in increassing order of the right most part, discarding the first 16bytes for the ordering.
> 
> This is possible in many langages, [including Javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), and I think we should support that as well.

Yes, I'm not proposing throwing the `comp` function pointer, but rather providing the two common use cases.
Actually, we can have only these 2 overloads:

```
function sort(<type>[] memory array, Order order) ... {}
function sort(<type>[] memory array, function(<type>, <type>) pure returns (bool) comp) ... {}
```

contracts/utils/Arrays.sol
Amxx: Also, this is a private function, so we should consider that documentation is mostly (only?) for us. Until recently, we did not even document these private functions.

contracts/utils/Arrays.sol
Amxx: done.

Amxx:
> Yes, I'm not proposing throwing the `comp` function pointer, but rather providing the two common use cases. Actually, we can have only these 2 overloads:
> 
> ```solidity
> function sort(<type>[] memory array, Order order) ... {}
> function sort(<type>[] memory array, function(<type>, <type>) pure returns (bool) comp) ... {}
> ```
> 
> The comparator function is still customizable

Sorry, I did not understand initially. 

I'm not sure. I like the fact that default (no second argument) is increasing order, just like its the case in js (~ish), c++ and python.

contracts/utils/Arrays.sol
ernestognw: Yeah this is true, although the original `sort` documentation is not as clear as this one. I think that's ok, but I'd rather be clear since all other functions are entry points to this

ernestognw (approved):
Some ideas like the `Order` enum and `begin/end` functions can improve later. This PR is good right now.

ernestognw:
> I'm not sure. I like the fact that default (no second argument) is increasing order, just like its the case in js (~ish), c++ and python.

I also like it. I'm not convinced we should have 3 overloads. Let's keep it for now, and reconsider if people ever ask for it.

ernestognw (approved):

PR #4887, opened 2024-02-08
Process and verify merkle proofs (and multiproof) with custom hash function

### LINKED ISSUE #4886: Add a MerkleProof.verify function that support arbitrary internal hashes
Some devs may want to verify merkle proof where internal nodes were computed using non-standard hashes.

Amxx:
Fixes #4886

The hasher should be view, but the existing versions are pure. This is resulved by
- Keeping the existing versions (pure)
- Add new (view) versions that use the a custom hasher
- Both for memory and calldata
- Both for simple proofs and multiproofs

This large amount of code duplication is handled through procedural generation

--- 
- [x] Tests
- [x] Documentation


ernestognw (commented):
I pushed some nits but the PR looks good imo. It may just need to be tested with custom hashing functions but that can't be done until we release the merkle tree library.

~I'd be fine leaving an issue and merging, wdyt @Amxx?~ I see codecov was flagging a drop in coverage. Investigating both #4977 and this drop

test/utils/cryptography/MerkleProof.test.js
ernestognw, line 1: To get 100% coverage we'll need to test with a custom hashing function. I think the best we can do is test it out again with a mock that uses the commutative `keccak256`function again and leave a TODO, so that we replace it when we release the Typescript library

ernestognw (dismissed):
LGTM, I left a TODO for when we release the Typescript library

test/utils/cryptography/MerkleProof.test.js
Amxx: refactored the tests to use two different (and commutative) hashes

contracts/mocks/MerkleProofCustomHashMock.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {MerkleProof} from "../utils/cryptography/MerkleProof.sol";
+
+// this count be a library, but then we would have to add it to the Stateless.sol mock for upgradeable tests
```
Amxx, line 7:
```suggestion
// This could be a library, but then we would have to add it to the Stateless.sol mock for upgradeable tests
```

contracts/utils/structs/MerkleTree.sol
```diff
@@ -19,6 +19,11 @@ import {Panic} from "../Panic.sol";
  * * Zero value: The value that represents an empty leaf. Used to avoid regular zero values to be part of the tree.
  * * Hashing function: A cryptographic hash function used to produce internal nodes. Defaults to {Hashes-commutativeKeccak256}.
  *
+ * proving the inclusion of a leaf is not possible with the {MerkleProof} library since it only supports
+ * NOTE: Building trees using non-commutative hashing functions (i.e. `H(a, b) != H(b, a)`) is supported. However,
```
Amxx, line 23:
```suggestion
 * NOTE: Building trees using non-commutative hashing functions (i.e. `H(a, b) != H(b, a)`) is supported. However,
 * proving the inclusion of a leaf is not possible with the {MerkleProof} library since it only supports
```

ernestognw (approved):
Thanks for updating the tests with the new Merkle Tree release.
LGTM

PR #4893, opened 2024-02-12
Add `bytes memory` version of `Math.modExp`

### LINKED ISSUE #4880: Extend `Math.modExp` to support `bytes memory`
From #1985

> I'm not sure if https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3298 's implementation is actually useful in the majority of the cases. The main reason why I see developers using `modexp` is because they're working with RSA signatures (not sure if there are other use cases).
>
> When working with RSA signatures, a private key consists of a modulus `m` and an exponent `e`. So for producing a signature, a signer should get a digest of the signed message and then do `modexp(digest, e, m)` in order to get a signature.
>
> Similarly, for verifying it's needed to do `modexp(signature, e, m)`.
> 
> Both operations don't fit in an `uint256` since the values of both `e` and `m` are typically 3072 bits (for 128-bit security).
>
> Although RSA signatures have never been popular on Ethereum, with the new Account Abstraction trend it can be used as an alternative authorization mechanism. So, I'm afraid that this implementation wouldn't solve that particular use case.
If there's another use case justifying this current implementation, we can consider merging it.

There are still use cases for higher results than `uint256`. And the account abstraction narrative might led us to prioritize this

ernestognw:
Fixes #4880
Fixes LIB-1209

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -341,7 +341,7 @@ library Math {
     function modExp(bytes memory b, bytes memory e, bytes memory m) internal view returns (bytes memory) {
         (bool success, bytes memory result) = tryModExp(b, e, m);
         if (!success) {
-            if (_zeroArray(m)) {
+            if (_zeroBytes(m)) {
                 Panic.panic(Panic.DIVISION_BY_ZERO);
             } else {
                 revert Address.FailedInnerCall();
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -336,6 +336,55 @@ library Math {
     }
 
     /**
+     * @dev Variant of {modExp} that supports inputs of arbitrary length.
+     */
+    function modExp(bytes memory b, bytes memory e, bytes memory m) internal view returns (bytes memory) {
+        (bool success, bytes memory result) = tryModExp(b, e, m);
+        if (!success) {
+            if (_zeroArray(m)) {
+                Panic.panic(Panic.DIVISION_BY_ZERO);
+            } else {
+                revert Address.FailedInnerCall();
+            }
+        }
+        return result;
+    }
+
+    /**
+     * @dev Variant of {tryModExp} that supports inputs of arbitrary length.
+     */
+    function tryModExp(
+        bytes memory b,
+        bytes memory e,
+        bytes memory m
+    ) internal view returns (bool success, bytes memory result) {
+        if (_zeroArray(m)) return (false, new bytes(0));
+
+        // Encode call args and move the free memory pointer
+        bytes memory args = abi.encodePacked(b.length, e.length, m.length, b, e, m);
+
+        // Given result <= modulus
+        result = new bytes(m.length);
+
+        /// @solidity memory-safe-assembly
+        assembly {
+            success := staticcall(gas(), 0x05, add(args, 0x20), mload(args), add(result, 0x20), mload(m))
+        }
+    }
+
+    /**
+     * @dev Returns whether the provided array is zero.
+     */
+    function _zeroArray(bytes memory array) private pure returns (bool) {
+        for (uint256 i; i < array.length; ++i) {
+            if (array[i] != 0) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /**
      * @dev Returns the square root of a number. If the number is not a perfect square, the value is rounded
      * towards zero.
      *
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -361,14 +361,21 @@ library Math {
         if (_zeroArray(m)) return (false, new bytes(0));
 
         // Encode call args and move the free memory pointer
-        bytes memory args = abi.encodePacked(b.length, e.length, m.length, b, e, m);
-
-        // Given result <= modulus
-        result = new bytes(m.length);
+        result = abi.encodePacked(b.length, e.length, m.length, b, e, m);
 
         /// @solidity memory-safe-assembly
         assembly {
-            success := staticcall(gas(), 0x05, add(args, 0x20), mload(args), add(result, 0x20), mload(m))
+            let mLen := mload(m)
+            // Write result on top of args to avoid allocating extra memory.
+            // | Offset    | Content      | Content (Hex)                                                      |
+            // |-----------|--------------|--------------------------------------------------------------------|
+            // | 0x00:0x1f | args length  | 0x<.......................................20+20+20+bLen+eLen+mLen> |
+            // | 0x20+mLen | result       | 0x<........................................................result> |
+            // | 0x..:0x.. | dirty bytes  | 0x<............................................20+20+20+bLen+eLen> |
+            success := staticcall(gas(), 0x05, add(result, 0x20), mload(result), add(result, 0x20), mLen)
+            // Overwrite the length.
+            // result.length > returndatasize() is guaranteed because returndatasize() == m.length
+            mstore(result, mLen)
         }
     }
 
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -349,7 +349,7 @@ library Math {
         bytes memory e,
         bytes memory m
     ) internal view returns (bool success, bytes memory result) {
-        if (_zeroBytes(m)) return (false, result);
+        if (_zeroBytes(m)) return (false, new bytes(0));
 
         uint256 mLen = m.length;
 
@@ -358,16 +358,14 @@ library Math {
 
         /// @solidity memory-safe-assembly
         assembly {
+            let dataPtr := add(result, 0x20)
             // Write result on top of args to avoid allocating extra memory.
-            // | Offset    | Content      | Content (Hex)                                                      |
-            // |-----------|--------------|--------------------------------------------------------------------|
-            // | 0x00:0x1f | args length  | 0x<.......................................20+20+20+bLen+eLen+mLen> |
-            // | 0x20+mLen | result       | 0x<........................................................result> |
-            // | 0x..:0x.. | dirty bytes  | 0x<............................................20+20+20+bLen+eLen> |
-            success := staticcall(gas(), 0x05, add(result, 0x20), mload(result), add(result, 0x20), mLen)
+            success := staticcall(gas(), 0x05, dataPtr, mload(result), dataPtr, mLen)
             // Overwrite the length.
             // result.length > returndatasize() is guaranteed because returndatasize() == m.length
             mstore(result, mLen)
+            // Set the memory pointer after the returned data.
+            mstore(0x40, add(dataPtr, mLen))
         }
     }
 
@@ -375,7 +373,7 @@ library Math {
      * @dev Returns whether the provided byte array is zero.
      */
     function _zeroBytes(bytes memory byteArray) private pure returns (bool) {
-        for (uint256 i; i < byteArray.length; ++i) {
+        for (uint256 i = 0; i < byteArray.length; ++i) {
             if (byteArray[i] != 0) {
                 return false;
             }
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -360,12 +360,13 @@ library Math {
     ) internal view returns (bool success, bytes memory result) {
         if (_zeroArray(m)) return (false, new bytes(0));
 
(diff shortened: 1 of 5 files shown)


contracts/utils/math/Math.sol
```diff
@@ -335,6 +335,55 @@ library Math {
...
+        // Encode call args and move the free memory pointer
+        bytes memory args = abi.encodePacked(b.length, e.length, m.length, b, e, m);
+
+        // Given result <= modulus
+        result = new bytes(m.length);
+
+        /// @solidity memory-safe-assembly
+        assembly {
+            success := staticcall(gas(), 0x05, add(args, 0x20), mload(args), add(result, 0x20), mload(m))
+        }
```
Amxx, line 372: This is going to leak more memory than needed.

- `args` are encoded, on new memory, that is allocate at the free memory pointer (0x40)
- then `result` is allocated (after args)
- in the end, `result` is returned, and `args` is never dealocated.

This means that if you do that in a loop, your memory pointer moves, and the gas cost increase quadratically.

I did implement a version that leaks less memory (only the result array is allocated) [here](https://github.com/Amxx/openzeppelin-contracts/blob/feature/math/modExp-bytes/contracts/utils/math/Math.sol#L356-L381). Any reason to not use that version?

ernestognw: This is becoming a guideline. Generally we should consider memory leaking in almost everything.

I'm out of home rn but will update once back. Thanks for flagging!

ernestognw:
I updated `tryModExp` to use the `result` as a pointer to avoid allocating extra memory. The function should be memory safe because it's not violating the memory safetiness rules and the only effect is that it's leaving dirty bytes between the result and the next memory pointer. Those shouldn't be reachable but I wonder if we should warn about it.

ernestognw:
Tests passing but coverage dropped. I tried mocking the `address(5)` with Hardhat's `setCode` utility but it didn't work:

```
it('is correctly reverting if the underlying call reverts', async function () {
  const base = 3n;
  const exponent = 200n;
  const modulus = 50n;

  await setCode('0x0000000000000000000000000000000000000005', '0xFD'); // Straight revert

  await expect(this.mock[this.fn](base, exponent, modulus)).to.be.revertedWithPanic(PANIC_CODES.DIVISION_BY_ZERO);
});
```

The revert's missing stack arguments but it should revert anyway and it's not.
I see no way of testing a failed underlying call to the modExp address and that's why the coverage is dropping. Not sure how to proceed tbh

ernestognw:
Alternatively, we can revert in any case `try` fails.

```solidity
function modExp(bytes memory b, bytes memory e, bytes memory m) internal view returns (bytes memory) {
    (bool success, bytes memory result) = tryModExp(b, e, m);
    if (!success) {
        Panic.panic(Panic.DIVISION_BY_ZERO);
    }
    return result;
}
```

Amxx:
```
function modExp(bytes memory b, bytes memory e, bytes memory m) internal view returns (bytes memory) {
    (bool success, bytes memory result) = tryModExp(b, e, m);
    if (!success) {
        Panic.panic(Panic.DIVISION_BY_ZERO);
    }
    return result;
}
```
I'm 100% ok with that. If we do it here, we should do it in both versions of `modExp` (uint256 and bytes memory).

contracts/utils/math/Math.sol
```diff
@@ -335,6 +335,63 @@ library Math {
...
+            // | Offset    | Content      | Content (Hex)                                                      |
+            // |-----------|--------------|--------------------------------------------------------------------|
+            // | 0x00:0x1f | args length  | 0x<.......................................20+20+20+bLen+eLen+mLen> |
+            // | 0x20+mLen | result       | 0x<........................................................result> |
+            // | 0x..:0x.. | dirty bytes  | 0x<............................................20+20+20+bLen+eLen> |
+            success := staticcall(gas(), 0x05, add(result, 0x20), mload(result), add(result, 0x20), mLen)
+            // Overwrite the length.
+            // result.length > returndatasize() is guaranteed because returndatasize() == m.length
+            mstore(result, mLen)
+        }
```
Amxx, line 380: This still leaks:

`result = abi.encodePacked(b.length, e.length, mLen, b, e, m);`  allocates a memory array that is 
- 0x20 for the length (of the bytes being encoded)
- 0x20 for the b.length
- 0x20 for the e.length
- 0x20 for the m.length
- 0x20 * b.length for b
- 0x20 * e.length for e
- 0x20 * m.length for m

For a total of `0x80 + 0x20 * (b.length + e.length + m.length)`. That is by how much the free memory pointer is moved.

For result, we only need:
- 0x20 for the length of result (that is value of m.length)
- 0x20 * m.length for the result.

So the current version moves the free memory ptr by `0x60 + 0x20 * (b.length + e.length)` more than needed. That is memory being allocated for no usage.

ernestognw: Yes, that's right. But I'm unsure of allocating the pointer after `m`. All the other elements haven't been zeroed out. That's a dangerous side effect.

Amxx: > All the other elements haven't been zeroed out. That's a dangerous side effect.

IMO that is not dangerous. Its clear that there no one should expect the bytes after the free memory ptr to be zeroed. 

See the warning [here](https://docs.soliditylang.org/en/latest/internals/layout_in_memory.html)
Capture d’écran du 2024-02-13 15-53-03

A typical example is [this function](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/Clones.sol#L68-L84). We write a lot of stuff in the free memory space, use that to do some computations, and never clean it. The space remains unallocated and is not zero.

contracts/utils/math/Math.sol
```diff
@@ -335,6 +330,59 @@ library Math {
...
+        result = abi.encodePacked(b.length, e.length, mLen, b, e, m);
+
+        /// @solidity memory-safe-assembly
+        assembly {
+            // Write result on top of args to avoid allocating extra memory.
+            // | Offset    | Content      | Content (Hex)                                                      |
+            // |-----------|--------------|--------------------------------------------------------------------|
+            // | 0x00:0x1f | args length  | 0x<.......................................20+20+20+bLen+eLen+mLen> |
+            // | 0x20+mLen | result       | 0x<........................................................result> |
+            // | 0x..:0x.. | dirty bytes  | 0x<............................................20+20+20+bLen+eLen> |
```
Amxx, line 366: I'm not sure this table really helps. The last line in particular. Offsets are not specified, and the hex content is not accurate.

contracts/utils/math/Math.sol
```diff
@@ -335,6 +330,59 @@ library Math {
...
+            // result.length > returndatasize() is guaranteed because returndatasize() == m.length
+            mstore(result, mLen)
+        }
+    }
+
+    /**
+     * @dev Returns whether the provided byte array is zero.
+     */
+    function _zeroBytes(bytes memory byteArray) private pure returns (bool) {
+        for (uint256 i; i < byteArray.length; ++i) {
```
Amxx, line 378:
```suggestion
        for (uint256 i = 0; i < byteArray.length; ++i) {
```

test/utils/math/Math.test.js
```diff
@@ -354,20 +339,126 @@ describe('Math', function () {
...
+        const exponent = 200n;
+        const modulus = 0n;
+
+        await expect(this.mock[this.fn](toBeHex(base), toBeHex(exponent), toBeHex(modulus))).to.be.revertedWithPanic(
+          PANIC_CODES.DIVISION_BY_ZERO,
+        );
+      });
 
-      expect(await this.mock.$modExp(base, exponent, modulus)).to.equal(base ** exponent % modulus);
+      for (const [baseExp, exponentExp, modulusExp] of product(range(0, 24, 4), range(0, 24, 4), range(0, 256, 64))) {
```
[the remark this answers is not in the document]
Amxx: These values are actually pretty small. The max value they take is
```
b: 1048576
e: 1048576
m: 6277101735386680763835789423207666416102355444464034512896
```
In comparaison, type(uint256).max is
```
115792089237316195423570985008687907853269984665640564039457584007913129639935
```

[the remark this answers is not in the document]
Amxx: wich makes me wonder:
- if all these fit in a uint256, why only test them by the bytes version, and not with the uint256 version
- This looks like fuzzing, but with very specific values (only powers of 2). 

Since the values used here are not bigger than then one supported by the foundry fuzzing, are way slower to run, and do not cover a bigger space ... do we need that at all ? I'd remove this part of the tests.

ernestognw:
Somewhat agree with [12ea8eb](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4893/commits/12ea8ebaabe92003fb0e3540d96428f3d6c6359b) but I think we should keep those test cases. The main reason is that the only benefit from the ethers tests vs Foundry is that we can test outside of the `uint256` space in Javascript.

I'll revert the change but reconfigure the function to avoid performing excessive operations.

ernestognw:
I readded the large `bytes` tests with these params:

```
product(
  range(4, 20, 6).map(i => 2n ** BigInt(i)),
  range(4, 20, 6).map(i => 2n ** BigInt(i)),
  range(256, 512, 64).map(i => 2n ** BigInt(i)),
)
```

The whole `Math.test.js` still runs in 5s tested locally, and we're still using decently large numbers for the tests.

contracts/utils/math/Math.sol
ernestognw: Yeah, I started with a bigger table when using your example. I removed it.

contracts/utils/math/Math.sol
```diff
@@ -335,6 +330,59 @@ library Math {
...
+
+    /**
+     * @dev Variant of {tryModExp} that supports inputs of arbitrary length.
+     */
+    function tryModExp(
+        bytes memory b,
+        bytes memory e,
+        bytes memory m
+    ) internal view returns (bool success, bytes memory result) {
+        if (_zeroBytes(m)) return (false, result);
```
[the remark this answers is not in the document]
ernestognw: Well, `a / 0 == undefined`.

Let's keep `new bytes(0)`, but imo the reason why we shouldn't care about reallocating the memory pointer if we're returning `new bytes(0)` should be because the memory leak is minimal (just the size slot) and not because the memory may be dirty.

contracts/utils/math/Math.sol
ernestognw: > A typical example is [this function](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/Clones.sol#L68-L84). We write a lot of stuff in the free memory space, use that to do some computations, and never clean it. The space remains unallocated and is not zero.

Right, this example convinced me. My main worry is that the modexp implementation may come as a report through Immunefi and I would've taken it as valid. The Clones example is evidence this is not an issue.

Note that I don't agree that developers should always know the default Solidity behavior. Imo they should know if they explicitly decide to deal with a particular feature that has side effects, but here the side effects are not obvious for regular language users.

Solving.

contracts/utils/math/Math.sol
```diff
@@ -335,6 +330,57 @@ library Math {
...
+            let dataPtr := add(result, 0x20)
+            // Write result on top of args to avoid allocating extra memory.
+            success := staticcall(gas(), 0x05, dataPtr, mload(result), dataPtr, mLen)
+            // Overwrite the length.
+            // result.length > returndatasize() is guaranteed because returndatasize() == m.length
+            mstore(result, mLen)
+            // Set the memory pointer after the returned data.
+            mstore(0x40, add(dataPtr, mLen))
+        }
+    }
```
Amxx, line 370: This works assuming the current solidity behavior:
- result is always allocated (at the free memory pointer)

If solidity ever changes its behavior, reusing memory gaps to put new arrays, then this would set the free memory pointer in an unsafe way.

That is unlikelly to happen for a white. Definitelly not in 0.8.x. So I think its ok.

Amxx (approved):

contracts/utils/math/Math.sol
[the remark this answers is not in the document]
Amxx: Technically, is you want to do `x = new bytes(0)`, you don't even need to allocate new data. It would be enough to point to 0x60 (that should always contain zero).

Its a memory array, so you can't push anyway.

PR #4895, opened 2024-02-13
Deduplicate code in `scripts/helpers` and `test/helpers/iterate`

Amxx:
#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/scripts/helpers.js
+++ b/scripts/helpers.js
@@ -30,7 +30,7 @@ module.exports = {
 
   // ================================================ Object helpers =================================================
 
-  // Create a new object by mapping the values trough a function, keeping the keys
+  // Create a new object by mapping the values through a function, keeping the keys
   // Example: mapValues({a:1,b:2,c:3}, x => x**2) → {a:1,b:4,c:9}
   mapValues: (obj, fn) => Object.fromEntries(Object.entries(obj).map(([k, v]) => [k, fn(v)])),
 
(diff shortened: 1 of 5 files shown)


ernestognw (approved):
I checked where the functions are used in the `scripts` folder and so far:

- `chunk`: Unused
- `product`: Unused
- `range`: In the SafeCast template
- `unique`: Unused
- `zip`: Unused
- `mapValues`: Unused
- `capitalize`: In multiple templates

Since these are used within automatic generation, these are already checked in the CI so I'm approving.

However, I wonder if it makes sense to keep the helpers in where they're used the most (i.e `tests/`)

ernestognw:
CI failing due to #4896. Rerunning

Amxx:
> However, I wonder if it makes sense to keep the helpers in where they're used the most (i.e `tests/`)

I considered keeping all that code in `test/helpers/iterate.js` and have the generation script point to that. It felt strange to have "production" scripts rely on testing code... but at the same time, `iterate.js` sounds like a better description of what the file contains.

Amxx:
[For the record, the "chunk" helpers was used in Forta's early days.](https://github.com/forta-network/forta-contracts/blob/master/scripts/deployments/deploy-token.js#L126-L149)

We wanted to relay a lot of transaction trough multicall, but not all at once (we did that in blocks of 16 so that we benefit from multicall while limiting the size of the multicall transaction)

ernestognw:
> It felt strange to have "production" scripts rely on testing code...

Right, that feels weird. I think this is better anyway since we don't ship those scripts. And they're checked in the CI.

ernestognw (approved):
LGTM

PR #4899, opened 2024-02-14
Update chai matchers

RenanSouza2:
Hey this PR updates the hardhat-chai-matchers library form version 2.0.3 to 2.0,5,

Rationale: version 2.0.5 also includes the ethers.Typed in the address verification, needed to simplify vesting wallet tests

version 2.0.4 changed the name of an internal variable from `ARITHMETIC_UNDER_OR_OVERFLOW` to `ARITHMETIC_OVERFLOW`

#### PR Checklist

- [ ] Tests
- [ ] Documentation


test/utils/Panic.test.js
```diff
@@ -15,7 +15,7 @@ describe('Panic', function () {
   for (const [name, code] of Object.entries({
     GENERIC: 0x0,
     ASSERT: PANIC_CODES.ASSERTION_ERROR,
-    UNDER_OVERFLOW: PANIC_CODES.ARITHMETIC_UNDER_OR_OVERFLOW,
+    UNDER_OVERFLOW: PANIC_CODES.ARITHMETIC_OVERFLOW,
```
ernestognw, line 18: We might want to update `UNDER_OVERFLOW` in `Panic.sol` as well.

.changeset/pretty-steaks-push.md
ernestognw, line 1: We don't need the changeset since the `Panic` library hasn't been released. We can safely remove it.

contracts/utils/Panic.sol
```diff
@@ -25,8 +25,8 @@ library Panic {
     uint256 internal constant GENERIC = 0x00;
     /// @dev used by the assert() builtin
     uint256 internal constant ASSERT = 0x01;
-    /// @dev arithmetic underflow or overflow
```
ernestognw, line 28: This is correct, it's just that the source defines it like that:

https://github.com/ethereum/solidity/blob/v0.8.24/libsolutil/ErrorCodes.h#L29

I'm fine with only having `overflow`, though. Just mentioning since we may want to send a PR to update the source, he.

ernestognw (approved):
Needs approval from @Amxx

ernestognw (changes requested):
Just realized we need to remove the changeset. Sorry 😅

contracts/utils/Panic.sol
RenanSouza2: Yes, hardhat chose to remove underflow since it is related to floating point not integers, solidity may follow the same path eventually

contracts/utils/Panic.sol
```diff
@@ -47,9 +47,9 @@ library Panic {
     function panic(uint256 code) internal pure {
         /// @solidity memory-safe-assembly
         assembly {
-            mstore(0x00, shl(0xe0, 0x4e487b71))
-            mstore(0x04, code)
-            revert(0x00, 0x24)
+            mstore(0x00, 0x4e487b71)
+            mstore(0x20, code)
+            revert(0x1c, 0x24)
```
ernestognw, line 52: Cool

contracts/utils/Panic.sol
ernestognw: Right, I think we should keep the comment you wrote. I readded it and I'm solving, just wanted to point out this as someone may come in the future with a PR updating it according to the source.

ernestognw (dismissed):
LGTM, thanks @RenanSouza2

Let's wait on @Amxx

Amxx:
> Hey this PR updates the hardhat-chai-matchers library form version 2.0.3 to 2.0,5,

This PR jumps to 2.0.6, not 2.0.5. 
 
> Rationale: version 2.0.5 also includes the ethers.Typed in the address verification, needed to simplify vesting wallet tests

I'm not seeing any simplification of the vesting wallet tests in this PR. Can you showcase how what the improvements are?

RenanSouza2:
> This PR jumps to 2.0.6, not 2.0.5.
> 
> >

yes, they released 2.0.6 in the meanwile, I should have updated the comment here



> I'm not seeing any simplification of the vesting wallet tests in this PR. Can you showcase how what the improvements are?

I didn't want to put the simplifications in this PR to no overblow it but I made the changes here:

https://github.com/RenanSouza2/openzeppelin-contracts/tree/possible-finance-pr/test/metatx

it removes the parts I was least sure about when migrating the finance tests

ernestognw:
If I'm getting it correctly, the update improves support for `Typed` arguments when using `.withArgs`, so we can remove the `argsVerify` argument of the `shouldBehaveLikeVesting` behavior.

The changes don't look too big. I'm fine either merging this PR and then making the test simplification or including it in this same PR. Consider I'll need to re-review if it's the later.

RenanSouza2:
> If I'm getting it correctly, the update improves support for `Typed` arguments when using `.withArgs`, so we can remove the `argsVerify` argument of the `shouldBehaveLikeVesting` behavior.
> 
> The changes don't look too big. I'm fine either merging this PR and then making the test simplification or including it in this same PR. Consider I'll need to re-review if it's the later.

Considering the assembly change in this PR I suggest using 2 PRs

ernestognw:
Right, almost forgot! Then I'd be in favor of merging this. Still requires approval from @Amxx

Amxx (approved):

PR #4901, opened 2024-02-15
add if(_owners[tokenId] != to) condition to save some gas on each NFT minting, burning or transferring

mnedelchev-vn:
**Files changed:**
1. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol

**Description of the change:**
Very small change of adding a check before always updating **[_owners[tokenId] = to;](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L265)** on each minting, burning or transfering of NFT.

### DIFF
--- a/contracts/token/ERC721/ERC721.sol
+++ b/contracts/token/ERC721/ERC721.sol
@@ -262,7 +262,9 @@ abstract contract ERC721 is Context, ERC165, IERC721, IERC721Metadata, IERC721Er
             }
         }
 
-        _owners[tokenId] = to;
+        if (_owners[tokenId] != to) {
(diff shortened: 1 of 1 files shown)


Amxx (commented):
This change adds an sload an a branch for all token updates. That sload alone cost 100gas.

If the value is already correct, it saves the sstore ... but doing an sstore that doesn't modify storage costs 100 gas.

In the end, the saving should be less than the added cost, which is what we observe in the automatic gas report
Capture d’écran du 2024-02-15 15-51-17

PR #4903, opened 2024-02-15
Rewrite comments using superscript for avoid confusion with xor operator

Amxx:
`^` can be confused with the xor operator. For clarity, we sould avoid using `^` to mean exponentiation. 
We should favor using `**` (that is the exponentiation operator in javascript and python), or use supercript numbers.

#### PR Checklist

- [ ] Tests
- [x] Documentation


ernestognw (approved):
LGTM

ernestognw:
There's a check missing but I don't see where is coming from.

PR #4907, opened 2024-02-18
feat: get paginated

Aliceonly:
New features 

paginated getter func in EnumerableSet.sol

startIndex from 1


Amxx:
Hello @Aliceonly

Usually, before implementing a feature like this, we like to open an issue to reference usecases and discuss implementation details. 

The PR in its current format is not following some requirements of our codebase. In particular, the codebase for the map and set structure is generated procedurally, and changes should be implemented through the template.

Openning an issue

PR #4913, opened 2024-02-20
Add CircularBuffer data structure

### LINKED ISSUE #4911: Implement a CircularBuffer
A circular buffer was initialy part of MerkleTree (#3617). This was removed, leaving the developper free to use any history-tracking structure they want.

We should implement a [CircularBuffer](https://en.wikipedia.org/wiki/Circular_buffer), which could be used in conjunction with MerkleTree, of for any other usecase.

Amxx:
This is a push-only structure that keeps track of last N values. If the structure is already full, old values are dropped to make room for the new ones.

Fixes #4911

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/utils/structs/CircularBuffer.sol
```diff
@@ -0,0 +1,114 @@
...
+     * dropped to make room for newer elements.
+     */
+    function last(Bytes32CircularBuffer storage self, uint256 i) internal view returns (bytes32) {
+        uint256 index = self._count;
+        uint256 length = self._data.length;
+        uint256 total = Math.min(index, length); // count(self)
+        if (i >= total) {
+            Panic.panic(Panic.ARRAY_OUT_OF_BOUNDS);
+        }
+        return Arrays.unsafeAccess(self._data, (index - i - 1) % self._data.length).value;
```
Amxx, line 97:
```suggestion
        return Arrays.unsafeAccess(self._data, (index - i - 1) % length).value;
```

contracts/utils/structs/CircularBuffer.sol
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * Struct members have an underscore prefix indicating that they are "private" and should not be read or written to
+     * directly. Use the functions provided below instead. Modifying the struct manually may violate assumptions and
+     * lead to unexpected behavior.
+     *
+     * The last item is at data[(index - 1) % data.length] and the last item is at data[index % data.length]. This
+     * range can wrap around.
+     */
+    struct Bytes32CircularBuffer {
+        uint256 _count;
```
ernestognw, line 38: I get that this value is needed to insert "wrap around" insertions, but I have a couple of thoughts:

- The value is uint256. Making the count uint256 is making the structure an array, it feels off that we don't have a smaller value. However, I'm not sure what we should use instead.
- The count seems to be relevant only when the structure is not full.
- Similar to Merkle Tree's root, one can track this value outside of the contract, and this seems rather an opinionated decision

I'm currently not opposed to keep it, but I'd like to reconsider it.

contracts/utils/structs/CircularBuffer.sol
```diff
@@ -0,0 +1,114 @@
...
+        if (i >= total) {
+            Panic.panic(Panic.ARRAY_OUT_OF_BOUNDS);
+        }
+        return Arrays.unsafeAccess(self._data, (index - i - 1) % length).value;
+    }
+
+    /**
+     * @dev Check if a given value is in the buffer.
+     */
+    function includes(Bytes32CircularBuffer storage self, bytes32 value) internal view returns (bool) {
```
ernestognw, line 103: This function is an unbounded loop. Shall we restrict the maximum length of the buffer?
The alternative may be just a warning

contracts/utils/structs/CircularBuffer.sol
```diff
@@ -0,0 +1,114 @@
...
+        clear(self);
+        Arrays.unsafeSetLength(self._data, length);
+    }
+
+    /**
+     * @dev Clear all data in the buffer, keeping the existing length.
+     */
+    function clear(Bytes32CircularBuffer storage self) internal {
+        self._count = 0;
+    }
```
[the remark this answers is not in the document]
Amxx: While this does not clear the storage, it reset the buffer to that, through the API, it behaves just like if it never contained any data.

Anything store in the array is ignored because count is 0, and whenever the count increases, the new values replace the old ones.

contracts/utils/structs/CircularBuffer.sol
Amxx: The loop is from i = 0 to total, with total being `Math.min(index, length)` ... so it is bounded by the length of the buffer already

contracts/utils/structs/CircularBuffer.sol
Amxx: 2. The count is always relevant. When the structure it point to the location where the next event should be pushed.
3. No, we need it internally to push new values.

Amxx: You think it would be clearer if we named it index, or pos ?

contracts/utils/structs/CircularBuffer.sol
ernestognw: Yes, but the buffer can be huge. This wasn't a problem in Merkle Tree originally because we restricted the maximum length to 255. Perhaps we should do the same here.

Amxx: You asked "Shall we restrict the maximum length of the buffer?". I' saying it already is.

Now I understand you are concerned this might be too large. Maybe. That is already documented at the top of the contract when we say that the complexity is ` * - inclusion ({includes}): O(N) (worst case)`

ernestognw: Fair, I'm just worried it's easy to make a mistake when choosing the size of the buffer even if the NatSpec mentions its complexity. 

Also, it's fair that CircularBuffer users might not need the `includes` function, otherwise, I may recommend using an EnumerableSet instead.

I'll clarify at the `setup` function that a selecting a `size` may render the `includes` function unusable

contracts/utils/structs/CircularBuffer.sol
ernestognw: Nah, I think `_count` is fine. I was trying to make a case for using a smaller variable size (e.g. `uint32`). There's no point in having a `uint256` here since I don't think any CircularBuffer will ever use all the uint256 space. However, there's also no reason to pack the variable.

Let's keep it for now, but now that we have the Packing library, a crazy idea would be to make "parallel" buffers (e.g. 8 channels of uint32's)

Let's keep it as is rn

contracts/utils/structs/CircularBuffer.sol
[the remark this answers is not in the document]
ernestognw: Yes, all good if the memory is only used through the API, but more experienced users may unexpectedly find dirty memory when they thought `clear` actually clears the data as the NatSpec suggests.

I would be better if we are a bit more specific

```suggestion
    /**
     * @dev Clear all data in the buffer without resetting memory, keeping the existing length.
     */
    function clear(Bytes32CircularBuffer storage self) internal {
        self._count = 0;
    }
```

ernestognw (approved):
LGTM

ernestognw (approved):

contracts/utils/structs/CircularBuffer.sol
1468907543: []()****

1468907543 (commented):
_@[]()__ _

PR #4915, opened 2024-02-22
Revert EIP-2098 support drop in ECDSA.sol

k06a:
This refactor allows dApp's to optimize calldata gas usage integrating EIP-2098 even after deployment. Specifically, it reduces the gas cost to 136 units by compressing the signature before it's passed into calldata. This compression strategy takes advantage of the gas pricing mechanism, where 1 non-zero byte costs 16 gas units, and each of the 31 padding zero bytes incurs a cost of 4 gas units.

### DIFF
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -54,10 +54,10 @@ library ECDSA {
      * - with https://docs.ethers.io/v5/api/signer/#Signer-signMessage[ethers]
      */
     function tryRecover(bytes32 hash, bytes memory signature) internal pure returns (address, RecoverError, bytes32) {
+        bytes32 r;
+        bytes32 s;
+        uint8 v;
         if (signature.length == 65) {
-            bytes32 r;
-            bytes32 s;
-            uint8 v;
             // ecrecover takes the signature parameters, and the only way to get them
             // currently is to use assembly.
             /// @solidity memory-safe-assembly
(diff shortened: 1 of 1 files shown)


Amxx:
Hello @k06a 

First, a bit of history:

The logic that you propose we introduce was part of the original PR #2582 that added support for short signatures in the library.

This feature was released in [4.1.0](https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v4.1.0) and then removed in patch [4.7.3](https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v4.7.3).

The reason for that removal is simple: some (IMO poorly written) contracts use the signature itself (or its hash) as an identifier. When the signature is consumed, they mark this identifier as consumed. 
```solidity
mapping(bytes => bool) private _used;

function doSomethingWithSignature(..., bytes calldata signature) public {
    require(!_used[signature], "signature already used");
    _used[signature] = true;

    require(hasRole(SOME_ROLE, ECDSA.recover(keccak256(abi.encode(......)), signature), "unauthorized");
    
    // ... do stuff ...
}
```

Supporting both types of signature means that an attacker is able to take one version of signature that was used, convert it from short to long, or from long to short, and replay the signature. This is terrible contract design, but I get why people think they are protected against signature malleability.

---

Now, what do we do?

I'd love to see short signature more widelly used, and I'd love to reintroduce native bytes support for short signature. But if we do that this bug is going to happen again. We'll need a lot of warning and documentation. We should have do that in 5.0, using the opportunity of the breaking change to change this assumption about signature uniqueness.

In 4.7.3 we went with security over features. I don't think that is ever a bad choice. Have things changed? 

@ernestognw @frangio

frangio:
The problem was that we had _promised_ non-malleability, and implemented non-malleability measures, which completely broke when transparent support for EIP-2098 was introduced.

I agree that I'd like to see this signature format used more widely.

I don't see any issue with adding malleability as long as we do it in a new function so that we don't break the promises of the existing function(s). We would also remove the other non-malleability measures in this new function because they become irrelevant. Ideally, there would be a documentation note about how to avoid the pitfalls of malleability in a contract (i.e., using a nonce).

k06a:
Thanks for the clarification @Amxx @frangio. Sad but true.

PR #4920, opened 2024-02-25
Fix typos in utilities and access control docs

enjoyooor:
Fixes #???? 

Was trying to set up the docs website and noticed that the `build` dist folder is not ignored, so adding to `.gitignore` file. In addition, fixed a bunch of typos in the docs.

#### PR Checklist

- [ ] Tests
- [x] Documentation
- [ ] Changeset entry (run `npx changeset add`) -- don't think this is necessary


ernestognw:
Hey, thanks for submitting the PR! I updated removing the changeset and ignored it since it's not needed imo

> Was trying to set up the docs website and noticed that the build dist folder is not ignored, so adding to .gitignore file. In addition, fixed a bunch of typos in the docs.

I suspect this may be required for a branch but we need to confirm. Perhaps is just a legacy requirement.

ernestognw (approved):
Sorry for the late review. I removed the .gitignore changes because I think it's an unrelated discussion. We can merge this once @Amxx approves.

Thanks!

Amxx (approved):

PR #4924, opened 2024-02-28
Update documentation in IAccessControl

### LINKED ISSUE #4923: The natspec comment of IAccessControl regarding `RoleGranted` event should be updated
**📝 Details**

Since OpenZeppelin Contracts 5.0 is introduced, `_setupRole(bytes32,address)` was deprecated in favor of `_grantRole(bytes32,address)`, so the natspec comment on IAccessControl should be updated accordingly as well:
```
// IAccessControl.sol

/**
* @dev Emitted when `account` is granted `role`.
*
* `sender` is the account that originated the contract call, an admin role
* bearer except when using {AccessControl-_setupRole}.
*/
event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
```

0xdevant:
Fixes #4923

As mentioned, this is to fix the natspec comment of IAccessControl on "RoleGranted" event to confirm with latest logic on AccessControl.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/access/IAccessControl.sol
+++ b/contracts/access/IAccessControl.sol
@@ -30,8 +30,8 @@ interface IAccessControl {
     /**
      * @dev Emitted when `account` is granted `role`.
      *
-     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted
-     * role), expect in cases where the role was granted using the internal {AccessControl-_grantRole}.
+     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted role).
+     * Expected in cases where the role was granted using the internal {AccessControl-_grantRole}.
      */
     event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
 
--- a/contracts/access/IAccessControl.sol
+++ b/contracts/access/IAccessControl.sol
@@ -30,7 +30,7 @@ interface IAccessControl {
     /**
      * @dev Emitted when `account` is granted `role`.
      *
-     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted 
+     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted
      * role), expect in cases where the role was granted using the internal {AccessControl-_grantRole}.
      */
     event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
--- a/contracts/access/IAccessControl.sol
+++ b/contracts/access/IAccessControl.sol
@@ -30,7 +30,8 @@ interface IAccessControl {
     /**
      * @dev Emitted when `account` is granted `role`.
      *
-     * `sender` is the account that originated the contract call i.e. the admin role bearer.
+     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted 
+     * role), expect in cases where the role was granted using the internal {AccessControl-_grantRole}.
      */
     event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender);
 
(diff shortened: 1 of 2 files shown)


ernestognw:
I removed the changeset because this is not something we'd like to document in the changelog.

ernestognw (approved):
Still requires review from @Amxx

contracts/access/IAccessControl.sol
```diff
@@ -30,8 +30,7 @@ interface IAccessControl {
     /**
      * @dev Emitted when `account` is granted `role`.
      *
-     * `sender` is the account that originated the contract call, an admin role
-     * bearer except when using {AccessControl-_setupRole}.
+     * `sender` is the account that originated the contract call i.e. the admin role bearer.
```
Amxx, line 33: When the role is granted using the internal `_grantRole` (for example during construction, or through any other process), then the `msg.sender` may not bear the admin rights for the role that is granted.

0xdevant: Oh right, my bad, so it should be 
```
`sender` is the account that originated the contract call i.e. the admin role bearer
except when using {AccessControl-_grantRole}.
```

Amxx:
```suggestion
     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted 
     * role), expect in cases where the role was granted using the internal {AccessControl-_grantRole}.
```

Amxx (approved):
@ernestognw is that good for you ?

ernestognw:
> @ernestognw is that good for you ?

Yes, more precise. Thanks!

contracts/access/IAccessControl.sol
0xdevant: If that's the case,  should we update the description for the event `RoleRevoked` as well then?

```
/**
 * @dev Emitted when `account` is revoked `role`.
 *
 * `sender` is the account that originated the contract call:
 *   - if using `revokeRole`, it bears the admin role (for the granted role)
 *   - if using `renounceRole`, it is the role bearer (i.e. `account`)
 */
event RoleRevoked(bytes32 indexed role, address indexed account, address indexed sender);
```

Amxx (dismissed):

ernestognw (approved):
Updated with a small nit.

contracts/access/IAccessControl.sol
ernestognw: I think the current description of `RoleRevoked` is fine. Given there are already contracts out there that may grant roles without emitting `RoleGranted`, it justifies clarifying only for that case.

ernestognw: I think the current description of `RoleRevoked` is fine. Given there are already contracts out there that may grant roles without emitting `RoleGranted`, it justifies clarifying only for that case.

contracts/access/IAccessControl.sol
```diff
@@ -30,8 +30,8 @@ interface IAccessControl {
     /**
      * @dev Emitted when `account` is granted `role`.
      *
-     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted
-     * role), expect in cases where the role was granted using the internal {AccessControl-_grantRole}.
+     * `sender` is the account that originated the contract call. This account bears the admin role (for the granted role).
```
Amxx, line 33: that line is 124 chars, should not pass linting.

ernestognw: Interesting, I ran `npm lint:fix` before pushing.

0xdevant: could be another issue to look into...

PR #4927, opened 2024-02-29
Optimize Clone.sol  yul

RenanSouza2:
This PR optmizes the yul in Clones.sol

This optmization can be done because solidity already verify dirty bytes in addresses from calldata

There are other optmizations that can be done but those woud require th use of unallocated memory area

Metrics (with `npm run gas-report`):
```
·--------------------------------------|--------|--------|------|---------·
| Method                               · Before · After  · Diff · % Diff  |
·······································|········|········|······|·········|
| Deployments                          · 271943 · 270215 · 1296 ·  0.635% |
·······································|········|········|······|·········|
| $clone(address)                      ·  64079 ·  64070 ·    9 ·  0.014% |
·······································|········|········|······|·········|
| $cloneDeterministic(address,bytes32) ·  64654 ·  64646 ·    8 .  0.012% |
·--------------------------------------|--------|--------|------|---------·
```

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -4,7 +4,7 @@
 pragma solidity ^0.8.20;
 
 /**
- * @dev https://eips.ethereum.org/EIPS/eip-1167[ERC-1167] is a standard for
+ * @dev https://eips.ethereum.org/EIPS/eip-1167 [ERC-1167] is a standard for
  * deploying minimal proxy contracts, also known as "clones".
  *
  * > To simply and cheaply clone contract functionality in an immutable way, this standard specifies
@@ -30,7 +30,7 @@ library Clones {
         assembly {
             // Cleans the upper 96 bits of the `implementation` word, then packs the first 3 bytes
             // of the `implementation` address with the bytecode before the address.
-            mstore(0x00, or(shr(0xe8, shl(0x60, implementation)), 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000))
+            mstore(0x00, or(shr(0x88, implementation), 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000))
             // Packs the remaining 17 bytes of `implementation` with the bytecode after the address.
             mstore(0x20, or(shl(0x78, implementation), 0x5af43d82803e903d91602b57fd5bf3))
             instance := create(0, 0x09, 0x37)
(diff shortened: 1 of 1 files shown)


ernestognw:
Optimized the title lmao

contracts/proxy/Clones.sol
```diff
@@ -4,7 +4,7 @@
 pragma solidity ^0.8.20;
 
 /**
- * @dev https://eips.ethereum.org/EIPS/eip-1167[ERC-1167] is a standard for
+ * @dev https://eips.ethereum.org/EIPS/eip-1167 [ERC-1167] is a standard for
```
Amxx, line 7: I think the space here is an typo
```suggestion
 * @dev https://eips.ethereum.org/EIPS/eip-1167[ERC-1167] is a standard for
```

RenanSouza2: It is to fix the link

ernestognw: > It is to fix the link

Do you mean in VS Code? This is how it works for the documentation in AsciiDoc, using [URL Macros](https://docs.asciidoctor.org/asciidoc/latest/macros/url-macro/)

ernestognw (commented):
Looks good to me. We need to fix the NatSpec and also get a second approval from @Amxx

contracts/proxy/Clones.sol
RenanSouza2: ohhh, majes sense

Amxx (approved):

Amxx:
For posterity, I tried:

```solidity
/// Alternative 1
mstore(0x00, shl(0x60, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73))
mstore(0x14, shl(0x60, implementation))
mstore(0x28, shl(0x88, 0x5af43d82803e903d91602b57fd5bf3))
instance := create(0, 0x00, 0x37)

/// Alternative 2
mstore(0x00, 0x3d602d80600a3d3981f3363d3d373d3d3d363d73000000000000000000000000)
mstore(0x14, shl(0x60, implementation))
mstore(0x28, 0x5af43d82803e903d91602b57fd5bf30000000000000000000000000000000000)
instance := create(0, 0x00, 0x37)
```

Using 0.8.24 I get:
- Alternative 1 is always 6 gas more expensive than this PR
- Alternative 2 is also 6 gas more expensive if optimisations are enabled, but is 6 gas cheaper when optimisations are turned off.
(Its difficult to isolate the memory preparation from the rest of the function, so devs may want to re-check my numbers with more tooling)

Amxx:
We are talking about 0.01% anyway

ernestognw:
@gitpoap-bot @RenanSouza2

PR #4937, opened 2024-03-01
Improve MerkleProof tests

alexfertel:
#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/test/utils/cryptography/MerkleProof.test.js
+++ b/test/utils/cryptography/MerkleProof.test.js
@@ -55,12 +55,12 @@ describe('MerkleProof', function () {
       const merkleTree = StandardMerkleTree.of(toElements('abc'), ['string']);
 
       const root = merkleTree.root;
-      const leaf = merkleTree.leafHash(['a']);
+      const hash = merkleTree.leafHash(['a']);
       const proof = merkleTree.getProof(['a']);
-      const badProof = proof.slice(0, proof.length - 5);
+      const badProof = proof.slice(0, proof.length - 1);
 
-      expect(await this.mock.$verify(badProof, root, leaf)).to.be.false;
-      expect(await this.mock.$verifyCalldata(badProof, root, leaf)).to.be.false;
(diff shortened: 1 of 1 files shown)


test/utils/cryptography/MerkleProof.test.js
```diff
@@ -55,12 +55,12 @@ describe('MerkleProof', function () {
       const merkleTree = StandardMerkleTree.of(toElements('abc'), ['string']);
 
       const root = merkleTree.root;
-      const leaf = merkleTree.leafHash(['a']);
+      const hash = merkleTree.leafHash(['a']);
```
alexfertel, line 58: The other tests on this file use `hash` instead of `leaf`.

test/utils/cryptography/MerkleProof.test.js
```diff
@@ -55,12 +55,12 @@ describe('MerkleProof', function () {
...
       const proof = merkleTree.getProof(['a']);
-      const badProof = proof.slice(0, proof.length - 5);
+      const badProof = proof.slice(0, proof.length - 1);
```
alexfertel, line 60: This `proof` is two hashes long, so `proof.slice(0, proof.length - 1)` makes `badProof` be empty. This is not really meaningful, but a bit more precise.

Amxx (approved):
LGTM.

We still a second review from @ernestognw. Also, I think we can simplify the PR name (that will show up in the git history)

alexfertel:
> I think we can simplify the PR name

Yeah, I should have split the changes, had a hard time naming this. How about "Make names consistent & slicing simpler in MerkleProof tests"?

ernestognw:
I'm fine with the current name. This is a change in tests that's fine if it's not precise, the PR gets linked so as long as Microsoft keeps Github running, we'll have a pointer to the conversations clarifying the intention behind a change.

ernestognw (approved):

ernestognw:
Thanks @alexfertel

PR #4938, opened 2024-03-01
Fix typo in ERC721Utils.sol

alexfertel:


Amxx (approved):
Thank you for this PR

PR #4941, opened 2024-03-05
Mask computed address in `Create2` and `Clones` libraries

byshape:
The `Create2` library has a `computeAddress` function and the `Clones` library has a `predictDeterministicAddress` function that return address as a result of the `keccak256`. Unmasked, this results in invalid value if used later in assembly code block.

Imagine that the `computeAddress` or `predictDeterministicAddress` function returns the address that is passed to another function that makes a call using that address as a parameter. If this assembly code, such as library, doesn’t clear the address, then this “address” will be invalid and the call will fail.

I suggest masking the result of the `keccak256` function to avoid possible problems with address usage.

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/mocks/ComputedAddressRemainderMock.sol
```diff
@@ -0,0 +1,36 @@
...
+        address deployer
+    ) external pure returns (uint256 remainder) {
+        address predicted = Clones.predictDeterministicAddress(implementation, salt, deployer);
+        remainder = _getRemainder(predicted);
+    }
+
+    function _getRemainder(address addr) internal pure returns (uint256 remainder) {
+        uint256 remainderMask = _REMAINDER_MASK;
+
+        assembly ("memory-safe") {
```
ernestognw, line 32: We generally use the annotation syntax. No particular reason other than the Solidity docs stating that:

> In case you are developing a library that is meant to be compatible across multiple versions of Solidity, you can use a special comment to annotate an assembly block as memory-safe

```suggestion
        /// @solidity memory-safe-assembly
        assembly ("memory-safe") {
```

No big issue since it's a mock.

contracts/mocks/ComputedAddressRemainderMock.sol
ernestognw, line 1: I'd split this mock in 2, one for Clones and the other for Create2, placing each file in their corresponding folder (e.g. /proxy and /utils)

ernestognw: Tried to split them, but realized they make more sense together after all. Solving

ernestognw (dismissed):
Thanks for this contribution @byshape!
Let's wait on @Amxx approval.

Amxx (dismissed):

Amxx (approved):
I don't really see a need to check that in hardhat, as coverage doesn't report that kind of operations

I replaced the mock and hardhat tests with a very short/simple fuzzing test.

ZumZoom:
Not sure about the visibility of the comment inside the resolved outdated changes so duplicating [it](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4941#discussion_r1528458911) here:

Is it still a good idea to stick with `/// @solidity memory-safe-assembly` annotations in favour of `assembly ("memory-safe")`?

ernestognw:
> Is it still a good idea to stick with /// @solidity memory-safe-assembly annotations instead of assembly ("memory-safe")?
That's a good question.

I checked on the Solidity repository and found out this is where the assembly dialect was added.
https://github.com/ethereum/solidity/pull/12620

My understanding is that it was added in a non-breaking way, so we're fine keeping it as a NatSpec annotation but you're right in that we may prefer the dialect. I don't see any concrete benefit but I think we're open to discuss it further. I'm opening an issue.

ernestognw (approved):
I'm merging since the memory-safe annotation can be added in another PR and I don't want to block this one.

PR #4942, opened 2024-03-06
Remove `async` from `describe` blocks and add missing `await` in tests

fvictorio:
When testing the latest version of Hardhat in the OZ repo, we noticed that some of the tests were suddenly failing. The problem was caused by a combination of some missing or extra `async`/`await` keywords. Mainly: the functions of some `describe` blocks had the `async` keyword (and they shouldn't), and some async assertions were missing the `await` keyword. This, combined with the different performance characteristics of the latest HH version, surfaced some issues that weren't showing up before just by coincidence. I fixed that in fe64f5c8.

After that, some other tests started failing. Most of them are related to event assertions that either had the wrong order of arguments, or had a missing argument. These were actually failing before, but since they weren't being awaited, they were floating rejected promises that didn't make the tests fail. I fixed that in 258a203d.

Now there is only one test failing, the `grants role to the user` test in `AccessManager.test.js`. This is asserting that a `RoleAccessRequested` was emitted but, as far as I can tell, that event doesn't exist anymore. I couldn't find an equivalent one so I just left it as it was; I'm sure you can fix it faster than me.

#### PR Checklist

This PR only affects tests, so no docs change or changeset entry is needed.

- [x] Tests
- [ ] Documentation

### DIFF
--- a/hardhat/test-sanity.js
+++ b/hardhat/test-sanity.js
@@ -1,4 +1,0 @@
-// Catches unhandled rejections (mainly from tests)
-process.on('unhandledRejection', reason => {
-  throw new Error(reason);
-});


fvictorio:
I forgot to mention:

- The wrong `async` keywords in describe functions can be prevented with [this one](https://github.com/lo1tuma/eslint-plugin-mocha/blob/main/docs/rules/no-async-describe.md).
- The missing `await` keywords are harder. Ideally you'd use typescript and the `no-floating-promises` rule, but I understand that that would be a huge undertaking.

ernestognw:
Hey @fvictorio, thank you for raising this. After we migrated tests to Ethers I think this was one of our major worries (getting ghost tests).

For the record, this comes from [this PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4710/files#diff-29eec4d6198c2ac37cf692929fdc77636999e93b1580ec73f34d418567e24408L1369-R1256). I guess the oversight came from the previous order of the arguments but I can't explain how `RoleAccessRequested` appeared.

I fixed the tests, and I think we'll need to discuss adding the `no-async-describe` rule (@Amxx ). Unfortunately for the `await` cases, I don't think we're supporting Typescript soon, not only is it a big commitment but is increasing the dependency surface, and the benefit comes only for testing purposes.

ernestognw (approved):
LGTM, we need to discuss the options to avoid it happening again, but the PR is good to go rn.

fvictorio:
Thanks for the review!

fvictorio:
Maybe you can consider adding this somewhere in the tests setup:

```js
process.on("unhandledRejection", (reason, promise) => {
  throw promise;
});
```

Amxx:
Changes to the test files look good. 

Is the hardhat update necessary, it comes with a big `package-lock.json` diff ...

ernestognw:
@fvictorio we see a lot of changes to the package-lock.json, is there any reason for upgrading Hardhat?

We may do this in a future PR but I wonder if we need it now

fvictorio:
No, no need to update it now. You might get a 10-15% performance improvement, but there isn't a strong reason for you to upgrade. I can rebase that commit out if you want.

ernestognw:
> No, no need to update it now. You might get a 10-15% performance improvement, but there isn't a strong reason for you to upgrade. I can rebase that commit out if you want.

Yeah I think we'd prefer to keep this PR short. We still want to upgrade for next hardfork though

fvictorio:
Done!

hardhat/env-artifacts.js
```diff
@@ -1,5 +1,9 @@
 const { HardhatError } = require('hardhat/internal/core/errors');
 
+process.on('unhandledRejection', reason => {
+  throw new Error(reason);
```
ernestognw, line 4: I couldn't find anything useful in the second argument of the listener (the promise). It doesn't give the stacktrace anyway and this one works fine.

The only drawback is that it halts the tests whenever a failure is found. We can tweak it to behave differently but I'm personally fine with it.

ernestognw (approved):
Thanks for all your work @fvictorio 🙏🏻 
I created an issue for the ESLint rule so we can discuss it later: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4943

Rn it's looking good again, let's wait on @Amxx approval.

hardhat/env-artifacts.js
Amxx: I wouldn't have put it in this file. IMO this should be its own file in `hardhat/`.

Maybe `hardhat/test-sanity.js`

test/proxy/beacon/BeaconProxy.test.js
```diff
@@ -34,7 +34,7 @@ describe('BeaconProxy', function () {
     it('non-compliant beacon', async function () {
       const badBeacon = await ethers.deployContract('BadBeaconNoImpl');
 
-      await expect(this.newBeaconProxy(badBeacon, '0x')).to.be.revertedWithoutReason;
+      await expect(this.newBeaconProxy(badBeacon, '0x')).to.be.reverted;
```
Amxx, line 37: I guess the issue here is a `ERC1967InvalidImplementation` with address(1). We should test that.

test/token/ERC1155/extensions/ERC1155Supply.test.js
```diff
@@ -111,7 +111,7 @@ describe('ERC1155Supply', function () {
 
   describe('other', function () {
     it('supply unaffected by no-op', async function () {
-      this.token.safeTransferFrom(ethers.ZeroAddress, ethers.ZeroAddress, firstTokenId, firstTokenValue, '0x');
+      this.token.$_update(ethers.ZeroAddress, ethers.ZeroAddress, [firstTokenId], [firstTokenValue]);
```
Amxx, line 114: This is missing an await, no?

Amxx:
```suggestion
      await this.token.$_update(ethers.ZeroAddress, ethers.ZeroAddress, [firstTokenId], [firstTokenValue]);
```

test/proxy/beacon/BeaconProxy.test.js
Amxx:
```suggestion
      await expect(this.newBeaconProxy(badBeacon, '0x'))
        .to.be.revertedWithCustomError(this.factory, 'ERC1967InvalidImplementation');
```

Amxx, line 37: Here the revert is:
- BadBeaconNoImpl does not implement `implementation()`
- ERC1967 utils try to get that in `_setBeacon`
- the call fail (no impelmentation of the selector + no fallback)

Amxx:
Moved 
```javascript
process.on('unhandledRejection', reason => {
  throw new Error(reason);
});
```
to a dedicated file + added a missing await (are there any more?)

test/proxy/beacon/BeaconProxy.test.js
ernestognw: It was previously:

```
await expect(this.newBeaconProxy(badBeacon, '0x')).to.be.revertedWithoutReason
```

Which seemed correct to me but I'm realizing it was missing the parenthesis:

```
await expect(this.newBeaconProxy(badBeacon, '0x')).to.be.revertedWithoutReason();
```

We can keep it with the explicit `revertedWithoutReason()` and keep the comments.

hardhat/env-artifacts.js
ernestognw: I'm fine with `test-sanity.js`

ernestognw:
> to a dedicated file + added a missing await (are there any more?)

Right, the one you spotted is a tricky one given is expecting no changes. The test should make sure the function is called before asserting, though. I don't expect other tests to be in similar cases.

ernestognw (approved):

Amxx (approved):

PR #4949, opened 2024-03-10
Remove root from MerkleTree

### LINKED ISSUE #4946: Consider removing the `_root` for the MerkleTree structure
In the MerkleTree structure introduced in #3617
- the "updated" root value is already returned by `push`
- dev should store that value "somewhere" (can be a ring buffer, a checkpoint, ...)

So storing this value in not really need. Its probably going to be a duplicate ... that cost 5k gas

Proposition is to:
- remove the `_root` entry in `MerkleTree`
- have `setup` return the zero-root.

ernestognw:
Fixes #4946

After discussing, we considered it rather unnecessary to store the root of the tree along with it, and its core functionality is enough for anyone to build use cases where the root is tracked. 

With this change the user may decide to keep track of the root for purposes that may not require on-chain availability (e.g. a whitelist that stores roots every N insertions) or that may require a lookup mechanism (e.g. Bitmaps) for which storing the root is already an opinionated decision.

We can consider adding a default recommendation, such as #4911.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/mocks/MerkleTreeMock.sol
+++ b/contracts/mocks/MerkleTreeMock.sol
@@ -9,19 +9,20 @@ contract MerkleTreeMock {
 
     MerkleTree.Bytes32PushTree private _tree;
 
-    event LeafInserted(bytes32 leaf, uint256 index, bytes32 root);
-    event TreeSetup(uint8 depth, bytes32 zero, bytes32 root);
+    // This mock only stored the latest root.
+    // Production contract may want to store historical values.
+    bytes32 public root;
 
-    function setup(uint8 _depth, bytes32 _zero) public returns (bytes32) {
-        bytes32 initialRoot = _tree.setup(_depth, _zero);
-        emit TreeSetup(_depth, _zero, initialRoot);
-        return initialRoot;
+    event LeafInserted(bytes32 leaf, uint256 index, bytes32 root);
+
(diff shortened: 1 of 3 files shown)


contracts/mocks/MerkleTreeMock.sol
```diff
@@ -10,18 +10,18 @@ contract MerkleTreeMock {
...
-    function push(bytes32 leaf) public {
-        (uint256 leafIndex, bytes32 currentRoot) = _tree.push(leaf);
+    function push(bytes32 leaf) public returns (uint256 leafIndex, bytes32 currentRoot) {
+        (leafIndex, currentRoot) = _tree.push(leaf);
         emit LeafInserted(leaf, leafIndex, currentRoot);
-    }
-
-    function root() public view returns (bytes32) {
-        return _tree.root();
+        return (leafIndex, currentRoot);
```
Amxx, line 24: Is the return actually needed here ? What was wrong with the previous version ?

Amxx:
That works, but thecking the event is a bit counterintuitive IMO, and it doesn't "show" the need for devs to store the root "outside the library"

I would have added a `bytes32 public latestRoot` in the Mock, and set it in `setup` and `push`. We would then check that using the getters and not trhough the events.

Amxx (approved):

PR #4951, opened 2024-03-12
Make SignatureChecker EIP-7377 safe

### LINKED ISSUE #4855: Consider checking code length of the signer before attempting an ecrecover in SignatureChecker.isValidSignatureNow
The current implementation of `SignatureChecker.isValidSignatureNow` is

```solidity
function isValidSignatureNow(address signer, bytes32 hash, bytes memory signature) internal view returns (bool) {
    (address recovered, ECDSA.RecoverError error, ) = ECDSA.tryRecover(hash, signature);
    return
        (error == ECDSA.RecoverError.NoError && recovered == signer) ||
        isValidERC1271SignatureNow(signer, hash, signature);
}
```

This implementation assumes that ecrecover will not identify the correct signer if the signer is a smart contract. This comes from the assumption that there contract addresses is generated in such a way that there is no (known) private key that derivates to the same address.

These assumption could be challenged by [EIP-7377](https://eips.ethereum.org/EIPS/eip-7377).

If EIP-7377 is deployed, a private key would be able to deploy code at its own address. In that case, there would be a known private key for the contract. This causes a governance issue. Some people may expect the contract to have some shared ownership (Multisig), be controlled by another wallet (Ownable), or just trustless. But using the private key that was used to deploy it, the deployer could generate a Permit signature (or similar) that could result in funds being drained out of the contract.

This attack will be feasable on historical implementations if the ecrecover precompile does not check the presence of code at the recovered location!

It as been proposed to not rely on the precompile possibly being changed, and change our implementation toward:

```solidity
function isValidSignatureNow(address signer, bytes32 hash, bytes memory signature) internal view returns (bool) {
    if (signer.code.length == 0) {
        (address recovered, ECDSA.RecoverError error, ) = ECDSA.tryRecover(hash, signature);
        return (error == ECDSA.RecoverError.NoError && recovered == signer);
    } else {
        return isValidERC1271SignatureNow(signer, hash, signature);
    }
}
```

However, doing this would significantly increase the gas cost when the signer is an EOA:
- If the signer is an EOA, then it is likelly that the signature is used in a transaction that is not performed by the signer itself. In that case, the signer account is cold, and the `signer.code.length == 0` checks costs 2600 gas! With signer being an EOA, it is unlikelly that it is called at any point, so "warming up" the account it not reused, and we should not anticipate saving elsewere as a result of this.
- If the signer is not an EOA, then we are going to call `IERC1271.isValidSignature` on it anyway. In that case the cost of the check is only really 100gas. The new version would also not try the unecessary ecrecover, with result is some savings. Overall this new version would be cheaper.

We expect the proposed version to be cheaper when signer is a contract, and signature verification is done through ERC-1271. However the proposed version should also be significantly more expensive (+2600gas) when the signer is an EOA, which is the vast majority of cases.

Amxx:
Fixes #4855

#### PR Checklist

### DIFF
--- a/.changeset/yellow-moles-hammer.md
+++ b/.changeset/yellow-moles-hammer.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-`SignatureChecker`: refactor `isValidSignatureNow` to make it safe if EIP-7377 (or similar) is ever implemented.
+`SignatureChecker`: refactor `isValidSignatureNow` to avoid validating ECDSA signatures if there is code deployed at the signer's address.
--- a/.changeset/yellow-moles-hammer.md
+++ b/.changeset/yellow-moles-hammer.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`SignatureChecker`: refactor `isValidSignatureNow` to make it safe if EIP-7377 (or similar) is ever implemented.
--- a/contracts/utils/cryptography/SignatureChecker.sol
+++ b/contracts/utils/cryptography/SignatureChecker.sol
@@ -20,10 +20,12 @@ library SignatureChecker {
      * change through time. It could return true at block N and false at block N+1 (or the opposite).
      */
     function isValidSignatureNow(address signer, bytes32 hash, bytes memory signature) internal view returns (bool) {
-        (address recovered, ECDSA.RecoverError error, ) = ECDSA.tryRecover(hash, signature);
-        return
-            (error == ECDSA.RecoverError.NoError && recovered == signer) ||
-            isValidERC1271SignatureNow(signer, hash, signature);
+        if (signer.code.length == 0) {
+            (address recovered, ECDSA.RecoverError err, ) = ECDSA.tryRecover(hash, signature);
+            return err == ECDSA.RecoverError.NoError && recovered == signer;
+        } else {
+            return isValidERC1271SignatureNow(signer, hash, signature);
+        }
     }
 
     /**


Amxx:
Capture d’écran du 2024-03-12 17-28-10
Relevant gas measurement:
- Best case becomes more expensive
- Worst case becomes cheaper
- On average (over our tests) it is slightly cheaper

Purpose of the change is mostly security, not gas costs

.changeset/yellow-moles-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`SignatureChecker`: refactor `isValidSignatureNow` to make it safe if EIP-7377 (or similar) is ever implemented.
```
ernestognw, line 5: I think the changelog should reflect the change rather than the rationale. This change looks more like:

```suggestion
`SignatureChecker`: refactor `isValidSignatureNow` to avoid validating smart contract signatures if the signer has code.
```

ernestognw (commented):
Looks good. Let's just discuss the changeset.

.changeset/yellow-moles-hammer.md
Amxx: I'd say "avoid validating ECDSA signatures" instead of "avoid validating smart contract signatures".

I think "smart contract signatures" can be confusing. Some people think of 1271 as smart contract signature.

ernestognw: > I'd say "avoid validating ECDSA signatures" instead of "avoid validating smart contract signatures".

Yeah, I agree with that concern. Let's go with your first proposal.

ernestognw (approved):

ernestognw (approved):

PR #4952, opened 2024-03-12
Implement RSA verification

Amxx:
Fix LIB-1238

Work in progress. Based on https://github.com/adria0/SolRsaVerify

Tests:
- parse `SigVer15_186-3.rsp` from [here](https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/dss/186-2rsatestvectors.zip) and automatically run all test examples.
- also run two extra tests from https://github.com/adria0/SolRsaVerify

For now, only sha256 is supported. `@nobles/hash` provides all the function we need to generate the digest, and `SigVer15_186-3.rsp` contains more tests.

Do we want to support more?

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,118 @@
...
+    function pkcs1Sha256(
+        bytes32 digest,
+        bytes memory sig,
+        bytes memory exp,
+        bytes memory mod
+    ) internal view returns (bool) {
+        unchecked {
+            // cache and check length
+            uint256 length = mod.length;
+            if (length < 0x40 || length != sig.length) {
```
ernestognw, line 46: I guess we reject length less than `0x40` because it wouldn't be secure. I wonder if `0x40` was arbitrarily chosen. If so, we need to evaluate it carefully, as far as I remember, RSA's security is `p * q` so a 512 bits signature is crackable in reasonable time.

Found this as a reference, but seems like 512 bits (0x40 bytes) signatures are pretty much broken.
https://github.com/tomrittervg/cloud-and-control/blob/master/gnfs-info/factoring-howto.txt

[RFC 3447](https://datatracker.ietf.org/doc/html/rfc8017) is from 2003 and was superseded by [RFC 8017](https://datatracker.ietf.org/doc/html/rfc8017), though, I couldn't find a recommendation for the mod length. Allegedly, 512 bits security was [first broken in 1999](https://crypto.stackexchange.com/a/3933), so my estimations say that we might increase this to `0x80` at least.

Needs discussion

ernestognw, line 46: `length != sig.length` corresponds to the first step of [RFC 3447 in section 8.2.2](https://datatracker.ietf.org/doc/html/rfc8017#section-8.2.2)

> 1. Length checking: If the length of the signature S is not k octets, output "invalid signature" and stop.

ernestognw (commented):
Managed to navigate through [RFC 8017](https://datatracker.ietf.org/doc/html/rfc8017) and mostly looks good.

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,118 @@
...
+                mask = 0xffffffffffffffffffffffffffffffffffffffff000000000000000000000000;
+            } else if (_unsafeReadBytes1(buffer, length - 48) == 0x2F) {
+                // case: sha256Implicit
+                offset = 0x32;
+                params = 0x00302f300b060960864801650304020104200000000000000000000000000000;
+                mask = 0xffffffffffffffffffffffffffffffffffff0000000000000000000000000000;
+            } else {
+                // unknown
+                return false;
+            }
```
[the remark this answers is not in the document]
ernestognw: Added comments explaining what each byte is

ernestognw:
I took the freedom of renaming the functions since one that receives a `bytes32` digest is not exclusively `sha256` as far as I can tell. Theoretically, a user can sign with RSA using any digest they like.

> For now, only sha256 is supported. @nobles/hash provides all the function we need to generate the digest, and SigVer15_186-3.rsp contains more tests.
>
> Do we want to support more?

Honestly, I don't think so, we'd require to adjust the param checks according to the length of the digest since we're counting backward to get the digest and allow optional digest info parameters.

I'm marking this as ready for review since I think it's already good enough and I already feel comfortable with its security.

ernestognw (commented):
In my opinion, I think the last thing to resolve is whether to restrict `n` to have a minimum length or not. I added a note (probably should be a warning) suggesting that no key with length less than 2048 bits should be used.

If we don't find a reason to continue allowing smaller lengths, then I'd advocate for restricting the function.

Aside from that, we need to cover the missing test cases (should be trivial) and it'll be good to go

.changeset/curvy-crabs-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`RSA`: Library to verify signatures according to RFC8017 Signature Verification Operation
```
Amxx, line 5:
```suggestion
`RSA`: Library to verify signatures according to RFC 8017 Signature Verification Operation
```

ernestognw:
[Line 106](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4952/files#diff-d5604af0f028f31eba96d4f25df528d80de6aa9b98f04b9cfd2607ab3f0b9391R106) is pretty difficult to reach as it requires to set up an attack where we forge two messages such that once signed we can derive a third signature that decrypts wrongly to the message. (Note by "decrypt" I'm referring to the `modExp` operation to recover the message)

I think it's ok if we don't have a test case for that rn. In the future we should look for adding fuzzing, however, we don't have an easy way of fuzzing keys unless we use `--ffi`.

ernestognw (approved):
Looks solid and I feel comfortable with the result, I'll leave it to your consideration @Amxx.

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,136 @@
...
+            uint256 length = n.length;
+            if (
+                length < 0x40 || // PKCS#1 padding is slightly less than 0x40 bytes at the bare minimum
+                length != s.length // signature must have the same length as the finite field
+            ) {
+                return false;
+            }
+
+            // RSAVP1 https://datatracker.ietf.org/doc/html/rfc8017#section-5.2.2
+            (bool success, bytes memory buffer) = Math.tryModExp(s, e, n);
```
cairoeth, line 52: do we have to check that buffer (`Math.tryModExp(s, e, n)`) = s to verify that the signature is valid for the public key (n, e)?

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,136 @@
...
+
+            // The padding has variable (arbitrary) length, so we check it byte per byte in a loop.
+            // This is required to ensure non-malleability. Not checking would allow an attacker to
+            // use the padding to manipulate the message in order to create a valid signature out of
+            // multiple valid signatures.
+            for (uint256 i = 2; i < paddingEnd; ++i) {
+                if (_unsafeReadBytes1(buffer, i) != 0xFF) {
+                    return false;
+                }
+            }
```
cairoeth, line 108: what do you think of how [this implementation](https://github.com/Legrandin/pycryptodome/blob/master/lib/Crypto/Signature/pkcs1_15.py#L133-L139) checks for this?


> By comparing the full encodings (as opposed to checking each
> of its components one at a time) we avoid attacks to the padding
> scheme like [Bleichenbacher's](https://medium.com/@c0D3M/bleichenbacher-attack-explained-bc630f88ff25)

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    function _unsafeReadBytes1(bytes memory array, uint256 offset) private pure returns (bytes1) {
+        return bytes1(_unsafeReadBytes32(array, offset));
+    }
+
+    function _unsafeReadBytes2(bytes memory array, uint256 offset) private pure returns (bytes2) {
+        return bytes2(_unsafeReadBytes32(array, offset));
+    }
+}
```
cairoeth, line 136: because `_unsafeReadBytes2` is only used once, would it be cleaner if we just do the bytes2 casting [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4952/files#diff-d5604af0f028f31eba96d4f25df528d80de6aa9b98f04b9cfd2607ab3f0b9391R112)?

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,136 @@
...
+                length != s.length // signature must have the same length as the finite field
+            ) {
+                return false;
+            }
+
+            // RSAVP1 https://datatracker.ietf.org/doc/html/rfc8017#section-5.2.2
+            (bool success, bytes memory buffer) = Math.tryModExp(s, e, n);
+            if (!success) {
+                return false;
+            }
```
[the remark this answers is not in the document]
Amxx: tryModExp doesn't do that check, and we have nothing that guarantee that s < n.

We should probably implement a bytes comparaison function, and use it here.

contracts/utils/cryptography/RSA.sol
Amxx: The result of `s ** e % n` is buffer. We don't check that it is equal to S. I don't think it should be equal to S.

We check that is it a properly formated "predicate" that contains the "digest", and all the other formating stuff.

contracts/utils/cryptography/RSA.sol
Amxx: I'm not sure what are the length of the objects being compared. This may not apply to solidity.

Amxx: If we were able to cap the size of the padding, we could keep track of all the possible hashes of string solely composed of `FF`, and compare the hash of that section ...

But I'm not really sure we can have a cap on paddingEnd (that requires a cap on length)

contracts/utils/cryptography/RSA.sol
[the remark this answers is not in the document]
Amxx: we could use something like 

```
enum CmpResult { Eq, Lt, Gt }

function compare(bytes memory a, bytes memory b) pure returns (CmpResult) {
    unchecked {
        uint256 i = 0; // pointer on the a buffer
        uint256 j = 0; // pointer on the b buffer
        if (a.length > b.length) {
            while (i < a.length - b.length) {
                if (a[i++] != 0) return CmpResult.Gt;
            }
        } else if (a.length < b.length) {
            while (j < b.length - a.length) {
                if (b[j++] != 0) return CmpResult.Lt;
            }
        }
        while (i < a.length) { // eq j < b.length
            bytes1 ai = a[i++];
            bytes1 bj = b[j++];
            if (ai < bj) {
                return CmpResult.Lt;
            } else if (ai > bj) {
                return CmpResult.Gt;
            }
        }
        return CmpResult.Eq;
    }
}
```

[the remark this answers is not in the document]
Amxx: @frangio 
I think we need this to avoid signature maleability ... but it also feels like an expensive check. WDYT?

We could write it in assembly ...
I'm going to check for other existing implementation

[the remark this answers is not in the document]
Amxx: or we could check that 
```solidity
keccak256(s) == keccak256(modExp(s, 1, n))
```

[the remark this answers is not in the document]
ernestognw: I'd be fine with `keccak256(s) == keccak256(modExp(s, 1, n))`.

`modExp` is only 200 gas if `e == 1`, and the cost of both keccak256 is 30.
The bottom line is that we're adding ~300 gas units, which is about 12 cents at 100gwei and ETH at 4000. 
I think this is good enough, and the loop might be even more costly if we end up accepting 4096-bit keys.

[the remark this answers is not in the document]
Amxx: There is also the cost of the two keccak. I suspect the loop might be cheaper (particularly is we do some fancy assembly that checks in blocks of 32 bytes)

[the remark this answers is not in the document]
ernestognw: Right, fair points. I also didn't consider under normal circumstances, the `compare` function finishes as early as possible.

If we're adding the `compare` function, I'd put it in `Math.sol`, wdyt?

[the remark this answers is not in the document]
ernestognw: Note that we have already validated that `s.length == n.length` above, do you think the following would work?

```
enum CmpResult { Eq, Lt, Gt }
function compare(bytes memory a, bytes memory b) pure returns (CmpResult) {
    uint256 higherA = 0;
    uint256 higherB = 0;
    assembly ("memory-safe") {
        higherA := mload(add(a, 0x20))
        higherB := mload(add(b, 0x20))
    }
    return
        higherA == higherB
            ? CmpResult.Eq
            : higherA < higherB
                ? CmpResult.Lt
                : CmpResult.Gt;
}
```

or at least:

```
uint256 i = 0; // pointer on the a buffer
uint256 j = 0; // pointer on the b buffer
if (a.length > b.length) {
    while (i < a.length - b.length) {
        if (a[i++] != 0) return CmpResult.Gt;
    }
} else if (a.length < b.length) {
    while (j < b.length - a.length) {
        if (b[j++] != 0) return CmpResult.Lt;
    }
}
while (i < a.length) { // eq j < b.length
    bytes1 ai = a[i++];
    bytes1 bj = b[j++];
    if (ai < bj) {
        return CmpResult.Lt;
    } else if (ai > bj) {
        return CmpResult.Gt;
    }
}
return CmpResult.Eq;
```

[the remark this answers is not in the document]
Amxx: first option only checks the first 32 bytes ... if the difference is further its not catching it. Arbitrary length means we need some loop somewhere.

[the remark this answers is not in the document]
ernestognw: worst case in the first option is that they're 32-bytes equal, would it be possible to just require an explicit `Gt` and revert otherwise?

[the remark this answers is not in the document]
Amxx: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4952/commits/abe598d94f12e51648b9ea0e741083043e834e92

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,140 @@
...
+
+            // Check that buffer is well encoded:
+            // buffer ::= 0x00 | 0x01 | PS | 0x00 | DigestInfo
+            //
+            // With
+            // - PS is padding filled with 0xFF
+            // - DigestInfo ::= SEQUENCE {
+            //    digestAlgorithm AlgorithmIdentifier,
+            //      [optional algorithm parameters]
```
ernestognw, line 76: Given the verification algorithm considers optional algorithm parameters, then it's dangerous to protect for replay attacks by using a hash even if we make sure `s` is less than `n`. We should warn about that in the docs.

contracts/utils/cryptography/RSA.sol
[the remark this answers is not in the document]
frangio: Do we know what would be the average cost of this check?

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,140 @@
...
+                length != s.length // signature must have the same length as the finite field
+            ) {
+                return false;
+            }
+
+            // verify that s < n
+            bool ok = false;
+            for (uint256 i = 0; i < length; i += 0x20) {
+                bytes32 si = _unsafeReadBytes32(s, Math.min(i, length - 0x20));
+                bytes32 ni = _unsafeReadBytes32(n, Math.min(i, length - 0x20));
```
frangio, line 55:
```suggestion
                uint256 p = Math.min(i, length - 0x20);
                bytes32 si = _unsafeReadBytes32(s, p);
                bytes32 ni = _unsafeReadBytes32(n, p);
```

This code reverts if length is < 0x20... but I don't think that's realistic is it?

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,140 @@
...
+                if (si < ni) {
+                    ok = true;
+                    break;
+                }
```
ernestognw, line 59: Shall we mention this?

```suggestion
                // if (si > ni) return false; // Assuming an honest input, this check makes it more expensive for the caller
                if (si < ni) {
                    ok = true;
                    break;
                }
```

contracts/utils/cryptography/RSA.sol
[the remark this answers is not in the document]
Amxx: The for could be long (64 iterations for a 4096 key)  ... but there is a break !

All real cases should have a difference somewhere in the first 32bytes (having such a large equality is not realisitc IMO).

One iteration should be way under 100 gas.

contracts/utils/cryptography/RSA.sol
ernestognw: We know `length >= 0x40` from a check above

contracts/utils/cryptography/RSA.sol
ernestognw: We know `length >= 0x40` from a check above

contracts/utils/cryptography/RSA.sol
Amxx: I don't understand the comment.

ernestognw: If `si < ni`, then it's not needed to check the next memory word (32 bytes).  However, this only optimizes for an attacker, and wouldn't be needed for an honest caller.

I just found it weird that it loops until the end.

Amxx: actually we had a bug here

contracts/utils/cryptography/RSA.sol
frangio: Ok nice! In that case this can be unchecked.

contracts/utils/cryptography/RSA.sol
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            // Verify that s < n to ensure there's only one valid signature for a given message
+            for (uint256 i = 0; i < length; i += 0x20) {
+                uint256 p = Math.min(i, length - 0x20);
+                bytes32 sp = _unsafeReadBytes32(s, p);
+                bytes32 np = _unsafeReadBytes32(n, p);
+                if (sp < np) {
+                    // s < n in the upper bits (everything before is equal) → s < n globally: ok
+                    break;
```
cairoeth, line 61: if the loop breaks as soon as it finds one segment where sp < np (say the first 32-byte segment), wouldn't it overlook segments where sp > np later in the array (say the second 32-byte segment), potentially leading to incorrect conclusions about the overall order of s and n?

cairoeth: maybe im not understand the assumption of `s < n globally`

test/utils/cryptography/RSA.test.js
```diff
@@ -0,0 +1,107 @@
...
+      const sig =
+        '0xa0073057133ff3758e7e111b4d7441f1d8cbe4b2dd5ee4316a14264290dee5ed7f175716639bd9bb43a14e4f9fcb9e84dedd35e2205caac04828b2c053f68176d971ea88534dd2eeec903043c3469fc69c206b2a8694fd262488441ed8852280c3d4994e9d42bd1d575c7024095f1a20665925c2175e089c0d731471f6cc145404edf5559fd2276e45e448086f71c78d0cc6628fad394a34e51e8c10bc39bfe09ed2f5f742cc68bee899d0a41e4c75b7b80afd1c321d89ccd9fe8197c44624d91cc935dfa48de3c201099b5b417be748aef29248527e8bbb173cab76b48478d4177b338fe1f1244e64d7d23f07add560d5ad50b68d6649a49d7bc3db686daaa7';
+      const exp = '0x03';
+      const mod =
+        '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000';
+      expect(await this.mock.$pkcs1Sha256(data, sig, exp, mod)).to.be.false;
```
cairoeth, line 104: this test is actually not reverting at the modExp step, but instead [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4952/files#diff-d5604af0f028f31eba96d4f25df528d80de6aa9b98f04b9cfd2607ab3f0b9391R62) because of `sp > np`. is there a way we can have more accurate checks when the return data is a boolean? maybe by measuring gas?

test/utils/cryptography/RSA.test.js
```diff
@@ -0,0 +1,107 @@
...
+      const data = ethers.toUtf8Bytes('hello world!');
+      const sig = '0x00112233';
+      const exp = '0x03';
+      const mod =
+        '0xe932ac92252f585b3a80a4dd76a897c8b7652952fe788f6ec8dd640587a1ee5647670a8ad4c2be0f9fa6e49c605adf77b5174230af7bd50e5d6d6d6d28ccf0a886a514cc72e51d209cc772a52ef419f6a953f3135929588ebe9b351fca61ced78f346fe00dbb6306e5c2a4c6dfc3779af85ab417371cf34d8387b9b30ae46d7a5ff5a655b8d8455f1b94ae736989d60a6f2fd5cadbffbd504c5a756a2e6bb5cecc13bca7503f6df8b52ace5c410997e98809db4dc30d943de4e812a47553dce54844a78e36401d13f77dc650619fed88d8b3926e3d8e319c80c744779ac5d6abe252896950917476ece5e8fc27d5f053d6018d91b502c4787558a002b9283da7';
+      expect(await this.mock.$pkcs1Sha256(data, sig, exp, mod)).to.be.false;
+    });
+
+    it('returns false if the modexp operation fails', async function () {
+      const data = ethers.toUtf8Bytes('hello world!');
```
cairoeth, line 98: if module is 0, then it will revert at the s < n check, so maybe we can force the precompile to always revert for this case:
```suggestion
      setCode("0x0500000000000000000000000000000000000000", "0x60006000fd");
      const data = ethers.toUtf8Bytes('hello world!');
```

contracts/utils/cryptography/RSA.sol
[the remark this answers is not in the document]
frangio: >64 iterations for a 4096 key

Are you sure about this? Isn't it 4096 bits, so 512 bytes, which is 16 32-byte words.

ernestognw (dismissed):
Added some docs as well. LGTM.

I'd wait for a second approval.

cairoeth (approved):
lgtm -- @ernestognw: I added a replay protection notice on the docs page

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -26,6 +30,32 @@ function _verify(bytes32 data, bytes memory signature, address account) internal
...
+    bytes32 data,
+    bytes memory signature,
+    bytes memory e,
+    bytes memory n
+) internal pure returns (bool) {
+    return data.pkcs1(signature, e, n);
+}
+----
+
+IMPORTANT: Always use keys of at least 2048 bits. Additionally, be aware that PKCS#1 v1.5 allows for replayability due to the possibility of arbitrary optional parameters. To prevent replay attacks, consider including an onchain nonce or unique identifier in the message.
```
ernestognw, line 57:
```suggestion
IMPORTANT: Always use keys of at least 2048 bits. Additionally, be aware that PKCS#1 v1.5 allows for replayability due to the possibility of arbitrary optional parameters added to the message construction before signing it. To prevent replay attacks, consider including an onchain nonce or unique identifier in the message.
```
```suggestion
IMPORTANT: Always use keys of at least 2048 bits. Additionally, be aware that PKCS#1 v1.5 allows for replayability due to the possibility of arbitrary optional parameters. To prevent replay attacks, consider including an onchain nonce or unique identifier in the message.
```
```suggestion
IMPORTANT: Always use keys of at least 2048 bits. Additionally, be aware that PKCS#1 v1.5 allows for replayability due to the possibility of arbitrary optional parameters. To prevent replay attacks, consider including an onchain nonce or unique identifier in the message.
```

ernestognw:
> lgtm -- @ernestognw: I added a replay protection notice on the docs page

Thanks, looking good. Approving again

ernestognw (approved):

PR #4955, opened 2024-03-13
Add a Slots library, and a transient variant of ReentrancyGuard

### LINKED ISSUE #4205: `ReentrancyGuard` with transient opcodes `TSTORE` and `TLOAD`
**🧐 Motivation**

Since [EIP-1153](https://eips.ethereum.org/EIPS/eip-1153) is confirmed to be included in the upcoming Cancun-Deneb upgrade, I would like to discuss the possibility of using the future transient opcodes `TSTORE` and `TLOAD` for the `ReentrancyGuard` (abstract) contract.

Furthermore, I would like to highlight the initiatives (partially driven by myself), to disable reentrancy by default at the compiler level:
- **Vyper:** https://github.com/vyperlang/vyper/issues/3380
- **Solidity:** https://github.com/ethereum/solidity/issues/12996

### LINKED ISSUE #4888: Use transient storage in ReentrancyGuard
For backward compatibility, and support of chains that do no implement EIP-1153, we whould have two version (the current one + a new transient one). Current one would be marked as deprecated

Amxx:
Fixes #4205
Fixes #4888

Note:
- **Hardhat update to 2.21.0:** Technically only 2.20.0 is needed to support cancun, but 2.21.0 includes EDR that improve test speed
- **hardhat.config.js:** Add a new entry `evm` / `evmVersion` that affects both the compiler (targeted evm version) and the node (executed evm version)
- **foundry.toml:** enable cancun (and set compiler to 0.8.24)

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/utils/TranscientReentrancyGuard.sol
```diff
@@ -0,0 +1,56 @@
...
+import {TranscientStorage} from "./TranscientStorage.sol";
+
+/**
+ * @dev Variant of {ReentrancyGuard} that uses transcient storage.
+ *
+ * NOTE: This variant only works on networks where EIP-1153 is available.
+ */
+abstract contract TranscientReentrancyGuard {
+    // keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.ReentrancyGuard")) - 1)) & ~bytes32(uint256(0xff))
+    bytes32 private constant ReentrancyGuardStorageLocation =
```
Amxx, line 14:
```suggestion
    bytes32 private constant REENTRANCY_GUARD_STORAGE =
```

scripts/generate/templates/TranscientStorage.js
```diff
@@ -0,0 +1,47 @@
+const format = require('../format-lines');
+const { capitalize } = require('../../helpers');
+
+const TYPES = ['address', 'bool', 'bytes32', 'uint256'];
+
+const header = `\
+pragma solidity ^0.8.24;
+
+/**
+ * @dev Library for reading and setting transcient storage.
```
Amxx, line 10:
```suggestion
 * @dev Library for reading and writting transcient storage.
```

allwin199 (commented):
@Amxx There is a typo in `writting` it should be `writing`

scripts/generate/templates/TransientStorage.js
```diff
@@ -0,0 +1,47 @@
...
+  /// @solidity memory-safe-assembly
+  assembly {
+      r := tload(slot)
+  }
+}
+`;
+
+const store = type => `\
+/**
+ * @dev Store \`value\` in the transient slot \`store\`.
```
ernestognw, line 30:
```suggestion
 * @dev Store a \`${type}`\ value in the transient slot \`store\`.
```

contracts/utils/TransientReentrancyGuard.sol
```diff
@@ -0,0 +1,56 @@
...
+pragma solidity ^0.8.24;
+
+import {TransientStorage} from "./TransientStorage.sol";
+
+/**
+ * @dev Variant of {ReentrancyGuard} that uses transient storage.
+ *
+ * NOTE: This variant only works on networks where EIP-1153 is available.
+ */
+abstract contract TransientReentrancyGuard {
```
ernestognw, line 12: What about `ReentrancyGuardTransient` instead? I feel the "transient" prefix captures the show:

```suggestion
abstract contract ReentrancyGuardTransient {
```

scripts/generate/templates/TransientStorage.js
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Note: This library only works on networks where EIP-1153 is available.
+ */
+`;
+
+const load = type => `\
+/**
+ * @dev Returns the transient slot \`slot\` as a \`${type}\`.
+ */
+function load${capitalize(type)}(bytes32 slot) internal view returns (${type} r) {
```
ernestognw, line 20: Any reason in particular to make the return value different to `StorageSlot`? The values there are defined as:

```
 struct Uint256Slot {
    uint256 value;
}
```

I definitely think this syntax is better, but I'm wondering if we should prefer one syntax over the other.

scripts/generate/templates/TransientStorage.js
```diff
@@ -0,0 +1,47 @@
...
+      r := tload(slot)
+  }
+}
+`;
+
+const store = type => `\
+/**
+ * @dev Store \`value\` in the transient slot \`store\`.
+ */
+function store(bytes32 slot, ${type} value) internal {
```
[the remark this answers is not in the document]
Amxx: Why? We don't need that.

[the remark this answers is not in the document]
ernestognw: Right. I was originally going to propose `tstore<Type>` as the name, and only `tstore` would conflict with the Yul instruction.

Let's solve this, I'm still making myself an opinion about its usage

scripts/generate/templates/TransientStorage.js
Amxx: AFAIK, Solidity supports structs in memory and in storage, but not transient ones. You can't mark it. Only thing we have is low level opcode access through assembly ?

scripts/generate/templates/TransientStorage.js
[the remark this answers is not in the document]
Amxx:
```suggestion
 * @dev Store ${type} \`value\` in the transient \`slot\` store.
```

scripts/generate/templates/TransientStorage.js
Amxx: Capture d’écran du 2024-03-14 20-49-50

Amxx: Capture d’écran du 2024-03-14 20-50-54

contracts/utils/TransientStorage.sol
frangio, line 1: Wouldn't it be nicer to provide an API more similar to the `StorageSlot` library?

A family of UDVT that encapsulate a transient location with `load()` and `store()` methods.

Amxx: Should we merge the two? (put the transient related tools in `StorageSlot` ?)

Amxx: I'm starting to think of something like that: https://gist.github.com/Amxx/86fc66c375a0ca73fa6d851066084466

scripts/generate/templates/TransientStorage.js
Amxx: See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4955/files#r1525423381

.changeset/honest-planets-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`TransientStorage`: Added a library to help read and write to transient storage (EIP-1153).
```
Amxx, line 5:
```suggestion
`Slots`: Added a library to help derivate, read and write to storage slots (both "normal" and transient).
```

.changeset/hungry-oranges-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`TransientReentrancyGuard`: Added a variant of `ReentrancyGuard` that uses transient storage.
```
Amxx, line 5:
```suggestion
`ReentrancyGuardTransient`: Added a variant of `ReentrancyGuard` that uses transient storage.
```

contracts/utils/README.adoc
```diff
@@ -13,6 +13,7 @@ Miscellaneous contracts and libraries containing utility functions you can use t
  * {MerkleProof}: Functions for verifying https://en.wikipedia.org/wiki/Merkle_tree[Merkle Tree] proofs.
  * {EIP712}: Contract with functions to allow processing signed typed structure data according to https://eips.ethereum.org/EIPS/eip-712[EIP-712].
  * {ReentrancyGuard}: A modifier that can prevent reentrancy during certain functions.
+ * {TransientReentrancyGuard}: Variant of {ReentrancyGuard} that uses transient storage (EIP-1153).
```
Amxx, line 16:
```suggestion
 * {ReentrancyGuardTransient}: Variant of {ReentrancyGuard} that uses transient storage (EIP-1153).
```

contracts/utils/README.adoc
```diff
@@ -29,6 +30,7 @@ Miscellaneous contracts and libraries containing utility functions you can use t
  * {Strings}: Common operations for strings formatting.
  * {ShortString}: Library to encode (and decode) short strings into (or from) a single bytes32 slot for optimizing costs. Short strings are limited to 31 characters.
  * {StorageSlot}: Methods for accessing specific storage slots formatted as common primitive types.
+ * {TransientStorage}: Methods for reading and writing to transient storage.
```
Amxx, line 33:
```suggestion
 * {Slots}: Methods for derivating, reading and writing to storage slots. This supports both "normal" and transient storage.
```

contracts/utils/README.adoc
```diff
@@ -64,6 +66,8 @@ Because Solidity does not support generic types, {EnumerableMap} and {Enumerable
 
 {{ReentrancyGuard}}
 
+{{TransientReentrancyGuard}}
```
Amxx, line 69:
```suggestion
{{ReentrancyGuardTransient}}
```

contracts/utils/README.adoc
```diff
@@ -110,6 +114,8 @@ Ethereum contracts have no native concept of an interface, so applications must
 
 {{StorageSlot}}
 
+{{TransientStorage}}
```
Amxx, line 117:
```suggestion
{{Slots}}
```

contracts/utils/Slots.sol
Amxx, line 1: @ernestognw @frangio Like this new approach?

Amxx:
Note: `StorageSlot` is more powerfull that the proposed `Slot`, because it supports operations sur as `+=`, `|=` or `&=`. It also supports non value types (because solidity is in charge of the code generation for reading/writting to the "slot", we just force its location).

A few question
- I think the slot derivation function are super usefull, but I'm wondering where to put them. Maybe they deserve a dedicated file.
- I think adding sstore and sload doesn't add complexity (just file length), so I'd keep that for completeness, even though it somehow overlaps with `StorageSlot`.
- Maybe `Slot.sol` needs a different name. Ideas?

frangio:
>I think adding sstore and sload doesn't add complexity (just file length), so I'd keep that for completeness, even though it somehow overlaps with StorageSlot.

I actually think it's better to remove those methods, and make the library purely about transient storage. It makes it a more sound abstraction for transient pointers, where there is only one correct dereferencing operation, and where it is not correct to cast into a normal storage pointer.

If those are removed `TransientSlot` may be a better name for the library. I think it's probably okay to leave the derivation methods in there since they are not necessary for storage slots? I'm not opposed to putting them in a separate file, but it would make it a little less convenient to use, since the current library only requires `using TransientSlot for *`.

Amxx:
THe derivation function have a use for normal storage, see the diff to `contracts/utils/Arrays.sol`

PR #4956, opened 2024-03-14
Update Solidity files in docs

ernestognw:
Found a couple of issues in the Solidity files from the documentation. I think it's easy to forget to update the files if they're not compiled.

I moved those files I thought convenient and fixed a couple of compilation issues. Also took the freedom to replace error strings with custom errors since that should be our go-to recommendation

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/contracts/mocks/docs/utilities/Base64NFT.sol
+++ b/contracts/mocks/docs/utilities/Base64NFT.sol
@@ -22,6 +22,6 @@ contract Base64NFT is ERC721 {
         // }
         bytes memory dataURI = abi.encodePacked("{", '"name": "Base64NFT #', tokenId.toString(), '"', "}");
 
-        return string.concat("data:application/json;base64,", Base64.encode(dataURI));
+        return string(abi.encodePacked("data:application/json;base64,", Base64.encode(dataURI)));
     }
 }
--- a/contracts/mocks/docs/utilities/Base64NFT.sol
+++ b/contracts/mocks/docs/utilities/Base64NFT.sol
@@ -1,11 +1,10 @@
-// contracts/Base64NFT.sol
 // SPDX-License-Identifier: MIT
 
 pragma solidity ^0.8.20;
 
-import {ERC721} from "../../../token/ERC721/ERC721.sol";
-import {Strings} from "../../../utils/Strings.sol";
-import {Base64} from "../../../utils/Base64.sol";
+import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+import {Base64} from "@openzeppelin/contracts/utils/Base64.sol";
 
 contract Base64NFT is ERC721 {
     using Strings for uint256;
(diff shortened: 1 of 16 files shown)


contracts/mocks/docs/utilities/Base64NFT.sol
```diff
@@ -0,0 +1,27 @@
...
+
+    // ...
+
+    function tokenURI(uint256 tokenId) public pure override returns (string memory) {
+        // Equivalent to:
+        // {
+        //   "name": "Base64NFT #1",
+        //   // Replace with extra ERC-721 Metadata properties
+        // }
+        bytes memory dataURI = abi.encodePacked("{", '"name": "Base64NFT #', tokenId.toString(), '"', "}");
```
Amxx, line 23: In these kind of application, I think using `string.concat` is clearer.
Also, things can be merges before.
```suggestion
        string memory dataURI = string.concat("{\"name\": \"Base64NFT #", tokenId.toString(), "\"}");
```

contracts/mocks/docs/utilities/Base64NFT.sol
```diff
@@ -0,0 +1,27 @@
...
+
+        return string(abi.encodePacked("data:application/json;base64,", Base64.encode(dataURI)));
```
Amxx, line 25:
```suggestion
        return string.concat("data:application/json;base64,", Base64.encode(dataURI));
```

ernestognw:
I tried the suggestions but they didn't compile. I could make it compile but since it doesn't have tests, I need to make sure they produce the same output. Will do tomorrow.

Amxx:
We ,eed to find a way to make tests and linting pass, otherwise this is going to affect all other PRs.

contracts/mocks/docs/utilities/Base64NFT.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
```
Amxx, line 5: We should not need the `@openzeppelin/contracts`. The `../../..` are supposed to be replaced automatically during doc generation

ernestognw:
The linter fails because there's a conflict between a rule configured for Prettier to avoid single quotes in Solidity files:

```json
{
  ...
  "singleQuote": true,
  "overrides": [
    {
      "files": "*.sol",
      "options": {
        "singleQuote": false
      }
    }
  ],
  "plugins": ["prettier-plugin-solidity"]
}
```

So whenever we run `npm run lint:fix`, it makes the following:

```diff
-        string memory dataURI = string.concat("{\"name\": \"Base64NFT #", tokenId.toString(), "\"}");
+        string memory dataURI = string.concat('{"name": "Base64NFT #', tokenId.toString(), '"}');
```

Which then conflicts with the override for Solidity files.

I added a `prettier-ignore` comment since I found it the less intrusive way to make the CI pass without affecting any other configuration.

Amxx (approved):

PR #4964, opened 2024-03-18
Remove outdated comment in ERC-20

alexfertel:


ernestognw (approved):
We are already explicit about this in other functions so I think this is fine.

As always, let's wait for @Amxx approval. Thanks!

Amxx (approved):

Issue #4696, opened 2023-10-21
ERC-7201 base contract


aviggiano:
**🧐 Motivation**

Hello

The "Namespaced Storage" pattern introduced in contracts v5 is a great feature, and will certainly help mitigate many problems associated with storage layout changes in upgradeable contracts.

However, the general issue still exists.

Since the implementation of ERC-7201 is only implemented in the OpenZeppelin contracts, and not enforced anywhere on users' contracts, developers can still make the same mistakes by messing with inherited contracts' storage variables.

Given that the thorough implementation of "Namespaced Storage" can be a bit challenging, requiring a little bit of inline assembly to fetch and update the storage, I suggest an ERC-7201 base contract that users can inherit from and use the diamond storage pattern everywhere.

**📝 Details**

As a Solidity developer, I would like to have a `ERC7201BaseContract` that all my custom contracts extend from, with helper methods that make it easy to retrieve and update data from storage.

ernestognw:
Hi @aviggiano!

I agree messing with the storage layout is still a risk when using upgradeable contracts. However, that can be avoided by using [@openzeppelin/upgrades](https://github.com/OpenZeppelin/openzeppelin-upgrades/#how-do-the-plugins-work) since they have upgrade safetiness checks that also support the namespaced storage pattern (soon to be out for Foundry users as well).

> However, the general issue still exists.

Yes, ideally we would have built-in safety instead of requiring checks. That's why we decided to implement EIP-7201 in upgradeable contracts, which is where the storage layout incompatibilities might happen. Aside from that, I don't see how it might be an issue for regular contract users.

Can you provide an example of how storage layout incompatibilities are still a risk even after introducing EIP-7201?

Regarding an ERC7201 base contract implementation, seems possible to do something like:

```
contract ERC7201Base {
  bytes private immutable STORAGE;

  constructor(string memory name) {
    STORAGE = keccak256(abi.encode(uint256(keccak256(name)) - 1)) & ~bytes32(uint256(0xff))
  }

    function getStorage() internal pure returns (InitializableStorage storage $) {
        assembly {
            $.slot := STORAGE
        }
    }
}
```

But, I think the storage struct should be declared outside and we already have an [storage slot](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3e6c86392c97fbc30d3d20a378a6f58beba08eba/contracts/utils/StorageSlot.sol) util.

aviggiano:
[...]

PR #4965, opened 2024-03-18
StorageSlot library : add storage slot derivation tooling and UDVT typed slot representation.

### LINKED ISSUE #4961: Library for derivation of slots
Solidity does the derivation of slot under the hood when trying to access a value in an array or a mapping. 
Users may want to do te same to compute slots for transient storage, which solidity does not support yet.

[This library](https://github.com/Amxx/EIP-5806/blob/main/packages/test-uniswap/contracts/libs/Slots.sol) does exactly that.

See it used [here](https://github.com/Amxx/EIP-5806/blob/main/packages/test-uniswap/contracts/libs/ERC20TemporaryAllowance.sol#L45-L51)

Amxx:
First part of #4955, Provides the building block for more transient storage work (reentrancy guard, temporary approvals, ...)

Fixes #4961
Addresses parts of #4696

Note: we decided to keep normal storage support to the UDVT representation, for the sake of compliteness, and because we don't have any security concern with it.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


hardhat.config.js
```diff
@@ -90,19 +96,21 @@ module.exports = {
...
+      hardfork: argv.evm,
       allowUnlimitedContractSize,
       initialBaseFeePerGas: argv.coverage ? 0 : undefined,
     },
   },
   exposed: {
     imports: true,
     initializers: true,
-    exclude: ['vendor/**/*', '**/*WithInit.sol'],
+    exclude: ['vendor/**/*', '**/*WithInit.sol', 'utils/StorageSlot.sol'],
```
Amxx, line 113:
```suggestion
    exclude: ['vendor/**/*', '**/*WithInit.sol', 'utils/StorageSlot*.sol'],
```

scripts/generate/templates/StorageSlot.js
```diff
@@ -1,17 +1,8 @@
...
-  { type: 'bytes32', isValueType: true },
-  { type: 'uint256', isValueType: true },
-  { type: 'string', isValueType: false },
-  { type: 'bytes', isValueType: false },
-].map(type => Object.assign(type, { struct: (type.name ?? capitalize(type.type)) + 'Slot' }));
+const { TYPES } = require('./StorageSlot.opts');
 
 const header = `\
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.24;
```
Amxx, line 5: This is technically a breaking change :/
I don't think its a major issue though

frangio:
> Note: we decided to keep normal storage support to the UDVT representation, for the sake of compliteness, and because we don't have any security concern with it.

I want to elaborate on this point just a little more in case I'm able to change your minds 🙂. My concern is with mixing these things is that correct use of the abstraction relies on remembering/checking at each use site that the right storage space (transient or normal) is being accessed. The wrong decision at a single use site can result in a massive bug. I would prefer it in my code if I only made this decision at the variable declaration site, and subsequent load/store operations involving this variable automatically access the correct storage space.

Amxx:
> I want to elaborate on this point just a little more in case I'm able to change your minds 🙂. My concern with mixing these things is that correct use of the abstraction relies on remembering/checking at each use site that the right storage space (transient or normal) is being accessed. The wrong decision at a single use site can result in a massive bug. I would prefer it in my code if I only made this decision at the variable declaration site, and subsequent load/store operations involving this variable automatically access the correct storage space.

That is an interresting view. Not sure I share that concern.

contracts/utils/StorageSlot.sol
```diff
@@ -29,28 +29,159 @@ pragma solidity ^0.8.20;
...
-        bool value;
+    /**
+     * @dev Derive an ERC-7201 slot from a string (path).
+     */
+    function erc7201slot(string memory path) internal pure returns (bytes32 slot) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0x00, sub(keccak256(add(path, 0x20), mload(path)), 1))
+            slot := and(keccak256(0x00, 0x20), not(0xff))
+        }
```
ernestognw, line 51: Wouldn't be a "namespace" instead?

```suggestion
    /**
     * @dev Derive an ERC-7201 slot from a string (namespace).
     */
    function erc7201slot(string memory namespace) internal pure returns (bytes32 slot) {
        /// @solidity memory-safe-assembly
        assembly {
            mstore(0x00, sub(keccak256(add(path, 0x20), mload(namespace)), 1))
            slot := and(keccak256(0x00, 0x20), not(0xff))
        }
```

scripts/generate/templates/StorageSlot.js
ernestognw: Yeah, I think it's fine

hardhat.config.js
```diff
@@ -90,19 +96,21 @@ module.exports = {
     '*': {
       'code-size': withOptimizations,
       'unused-param': !argv.coverage, // coverage causes unused-param warnings
+      'transient-storage': false,
```
ernestognw, line 99: I couldn't find this option in the docs, where's it coming from?

scripts/generate/templates/StorageSlot.js
```diff
@@ -39,40 +30,179 @@ pragma solidity ^0.8.20;
...
+function offset(bytes32 slot, uint256 pos) internal pure returns (bytes32 result) {
+  unchecked {
+      return bytes32(uint256(slot) + pos);
+  }
+}
+
+/**
+ * @dev Derive the location of the first element in an array from the slot where the length is stored.
+ *
+ * See: https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays.
```
ernestognw, line 67: I think we should reference these as a link (same applies to others):

```suggestion
 * See https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays[Solidity docs for mappings and dynamic arrays.]
```

ernestognw:
After reviewing I see @frangio's concerns about mixing transient and regular storage.
In my opinion, it's not a big deal but after reviewing I felt a bit lost navigating the StorageSlot library on its own.

From the Solidity file, there are ~500 lines of code only for "storage slots". I would say it's expected if users get confused with all the options in the library.

Currently, I don't have a strong opinion of how to split this, but splitting Transient from regular Storage (eg. TransientSlots vs StorageSlots) seems a good starting point.

Also side note, but, I couldn't find any plans for Solidity to add a language feature that enables transient storage in a variable. I'm wondering how much this work may overlap with such a feature if they ever release it. I'm feeling this PR is adding a bunch of features that might be as well the language responsibility.

hardhat.config.js
Amxx: its a new one introduced in 0.8.24, I had to make a PR to add this https://github.com/frangio/hardhat-ignore-warnings/pull/6/files

Amxx:
> Also side note, but, I couldn't find any plans for Solidity to add a language feature that enables transient storage in a variable. I'm wondering how much this work may overlap with such a feature if they ever release it. I'm feeling this PR is adding a bunch of features that might be as well the language responsibility.

Its not clear to me how 0.8 is evolving, and at what pace. I think most of the effort is going toward the overall redesign. I agree it would be better is solidity just supported this out of the box ... but we have no clue when that will be possible, and people are activelly asking for some solution to handle transient storage.

Amxx:
> From the Solidity file, there are ~500 lines of code only for "storage slots". I would say it's expected if users get confused with all the options in the library.

SafeCast is over twice as long. I think its ok to have long file if its just many clear variation of a few simple features. 

> Currently, I don't have a strong opinion of how to split this, but splitting Transient from regular Storage (eg. TransientSlots vs StorageSlots) seems a good starting point.

If we do that, do we :
- remove sstore/sload from the UDVT approach?
  - If not, where do we define the UDVT types? 
    - In `StorageSlot.sol` (with the sstore/ssload) and then import them in TransiantSlot.sol?
    - elsewhere?
- where do we put the namespace & derivation functions ?

ernestognw:
The PR is looking good but I'm still not convinced about the design. Sorry for the back and forth but after reviewing I came to the following conclusions:

- `TransientSlot` is not a "transient" version of `StorageSlot`, they do very different things.
- Even if we define the custom UDVTs in each `StorageSlot` and `TransientSlot` libraries, it shouldn't be a problem because I don't think we want to encourage using them indistinctly (i.e. it's ok that `StorageSlot.Uint256SlotType` is different to `TransientSlot.Uint256SlotType` for the compiler)
- Providing any of the StorageSlot struct types (e.g. `struct AddressSlot {...}`) as an argument to a function requires `memory` (more on this below)

Right now I'm leaning towards the following 3 libraries:

- `StorageSlot`: Leave it as it is right now.
- `SlotDerivation`: Leave it as it is right now
- `TransientSlot`: 
  - Rename this one to `TransientStorage`: This reflects that the library is for actually storing.
  - We leave the types from `TypedSlot` here and rename them as `AddressTransientType` instead of `AddressSlotType`.
  - Similarly, we rename the functions from `asAddressSlot` to `toTransientAddressSlot` (or `asTransientAddressSlot`)


If my framing is correct, this sets apart transient storage operations from regular storage ones, and the only common thing between them is the library with the derivation functions. I wouldn't try to make both `TransientStorage` and `StorageSlot` 
 to use the same value types, this is because assigning would work very differently depending on the type of storage:

```solidity
// For regular storage
getAddressSlot().value = '0x...'

// For transient storage
slot.toTransientAddress().tstore('0x...')
```

Trying to use the same types we already have in StorageSlot would result in (iugh, memory):

```solidity
struct AddressSlot {
    address value;
}

function tstore(TypedSlot.AddressSlotType slot, AddressSlot memory v) internal {
    address value = v.value;
    assembly {
        tstore(slot, value)
    }
}
```

I see that a common interface between `TransientStorage` and `StorageSlots` is possible without using memory if we share the UDVTs, but I'd be fine if they are not compatible out of the box.

scripts/generate/templates/TypedSlot.js
```diff
@@ -0,0 +1,32 @@
...
+ * @dev UDVT that represent a slot holding a ${type}.
+ */
+type ${name}SlotType is bytes32;
+
+/**
+ * @dev Cast an arbitrary slot to a ${name}SlotType.
+ */
+function as${name}Slot(bytes32 slot) internal pure returns (${name}SlotType) {
+  return ${name}SlotType.wrap(slot);
+}
```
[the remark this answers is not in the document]
Amxx: What usecase do you see ?
I think once a slot is typed, its ready to use to load/store. I'm not sure it makes sens to go back to bytes32 to derive further. 

Worst case, it someone really wants to do that, they can do `TypedSlot.Uint256SlotType.unwrap(...)`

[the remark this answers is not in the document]
ernestognw: > Worst case, it someone really wants to do that, they can do TypedSlot.Uint256SlotType.unwrap(...)

Fair, I didn't consider `unwrap` is already chain-able.

scripts/generate/templates/TypedSlot.js
```diff
@@ -0,0 +1,32 @@
...
+const udvt = ({ type, name }) => `\
+/**
+ * @dev UDVT that represent a slot holding a ${type}.
+ */
+type ${name}SlotType is bytes32;
+
+/**
+ * @dev Cast an arbitrary slot to a ${name}SlotType.
+ */
+function as${name}Slot(bytes32 slot) internal pure returns (${name}SlotType) {
```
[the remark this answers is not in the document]
Amxx: I'm not sure. I think "asUint256Slot" is more representative.

`toDelay` takes a duration that is a uint32, and transform it to a delay object, that is a uint112, and that represent more (there is space for the pending value, the effect timepoint...). So IMO its a "to" because there is a transformation (even if its just adding zeros)

In this case, its a bytes32 that remains a bytes32 ... just with a wrapper. Its really like `as` in typescript.

Amxx:
Renaming `TransientSlot` to `TransientStorage` is good to me.

However, I'm not really onboard with the other proposed changes. I think its a question of vision. IMO the storage and transient storage spaces are like 2 view, with different lifecycle, of the same thing. Like 2 plans on top of one another.

Consequently, I believe that a (derived) storage location is equally valid in storage and in transiant storage. One example is the ReentrancyGuard contract, where I think the slot that is currently used with normal storage should be used exactly the same, but in the transiant storage space. This is why I initially imagined the Typed-slot UDVT would provide sload and sstore in addition to tstore and tload. I'm ok with only implementing the transient side for now, and putting that in a dedicated library (TransientStorage) ... but I would not go as far as using a transient specific UDVT.

Again, I think the UDVT represent a "location + type" that is valid in both spaces.

ernestognw:
I agree that a location is valid for both storage and transient storage, but I guess the main difference here comes from the Solidity. Given the language doesn't treat both types of storage in the same way, we're limited by the current syntax of each one. It would make sense to share an interface if both were the same at the language level, but they're only very similar in assembly.

The way I read this situation is that in order to make both share the same interface, the only way is to use assembly for both. However, the native Solidity syntax is preferred, so that sets transient storage in a different category only because it can't be used with the same language features

>  I'm ok with only implementing the transient side for now, and putting that in a dedicated library (TransientStorage) ... but I would not go as far as using a transient specific UDVT.

The UDVT is useful here because we don't have a way to tell the compiler what type we're using (e.g. `address`), so we're using the UDVT as a wrapper to circumvent this limitation. This makes it more specific to transient storage than to "slots". Wouldn't you say `as<type>Slot` has a name conflict with `StorageSlot.<type>Slot`?

Amxx:
> The UDVT is useful here because we don't have a way to tell the compiler what type we're using (e.g. address), so we're using the UDVT as a wrapper to circumvent this limitation. This makes it more specific to transient storage than to "slots". Wouldn't you say as<type>Slot has a name conflict with StorageSlot.<type>Slot?

The need for UDVT, as an approach, is specific to transient storage for sure, but no, I don't think the UDVT types themselves are specific to transient storage. Again they just represent a "position+type".

Yes naming can be confusing. We already have a "conflict" between `AddressSlotType` (UDVT) and `AddressSlot` (struct). Maybe we can rename `as<type>Slot` to something like `as<type>TypedSlot` ... but again I would avoid mentionning "Transient" here, because there is nothing specific to "transient" in the UDVT definition.

Amxx:
Replaced by #4980

PR #4968, opened 2024-03-21
Remove outdated comment in ERC-20

alexfertel:
Missed this other outdated comment in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4964

### DIFF
--- a/contracts/token/ERC20/ERC20.sol
+++ b/contracts/token/ERC20/ERC20.sol
@@ -134,7 +134,7 @@ abstract contract ERC20 is Context, IERC20, IERC20Metadata, IERC20Errors {
      * @dev See {IERC20-transferFrom}.
      *
      * Skips emitting an {Approval} event indicating an allowance update. This is not
-     * required by the ERC. See {xref-ERC20-_approve-address-address-uint256-bool-}.
+     * required by the ERC. See {xref-ERC20-_approve-address-address-uint256-bool-}[_approve].
      *
      * NOTE: Does not update the allowance if the current allowance
      * is the maximum `uint256`.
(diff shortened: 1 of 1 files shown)


contracts/token/ERC20/ERC20.sol
```diff
@@ -133,9 +133,6 @@ abstract contract ERC20 is Context, IERC20, IERC20Metadata, IERC20Errors {
     /**
      * @dev See {IERC20-transferFrom}.
      *
-     * Emits an {Approval} event indicating the updated allowance. This is not
-     * required by the ERC. See the note at the beginning of {ERC20}.
-     *
```
ernestognw, line 138: In my opinion, this section is important because it makes it explicit that the ERC does not require the event. That's still true. Let's rephrase it to: 

```suggestion
Skips emitting an {Approval} event indicating an allowance update. This is not
required by the ERC. See {xref-ERC20-_approve-address-address-uint256-bool-}.
```

ernestognw (commented):
On a side note, I was reviewing to make sure we're not missing something again in the docs, and turns out we're still warning users about [`approve` frontrunning:](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/7bd2b2aaf68c21277097166a9a51eb72ae239b34/contracts/token/ERC20/IERC20.sol#L58-L63) when we already deprecated the non-standard `increaseAllowance` and `decreaseAllowance` functions.

```
* IMPORTANT: Beware that changing an allowance with this method brings the risk
* that someone may use both the old and the new allowance by unfortunate
* transaction ordering. One possible solution to mitigate this race
* condition is to first reduce the spender's allowance to 0 and set the
* desired value afterwards:
* https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
```

I wonder if we should remove the warning. We agree is a theoretical issue but doesn't hurt anyone 🤔

ernestognw (approved):
LGTM. Merging if CI pass

alexfertel:
Thank you!

Amxx (approved):

ernestognw:
Testing @gitpoap-bot @alexfertel

PR #4969, opened 2024-03-21
Fix excessive gas usage when at or near cap.

### LINKED ISSUE #4967: ERC20 _update customization is not practical.
You can't actually modify behavior due to the variables being private.

    /**
     * @dev Transfers a `value` amount of tokens from `from` to `to`, or alternatively mints (or burns) if `from`
     * (or `to`) is the zero address. All customizations to transfers, mints, and burns should be done by overriding
     * this function.
     *
     * Emits a {Transfer} event.
     */
    function _update(address from, address to, uint256 value) internal virtual {

Let's say I want to customize the burn to increment the balance of the 0 adresss and decrement the total supply, It's not currently possible.

**💻 Environment**

OpenZeppelin Contracts (last updated v5.0.0) (token/ERC20/ERC20.sol)

**📝 Details**

**🔢 Code to reproduce bug**

BillSchumacher:
Fixes #4967

Saves gas by completely avoiding _update if the cap will be exceeded.
image

The "After" test is the current usage and the non-After OverCap test is the PR's usage.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC20/extensions/ERC20Capped.sol
+++ b/contracts/token/ERC20/extensions/ERC20Capped.sol
@@ -43,14 +43,14 @@ abstract contract ERC20Capped is ERC20 {
      * @dev See {ERC20-_update}.
      */
     function _update(address from, address to, uint256 value) internal virtual override {
-        super._update(from, to, value);
-
         if (from == address(0)) {
             uint256 maxSupply = cap();
             uint256 supply = totalSupply();
-            if (supply > maxSupply) {
+            if (supply + value > maxSupply) {
                 revert ERC20ExceededCap(supply, maxSupply);
             }
         }
+        
+        super._update(from, to, value);
     }
 }


ernestognw:
Sorry, I'll close. Let me be clear about a couple of things:

- https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4967 is a non-issue and seems unrelated to this PR aside from the "gas savings" (which are none).
- There's no point in optimizing the gas consumption whenever the function will revert. If the reasoning is that a user will pay less if the transaction fails, that's completely wrong! Any wallet will prevent the user from executing an operation if it'll fail. This is easy to catch with a gas estimation that every wallet does.
- The order of operations is important. By accepting this change we need to make sure that every possible side effect from the order of execution keeps the invariants (possibly yes but there's still no point in these changes). This is why we book an audit before a release.
- A screenshot of your tests is **no proof** of any gas improvement, even if it was true. We can't merge without the code and a proper analysis of what you're doing, including compiler settings.

PR #4975, opened 2024-03-26
Add slot derivation library

Amxx:
This takes the non-controversial part of #4965, while we figure out how we want to handle transiant storage.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/SlotDerivation.sol
+++ b/contracts/utils/SlotDerivation.sol
@@ -20,7 +20,7 @@ pragma solidity ^0.8.20;
  *     string private constant _NAMESPACE = "<namespace>" // eg. OpenZeppelin.Slot
  *
  *     function storagePointer() internal view returns (bytes32) {
- *         return _NAMESPACE.erc7201Slot(); // or erc1967Slot()
+ *         return _NAMESPACE.erc7201Slot();
  *     }
  * }
  * ```
--- a/contracts/utils/SlotDerivation.sol
+++ b/contracts/utils/SlotDerivation.sol
@@ -14,13 +14,18 @@ pragma solidity ^0.8.20;
  * ```solidity
  * contract Example {
  *     // Add the library methods
+ *     using StorageSlot for bytes32;
  *     using SlotDerivation for bytes32;
  *
  *     // Declare a namespace
  *     string private constant _NAMESPACE = "<namespace>" // eg. OpenZeppelin.Slot
  *
- *     function storagePointer() internal view returns (bytes32) {
- *         return _NAMESPACE.erc7201Slot();
+ *     function setValueInNamespace(uint256 key, address newValue) internal {
+ *         _NAMESPACE.erc7201Slot().deriveMapping(key).getAddressSlot().value = newValue;
+ *     }
+ *
+ *     function getValueInNamespace(uint256 key) internal view returns (address) {
+ *         return _NAMESPACE.erc7201Slot().deriveMapping(key).getAddressSlot().value;
  *     }
  * }
  * ```
--- a/contracts/utils/SlotDerivation.sol
+++ b/contracts/utils/SlotDerivation.sol
@@ -9,12 +9,29 @@ pragma solidity ^0.8.20;
  * the solidity language / compiler.
  *
  * See https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays[Solidity docs for mappings and dynamic arrays.].
+ *
+ * Example usage:
+ * ```solidity
+ * contract Example {
+ *     // Add the library methods
+ *     using SlotDerivation for bytes32;
+ *
+ *     // Declare a namespace
+ *     string private constant _NAMESPACE = "<namespace>" // eg. OpenZeppelin.Slot
+ *
+ *     function storagePointer() internal view returns (bytes32) {
+ *         return _NAMESPACE.erc7201Slot(); // or erc1967Slot()
+ *     }
+ * }
+ * ```
+ *
+ * TIP: Consider using this library along with {StorageSlot}.
  */
 library SlotDerivation {
     /**
      * @dev Derive an ERC-1967 slot from a string (namespace).
      */
-    function erc1967slot(string memory namespace) internal pure returns (bytes32 slot) {
+    function erc1967Slot(string memory namespace) internal pure returns (bytes32 slot) {
         /// @solidity memory-safe-assembly
         assembly {
             slot := sub(keccak256(add(namespace, 0x20), mload(namespace)), 1)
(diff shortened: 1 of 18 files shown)


Amxx:
We used to have `codecov` actions, but they don't show up anymore

scripts/generate/templates/SlotDerivation.js
```diff
@@ -0,0 +1,101 @@
...
+
+const header = `\
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Library for computing storage (and transient storage) locations from namespaces and deriving slots
+ * corresponding to standard patterns. The derivation method for array and mapping matches the storage layout used by
+ * the solidity language / compiler.
+ *
+ * See https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays[Solidity docs for mappings and dynamic arrays.].
```
ernestognw, line 12: Let's add an example usage here:

```suggestion
 * See https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html#mappings-and-dynamic-arrays[Solidity docs for mappings and dynamic arrays.].
 *
 * Example usage:
 * \`\`\`solidity
 * contract Example {
 *     // Add the library methods
 *     using SlotDerivation for bytes32;
 *
 *     // Declare a namespace
 *     string private constant _NAMESPACE = "<namespace>" // eg. OpenZeppelin.Slot
 *  
 *     function storagePointer() internal view returns (bytes32) {
 *         return _NAMESPACE.erc7201Slot();
 *     }
 * }
 * \`\`\`
```

scripts/generate/templates/SlotDerivation.js
```diff
@@ -0,0 +1,101 @@
...
+ */
```
ernestognw, line 13: Would it make sense to recommend using along with StorageSlot?

```suggestion
 * TIP: Consider using this library along with {StorageSlot}
 */
```

scripts/generate/templates/SlotDerivation.js
```diff
@@ -0,0 +1,101 @@
...
+`;
+
+const namespace = `\
+/**
+ * @dev Derive an ERC-1967 slot from a string (namespace).
+ */
+function erc1967slot(string memory namespace) internal pure returns (bytes32 slot) {
```
ernestognw, line 20: We write function names in camel case, shouldn't "slot" be capitalized?

```suggestion
function erc1967Slot(string memory namespace) internal pure returns (bytes32 slot) {
```

scripts/generate/templates/SlotDerivation.js
```diff
@@ -0,0 +1,101 @@
...
+  /// @solidity memory-safe-assembly
+  assembly {
+      slot := sub(keccak256(add(namespace, 0x20), mload(namespace)), 1)
+  }
+}
+
+/**
+ * @dev Derive an ERC-7201 slot from a string (namespace).
+ */
+function erc7201slot(string memory namespace) internal pure returns (bytes32 slot) {
```
ernestognw, line 30: Same

```suggestion
function erc7201Slot(string memory namespace) internal pure returns (bytes32 slot) {
```

contracts/utils/SlotDerivation.sol
```diff
@@ -0,0 +1,163 @@
...
+library SlotDerivation {
+    /**
+     * @dev Derive an ERC-1967 slot from a string (namespace).
+     */
+    function erc1967Slot(string memory namespace) internal pure returns (bytes32 slot) {
+        /// @solidity memory-safe-assembly
+        assembly {
+            slot := sub(keccak256(add(namespace, 0x20), mload(namespace)), 1)
+        }
+    }
```
ernestognw, line 39: For ERC1967 there are only 3 valid namespaces:

- 'eip1967.proxy.implementation'
- 'eip1967.proxy.admin'
- 'eip1967.proxy.beacon'

I think we should not expose this function since it may be dangerous (i.e. is it collision resistant as erc7201 is?), and rather only expose the slots that are already defined in the EIP. Our go-to recommendation for deriving slots should be always ERC7201

ernestognw (changes requested):
I improved the docs including examples for both SlotDerivation and StorageSlot (and Merkle Tree, je).

I'd like to reconsider exposing the `erc1967Slot`

contracts/utils/SlotDerivation.sol
ernestognw: Concretely, we might do:

```solidity
function erc1967Admin(string memory namespace) internal pure returns (bytes32 slot) {
    return 0xb53127684a568b3173ae13b9f8a6016e243e63b6e8ee1178d6a717850b5d6103;
}

function erc1967Implementation(string memory namespace) internal pure returns (bytes32 slot) {
    return 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
}

function erc1967Beacon(string memory namespace) internal pure returns (bytes32 slot) {
    return 0xa3f0ad74e5423aebfd80d3ef4346578335a9a72aeaee59ff6cb3582b35133d50;
}
```

contracts/utils/SlotDerivation.sol
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * @dev Add an offset to a slot to get the n-th element of a structure or an array.
+     */
+    function offset(bytes32 slot, uint256 pos) internal pure returns (bytes32 result) {
+        unchecked {
+            return bytes32(uint256(slot) + pos);
+        }
```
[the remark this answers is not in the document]
Amxx: Slot location do wrap around in solidity. Just try this in remix:

```solidity

pragma solidity ^0.8.20;

contract Test {
    struct Uint256Slot { uint256 value; }
    Uint256Slot[] internal m_array;

    constructor() {
        assembly { sstore(m_array.slot, not(0)) }
    }

    function length() public view returns (uint256) {
        return m_array.length;
    } 

    function firstSlot() public view returns (uint256 result) {
        return slotId(m_array[0]);
    } 
    
    function lastSlot() public view returns (uint256 result) {
        return slotId(m_array[m_array.length - 1]);
    } 

    function zeroSlot() public view returns (uint256 result) {
        return slotId(m_array[type(uint256).max - firstSlot() + 1]);
    } 

    function slotId(Uint256Slot storage obj) private pure returns (uint256 slot) {
        assembly { slot := obj.slot }
    }
}
```

In any real usecase, the wrap around should not happen. It is highly unexpected, and potentially dangerous ... but its also solidity's behavior, and I think we should reproduce it.

contracts/utils/SlotDerivation.sol
frangio: I'm more on the side of @ernestognw on this one. It's true that the formula is used in other places, but I don't know that I would call those "ERC-1967".

Amxx: So I guess we remove this function then

contracts/utils/StorageSlot.sol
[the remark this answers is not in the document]
Amxx: added

ernestognw (approved):
LGTM! 🚀

PR #4976, opened 2024-03-27
Branchless ternary, min and max methods

Lohann:
# Description
Follow a Branchless Min and Max functions, all of them consumes less gas than the current implementation, and have a constant gas cost independent of the input value.

# Motivation
I'm working in the [Analog Timechain](https://www.analog.one/), which is a cross-chain protocol that allow interoperability between different blockchains.

At some point our protocol needs for a given input calculate the exact execution gas cost, but this was hard due a lot of branching, branching that should not even exist, for example the `Math.min` function use more or less gas when `a < b` or `a >= b`, either compiling in debug and optimized mode:
```solidity
pragma solidity >=0.7.0 <0.9.0;

contract Math {
    function min(uint256 a, uint256 b) external pure returns (uint256 result) {
        result = a < b ? a : b;
    }
}
```
`a < b` consumes `306 gas`
`a >= b` consumes `316 gas`

This was annoying because for every change in the contract I had to dive into the EVM assembly to understand how solidity generate different branches to understand how it affect the gas used. So I decided to take a different approach and make everything constant gas cost, as result I implemented the [BranchlessMath.sol](https://github.com/Analog-Labs/analog-gmp/blob/42a7223b44141a9028f39c7aff9f6cd9c75c1196/src/utils/BranchlessMath.sol) library, which have a constant gas cost for any input, and surprisingly I also noticed significant gas savings using those methods compared to the OpenZeppeling's Math library, so I decided to make it available to everyone by contributing here.

In comparison the branchless version reduces the final bytecode size, and also consumes less gas:
```solidity
contract Math {
    function min(uint256 a, uint256 b) external pure returns (uint256 result) {
        assembly ("memory-safe") {
            // min(a,b) = b ^ ((a ^ b) * (a < b))
            result := xor(b, mul(xor(a, b), lt(a, b)))
        }
    }
}
```
Always consumes `291 gas`, independently if `a < b` or `a >= b`.

#### PR Checklist
- [x] Tests
- [x] Documentation


Amxx (commented):
Hello @Lohann and thank you for this PR.

The min/max option is interresting, because branches are indeed notoriously bad. Code readability is not that great though. If we don't have them yet, this would definitelly need to be fuzzed.

I'm not that fan of the rest. The objective of this library has always been security, readability, and auditability. Regardless of what some other project may think, we strongly believe in readability (and documentation through comments) of the code.

Therefore, I'm not leaning toward a full inline assembly version, nor am I willing to remove the documentation comments (that we took so long to make sure are clear)

Lohann:
@Amxx thanks for the feedback, will remove some assembly at the cost of a consuming a bit more gas, but essentially what the sqrt does is:

```solidity
function sqrt(x) public pure returns (uint256) {
  unchecked {
    uint256 xn = 2 ** (Math.log2(x) / 2);
    
    // 7 iterations of Netwon's method
    xn = (xn + a / xn) >> 1;
    xn = (xn + a / xn) >> 1;
    xn = (xn + a / xn) >> 1;
    xn = (xn + a / xn) >> 1;
    xn = (xn + a / xn) >> 1;
    xn = (xn + a / xn) >> 1; 
    xn = (xn + a / xn) >> 1;
    
    // Round down
    xn = Math.min(xn, a / xn);
    
    return xn;
  }
}
```

Amxx:
[This is almost exactly what we used to have](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/96e5c0830a83b61197dd4e29df59cb56499600ef/contracts/utils/math/Math.sol#L390-L421)

It got changed in a long and painfull (to review) PR. This included documenting a lot of math to prove (and not just feel) that the new version is correct (in addition to being cheaper).

If we are to go back on sqrt, I'll do it in a separate issue/PR. I'd rather split the review effort and the frustration. I really see a point to merging min/max decently fast ... and I fear sqrt will take much more discussion.

Lohann:
@Amxx makes sense, I removed the sqrt changes, I also removed all assembly code maintaining the same gas efficiency to `min` and `max` methods.

I also made it a bit more generic by introducing the `choice` function, which is the core principle on how the branchless min/max works, it can be used to replace any simple ternary operations.
Because it is a new method, I incremented the `minor` instead of the `patch`, but if you think we should not expose this function, I can make the `choice` private, even so I think it is quite useful.

Amxx:
> Because it is a new method, I incremented the minor instead of the patch.

Anythink that is not a security fix goes into a minor anyway.

> but if you think we should not expose this function, I can make the choice private, even so I think it is quite useful.

We'll discuss that.

Thank you for your effort ... though please don't be frustrated if this takes time to merge. Our processes are ususally considered slow by external contributors, but we fell that is how we achieve maximum quality/security.

contracts/utils/math/Math.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Contracts (last updated v5.0.0) (utils/math/Math.sol)
+// OpenZeppelin Contracts (last updated v5.0.2) (utils/math/Math.sol)
```
Amxx, line 2: This is going to be updated automatically during the release process, we should not manually touch it.
```suggestion
// OpenZeppelin Contracts (last updated v5.0.0) (utils/math/Math.sol)
```

contracts/utils/math/Math.sol
```diff
@@ -76,15 +76,32 @@ library Math {
...
-    function min(uint256 a, uint256 b) internal pure returns (uint256) {
-        return a < b ? a : b;
+    function min(uint256 a, uint256 b) internal pure returns (uint256 result) {
+        return choice(a < b, a, b);
+    }
+
+    /**
+     * @dev If `condition` is true, returns `a`, otherwise returns `b`.
+     */
+    function choice(bool condition, uint256 a, uint256 b) internal pure returns (uint256) {
```
Amxx, line 93: I really like this function. 

I would rename that `ternary(bool, uint256, uint256)` though. @ernestognw, WDYT?

Lohann: Another suggestion is using the name `select`, this is the name used in [openssl constant time library](https://github.com/openssl/openssl/blob/openssl-3.2.0/include/internal/constant_time.h#L276-L281).

contracts/utils/math/Math.sol
```diff
@@ -114,7 +131,7 @@ library Math {
         // but the largest value we can obtain is type(uint256).max - 1, which happens
         // when a = type(uint256).max and b = 1.
         unchecked {
-            return a == 0 ? 0 : (a - 1) / b + 1;
+            return choice(a == 0, 0, (a - 1) / b + 1);
```
Amxx, line 134: looking at this I'm strating to have mixed feelings.

When you do a ternary choice, the compiler can "optimize" it so that only one branch is computer. However, when you use a dedicated function, values for both side must be computed. This remove the opportunity for lazy evaluation.

If both side are very simple math (or at least if the side that is often ignore is cheap enough) then that may make sens ... but if both branches are storage read, then using choice might be a very bad idea.

They may also be cases where the branch that is not needed cannot be computed (overflow, out of array access, ...)

I think we must evaluate gas costs of all the functions where we plan on using choice individually.

Lohann: That's why it should be used only for simple ternary operations, definitively no recommended for nested ternary operations for example, but in this case specifically it is still more optimized than the regular method, once one subtraction, addition and division are cheaper than one jump.

[SUB](https://www.evm.codes/#03?fork=cancun) and [ADD](https://www.evm.codes/#01?fork=cancun) cost `3 gas` units, [DIV](https://www.evm.codes/#04?fork=cancun) cost 5 gas units.

While a single [JUMPI](https://www.evm.codes/#01?fork=cancun) cost `10 gas units`, plus 1 [JUMPDEST](https://www.evm.codes/#5b?fork=cancun), plus 3 gas unit for pushing the PC pointer onto the stack.

Amxx: I'm with you on all this.

Amxx: That raises the issue of input sampling in (fuzzing) tests. I believe 0 and 1 and over represent compared to "real" use cases ... but again it's hard to figure out what the real use cases are.

Lohann: IMO by default should be assumed an uniform distribution, it means that no inputs are special, zero is not more likely to be used than `7014971116814175` for example:
```solidity
if (a == 0) return 0;
```
Is no better than doing this:
```solidity
if (a == 7014971116814175) return 83755424;
```
This means that optimizations and branching must be guided on reduce the average gas cost of all inputs.

Lohann: Actually here there's an even better way to optimize this code using branchless method, this method works for any `cond ? a : b` when `a == 0 || b == 0`:
```solidity
unchecked {
    return SafeCast.toUint(a > 0) * ((a - 1) / (b + 1));
}
```

Lohann: Done

This consumes `372 gas` when `a !=0` and  `324` when `a == 0`
```solidity
contract CeilDiv1 {
    function ceilDiv(uint256 a, uint256 b) external pure returns (uint256) {
        if (b == 0) {
            Panic.panic(Panic.DIVISION_BY_ZERO);
        }
        unchecked {
            return a == 0 ? 0 : (a - 1) / b + 1;
        }
    }
}
```

This consumes `355 gas` regardless the value of `a`.
```
contract BranchlessCeilDiv {
    function ceilDiv(uint256 a, uint256 b) external pure returns (uint256) {
        if (b == 0) {
            Panic.panic(Panic.DIVISION_BY_ZERO);
        }
        unchecked {
            return SafeCast.toUint(a > 0) * ((a - 1) / b + 1);
        }
    }
}
```

For simple code like this, this is how I compare the gas without fuzzing:
```python
total = 2 ** 256 // number of possible inputs

// a == 0
// gas cost multiplied by how many inputs can reach this branch.
branch1_weight = 324 * 1

// a != 0
branch2_weight = 374 * (total - 1)

avg_gas_cost = (branch1_weight + branch2_weight) / 2
```

Amxx: IMO by default should be assumed an uniform distribution

I would dissagree with that. I'm readdy to bet that, considering all the number manipualted onchain trough oracles, tokens, NFTs, ... there is significantly more between `0` and `2**255 -1` than between `2**255-1` and `2**256 - 1`.

Most practical values are probably < 10**36

Lohann: > Most practical values are probably < 10**36

If this is the case, and you know the distribution, just need plug the weight of the inputs, example.

Given all inputs `x`:
- `x == 0` weight is `1000` _(this basically means x==0 happens 100x more often than  `0 < x < 10**36`_
- `0 < x < 10**36` weight is `10`
- `x >= 10**36` weight is `1`

Then we do the same weighted gas calculation, with a more scientific way to decide which implementation is more efficient.
The formula is sum of all `input_count * weight * gas_cost` divided by the total
```python
input_1 = 1 # x == 0
input_2 = (2 ** 36) - input_0 # x > 0
input_3 = (2 ** 256) - input_1 - input_0 # x > (2 ** 36)

branch1_weight = input_0 * 1000
branch2_weight = (input_1 * 10) + (input_2 * 1)

total = branch1_weight + branch2_weight

branch1_gas_cost = branch1_weight * 304
branch2_gas_cost = branch2_weight * 350

avg_gas_cost = (branch1_gas_cost + branch2_gas_cost) / total
```
Using this formula, the final gas cost is:
`349` for `a == 0 ? 0 : (a - 1) / b + 1`
`333` for  `SafeCast.toUint(a != 0) * ((a - 1) / b + 1);`

Lohann: Also big values are more common when using decimal arithmetic, for example calculate sqrt with 10 decimal places precision:
```solidity
uint256 internal constant DECIMALS = 10; // precision in decimals

function sqrt(uint256 x) public pure returns (uint256 result) {
    x *= 10 ** (DECIMALS * 2);
    result = Math.sqrt(x);
}

function divCeil(uint256 x) public pure returns (uint256 result) {
    x *= 10 ** DECIMALS;
    result = Math.divCeil(x);
}
```

Lohann:
>When you do a ternary choice, the compiler can "optimize" it so that only one branch is computer. However, when you use a dedicated function, values for both side must be computed. This remove the opportunity for lazy evaluation.

I can make the `select` method private or use it only in simple ternary operations, if we are unsure about expose it or not.

Amxx:
I don't think it should be private. We want to be able to use it in different places/libraries without re-declaring it.

I think it comes down to documentation that it is not always cheaper than a ternary, and that both branches are always evaluate.

ernestognw:
Catching up with this. I'd like to highlight that @Lohann started this looking for constant cost in math operations. Although it's not the kind of priorities OpenZeppelin Contracts has, it's an honest requirement. I also agree with @Amxx with that we prefer more readable code.

I would consider creating an issue for branchless math if this requirement of constant costs ever comes back.

I'll be leaving comments while I read the conversation.

contracts/utils/math/Math.sol
ernestognw: I like both but I feel like `ternary` is more explicit than `select`.

Note that Solidity is also frequently used by backend developers who may be used to the meaning of `select` in SQL. My opinion is that `ternary` is better scoped to what it does and its arguments follow the same order as in a ternary (e.g. `a ? b : c` vs `ternary(a, b, c)`)

contracts/utils/math/Math.sol
ernestognw: > we don't even know the input distribution

This is the major source of uncertainty when deciding to implement gas optimizations. Most of the times we try to estimate what's better, and I agree that for most practical purposes is < 10 ** 36. This opinion is based exclusively on how popular 1e18 decimals are.

> Also big values are common when using decimal arithmetic

This is a very interesting observation. I tend to agree with this concern but I also have 0 evidence of a sqrt operation being widely used in production with big values, I'd appreciate references if you have.

Considering Uniswap's Universal Router is the #1 mainnet gas consumer, I tried searching for `sqrt` operations and only found sqrt for large values when [minting liquidity in V2](https://github.com/Uniswap/v2-core/blob/ee547b17853e71ed4e0101ccfd52e70d5acded58/contracts/UniswapV2Pair.sol#L95) and on the [TWAMM math of v4](https://github.com/Uniswap/v4-periphery/blob/6616b12db25257ffb3c562f131612ebb2fd89082/contracts/libraries/TWAMM/TwammMath.sol#L44C14-L44C32). I don't consider none of these cases to be a big deal.

Lohann: In this case specifically, regardless of the distribution doing, doing this:
```solidity
// Consumes 333 gas units
SafeCast.toUint(a > 0) * ((a - 1) / b + 1);
```
is cheaper, but maybe less clear, than doing this:
```solidity
// Consumes 302 gas when a == 0, and 350 when a != 0
a == 0 ? 0 : (a - 1) / b + 1;
```

Lohann: In this case specifically, regardless of the distribution, doing this:
```solidity
// Consumes 333 gas units
SafeCast.toUint(a > 0) * ((a - 1) / b + 1);
```
is cheaper, but maybe less clear, than doing this:
```solidity
// Consumes 302 gas when a == 0, and 350 when a != 0
a == 0 ? 0 : (a - 1) / b + 1;
```

ernestognw: Yes. Agree with this one specifically.

More generally, I don't think we have enough data to even estimate the total economic cost of this. The savings are 17 gas in this case (can be extrapolated to others) and will require ~10,000 operations to even cost 51 USD on mainnet (at 100 gwei and ETH trading at 3000).

ernestognw:
> When you do a ternary choice, the compiler can "optimize" it so that only one branch is computer. However, when you use a dedicated function, values for both side must be computed. This remove the opportunity for lazy evaluation.

Is this true? I see there's an open issue about it:

> When you do a ternary choice, the compiler can "optimize" it so that only one branch is computer. However, when you use a dedicated function, values for both side must be computed. This remove the opportunity for lazy evaluation.

Is this true? I see there's an open issue about it:
https://github.com/ethereum/solidity/issues/12930

ernestognw (commented):
I also have mixed feelings regarding compiler branchless optimization, but I think it's not implemented and, in general, Solidity developers complain about obvious optimizations not performed, so I see value in giving some control to the user.

contracts/utils/math/Math.sol
Lohann: Honestly we are only discussing this because most of those optimizations, and actually most of things that `Math.sol` does should be handled by Solidity, not by a library, because most optimization depends on the context where it is being used.
Example calculate the log2 of an `uint128` is cheaper than calculate of an `uint256`, but consider this:
```solidity
uint256 result = Math.log2(Math.sqrt(uint256(value)));
```
The sqrt of an `uint256` cannot return a value greater than `uint128`, sqrt of `uint128` returns an `uint64`, etc... but openzepellin doesn't know about that, but the compiler does.. that's why I think those mathematical function should be built-in the compiler, same way as overflow check, not by a library.

Lohann: As result this cost increases exponentially, so 13 gas unit here in a simple operation, become 100x in more complex formulas.

contracts/utils/math/Math.sol
Lohann: Ok, will change to `ternary` then.

Lohann: done

.changeset/spotty-falcons-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Replaces simple ternary operations `cond ? a : b` by a more efficient branchless `ternary` function, now supported by `Math.sol` and `SignedMath.sol`.
```
ernestognw, line 5:
```suggestion
`Math`, `SafeMath`: Replace simple ternary operations `cond ? a : b` by a more efficient branchless `ternary` function.
```

.changeset/spotty-falcons-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Math`, `SafeMath`: Replace simple ternary operations `cond ? a : b` by a more efficient branchless `ternary` function.
```
ernestognw, line 5: Just came to the conclusion that the optimization itself is not a substantial change. We should focus the changelog entry on the new `ternary` function and the reasons why someone would use it.

```suggestion
`Math`, `SignedMath`: Add a branchless `ternary` function that computes`cond ? a : b` in constant gas costs.
```

ernestognw (commented):
Small nit with the changeset

ernestognw (approved):
I tried looking for ternary branchless optimization in Solidity without luck, so I'd expect such optimization to not exist. Regardless, we can't ensure it'll be more efficient in the future, so I focused the changeset on the new `ternary` function and its "constant gas" property.

For the controversial part of `Math.sol`, the discussion will continue back and forth without strong evidence of the distribution in applications out there, so I wouldn't focus on it if the result is more-less 20 gas units.

LGTM, but let's wait on @Amxx since the name change to `ternary` wasn't agreed with him.

Amxx (approved):

PR #4992, opened 2024-04-04
Add Packing library

Amxx:
Will be usefull for processing "packed" values such as `accountGasLimits` and `gasFees` as defined in [ERC-4337](https://eips.ethereum.org/EIPS/eip-4337#entrypoint-definition)

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/heavy-baboons-give.md
+++ b/.changeset/heavy-baboons-give.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Packing`: Added a new utility for packing and unpacking multiple values into a single bytes32.
--- a/.changeset/heavy-baboons-give.md
+++ b/.changeset/heavy-baboons-give.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-`Packing`: Added a new utility for packing and unpacking multiple values into a single bytes32.
+`Packing`: Added a new utility for packing and unpacking multiple values into a single bytes32. Includes initial support for packing two `uint128` in an `Uint128x2` type.
--- a/contracts/utils/Packing.sol
+++ b/contracts/utils/Packing.sol
@@ -19,22 +19,22 @@ library Packing {
     }
 
     /// @dev Pack two uint128 into a Uint128x2
-    function pack(uint128 high128, uint128 low128) internal pure returns (Uint128x2) {
-        return Uint128x2.wrap(bytes32(bytes16(high128)) | bytes32(uint256(low128)));
+    function pack(uint128 first128, uint128 second128) internal pure returns (Uint128x2) {
+        return Uint128x2.wrap(bytes32(bytes16(first128)) | bytes32(uint256(second128)));
     }
 
     /// @dev Split a Uint128x2 into two uint128
     function split(Uint128x2 self) internal pure returns (uint128, uint128) {
-        return (high(self), low(self));
+        return (first(self), second(self));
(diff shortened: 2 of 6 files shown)


contracts/utils/Packing.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Helper library packing and unpacking multiple values into bytes32
+ */
+library Packing {
+    type Uint128x2 is bytes32;
```
Amxx, line 9: For now we only have `(uint128, uint128) <> bytes32`, but we can imagine having other types, such as `uint64x4`

Everything here is designed to scale to oither types.

test/utils/Packing.t.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Test} from "forge-std/Test.sol";
+import {Packing} from "@openzeppelin/contracts/utils/Packing.sol";
+
+contract Base64Test is Test {
```
Amxx, line 8:
```suggestion
  contract PackingTest is Test {
```

ernestognw (commented):
Neat! I left only one comment regarding the `high` and `low` functions in the context of other types. Also, how would we name other types like `uint64|uint64|uint128`?

contracts/utils/Packing.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    /// @dev Get the first element of a Uint128x2 (high part)
+    function high(Uint128x2 self) internal pure returns (uint128) {
+        return uint128(bytes16(Uint128x2.unwrap(self)));
+    }
+
+    /// @dev Get the second element of a Uint128x2 (low part)
+    function low(Uint128x2 self) internal pure returns (uint128) {
+        return uint128(uint256(Uint128x2.unwrap(self)));
+    }
```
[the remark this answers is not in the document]
Amxx: There are places where doing a full split, and keeping only part of the result is a bit more expensive, and also less clean in terms of code.

I would keep these two function, but we can rename them.

I considered `first()` and `second()`, which may scale better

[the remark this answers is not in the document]
ernestognw: Right, also I realized it would be easy to reach a `Stack too deep` error if we pack too many values, so I'd be in favor of having single getters.

Agree with `first` and `second`, but the distinction with `high` and `low` is that the former doesn't express the order (i.e. first ltr or rtl?). Shall we make it explicit? Maybe by exposing `first()` and `firstRtl()`

ernestognw:
> Also, how would we name other types like uint64|uint64|uint128?

Just realized that if we have an `Uint64x2`, we can pack 4 of those in a `Uint128x2` . Similar for every other combination down to the boolean. Great!

.changeset/heavy-baboons-give.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Packing`: Added a new utility for packing and unpacking multiple values into a single bytes32.
```
ernestognw, line 5:
```suggestion
`Packing`: Added a new utility for packing and unpacking multiple values into a single bytes32. Includes initial support for packing two `uint128` in an `Uint128x2` type.
```

ernestognw (approved):

PR #5004, opened 2024-04-08
Fix broken references in doc-site

### LINKED ISSUE #5003: Fix API documentation broken links in the doc-site
There are a few places where the autogenerated link is either broken or miss-formatted:

I noticed three main cases:
1. The scoped module is missing (ex: `{renounceRole}` instead of `{AccessControl-renounceRole}`).
2. Extra pair of brackets (ex: `{{renounceRole}}` instead of `{renounceRole}`)
3. References to private functions that are not automatically documented, becoming broken references in the doc-site (ex: `{_getAdminRestrictions}` in AccessManager).

The first two should be directly fixed, or a rule can be created to check them (asciidoctor detects the broken references for ex).

For the latter, I think a design to avoid private function references in the doc-site would be helpful (something as simple as using `privateFunctionName` instead of {privateFunctionName} could help.

EDIT: [Here is a link](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20#ERC20Votes) to some broken references in the doc-site (third paragraph).

ericnordelo:
Partially fixes #5003

### DIFF
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -110,6 +110,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
 
     /**
      * @dev Check that the caller is authorized to perform the operation.
+     * See {AccessManager} description for a detailed breakdown of the authorization logic.
      */
     modifier onlyAuthorized() {
         _checkAuthorized();
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -470,7 +470,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
     /**
      * @dev Reverts if the operation is currently scheduled and has not expired.
      *
-     * Note: This function was introduced due to stack too deep errors in schedule.
+     * NOTE: This function was introduced due to stack too deep errors in schedule.
      */
     function _checkNotScheduled(bytes32 operationId) private view {
         uint48 prevTimepoint = _schedules[operationId].timepoint;
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -55,8 +55,8 @@ import {Time} from "../../utils/types/Time.sol";
  * will be {AccessManager} itself.
  *
  * WARNING: When granting permissions over an {Ownable} or {AccessControl} contract to an {AccessManager}, be very
- * mindful of the danger associated with functions such as {{Ownable-renounceOwnership}} or
- * {{AccessControl-renounceRole}}.
+ * mindful of the danger associated with functions such as {Ownable-renounceOwnership} or
+ * {AccessControl-renounceRole}.
  */
 contract AccessManager is Context, Multicall, IAccessManager {
     using Time for *;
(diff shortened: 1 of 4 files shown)


contracts/access/IAccessControl.sol
```diff
@@ -31,7 +31,7 @@ interface IAccessControl {
      * @dev Emitted when `account` is granted `role`.
      *
      * `sender` is the account that originated the contract call, an admin role
-     * bearer except when using {AccessControl-_setupRole}.
+     * bearer except when using {AccessControl-_grantRole}.
```
ericnordelo, line 34: Outdated reference.

contracts/access/manager/AccessManager.sol
```diff
@@ -109,8 +109,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
     bytes32 private _executionId;
 
     /**
-     * @dev Check that the caller is authorized to perform the operation, following the restrictions encoded in
-     * {_getAdminRestrictions}.
+     * @dev Check that the caller is authorized to perform the operation.
```
ericnordelo, line 112: Maybe the comment can be kept but we can change brackets for `` ? Private references should be treated differently.

contracts/access/manager/AccessManager.sol
```diff
@@ -470,7 +469,8 @@ contract AccessManager is Context, Multicall, IAccessManager {
 
     /**
      * @dev Reverts if the operation is currently scheduled and has not expired.
-     * (Note: This function was introduced due to stack too deep errors in schedule.)
+     *
+     * NOTE: This function was introduced due to stack too deep errors in schedule.
```
ericnordelo, line 473: For conistency

contracts/access/IAccessControl.sol
ernestognw: This was already fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4924, though we didn't cherry pick it to the releaase branch. Let's merge this and then cherry pick to the release branch so the documentation gets updated accordingly.

contracts/access/manager/AccessManager.sol
ernestognw: It's true private references should be treated differently, although in this case I think a reference to where the rules are defined is more important than the reference itself.

Let's do:

```suggestion
     * @dev Check that the caller is authorized to perform the operation.
     * See {AccessManager} description for a detailed breakdown of the authorization logic.
```

contracts/access/manager/IAccessManager.sol
```diff
@@ -107,9 +107,6 @@ interface IAccessManager {
      * NOTE: The IAuthority interface does not include the `uint32` delay. This is an extension of that interface that
      * is backward compatible. Some contracts may thus ignore the second return argument. In that case they will fail
      * to identify the indirect workflow, and will consider calls that require a delay to be forbidden.
-     *
-     * NOTE: This function does not report the permissions of this manager itself. These are defined by the
-     * {_canCallSelf} function instead.
```
ernestognw, line 112: Similarly, I think it's valuable to reference AccessManager's own permissions rather than the private function. Let's do:

```suggestion

     * NOTE: This function does not report the permissions of this manager itself. These are defined by the
     * {AccessManager} documentation.
```

ernestognw (approved):
Thanks! ✨

Amxx (approved):

PR #5008, opened 2024-04-08
Expose an initializer function for non-abstract contracts in their upgradeable versions

### LINKED ISSUE #5002: Mark AccessManagerUpgradeable abstract or implement initializer function
**🧐 Motivation**
I want to use AccessManagerUpgradeable the same way I can use AccessManager contract.

**📝 Details**
It is impossible to use AccessManagerUpgradeable as is right know. It lacks initializer function, so it is not possible to deploy it correctly. And also it is not marked abstract to force implement initializer function in sub-contract.

ernestognw:
Fixes #5002 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/thin-walls-drop.md
+++ b/.changeset/thin-walls-drop.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`AccessManager`, `VestingWallet`, `TimelockController` and `ERC2771Forwarder`: Added an `initializer` function in their corresponding upgradeable variants.
--- a/.changeset/thin-walls-drop.md
+++ b/.changeset/thin-walls-drop.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-`AccessManager`, `VestingWallet`, `TimelockController` and `ERC2771Forwarder`: Added an `initializer` function in their corresponding upgradeable variants.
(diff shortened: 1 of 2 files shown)


scripts/upgradeable/transpile.sh
```diff
@@ -39,6 +39,10 @@ npx @openzeppelin/upgrade-safe-transpiler -D \
   -x '!contracts/proxy/utils/UUPSUpgradeable.sol' \
   -x '!contracts/proxy/beacon/IBeacon.sol' \
   -p 'contracts/**/presets/**/*' \
```
Amxx, line 41: We should remove that line since we don't have any presets any more, right?
```suggestion
```

ernestognw: Yeah, that's right!

Amxx (dismissed):
LGTM

.changeset/thin-walls-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`AccessManager`, `VestingWallet`, `TimelockController` and `ERC2771Forwarder`: Added an `initializer` function in their corresponding upgradeable variants.
```
Amxx, line 5:
```suggestion
`AccessManager`, `VestingWallet`, `TimelockController` and `ERC2771Forwarder`: Added a public `initializer` function in their corresponding upgradeable variants.
```

Amxx (approved):

PR #5014, opened 2024-04-10
Extend `onlyAuthorized` to support extra functions in AccessManager

ernestognw:
This PR adds the ability to extend the AccessManager and use the `onlyAuthorized` modifer for those functions as well.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/light-news-listen.md
+++ b/.changeset/light-news-listen.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`AccessManager`: Allow the `onlyAuthorize` modifier to restrict functions added to the manager.
--- a/.changeset/light-news-listen.md
+++ b/.changeset/light-news-listen.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-`AccessManager`: Allow the `onlyAuthorize` modifier to restrict functions added to the manager.
+`AccessManager`: Allow the `onlyAuthorized` modifier to restrict functions added to the manager.
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -586,14 +586,15 @@ contract AccessManager is Context, Multicall, IAccessManager {
 
     // ================================================= ADMIN LOGIC ==================================================
     /**
-     * @dev Check if the current call is authorized according to admin logic.
+     * @dev Check if the current call is authorized according to admin and roles logic.
+     *
+     * WARNING: Carefully review the considerations of {AccessManaged-restricted} since they apply to this modifier.
      */
     function _checkAuthorized() private {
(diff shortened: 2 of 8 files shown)


.changeset/light-news-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`AccessManager`: Allow the `onlyAuthorize` modifier to restrict functions added to the manager.
```
ernestognw, line 5:
```suggestion
`AccessManager`: Allow the `onlyAuthorized` modifier to restrict functions added to the manager.
```

test/access/manager/AccessManager.test.js
```diff
@@ -1670,6 +1671,60 @@ describe('AccessManager', function () {
     });
   });
 
+  describe('access managed self operations', function () {
+    describe('when calling a restricted target function', function () {
+      beforeEach('set required role', function () {
+        this.method = this.target.fnRestricted.getFragment();
+        this.role = { id: 785913n };
+        this.manager.$_setTargetFunctionRole(this.target, this.method.selector, this.role.id);
```
Amxx, line 1679: Should we set the function role for `this.manager, this.method.selector` ? It feels like we are testing restricted operation on the target when we should test the restricted operation on the manager.

contracts/access/manager/AccessManager.sol
```diff
@@ -611,7 +613,7 @@ contract AccessManager is Context, Multicall, IAccessManager {
      */
     function _getAdminRestrictions(
         bytes calldata data
-    ) private view returns (bool restricted, uint64 roleAdminId, uint32 executionDelay) {
+    ) private view returns (bool adminRestricted, uint64 roleAdminId, uint32 executionDelay) {
         if (data.length < 4) {
             return (false, 0, 0);
```
Amxx, line 618: maybe we want to do
`return (false, getTargetFunctionRole(address(this), bytes4(0)), 0);`
so that selector 0 can be used to set restriction on the `receive()` function

Amxx:
```suggestion
            return (false, getTargetFunctionRole(address(this), bytes4(0)), 0);
```

Amxx: We need to decide on this, but otherwize the PR is good to me.

ernestognw:
We were loosing coverage for this section:

```solidity
// isTragetClosed apply to non-admin-restricted function
if (!adminRestricted && isTargetClosed(address(this))) {
    return (false, 0);
}
```

Turns out closing the manager is not allowed. We decided to allow closing the manager so that the flag applies only to non-admin functions (i.e. functions added to the manager). This way the contract can close access to itself without getting it locked.

Amxx (approved):

PR #5020, opened 2024-04-18
Update delegation method to perform additional check

dimitriospapathanasiou:
Adding a check to exit the function early if the delegatee is the same as the current delegate. This prevents emitting events and executing the vote movement logic unnecessarily when the delegate doesn't change, aiming to solve potential gas costs, especially in scenarios where the delegate is frequently updated.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/weak-wolves-turn.md
+++ b/.changeset/weak-wolves-turn.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Add check to '\_delegate' to exit early if the delegate does not change.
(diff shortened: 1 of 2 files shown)


ernestognw:
Hi @dimitriospapathanasiou,

This is saving gas for cases that can be avoided before sending the transaction at the expense of adding extra gas for cases where the new delegate is different.

I don't think this check is reasonable unless there's evidence of a high volume of users constantly redelegating to the same address.

ernestognw:
Got it. I agree there may be honest cases where the check is helpful but I don't think we should consider merging this until there's data proving is saving more than wasting (I don't expect so).

I'm closing, but thanks for raising!

PR #5021, opened 2024-04-19
Update certora-cli to 7.3.0

ernestognw:
Replaces #4957 and #5019

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/certora/specs/AccessManaged.spec
+++ b/certora/specs/AccessManaged.spec
@@ -7,6 +7,9 @@ methods {
     function authority_canCall_immediate(address) external returns (bool);
     function authority_canCall_delay(address)     external returns (uint32);
     function authority_getSchedule(address)       external returns (uint48);
+
+    // Summaries
+    function _.setAuthority(address)               external => DISPATCHER(true);
 }
 
 invariant isConsumingScheduledOpClean()
--- a/certora/specs/AccessManaged.spec
+++ b/certora/specs/AccessManaged.spec
@@ -7,9 +7,10 @@ methods {
     function authority_canCall_immediate(address) external returns (bool);
     function authority_canCall_delay(address)     external returns (uint32);
     function authority_getSchedule(address)       external returns (uint48);
+    function _hasCode(address)                    external returns (bool) envfree;
 
     // Summaries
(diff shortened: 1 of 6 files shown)


ernestognw:
The CI uses 0.8.24. When I compile locally with 0.8.25 everything works fine but switching to 0.8.24 produce the same results as the CI.

I'll keep investigating

certora/specs/AccessManaged.spec
```diff
@@ -32,3 +36,24 @@ rule callRestrictedFunction(env e) {
...
+
+    address previousAuthority = authority();
+
+    setAuthority@withrevert(e, newAuthority);
+    bool success = !lastReverted;
+
+    assert (success && authority() == newAuthority) <=> (
+        previousAuthority == e.msg.sender &&
+        _hasCode(newAuthority)
+    );
```
Amxx, line 58: I would do it like this:

```suggestion
    assert success <=> (
        previousAuthority == e.msg.sender &&
        _hasCode(newAuthority)
    );
    assert success => newAuthority == authority();
```

Reason: "success && authority() == newAuthority" implies that there is a possibility that both values are different. Like the call did not revert, but the authority was not set...

Amxx:
Note: ERC721 safeMint and safeTransferFrom rules fail because failure of the underlying `checkOnERC721Received` is not modeled in the `assert success <=> (...);` description on liveness.

certora/specs/AccessManaged.spec
ernestognw: Right, it's not possible to make a successful call and having the authority not set. I was thinking the prover will fail for the case when the new authority is the same as the previous one.

I'm fine with this one.

Amxx:
I'm not sure why that happens, but clearly there is an issue here

See [this](https://prover.certora.com/output/857050/ed327a47bcef450e8a8ef7e27492f6f5?anonymousKey=c35c617098bdfed45eb6160afd16ded0c0d30176).

Capture d’écran du 2024-04-25 10-08-45

Capture d’écran du 2024-04-25 10-10-39

PR #5029, opened 2024-04-29
Support Saturation Arithmetic Operations

Lohann:
# Description
Adds gas efficient [saturating arithmetic](https://en.wikipedia.org/wiki/Saturation_arithmetic) operators, those operators are useful when both overflow and revert are not desired.
This PR also optimizes and remove unnecessary branching from various Math methods.

# Motivation
If you want to write some formula, but don't want neither wrapping and reverts, the only option is using the `Math.try*` methods, which can't be used in chain and make the code less readable.

Ex: Make sure there're at least `100_000` gas units left before calling an external contract, otherwise revert.
```solidity
// “all but one 64th", reference: https://eips.ethereum.org/EIPS/eip-150
uint256 gasAvailable = gasleft().saturatingSub(5000).saturatingMul(63) / 64;
require(gasAvailable >= 100_000, "not enough gas left to call contract");
// ... call contract
```

The equivalent code using `Math.try*` is less readable and less efficient.
```solidity
(bool success, uint256 gasAvailable) = gasleft().trySub(5000);
gasAvailable = Math.ternary(success, gasAvailable, 0);

// “all but one 64th", reference: https://eips.ethereum.org/EIPS/eip-150
(success, gasAvailable) = gasAvailable.tryMul(63);
gasAvailable = Math.ternary(success, gasAvailable, type(uint256).max);
gasAvailable /= 64;

require(gasAvailable >= 100_000, "not enough gas left to call contract");
// ... call contract
```


contracts/utils/math/Math.sol
```diff
@@ -17,14 +17,36 @@ library Math {
...
+        }
+    }
+
+    /**
+     * @dev Unsigned saturating subtraction, bounds to zero instead of overflowing.
+     */
+    function saturatingSub(uint256 a, uint256 b) internal pure returns (uint256) {
+        unchecked {
+            // equivalent to: a >= b ? a - b : 0
+            return (a - b) * SafeCast.toUint(a > b);
```
Amxx, line 49: We could also do.
```suggestion
            return ternary(a > b, a - b, 0);
```

Lohann: Otherwise I can refactor it for use the `ternary` at a little gas cost, but still better than branching.

contracts/utils/math/Math.sol
```diff
@@ -33,8 +55,25 @@ library Math {
...
+            success = a >= b;
+            // equivalent to: success ? (a - b) : 0
+            result = SafeCast.toUint(success) * (a - b);
+        }
+    }
+
+    /**
+     * @dev Unsigned saturating multiplication, bounds to `2 ** 256 - 1` instead of overflowing.
+     */
+    function saturatingMul(uint256 a, uint256 b) internal pure returns (uint256) {
```
Amxx, line 67: use of assembly may be efficient, but readability is not great. I'd be more confortable with that function being

```solidity
(bool success, uint256 result) = tryMul(a, b);
return ternary(success, result, type(uint256).max);
```

Lohann: Yeah I had to use assembly because `c / a` reverts, even inside a `unchecked` statement, while in pure evm a division by zero always result in zero.

contracts/utils/math/Math.sol
```diff
@@ -17,14 +17,36 @@ library Math {
...
     }
 
+    /**
+     * @dev Unsigned saturating addition, bounds to `2 ** 256 - 1` instead of overflowing.
+     */
+    function saturatingAdd(uint256 a, uint256 b) internal pure returns (uint256) {
+        unchecked {
+            uint256 c = a + b;
+            // equivalent to: c < a ? type(uint256).max : c
+            return c | (0 - SafeCast.toUint(c < a));
```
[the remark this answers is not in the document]
Lohann: Yeah the way I designed all methods in this PR was by first write everything in pure assembly using https://www.evm.codes/playground to find the most efficient solution possible, then I write the solidity code that generate the equivalent code. For this one specifically this is the most optimized way I've found:
```
PUSH2 0xdead  // b
PUSH2 0xbeef  // a b
DUP2          // b a b
ADD           // c b      where c = a + b        
SWAP1         // b c
DUP2          // c b c
LT            // d c      where d is 1 when an overflow occur and 0 when doesn't
PUSH0         // 0 d c
SUB           // e c      where e is either zero or `type(uint256).max` if an overflow occur
OR            // r     

// total gas used starting from DUP2 = 23 gas
```

contracts/utils/math/SignedMath.sol
```diff
@@ -65,4 +65,74 @@ library SignedMath {
...
+            // | false       | true    | Ok
+            bool sign = b >= 0;
+            bool overflow = a > c == sign;
+
+            // Efficient branchless method to retrieve the boundary limit:
+            // (1 << 255)     == type(int256).min
+            // (1 << 255) - 1 == type(int256).max
+            uint256 limit = (SafeCast.toUint(overflow) << 255) - SafeCast.toUint(sign);
+
+            return ternary(overflow, int256(limit), c);
```
Lohann, line 100: For this one, this is the assembly implementation, it uses `64` gas.
```
// b = type(int256).min
PUSH32 0x8000000000000000000000000000000000000000000000000000000000000000
// a = -1
PUSH32 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff

// c = a + b
DUP2 // b a b
DUP2 // a b a b
ADD  // c a b

// sign = b >= 0
SWAP2       // b a c
PUSH0       // 0 b a c
SGT         // (0 > b) a c
ISZERO      // sign a c

// overflow = (c < a) == sign
SWAP1       // a sign c
DUP3        // c a sign c
SLT         // (c < a) sign c
DUP2        // sign (c > a) sign c
EQ          // overflow sign c

// limit = (overflow << 255) - sign
SWAP1       // sign overflow c
DUP2        // overflow sign overflow c
PUSH1 0xff  // 255 overflow sign overflow c
SHL         // (overflow << 255) sign overflow c
SUB         // limit overflow c

// result = overflow ? limit : c
DUP3        // c limit overflow c
XOR         // (c ^ limit) overflow c
MUL         // ((c ^ limit) * overflow) c
XOR         // result
```

contracts/utils/math/Math.sol
Lohann: Btw I opened an issue in the solidity compiler, division should not revert in unchecked blocks:
https://github.com/ethereum/solidity/issues/15200

contracts/utils/math/Math.sol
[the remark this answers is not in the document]
Amxx: For the record, the ternary option would be
```suggestion
            return (a < c).ternary(c, type(uint256).max);
```

contracts/utils/math/Math.sol
```diff
@@ -17,14 +17,36 @@ library Math {
...
+        }
+    }
+
+    /**
+     * @dev Unsigned saturating subtraction, bounds to zero instead of overflowing.
+     */
+    function saturatingSub(uint256 a, uint256 b) internal pure returns (uint256) {
+        unchecked {
+            // equivalent to: a > b ? a - b : 0
+            return (a - b) * SafeCast.toUint(a > b);
```
Amxx, line 49:
```suggestion
            return SafeCast.toUint(b < a) * (a - b);
```

Amxx, line 49:
```suggestion
            // equivalent to: a > b ? a - b : 0
            return SafeCast.toUint(a > b) * (a - b);
```

Amxx:
Signed part of this PR is merged as part of #5527. The unsigned part (more complex) is now planned for 5.4

Amxx:
Unsigned implemented in #5527

We'll need a new PR for the signed part

Lohann:
Hi @Amxx, sorry for the late reply, awesome I'll open an PR including the signed part.

PR #5034, opened 2024-05-08
Add Halmos support

ernestognw:
So far we've been maintaining our Formal Verification workflows with Certora. Though, Halmos has got lots of improvements lately. This PR is to test it out in our current fuzz tests. Note that our rule of thumb so far has been to formally verify when possible and fuzz otherwise, so it's expected that some of theses tests won't be working. 

Currently, we added a decent amount of extra Formal Verification tests that are working good and really fast within our regular checks pipeline.

An important note from the Halmos team is that they don't recommend removing fuzz tests even if they're already formally verified. This is because the bounded model in which both Certora and Halmos only test for fixed size arrays.

#### PR Checklist

- [x] Tests
- [ ] Documentation


ernestognw:
The approach of this PR was to formally verify those cases when there's a fuzz test, some of these obviously didn't work given they were also difficult to formally verify with Certora in the first place.

These are my notes about the tests that didn't work so were skipped:

## Base64
The algorithm is a loop. Although attempted, tests failed with something similar to:

> Encountered symbolic memory offset: 289 + Concat(0, Extract(63, 58, halmos_EncodeInput_bytes_01))

## ShortStrings

In the case of testing if `toShortString` reverted with long strings, there seems to be an edge case with the selector of some function (perhaps when calling `this.shortString`). Couldn't figure out what's going on so I decided to move forward. For reference, this is what I got

>Encountered Unsupported cheat code: calldata = Concat(0xf28dceb300000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000144305a27a900000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000100, halmos_RevertLongInput_string_01(), 0x00000000000000000000000000000000000000000000000000000000)

## Governor

Doesn't make sense to formally verify for the description with proposer since we need to increase the loop bound and the string length less than 52 (which is 12 loops given 52 - 40) will always return true. Using a bigger string makes the test timeout.

## Checkpoints

Depends heavily on loops. Some variables might be worth making symbolic but not moving into that yet imo.

## SignedMath

### Average

>Encountered Unsupported cheat code: calldata = Concat(0xa322c40e, Extract(+(0x4000000000000000000000000000000000000000000000000000000000000000, *(0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff, Concat(0x00, +(0x4000000000000000000000000000000000000000000000000000000000000001, ~(Extract(p_a_int256()))))))), +(0xff, *(0xff, ~(Extract(p_a_int256())))))

## Math

### CeilDiv / Sqrt / MulDiv

> WARNING:Halmos:Counterexample: unknown
(see https://github.com/a16z/halmos/wiki/warnings#counterexample-unknown)
[TIMEOUT] check_CeilDiv(uint256,uint256) (paths: 7, time: 1.44s, bounds: [])

### InvMod

Depends heavily on loops

### Log2 / Log10 / Log256

Log2 works locally but times out in the CI. I don't think it makes sense to keep them formally verified.

>WARNING:Halmos:Counterexample (potentially invalid):
    p_input_uint256 = 0x00000290967f152f6cc310785e8e806ebcb3a077e59a091c0000000000001005 (17701574136214187778346309088771581813280962368749720433162010073501701)
    p_r_uint8 = 0x0000000000000000000000000000000000000000000000000000000000000000 (0)
(see https://github.com/a16z/halmos/wiki/warnings#counterexample-invalid)

### MulDivDomain

>WARNING:Halmos:Encountered Unsupported cheat code: calldata = 0xf28dceb3000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000244e487b71000000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)

### Modexp

>[ERROR] check_ModExp(uint256,uint256,uint256) (paths: 2, time: 0.03s, bounds: [])
WARNING:Halmos:Encountered Unsupported cheat code: calldata = 0xf28dceb3000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000244e487b71000000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000

### TryModexp

>[ERROR] check_TryModExp(uint256,uint256,uint256) (paths: 2, time: 0.02s, bounds: [])
WARNING:Halmos:Encountered Unknown contract call: to = 0x0000000000000000000000000000000000000005; calldata = Concat(0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020, p_b_uint256(), p_e_uint256(), p_m_uint256()); callvalue = 0x0000000000000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)

### ModExpMemory

>Running 1 tests for test/utils/math/Math.t.sol:MathTest
[ERROR] check_ModExpMemory(uint256,uint256,uint256) (paths: 4, time: 0.07s, bounds: [])
WARNING:Halmos:Encountered Unsupported cheat code: calldata = 0xf28dceb3000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000244e487b71000000000000000000000000000000000000000000000000000000000000001200000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:Halmos:Encountered Unknown contract call: to = 0x0000000000000000000000000000000000000005; calldata = Concat(0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020, p_b_uint256(), p_e_uint256(), p_m_uint256()); callvalue = 0x0000000000000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:Halmos:Encountered Unknown contract call: to = 0x0000000000000000000000000000000000000005; calldata = Concat(0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020, p_b_uint256(), p_e_uint256(), p_m_uint256()); callvalue = 0x0000000000000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:Halmos:Encountered Unknown contract call: to = 0x0000000000000000000000000000000000000005; calldata = Concat(0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020, p_b_uint256(), p_e_uint256(), p_m_uint256()); callvalue = 0x0000000000000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:Halmos:check_ModExpMemory(uint256,uint256,uint256): paths have not been fully explored due to the loop unrolling bound: 2
(see https://github.com/a16z/halmos/wiki/warnings#loop-bound)
Symbolic test result: 0 passed; 1 failed; time: 0.09s

### TryModExpMemory

> Running 1 tests for test/utils/math/Math.t.sol:MathTest
[ERROR] check_TryModExpMemory(uint256,uint256,uint256) (paths: 3, time: 0.06s, bounds: [])
WARNING:Halmos:Encountered Unknown contract call: to = 0x0000000000000000000000000000000000000005; calldata = Concat(0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020, p_b_uint256(), p_e_uint256(), p_m_uint256()); callvalue = 0x0000000000000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:Halmos:Encountered Unknown contract call: to = 0x0000000000000000000000000000000000000005; calldata = Concat(0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020, p_b_uint256(), p_e_uint256(), p_m_uint256()); callvalue = 0x0000000000000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:Halmos:Encountered Unknown contract call: to = 0x0000000000000000000000000000000000000005; calldata = Concat(0x000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000020, p_b_uint256(), p_e_uint256(), p_m_uint256()); callvalue = 0x0000000000000000000000000000000000000000000000000000000000000000
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)
WARNING:Halmos:check_TryModExpMemory(uint256,uint256,uint256): paths have not been fully explored due to the loop unrolling bound: 2
(see https://github.com/a16z/halmos/wiki/warnings#loop-bound)
Symbolic test result: 0 passed; 1 failed; time: 0.07s

## ERC2771Forwarder

### ExecuteAvoidsETHStuck
>[ERROR] check_ExecuteAvoidsETHStuck(uint256,uint256,bool) (paths: 163, time: 13.57s, bounds: [])
WARNING:Halmos:Encountered Unsupported cheat code: calldata = 0xf4844814
(see https://github.com/a16z/halmos/wiki/warnings#internal-error)

The batch version is a loop.

## ERC712Consecutive

Loops

ernestognw:
idk why netlify preview is failing, I asked for access but I think it's ready for review in the meantime @Amxx

fv-requirements.txt
Amxx, line 1: Any reason to rename from `requirements.txt` (standard in python) to `fv-requirements.txt` ?

.github/workflows/formal-verification.yml
```diff
@@ -66,3 +67,20 @@ jobs:
...
           CERTORAKEY: ${{ secrets.CERTORAKEY }}
+
+  halmos:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up environment
+        uses: ./.github/actions/setup
+      - name: Install python
+        uses: actions/setup-python@v4
```
Amxx, line 78: Certora uses v5, mlaybe we should use the same version everywhere.
```suggestion
        uses: actions/setup-python@v5
```

.github/workflows/formal-verification.yml
```diff
@@ -66,3 +67,20 @@ jobs:
...
+        with:
+          python-version: ${{ env.PIP_VERSION }}
+          cache: 'pip'
+          cache-dependency-path: 'fv-requirements.txt'
+      - name: Install python packages
+        run: pip install -r fv-requirements.txt
+      - name: Run Halmos
+        run: halmos --match-test '^check_|^testSymbolic' -vv
```
Amxx, line 86: The `check_` pattern requires adding `solhint-disable func-name-mixedcase` in many test files. Why not use a different prefix. I'm proposing:
```suggestion
        run: halmos --match-test '^checkSymbolic|^testSymbolic' -vv
```

Amxx: Another, potentially better, option would be to have:
- `testXxxxxx` for foundry fuzzing test
- `symbolicXxxxxx` for halmos

So that there is no overlap, and everything is clear.

test/utils/Arrays.t.sol
```diff
@@ -2,12 +2,30 @@
...
 
-contract ArraysTest is Test {
+contract ArraysTest is Test, SymTest {
     function testSort(uint256[] memory values) public {
         Arrays.sort(values);
+        _assertSort(values);
+    }
+
+    function check_Sort() public {
+        uint256[] memory values = new uint256[](3);
```
Amxx, line 18: Any reason to use 3 ? If possible we should use something bigger (ideally an arbitrary value - within some bounds)

test/utils/SlotDerivation.t.sol
```diff
@@ -3,18 +3,20 @@
...
     using SlotDerivation for bytes32;
 
     bytes[] private _array;
 
-    function testDeriveArray(uint256 length, uint256 offset) public {
-        length = bound(length, 1, type(uint256).max);
-        offset = bound(offset, 0, length - 1);
+    function testSymbolicDeriveArray(uint256 length, uint256 offset) public {
+        vm.assume(length > 0);
+        vm.assume(offset < length);
```
Amxx, line 19: Using `vm.assume` drops fuzzing branches, and thus reduces the size of the fuzzing "coverage". `bound` is the recommanded option when. Any good reason for that change?

Amxx: I see that Halmos recommend using vm.assume. 

It may be better to have 2 version, one with vm.assume for halmos, and one with bound for fuzzing

fv-requirements.txt
ernestognw: Yes, the reason is that otherwise Netlify will try to install Python dependencies but the newest version it supports is 3.8. The current halmos version is incompatible with Python 3.8.

This is the only way I could of tricking Netlify to ignore the dependencies.

test/utils/Arrays.t.sol
ernestognw: Unfortuanately 3 is the largest value that didn't timeout. The sorting algorithm's path explosion after 4 seems to be too much already, and make sense.

I would even consider removing this check whatsoever.

test/utils/SlotDerivation.t.sol
ernestognw: Yeah good observation, I splitted the test in two so that the fuzzing version keeps `bound`

Amxx:
Looks mostly good. Before merging I'd like to discuss the prefixes. Currently we have:
- `testXxxx`
- `testSymbolicXxxx`
- `checkSymbolicXxxx`

I'd like use to consider alternatives, maybe
- `testXxxx`
- `testSymbolicXxxx`
- `symbolicXxxx`

Lets discuss that on today's call.

ernestognw:
> I'd like use to consider alternatives, maybe
> 
> * `testXxxx`
> * `testSymbolicXxxx`
> * `symbolicXxxx`

I'd be fine with `symbolic` though. Just proposed `checkSymbolic` as it was my best alternative lol
Will update

Amxx (approved):

PR #5035, opened 2024-05-09
Add 512bits add and mult operations

### LINKED ISSUE #5433: In the Math library, add support for computing `x * y >> n`
**🧐 Motivation**
A function which calculates `x* y >> n`, allowing for the intermediate value of `x * y` to be larger than 256 bits.

This is of course already covered via `mulDiv(x, y, 1 << n, Rounding.Floor)`.

But it (the method below) would consume a lot less gas, so you may consider it complementary for performance optimization.

**📝 Details**
```
function mulShr(uint256 x, uint256 y, uint8 n) internal pure returns (uint256) {
    unchecked {
        (uint256 xyh, uint256 xyl) = mul512(x, y);
        require(xyh < 1 << n);
        return (xyh << (256 - n)) | (xyl >> n);
    }
}

function mul512(uint256 x, uint256 y) private pure returns (uint256, uint256) {
    unchecked {
        uint256 p = mulmod(x, y, type(uint256).max);
        uint256 q = x * y;
        if (p >= q)
            return (p - q, q);
        return (p - q - 1, q);
    }
}
```
I'm pretty sure that you already have some version of `mul512` (above) implemented in the `Math` library.
For example, at the beginning of function `mulDiv`:
```
uint256 prod0 = x * y; // Least significant 256 bits of the product
uint256 prod1; // Most significant 256 bits of the product
assembly {
    let mm := mulmod(x, y, not(0))
    prod1 := sub(sub(mm, prod0), lt(mm, prod0))
}
```
So you might want to reuse that piece of code in some way.

---

An alternative which won't require updating the existing API is to add this in your `mulDiv` function:
```
if (denominator & (denominator - 1) == 0) {
    return (prod1 * (type(uint256).max / denominator + 1)) | (prod0 / denominator);
}
```
However, there are two issues with this option:
1. Minor issue: for cases which enter the `if` statement, it is slightly more expensive than the original suggestion
2. Major issue: it will slightly increase the cost for all other cases, which means worsening the worst-case scenario

Amxx:
Fixes #5433

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/utils/math/Math.sol
```diff
@@ -229,6 +249,26 @@ library Math {
...
+    /**
+     * @dev Calculates floor(x * y >> n) with full precision. Throws if result overflows a uint256.
+     */
+    function mulShr(uint256 x, uint256 y, uint8 n) internal pure returns (uint256 result) {
+        unchecked {
+            (uint256 prod1, uint256 prod0) = mul512(x, y);
+            if (prod1 >= 1 << n) {
+                Panic.panic(Panic.UNDER_OVERFLOW);
+            }
+            return (prod1 << (256 - n)) | (prod0 >> n);
```
arr00, line 261: I like `high`, `low` better as used in the function definition.

Amxx: For reference, prod1 and prod0 has been historically used in mulDiv 

Would you also change that part ?

arr00: Let's be consistent with whatever we choose. I think high and low is better for readability, do you agree? If so let's change that too.

arr00 (dismissed):

contracts/utils/math/Math.sol
Amxx: changed

arr00 (approved):

PR #5036, opened 2024-05-09
Gas loop optimization in `MerkleProof.sol`

### LINKED ISSUE #5028: Loop gas optimisation
The MerkleProof.sol contract by OZ is used very widely in contracts for EVM chains. Some of the functions are used for-loop. And the way that this for-loop is not the gas-optimal

**📝 Details**

After the gas optimized code will consume 100 gas units less per each element of an array.

**🔢 Code to reproduce bug**

The overheads outlined below are PER LOOP, excluding the first loop

storage arrays incur a Gwarmaccess (100 gas) memory arrays use MLOAD (3 gas) calldata arrays use CALLDATALOAD (3 gas) Caching the length changes each of these to a DUP<N> (3 gas), and gets rid of the extra DUP<N> needed to store the stack offset

```solidity
pragma solidity ^0.8.13;

contract Counter {

    uint256[] public proof;

    constructor() {
        for(uint256 i = 0; i < 100; i++) {
            proof.push(1);
        }
    }


    function loop_not_opt() public {
        uint256 j;
            j++;
        }
    }

    function loop_opt() public {
        uint256 j;
        uint256 len = proof.length;
        for (uint256 i = 0; i < len; i++) {
            j++;
        }
    }
}
```
As you see on a screenshot, the difference between these two calls is around 10_000 gas units, for a array length of a 100 elements.

Which is not pretty much for one call, But since a lot of contracts use this OZ library, the total amount of overspent gas is high.

**Code snippet:**
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c80b675b8db1d951b8b3734df59530d0d3be064b/contracts/utils/cryptography/MerkleProof.sol#L64 https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c80b675b8db1d951b8b3734df59530d0d3be064b/contracts/utils/cryptography/MerkleProof.sol#L53

**Recomendation**
And Recommendation is to calculate the length of an array one's and keep it locally
```solidity
 uint256 len = proof.length;
        for (uint256 i = 0; i < len; i++) {
            j++;
        }
```

CarsonCase:
Closes #5028 by caching the length of the proof to prevent an unnecessary DUP call (3 gas per iteration) 

#### PR Checklist

- [x] Tests
- [x] Documentation


cairoeth:
@CarsonCase reverted back the `unchecked` block because it's done automatically by Solidity 0.8.22 onwards, and also made the `i` variable explicit

cairoeth:
Gas difference:

|                                        | Normal             | Optimizer (1000 runs) | Via IR               |
|----------------------------------------|--------------------|-----------------------|----------------------|
| testProcessProof (unoptimized)         | μ: 41703, ~: 41638 | μ: 41699, ~: 41638    | μ: 34632, ~: 34667   |
| testProcessProof (optimized)           | μ: 41335, ~: 41274 | μ: 41331, ~: 41274    | μ: 37574, ~: 37602   |
| diff                                   | μ: -368, ~: -364   | μ: -368, ~: -364      | μ: +2,942, ~: +2,935 |
|                                        |                    |                       |                      |
| testprocessProofCalldata (unoptimized) | μ: 31410, ~: 31343 | μ: 31362, ~: 31299    | μ: 24572, ~: 24567   |
| testprocessProofCalldata (optimized)   | μ: 31418, ~: 31351 | μ: 31370, ~: 31307    | μ: 24572, ~: 24567   |
| diff                                   | μ: +8, ~: +8       | μ: +8, ~: +8          | μ: 0, ~: 0           |

ernestognw (commented):
It does look good to me but I am not sure if the tradeoff makes sense.

From the gas benchmarks we can see `--via-ir` is even worse in the case of memory. For the `calldata` one I see no difference, which is consistent with how `--via-ir` works as it can assume that calldata never changes, thus it can be taken out of the loop.

At the moment, seems like we're largely affecting projects using `--via-ir` but improving performance for those who don't, and I don't have an opinion yet 🤔

Amxx:
As mentionned by @cairoeth, recent version of the compiler now automatically cache the value (and perform "unchecked" incrementation of the counter). IMO we should keep the code as clean as possible and leave that to the compiler.

Additionnally, this is clearly a micro optimization. I would expect developpers who value this kind of optimisation to use the --via-ir pipeline. If this "optimization" doesn't provide good results when --via-ir is enabled, and even degrades performances, then I think we should definitelly not implement it.

PR #5044, opened 2024-05-14
develop read-optimized storage

havi-kim:
### Idea
The ReadOptimizedStorage library is designed to efficiently store and retrieve data that does not change frequently but needs to be read frequently. This library leverages the concept of copying contract code to transient storage when read, which can improve read performance in some scenarios.

### How It Works
- Writing Data
  - When data is added or modified, it is encoded into bytes and used to create a new storage contract. This storage contract holds the key-value pairs. The storage contract is then stored in a dedicated storage slot.

- Reading Data
  - When data is read, the extcodecopy operation is used to extract the key-value pairs from the storage contract. These key-value pairs are decoded and stored in transient storage in bulk. This approach ensures that subsequent read operations are efficient.

### Benefits
This method is particularly useful for protocols that:

- Choose not to adopt the upgradeable delegate call-based proxy pattern and prefer separate storage contracts.
- Need to read configuration or similar values that do not change often but require multiple storage slots (more than four) to read.

### DIFF
--- a/contracts/utils/storage/ExternalStorage.sol
+++ b/contracts/utils/storage/ExternalStorage.sol
@@ -1,0 +1,113 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "contracts/utils/storage/IStorage.sol";
+
+library ExternalStorage {
+    IExternalStorage private constant _STORAGE_CONTRACT =
+        IExternalStorage(address(0x5615dEB798BB3E4dFa0139dFa1b3D433Cc23b72f));
+
+    function writeUint(IStorage _key, uint256 _value) internal {
+        _STORAGE_CONTRACT.writeUint(_key, _value);
+    }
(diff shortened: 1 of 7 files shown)


OKEAMAH (commented):
Cool

cairoeth:
We already have [`TransientSlot`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/TransientSlot.sol), [`StorageSlot`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/StorageSlot.sol), and [`SlotDerivation`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/SlotDerivation.sol), which cover most of these use cases. Therefore, closing this PR. If you still think there are some use cases that should be implemented, please open a issue and describe the motivation + idea first.

PR #5045, opened 2024-05-15
Add GovernorCountingFractional

### LINKED ISSUE #4958: Add Flexible Voting to Governor
**🧐 Motivation**

Scopelift's [Flexible Voting extension](https://flexiblevoting.com/) ([github](https://github.com/ScopeLift/flexible-voting)) has been live for ~1.5 years. Several Governor implementations use it (Frax, Gitcoin, and PoolTogether) with more in the pipeline (Wormhole, Arbitrum, RadWorks). 

The extension supports several new use cases, such as voting from a [DeFi protocol](https://scopelift.co/blog/how-scopelift-built-a-flex-voting-atoken-on-aave) and [multichain voting](https://github.com/ScopeLift/l2-flexible-voting). I have also talked to projects interested in building private voting or liquid delegation features on top of Flexible Voting. 

I would be in favor of merging the Flexible Voting Extension into OZ Governor contracts to solidify the standard and make it easier for builders to build new use cases.  I understand that there are other possible standards for fractional voting, but this one seems to be the one with by far the most traction.

**📝 Details**
The changes to support Flexible Voting are in [this repository](https://github.com/ScopeLift/flexible-voting/blob/master/src/GovernorCountingFractional.sol). If there is interest in this proposal, the ScopeLift team would probably be able to help advise on next steps for merging the code.

Amxx:
Fixes #4958 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -135,10 +135,14 @@ abstract contract GovernorCountingFractional is Governor {
         }
 
         // For clarity of event indexing, fractional voting must be clearly advertised in the "support" field.
+        //
+        // Supported `support` value must be:
+        // - "Full" voting: `support = 0` (Against), `1` (For) or `2` (Abstain), with empty params.
+        // - "Fractional" voting: `support = 255`, with 48 bytes params.
         if (support <= uint8(GovernorCountingSimple.VoteType.Abstain)) {
             if (params.length != 0) revert GovernorInvalidVoteParams();
             return _countVoteNominal(proposalId, account, support, remainingWeight);
-        } else if (support == uint8(GovernorCountingSimple.VoteType.Fractional)) {
+        } else if (support == type(uint8).max) {
             if (params.length != 0x30) revert GovernorInvalidVoteParams();
             return _countVoteFractional(proposalId, account, params, remainingWeight);
         } else {
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -7,12 +7,17 @@ import {GovernorCountingSimple} from "./GovernorCountingSimple.sol";
 import {Math} from "../../utils/math/Math.sol";
 
 /**
- * @notice Extension of {Governor} for 3 option fractional vote counting. When voting, a delegate may split their vote
- * weight between Against/For/Abstain. This is most useful when the delegate is itself a contract, implementing its
- * own rules for voting. By allowing a contract-delegate to split its vote weight, the voting preferences of many
- * disparate token holders can be rolled up into a single vote to the Governor itself. Some example use cases include
- * voting with tokens that are held by a DeFi pool, voting from L2 with tokens held by a bridge, or voting privately
- * from a shielded pool using zero knowledge proofs.
+ * @dev Extension of {Governor} for fractional 3 option vote counting.
+ *
+ * Voters can split their voting power amongst 3 options: For, Against and Abstain.
+ * This is mostly useful when the delegate is a contract that implements its own rules for voting. These delegate-contracts
+ * can cast fractional votes according to the preferences of multiple entities delegating their voting power.
+ *
+ * Some example use cases include:
+ *
+ * * Voting from tokens that are held by a DeFi pool
+ * * Voting from an L2 with tokens held by a bridge
+ * * voting privately from a shielded pool using zero knowledge proofs.
  *
  * Based on ScopeLift's https://github.com/ScopeLift/flexible-voting/blob/master/src/GovernorCountingFractional.sol
  */
@@ -52,7 +57,7 @@ abstract contract GovernorCountingFractional is Governor {
     }
 
     /**
-     * @dev Get the number of votes cast thus far on proposal `proposalId` by account `account`. Useful for
+     * @dev Get the number of votes already cast by `account` for a proposal with `proposalId`. Useful for
      * integrations that allow delegates to cast rolling, partial votes.
      */
     function usedVotes(uint256 proposalId, address account) public view virtual returns (uint256) {
@@ -60,7 +65,7 @@ abstract contract GovernorCountingFractional is Governor {
     }
 
     /**
-     * @dev Accessor to the internal vote counts.
+     * @dev Get current distribution of votes for a given proposal.
      */
     function proposalVotes(
         uint256 proposalId
@@ -102,6 +107,10 @@ abstract contract GovernorCountingFractional is Governor {
      *
      * The sum total of the three decoded vote weights _must_ be less than or equal to the delegate's remaining weight
      * on the proposal, i.e. their checkpointed total weight minus votes already cast on the proposal.
+     *
+     * NOTE: Consider the number of votes is restricted to 128 bits. Depending on how many decimals the underlying token
+     * has, a single voter may require to split their vote into multiple transactions. For precision higher than
+     * ~30 decimals, large token holders may require an exponentially large number of transactions to cast their vote.
      *
      * See `_countVoteNominal` and `_countVoteFractional` for more details.
      */
@@ -158,17 +167,19 @@ abstract contract GovernorCountingFractional is Governor {
     /**
      * @dev Count votes with fractional weight.
      *
-     * `params` is expected to be tree packed uint128s:
+     * The `params` argument is expected to be three packed `uint128`:
      * `abi.encodePacked(uint128(againstVotes), uint128(forVotes), uint128(abstainVotes))`
      *
-     * This function can be called multiple times for the same account and proposal, i.e. partial/rolling votes are
-     * allowed. For example, an account with total weight of 10 could call this function three times with the
+     * This function can be called multiple times for the same account and proposal (i.e. partial/rolling votes are
+     * allowed). For example, an account with total weight of 10 could call this function three times with the
      * following vote data:
-     *   - against: 1, for: 0, abstain: 2
-     *   - against: 3, for: 1, abstain: 0
-     *   - against: 1, for: 1, abstain: 1
-     * The result of these three calls would be that the account casts 5 votes AGAINST, 2 votes FOR, and 3 votes
-     * ABSTAIN on the proposal. Though partial, votes are still final once cast and cannot be changed or overridden.
+     *
+     *   * Against: 1, For: 0, Abstain: 2
+     *   * Against: 3, For: 1, Abstain: 0
+     *   * Against: 1, For: 1, Abstain: 1
+     *
+     * Casting votes like this will make the calling account to cast a total of 5 `Against` votes, 2 `For` votes
+     * and 3 `Abstain` votes. Though partial, votes are still final once cast and cannot be changed or overridden.
      * Subsequent partial votes simply increment existing totals.
      */
     function _countVoteFractional(
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -91,9 +91,7 @@ abstract contract GovernorCountingFractional is Governor {
     }
 
     /**
-     * @notice See {Governor-_countVote}.
-     *
-     * @dev Function that records the delegate's votes.
+     * @dev See {Governor-_countVote}. Function that records the delegate's votes.
      *
      * If the `params` bytes parameter is empty, then this module behaves identically to {GovernorCountingSimple} for
      * the remaining weight. That is, it assigns the remaining weight of the delegate to the `support` parameter,
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -13,7 +13,7 @@ import {Math} from "../../utils/math/Math.sol";
  * Against, For, Abstain. Additionally, it includes a fourth option: Fractional, which allows voters to split their voting
  * power amongst the other 3 options.
  *
- * Votes cast with the Fractional support must be accompanied by a `params` argument that is tree packed `uint128` values
+ * Votes cast with the Fractional support must be accompanied by a `params` argument that is three packed `uint128` values
  * representing the weight the delegate assigns to Against, For, and Abstain respectively. For those votes cast for the other
  * 3 options, the `params` argument must be empty.
  *
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -19,7 +19,7 @@ import {Math} from "../../utils/math/Math.sol";
  * * Voting from an L2 with tokens held by a bridge
  * * voting privately from a shielded pool using zero knowledge proofs.
  *
- * Based on ScopeLift's https://github.com/ScopeLift/flexible-voting/blob/master/src/GovernorCountingFractional.sol
+ * Based on ScopeLift's https://github.com/ScopeLift/flexible-voting/blob/e5de2efd1368387b840931f19f3c184c85842761/src/GovernorCountingFractional.sol
  */
 abstract contract GovernorCountingFractional is Governor {
     using Math for *;
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -145,7 +145,6 @@ abstract contract GovernorCountingFractional is Governor {
             if (params.length != 0) revert GovernorInvalidVoteParams();
             return _countVoteNominal(proposalId, account, support, remainingWeight);
         } else if (support == VOTE_TYPE_FRACTIONAL) {
-            if (params.length != 0x30) revert GovernorInvalidVoteParams();
             return _countVoteFractional(proposalId, account, params, remainingWeight);
         } else {
             revert GovernorInvalidVoteType();
@@ -202,9 +201,7 @@ abstract contract GovernorCountingFractional is Governor {
         bytes memory params,
         uint256 weight
     ) private returns (uint256) {
-        uint128 againstVotes = _extractUint128(params, 0);
-        uint128 forVotes = _extractUint128(params, 1);
-        uint128 abstainVotes = _extractUint128(params, 2);
+        (uint128 againstVotes, uint128 forVotes, uint128 abstainVotes) = _unpackParams(params);
 
         uint256 usedWeight = againstVotes + forVotes + abstainVotes;
         if (usedWeight > weight) {
@@ -226,9 +223,18 @@ abstract contract GovernorCountingFractional is Governor {
         return usedWeight;
     }
 
-    function _extractUint128(bytes memory data, uint256 pos) private pure returns (uint128 result) {
-        assembly {
-            result := shr(128, mload(add(data, add(0x20, mul(0x10, pos)))))
+    function _unpackParams(
+        bytes memory data
+    ) private pure returns (uint128 againstVotes, uint128 forVotes, uint128 abstainVotes) {
+        if (data.length != 0x30) {
+            revert GovernorInvalidVoteParams();
+        }
+
+        assembly ("memory-safe") {
+            let ptr := add(data, 0x20)
+            againstVotes := shr(128, mload(ptr))
+            forVotes := shr(128, mload(add(ptr, 0x10)))
+            abstainVotes := shr(128, mload(add(ptr, 0x20)))
         }
     }
 }
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -31,6 +31,8 @@ import {Math} from "../../utils/math/Math.sol";
 abstract contract GovernorCountingFractional is Governor {
     using Math for *;
 
+    uint8 internal constant VOTE_TYPE_FRACTIONAL = 255;
+
     struct ProposalVote {
         uint256 againstVotes;
         uint256 forVotes;
@@ -53,7 +55,7 @@ abstract contract GovernorCountingFractional is Governor {
      */
     // solhint-disable-next-line func-name-mixedcase
     function COUNTING_MODE() public pure virtual override returns (string memory) {
-        return "support=bravo,params&quorum=for,abstain&params=fractional";
+        return "support=bravo,fractional&quorum=for,abstain&params=fractional";
     }
 
     /**
@@ -142,7 +144,7 @@ abstract contract GovernorCountingFractional is Governor {
         if (support <= uint8(GovernorCountingSimple.VoteType.Abstain)) {
             if (params.length != 0) revert GovernorInvalidVoteParams();
             return _countVoteNominal(proposalId, account, support, remainingWeight);
-        } else if (support == type(uint8).max) {
+        } else if (support == VOTE_TYPE_FRACTIONAL) {
             if (params.length != 0x30) revert GovernorInvalidVoteParams();
             return _countVoteFractional(proposalId, account, params, remainingWeight);
         } else {
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -31,8 +31,10 @@ abstract contract GovernorCountingFractional is Governor {
      */
     mapping(uint256 => ProposalVote) private _proposalVotes;
 
-    error GovernorInvalidParamsFormat(address voter);
-    error GovernorUsedVotesExceedRemainingWeight(address voter, uint256 usedVotes, uint256 remainingWeight);
+    /**
+     * @dev A fractional vote params uses more votes than are available for that user.
+     */
+    error GovernorExceedRemainingWeight(address voter, uint256 usedVotes, uint256 remainingWeight);
 
     /**
      * @dev See {IGovernor-COUNTING_MODE}.
@@ -88,9 +90,9 @@ abstract contract GovernorCountingFractional is Governor {
      *
      * @dev Function that records the delegate's votes.
      *
-     * If the `params` bytes parameter is empty, then this module behaves identically to {GovernorCountingSimple}.
-     * That is, it assigns the remaining weight of the delegate to the `support` parameter, which follows the
-     * `VoteType` enum from Governor Bravo (as defined in {GovernorCountingSimple}).
+     * If the `params` bytes parameter is empty, then this module behaves identically to {GovernorCountingSimple} for
+     * the remaining weight. That is, it assigns the remaining weight of the delegate to the `support` parameter,
+     * which follows the `VoteType` enum from Governor Bravo (as defined in {GovernorCountingSimple}).
      *
      * If the `params` bytes parameter is not zero, then it _must_ be tree packed uint128s, totaling 48 bytes,
      * representing the weight the delegate assigns to Against, For, and Abstain respectively. This format can be
@@ -121,7 +123,7 @@ abstract contract GovernorCountingFractional is Governor {
         } else if (params.length == 0x30) {
             return _countVoteFractional(proposalId, account, params, remainingWeight);
         } else {
-            revert GovernorInvalidParamsFormat(account);
+            revert GovernorInvalidVoteParams();
         }
     }
 
@@ -181,7 +183,7 @@ abstract contract GovernorCountingFractional is Governor {
 
         uint256 usedWeight = againstVotes + forVotes + abstainVotes;
         if (usedWeight > weight) {
-            revert GovernorUsedVotesExceedRemainingWeight(account, usedWeight, weight);
+            revert GovernorExceedRemainingWeight(account, usedWeight, weight);
         }
 
         ProposalVote storage details = _proposalVotes[proposalId];
--- a/contracts/governance/extensions/GovernorCountingFractional.sol
+++ b/contracts/governance/extensions/GovernorCountingFractional.sol
@@ -48,14 +48,14 @@ abstract contract GovernorCountingFractional is Governor {
      * @dev See {IGovernor-hasVoted}.
      */
     function hasVoted(uint256 proposalId, address account) public view virtual override returns (bool) {
-        return voteWeightCast(proposalId, account) > 0;
+        return usedVotes(proposalId, account) > 0;
     }
 
     /**
      * @dev Get the number of votes cast thus far on proposal `proposalId` by account `account`. Useful for
      * integrations that allow delegates to cast rolling, partial votes.
      */
-    function voteWeightCast(uint256 proposalId, address account) public view virtual returns (uint256) {
+    function usedVotes(uint256 proposalId, address account) public view virtual returns (uint256) {
         return _proposalVotes[proposalId].usedVotes[account];
     }
 
@@ -113,7 +113,7 @@ abstract contract GovernorCountingFractional is Governor {
         bytes memory params
     ) internal virtual override returns (uint256) {
         // Compute number of remaining votes. Returns 0 on overflow.
-        (, uint256 remainingWeight) = totalWeight.trySub(voteWeightCast(proposalId, account));
+        (, uint256 remainingWeight) = totalWeight.trySub(usedVotes(proposalId, account));
         if (remainingWeight == 0) {
             revert GovernorAlreadyCastVote(account);
         }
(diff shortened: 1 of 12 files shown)


test/governance/extensions/GovernorCountingFractional.test.js
```diff
@@ -0,0 +1,195 @@
...
+            }),
+          )
+            .to.emit(this.mock, 'VoteCastWithParams')
+            .withArgs(
+              this.voter2,
+              this.proposal.id,
+              0,
+              ethers.parseEther('7'), // total weight for voter2
+              'no particular reason',
+              params,
```
Amxx, line 119: Events are not great. Indexing is likelly going to be broken :/

Amxx: Asked Tally how they are handling it.

Amxx:
[12123b0](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5045/commits/12123b0255e9bd5abd7e7511975d26e79c326a7b) fixes the events, but is a breaking change. I don't think its a big deal (it wont be silent, compiler fail) ... but its not ideal.

@ernestognw @frangio WDYT?

frangio:
I don't see a way to fix this without a breaking change. Another possible fix would be to add a new internal function. But I think just adding a return value is a better option.

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,201 @@
...
+        mapping(address voter => uint256) usedVotes;
+    }
+
+    /**
+     * @dev Mapping from proposal ID to vote tallies for that proposal.
+     */
+    mapping(uint256 => ProposalVote) private _proposalVotes;
+
+    error GovernorInvalidParamsFormat(address voter);
+    error GovernorUsedVotesExceedRemainingWeight(address voter, uint256 usedVotes, uint256 remainingWeight);
```
ernestognw, line 35: This name feels weird. I'd say this follows our guidelines better:

```suggestion
    error GovernorExceedRemainingWeight(address voter, uint256 usedVotes, uint256 remainingWeight);
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,201 @@
...
+ * Based on ScopeLift's https://github.com/ScopeLift/flexible-voting/blob/master/src/GovernorCountingFractional.sol
+ */
+abstract contract GovernorCountingFractional is Governor {
+    using Math for *;
+
+    struct ProposalVote {
+        uint256 againstVotes;
+        uint256 forVotes;
+        uint256 abstainVotes;
+        mapping(address voter => uint256) usedVotes;
```
ernestognw, line 26: I'm confused why this has a different name internally to externally. Honestly I think it's better to have this for consistency, otherwise I need to keep a mental link between the getter and the variable while reading

```suggestion
        mapping(address voter => uint256) voteWeightCast;
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * If the `params` bytes parameter is empty, then this module behaves identically to {GovernorCountingSimple}.
+     * That is, it assigns the remaining weight of the delegate to the `support` parameter, which follows the
+     * `VoteType` enum from Governor Bravo (as defined in {GovernorCountingSimple}).
+     *
+     * If the `params` bytes parameter is not zero, then it _must_ be tree packed uint128s, totaling 48 bytes,
+     * representing the weight the delegate assigns to Against, For, and Abstain respectively. This format can be
+     * produced using:
+     *
+     * `abi.encodePacked(uint128(againstVotes), uint128(forVotes), uint128(abstainVotes))`
```
ernestognw, line 99: I'd challenge this format a bit. It essentially means that any casted vote with values above `type(uint128).max` will need to vote twice. These may be handled on the app side, but are highly unlikely and most likely that justifies this strategy in the first place.

Shall we mention that in the docs? 2**128 = 3.4e38 and most tokens might have supply of < 1e15 with 18 decimals totalling e23. I think there's room enough to not be an issue but this is the kind of things Code4rena auditors look for 👀

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,201 @@
...
+        ProposalVote storage proposalVote = _proposalVotes[proposalId];
+        return proposalVote.forVotes > proposalVote.againstVotes;
+    }
+
+    /**
+     * @notice See {Governor-_countVote}.
+     *
+     * @dev Function that records the delegate's votes.
+     *
+     * If the `params` bytes parameter is empty, then this module behaves identically to {GovernorCountingSimple}.
```
ernestognw, line 91: This is partially true

```suggestion
     * If the `params` bytes parameter is empty, then this module behaves identically to {GovernorCountingSimple} for the remaining weight.
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,201 @@
...
+        uint256 abstainVotes;
+        mapping(address voter => uint256) usedVotes;
+    }
+
+    /**
+     * @dev Mapping from proposal ID to vote tallies for that proposal.
+     */
+    mapping(uint256 => ProposalVote) private _proposalVotes;
+
+    error GovernorInvalidParamsFormat(address voter);
```
ernestognw, line 34: I'd consider moving this error to Governor's even if it's not used there. It would keep the error consistent if we add other Governor extensions relying in the params format.

contracts/governance/extensions/GovernorCountingFractional.sol
ernestognw, line 1: I feel this contract is adding two functionalities that can be split:

1. A partial voting mechanism that substracts from the voter's total weight on each vote
2. A custom voting format

By abstracting the first one away we'd have a base governor contract from which anyone can extend from and add their own voting formats. In that case this contract would be like a "basic" fractional voting.

Not sure if it's even worth pursuing the idea, but I wanted to share it.

ernestognw:
> But I think just adding a return value is a better option.

Yeah same, it's one of those breaking changes that are easy to fix imo.

test/governance/extensions/GovernorCountingFractional.test.js
Amxx: Answer: 
- if params is not empty
  - disregard `support` and `weight` from the event
  - decode the params as 3 uint128
  - use these values
- if params is empty
  - use support and weight from the event

This confirms that we should fix the event to emit the correct weight when doing a "norminal" vote with the remaining power.

One question remain: How do we deal with support if params are not empty. 
- Is it ok to pass the value sent by the user without checking it.
- Should we add a 4th value (Against, For, Abstain, Fractional), and require that if params is not empty, then support must be Fractional ? This would possibly break compatibility existing modules for fractional counting that do not support that 4th option.

contracts/governance/extensions/GovernorCountingFractional.sol
Amxx: This format was "decided" by Scopelift, and is supported by Tally. If we change it, we break compatibility.

contracts/governance/extensions/GovernorCountingFractional.sol
Amxx: `voteWeightCast` is a public part of the Scopelift module. For compatibility reason, I think we should probably keep the name.

But I hate that name. At the bare minimum it should be `voteWeightCasted`. Lets discuss that on a call.

test/governance/extensions/GovernorCountingFractional.test.js
frangio: >This would possibly break compatibility existing modules for fractional counting that do not support that 4th option.

It would be different behavior, but we can specify that support should be ignored and in that sense it would remain compatible.

Amxx: I was afraid that Scopelift's version would not accept any support that is not 0, 1 or 2. But that is only true if there is no params. If they have params, then the support is discarded (just like we do now).

So right now, where there are params, both version (scopelift and this PR) will accept any support. The question is do we want to force the support to a specific value (which scopelift doesn't do) so that event are "clearer" ?

contracts/governance/extensions/GovernorCountingFractional.sol
ernestognw: We agreed keeping this out of scope right now since it may come with breaking changes and the benefits aren't clear.

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+import {GovernorCountingSimple} from "./GovernorCountingSimple.sol";
+import {Math} from "../../utils/math/Math.sol";
+
+/**
+ * @notice Extension of {Governor} for 3 option fractional vote counting. When voting, a delegate may split their vote
+ * weight between Against/For/Abstain. This is most useful when the delegate is itself a contract, implementing its
+ * own rules for voting. By allowing a contract-delegate to split its vote weight, the voting preferences of many
+ * disparate token holders can be rolled up into a single vote to the Governor itself. Some example use cases include
+ * voting with tokens that are held by a DeFi pool, voting from L2 with tokens held by a bridge, or voting privately
+ * from a shielded pool using zero knowledge proofs.
```
ernestognw, line 15: Note `@dev` is needed, otherwise we won't have the description in the API docs

```suggestion
 * @dev Extension of {Governor} for fractional 3 option vote counting.
 *
 * Voters can split their voting power amongst 3 options: For, Against and Abstain.
 * This is mostly useful when the delegate is a contract that implements its own rules for voting. These delegate-contracts
 * can cast fractional votes according to the preferences of multiple entities delegating their voting power.
 *
 * Some example use cases include:
 *
 * * Voting from tokens that are held by a DeFi pool
 * * Voting from an L2 with tokens held by a bridge
 * * voting privately from a shielded pool using zero knowledge proofs.
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+    /**
+     * @dev See {IGovernor-hasVoted}.
+     */
+    function hasVoted(uint256 proposalId, address account) public view virtual override returns (bool) {
+        return usedVotes(proposalId, account) > 0;
+    }
+
+    /**
+     * @dev Get the number of votes cast thus far on proposal `proposalId` by account `account`. Useful for
+     * integrations that allow delegates to cast rolling, partial votes.
```
ernestognw, line 56:
```suggestion
     * @dev Get the number of votes already cast by `account` for a proposal with `proposalId`. Useful for
     * integrations that allow delegates to cast rolling, partial votes.
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+     */
+    function usedVotes(uint256 proposalId, address account) public view virtual returns (uint256) {
+        return _proposalVotes[proposalId].usedVotes[account];
+    }
+
+    /**
+     * @dev Accessor to the internal vote counts.
```
ernestognw, line 63:
```suggestion
     * @dev Get current distribution of votes for a given proposal.
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+    /**
+     * @dev See {Governor-_voteSucceeded}. In this module, forVotes must be > againstVotes.
+     */
+    function _voteSucceeded(uint256 proposalId) internal view virtual override returns (bool) {
+        ProposalVote storage proposalVote = _proposalVotes[proposalId];
+        return proposalVote.forVotes > proposalVote.againstVotes;
+    }
+
+    /**
+     * @notice See {Governor-_countVote}.
```
ernestognw, line 89: This line won't show up in the docs so the format is unnecessary, why not using the same as we've done in the past?:

```
@dev See {[ref]}. ][Some extra info]
```

contracts/governance/extensions/GovernorCountingFractional.sol
ernestognw: Agree. What about?

```suggestion
     * If the `params` bytes parameter is not zero, then it _must_ be tree packed uint128s, totaling 48 bytes,
     * representing the weight the delegate assigns to Against, For, and Abstain respectively. This format can be
     * produced using:
     *
     * `abi.encodePacked(uint128(againstVotes), uint128(forVotes), uint128(abstainVotes))`
     *
     * NOTE: Consider the number of votes is restricted to 128 bits. Depending on how many decimals the underlying token
     * has, a single voter may require to split their vote into multiple transactions. For precision higher than
     * ~30 decimals, large token holders may require an exponentially large number of transactions to cast their vote.
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+            revert GovernorAlreadyCastVote(account);
+        }
+
+        if (params.length == 0) {
+            return _countVoteNominal(proposalId, account, support, remainingWeight);
+        } else if (params.length == 0x30) {
+            return _countVoteFractional(proposalId, account, params, remainingWeight);
+        } else {
+            revert GovernorInvalidVoteParams();
+        }
```
ernestognw, line 127: nit, I just think this is clearer

```suggestion
        }
        revert GovernorInvalidVoteParams();
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * This function can be called multiple times for the same account and proposal, i.e. partial/rolling votes are
+     * allowed. For example, an account with total weight of 10 could call this function three times with the
+     * following vote data:
+     *   - against: 1, for: 0, abstain: 2
+     *   - against: 3, for: 1, abstain: 0
+     *   - against: 1, for: 1, abstain: 1
+     * The result of these three calls would be that the account casts 5 votes AGAINST, 2 votes FOR, and 3 votes
+     * ABSTAIN on the proposal. Though partial, votes are still final once cast and cannot be changed or overridden.
+     * Subsequent partial votes simply increment existing totals.
```
ernestognw, line 172:
```suggestion
     * The `params` argument is expected to be three packed `uint128`:
     * `abi.encodePacked(uint128(againstVotes), uint128(forVotes), uint128(abstainVotes))`
     *
     * This function can be called multiple times for the same account and proposal (i.e. partial/rolling votes are
     * allowed). For example, an account with total weight of 10 could call this function three times with the
     * following vote data:
     *
     *   * Against: 1, For: 0, Abstain: 2
     *   * Against: 3, For: 1, Abstain: 0
     *   * Against: 1, For: 1, Abstain: 1
     *
     * Casting votes like this will make the calling account to cast a total of 5 `Against` votes, 2 `For` votes 
     * and 3 `Abstain` votes. Though partial, votes are still final once cast and cannot be changed or overridden.
     * Subsequent partial votes simply increment existing totals.
```

ernestognw (commented):
Almost there!

test/governance/extensions/GovernorCountingFractional.test.js
ernestognw: In my opinion we should make clear what the event represents. Given that the optional parameters are arbitrary bytes we can't specify a format, so the general guideline here should to be look for making the optional parameters as clear as possible.

I'd be in favor of changing the format, we're not restricted yet by backwards compatibility and the only constraint would be that Tally (or others) should update to support this variant. I think that's better than us starting with a design we're not comfortable with.

What do you think about restricting nominal voting to only when no previous vote was casted? That'd make it consistent, either it's a single nominal vote or multiple fractional ones.

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+
+/**
+ * @notice Extension of {Governor} for 3 option fractional vote counting. When voting, a delegate may split their vote
+ * weight between Against/For/Abstain. This is most useful when the delegate is itself a contract, implementing its
+ * own rules for voting. By allowing a contract-delegate to split its vote weight, the voting preferences of many
+ * disparate token holders can be rolled up into a single vote to the Governor itself. Some example use cases include
+ * voting with tokens that are held by a DeFi pool, voting from L2 with tokens held by a bridge, or voting privately
+ * from a shielded pool using zero knowledge proofs.
+ *
+ * Based on ScopeLift's https://github.com/ScopeLift/flexible-voting/blob/master/src/GovernorCountingFractional.sol
```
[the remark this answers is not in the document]
Amxx:
```suggestion
 * Based on ScopeLift's https://github.com/ScopeLift/flexible-voting/blob/e5de2efd1368387b840931f19f3c184c85842761/src/GovernorCountingFractional.sol
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,209 @@
...
+     */
+    function _voteSucceeded(uint256 proposalId) internal view virtual override returns (bool) {
+        ProposalVote storage proposalVote = _proposalVotes[proposalId];
+        return proposalVote.forVotes > proposalVote.againstVotes;
+    }
+
+    /**
+     * @notice See {Governor-_countVote}.
+     *
+     * @dev Function that records the delegate's votes.
```
Amxx, line 96:
```suggestion
     * @dev Function that records the delegate's votes. See {Governor-_countVote}.
```

test/governance/extensions/GovernorCountingFractional.test.js
Amxx: Added "Fractional" to the VoteType enum following our discussion. WDYT?

contracts/governance/extensions/GovernorCountingSimple.sol
```diff
@@ -15,7 +15,8 @@ abstract contract GovernorCountingSimple is Governor {
     enum VoteType {
         Against,
         For,
-        Abstain
+        Abstain,
+        Fractional
```
ernestognw, line 19: I'm not convinced it's fine to extend the previous enum. I think it should be generally a bad practice to extend enums because it's assigning an arbitrary meaning to an unused `uint8` (in the context of `GovernorCountingSimple`).

This would make sense if `GovernorCountingFractional` extended from `GovernorCountingSimple` or a common implementation, but we've deliberately decided to keep their implementations separate so I think it may be better to have a vote type enum exclusive for `GovernorCountingFractional`.

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * * Voting from tokens that are held by a DeFi pool
+ * * Voting from an L2 with tokens held by a bridge
+ * * voting privately from a shielded pool using zero knowledge proofs.
+ *
+ * Based on ScopeLift's https://github.com/ScopeLift/flexible-voting/blob/e5de2efd1368387b840931f19f3c184c85842761/src/GovernorCountingFractional.sol
+ */
+abstract contract GovernorCountingFractional is Governor {
+    using Math for *;
+
```
ernestognw, line 33: Following my previous comment, I'd suggest:

```suggestion
enum FractionalVoteType {
  Against
  For
  Abstain
  Fractional
}
```

I still believe it's handy to keep the Bravo vote type ordering.

test/helpers/enums.js
```diff
@@ -5,7 +5,7 @@ function Enum(...options) {
 module.exports = {
   Enum,
   ProposalState: Enum('Pending', 'Active', 'Canceled', 'Defeated', 'Succeeded', 'Queued', 'Expired', 'Executed'),
-  VoteType: Enum('Against', 'For', 'Abstain'),
+  VoteType: Enum('Against', 'For', 'Abstain', 'Fractional'),
```
ernestognw, line 8:
```suggestion
  VoteType: Enum('Against', 'For', 'Abstain'),
  FractionalVoteType: Enum('Against', 'For', 'Abstain', 'Fractional'),
```

test/governance/extensions/GovernorCountingFractional.test.js
ernestognw: Addition looks good to me. Left a few comments but I'd say we can close this.

contracts/governance/extensions/GovernorCountingFractional.sol
Amxx: I'm not sure I understand. Current change keeps the Bravo ordering.

contracts/governance/extensions/GovernorCountingSimple.sol
Amxx: I think there is a value in having the enum declared only once.

If we declare 
```
// in GovernorCountingSimple
enum VoteType { Against, For, Abstain }
// in GovernorCountingFractional
enum VoteTypeFractional { Against, For, Abstain, Fractional }
```
then we run the risk of someone else, not aware of `GovernorCountingFractional`, extending `VoteType` with another 4th option that is not equivalent to Fractional. This would be terrible to indexers.

IMO the only good options are:
- declaring Fractional as a "core" enum entry. Ideally I would declare the enum elsewhere, but that would be a big breaking change
- not delcaring Franctional as an enum entry, and use `type(uint8).max` instead.

ernestognw: Yeah, I think it's terrible for general-purpose indexers, but there is already a clarification in [the event](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/93f460f8854bbfcdd0d1150c786b61ce7646330c/contracts/governance/IGovernor.sol#L145):

> `support` values should be seen as buckets. Their interpretation depends on the voting module used.

I don't think it's a big deal for indexers who're tracking their own Governor, since they will never have a conflict definition for `support`. And even if that's the case, I think there might be already other uses of `support` we're not aware of.

> not declaring Fractional as an enum entry, and use type(uint8).max instead.

This sounds like a good option. I don't particularly prefer one over the other.

contracts/governance/extensions/GovernorCountingFractional.sol
ernestognw: The suggestion is just to add the enum definition here because I think it's specific to `GovernorCountingFractional`. I say I think it's handy that it uses the same order as Bravo although that's not strictly required.

contracts/governance/extensions/GovernorCountingSimple.sol
Amxx: I implemented the `type(uint8).max` option.

I'd love @frangio 's feedback.

contracts/governance/extensions/GovernorCountingFractional.sol
Amxx: I expect most voters will not use params, and will instead relly on the canonical (bravo) methods. I think it is essential to not change the meaning of support for them.

ernestognw: So I do. That's why this is **not suggesting to change the order**

Amxx: My comment was on "although that's not strictly required." ... IMO it is.

We all agree on not changing it!

contracts/governance/extensions/GovernorCountingSimple.sol
frangio: > then we run the risk of someone else, not aware of `GovernorCountingFractional`, extending `VoteType` with another 4th option that is not equivalent to Fractional. This would be terrible to indexers.

The point of `COUNTING_MODE` was that contracts can declare how the `support` value should be interpreted. As long as `COUNTING_MODE` is different, the enums can be completely different.

IMO this contract should have a separate enum, preserving the order and adding the new `Fractional` variant, and declare something like `support=fractional` in `COUNTING_MODE`.

frangio: > preserving the order and adding the new `Fractional` variant

Actually, I don't have a strong preference for this. It would be fine to use `type(uint8).max`, but this still requires changing the value of `support` in the counting mode!

Amxx: >but this still requires changing the value of support in the counting mode!

current commit uses:
```solidity
    function COUNTING_MODE() public pure virtual override returns (string memory) {
        return "support=bravo,params&quorum=for,abstain&params=fractional";
    }
```

do you feel that is ok? would you do `support=bravo+params` or something completelly different ?

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,232 @@
...
+import {Math} from "../../utils/math/Math.sol";
+
+/**
+ * @dev Extension of {Governor} for fractional voting.
+ *
+ * Similar to {GovernorCountingSimple}, this contract is a votes counting module for {Governor} that supports 3 options:
+ * Against, For, Abstain. Additionally, it includes a fourth option: Fractional, which allows voters to split their voting
+ * power amongst the other 3 options.
+ *
+ * Votes cast with the Fractional support must be accompanied by a `params` argument that is tree packed `uint128` values
```
frangio, line 16:
```suggestion
 * Votes cast with the Fractional support must be accompanied by a `params` argument that is three packed `uint128` values
```

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,232 @@
...
+ * This is mostly useful when the delegate is a contract that implements its own rules for voting. These delegate-contracts
+ * can cast fractional votes according to the preferences of multiple entities delegating their voting power.
+ *
+ * Some example use cases include:
+ *
+ * * Voting from tokens that are held by a DeFi pool
+ * * Voting from an L2 with tokens held by a bridge
+ * * Voting privately from a shielded pool using zero knowledge proofs.
+ *
+ * Based on ScopeLift's GovernorCountingFractional[https://github.com/ScopeLift/flexible-voting/blob/e5de2efd1368387b840931f19f3c184c85842761/src/GovernorCountingFractional.sol]
```
frangio, line 29: Are there any important differences that people should consider if they migrate from ScopeLift's implementation to this one, or if an interface has to support both? I think it may be worth listing them here.

contracts/governance/extensions/GovernorCountingSimple.sol
frangio: Yeah I just saw that. That's good, although it should be added to the documentation for `COUNTING_MODE` in `IGovernor`.

The comma is good.

frangio: Sorry, second thoughts... I don't understand the rationale for `params`. I would've named it `support=bravo,fractional`.

frangio: Is the idea that `,params` means a value of `type(uint8).max` defers the meaning of the support to the content of params? Maybe that's good. I'll wait to see the documentation.

Amxx: Rational comes from here: https://github.com/ScopeLift/flexible-voting/blob/e5de2efd1368387b840931f19f3c184c85842761/src/GovernorCountingFractional.sol#L55

Will check with tally if they use it, and if `support=bravo,fractional` would work for them

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,232 @@
...
+        }
+        if (abstainVotes > 0) {
+            details.abstainVotes += abstainVotes;
+        }
+        details.usedVotes[account] += usedWeight;
+
+        return usedWeight;
+    }
+
+    function _extractUint128(bytes memory data, uint256 pos) private pure returns (uint128 result) {
```
frangio, line 227: This function concerns me because the length of `data` is ignored and `pos` may be reading out of bounds. The assembly block correctly omits the `memory-safe` annotation, but like I said in a recent issue this globally disables optimizations so it should be avoided.

Is it possible to just check the validity of `pos` against the data length? Sadly I do see it would make the code less efficient by checking the length three times...

contracts/governance/extensions/GovernorCountingSimple.sol
frangio: I think `support=bravo,params` is okay though!

contracts/governance/extensions/GovernorCountingFractional.sol
ernestognw: So far:

- The reference implementation is in OpenZeppelin Contracts v4
- It includes a `nonce` in the `params` object given v4 had no nonce protection
- The `COUNTING_MODDE` is different

The voting format may vary after the reviews. I think it's currently fine since it reads it's "based on", so no compatibility should be assumed from the get go

contracts/governance/extensions/GovernorCountingSimple.sol
ernestognw: Also agree with the current `COUNTING_MODE`
I'm solving since `type(uint8).max` option was used and we all seem aligned

ernestognw (dismissed):
LGTM

contracts/governance/extensions/GovernorCountingFractional.sol
Amxx: This is private, and everywhere we call it we first validate the length. Is that not good enough?

ernestognw: GIven how frequent it is for people to copy-paste from our code I think we should try to make functions self-contained and well documented even if they're private.

In this case, not only that I think this is the cleanest way of making the assembly block `memory-safe` to avoid disabling optimizations, but also I share the concern with @frangio; although the function is unreachable with invalid parameters, I don't feel comfortable having such a function around.

ernestognw:
I'd be open to change it as long as we either remove the assembly or make the block `memory-safe`.

I agree it's not an issue itself given the function is private and can't be reached with invalid arguments, it's just that I wouldn't feel comfortable given how people might use the code.

frangio:
I think it would be ok to leave the previous `_extractUint128` if we mark it as `memory-safe`, along with a comment warning that the caller must validate `pos`.

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,193 @@
...
+     * The sum total of the three decoded vote weights _must_ be less than or equal to the delegate's remaining weight
+     * on the proposal, i.e. their checkpointed total weight minus votes already cast on the proposal.
+     *
+     * NOTE: Consider that fractional voting restricts the number of casted vote (in each category) to 128 bits.
+     * Depending on how many decimals the underlying token has, a single voter may require to split their vote into
+     * multiple vote operations. For precision higher than ~30 decimals, large token holders may require an
+     * potentially large number of calls to cast all their votes. The voter has the possibility to cast all the
+     * remaining votes in a single operation using the traditional "bravo" vote.
+     */
+    // slither-disable-next-line cyclomatic-complexity
```
ernestognw, line 131: For the record, we both feel fine by keeping this function self contained

contracts/governance/extensions/GovernorCountingFractional.sol
```diff
@@ -0,0 +1,193 @@
...
+        uint256 usedWeight;
+
+        // For clarity of event indexing, fractional voting must be clearly advertised in the "support" field.
+        //
+        // Supported `support` value must be:
+        // - "Full" voting: `support = 0` (Against), `1` (For) or `2` (Abstain), with empty params.
+        // - "Fractional" voting: `support = 255`, with 48 bytes params.
+        if (support == uint8(GovernorCountingSimple.VoteType.Against)) {
+            if (params.length != 0) revert GovernorInvalidVoteParams();
+            usedWeight = againstVotes = remainingWeight;
```
ernestognw, line 157: Similarly, we agreed that this double assignment is clear enough

ernestognw (approved):

ernestognw (approved):
LFG 🚀

simplyoptimistic:
Seeing as the main benefit of this is to allow for proxy voting of locked tokens, are there any plans on adding a FlexVotingClient like the one in ScopeLift that users can use with this governor? At the moment, I want to use this, but it seems like I will have to adapt the client from ScopeLift in order to make it work (after reviewing the diff from this and theirs).

I am unsure if I should open an issue for this. 

https://github.com/ScopeLift/flexible-voting/blob/master/src/FlexVotingClient.sol

Amxx:
> I am unsure if I should open an issue for this.

Yes please, open an issue describing the behavior desired as precisely as possible.

PR #5048, opened 2024-05-19
Fix typo in Governance docs

GianfrancoBazzani:


ernestognw:
LGTM, will approve once I get the CI to pass

> @ernestognw, is there any chance that the emoji in the title is causing the CI issues ?

I removed it and it keeps failing so pretty much not the case. I can't se the logs either so I wonder if there's an issue on the Netlify side. I'll wait a bit and retry.

PR #5051, opened 2024-05-21
Procedurally generate packing libraries

ernestognw:
### Description

This PR extends the new (unreleased) Packing library to be automatically generated. It allows for packing values into `bytes32`, `bytes16`, `bytes8` and `bytes4`.

I decided to split the contract in `PackingBytes32`, `PackingBytes16`, `PackingBytes8` and `PackingBytes4`. The reason is that when function identifiers collide (i.e. there are multiple definitions of `at` and `pack` so the compiler can't dissambiguate).

The alternative is to give users instructions of how to apply the `using` syntax to only apply the functions to those types intended. This is simpler imo.

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/.changeset/heavy-baboons-give.md
+++ b/.changeset/heavy-baboons-give.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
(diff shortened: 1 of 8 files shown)


ernestognw:
Unit tests are missing. Will update later

ernestognw:
In favor of #5056

PR #5052, opened 2024-05-22
Bubble up `returndata` from reverted Create2 deployments

### LINKED ISSUE #5046: Create2 doesn't bubble up returndata on revert
When a contract deployment fails in `Create2`, returndata is ignored even though it may contain a revert reason that should probably be bubbled up like the `Address` library does for calls. I haven't checked if Solidity does this for native contract creation.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/255e27e6d22934ddaf00c7f279039142d725382d/contracts/utils/Create2.sol#L49

dimitriospapathanasiou:
Fixes #5046 

I am uploading this PR to propose a solution for issue #5046. The current implementation simply reverts with a generic FailedDeployment error, which obscures potentially useful debugging information contained in the returndata.
So in the changes, deploy function was enhanced in the following way:

- returndata is captured when a create2 deployment fails.
- If returndata is available, it is used to revert, providing the original error message from the failed deployment.
- If no returndata is available, the function reverts with the existing generic FailedDeployment error.

The changes were tested locally, and the expected output was given.

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Create2.sol
+++ b/contracts/utils/Create2.sol
@@ -42,28 +42,18 @@ library Create2 {
             revert Create2EmptyBytecode();
         }
 
-        uint256 returndataSize;
-        bytes memory returndata;
-
         /// @solidity memory-safe-assembly
         assembly {
             addr := create2(amount, add(bytecode, 0x20), mload(bytecode), salt)
-            if iszero(addr) {
-                returndataSize := returndatasize()
-                returndata := mload(0x40)
-                mstore(0x40, add(returndata, add(returndataSize, 0x20)))
-                returndatacopy(returndata, 0, returndataSize)
+            // if no address was created, and returndata is not empty, bubble revert
+            if and(iszero(addr), not(iszero(returndatasize()))) {
+                returndatacopy(0, 0, returndatasize())
+                revert(0, returndatasize())
             }
         }
 
         if (addr == address(0)) {
-            if (returndataSize > 0) {
-                assembly {
-                    revert(add(returndata, 0x20), returndataSize)
-                }
-            } else {
-                revert Errors.FailedDeployment();
-            }
+            revert Errors.FailedDeployment();
         }
     }
 
--- a/contracts/utils/Create2.sol
+++ b/contracts/utils/Create2.sol
@@ -41,7 +41,6 @@ library Create2 {
         if (bytecode.length == 0) {
             revert Create2EmptyBytecode();
         }
-
         /// @solidity memory-safe-assembly
         assembly {
             addr := create2(amount, add(bytecode, 0x20), mload(bytecode), salt)
@@ -51,7 +50,6 @@ library Create2 {
                 revert(0, returndatasize())
             }
         }
-
         if (addr == address(0)) {
             revert Errors.FailedDeployment();
         }
--- a/contracts/utils/Create2.sol
+++ b/contracts/utils/Create2.sol
@@ -41,12 +41,29 @@ library Create2 {
         if (bytecode.length == 0) {
             revert Create2EmptyBytecode();
         }
+
+        uint256 returndataSize;
+        bytes memory returndata;
+
         /// @solidity memory-safe-assembly
         assembly {
             addr := create2(amount, add(bytecode, 0x20), mload(bytecode), salt)
+            if iszero(addr) {
+                returndataSize := returndatasize()
+                returndata := mload(0x40)
+                mstore(0x40, add(returndata, add(returndataSize, 0x20)))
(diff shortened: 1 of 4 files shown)


contracts/utils/Create2.sol
```diff
@@ -41,12 +41,29 @@ library Create2 {
...
+        uint256 returndataSize;
+        bytes memory returndata;
+
         /// @solidity memory-safe-assembly
         assembly {
             addr := create2(amount, add(bytecode, 0x20), mload(bytecode), salt)
+            if iszero(addr) {
+                returndataSize := returndatasize()
+                returndata := mload(0x40)
+                mstore(0x40, add(returndata, add(returndataSize, 0x20)))
```
Amxx, line 54: returndata is going to be used for the revert. It has no long term purpose, so we can do without the memory ptr update.

contracts/utils/Create2.sol
```diff
@@ -41,12 +41,29 @@ library Create2 {
...
+                returndatacopy(returndata, 0, returndataSize)
+            }
```
Amxx, line 56:
```suggestion
            if iszero(addr) {
                returndata := mload(0x40)
                mstore(returndata, returndatasize())
                returndatacopy(add(returndata, 0x20), 0, returndatasize())
            }
```

contracts/utils/Create2.sol
```diff
@@ -41,12 +41,29 @@ library Create2 {
...
         }
+
         if (addr == address(0)) {
-            revert Errors.FailedDeployment();
+            if (returndataSize > 0) {
+                assembly {
+                    revert(add(returndata, 0x20), returndataSize)
```
Amxx, line 62:
```suggestion
                    revert(add(returndata, 0x20), mload(returndata))
```

Amxx:
Thank you @dimitriospapathanasiou for this PR.

I like the idea ... but I'm not sure I like the current implementation.

contracts/utils/Create2.sol
```diff
@@ -41,12 +41,29 @@ library Create2 {
...
+                }
+            } else {
+                revert Errors.FailedDeployment();
+            }
```
Amxx, line 66:
```suggestion
            revert Errors.FailedDeployment();
```

contracts/utils/Create2.sol
```diff
@@ -41,12 +41,29 @@ library Create2 {
         if (bytecode.length == 0) {
             revert Create2EmptyBytecode();
         }
+
+        uint256 returndataSize;
+        bytes memory returndata;
+
```
Amxx, line 47:
```suggestion
```

ernestognw (commented):
LGTM. The CI should be fine after we merge #5055.

I shared some thoughts in the original issue. Although I'm approving, it may make sense to reconsider.

ernestognw (approved):

Amxx (approved):

contracts/utils/Create2.sol
```diff
@@ -44,6 +44,11 @@ library Create2 {
         /// @solidity memory-safe-assembly
         assembly {
             addr := create2(amount, add(bytecode, 0x20), mload(bytecode), salt)
+            // if no address was created, and returndata is not empty, bubble revert
+            if and(iszero(addr), not(iszero(returndatasize()))) {
+                returndatacopy(0, 0, returndatasize())
+                revert(0, returndatasize())
```
frangio, line 50: This snippet is explicitly mentioned in Solidity documentation as something that is not memory-safe.

https://docs.soliditylang.org/en/latest/assembly.html#memory-safety

cc @Amxx @ernestognw

frangio: See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5057

ernestognw: Been thinking about the consequences of this, but does it even matter given it's reverting? It's "memory unsafe" because it may overwrite memory but there's no way to do something unsafe afterwards.

frangio: Yeah the docs say it matters if the assembly block reverts and this exact snippet is used to illustrate the point.

frangio: image

ernestognw: Yeah, it's kinda unclear what's the relationship between being "mainly about the optimizer" and still needing to ensure memory safe

frangio: Yeah I should look into this for my next blog post to give you a reason to read it. 😁

PR #5057, opened 2024-05-27
Use free memory pointer for returndatacopy

frangio:
Fixing memory safety as explained here: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5052/files#r1616302783

The Solidity docs state:

> these restrictions still need to be followed, even if the assembly block reverts or terminates.


frangio:
Does this need a changeset?

ernestognw (approved):
> Does this need a changeset?

I don't think so

Amxx (approved):

PR #5058, opened 2024-05-27
Fix invalid memory-safe annotation

Amxx:
See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5052#discussion_r1616302783

### DIFF
--- a/contracts/utils/Create2.sol
+++ b/contracts/utils/Create2.sol
@@ -41,7 +41,6 @@ library Create2 {
         if (bytecode.length == 0) {
             revert Create2EmptyBytecode();
         }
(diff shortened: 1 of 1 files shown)


frangio:
Removing the annotation is not good because it globally disables optimizations in any contract that includes this code. It's better to make the code memory safe as in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5057

frangio:
In Proxy.sol it would be interesting to test the performance of making the block memory-safe by using the free memory pointer instead of writing to memory pointer 0. Although I think the optimization consequences only apply in IR mode.

PR #5067, opened 2024-06-03
Add utility function for converting an address to checksummed string

### LINKED ISSUE #4633: Util function for converting an address to checksummed string
🧐 **Motivation**
Some applications might need to have the checksummed address be converted to string. In my usage for instance, I need to send some tokens to an address whose hash was committed earlier, so I need the checksummed string.

📝 **Details**
Implement EIP-55 to get the checksummed string corresponding to the address. (I have already implementated for my application on finding no short cut online. More people may need this.)

Some old discussion on this:
https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2197#issuecomment-631005323

cairoeth:
Closes #4633 

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -97,7 +97,6 @@ library Strings {
         bytes memory buffer = new bytes(42);
         buffer[0] = "0";
         buffer[1] = "x";
-        uint160 addrValue = uint160(addr);
         uint160 hashValue = uint160(bytes20(hashedAddr));
         for (uint256 i = 41; i > 1; --i) {
             if (hashValue & 0xf > 7) {
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -64,17 +64,15 @@ library Strings {
      * @dev Converts a `uint256` to its ASCII `string` hexadecimal representation with fixed length.
      */
     function toHexString(uint256 value, uint256 length) internal pure returns (string memory) {
-        uint256 localValue = value;
+        if (length < Math.log256(value) + 1) {
+            revert StringsInsufficientHexLength(value, length);
+        }
+
         bytes memory buffer = new bytes(2 * length + 2);
         buffer[0] = "0";
         buffer[1] = "x";
-        for (uint256 i = 2 * length + 1; i > 1; --i) {
-            buffer[i] = HEX_DIGITS[localValue & 0xf];
-            localValue >>= 4;
-        }
-        if (localValue != 0) {
-            revert StringsInsufficientHexLength(value, length);
-        }
+        _setHexString(buffer, 2, value);
+
         return string(buffer);
     }
 
@@ -91,12 +89,8 @@ library Strings {
      * representation, according to EIP-55.
      */
     function toChecksumHexString(address addr) internal pure returns (string memory) {
-        uint160 localValue = uint160(addr);
         bytes memory lowercase = new bytes(40);
-        for (uint256 i = 40; i > 0; --i) {
-            lowercase[i - 1] = HEX_DIGITS[localValue & 0xf];
-            localValue >>= 4;
-        }
+        _setHexString(lowercase, 0, uint160(addr));
         bytes32 hashedAddr = keccak256(abi.encodePacked(lowercase));
 
         bytes memory buffer = new bytes(42);
@@ -121,5 +115,15 @@ library Strings {
      */
     function equal(string memory a, string memory b) internal pure returns (bool) {
         return bytes(a).length == bytes(b).length && keccak256(bytes(a)) == keccak256(bytes(b));
+    }
+
+    /**
+     * @dev Sets the hexadecimal representation of a value in the specified buffer starting from the given offset.
+     */
+    function _setHexString(bytes memory buffer, uint256 offset, uint256 value) private pure {
+        for (uint256 i = buffer.length; i > offset; --i) {
+            buffer[i - 1] = HEX_DIGITS[value & 0xf];
+            value >>= 4;
+        }
     }
 }
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -71,7 +71,7 @@ library Strings {
         bytes memory buffer = new bytes(2 * length + 2);
         buffer[0] = "0";
         buffer[1] = "x";
-        _setHexString(buffer, 2, value);
+        _unsafeSetHexString(buffer, 2, value);
 
         return string(buffer);
     }
@@ -90,8 +90,8 @@ library Strings {
      */
     function toChecksumHexString(address addr) internal pure returns (string memory) {
         bytes memory lowercase = new bytes(40);
-        uint160 addrValue = uint160(address)
-        _setHexString(lowercase, 0, addrValue);
+        uint160 addrValue = uint160(addr);
+        _unsafeSetHexString(lowercase, 0, addrValue);
         bytes32 hashedAddr = keccak256(abi.encodePacked(lowercase));
 
         bytes memory buffer = new bytes(42);
@@ -99,11 +99,8 @@ library Strings {
         buffer[1] = "x";
         uint160 hashValue = uint160(bytes20(hashedAddr));
         for (uint256 i = 41; i > 1; --i) {
-            if (hashValue & 0xf > 7) {
-                buffer[i] = HEX_DIGITS_CAPITAL[addrValue & 0xf];
-            } else {
-                buffer[i] = HEX_DIGITS[addrValue & 0xf];
-            }
+            uint8 digit = uint8(addrValue & 0xf);
+            buffer[i] = hashValue & 0xf > 7 ? HEX_DIGITS_CAPITAL[digit] : HEX_DIGITS[digit];
             addrValue >>= 4;
             hashValue >>= 4;
         }
@@ -119,8 +116,11 @@ library Strings {
 
     /**
      * @dev Sets the hexadecimal representation of a value in the specified buffer starting from the given offset.
+     *
+     * NOTE: This function does not check that the `buffer` can allocate `value` without overflowing. Make sure
+     * to check whether `Math.log256(value) + 1` is larger than the specified `length`.
      */
-    function _setHexString(bytes memory buffer, uint256 offset, uint256 value) private pure {
+    function _unsafeSetHexString(bytes memory buffer, uint256 offset, uint256 value) private pure {
         for (uint256 i = buffer.length; i > offset; --i) {
             buffer[i - 1] = HEX_DIGITS[value & 0xf];
             value >>= 4;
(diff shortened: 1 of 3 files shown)


cairoeth:
There's some re-used logic [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5067/files#diff-e402541b45f70ec8d3de9b9f87542d9eb52920601ac011a340e723c5b4b9d659R94-R99) from `toHexString` but the main difference is removing the `0x` from the string. `toHexString` adds `0x` to the buffer by default, so to use this function it would require removing it form the returned string which would be extra computation just for reusability.

ernestognw (commented):
I've been trying to come up with a way of reusing the logic of `toHexString` and I noticed some old behaviors we've kept in the library like providing the expected `length` of the hex string (which we already know by `Math.log256(value) + 1`).

Perhaps this PR derives in a broader discussion about how to implement this, but I'd say the simplest option is to rewrite `toHexString` in the following way:

```solidity
function toHexString(uint256 value, uint256 length) internal pure returns (string memory) {
    if (length < Math.log256(value) + 1) {
        revert StringsInsufficientHexLength(value, length);
    }

    bytes memory buffer = new bytes(2 * length + 2);
    buffer[0] = "0";
    buffer[1] = "x";
    _setHexString(buffer, 2, value);

    return string(buffer);
}

function _setHexString(bytes memory buffer, uint256 offset, uint256 value) internal pure {
    for (uint256 i = buffer.length - 1; i >= offset; --i) {
        buffer[i] = HEX_DIGITS[value & 0xf];
        value >>= 4;
    }
}
``` 

This way, we can reuse `_setHexString` in the new function. Similar to:

```solidity
function toChecksumHexString(address addr) internal pure returns (string memory) {
    bytes memory lowercase = new bytes(40);
    _setHexString(lowercase, 0, uint160(address);
    bytes32 hashedAddr = keccak256(abi.encodePacked(lowercase));
    ...
}
```

What do you think?

cairoeth:
@ernestognw Yes, I was thinking something like that as well -- should we should make `_setHexString` private as it's not designed to be used outside the library (mainly because it doesn't have any checks on value after the operation)?

contracts/utils/Strings.sol
```diff
@@ -85,10 +84,46 @@ library Strings {
         return toHexString(uint256(uint160(addr)), ADDRESS_LENGTH);
     }
 
+    /**
+     * @dev Converts an `address` with fixed length of 20 bytes to its checksummed ASCII `string` hexadecimal
+     * representation, according to EIP-55.
+     */
+    function toChecksumHexString(address addr) internal pure returns (string memory) {
+        bytes memory lowercase = new bytes(40);
+        _setHexString(lowercase, 0, uint160(addr));
```
ernestognw, line 93: We have the `uint160(addr)` twice. This should be better for both readability and perhaps saving one `PUSH` here and there.

```suggestion
        uint160 addrValue = uint160(address)
        _setHexString(lowercase, 0, addrValue);
```

cairoeth: good catch

contracts/utils/Strings.sol
```diff
@@ -11,6 +11,7 @@ import {SignedMath} from "./math/SignedMath.sol";
  */
 library Strings {
     bytes16 private constant HEX_DIGITS = "0123456789abcdef";
+    bytes16 private constant HEX_DIGITS_CAPITAL = "0123456789ABCDEF";
```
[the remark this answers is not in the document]
cairoeth: i think they are synonyms, but makes more sense to refer to uppercase to a sentence/word instead of the first letter.

contracts/utils/Strings.sol
```diff
@@ -85,10 +84,46 @@ library Strings {
...
+        buffer[0] = "0";
+        buffer[1] = "x";
+        uint160 addrValue = uint160(addr);
+        uint160 hashValue = uint160(bytes20(hashedAddr));
+        for (uint256 i = 41; i > 1; --i) {
+            if (hashValue & 0xf > 7) {
+                buffer[i] = HEX_DIGITS_CAPITAL[addrValue & 0xf];
+            } else {
+                buffer[i] = HEX_DIGITS[addrValue & 0xf];
+            }
```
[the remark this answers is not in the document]
cairoeth: ah yes, much cleaner now

test/utils/Strings.test.js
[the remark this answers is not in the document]
cairoeth: totally -- fuzzing should be used in some of the other utils as well.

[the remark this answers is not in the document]
ernestognw: Agree, feel free to open PRs adding fuzzing or Halmos FV to those utils you consider make sense

test/utils/Strings.test.js
```diff
@@ -120,6 +120,18 @@ describe('Strings', function () {
...
+    it('converts a random address', async function () {
+      const addr = '0xa9036907dccae6a1e0033479b12e837e5cf5a02f';
+      expect(await this.mock.getFunction('$toChecksumHexString(address)')(addr)).to.equal(ethers.getAddress(addr));
+    });
+
+    it('converts an address with leading zeros', async function () {
+      const addr = '0x0000e0ca771e21bd00057f54a68c30d400000000';
+      expect(await this.mock.getFunction('$toChecksumHexString(address)')(addr)).to.equal(ethers.getAddress(addr));
+    });
+  });
```
ernestognw, line 133: There's a little consideration in the [documentation of `getAddress`](https://docs.ethers.org/v6/api/address/#getAddress):

> *  If %%address%% contains both upper-case and lower-case, it is
> *  assumed to already be a checksum address and its checksum is
> *  validated, and if the address fails its expected checksum an
> *  error is thrown.

None of these tests are using mixed-case letters so I'd recommend adding `.toLowerCase()` according to the same docs:

> * If you wish the checksum of %%address%% to be ignore, it should
> *  be converted to lower-case (i.e. ``.toLowercase()``) before
> *  being passed in.

Even better, let's rewrite these tests:

```
const addresses = [...]

describe('toChecksumHexString address', function () {
  for (const addr of addresses) {
    it(`converts ${addr}`, async function () {
      expect(await this.mock.getFunction('$toChecksumHexString(address)')(addr.toLowerCase())).to.equal(
        ethers.getAddress(addr),
      );
    });
  }
});
```

I'm pushing a commit

ernestognw (dismissed):
LGTM, we'll need an approval from @Amxx imo

contracts/utils/Strings.sol
```diff
@@ -63,17 +64,15 @@ library Strings {
      * @dev Converts a `uint256` to its ASCII `string` hexadecimal representation with fixed length.
      */
     function toHexString(uint256 value, uint256 length) internal pure returns (string memory) {
-        uint256 localValue = value;
+        if (length < Math.log256(value) + 1) {
```
Amxx, line 67: computing the log is quite expensive. Checking at the end is less expensive (when everything works fine, which is the case we should optimize for). Any reason to not keep it?

contracts/utils/Strings.sol
```diff
@@ -85,10 +84,46 @@ library Strings {
...
+
+    /**
+     * @dev Sets the hexadecimal representation of a value in the specified buffer starting from the given offset.
+     *
+     * NOTE: This function does not check that the `buffer` can allocate `value` without overflowing. Make sure
+     * to check whether `Math.log256(value) + 1` is larger than the specified `length`.
+     */
+    function _unsafeSetHexString(bytes memory buffer, uint256 offset, uint256 value) private pure {
+        for (uint256 i = buffer.length; i > offset; --i) {
+            buffer[i - 1] = HEX_DIGITS[value & 0xf];
```
Amxx, line 125: I'm concerned each write may check that `i-1` is within the bound of the array. If that is the case, its a major waste. Maybe we could optimize that.

contracts/utils/Strings.sol
```diff
@@ -85,10 +84,46 @@ library Strings {
...
 
+    /**
+     * @dev Converts an `address` with fixed length of 20 bytes to its checksummed ASCII `string` hexadecimal
+     * representation, according to EIP-55.
+     */
+    function toChecksumHexString(address addr) internal pure returns (string memory) {
+        bytes memory lowercase = new bytes(40);
+        uint160 addrValue = uint160(addr);
+        _unsafeSetHexString(lowercase, 0, addrValue);
+        bytes32 hashedAddr = keccak256(abi.encodePacked(lowercase));
```
Amxx, line 95: You often to `abi.encodePacked(x)` when x is already a `bytes`. (here and on line 107)

Used like this, `abi.encodePacked` is the identity function ... the output exactly match the input ... so it is not necessary. One thing that happens however, is that memory is allocated for this copy... there is also a copy loop (or a mcopy if we are lucky).

Anyway, this increasse costs and leaks memory, so we should avoid it!

contracts/utils/Strings.sol
```diff
@@ -85,10 +84,46 @@ library Strings {
...
+        buffer[1] = "x";
+        uint160 hashValue = uint160(bytes20(hashedAddr));
+        for (uint256 i = 41; i > 1; --i) {
+            uint8 digit = uint8(addrValue & 0xf);
+            buffer[i] = hashValue & 0xf > 7 ? HEX_DIGITS_UPPERCASE[digit] : HEX_DIGITS[digit];
+            addrValue >>= 4;
+            hashValue >>= 4;
+        }
+        return string(abi.encodePacked(buffer));
+    }
```
Amxx, line 108: This entier function should be node "in place". Allocating two buffers is a waste.

contracts/utils/Strings.sol
Amxx, line 1: Redesigned `toChecksumHexString` to avoid double allocation.
- removed the need for `_unsafeSetHexString`
- removed the need for `HEX_DIGITS_UPPERCASE`

contracts/utils/Strings.sol
```diff
@@ -85,6 +85,30 @@ library Strings {
...
+    /**
+     * @dev Converts an `address` with fixed length of 20 bytes to its checksummed ASCII `string` hexadecimal
+     * representation, according to EIP-55.
+     */
+    function toChecksumHexString(address addr) internal pure returns (string memory) {
+        bytes memory buffer = bytes(toHexString(addr));
+
+        // hash the hex part of buffer (skip length + 2 bytes, length 40)
+        uint256 hashValue;
+        assembly ("memory-safe") {
```
Amxx, line 97: note: this is safe because we know buffer is 42 bytes long.

contracts/utils/Strings.sol
ernestognw: Right this is extremely cleaner. Thanks!

ernestognw (approved):
LGTM again

contracts/utils/Strings.sol
cairoeth: great changes, thxs!

Amxx (approved):

PR #5068, opened 2024-06-06
use arrays.unsafeMemoryAccess in ERC1155Supply

RenanSouza2:
This pr makes ERC1155Supply use the same pattern to access arrays as the rest of the ERC1155 contracts 

#### PR Checklist

- [X] Tests
- [X] Documentation

### DIFF
--- a/contracts/token/ERC1155/extensions/ERC1155Supply.sol
+++ b/contracts/token/ERC1155/extensions/ERC1155Supply.sol
@@ -4,6 +4,7 @@
 pragma solidity ^0.8.20;
 
 import {ERC1155} from "../ERC1155.sol";
+import {Arrays} from "../../../utils/Arrays.sol";
 
 /**
  * @dev Extension of ERC-1155 that adds tracking of total supply per id.
@@ -19,6 +20,8 @@ import {ERC1155} from "../ERC1155.sol";
  * CAUTION: This extension should not be added in an upgrade to an already deployed contract.
  */
 abstract contract ERC1155Supply is ERC1155 {
+    using Arrays for uint256[];
+
     mapping(uint256 id => uint256) private _totalSupply;
     uint256 private _totalSupplyAll;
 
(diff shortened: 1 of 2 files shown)


Amxx:
Thank you @RenanSouza2 for this proposal

Gas optimization don't need a changelog entry.

contracts/token/ERC1155/extensions/ERC1155Supply.sol
```diff
@@ -57,9 +60,9 @@ abstract contract ERC1155Supply is ERC1155 {
         if (from == address(0)) {
             uint256 totalMintValue = 0;
             for (uint256 i = 0; i < ids.length; ++i) {
-                uint256 value = values[i];
+                uint256 value = values.unsafeMemoryAccess(i);
```
Amxx, line 63: This assumes `values.length == ids.length`. It is something that is checked in `ERC1155._update`, but that may be lost if another override of `_update` (between this and `ERC1155` in the linearized list or contracts) fails to call `super`.

I'm not too concerned about that.

Amxx (approved):
LGTM.

@ernestognw ?

ernestognw (approved):
LGTM

contracts/token/ERC1155/extensions/ERC1155Supply.sol
Amxx: Well, if we merge that change it will not panic. 
This PR removes the check that would cause the panic to happen

ernestognw: Right, got confused.
Then what do you think of adding that requirement in the fn natspec?

cairoeth (approved):
lgtm

contracts/token/ERC1155/extensions/ERC1155Supply.sol
ernestognw: Thought about it and concluded it's not necessary since it's already checked in ERC1155 and we're already warning people about overrides without calling `super`.

Solving

PR #5069, opened 2024-06-06
Fix dirty bits in upper bits in implementation address in `Clones.sol`

Amxx:
Revert 8b2f29ceb0db64957a49deb63af7792772bd7a5d (#4927) that introduced vulnerability to dirty upper bits in the implementation address

#### PR Checklist

- [x] Tests

### DIFF
--- a/test/proxy/Clones.t.sol
+++ b/test/proxy/Clones.t.sol
@@ -2,10 +2,52 @@
 
 pragma solidity ^0.8.20;
 
-import {Test} from "forge-std/Test.sol";
+import {Test, console2} from "forge-std/Test.sol";
 import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
 
+contract DummyContract {
+    function getNumber() external pure returns (uint256) {
+        return 42;
+    }
+}
+
+contract ClonesMock {
+    function cloneContract(address target) external returns (address) {
+        return Clones.clone(target);
+    }
+
+    function cloneDirtyTarget(address target) external returns (address) {
+        // Add dirty bits; will start with (0xff..)
+        assembly {
+            // Get type(uint256).max and shift to the left by 20 bytes => `or` with target
+            target := or(shl(160, not(0)), target)
+        }
+        return Clones.clone(target);
+    }
+
+    function cloneContractDeterministic(address target) external returns (address) {
+        return Clones.clone(target);
+    }
+
+    function cloneDirtyTargetDeterministic(address target) external returns (address) {
+        // Add dirty bits; will start with (0xff..)
+        assembly {
+            // Get type(uint256).max and shift to the left by 20 bytes => `or` with target
+            target := or(shl(160, not(0)), target)
+        }
+        return Clones.clone(target);
+    }
+}
+
 contract ClonesTest is Test {
+    DummyContract private dummy;
+    ClonesMock private clonesWrapper;
+
+    function setUp() public {
+        dummy = new DummyContract();
+        clonesWrapper = new ClonesMock();
+    }
+
     function testSymbolicPredictDeterministicAddressSpillage(address implementation, bytes32 salt) public {
         address predicted = Clones.predictDeterministicAddress(implementation, salt);
         bytes32 spillage;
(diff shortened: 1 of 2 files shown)


test/proxy/Clones.t.sol
```diff
@@ -2,10 +2,54 @@
...
+        // Add dirty bits; will start with (0xff..)
+        assembly {
+            // Get type(uint256).max and shift to the left by 20 bytes => `or` with target
+            target := or(shl(160, not(0)), target)
+        }
+        return Clones.clone(target);
+    }
+
+    function cloneContractDeterministic(address target) external returns (address) {
+        return Clones.clone(target);
```
Amxx, line 31: that is not deterministic

test/proxy/Clones.t.sol
```diff
@@ -15,4 +19,42 @@ contract ClonesTest is Test {
...
+    function testPredictDeterministicAddressDirty(bytes32 salt) external {
+        address predictClean = Clones.predictDeterministicAddress(address(this), salt);
+        address predictDirty = Clones.predictDeterministicAddress(_dirty(address(this)), salt);
+
+        //prediction should be similar
+        assertEq(predictClean, predictDirty);
+    }
+
+    function _dirty(address input) private pure returns (address output) {
+        assembly {
```
ernestognw, line 56:
```suggestion
        assembly("memory-safe") {
```

test/proxy/Clones.t.sol
```diff
@@ -15,4 +19,42 @@ contract ClonesTest is Test {
...
+        assertEq(keccak256(cloneClean.code), keccak256(cloneDirty.code));
+
+        // both clones behave as expected
+        assertEq(ClonesTest(cloneClean).getNumber(), this.getNumber());
+        assertEq(ClonesTest(cloneDirty).getNumber(), this.getNumber());
+    }
+
+    function testCloneDeterministicDirty() external {
+        address cloneClean = Clones.cloneDeterministic(address(this), bytes32(block.prevrandao));
+        address cloneDirty = Clones.cloneDeterministic(_dirty(address(this)), bytes32(~block.prevrandao));
```
ernestognw, line 37: Why not fuzzing this?

```suggestion
    function testCloneDeterministicDirty(bytes32 salt) external {
        address cloneClean = Clones.cloneDeterministic(address(this), salt);
        address cloneDirty = Clones.cloneDeterministic(_dirty(address(this)), ~salt);
```

test/proxy/Clones.t.sol
```diff
@@ -15,4 +19,42 @@ contract ClonesTest is Test {
...
+        assertEq(ClonesTest(cloneDirty).getNumber(), this.getNumber());
+    }
+
+    function testPredictDeterministicAddressDirty(bytes32 salt) external {
+        address predictClean = Clones.predictDeterministicAddress(address(this), salt);
+        address predictDirty = Clones.predictDeterministicAddress(_dirty(address(this)), salt);
+
+        //prediction should be similar
+        assertEq(predictClean, predictDirty);
+    }
```
ernestognw, line 53: What is this testing? Since the address is still of `address` type it'll pass the test even if dirty.
I have the sense this is testing the same as `testSymbolicPredictDeterministicAddressSpillage`.

ernestognw:
Tried using `testSymbolic` to formally verify it with halmos but turns out `0x3c` is an unsupported opcode.
Fuzzing is enough here imo

test/proxy/Clones.t.sol
Amxx: This tests that `predictDeterministicAddress` returns the same value if the input address (implementation) is dirty.

test/proxy/Clones.t.sol
```diff
@@ -20,9 +20,11 @@ contract ClonesTest is Test {
         assertEq(spillage, bytes32(0));
     }
 
-    function testCloneDirty() external {
+    function testCloneDirty(address caller) external {
+        vm.startPrank(caller);
```
Amxx, line 24: This has no effect.

prank changes the address of `msg.sender` for outgoing calls. Here we are doing an internal library call and prank has no effect on it.

test/proxy/Clones.t.sol
```diff
@@ -15,4 +19,44 @@ contract ClonesTest is Test {
         }
         assertEq(spillage, bytes32(0));
     }
+
+    function testCloneDirty(address caller) external {
+        vm.startPrank(caller);
+        address cloneClean = Clones.clone(address(this));
+        address cloneDirty = Clones.clone(_dirty(address(this)));
+        vm.stopPrank();
```
Amxx, line 27:
```suggestion
    function testCloneDirty() external {
        address cloneClean = Clones.clone(address(this));
        address cloneDirty = Clones.clone(_dirty(address(this)));
```

test/proxy/Clones.t.sol
ernestognw: Are you sure? The `clone` function uses `create` under the hood changing the resulting address.
I was thinking the test should prove that for any address, even if it's dirty, there's no side effect. I'm open to remove it.

Amxx: I'm sure of that part.

- `testCloneDirty` is a function that is executed bith some msg.sender
- doing `vm.startPrank(caller)` means that any outgoing call will see `caller` as their `msg.sender` (instead of `address(this)`
- then we do `Clone.clone` that is an internal function in a library, so basically its code is part of `address(this)`'s code, and it is executed in `address(this)` context.

Note that when a contract does a create, the caller (msg.sender) is irrelevant anyway.

ernestognw: > Note that when a contract does a create, the caller (msg.sender) is irrelevant anyway.

Yes! Thanks for the clear explanation, you're right and this doesn't add any value as CREATE will result in the same address regardless

test/proxy/Clones.t.sol
ernestognw: Right, the other one is just testing the spillage. Clearer now, solving

ernestognw (approved):
LGTM

PR #5070, opened 2024-06-06
Gas optimize processMultiProof function

dimitriospapathanasiou:
This PR includes a minor change which could be considered to potentially gas optimize the processMultiProof function in MerkleProof.sol file.

In the proposed changes we are accessing values from the hashes array directly using hashes[i - leavesLen] instead of maintaining a separate counter (hashPos). This change aims to reduce gas by eliminating the need for an additional counter variable and conditional checks.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/sour-terms-fetch.md
+++ b/.changeset/sour-terms-fetch.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`MerkleProof`: Gas-optimization through counter removal.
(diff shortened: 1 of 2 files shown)


Amxx:
Hello @dimitriospapathanasiou 

This PR assumes that `hashPos = i - leavesLen`. I don't understand where that assumption comes from.

A few things to note: 
- `i - leavesLen` will often overflow, causing a panic
- `i - leavesLen` increasses by exactly 1 for each iterration. However, `hashPos` may increase by 0, 1 or 2, depending on a many factors, at each iteration.

Unsurprisingly,  the test are broken (we see the panic code corresponding to the overflow).

dimitriospapathanasiou:
Thank you for the insightful review. I understood hashes[i - leavesLen] refers to accessing the appropriate hash from the hashes array, pointing to the position determined by hashPos. But I see how they can diverge based on the iterations and conditions.

I am sorry for the misunderstanding, and thank you for the clarification.

PR #5076, opened 2024-06-11
Add `MaxHeap`

### LINKED ISSUE #3410: Add priority queue as a data structure to be able to implement max/min heaps

cairoeth:
Fixes #3410 — WIP 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/structs/MaxHeap.sol
+++ b/contracts/utils/structs/MaxHeap.sol
@@ -1,0 +1,162 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/**
+ *  TODO:
+ *  - optimizations
+ *  - changeset
+ *  - add tests
+ *  - add docs
+ *  - add base impl link here
+ *
+ * @dev Library for managing a max heap data structure.
+ *
+ * A max heap is a complete binary tree where each node has a value greater than or equal to its children.
+ * The root node contains the maximum value in the heap.
+ *
+ * This library provides functions to insert, update, and remove elements from the max heap, as well as to
+ * retrieve the maximum element (peek) and check the validity of the heap.
+ *
+ * The max heap has the following properties:
+ *
+ * - Insertion: O(log n)
+ * - Deletion of maximum element: O(log n)
+ * - Retrieval of maximum element (peek): O(1)
+ * - Update of an element: O(log n)
+ *
+ * The max heap is implemented using two mappings:
+ * - `tree`: Maps the position in the heap to the item ID.
+ * - `items`: Maps the item ID to its corresponding `Node` struct, which contains the value and heap index.
+ *
+ * Example usage:
+ *
+ * ```solidity
+ * contract Example {
+ *     using MaxHeap for MaxHeap.MaxHeap;
+ *
+ *     MaxHeap.MaxHeap private heap;
+ *
+ *     function addItem(uint256 itemId, uint256 value) public {
+ *         heap.insert(itemId, value);
+ *     }
+ *
+ *     function removeMax() public returns (uint256, uint256) {
+ *         return heap.pop();
+ *     }
+ *
+ *     function getMax() public view returns (uint256, uint256) {
+ *         return heap.peek();
+ *     }
+ * }
+ * ```
+ */
+library MaxHeap {
+    /**
+     * @dev The position doesn't have a _parent as it's the root.
+     */
+    error InvalidPositionZero();
+
+    struct Node {
+        uint256 value;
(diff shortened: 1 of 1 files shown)


contracts/utils/structs/MaxHeap.sol
```diff
@@ -0,0 +1,162 @@
...
+library MaxHeap {
+    /**
+     * @dev The position doesn't have a _parent as it's the root.
+     */
+    error InvalidPositionZero();
+
+    struct Node {
+        uint256 value;
+        uint256 heapIndex;
+    }
```
Amxx, line 63: Maybe we can do something like:
- assuming the `heapIndex` fits into something smaller. I'm not expecting any onchain structure to have more than `type(uint32).max` elements. Maybe just `type(uint16).max` is enough (that is 65k elements).
- limit the value to uint224

So that `Node` fits into a single slot.

contracts/utils/structs/MaxHeap.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    struct MaxHeap {
+        mapping(uint256 => uint256) tree;
+        mapping(uint256 => Node) items;
```
Amxx, line 67: I'm expecting one of these mapping (maybe both) can be implemented using array. That would result in better data locality when we move to verkle.

Amxx: also, using array with a value smaller than uint256 (for `tree`) would compress the data (shared slots)

contracts/utils/structs/MaxHeap.sol
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    function _parent(uint256 pos) private pure returns (uint256) {
+        if (pos == 0) revert InvalidPositionZero();
+        return (pos - 1) / 2;
+    }
+
+    function _swap(MaxHeap storage heap, uint256 fpos, uint256 spos) private {
+        (heap.tree[fpos], heap.tree[spos]) = (heap.tree[spos], heap.tree[fpos]);
+        (heap.items[heap.tree[fpos]].heapIndex, heap.items[heap.tree[spos]].heapIndex) = (fpos, spos);
```
Amxx, line 78:
```suggestion
        (
            heap.tree[fpos], heap.items[heap.tree[fpos]].heapIndex,
            heap.tree[spos], heap.items[heap.tree[spos]].heapIndex
        ) = (
            heap.tree[spos], spos, 
            heap.tree[fpos], fpos
        );
```

contracts/utils/structs/MaxHeap.sol
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    function heapify(MaxHeap storage heap, uint256 pos) internal {
+        if (pos >= (heap.size / 2) && pos <= heap.size) return;
```
Amxx, line 82: I think `pos <= heap.size` should be `pos < heap.size`. 
Also `pos >= heap.size` should be an error.

contracts/utils/structs/MaxHeap.sol
```diff
@@ -0,0 +1,162 @@
...
+     * @dev The position doesn't have a _parent as it's the root.
+     */
+    error InvalidPositionZero();
+
+    struct Node {
+        uint256 value;
+        uint256 heapIndex;
+    }
+
+    struct MaxHeap {
```
Amxx, line 65: Random though: maybe we want a Heap, instead of a MaxHeap, with a function pointer for the comparator.

contracts/utils/structs/MaxHeap.sol
```diff
@@ -0,0 +1,162 @@
...
+                _swap(heap, pos, left);
+                heapify(heap, left);
+            } else {
+                _swap(heap, pos, right);
+                heapify(heap, right);
+            }
+        }
+    }
+
+    function insert(MaxHeap storage heap, uint256 itemId, uint256 value) internal {
```
Amxx, line 102: My understand is that itemId must be unique, so that all items are clearly identified ... but is then useless to the user. Maybe it should be automatically generated ... so that the user doesn't have to worry about it, and can just push a value.

cairoeth:
Closing in favor of improved implementation at #5084

PR #5077, opened 2024-06-12
Add bytes6 to the Packing library.

Amxx:
In ERC-4337, the validation data is a uint256 that is composed of:
- an address: `aggregator`
- a uint48: `validUntil`
- a uint48: `validAfter`

This means `bytes20` + `bytes6` + `bytes6`.

This justifies that we have `bytes6` to the length supported by the packing library.

#### PR Checklist

- [x] Tests

### DIFF
--- a/contracts/utils/Packing.sol
+++ b/contracts/utils/Packing.sol
@@ -45,6 +45,24 @@ library Packing {
         }
     }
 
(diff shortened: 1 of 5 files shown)


ernestognw (approved):
Thanks!

PR #5078, opened 2024-06-12
Store AccessManager's executionId in transient storage

### LINKED ISSUE #4993: Consider using transient storage in Governor and AccessManager
**📝 Details**

Similar to #4888, but regarding the following variables:

- [`_governanceCall`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/Governor.sol#L55) in Governor
- [`executionId`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3e6c86392c97fbc30d3d20a378a6f58beba08eba/contracts/access/manager/AccessManager.sol#L109) in AccessManager

The Governor call is a Bytes32Deque, which makes me wonder if we may add a transient version of some libraries.
Both are technically breaking changes, but I don't see how they would affect the contract given the values are private.

Amxx:
Partially fixes #4993

Note: This will make the AccessManager contract incompatible with blockchains that do not implement EIP-1153. That is an  issue IMO. Maybe we should wait before we commit to that.

Also, testing should be fixed.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -8,6 +8,7 @@ import {IAccessManaged} from "./IAccessManaged.sol";
 import {Address} from "../../utils/Address.sol";
 import {Context} from "../../utils/Context.sol";
 import {Multicall} from "../../utils/Multicall.sol";
+import {StorageSlot} from "../../utils/StorageSlot.sol";
 import {Math} from "../../utils/math/Math.sol";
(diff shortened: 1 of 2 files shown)


gonzaotc:
Alternative using solidity transient keyword; https://github.com/OpenZeppelin/openzeppelin-contracts/pull/6660

PR #5080, opened 2024-06-12
Avoid storing hashing function pointers in storage make MerkleTree structure upgrade-safe

### LINKED ISSUE #5079: Resolve issue with function pointers in MerkleTree not being upgrade safe

Amxx:
Fixes #5079

This is change to an unreleased feature, no need for a changeset entry.

#### PR Checklist

- [x] Tests

### DIFF
--- a/contracts/utils/structs/MerkleTree.sol
+++ b/contracts/utils/structs/MerkleTree.sol
@@ -52,25 +52,24 @@ library MerkleTree {
      *
      * Calling this function on MerkleTree that was already setup and used will reset it to a blank state.
      *
+     * Once a tree is setup, any push to it must use the same hashing function. This means that values
+     * should be pushed to it using the default {xref-MerkleTree-push-struct-MerkleTree-Bytes32PushTree-bytes32-}[push] function.
+     *
      * IMPORTANT: The zero value should be carefully chosen since it will be stored in the tree representing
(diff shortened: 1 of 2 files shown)


contracts/utils/structs/MerkleTree.sol
```diff
@@ -96,11 +102,31 @@ library MerkleTree {
      *
      * Hashing the leaf before calling this function is recommended as a protection against
      * second pre-image attacks.
+     *
+     * This variant uses {Hashes-commutativeKeccak256} to hash internal nodes. It should only be used on merkle trees
+     * setup using the same (default) ahshing function.
```
cairoeth, line 107:
```suggestion
     * This variant uses {Hashes-commutativeKeccak256} to hash internal nodes. It should only be used on merkle trees
     * that were setup using the same (default) hashing function.
```

ernestognw (approved):
I expanded a bit on the documentation, I felt we needed to give an official recommendation of how to store the function as it may seem pretty easy to just put it in storage as we were doing.

Also, I added cross-references where possible but the custom `setup` and `push` can't be referenced since adding two `-` (hyphens) gets encoded as `—%E2%80%8B` and can't find where the issue is (likely in https://github.com/OpenZeppelin/solidity-docgen/)

LGTM

PR #5081, opened 2024-06-14
`MessageHashUtils`: Add `toDataWithIntendedValidatorHash(address, bytes32)`

StackOverflowExcept1on:
#### PR Checklist

- [x] Tests: `toDataWithIntendedValidatorHash()` has already been tested, but I added tests for `bytes32` version
- [x] Documentation


Amxx:
Postponning this as there are not tests, and no offchain tooling to provide signature for such hashes.

StackOverflowExcept1on:
@Amxx Can you elaborate on what kind of tests I need to write? Maybe I can copy them from somewhere. I'm not very familiar with Hardhat ecosystem.

Amxx:
Update: actually the testing part we already have a 
```
describe('toDataWithIntendedValidatorHash', function () {
``` 
that tests with random message of 128 bytes. So we would only need to duplicate that with a bytes32 object.

---

But overall, what I'd like to understand is why is that function really needed? What is the issue with calling the existing `toDataWithIntendedValidatorHash(address,bytes)` with abi.encodePacked manually? I'm not sure many people will call either version of the function. For sure manually adding the `abi.encodePacked(...)` is not such a big deal.

StackOverflowExcept1on:
@Amxx 

> For sure manually adding the `abi.encodePacked(...)` is not such a big deal.

It's okay to have two versions: one with message hash (digest), second with message. Also maybe we could think about providing more optimized function like `toEthSignedMessageHash` does now.

StackOverflowExcept1on:
@Amxx I fixed tests and added optimization that does not require memory allocation for the version with `bytes32 messageHash`. Could you please take a look?

.changeset/quiet-shrimps-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"openzeppelin-solidity": patch
+---
+
+feat: add `toDataWithIntendedValidatorHash()` with `messageHash`
```
Amxx, line 5:
```suggestion
`MessageHashUtils`: Add `toDataWithIntendedValidatorHash(address, bytes32)`.
```

StackOverflowExcept1on:
@Amxx I found this useful for our smart contract using custom Schnorr signatures, but it might also make sense for saving gas with ECDSA signatures.

contracts/utils/cryptography/MessageHashUtils.sol
```diff
@@ -63,6 +63,27 @@ library MessageHashUtils {
...
     }
 
+    /**
+     * @dev Returns the keccak256 digest of an ERC-191 signed data with version
+     * `0x00` (data with intended validator).
+     *
+     * The digest is calculated by prefixing a bytes32 `messageHash` with `"\x19\x00"` and the intended
+     * `validator` address. Then hashing the result.
+     *
+     * See {ECDSA-recover}.
```
Amxx, line 73:
```suggestion
     * @dev Variant of {toDataWithIntendedValidatorHash-address-bytes-} optimized for cases where `data` is a bytes32.
```

contracts/utils/cryptography/MessageHashUtils.sol
```diff
@@ -63,6 +63,21 @@ library MessageHashUtils {
...
+    /**
+     * @dev Variant of {toDataWithIntendedValidatorHash-address-bytes-} optimized for cases where `data` is a bytes32.
+     */
+    function toDataWithIntendedValidatorHash(
+        address validator,
+        bytes32 messageHash
+    ) internal pure returns (bytes32 digest) {
+        assembly ("memory-safe") {
+            mstore(0x00, hex"19_00")
+            mstore(0x02, shl(96, validator))
```
Amxx, line 75: TODO: sanitize address

StackOverflowExcept1on: wdym?

Amxx: I added that note during a meeting. Someone brote that up. Often when we write memory in assembly we have so assume that the input values may contain dirty upper bits.

However now that I look at it more carefully, the `shl(96, ...)` part should already fix that. 

@ernestognw can you confirm its fine?

Amxx: The first part is not always true. Its quite easy to generate "dirty" addresses that have non-zero upper bits. The `shl` should indeed address that.

I'm adding fuzzing tests

Amxx (dismissed):

contracts/utils/cryptography/MessageHashUtils.sol
```diff
@@ -63,6 +63,21 @@ library MessageHashUtils {
         return keccak256(abi.encodePacked(hex"19_00", validator, data));
     }
 
+    /**
+     * @dev Variant of {toDataWithIntendedValidatorHash-address-bytes-} optimized for cases where `data` is a bytes32.
```
arr00, line 67:
```suggestion
     * @dev Variant of {toDataWithIntendedValidatorHash-address-bytes} optimized for cases where `data` is a bytes32.
```

arr00 (approved):

Amxx (approved):

PR #5084, opened 2024-06-16
Add Binary heap structure

### LINKED ISSUE #3410: Add priority queue as a data structure to be able to implement max/min heaps

Amxx:
Fixes #3410
Alternative to #5076

This PR uses names from https://en.wikipedia.org/wiki/Heap_(data_structure)#Operations
- Basic
  - peek
  - insert
  - pop 
  - replace
- Internal
  - siftUp
  - siftDown

For the inspection part, we use `length` instead of `size` for consistencty with all other structures in the `utils/struct` folder.

#### PR Checklist

- [x] Tests
- [x] Tests Fuzzing
- [x] Documentation


contracts/utils/structs/Heap.sol
```diff
@@ -0,0 +1,228 @@
...
+     */
+    function _heapifyDown(
+        Uint256Heap storage self,
+        uint32 size,
+        uint32 pos,
+        uint256 value,
+        function(uint256, uint256) view returns (bool) comp
+    ) private {
+        uint32 left = 2 * pos + 1;
+        uint32 right = 2 * pos + 2;
```
Amxx, line 181: this could overflow uint32. the checks just after guarantee that they are smaller then size which is uint32.
maybe we could compute as uint256 and then cast down in the ifs.

I don't think its a realistic issue though.

contracts/utils/structs/Heap.sol
```diff
@@ -0,0 +1,228 @@
...
+     * during the lifecycle of a heap will result in undefined behavior.
+     */
+    function insert(
+        Uint256Heap storage self,
+        uint256 value,
+        function(uint256, uint256) view returns (bool) comp
+    ) internal {
+        uint32 size = length(self);
+        self.data.push(Uint256HeapNode({index: size, lookup: size, value: value}));
+        _heapifyUp(self, size, value, comp);
```
Amxx, line 133: is length is type(uint32).max, the insert works, but then the size is increassed, and the length function reverts. Maybe we want to check here and revert ?

That check would hewever have a cost, and its won't realistically be triggered.

.githooks/pre-push
```diff
@@ -3,5 +3,6 @@
 set -euo pipefail
 
 if [ "${CI:-"false"}" != "true" ]; then
+	npm run generate
```
Amxx, line 6:
```suggestion
  npm run generate
```

test/utils/structs/Heap.test.js
```diff
@@ -0,0 +1,130 @@
...
+        await expect(this.helper.pop()).to.emit(this.mock, popEvent).withArgs(42n);
+        await expect(this.helper.pop()).to.emit(this.mock, popEvent).withArgs(42n);
+        await expect(this.helper.pop()).to.emit(this.mock, popEvent).withArgs(42n);
+        await expect(this.helper.pop()).to.emit(this.mock, popEvent).withArgs(42n);
+
+        // popping a 6th time panics
+        await expect(this.helper.pop()).to.be.revertedWithPanic(PANIC_CODES.POP_ON_EMPTY_ARRAY);
+      });
+
+      it('insert, pop ansd replace', async function () {
```
Amxx, line 77:
```suggestion
      it('insert, pop and replace', async function () {
```

test/utils/structs/Heap.test.js
```diff
@@ -0,0 +1,131 @@
...
+          { op: 'insert', value: 712 }, // [712]
+          { op: 'insert', value: 20 }, // [20, 712]
+          { op: 'insert', value: 4337 }, // [20, 712, 4437]
+          { op: 'pop' }, // 20, [712, 4437]
+          { op: 'insert', value: 1559 }, // [712, 1559, 4437]
+          { op: 'insert', value: 165 }, // [165, 712, 1559, 4437]
+          { op: 'insert', value: 155 }, // [155, 165, 712, 1559, 4437]
+          { op: 'insert', value: 7702 }, // [155, 165, 712, 1559, 4437, 7702]
+          { op: 'pop' }, // 155, [165, 712, 1559, 4437, 7702]
+          { op: 'replace', value: 721 }, // [712, 721, 1559, 4437, 7702]
```
Amxx, line 89:
```suggestion
          { op: 'replace', value: 721 }, // 165, [712, 721, 1559, 4437, 7702]
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,310 @@
...
+import {Panic} from "../Panic.sol";
+`;
+
+const generate = ({ struct, node, valueType, indexType, blockSize }) => `\
+/**
+ * A Heap is represented as an array of Node objects. In this array we store two overlapping structures:
+ * - A tree structure, where index 0 is the root, and for each index i, the children are 2*i+1 and 2*i+2.
+ *   For each index in this tree we have the \`index\` pointer that gives the position of the corresponding value.
+ * - An array of values (payload). At each index we store a ${valueType} \`value\` and \`lookup\`, the index of the node
+ *   that points to this value.
```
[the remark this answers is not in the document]
ernestognw: I feel this block should be in the top NatSpec and not in the struct docs

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,310 @@
...
+ *   \`\`\`
+ *
+ * The structure is order so that each node is bigger then its parent. An immediate consequence is that the
+ * smallest value is the one at the root. It can be retrieved in O(1) at \`heap.data[heap.data[0].index].value\`
+ *
+ * This structure is designed for the following complexities:
+ * - peek (get the smallest value in set): O(1)
+ * - insert (insert a value in the set): 0(log(n))
+ * - pop (remove the smallest value in set): O(log(n))
+ * - replace (replace the smallest value in set with a new value): O(log(n))
```
ernestognw, line 35:
```suggestion
 * The structure is designed to perform the following operations with its corresponding complexity:
 *
 * * peek (get the smallest value in set): O(1)
 * * insert (insert a value in the set): 0(log(n))
 * * pop (remove the smallest value in set): O(log(n))
 * * replace (replace the smallest value in set with a new value): O(log(n))
```

contracts/mocks/Stateless.sol
```diff
@@ -22,6 +22,7 @@ import {ERC165} from "../utils/introspection/ERC165.sol";
 import {ERC165Checker} from "../utils/introspection/ERC165Checker.sol";
 import {ERC1967Utils} from "../proxy/ERC1967/ERC1967Utils.sol";
 import {ERC721Holder} from "../token/ERC721/utils/ERC721Holder.sol";
+import {Heap} from "../utils/structs/Heap.sol";
```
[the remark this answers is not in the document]
Amxx: It was overlooked, and since there are currently no tests for the `Comparators` library, it missing did not triger any issue.

Lets start by discussing weither we want the comparator library or not ... then we can add tests and solve that

scripts/generate/templates/Heap.js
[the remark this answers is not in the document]
Amxx: lets discuss that part on our call

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+            _siftDown(self, size, lIndex, value, comp);
+        }
+    }
+}
+
+/**
+ * @dev Perform heap maintenance on \`self\`, starting at position \`pos\` (with the \`value\`), using \`comp\` as a
+ * comparator, and moving toward the root of the underlying tree.
+ *
+ * Note: This is a private function that is called in a trusted context with already cached parameters. \`value\`
```
ernestognw, line 278:
```suggestion
 * NOTE: This is a private function that is called in a trusted context with already cached parameters. \`value\`
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+    // update lookup pointers for consistency
+    _unsafeNodeAccess(self, ii).lookup = j;
+    _unsafeNodeAccess(self, jj).lookup = i;
+}
+
+/**
+ * @dev Perform heap maintenance on \`self\`, starting at position \`pos\` (with the \`value\`), using \`comp\` as a
+ * comparator, and moving toward the leafs of the underlying tree.
+ *
+ * Note: This is a private function that is called in a trusted context with already cached parameters. \`length\`
```
ernestognw, line 234:
```suggestion
 * NOTE: This is a private function that is called in a trusted context with already cached parameters. \`length\`
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+ */
+function replace(${struct} storage self, ${valueType} newValue) internal returns (${valueType}) {
+    return replace(self, newValue, Comparators.lt);
+}
+
+/**
+ * @dev Return the root element for the heap, and replace it with a new value, using the provided comparator.
+ * This is equivalent to using {pop} and {insert}, but requires only one rebalancing operation.
+ *
+ * Note: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```
ernestognw, line 172:
```suggestion
 * NOTE: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+
+    self.data.push(${struct}Node({index: size, lookup: size, value: value}));
+    _siftUp(self, size, value, comp);
+}
+
+/**
+ * @dev Return the root element for the heap, and replace it with a new value, using the default comparator.
+ * This is equivalent to using {pop} and {insert}, but requires only one rebalancing operation.
+ *
+ * Note: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```
ernestognw, line 161:
```suggestion
 * NOTE: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+ * during the lifecycle of a heap will result in undefined behavior.
+ */
+function insert(${struct} storage self, ${valueType} value) internal {
+    insert(self, value, Comparators.lt);
+}
+
+/**
+ * @dev Insert a new element in the heap using the provided comparator.
+ *
+ * Note: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```
ernestognw, line 142:
```suggestion
 * NOTE: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+
+        // return root value
+        return rootDataValue;
+    }
+}
+
+/**
+ * @dev Insert a new element in the heap using the default comparator.
+ *
+ * Note: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```
ernestognw, line 132:
```suggestion
 * NOTE: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+ * during the lifecycle of a heap will result in undefined behavior.
+ */
+function pop(${struct} storage self) internal returns (${valueType}) {
+    return pop(self, Comparators.lt);
+}
+
+/**
+ * @dev Remove (and return) the root element for the heap using the provided comparator.
+ *
+ * Note: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```
ernestognw, line 80:
```suggestion
 * NOTE: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+ */
+function peek(${struct} storage self) internal view returns (${valueType}) {
+    // self.data[0] will \`ARRAY_ACCESS_OUT_OF_BOUNDS\` panic if heap is empty.
+    return _unsafeNodeAccess(self, self.data[0].index).value;
+}
+
+/**
+ * @dev Remove (and return) the root element for the heap using the default comparator.
+ *
+ * Note: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```
ernestognw, line 70:
```suggestion
 * NOTE: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,322 @@
...
+        ${indexType} lastIdx = lastNode.index;
+        ${valueType} lastValue = _unsafeNodeAccess(self, lastIdx).value;
+
+        // move the last leaf to the root, pop last leaf ...
+        rootNode.index = lastIdx;
+        _unsafeNodeAccess(self, lastIdx).lookup = 0;
+        self.data.pop();
+
+        // ... and heapify
+        _siftDown(self, last, 0, lastValue, comp);
```
ernestognw, line 122: Wouldn't this be needed only if `rootIdx !== last` as well? Otherwise my understanding is that the order is preserved so it may be unnecessary

contracts/utils/structs/Heap.sol
```diff
@@ -0,0 +1,570 @@
...
+     * This is equivalent to using {pop} and {insert}, but requires only one rebalancing operation.
+     *
+     * Note: All inserting and removal from a heap should always be done using the same comparator. Mixing comparator
+     * during the lifecycle of a heap will result in undefined behavior.
+     */
+    function replace(
+        Uint256Heap storage self,
+        uint256 newValue,
+        function(uint256, uint256) view returns (bool) comp
+    ) internal returns (uint256) {
```
ernestognw, line 176: I'm not convinced about `replace` for this one, is this mimicking any C++ library?

contracts/utils/structs/Heap.sol
```diff
@@ -0,0 +1,570 @@
...
+import {SafeCast} from "../math/SafeCast.sol";
+import {Comparators} from "../Comparators.sol";
+import {Panic} from "../Panic.sol";
+
+/**
+ * @dev Library for managing https://en.wikipedia.org/wiki/Binary_heap[binary heap] that can be used as
+ * https://en.wikipedia.org/wiki/Priority_queue[priority queue].
+ *
+ * Heaps are represented as an array of Node objects. This array stores two overlapping structures:
+ * * A tree structure where the first element (index 0) is the root, and where the node at index i is the child of the
```
cairoeth, line 15: if we use index 1 instead of 0 it can simplify operations of parent, left-child and right-child from (n-1)/2, 2n+1 and 2n+2 to n/2, 2n and 2n + 1, respectively.

this makes parent and left-child more optimal. if p points to a 0-based array and q = p - 1 then we can access p[0] as q[1] so there is no overhead in using a 1-based array

contracts/utils/structs/Heap.sol
Amxx: All names come from here: https://en.wikipedia.org/wiki/Heap_(data_structure)#Operations

contracts/utils/structs/Heap.sol
Amxx: This makes is so that `array.length` is no longer the size of the Heap. It also creates a possible issue with empty array and array with one elements both being somehow empty.

The complexity of doing `(n-1)/2` or `2n+1`, `2n+2` is not that much more complexe compared to the special cases we would have to consider

scripts/generate/templates/Heap.js
Amxx: if `rootIdx !== last`, what we did is move the root value to the end of the array, so it can safelly be popped. we shifted around the `value` part of the array, and made sure the indices are still correct. 

For that last part we made sure that the index that used to point to the last value now point to the location that the root used to point to (that stored the data that we pop).

This shifting in the values do NOT change the tree. Each node of the tree still points to the same value (but possibly at a different location).

Just before pop-ing the array we did `rootNode.index = lastIdx;`, which is basically saying 
> we take the last leaf, which could be anything, put it where the root was, and then propagate it down. 

There is no reason to think that this last root is already properly placed. As far as we know it could be the biggest value in the heap.

We have fuzzing tests, so fell free to give it a try. I don't think your logic is good, but maybe I'm missing something.

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,327 @@
...
+        ${valueType} lValue = _unsafeNodeAccess(self, _unsafeNodeAccess(self, lIndex).index).value;
+        ${valueType} rValue = _unsafeNodeAccess(self, _unsafeNodeAccess(self, rIndex).index).value;
+        if (comp(lValue, value) || comp(rValue, value)) {
+            if (comp(lValue, rValue)) {
+                _swap(self, pos, lIndex);
+                _siftDown(self, size, lIndex, value, comp);
+            } else {
+                _swap(self, pos, rIndex);
+                _siftDown(self, size, rIndex, value, comp);
+            }
```
ernestognw, line 266: I think is more readable to do:

```suggestion
            index = ternary(comp(lValue, rValue), lIndex, rIndex);
            _swap(self, pos, index);
            _siftDown(self, size, index, value, comp);
```

Though, I understand it may require an extra value in the stack so it's not more optimal. Just leaving the suggestion

ernestognw (approved):
Added some docs, mostly minor things. LGTM

I can reapprove if suggestions are applied

scripts/generate/templates/Heap.opts.js
```diff
@@ -0,0 +1,13 @@
...
+  node: `Uint${valueSize}HeapNode`,
+  valueSize,
+  valueType: `uint${valueSize}`,
+  indexSize,
+  indexType: `uint${indexSize}`,
+  blockSize: Math.ceil((valueSize + 2 * indexSize) / 256),
+});
+
+module.exports = {
+  TYPES: [makeType(256, 32), makeType(208, 24)],
```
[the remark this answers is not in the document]
Amxx: I dont think its realistic to push 4.2 billion values without poping any in the meantime. That would be 8.4 billions storage slot stores that are written to. That is the equivalent of 180 000 eth worth of gas (considering a very con servative 1 gwei/gas) ... just for writting to the data to the heap (not even including the rebalancing of the heap (and the other operations done at the same time). If you include the gas cost of all the reordering along the way, you are probable closer to 300 000 eth consumed on gas.

Note that if you heap has ~ 2**32 items already, pushing a new one means shitfing up ~32 levels of the tree. Each level requires reading the index + the value, which are 2 slots in the case of uint256. That is ~134k of gas just for warlming the slots. If you add the writes to reorder on top of that ... and its JUST for one structure push when the structure is close to full.

The idea was that both version use the same type for length so that the dev experience is similar, and IMO the limits are way sufficient.

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,327 @@
...
+/**
+ * @dev Binary heap that support values of type ${valueType}.
+ *
+ * Each element of that structures uses ${blockSize} storage slots.
+ */
+struct ${struct} {
+    ${node}[] data;
+}
+
+struct ${node} {
```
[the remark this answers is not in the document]
Amxx:
```suggestion
/**
 * @dev Internal node type for ${struct}. Stores a value of type ${valueType}.
 */
struct ${node} {
```

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,327 @@
...
+ *   node at index (i-1)/2 and the father of nodes at index 2*i+1 and 2*i+2. Each node stores the index (in the array)
+ *   where the corresponding value is stored.
+ * * A list of payloads values where each index contains a value and a lookup index. The type of the value depends on
+ *   the variant being used. The lookup is the index of the node (in the tree) that points to this value.
+ *
+ * Some invariants:
+ *   \`\`\`
+ *   i == heap.data[heap.data[i].index].lookup // for all indices i
+ *   i == heap.data[heap.data[i].lookup].index // for all indices i
+ *   \`\`\`
```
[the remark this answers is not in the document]
Amxx: Created an issue so we do that in a further PR (during the audit?)

scripts/generate/templates/Heap.js
```diff
@@ -0,0 +1,327 @@
...
+ * The structure is ordered so that each node is bigger than its parent. An immediate consequence is that the
+ * smallest value is the one at the root. This value can be lookup up in constant time (O(1)) at
+ * \`heap.data[heap.data[0].index].value\`
+ *
+ * The structure is designed to perform the following operations with the corresponding complexities:
+ *
+ * * peek (get the smallest value in set): O(1)
+ * * insert (insert a value in the set): 0(log(n))
+ * * pop (remove the smallest value in set): O(log(n))
+ * * replace (replace the smallest value in set with a new value): O(log(n))
```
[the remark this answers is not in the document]
Amxx: I'm honestly not a fan of "most important".

There may be cases where you use a Heap to keep the "important" element at the end, and "dump" the one you don't care about. I've got a few examples that we can discuss. In that case arguably the root is the "least important", and you want to pop/replace it with more "important" ones.

While "smallest" assumes the default comparator is used, I think its really clear and easy to understand. 

If you insist on changing, what would you think of "higest priority".

scripts/generate/templates/Heap.js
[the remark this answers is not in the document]
Amxx: Note that this is "manually" checked during the fuzzing tests: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5084/files#diff-c3eff7ae8b7d32a0c90af26a40d0bda533d4021ca020ee4e985576e3c4e705d9R19-R20

scripts/generate/templates/Heap.js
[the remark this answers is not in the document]
ernestognw: Yes, "important" is subjective and I think "highest priority" conveys what we're trying to do. I'm updating:

```
 * most important value is the one at the root. This value can be lookup up in constant time (O(1)) at
 * \`heap.data[heap.data[0].index].value\`
 *
 * The structure is designed to perform the following operations with the corresponding complexities:
 *
 * * peek (get the most important value in set): O(1)
 * * insert (insert a value in the set): 0(log(n))
 * * pop (remove the most important value in set): O(log(n))
 * * replace (replace the most important value in set with a new value): O(log(n))
```

scripts/generate/templates/Heap.opts.js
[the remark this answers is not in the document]
ernestognw: Right the numbers are huge, but imo is not unthinkable in the context of high performance L2s (for e.g. gaming)

Regardless, I agree that it's an edge case and I'd rather wait for a user to provide us feedback

scripts/generate/templates/Heap.js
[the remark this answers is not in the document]
cairoeth: yep, we also have a couple of other invariants that we can check, for example: if there exists a node n at depth h, then all 2^(h-1) possible node of depth h–1 exist in the tree, along with every possible node to the left of n of depth h. it follows that if h is the maximum depth of a node in the tree, the leaves of the tree occur only at depths h and h–1.

ernestognw (approved):

scripts/generate/templates/Heap.js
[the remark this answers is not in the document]
ernestognw: Right, that'd be a good invariant to check too. Once the audit starts we can analyze where we can use invariant testing within our internal advanced testing tracking on Notion

cairoeth (approved):
lgtm!

PR #5088, opened 2024-06-17
Fix repeated word in `ERC2771Forwarder` comments

GianfrancoBazzani:
fix repeated word


ernestognw (approved):
Thanks @GianfrancoBazzani!

PR #5091, opened 2024-06-18
Low level call and memory management

### LINKED ISSUE #5013: Low level call library
**🧐 Motivation**
[this library](https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/utils/Exec.sol) provides a solidity level interface to execute calls/staticcalls/delegate call without automatically dupping all the return data to memory. This makes it return-bomb safe.

We should build a similar library an use it in places such as the Forwarder.

Amxx:
Fixes #5013

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -84,10 +84,10 @@ abstract contract ERC4626 is ERC20, IERC4626 {
      * @dev Attempts to fetch the asset decimals. A return value of false indicates that the attempt failed in some way.
      */
     function _tryGetAssetDecimals(IERC20 asset_) private view returns (bool, uint8) {
```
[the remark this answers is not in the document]
ernestognw: Isn't it because the return value may revert at decoding? A `try IERC20Metadata(address(asset_)).decimals() returns (uint8 value)` would've failed if the token returned an `uint256`. Nothing in the ERC20 restricts the return value to uint8

Amxx:
I tried using a library like this one https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/utils/Exec.sol. The result did not really look good to me. I found the use of assembly actually cleaner ... because we automatically copy X bytes of the returndata, without having to call an additional `returndatacopy`. 

I found that the most valuable thing from the 4337 draft PR was the free memory management, to dealocate memory.

contracts/access/manager/AuthorityUtils.sol
```diff
@@ -17,16 +18,21 @@ library AuthorityUtils {
...
+
+        bytes memory params = abi.encodeCall(IAuthority.canCall, (caller, target, selector));
+        assembly ("memory-safe") {
+            let success := staticcall(not(0), authority, add(params, 0x20), mload(params), 0, 0x40)
+            if success {
+                if gt(returndatasize(), 0x1F) {
+                    immediate := gt(mload(0x00), 0)
+                }
+                if gt(returndatasize(), 0x3F) {
+                    delay := and(mload(0x20), 0xFFFFFFFF)
```
Amxx, line 31: Should this revert if delay is too large ?
```suggestion
                    delay := mload(0x20)
                    if gt(delay, 0xFFFFFFFF) { revert(0,0) }
```

contracts/utils/Memory.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Helper library for deallocating memory reserved by abi.encode or low level calls.
+ */
+library Memory {
+    type FreePtr is bytes32;
```
ernestognw, line 9: I think this should be just `Pointer`, it doesn't have anything to do with the free memory pointer strictly talking (e.g. the library may add functions for dealing with pointers)

```suggestion
    type Pointer is bytes32;
```

contracts/utils/Memory.sol
```diff
@@ -0,0 +1,22 @@
...
+
+    function save() internal pure returns (FreePtr ptr) {
```
ernestognw, line 11: Similarly, I think this should be very explicit, otherwise reading `Memory.save()` doesn't tell much.

```suggestion
    function saveFreePointer() internal pure returns (Pointer ptr) {
```

contracts/utils/Memory.sol
```diff
@@ -0,0 +1,22 @@
...
+        assembly ("memory-safe") {
+            ptr := mload(0x40)
+        }
+    }
+
+    function load(FreePtr ptr) internal pure {
```
ernestognw, line 17: Accordingly

```suggestion
    function loadFreePointer(Pointer ptr) internal pure {
```

contracts/access/manager/AuthorityUtils.sol
ernestognw, line 31:
```suggestion
                let size := returndatasize()
                if gt(size, 0x1F) {
                    immediate := gt(mload(0x00), 0)
                }
                if gt(size, 0x3F) {
                    delay := and(mload(0x20), 0xFFFFFFFF)
```

ernestognw (commented):
Left some comments, though I've been thinking about the set of "low level" functions we can offer, and I feel there are essentially 3 tricks people would do with this library:

- Calling and ignoring the return data
- Caling and saving the returndata in scratch space
- Calling and saving the returndata on top of another another memory pointer

So, I feel it'd make sense to offer a library with the following functions:

```solidity
library LowLevelCall {
  function callRaw(address target, uint256 value, bytes memory data) internal returns (bool success) {
        assembly ("memory-safe") {
            success := call(gas(), target, value, add(data, 0x20), mload(data), 0, 0)
        }
    }

    function callReturnBytes32(
        address target,
        uint256 value,
        bytes memory data
    ) internal returns (bool success, bytes32 result) {
        assembly ("memory-safe") {
            success := call(gas(), target, value, add(data, 0x20), mload(data), 0, 0x20)
            result := mload(0)
        }
    }

    function callReturnBytes32Tuple(
        address target,
        uint256 value,
        bytes memory data
    ) internal returns (bool success, bytes32 result1, bytes32 result2) {
        assembly ("memory-safe") {
            success := call(gas(), target, value, add(data, 0x20), mload(data), 0, 0x40)
            result1 := mload(0)
            result2 := mload(0x20)
        }
    }

    function callReturnOverride(
        address target,
        uint256 value,
        bytes memory data,
        bytes memory resultPtr
    ) internal returns (bool success) {
        // WARNING: Memory safetiness depends on the caller to ensure that writing to `resultPtr` is safe
        assembly ("memory-safe") {
            let maxSize := mload(resultPtr)
            success := call(gas(), target, value, add(data, 0x20), mload(data), resultPtr, maxSize)
        }
    }
}
```

The changes for this PR should be doable with these variants so far, I'll iterate this.

contracts/access/manager/AuthorityUtils.sol
ernestognw: `abi.decode` reverts, so I guess we should revert too to keep backwards compatibility

PR #5094, opened 2024-06-20
Implement LowLevelCall library

### LINKED ISSUE #5013: Low level call library

ernestognw:
Fixes #5013

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC20/extensions/ERC4626.sol
+++ b/contracts/token/ERC20/extensions/ERC4626.sol
@@ -7,6 +7,8 @@ import {IERC20, IERC20Metadata, ERC20} from "../ERC20.sol";
 import {SafeERC20} from "../utils/SafeERC20.sol";
 import {IERC4626} from "../../../interfaces/IERC4626.sol";
 import {Math} from "../../../utils/math/Math.sol";
+import {Memory} from "../../../utils/Memory.sol";
+import {LowLevelCall} from "../../../utils/LowLevelCall.sol";
 
 /**
  * @dev Implementation of the ERC-4626 "Tokenized Vault Standard" as defined in
@@ -75,25 +77,21 @@ abstract contract ERC4626 is ERC20, IERC4626 {
      * @dev Set the underlying asset contract. This must be an ERC20-compatible contract (ERC-20 or ERC-777).
      */
     constructor(IERC20 asset_) {
-        (bool success, uint8 assetDecimals) = _tryGetAssetDecimals(asset_);
-        _underlyingDecimals = success ? assetDecimals : 18;
+        _underlyingDecimals = _tryGetAssetDecimalsWithFallback(asset_, 18);
         _asset = asset_;
     }
 
-    /**
-     * @dev Attempts to fetch the asset decimals. A return value of false indicates that the attempt failed in some way.
-     */
-    function _tryGetAssetDecimals(IERC20 asset_) private view returns (bool, uint8) {
-        (bool success, bytes memory encodedDecimals) = address(asset_).staticcall(
-            abi.encodeCall(IERC20Metadata.decimals, ())
-        );
-        if (success && encodedDecimals.length >= 32) {
-            uint256 returnedDecimals = abi.decode(encodedDecimals, (uint256));
-            if (returnedDecimals <= type(uint8).max) {
-                return (true, uint8(returnedDecimals));
-            }
-        }
-        return (false, 0);
+    function _tryGetAssetDecimalsWithFallback(IERC20 asset_, uint8 defaultValue) private view returns (uint8) {
+        Memory.Pointer ptr = Memory.saveFreePointer();
+        bytes memory params = abi.encodeCall(IERC20Metadata.decimals, ());
+
+        (bool success, bytes32 rawValue) = LowLevelCall.staticCallReturnBytes32(address(asset_), params);
+        uint256 length = LowLevelCall.returnDataSize();
+        uint256 value = uint256(rawValue);
+
+        Memory.loadFreePointer(ptr);
+
+        return uint8(Math.ternary(success && length >= 0x20 && value <= type(uint8).max, value, defaultValue));
     }
 
     /**
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -4,6 +4,7 @@
 pragma solidity ^0.8.20;
 
 import {Errors} from "./Errors.sol";
+import {LowLevelCall} from "./LowLevelCall.sol";
 
 /**
  * @dev Collection of functions related to the address type
@@ -35,7 +36,7 @@ library Address {
             revert Errors.InsufficientBalance(address(this).balance, amount);
         }
 
-        (bool success, ) = recipient.call{value: amount}("");
+        bool success = LowLevelCall.callRaw(recipient, amount, "");
         if (!success) {
             revert Errors.FailedCall();
         }
--- a/contracts/utils/LowLevelCall.sol
+++ b/contracts/utils/LowLevelCall.sol
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Errors} from "./Errors.sol";
+
+/**
+ * @dev Library of low level call functions that implement different calling strategies to deal with the return data.
+ */
+library LowLevelCall {
+    /// === CALL ===
+
+    /// @dev Performs a Solidity function call using a low level `call` and ignoring the return data.
+    function callRaw(address target, uint256 value, bytes memory data) internal returns (bool success) {
+        assembly ("memory-safe") {
+            success := call(gas(), target, value, add(data, 0x20), mload(data), 0, 0)
+        }
+    }
+
+    /// @dev Performs a Solidity function call using a low level `call` and returns the first 32 bytes of the result
+    /// in the scratch space of memory. Useful for functions that return a single-word value.
+    ///
+    /// WARNING: Do not assume that the result is zero if `success` is false. Memory can be already allocated
+    /// and this function doesn't zero it out.
+    function callReturnBytes32(
+        address target,
+        uint256 value,
+        bytes memory data
+    ) internal returns (bool success, bytes32 result) {
+        assembly ("memory-safe") {
+            success := call(gas(), target, value, add(data, 0x20), mload(data), 0, 0x20)
+            result := mload(0)
+        }
+    }
+
+    /// @dev Performs a Solidity function call using a low level `call` and returns the first 64 bytes of the result
+    /// in the scratch space of memory. Useful for functions that return a tuple of single-word values values.
+    ///
+    /// WARNING: Do not assume that the results are zero if `success` is false. Memory can be already allocated
+    /// and this function doesn't zero it out.
+    function callReturnBytes32Tuple(
+        address target,
+        uint256 value,
+        bytes memory data
+    ) internal returns (bool success, bytes32 result1, bytes32 result2) {
+        assembly ("memory-safe") {
+            success := call(gas(), target, value, add(data, 0x20), mload(data), 0, 0x40)
+            result1 := mload(0)
+            result2 := mload(0x20)
+        }
+    }
+
+    /// @dev Performs a Solidity function call using a low level `call` and writes the result to the memory location
+    /// specified by `resultPtr`.
+    ///
+    /// IMPORTANT: This function assumes that the length of the memory array is stored in the first 32 bytes of the array and uses it for truncating
+    /// returndata if it's longer than the allocated memory to avoid corrupting to further places in memory. The `resultPtr` should be a
+    /// memory location that is already allocated with a predefined length.
+    ///
+    /// WARNING: Do not use if writing to `resultPtr` is not safe according to
+    /// the https://docs.soliditylang.org/en/latest/assembly.html#memory-safety[Solidity documentation].
+    function callReturnOverride(
+        address target,
+        uint256 value,
+        bytes memory data,
+        bytes memory resultPtr
+    ) internal returns (bool success) {
+        assembly ("memory-safe") {
+            let maxSize := mload(resultPtr)
+            success := call(gas(), target, value, add(data, 0x20), mload(data), resultPtr, maxSize)
+        }
+    }
+
+    /// === STATICCALL ===
+
+    /// @dev Performs a Solidity function call using a low level `staticcall` and ignoring the return data.
+    function staticCallRaw(address target, bytes memory data) internal view returns (bool success) {
+        assembly ("memory-safe") {
+            success := staticcall(gas(), target, add(data, 0x20), mload(data), 0, 0)
+        }
+    }
+
+    /// @dev Performs a Solidity function call using a low level `staticcall` and returns the first 32 bytes of the result
+    /// in the scratch space of memory. Useful for functions that return a single-word value.
+    ///
+    /// WARNING: Do not assume that the result is zero if `success` is false. Memory can be already allocated
+    /// and this function doesn't zero it out.
+    function staticCallReturnBytes32(
+        address target,
+        bytes memory data
+    ) internal view returns (bool success, bytes32 result) {
+        assembly ("memory-safe") {
+            success := staticcall(gas(), target, add(data, 0x20), mload(data), 0, 0x20)
+            result := mload(0)
+        }
+    }
+
+    /// @dev Performs a Solidity function call using a low level `staticcall` and returns the first 64 bytes of the result
+    /// in the scratch space of memory. Useful for functions that return a tuple of single-word values values.
+    ///
+    /// WARNING: Do not assume that the results are zero if `success` is false. Memory can be already allocated
+    /// and this function doesn't zero it out.
+    function staticCallReturnBytes32Tuple(
+        address target,
+        bytes memory data
+    ) internal view returns (bool success, bytes32 result1, bytes32 result2) {
+        assembly ("memory-safe") {
+            success := staticcall(gas(), target, add(data, 0x20), mload(data), 0, 0x40)
+            result1 := mload(0)
+            result2 := mload(0x20)
+        }
+    }
+
+    /// @dev Performs a Solidity function call using a low level `staticcall` and writes the result to the memory location
+    /// specified by `resultPtr`.
+    ///
+    /// IMPORTANT: This function assumes that the length of the memory array is stored in the first 32 bytes of the array and uses it for truncating
+    /// returndata if it's longer than the allocated memory to avoid corrupting to further places in memory. The `resultPtr` should be a
+    /// memory location that is already allocated with a predefined length.
+    ///
+    /// WARNING: Do not use if writing to `resultPtr` is not safe according to
+    /// the https://docs.soliditylang.org/en/latest/assembly.html#memory-safety[Solidity documentation].
+    function staticCallReturnOverride(
+        address target,
+        bytes memory data,
+        bytes memory resultPtr
+    ) internal view returns (bool success) {
+        assembly ("memory-safe") {
+            let maxSize := mload(resultPtr)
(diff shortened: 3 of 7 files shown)


contracts/utils/LowLevelCall.sol
```diff
@@ -0,0 +1,152 @@
...
+        address target,
+        bytes memory data,
+        uint256 value
+    ) internal returns (bool success, bytes32 result1, bytes32 result2) {
+        assembly ("memory-safe") {
+            success := call(gas(), target, value, add(data, 0x20), mload(data), 0, 0x40)
+            result1 := mload(0)
+            result2 := mload(0x20)
+        }
+    }
```
Amxx: 2 comments about the `callReturnBytes32` and `callReturnBytes32Pair`:

- If you don't know what is in scratch space before the call, then there is no way to distinguish between `result1` and `result2` comming from the call versus them just being the old value without looking at the returndatasize. I would not expect the users of this library to manually (assembly) set/clean the scratch space before doing the call, which means that the best way for them to validate the results is to also get the length of the return. Considering how crutial it is to check it, I would return it all the time.

- Given that the cost of an mload is just 3 gas, and that asking for 0x40 return length (instead of 0x20) doesn't change the gas cost of the call I would be in favor of replacing them with a single
```
function callReturnScratch(address target, bytes memory data, uint256 value) internal returns (bool success, uint256 returnlength, bytes32 result1, bytes32 result2)
```

Amxx: obviously this comment also applies to the static version.

contracts/utils/LowLevelCall.sol
```diff
@@ -0,0 +1,152 @@
...
+ * WARNING: Using this library requires an advanced understanding of Solidity and how the EVM works. It is recommended
+ * to use the {Address} library instead.
+ */
+library LowLevelCall {
+    using Memory for *;
+
+    /// === CALL ===
+
+    /// @dev Performs a Solidity function call using a low level `call` and ignoring the return data.
+    function callRaw(address target, bytes memory data) internal returns (bool success) {
```
Amxx: I'd name that `...ReturnEmpty` or `...NoReturn` for consistency with `...ReturnBytes32`
```suggestion
    function callReturnEmpty(address target, bytes memory data) internal returns (bool success) {
```

contracts/utils/LowLevelCall.sol
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /// @dev Returns the return data buffer.
+    function returnData() internal pure returns (bytes memory data) {
+        uint256 size = returnDataSize();
+        assembly ("memory-safe") {
+            data := mload(0x40)
+            returndatacopy(data, 0, size)
+        }
+    }
```
Amxx: This does not return a valid `bytes memory` object:
- the length is not stored in the buffer as solidity expectes
- the memory is not allocated (by moving the fmp)

Using it for anything beside the `bubbleRevert` function (which i would write differently anyway) would be super dangerous.

```suggestion
    function returnData() internal pure returns (bytes memory data) {
        uint256 size = returnDataSize();
        assembly ("memory-safe") {
            data := mload(0x40)
            mstore(data, size)
            returndatacopy(add(data, 0x20), 0, size)
            mstore(0x40, add(data, add(0x20, size)))
        }
    }
```

contracts/utils/LowLevelCall.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    /// === BUBBLE UP REVERT REASON ===
+
+    /**
+     * @dev Bubbles up the revert reason. This function is useful to bubble up the revert reason
+     * from a low level call. Developers can provide any bytes memory pointer to the revert reason
+     * since the function will revert regardless of the pointer value.
+     */
+    function bubbleRevert(bytes memory ptr) internal pure {
+        bubbleRevert(ptr.asPointer().asBytes32());
```
Amxx: if `bytes memory ptr` is not the same length as `returnDataSize`, this will not work.

contracts/utils/LowLevelCall.sol
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /// @dev Same as {bubbleRevert-bytes}, but for `bytes32` pointers.
+    function bubbleRevert(bytes32 ptr) internal pure {
```
Amxx: Does this have any usecase beside bubbling return data from the last lowlevelcall ?

If not, I would consider droping the ptr argument and doing 

```solidity
    function bubble() internal pure {
        assembly ("memory-safe") {
            returndatacopy(0x00, 0x00, returndatasize())
            revert(0, returndatasize())
        }
    }
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -311,9 +315,7 @@ abstract contract AccountERC7579 is Account, IERC1271, IERC7579Execution, IERC75
 
         if (success) return returndata;
 
-        assembly ("memory-safe") {
-            revert(add(returndata, 0x20), mload(returndata))
-        }
+        returndata.asPointer().addOffset(0x20).asBytes().bubbleRevert();
```
Amxx: I honestly think this is way to complex. It should just be `bubble(returndata)` or `bubble()`

Amxx (commented):
Is there any case where the bubble is something other that the return data that we received from a call that was not successfull ?

I think that by definition of "bubble", it implies that its already an error that was reverted by an unsuccessfull call. Therefore, I'm would argue maybe the "low level call" function should include a boolean to select is yes/no we want to bubble any revert we get.

It could look like
```solidity
    function callEmpty(address target, bytes memory data, uint256 value, bool bubble) internal returns (bool success) {
        assembly ("memory-safe") {
            success := call(gas(), target, value, add(data, 0x20), mload(data), 0x00, 0x00)
            if and(iszero(success), bubble) {
                let ptr := mload(0x40)
                returndatacopy(ptr, 0, returndatasize())
                revert(ptr, returndatasize())
            }
        }
    }

    function callScratch(address target, bytes memory data, uint256 value, bool bubble) internal returns (bool success, uint256 size, bytes32 x, bytes32 y) {
        assembly ("memory-safe") {
            success := call(gas(), target, value, add(data, 0x20), mload(data), 0x00, 0x40)
            if and(iszero(success), bubble) {
                let ptr := mload(0x40)
                returndatacopy(ptr, 0, returndatasize())
                revert(ptr, returndatasize())
            }
            size := returndatasize()
            x := mload(0x00)
            y := mload(0x20)
        }
    }

    function callFull(address target, bytes memory data, uint256 value, bool bubble) internal returns (bool success, bytes memory returndata) {
        assembly ("memory-safe") {
            success := call(gas(), target, value, add(data, 0x20), mload(data), 0x00, 0x00)

            returndata := mload(0x40)
            returndatacopy(add(returndata, 0x20), 0, returndatasize())
            if and(iszero(success), bubble) {
                revert(add(returndata, 0x20), returndatasize())
            }
            mstore(returndata, returndatasize())
            mstore(0x40, add(add(returndata, 0x20), returndatasize()))
        }
    }
```

--- 

Note: The `RETURNDATASIZE` opcode is only 2 gas, and `DUPX` are 3 gas. So it turns out caching the returndatasize in the stack and dupping it is counterproductive !

Amxx:
I updated the PR following our discussion of the day. I'm curious what the gas comparaison script tells us.

Note: the changes for SafeERC20 will be overwritten by the SafeERC20 PR.

Amxx:
[Gas number form CI](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/16192780880?pr=5094)
- Functions in `Address.sol` get cheaper by about 170gas (call) or 75gas (delegatecall)
- This impact other contract that use the Address library
  - `AccessManager.execute`
  - `ERC20Multicall.multicall`, `BatchCaller.execute`, `TransientSlot.multicall` ...
- SafeERC20 becomes more expensive by about ~200 gas. That is bad. It will be fixed by #5771. We could
  - Merge that first, then do #5771
  - Remove the SafeERC20 changes from this PR, then fo #5771
  - do #5771 first, then do this (without any change to SafeERC20.

Amxx:
Current remarquable numbers:

- `AccessManager.execute(address,bytes)`: -124 gas on average
- `AccountERC7579.validateUserOp`: -88 gas on average (-1833 in the worst case)
- `Create2.deploy`: -7500 gas on average
- `ERC20Multicall`: -530 gas (for batches of 2 calls)

contracts/utils/Address.sol
```diff
@@ -76,26 +80,50 @@ library Address {
...
+        bool success = LowLevelCall.callNoReturn(target, value, data);
+        if (success && (LowLevelCall.returnDataSize() > 0 || target.code.length > 0)) {
+            return LowLevelCall.returnData();
+        } else if (success) {
+            revert AddressEmptyCode(target);
+        } else if (LowLevelCall.returnDataSize() > 0) {
+            LowLevelCall.bubbleRevert();
+        } else {
+            revert Errors.FailedCall();
+        }
```
james-toussaint, line 92: This saves a few more gas compared to [your numbers here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5094#issuecomment-3072461643).
Same pattern in `functionCallWithValue(..)` & `functionStaticCall(..)` saves gas.
```suggestion
        if (LowLevelCall.callNoReturn(target, value, data)) {
            if (LowLevelCall.returnDataSize() > 0 || target.code.length > 0) {
                return LowLevelCall.returnData();
            }
            revert AddressEmptyCode(target);
        }
        if (LowLevelCall.returnDataSize() > 0) {
            LowLevelCall.bubbleRevert();
        } else {
            revert Errors.FailedCall();
        }
```

contracts/utils/Address.sol
```diff
@@ -76,26 +80,50 @@ library Address {
         if (address(this).balance < value) {
             revert Errors.InsufficientBalance(address(this).balance, value);
         }
-        (bool success, bytes memory returndata) = target.call{value: value}(data);
-        return verifyCallResultFromTarget(target, success, returndata);
```
james-toussaint, line 80: I think `verifyCallResultFromTarget(..)` is not used internally anymore (and not tested).
We can eventually in 5.x add unit tests for it (and deprecate usage until 6.0 removal if it is considered to be useless for developer, which am not sure about)

contracts/utils/Address.sol
```diff
@@ -34,10 +35,13 @@ library Address {
...
-        (bool success, bytes memory returndata) = recipient.call{value: amount}("");
-        if (!success) {
-            _revert(returndata);
+        if (LowLevelCall.callNoReturn(recipient, amount, "")) {
+            // call successful, nothing to do
+            return;
+        } else if (LowLevelCall.returnDataSize() == 0) {
+            revert Errors.FailedCall();
+        } else {
+            LowLevelCall.bubbleRevert();
```
james-toussaint: For consistency purposes with other functions only, how about swapping reverts order:
```suggestion
        }
        if (LowLevelCall.returnDataSize() > 0) {
            LowLevelCall.bubbleRevert();
        } else {
            revert Errors.FailedCall();
        }
```

contracts/utils/Address.sol
```diff
@@ -76,26 +80,50 @@ library Address {
         if (address(this).balance < value) {
```
james-toussaint, line 80: How about using `functionCallWithValue(..)` instead of `verifyCallResult(..)` in Governor & Timelock?

contracts/utils/Address.sol
```diff
@@ -138,10 +166,7 @@ library Address {
     function _revert(bytes memory returndata) private pure {
```
james-toussaint: In the current state, this private `_revert(..)` function can be removed/merged in the calling function.

contracts/utils/Address.sol
Amxx: I personally find it less readable. How must does it save ? I want cases do we save ?

Currently the "happy path" (the one that leads to `return LowLevelCall.returnData();`, and that we want to optimize), has a single `if`. Your proposal adds a second `if` to that path. I'm worried we might be optimising error costs at the expense of the happy path.

contracts/utils/Address.sol
Amxx: this function is called 2 times. Are you saying we should manually inline/duplicate ?

contracts/utils/Address.sol
Amxx: `functionCallWithValue` will revert with `AddressEmptyCode` if you try to do a call to an EOA. If we use it in Governor/Timelock, then they would not be able to send ETH to an EOA. We have a test for that.

Amxx (approved):

Amxx (approved):

contracts/utils/LowLevelCall.sol
```diff
@@ -0,0 +1,126 @@
...
+            mstore(0x40, add(result, add(0x20, returndatasize())))
+        }
+    }
+
+    /// @dev Revert with the return data from the last call.
+    function bubbleRevert() internal pure {
+        assembly ("memory-safe") {
+            let fmp := mload(0x40)
+            returndatacopy(fmp, 0, returndatasize())
+            revert(fmp, returndatasize())
```
james-toussaint, line 117: Should we increase/update free memory pointer before reverting the error (which I would call `revert(result, returndatasize())` in my comment) in case the revert is caught by the client and execution flow continues?

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -461,6 +461,49 @@ await instance.multicall([
...
+function example(address target, bytes memory data) internal {
+    bool success;
+    bytes32 result1;
+    bytes32 result2;
+
+    // Ignore return data
+    success = target.callNoReturn(data);
+
+    // Extract single 32-byte value
+    (success, result1, ) = target.callReturn64Bytes(data);
```
james-toussaint, line 483: I like it that way tbh, just to make sure we wouldn't want to propose the following in addition to this `callReturn64Bytes(data)`:
- `(success, result) = target.callReturnFirst32Bytes(data);`
- `(success, result) = target.callReturnSecond32Bytes(data);`

test/utils/Address.test.js
```diff
@@ -5,6 +5,9 @@ const { PANIC_CODES } = require('@nomicfoundation/hardhat-chai-matchers/panic');
 
 const coder = ethers.AbiCoder.defaultAbiCoder();
 
+const fakeContract = { interface: ethers.Interface.from(['error SomeCustomErrorWithoutArgs()']) };
+const returndata = fakeContract.interface.encodeErrorResult('SomeCustomErrorWithoutArgs');
```
james-toussaint, line 9: For user "convenience", would it be a shame to add an almost no-op function in the _original_ lib contract which only reverts the error? (not a big fan though)

test/utils/LowLevelCall.test.js
```diff
@@ -0,0 +1,220 @@
...
+      it("calls the requested function and returns false if the caller doesn't have enough balance", async function () {
+        const tx = this.mock.$callReturn64Bytes(
+          this.target,
+          ethers.Typed.uint256(value),
+          this.target.interface.encodeFunctionData('mockFunctionWithArgsReturn', [returnValue1, returnValue2]),
+        );
+        await expect(tx).to.changeEtherBalances([this.mock, this.target], [0n, 0n]);
+        await expect(tx)
+          .to.emit(this.mock, 'return$callReturn64Bytes_address_uint256_bytes')
+          .withArgs(false, ethers.ZeroHash, ethers.ZeroHash);
```
james-toussaint, line 102: Should we add tests where `success` is false but `result1&2` are not empty?

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -84,16 +86,17 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
             abi.encodeCall(IERC20Metadata.decimals, ())
         );
-        if (success && encodedDecimals.length >= 32) {
-            uint256 returnedDecimals = abi.decode(encodedDecimals, (uint256));
-            if (returnedDecimals <= type(uint8).max) {
-                return (true, uint8(returnedDecimals));
-            }
-        }
-        return (false, 0);
+        Memory.setFreeMemoryPointer(ptr);
```
james-toussaint, line 94: Why do we take care of fmp after this low level call but not really after others low level calls?

contracts/utils/LowLevelCall.sol
Amxx: the revert data goes into the returndata (just like a return). It doesn't stay in memory, so no need to allocate it.

docs/modules/ROOT/pages/utilities.adoc
Amxx: I think we should avoid having to many variants :
- reduced discoverability for user
- more code to maintain for us

test/utils/Address.test.js
Amxx: I tried that, but unfortunatelly it does work like that.

Lets assume we have this library
```solidity 
library Lib {
    error SomeError();
    function selector() internal pure returns (bytes4) { return SomeError.selector;}
    function trigger() internal pure { revert SomeError(); }
}
```

and the following contracts:

```solidity
contract Test1 {
    function selector() public pure returns (bytes4) {
        return Lib.selector();
    }
}
contract Test2 is Test1 {
    function trigger() public pure {
        Lib.trigger();
    }
}
```

The compiler will include the error definition in `Test2` (because the error is reachable in Test2.trigger -> Lib.trigger).
The compiler will NOT include the error definition in Test1, because it doesn't see how the error could be triggered by calling Test1 functions.

So if we were to make the error part of the contract, we need the user to use that no-op function somewhere in there contract. 

Note that if we change the visibility of Test2.trigger to internal or private, the error is no longer part of the ABI.

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: Not sure about the other places, but in this particular case, we do it to dealocate the memory that was allocated by the `abi.encodeCall(IERC20Metadata.decimals, ())`

contracts/utils/LowLevelCall.sol
Amxx: Basically
- the called contract either does a return (success) or revert (failure) with some memory location (in its context)
- in both cases, the result is copied (by the evm) to "returndata"
- we go back to the context of the caller contract, they receive a boolean (success/faillure) and access to the returndata.

Note that when you do a call, you chagne the execution context, and that means you are in a "new memory space". This makes sure that the called contract cannot interfer with the stack and memory of the caller.

I invite you to run the following code on remix:

```solidity
contract T1 {
    function getFMP() public pure returns (uint256 fmp) {
        assembly ("memory-safe") {
            fmp := mload(0x40)
        }
    }

    function allocateAndGetFMP(uint256 size) public pure returns (uint256 fmp) {
        bytes memory buffer = new bytes(size);
        buffer;
        return getFMP(); // this cannot be catched (local jump)
    }

        function allocateAndCallGetFMP(uint256 size) public view returns (uint256 fmp) {
        bytes memory buffer = new bytes(size);
        buffer;
        return this.getFMP(); // this is an EVM call that can be catched.
    }
}
```

test/utils/LowLevelCall.test.js
Amxx: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5094/commits/ab88c0fd91501a543c98a6db71eb21d2256f6f46

contracts/utils/LowLevelCall.sol
james-toussaint: Thank you 👍

james-toussaint (dismissed):

PR #5095, opened 2024-06-20
Add remaining fuzzing tests and remove pseudo-random fuzzing

### LINKED ISSUE #4894: Remove pseudo-fuzzing tests in hardhat in favor of foundry fuzzing.
Some unit tests are using loops generating random values. In some cases, this could be replaced with proper fuzzing. There are even cases where the fuzzing tests have already been written.

I propose we remove the random and/or looped unit tests when an equivalent fuzzing tests exists.

cairoeth:
Fixes #4894 by removing unit tests that are already equivalently running with fuzzing via Foundry and adding remaining fuzzing unit tests. 

#### PR Checklist

- [ ] Checkpoints
- [ ] CircularBuffer
- [ ] AccessControlEnumerable
- [ ] GovernorVotesQuorumFraction
- [x] GovernorCountingFractional
- [x] AuthorityUtils
- [x] CREATE2
- [x] ECDSA
- [x] Strings
- [ ] ERC20Votes
- [ ] ERC20Snapshot
- [ ] ERC721Votes

<br>

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/access/manager/AuthorityUtils.test.js
+++ b/test/access/manager/AuthorityUtils.test.js
@@ -64,24 +64,6 @@ describe('AuthorityUtils', function () {
       });
     });
 
-    describe('when authority replies with a delay', function () {
-      beforeEach(async function () {
-        this.authority = this.authorityDelayMock;
-      });
-
-      for (const immediate of [true, false]) {
-        for (const delay of [0n, 42n]) {
-          it(`returns (immediate=${immediate}, delay=${delay})`, async function () {
-            await this.authority._setImmediate(immediate);
-            await this.authority._setDelay(delay);
-            const result = await this.mock.$canCallWithDelay(this.authority, this.user, this.other, '0x12345678');
-            expect(result.immediate).to.equal(immediate);
-            expect(result.delay).to.equal(delay);
-          });
-        }
-      }
-    });
-
     describe('when authority replies with empty data', function () {
       beforeEach(async function () {
         this.authority = this.authorityNoResponse;
--- a/test/token/ERC721/extensions/ERC721Consecutive.test.js
+++ b/test/token/ERC721/extensions/ERC721Consecutive.test.js
@@ -142,12 +142,6 @@ describe('ERC721Consecutive', function () {
       describe('ERC721 behavior', function () {
         const tokenId = offset + 1n;
 
-        it('core takes over ownership on transfer', async function () {
-          await this.token.connect(this.alice).transferFrom(this.alice, this.receiver, tokenId);
-
-          expect(await this.token.ownerOf(tokenId)).to.equal(this.receiver);
-        });
-
         it('tokens can be burned and re-minted #1', async function () {
           await expect(this.token.connect(this.alice).$_burn(tokenId))
             .to.emit(this.token, 'Transfer')
--- a/test/utils/Strings.test.js
+++ b/test/utils/Strings.test.js
@@ -108,45 +108,6 @@ describe('Strings', function () {
     });
   });
 
-  describe('addresses', function () {
-    const addresses = [
-      '0xa9036907dccae6a1e0033479b12e837e5cf5a02f', // Random address
-      '0x0000e0ca771e21bd00057f54a68c30d400000000', // Leading and trailing zeros
-      // EIP-55 reference
-      '0x5aAeb6053F3E94C9b9A09f33669435E7Ef1BeAed',
-      '0xfB6916095ca1df60bB79Ce92cE3Ea74c37c5d359',
-      '0xdbF03B407c01E7cD3CBea99509d93f8DDDC8C6FB',
-      '0xD1220A0cf47c7B9Be7A2E6BA89F429762e7b9aDb',
-      '0xfb6916095ca1df60bb79ce92ce3ea74c37c5d359',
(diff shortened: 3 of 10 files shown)


test/utils/math/SignedMath.test.js
Amxx, line 1: We need to check that this doesn't reduce coverage. It may be necessary to keep at least one test. 
(foundry tests are not registered when doing coverage)

test/utils/cryptography/ECDSA.t.sol
```diff
@@ -0,0 +1,28 @@
...
+contract ECDSATest is Test {
+    function testRecoverWithValidSignature(string calldata seed, string calldata message) public {
+        (address signer, uint256 key) = makeAddrAndKey(seed);
+
+        bytes32 hash = keccak256(abi.encode(message));
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(key, hash);
+
+        bytes memory signature = abi.encodePacked(r, s, v);
+        assertEq(signer, ECDSA.recover(hash, signature));
+    }
```
Amxx, line 17:
```suggestion
    function testRecoverWithValidSignature(string calldata seed, bytes32 digest) public {
        (address signer, uint256 key) = makeAddrAndKey(seed);
        (uint8 v, bytes32 r, bytes32 s) = vm.sign(key, digest);

        bytes memory signature = abi.encodePacked(r, s, v);
        assertEq(signer, ECDSA.recover(digest, signature));
    }
```

test/utils/cryptography/ECDSA.t.sol
```diff
@@ -0,0 +1,28 @@
...
+
+    function testRecoverWithWrongMessage(string calldata seed, string calldata message, bytes32 digest) public {
+        (address signer, uint256 key) = makeAddrAndKey(seed);
+
+        bytes32 validDigest = keccak256(abi.encode(message));
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(key, validDigest);
+
+        bytes memory signature = abi.encodePacked(r, s, v);
+        assertTrue(signer != ECDSA.recover(digest, signature));
+    }
```
Amxx, line 27:
```suggestion
    function testRecoverWithWrongMessage(string calldata seed, bytes32 digest, bytes32 otherDigest) public {
        vm.assume(digest != otherDigest);
        
        (address signer, uint256 key) = makeAddrAndKey(seed);
        (uint8 v, bytes32 r, bytes32 s) = vm.sign(key, digest);

        bytes memory signature = abi.encodePacked(r, s, v);
        assertFalse(signer == ECDSA.recover(otherDigest, signature));
    }
```

test/utils/Strings.test.js
Amxx, line 1: We need to check that this doesn't reduce coverage. It may be necessary to keep at least one test.
(foundry tests are not registered when doing coverage)

Note: I like that unit test cover the EIP-55 references specifically

test/token/ERC721/extensions/ERC721Consecutive.test.js
```diff
@@ -142,12 +142,6 @@ describe('ERC721Consecutive', function () {
       describe('ERC721 behavior', function () {
         const tokenId = offset + 1n;
 
-        it('core takes over ownership on transfer', async function () {
-          await this.token.connect(this.alice).transferFrom(this.alice, this.receiver, tokenId);
-
-          expect(await this.token.ownerOf(tokenId)).to.equal(this.receiver);
-        });
-
```
Amxx, line 150: why is this test removed ?

test/access/manager/AuthorityUtils.test.js
```diff
@@ -64,24 +64,6 @@ describe('AuthorityUtils', function () {
...
-            await this.authority._setImmediate(immediate);
-            await this.authority._setDelay(delay);
-            const result = await this.mock.$canCallWithDelay(this.authority, this.user, this.other, '0x12345678');
-            expect(result.immediate).to.equal(immediate);
-            expect(result.delay).to.equal(delay);
-          });
-        }
-      }
-    });
-
```
Amxx, line 84: This test is not so much fuzzing as it is testing 4 clearly different cases. Not saying fuzzing would not cover them, but we don't really need 5000 fuzzing runs to cover the same code.

Also, I'm not sure how that change would affect coverage.

Amxx (commented):
Our guideline what far was:
- unit test first, with as close to 100% coverage as possible
- fuzzing and formal verification is some places.
I still think this applies. 

A direct consequence is that even when there is some fuzzing, we still want to have unit test that cover the code. In particular the corner cases whe have identified (and that we are not 100% sure fuzzing will explore).

The idea being the issue to avoid unit tests that loop over (large amount of) randomly generated values. However, that doesn't mean all loops in the tests should be removed. Following the [swiss chesse model](https://en.wikipedia.org/wiki/Swiss_cheese_model), fuzzing should be an extra layer, but we should not add hole in the unit test layers with the hope that fuzzing covers them.

 My initial impression reading this PR is that the added foundry tests are great ... but I wouldn't necessarily remove (all) the corresponding unit tests. For example, there may to too many different input for testing SignedMath.average, but I still think we should test the combinaison of MinInt256 and MaxInt256 with each other, and with some smaller values.

PR #5097, opened 2024-06-23
Update hashProposal visibility to view

0xDiscotech:
Fixes

Update `hashProposal()` function visibility to `view` instead of `pure` to allow an override using `block.chainid` and/or `address(this)` in order to guarantee more uniqueness on the same proposal but on different networks.

Can be very useful when the proposal is used as a kind of proof to validate the vote, and the same proposal is created on different networks.  This allows the developer to mitigate a replay attack over the same `proposalId` in a clean way by overriding the function to read from the environment or the contract's state.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/popular-horses-tell.md
+++ b/.changeset/popular-horses-tell.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+- update `hashProposal()` visibility to view instead of pure to allow `address(this)` and `block.chainid` to be part of the proposal id computation if more uniqueness between different chains is desired
--- a/.changeset/popular-horses-tell.md
+++ b/.changeset/popular-horses-tell.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-- update `hashProposal()` visibility to view instead of pure to allow `address(this)` and `block.chainid` to be part of the proposal id computation if more uniqueness between different chains is desired
+- Update `hashProposal()` visibility to view instead of pure to allow `address(this)` and `block.chainid` to be part of the `proposalId` computation if more uniqueness between different chains is desired
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -117,10 +117,8 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
      * can be produced from the proposal data which is part of the {ProposalCreated} event. It can even be computed in
      * advance, before the proposal is submitted.
      *
-     * Note that the chainId and the governor address are not part of the proposal id computation. Consequently, the
-     * same proposal (with same operation and same description) will have the same id if submitted on multiple governors
-     * across multiple networks. This also means that in order to execute the same operation twice (on the same
-     * governor) the proposer will have to change the description in order to avoid proposal id conflicts.
+     * Note that the chainId and the governor address are not part of the proposal id computation, but can be added to
+     * guarantee more uniqueness if the same proposal is to submitted on multiple governors across multiple networks.
      */
     function hashProposal(
         address[] memory targets,
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -127,7 +127,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         uint256[] memory values,
         bytes[] memory calldatas,
         bytes32 descriptionHash
-    ) public pure virtual returns (uint256) {
+    ) public view virtual returns (uint256) {
         return uint256(keccak256(abi.encode(targets, values, calldatas, descriptionHash)));
     }
 
--- a/contracts/governance/IGovernor.sol
+++ b/contracts/governance/IGovernor.sol
@@ -207,7 +207,7 @@ interface IGovernor is IERC165, IERC6372 {
         uint256[] memory values,
         bytes[] memory calldatas,
         bytes32 descriptionHash
-    ) external pure returns (uint256);
+    ) external view returns (uint256);
 
     /**
      * @notice module:core


ernestognw (commented):
Hey, thanks for the contribution! Some thoughts

Changing to `view` allows to extend the hash and include `block.chainid` or `addres(this)`, but also enables developers to read from storage, which concerns me more.

As opposed to pure functions, a view function depends on the state, which means that the `proposalId` may unexpectedly change if a state variable is added to the formula. In that case, the proposal may end up stuck (i.e. not deriving the same `proposalId` in `queue` or `execute`)

Though the concern of replayability does make sense in the context of signatures, I would add a nonce to the signature instead.

.changeset/popular-horses-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+- Update `hashProposal()` visibility to view instead of pure to allow `address(this)` and `block.chainid` to be part of the `proposalId` computation if more uniqueness between different chains is desired
```
Amxx, line 5:
```suggestion
- `Governor`: change `hashProposal()` visibility to view to allow `address(this)` and `block.chainid` to be part of the `proposalId` computation if more uniqueness between different chains is desired
```

Amxx:
Thank you @0xDiscotech for this proposal.

I'm not sure what the consequences of changing the proposalId computation would be on UI such as Tally ... but given it is virtual, I don't see why we would allow some changes but not that change.

So its good to me

Amxx:
@0xDiscotech, we discussed this PR internally, and we have some concern with the consequences of using pure vs view.

Using pure would allow change that make the fonction change value independently of the input parameter (that is the whole point of pure vs view). This means reading state variable, reading the time, reading anything that is not constant. 

If this was to happen, then the proposalId may not be consistently computed through the lifecycle of a proposal, and that would be bad. Potentially, it could be used to hide some attack vector that allows a priviledge user to prevent a proposal from being executed. This could be happen by mistake, or intentionally by a malicious dev.

On the other hand, we don't really understand the usecase. If two governors, on the same chain, or on two different chains, are planning to execute the exact same logic, what is the issue using the same identifier for both proposals ? The same proposalId means the same input was proposed. The execution outcome may differ depending on the governor instance doing it, but the input remains the same.

Whe voting on it, the voting scheme (that is based on EIP-721) does already prevent cross-instance / cross-chain replay.

With all that in mind, we see a not very likelly but still very clear risk ... and we don't really see an upside to that.

0xDiscotech:
Thanks for you response @Amxx @ernestognw!

The idea of adding it to view came from a project where our team created a Governor extension interacting with Worldcoin's World ID protocol. In this project, we used the `proposalId` as a key part of the input to verify that a user is a unique human. Since Worldcoin preserves user information and only verifies if the user is human, it always returns the same `nullifierHash` for the input action. It makes sense to use the `proposalId` as input and store that `nullifierHash` to avoid double-spending by the same user. However, you must ensure that the `proposalId` is unique to prevent replay attacks.

In our solution, we had to concatenate the description with a kind of 'signature' using the chain ID and the Governor's address. This breaks the functionality of the `hashProposal()` function because it won't return the exact same `proposalId` after calling `propose()`. You can see our implementation here: [GovernorWorldID.sol](https://github.com/defi-wonderland/world-id-governor-core/blob/dev/src/contracts/GovernorWorldID.sol#L183). The solution was somewhat 'hacky' because we couldn't add that data inside the `hashProposal()` function, which would have been the cleanest way.

This approach might be very specific to the mentioned use case, but it's not uncommon for future Governors to use similar protocols to get proof-of-personhood while preserving user privacy.

In summary, my point is that developers can still modify the logic behind `proposalId` generation. However, restricting this to pure functions makes things more difficult and complex to handle. The benefit of allowing more flexibility in these cases is that it ensures guaranteed uniqueness not only across different chains but also across different DAOs. Currently, the same `proposalId` can be generated in two different DAOs with the same proposal as input.

0xDiscotech:
> Thanks for you response @Amxx @ernestognw!
> 
> The idea of adding it to view came from a project where our team created a Governor extension interacting with Worldcoin's World ID protocol. In this project, we used the `proposalId` as a key part of the input to verify that a user is a unique human. Since Worldcoin preserves user information and only verifies if the user is human, it always returns the same `nullifierHash` for the input action. It makes sense to use the `proposalId` as input and store that `nullifierHash` to avoid double-spending by the same user. However, you must ensure that the `proposalId` is unique to prevent replay attacks.
> 
> In our solution, we had to concatenate the description with a kind of 'signature' using the chain ID and the Governor's address. This breaks the functionality of the `hashProposal()` function because it won't return the exact same `proposalId` after calling `propose()`. You can see our implementation here: [GovernorWorldID.sol](https://github.com/defi-wonderland/world-id-governor-core/blob/dev/src/contracts/GovernorWorldID.sol#L183). The solution was somewhat 'hacky' because we couldn't add that data inside the `hashProposal()` function, which would have been the cleanest way.
> 
> This approach might be very specific to the mentioned use case, but it's not uncommon for future Governors to use similar protocols to get proof-of-personhood while preserving user privacy.
> 
> In summary, my point is that developers can still modify the logic behind `proposalId` generation. However, restricting this to pure functions makes things more difficult and complex to handle. The benefit of allowing more flexibility in these cases is that it ensures guaranteed uniqueness not only across different chains but also across different DAOs. Currently, the same `proposalId` can be generated in two different DAOs with the same proposal as input.

Also we could use the `address(this)` and `block.chainid` as the default implementation of `hashProposal()`, but I don't know if it something you are interested in. Maybe I can open a new PR with that change (only in case you agree on moving forward with the  `pure` to `view` modification).

Amxx:
> Also we could use the `address(this)` and `block.chainid` as the default implementation of `hashProposal()`, but I don't know if it something you are interested in.

This is something that was discussed in the past, and was rejected. We were concerned that changing the hash computation could break tools such as Tally, and we did not see a clear advantage (as discussed above). It would also cause issues if a governor contract is upgraded while proposals are active.

It is not inconceivable that we make the function view, so that overrides can use address(this) ... but for sure we are not going to change the default function.

0xDiscotech:
> > Also we could use the `address(this)` and `block.chainid` as the default implementation of `hashProposal()`, but I don't know if it something you are interested in.
> 
> This is something that was discussed in the past, and was rejected. We were concerned that changing the hash computation could break tools such as Tally, and we did not see a clear advantage (as discussed above). It would also cause issues if a governor contract is upgraded while proposals are active.
> 
> It is not inconceivable that we make the function view, so that overrides can use address(this) ... but for sure we are not going to change the default function.

Awesome, makes sense. Good to know 👍

ernestognw:
> The idea of adding it to view came from a project where our team created a Governor extension interacting with Worldcoin's World ID protocol. In this project, we used the proposalId as a key part of the input to verify that a user is a unique human

This is an interesting insight. I guess the problem here is that we never intended `proposalId` to be something "non-replayable" outside of the context of this contract. The security assumptions here indicate that there shouldn't be an issue, though using the `proposalId` in the way you describe definitely requires replayability protection.

From your code I see you added [some information to the returned proposalId](https://github.com/defi-wonderland/world-id-governor-core/blob/dev/src/contracts/GovernorWorldID.sol#L183). That's fine in my opinion since it's not altering the internal assumptions the Governor does about the `proposalId`.

I would be fine making the function `view`, it's just that most of the lifecycle of the governor assumes that the proposal id function is a `pure` function (in the sense that it doesn't have any side effect), and allowing to remove such guarantee sounds like a big commitment imo

0xDiscotech:
> That's fine in my opinion since it's not altering the internal assumptions the Governor does about the proposalId.
Yes, even though that approach works, the off-chain interaction is affected since `hashProposal()` won't return the same `proposalId` as `_propose()`.

My point of view is that by setting it to `view`, you provide more flexibility to cover scenarios requiring more uniqueness. This allows developers to implement it straightforwardly without breaking any off-chain interaction.

That said, you make a good point. While I maintain my perspective, I respect and understand any decision you make 🫡

contracts/governance/Governor.sol
```diff
@@ -117,17 +117,15 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
      * can be produced from the proposal data which is part of the {ProposalCreated} event. It can even be computed in
      * advance, before the proposal is submitted.
      *
-     * Note that the chainId and the governor address are not part of the proposal id computation. Consequently, the
-     * same proposal (with same operation and same description) will have the same id if submitted on multiple governors
-     * across multiple networks. This also means that in order to execute the same operation twice (on the same
-     * governor) the proposer will have to change the description in order to avoid proposal id conflicts.
+     * Note that the chainId and the governor address are not part of the proposal id computation, but can be added to
+     * guarantee more uniqueness if the same proposal is to submitted on multiple governors across multiple networks.
```
ernestognw, line 121: I feel some context in the previous comment was relevant, let's do:

```suggestion
     * Note that the chainId and the governor address are not part of the proposal id computation. Consequently, the
     * same proposal will share its id with another governor in a different network where the same proposal was
     * submitted. We recommend using the description field as a discriminator to avoid proposal id conflicts.
     * Similarly, developers can add both `chainId` and `address(this)` to the hash computation by overriding
     * this function.
     *
     * IMPORTANT: Overriding this function may have unexpected effects after an upgrade. Consider to verify
     * that the new implementation doesn't allow a proposer to forge an id.
```

ernestognw:
Hi @0xDiscotech,

We've been discussing this change since #5290 came up as a requirement for some other DAOs too. Those changes may likely end up being even more aggressive than this visibility change so we might get this merged before.

0xDiscotech:
> Hi @0xDiscotech,
> 
> We've been discussing this change since #5290 came up as a requirement for some other DAOs too. Those changes may likely end up being even more aggressive than this visibility change so we might get this merged before.

Hi @ernestognw,
Thanks for letting me know, this sounds good!
Do you need some change from side on the code?

ernestognw:
> Hi @ernestognw, Thanks for letting me know, this sounds good! Do you need some change from side on the code?

All good, just an update :D

arr00:
Just wanted to leave an update here--we are currently leaning towards just merging #5290 which moves in the direction of deprecating `hashProposal` eventually in favor of a `view` function named `getProposalId`.

PR #5098, opened 2024-06-24
Add Foundry tests to coverage

cairoeth:
Adds Foundry tests to coverage and allows Codecov to combine both reports. The `coverage` script command removes zero hits in the Foundry report so that both reports can be merged exclusively, otherwise Codecov will mark lines as not covered, when in reality they are covered by Hardhat.

It increases coverage by 0.07% overall, 0.98% in Governor and 0.65% in Clones. Time increase for workflow is very small.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -91,8 +91,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up environment
         uses: ./.github/actions/setup
-      - name: Install LCOV
-        run: sudo apt-get install lcov
       - name: Run coverage
         run: npm run coverage
       - uses: codecov/codecov-action@v4
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -91,6 +91,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up environment
         uses: ./.github/actions/setup
+      - name: Install LCOV
+        run: sudo apt-get install lcov
       - name: Run coverage
         run: npm run coverage
       - uses: codecov/codecov-action@v4
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "compile": "hardhat compile",
     "compile:harnesses": "env SRC=./certora/harnesses hardhat compile",
-    "coverage": "env COVERAGE=true hardhat coverage && FOUNDRY_FUZZ_RUNS=10 forge coverage --report lcov && lcov -o lcov.info --remove lcov.info 'test/*' 'contracts/mocks/*' && awk '!/,0/' lcov.info > temp && mv temp lcov.info",
+    "coverage": "scripts/checks/coverage.sh",
     "docs": "npm run prepare-docs && oz-docs",
     "docs:watch": "oz-docs watch contracts docs/templates docs/config.js",
     "prepare": "git config --local core.hooksPath .githooks",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "compile": "hardhat compile",
     "compile:harnesses": "env SRC=./certora/harnesses hardhat compile",
-    "coverage": "env COVERAGE=true hardhat coverage",
+    "coverage": "env COVERAGE=true hardhat coverage && forge coverage --report lcov --ir-minimum",
     "docs": "npm run prepare-docs && oz-docs",
     "docs:watch": "oz-docs watch contracts docs/templates docs/config.js",
     "prepare": "git config --local core.hooksPath .githooks",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "compile": "hardhat compile",
     "compile:harnesses": "env SRC=./certora/harnesses hardhat compile",
-    "coverage": "env COVERAGE=true hardhat coverage && forge coverage --report lcov --ir-minimum && lcov -o lcov.info --remove lcov.info --rc lcov_branch_coverage=1 --rc lcov_function_coverage=1 'test/*' 'contracts/mocks/*'",
+    "coverage": "env COVERAGE=true hardhat coverage && forge coverage --report lcov && lcov -o lcov.info --remove lcov.info --rc lcov_branch_coverage=1 --rc lcov_function_coverage=1 'test/*' 'contracts/mocks/*'",
     "docs": "npm run prepare-docs && oz-docs",
     "docs:watch": "oz-docs watch contracts docs/templates docs/config.js",
     "prepare": "git config --local core.hooksPath .githooks",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "compile": "hardhat compile",
     "compile:harnesses": "env SRC=./certora/harnesses hardhat compile",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "compile": "hardhat compile",
     "compile:harnesses": "env SRC=./certora/harnesses hardhat compile",
-    "coverage": "env COVERAGE=true hardhat coverage",
+    "coverage": "env COVERAGE=true hardhat coverage && FOUNDRY_FUZZ_RUNS=10 forge coverage --report lcov && lcov -o lcov.info --remove lcov.info 'test/*' 'contracts/mocks/*' && awk '!/,0/' lcov.info > temp && mv temp lcov.info",
```
Amxx, line 13: If possible I'd like to avoid having to much logic in the package.json

Is it possible to have a specific profile in the `foundry.toml` file that specify this number of runs ? IMO it would be better if we could remove it from this line.

Also, I'm wondering it we should put that logic in a `./script/coverage.js`.

Amxx: [The documentation](https://book.getfoundry.sh/reference/cli/forge/coverage?highlight=coverage#forge-coverage) mentions a `--ir-minimum` option that I suspect will be needed when doing founcry coverage or P256 (that is already at the limit of stack too deep).

Amxx:
I tried running this, I'm getting an error
> lcov: ERROR: (inconsistent) "contracts/token/ERC721/extensions/ERC721Royalty.sol":24:  function ERC721Royalty.supportsInterface found on line but no corresponding 'line' coverage data point.  Cannot derive function end line.  See lcovrc man entry for 'derive_function_end_line'.
	(use "lcov --ignore-errors inconsistent ..." to bypass this error)

Also, I'm missing how this is working. Should the `forge coverage` and "lcov processing" be done BEFORE the `hardhat coverage` so that both reports can be merged ?

package.json
cairoeth: yes, originally I added `--ir-minimum` but I did a test with P256 and it wasn't necessary so I removed it -- maybe we have to bring it back, although it increases time by a good amount.

cairoeth:
> I tried running this, I'm getting an error
> 
> > lcov: ERROR: (inconsistent) "contracts/token/ERC721/extensions/ERC721Royalty.sol":24:  function ERC721Royalty.supportsInterface found on line but no corresponding 'line' coverage data point.  Cannot derive function end line.  See lcovrc man entry for 'derive_function_end_line'.
> > (use "lcov --ignore-errors inconsistent ..." to bypass this error)
> 
> I'm using `LCOV version 2.1-1`
> 
> Also, I'm missing how this is working. Should the `forge coverage` and "lcov processing" be done BEFORE the `hardhat coverage` so that both reports can be merged ?

the lcov processing is only needed for the foundry report, so it has to be done after that. i also don't like that the script command in `package.json` is very long and complicated. maybe we can move it to a sh script and call the script in `package.json`?

Amxx:
> maybe we can move it to a sh script and call the script in package.json?

yes

Amxx:
> the lcov processing is only needed for the foundry report, so it has to be done after that

That I inderstand, but afdter that processing nothing gets done with the `lcov.info`. I was thinking that for the hardhat task to merge the report produced by foundry (and post processed) ... then the `harhdat coverage` run should be done at the end.

cairoeth:
> nothing gets done with the `lcov.info`

the workflow uploads `lcov.info` and the other coverage reports to codecov automatically, which then [merges it](https://about.codecov.io/product/feature/report-merging/). solidity-coverage doesn't support merging Hardhat and Foundry reports sadly (https://github.com/sc-forks/solidity-coverage/issues/734).

Amxx:
> > nothing gets done with the `lcov.info`
> 
> the workflow uploads `lcov.info` and the other coverage reports to codecov automatically, which then [merges it](https://about.codecov.io/product/feature/report-merging/). solidity-coverage doesn't support merging Hardhat and Foundry reports sadly ([sc-forks/solidity-coverage#734](https://github.com/sc-forks/solidity-coverage/issues/734)).

Hoooo ... then I don't think the foundry part should run when doing coverage locally on our machine (unless we find a way to report an aggregate).

If we have a script it could distinguish between CI and local, and only run the foundry + post processing is in CI.

scripts/checks/coverage.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+export COVERAGE=true
+export FOUNDRY_FUZZ_RUNS=10
+
+# Hardhat coverage
+npx hardhat coverage
```
Amxx, line 7: npx should not needed here (see `scripts/prepare-docs.sh`)

```suggestion
hardhat coverage
```

scripts/checks/coverage.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Foundry coverage
+forge coverage --report lcov
+# Remove test and mock data
+lcov --rc derive_function_end_line=0 -o lcov.info --remove lcov.info 'test/*' 'contracts/mocks/*'
+# Remove zero hits
+awk '!/,0/' lcov.info > temp && mv temp lcov.info
```
Amxx, line 14: similar to `.githooks/pre-push` (with == instead of !=)
```suggestion
if [ "${CI:-"false"}" == "true" ]; then
  # Remove test and mock data
  lcov --rc derive_function_end_line=0 -o lcov.info --remove lcov.info 'test/*' 'contracts/mocks/*'
  # Remove zero hits
  awk '!/,0/' lcov.info > temp && mv temp lcov.info
fi
```

Amxx, line 14: similar to `.githooks/pre-push` (with == instead of !=)
```suggestion
if [ "${CI:-"false"}" == "true" ]; then
  # Foundry coverage
  forge coverage --report lcov
  # Remove test and mock data
  lcov --rc derive_function_end_line=0 -o lcov.info --remove lcov.info 'test/*' 'contracts/mocks/*'
  # Remove zero hits
  awk '!/,0/' lcov.info > temp && mv temp lcov.info
fi
```

Amxx: Please note the "inplace" deletion using sed. IMO that is clearer/cleaner than using awk with a temp file.

scripts/checks/coverage.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
```
Amxx, line 2:
```suggestion

set -euo pipefail

```

.github/workflows/checks.yml
```diff
@@ -91,6 +91,8 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up environment
         uses: ./.github/actions/setup
+      - name: Install LCOV
+        run: sudo apt-get install lcov
```
ernestognw, line 95: I'm afraid this is adding a dependency to the CI and we should be very careful about it. Is there any way to verify `lcov` is trusted?

ernestognw (approved):
LGTM, thanks @cairoeth

On a related topic, I'm intrigued by the increase in coverage for `Clones.sol`, it seems like it's hitting a case I never knew how to replicate in unit tests. [See here](https://app.codecov.io/gh/OpenZeppelin/openzeppelin-contracts/pull/5098/indirect-changes#623e1d12d421269163056c346fe856d7-R90)

PR #5099, opened 2024-06-26
Clarify hardhat & foundry configs and enable optimisation by default.

Amxx:
### Before this PR:

- Local tests on our machine run in development mode without `GAS` or `COVERAGE` (default). This results in:
  - optimization disabled
  - unlimited contract size
- Non upgradeable tests on the CI set the GAS flag (to produce a report). This results in:
  - optimization enabled
    - code size warning
  - unlimited contract size
- Coverage tests on the CI set the COVERAGE flag. This results in:
  - optimization enabled
    - code size warning
  - unlimited contract size
  - some additional stuff (unused params, initialBaseFeePerGas)
- Upgradeable tests do not set any flag. They run in development mode. This results in:
  - optimization disabled
  - unlimited contract size

All this was not necessarily clear. In particular, the fact that by default, out local tests run in a mode that doesnt match the CI (unless we have some `.env` to override that). Also, setting the `IR` flag has no effect unless you also enable optimization (either through the production mode, or implicitelly when doing gas or coverage checks)

### This PR does the following changes:

- All compilation is done with optimization. That is what our users should all use in production, so that is what we should test.
- Make code size unlimited by default. That is what we had previously, but it was "hidden". This PR makes it obvious what is happening here.
- Gas no longer affect the compiler settings, it only enables/disables the gas module.
- No changes to the config specific to coverage

### Note

- This PR **does not** change the compiler settings during CI, so the gas reports continue to be valid. 
- This PR **does** change the compiler settings when running tests locally.

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -1,12 +1,12 @@
 /// ENVVAR
-// - COMPILE_VERSION:   compiler version (default: 0.8.20)
-// - SRC:               contracts folder to compile (default: contracts)
-// - COMPILE_MODE:      production modes enables optimizations (default: development)
-// - IR:                enable IR compilation (default: false)
-// - COVERAGE:          enable coverage report
-// - ENABLE_GAS_REPORT: enable gas report
-// - COINMARKETCAP:     coinmarkercat api key for USD value in gas report
-// - CI:                output gas report to file instead of stdout
+// - COMPILER:      compiler version (default: 0.8.24)
+// - SRC:           contracts folder to compile (default: contracts)
+// - RUNS:          number of optimization runs (default: 200)
+// - IR:            enable IR compilation (default: false)
+// - COVERAGE:      enable coverage report (default: false)
+// - GAS:           enable gas report (default: false)
+// - COINMARKETCAP: coinmarketcap api key for USD value in gas report
+// - CI:            output gas report to file instead of stdout
 
 const fs = require('fs');
 const path = require('path');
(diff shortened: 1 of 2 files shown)


hardhat.config.js
```diff
@@ -25,11 +25,10 @@ const { argv } = require('yargs/yargs')()
...
     },
-    mode: {
-      alias: 'compileMode',
-      type: 'string',
-      choices: ['production', 'development'],
-      default: 'development',
+    runs: {
+      alias: 'optimizationRuns',
+      type: 'number',
+      default: 200,
```
ernestognw, line 31: Any reason for picking 200 in particular? Feels pretty arbitrary and I'd rather just do 1.
Ideally we should have tests and benchmarks with both few and a lot of runs

ernestognw (commented):
This is way cleaner, thanks! Just left a qustion

hardhat.config.js
Amxx: I think 200 is the canonical value that "most" project use. I don't know where that value comes from exactly, but it has been around since I started using truffle in 2017. 

For example, its the default value used by remix.

Amxx: That is also the value we used before when running with optimization (for example when doing gas comparaison).

cairoeth: it's also the default value of `solc --optimize`: https://docs.soliditylang.org/en/latest/using-the-compiler.html#optimizer-options

cairoeth:
should we also clarify `foundry.toml`, like the optimizer runs? it has the same default behavior already.

cairoeth (approved):
lgtm

ernestognw (approved):

PR #5101, opened 2024-07-01
Treat code-size warnings as errors

frangio:
Follow up to #5099.

The current config is treating code size as a warning (`true` is equivalent to `'warn'`), which would not cause CI to fail. This config removes the line so that it's considered an error (according to `default: 'error'` below).

Feel free to close this if it's not the desired behavior.


Amxx:
I'd say this is desirable, unless it triggers when compiling the "coverage" contract that are way bigger (because of the added events).

Amxx (approved):

PR #5102, opened 2024-07-01
Clones.sol: Changed '96 bits' to '12 bytes'

0xScratch:
This change is made to make a consistent usage of units within the comments which makes it easy for devs to relate what's going on (Although, I understand it's not a rocket science to convert 96 bits to 12 bytes for a dev unless he doesn't know much about it...but still we got the point)

While going through the code myself, found it puzzling why `bits` is included in that comment despite the usage of `bytes` 2 times in the same comment. Maybe it's done by mistake or there could be an important reason behind this which I might not be aware of but still we devs find `bytes` more promising than `bits` as far as I am concerned!


Vectorized:
@0xScratch I was the one who wrote the comment.

This is because the `shl` opcode takes in the number of bits to shift, not the number of bytes.

The nomenclature of cleaning upper bits vs upper bytes is chosen to be more similar with the Solidity documentation
https://docs.soliditylang.org/en/latest/internals/variable_cleanup.html.

0xScratch:
Got it, Thanks @Vectorized

PR #5109, opened 2024-07-10
Add clone variant with per-instance immutable arguments

Amxx:
#### Notes about EOF:

- The clones are not EOF objects. They are legacy contract. As a consequence, they can be used with either EOF or Legacy implementations. 
- Legacy implementations are not an issue. The clone being a legacy contract, it is observable.
- EOF implementations will not be able to read the params directly. They could however do it indirectly by doing a static call on an helper (Legacy) contract, that would do an `extcodecopy` on the caller.

```solidity
/// compiled in legacy mode
contract ImmutableArgsReader {
    function selfFetchCloneArgs() external returns (bytes memory) {
        return Clones.fetchCloneArgs(msg.sender); 
    }
    
    function fetchCloneArgs(address instance) external returns (bytes memory) {
        return Clones.fetchCloneArgs(instance);
    }
}

/// compiled in EOF 
contract MyEOFImplementation {
    ImmutableArgsReader private immutable immutableArgsReader = ImmutableArgsReader(...);
    
    function somefunction(...) public {
       (uint256 arg1, address arg2, ...) = abi.decode(immutableArgsReader.selfFetchCloneArgs(), (uint256, address, ...));
       //...
    }
}
```

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -229,12 +229,12 @@ library Clones {
      * - If `instance` is NOT a clone deployed using this library, the behavior is undefined. This
      *   function should only be used to check addresses that are known to be clones.
      */
-    function fetchCloneArgs(address instance) internal view returns (bytes memory result) {
-        uint256 argsLength = instance.code.length - 0x2d; // revert if length is too short
-        result = new bytes(argsLength);
+    function fetchCloneArgs(address instance) internal view returns (bytes memory) {
+        bytes memory result = new bytes(instance.code.length - 0x2d); // revert if length is too short
         assembly ("memory-safe") {
-            extcodecopy(instance, add(result, 0x20), 0x2d, argsLength)
+            extcodecopy(instance, add(result, 0x20), 0x2d, mload(result))
         }
+        return result;
     }
 
     /**
@@ -243,18 +243,19 @@ library Clones {
      * An assembly variant of this function requires copying the `args` array, which can be efficiently done using
      * `mcopy`. Unfortunately, that opcode is not available before cancun. A pure solidity implementation using
      * abi.encodePacked is more expensive but also more portable and easier to review.
+     *
+     * NOTE: https://eips.ethereum.org/EIPS/eip-3860[EIP-3860] limits the length of the `initcode` to 49152 bytes.
+     * With the proxy code taking 45 bytes, that limits the length of the immutable args to 49107 bytes.
      */
     function _cloneWithImmutableArgsCode(
         address implementation,
         bytes memory args
     ) private pure returns (bytes memory) {
-        uint256 initCodeLength = args.length + 0x2d;
-        // initcode is limited to 49152 bytes as per https://eips.ethereum.org/EIPS/eip-3860[EIP-3860]
-        if (initCodeLength > 0xc000) revert ImmutableArgsTooLarge();
+        if (args.length > 49107) revert ImmutableArgsTooLarge();
         return
             abi.encodePacked(
                 hex"61",
-                uint16(initCodeLength),
+                uint16(args.length + 0x2d),
                 hex"3d81600a3d39f3363d3d373d3d3d363d73",
                 implementation,
                 hex"5af43d82803e903d91602b57fd5bf3",
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -3,6 +3,7 @@
 
 pragma solidity ^0.8.20;
 
+import {Create2} from "../utils/Create2.sol";
 import {Errors} from "../utils/Errors.sol";
 
 /**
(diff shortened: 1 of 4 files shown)


contracts/proxy/Clones.sol
```diff
@@ -121,4 +124,142 @@ library Clones {
...
+        return
+            abi.encodePacked(
+                hex"3d61",
+                uint16(initCodeLength),
+                hex"80600b3d3981f3363d3d373d3d3d363d73",
+                implementation,
+                hex"5af43d82803e903d91602b57fd5bf3",
+                args
+            );
+    }
```
Vectorized, line 264: 1 byte less:

```solidity
abi.encodePacked(
    hex"61",
    uint16(initcodeLength),
    hex"3d81600a3d39f3363d3d373d3d3d363d73",
    implementation,
    hex"5af43d82803e903d91602b57fd5bf3",
    args
)
```

https://github.com/Vectorized/solady/blob/2cead9a403ef788b5467e16fc5a166497be4f820/test/LibClone.t.sol#L632

Vectorized:
If you need an ERC1967 minimal proxy with immutable args, this is the formulation of the init code.

```solidity
abi.encodePacked(
    hex"61",
    uint16(args.length + 0x3d),
    hex"3d8160233d3973",
    implementation,
    hex"60095155f3363d3d373d3d363d7f360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc545af43d6000803e6038573d6000fd5b3d6000f3",
    args
)
```

For minimalism, this does not emit the `Upgraded` event on deployment.

contracts/proxy/Clones.sol
```diff
@@ -121,4 +124,142 @@ library Clones {
...
+            mstore(result, argsLength)
+            extcodecopy(instance, add(result, 0x20), 0x2d, argsLength)
+        }
+    }
+
+    /**
+     * @dev Helper that prepares the initcode of the proxy with immutable args.
+     *
+     * An assembly variant of this function requires copying the `args` array, which can be efficiently done using
+     * `mcopy`. Unfortunatelly, that opcode is not available before cancun. A pure solidity implemenation using
```
Amxx, line 246:
```suggestion
     * `mcopy`. Unfortunately, that opcode is not available before cancun. A pure solidity implementation using
```

ernestognw (commented):
Mostly looks good, but I'd rather have `WithImmutableArgs` as a suffix (e.g. `predictDeterministicAddressWithImmutableArgs` instead of `predictWithImmutableArgsDeterministicAddress`)

contracts/proxy/Clones.sol
```diff
@@ -121,4 +124,142 @@ library Clones {
...
+     * - If `instance` is a clone that was deployed using `clone` or `cloneDeterministic`, this
+     *   function will return an empty array.
+     * - If `instance` is a clone that was deployed using `cloneWithImmutableArgs` or
+     *   `cloneWithImmutableArgsDeterministic`, this function will return the args array used at
+     *   creation.
+     * - If `instance` is NOT a clone deployed using this library, the behavior is undefined. This
+     *   function should only be used to check addresses that are known to be clones.
+     */
+    function fetchCloneArgs(address instance) internal view returns (bytes memory result) {
+        uint256 argsLength = instance.code.length - 0x2d; // revert if length is too short
```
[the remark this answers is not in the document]
ernestognw: Solving, I thought it was "too long" and was a missing check

contracts/proxy/Clones.sol
```diff
@@ -121,4 +124,142 @@ library Clones {
...
+     * An assembly variant of this function requires copying the `args` array, which can be efficiently done using
+     * `mcopy`. Unfortunately, that opcode is not available before cancun. A pure solidity implementation using
+     * abi.encodePacked is more expensive but also more portable and easier to review.
+     */
+    function _cloneWithImmutableArgsCode(
+        address implementation,
+        bytes memory args
+    ) private pure returns (bytes memory) {
+        uint256 initCodeLength = args.length + 0x2d;
+        if (initCodeLength > type(uint16).max) revert ImmutableArgsTooLarge();
```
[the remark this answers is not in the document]
Amxx: Interresting. We will never reach the limit of push2.

That leaves 2 options:
- keep the check but change the limit to `0xc000`
- remove the check, and rely on the fact that creation will fail anyway.

I'd go with the first option, just to make sure that `predict` fails if what is being predicted is not deployable.

test/proxy/Clones.test.js
```diff
@@ -43,53 +82,94 @@ describe('Clones', function () {
...
+              .to.emit(this.factory, 'return$cloneDeterministic_address_bytes32')
+              .withArgs(predicted);
+          }
+        });
+      });
     });
+  }
+
+  it('EIP-170 limit on immutable args', async function () {
+    // EIP-170 limits the contract code size to 0x60000
```
Amxx, line 161:
```suggestion
    // EIP-170 limits the contract code size to 0x6000
```

ernestognw (dismissed):

ernestognw (dismissed):

ernestognw (commented):
Before merging we've been discussing how the code size limit error for the clone contract should be.

Our guidelines indicate to follow the ERC6093 rationale for custom errors, I'm personally fine with `EIP170ContractCodeSizeLimit` but it's missing the "error prefix". I'd like to propose the following:

- `EIP170ExceededContractSizeLimit` (same rationale as in other errors like `ERC20ExceededCap` or `ERC4626ExceededMaxWithdraw`)
- `CloneArgumentsTooLong` (similar rationale to `StringTooLong`)

Perhaps we may go even more generic and just have a generic `InitcodeSizeExceeded` in `Errors.sol`.

wdyt @Amxx ?

Amxx:
My preference goes for `CloneArgumentsTooLong` because it is clear what the issue is.

If you get `EIP170ExceededContractSizeLimit`, you must understand that there is a contract size limit, then understand that the immutable args are part of the code, then understand that the immutable args are so long that they cause the size of the proxy to exceed the limit ... 

`CloneArgumentsTooLong` is clear: the argument are too long, you don't need to know why, just that its not supported.

ernestognw (approved):
Sounds good.

I think `CloneArgumentsTooLong` is good enough and I don't want to block this anymore. Ideally, I'd like the naming to meet our guidelines for custom errors and be `CloneTooLongArguments` but I understand you dislike it because it's not as clearly readable as the current naming. I agree, though.

I'm approving since I don't have a major issue and error backwards compatibility is not guaranteed anyway.

PR #4991, opened 2024-04-04
ERC-4337 & ERC-7579 compliant modular smart accounts

Amxx:
Fixes LIB-1237
Fixes LIB-1236
Fixes LIB-1231
Fixes LIB-1230
Fixes LIB-1229
Fixes LIB-1228
Fixes LIB-1227

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changeset entry (run `npx changeset add`)

PR #5119, opened 2024-07-22
ERC-4337 & ERC-7579 compliant modular smart accounts with ERC7579 signer/validator support

Amxx:
Variant of #4991 with full 7579 based userop verification

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ernestognw (commented):
I've been reviewing and it's looking good if we only consider the ERC7579 view of how to implement accounts.

I'd like to challenge the idea that ERC7579 is indeed the minimal version of an account; IMO, the `validators` are not a good fit for identities since they're not something you change frequently as you may do with a module. That would require a lot of operations during deployment (e.g. deploying multiple identities and enabling multiple modules).

From my point of view, identities are core to the functionality and we already have a standard for signature interoperability (ERC1271) so I'd like the identities approach to be more focused on making ERC1271 work out of the box rather than through a validator module.

I opened [this PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5182/files) to demonstrate how it simpler if we allow a generic identity to be plugged into an account (see AccountIdentity) or multiple (AccountMultiIdentity). I feel that approach is way leaner, but we can still write ERC7579 modules for more complex requirements (e.g. Governor as an executor).

ernestognw:
Closing in favor of community-contracts

PR #5123, opened 2024-07-24
Update ReentrancyGuard.sol (#1)

sueun-dev:
I conducted tests comparing the previous code with the optimized version and was able to reduce gas costs by 0.024%. Therefore, I am submitting a pull request.

Gas used for increment (Original) 1: 45914
Gas used for increment (Original) 2: 28814
Gas used for increment (Original) 3: 28814
...

Gas used for increment (After Change Code) 1: 45903 
Gas used for increment (After Change Code) 2: 28803
Gas used for increment (After Change Code) 3: 28803
...

Therefore, This code is more efficient.

Fixes #???? 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/ReentrancyGuard.sol
+++ b/contracts/utils/ReentrancyGuard.sol
@@ -62,19 +62,25 @@ abstract contract ReentrancyGuard {
     }
 
     function _nonReentrantBefore() private {
-        // On the first call to nonReentrant, _status will be NOT_ENTERED
-        if (_status == ENTERED) {
-            revert ReentrancyGuardReentrantCall();
+        assembly {
+            // Load the value of _status
+            let status := sload(_status.slot)
+            // If status is ENTERED, throw an error
+            if eq(status, 2) {
+                mstore(0x00, 0x4e487b71) // keccak256("Error(string)")
+                mstore(0x20, 0x20) // Data length (32 bytes)
+                revert(0x00, 0x24) // Revert with (0x00 to 0x24)
+            }
+            // Set _status to ENTERED
+            sstore(_status.slot, 2)
         }
-
-        // Any calls to nonReentrant after this point will fail
-        _status = ENTERED;
     }
 
     function _nonReentrantAfter() private {
-        // By storing the original value once again, a refund is triggered (see
-        // https://eips.ethereum.org/EIPS/eip-2200)
-        _status = NOT_ENTERED;
+        assembly {
+            // Set _status to NOT_ENTERED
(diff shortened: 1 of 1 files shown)


contracts/utils/ReentrancyGuard.sol
```diff
@@ -62,26 +62,37 @@ abstract contract ReentrancyGuard {
...
-        if (_status == ENTERED) {
-            revert ReentrancyGuardReentrantCall();
+        assembly {
+            // Load the value of _status
+            let status := sload(_status.slot)
+            // If status is ENTERED, throw an error
+            if eq(status, 2) {
+                mstore(0x00, 0x4e487b71) // keccak256("Error(string)")
+                mstore(0x20, 0x20) // Data length (32 bytes)
+                revert(0x00, 0x24) // Revert with (0x00 to 0x24)
```
cairoeth, line 72: should be reverting `ReentrancyGuardReentrantCall`

contracts/utils/ReentrancyGuard.sol
```diff
@@ -62,26 +62,37 @@ abstract contract ReentrancyGuard {
...
 
     function _nonReentrantBefore() private {
-        // On the first call to nonReentrant, _status will be NOT_ENTERED
-        if (_status == ENTERED) {
-            revert ReentrancyGuardReentrantCall();
+        assembly {
+            // Load the value of _status
+            let status := sload(_status.slot)
+            // If status is ENTERED, throw an error
+            if eq(status, 2) {
```
cairoeth, line 69: can be combined into:
```suggestion
            // If status is ENTERED, throw an error
            if eq(sload(_status.slot), 2) {
```

contracts/utils/ReentrancyGuard.sol
```diff
@@ -62,26 +62,37 @@ abstract contract ReentrancyGuard {
...
+                mstore(0x00, 0x4e487b71) // keccak256("Error(string)")
+                mstore(0x20, 0x20) // Data length (32 bytes)
+                revert(0x00, 0x24) // Revert with (0x00 to 0x24)
+            }
+            // Set _status to ENTERED
+            sstore(_status.slot, 2)
```
cairoeth, line 75: in general, we should be using `ENTERED` and `NOT_ENTERED` constants instead of the defined value.

cairoeth (changes requested):
after addressing these comments, i imagine gas savings will be lower so not sure if they are worth it in exchange for less code clarity.

PR #5126, opened 2024-07-24
[Optimization] Handling zero length batch transfer in ERC-1155

### LINKED ISSUE #5125: Potential Optimization on ERC-1155 `update` event
**🧐 Motivation**
The current [`_update` function](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/932fddf69a699a9a80fd2396fd1a2ab91cdda123/contracts/token/ERC1155/ERC1155.sol#L147-L148) in ERC-1155 emits two types of events:
- `TransferSingle` when `ids.length == 1`
- `TransferBatch` otherwise

However, it does not handle the case when `ids.length == 0`, which can occur with `safeBatchTransferFrom`. Currently, it emits `TransferBatch` even when there are no transfers, leading to inconsistency.

**📝 Details**
We can introduce a separate event, such as `TransferZero`, to accurately reflect the transaction behavior when empty batch is transferred. Also, may save gas slightly(based on my test PR)

I propose a PR #5126  here. Please check, and thank you!

QiLOL:
Fixes #5125  
For empty batch transfer in ERC1155

Before:
```
[PASS] testTransferZero() (gas: 14037)
Traces:
  [14037] MERC1155Test::testTransferZero()
    ├─ [8363] MERC1155::safeBatchTransferFrom(MERC1155Test: [0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496], 0x0000000000000000000000000000000000000042, [], [], 0x)
    │   ├─ emit TransferBatch(operator: MERC1155Test: [0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496], from: MERC1155Test: [0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496], to: 0x0000000000000000000000000000000000000042, ids: [], values: [])
    │   └─ ← [Stop]
    └─ ← [Stop]
```

After:
```
[PASS] testTransferZero() (gas: 12681)
Traces:
  [12681] MERC1155Test::testTransferZero()
    ├─ [7007] MERC1155::safeBatchTransferFrom(MERC1155Test: [0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496], 0x0000000000000000000000000000000000000042, [], [], 0x)
    │   ├─ emit TransferZero(operator: MERC1155Test: [0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496], from: MERC1155Test: [0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496], to: 0x0000000000000000000000000000000000000042)
    │   └─ ← [Stop]
    └─ ← [Stop]
```

Test contracts:
```
// Compatible with OpenZeppelin Contracts ^5.0.0
pragma solidity ^0.8.20;

import "@openzeppelin/contracts/token/ERC1155/ERC1155.sol";
import "@openzeppelin/contracts/access/Ownable.sol";

contract MERC1155 is ERC1155, Ownable {
    constructor(address initialOwner) ERC1155("") Ownable(initialOwner) {}

    function mint(address account, uint256 id, uint256 amount, bytes memory data)
        public
        onlyOwner
    {
        _mint(account, id, amount, data);
    }

    function mintBatch(address to, uint256[] memory ids, uint256[] memory amounts, bytes memory data)
        public
        onlyOwner
    {
        _mintBatch(to, ids, amounts, data);
    }
}
```

```
pragma solidity ^0.8.13;

import {Test, console} from "forge-std/Test.sol";
import {MERC1155} from "../src/merc1155.sol";

contract MERC1155Test is Test {

    MERC1155 mtoken;

    function setUp() public {
        mtoken = new MERC1155(address(this));
    }

    function testTransferZero() public {
        uint256[] memory ids;
        uint256[] memory values;
        mtoken.safeBatchTransferFrom(address(this), address(0x42), ids, values, "");
    }

}
```

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/token/ERC1155/IERC1155.sol
+++ b/contracts/token/ERC1155/IERC1155.sol
@@ -11,6 +11,11 @@ import {IERC165} from "../../utils/introspection/IERC165.sol";
  */
 interface IERC1155 is IERC165 {
     /**
+     * @dev Emitted when zero amount of token transferred from `from` to `to` by `operator` with safeBatchTransferFrom.
+     */
+    event TransferZero(address indexed operator, address indexed from, address indexed to);
+    
+    /**
      * @dev Emitted when `value` amount of tokens of type `id` are transferred from `from` to `to` by `operator`.
      */
     event TransferSingle(address indexed operator, address indexed from, address indexed to, uint256 id, uint256 value);
--- a/contracts/token/ERC1155/ERC1155.sol
+++ b/contracts/token/ERC1155/ERC1155.sol
@@ -172,7 +172,10 @@ abstract contract ERC1155 is Context, ERC165, IERC1155, IERC1155MetadataURI, IER
             }
         }
 
-        if (ids.length == 1) {
+        if (ids.length == 0) {
+            emit TransferZero(operator, from, to);
+        }
+        else if (ids.length == 1) {
             uint256 id = ids.unsafeMemoryAccess(0);
             uint256 value = values.unsafeMemoryAccess(0);
             emit TransferSingle(operator, from, to, id, value);


contracts/token/ERC1155/IERC1155.sol
```diff
@@ -10,6 +10,11 @@ import {IERC165} from "../../utils/introspection/IERC165.sol";
  * https://eips.ethereum.org/EIPS/eip-1155[ERC].
  */
 interface IERC1155 is IERC165 {
+    /**
+     * @dev Emitted when zero amount of token transferred from `from` to `to` by `operator` with safeBatchTransferFrom.
+     */
+    event TransferZero(address indexed operator, address indexed from, address indexed to);
+
```
Amxx, line 17: This is not part of the ERC-1155 standard. Even if we want it (I don't think we do) it should no be here.

Amxx:
Hello @QiLOL 

The additional check will increase the gas cost of all "non empty" batch transfers. I don't think we want that. Additionally, the new event you introduce is non standard. Using that is an opinionated choice we should not be doing, as it will be poorly supported by indexers.

I'd like to understand what the issue is with the current implementation. Emitting a BatchTransfer event with empty arrays for ids and values doesn't sound like something that should cause any indexing issues.

Amxx:
Note that both solmate and solady do emit BatchTransfer events with empty arrays

QiLOL:
> Note that both solmate and solady do emit BatchTransfer events with empty arrays

Thanks for the explanation. I'm just confused about why we use `TransferSingle` exclusively instead of a list with a single element. However, I understand that this might be more related to the EIP standard than the implementation. Do you have any idea?

Amxx:
> I understand that this might be more related to the EIP standard than the implementation

Yes. ERC are written with standard events, and we use them.

In that case, it may be because simple event are the most commonly used, and creating an event with a simple value (instead of an array) save gas. So they optimise the most commonly used case.

The case were the array are both empty should be extremelly rare, and so we don't usually want to optimise it, particularly if it increases the cost of other (more frequent) cases.

QiLOL:
@Amxx Thank you, I will close both PR and the issue, I really appreciate your explanation.

PR #5129, opened 2024-07-29
Add note about opinionated design to ERC4626Fees

ernestognw:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/mocks/docs/ERC4626Fees.sol
+++ b/contracts/mocks/docs/ERC4626Fees.sol
@@ -8,6 +8,10 @@ import {SafeERC20} from "../../token/ERC20/utils/SafeERC20.sol";
 import {Math} from "../../utils/math/Math.sol";
 
 /// @dev ERC-4626 vault with entry/exit fees expressed in https://en.wikipedia.org/wiki/Basis_point[basis point (bp)].
+///
+/// NOTE: The contract charges fees in terms of assets, not shares. This means that the fees are calculated based on the
+/// amount of assets that are being deposited or withdrawn, and not based on the amount of shares that are being minted or
+/// redeemed. This is an opinionated design decision that should be taken into account when integrating this contract.
 abstract contract ERC4626Fees is ERC4626 {
     using Math for uint256;
 
--- a/contracts/mocks/docs/ERC4626Fees.sol
+++ b/contracts/mocks/docs/ERC4626Fees.sol
@@ -12,6 +12,8 @@ import {Math} from "../../utils/math/Math.sol";
 /// NOTE: The contract charges fees in terms of assets, not shares. This means that the fees are calculated based on the
 /// amount of assets that are being deposited or withdrawn, and not based on the amount of shares that are being minted or
 /// redeemed. This is an opinionated design decision that should be taken into account when integrating this contract.
+///
+/// WARNING: This contract has not been audited and shouldn't be considered production ready. Consider using it with caution.
 abstract contract ERC4626Fees is ERC4626 {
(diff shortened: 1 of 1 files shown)


contracts/mocks/docs/ERC4626Fees.sol
```diff
@@ -8,6 +8,10 @@ import {SafeERC20} from "../../token/ERC20/utils/SafeERC20.sol";
 import {Math} from "../../utils/math/Math.sol";
 
 /// @dev ERC-4626 vault with entry/exit fees expressed in https://en.wikipedia.org/wiki/Basis_point[basis point (bp)].
+///
+/// NOTE: The contract charges fees in terms of assets, not shares. This means that the fees are calculated based on the
+/// amount of assets that are being deposited or withdrawn, and not based on the amount of shares that are being minted or
+/// redeemed. This is an opinionated design decision that should be taken into account when integrating this contract.
```
cairoeth, line 14: > taken into account when integrating this contract

we should instead say this contract should not be integrated with and that is not ready for production usage nor audited.

cairoeth (commented):
this is going to get propagated to https://docs.openzeppelin.com/contracts/4.x/erc4626#fees as well, right? maybe we should have a ASCIIdoc note outside and before the contract section to make it more noticeable

ernestognw:
> this is going to get propagated to https://docs.openzeppelin.com/contracts/4.x/erc4626#fees as well, right?

We need to cherry-pick the commit to the `release-v5.0` branch after merging so that it makes it to the docs.

> maybe we should have a ASCIIdoc note outside and before the contract section to make it more noticeable

I felt it a bit redundant given the docs in `docs/modules/ROOT/pages/erc4626.adoc` are already concise and explain the mechanism quite well imo. I am mostly worried for those who use VSCode search in Foundry since they're the most likely to pick up this contract without reading the docs

contracts/mocks/docs/ERC4626Fees.sol
ernestognw: I would say it's ready for production usage under certain conditions that are up to the developer. Instead, I'd focus on a list of considerations to take before using it, I feel the current note is already precise enough, wdyt?

cairoeth: imo: not audited = not production ready, unless the dev wants to take the risk. also we have the considerations that `maxWithdraw` and `maxRedeem` are not considering fees, so the `ERC4626Exceeded` checks don't work correctly.

Amxx: I agree with Monchi, this is in mocks, and its not released in packages. It should be used "at your own risks", similar to the community contracts.

ernestognw: Agree with both, I just think it's more precise to prompt people to _use under their own responsibility since it hasn't been audited_ than warning about _not being production ready_

> also we have the considerations that maxWithdraw and maxRedeem are not taking into account fees

Also true. We can clarify that `maxWithdraw` and `maxRedeem` are capped to the user's balance; I don't see a way where it could fail other than if they're customized.

cairoeth (approved):
lgtm -- i can open another PR to fix `maxWithdraw` and `maxRedeem`

ernestognw:
> lgtm -- i can open another PR to fix `maxWithdraw` and `maxRedeem`

Sounds good, that's the only concern missing but it's fine if we tackle it in a subsequent PR imo

PR #5130, opened 2024-07-30
Refactor ERC4626-maxWithdraw to reflect other functions overrides

Amxx:
@cairoeth WDYT?

#### PR Checklist

### DIFF
--- a/contracts/token/ERC20/extensions/ERC4626.sol
+++ b/contracts/token/ERC20/extensions/ERC4626.sol
@@ -44,6 +44,18 @@ import {Math} from "../../../utils/math/Math.sol";
  *
  * To learn more, check out our xref:ROOT:erc4626.adoc[ERC-4626 guide].
  * ====
+ *
+ * Note: When overriding this contract, some elements have to be considered
+ *
+ * * When overriding the behavior of the deposit or withdraw mechanism, one should override the internal functions.
+ * Overriding {_deposit} automatically affects both {deposit} and {mint}. Similarly, overriding {_withdraw}
+ * automatically affects both {withdraw} and {redeem}. Having inconsistent behavior between the {deposit} and {mint}
+ * or between {withdraw} and {redeem} can lead to bugs. Overall it is not recommended to override the public facing
+ * functions.
+ *
+ * * {maxWithdraw} depends on {maxRedeem}. Therefore, overriding {maxRedeem} only is enough. On the other hand,
+ * overriding {maxWithdraw} only would have no effect on {maxRedeem}, and could create an inconsistency between the two
+ * functions.
  */
 abstract contract ERC4626 is ERC20, IERC4626 {
     using Math for uint256;
--- a/contracts/token/ERC20/extensions/ERC4626.sol
+++ b/contracts/token/ERC20/extensions/ERC4626.sol
@@ -49,7 +49,7 @@ import {Math} from "../../../utils/math/Math.sol";
  * ====
  * When overriding this contract, some elements must to be considered:
  *
- * * When overriding the behavior of the deposit or withdraw mechanism, it is recommended to override the internal
+ * * When overriding the behavior of the deposit or withdraw mechanisms, it is recommended to override the internal
  * functions. Overriding {_deposit} automatically affects both {deposit} and {mint}. Similarly, overriding {_withdraw}
  * automatically affects both {withdraw} and {redeem}. Overall it is not recommended to override the public facing
  * functions since that could lead to inconsistent behaviors between the {deposit} and {mint} or between {withdraw} and
--- a/contracts/token/ERC20/extensions/ERC4626.sol
+++ b/contracts/token/ERC20/extensions/ERC4626.sol
@@ -45,17 +45,25 @@ import {Math} from "../../../utils/math/Math.sol";
  * To learn more, check out our xref:ROOT:erc4626.adoc[ERC-4626 guide].
  * ====
  *
- * Note: When overriding this contract, some elements have to be considered
+ * [NOTE]
+ * ====
+ * When overriding this contract, some elements must to be considered:
  *
- * * When overriding the behavior of the deposit or withdraw mechanisms, it is recommended to override the internal
+ * * When overriding the behavior of the deposit or withdraw mechanism, it is recommended to override the internal
  * functions. Overriding {_deposit} automatically affects both {deposit} and {mint}. Similarly, overriding {_withdraw}
  * automatically affects both {withdraw} and {redeem}. Overall it is not recommended to override the public facing
  * functions since that could lead to inconsistent behaviors between the {deposit} and {mint} or between {withdraw} and
  * {redeem}, which is documented to have lead to loss of funds.
  *
+ * * Overrides to the deposit or withdraw mechanism must be reflected in the preview functions as well.
+ *
  * * {maxWithdraw} depends on {maxRedeem}. Therefore, overriding {maxRedeem} only is enough. On the other hand,
  * overriding {maxWithdraw} only would have no effect on {maxRedeem}, and could create an inconsistency between the two
  * functions.
+ *
+ * * If {previewRedeem} is overriden to revert, {maxWithdraw} must be overridden as necessary to ensure it
+ * always return successfully.
+ * ====
  */
 abstract contract ERC4626 is ERC20, IERC4626 {
     using Math for uint256;
--- a/contracts/token/ERC20/extensions/ERC4626.sol
+++ b/contracts/token/ERC20/extensions/ERC4626.sol
@@ -47,11 +47,11 @@ import {Math} from "../../../utils/math/Math.sol";
  *
  * Note: When overriding this contract, some elements have to be considered
  *
- * * When overriding the behavior of the deposit or withdraw mechanism, one should override the internal functions.
- * Overriding {_deposit} automatically affects both {deposit} and {mint}. Similarly, overriding {_withdraw}
- * automatically affects both {withdraw} and {redeem}. Having inconsistent behavior between the {deposit} and {mint}
- * or between {withdraw} and {redeem} can lead to bugs. Overall it is not recommended to override the public facing
- * functions.
+ * * When overriding the behavior of the deposit or withdraw mechanisms, it is recommended to override the internal
+ * functions. Overriding {_deposit} automatically affects both {deposit} and {mint}. Similarly, overriding {_withdraw}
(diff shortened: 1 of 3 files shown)


contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -97,77 +97,77 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
+    /// @inheritdoc IERC4626
     function maxMint(address) public view virtual returns (uint256) {
         return type(uint256).max;
     }
 
-    /** @dev See {IERC4626-maxWithdraw}. */
+    /// @inheritdoc IERC4626
     function maxWithdraw(address owner) public view virtual returns (uint256) {
-        return _convertToAssets(balanceOf(owner), Math.Rounding.Floor);
+        return previewRedeem(maxRedeem(owner));
```
Amxx, line 142: This is the change I propose.

In the case of ERC4626Fees, the fees inclusion in `previewRedeem` woudl reflect here.
Additionally, if `maxRedeem` is ever overriden, it would reflect here (but the inverse is not true)

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -97,77 +97,77 @@ abstract contract ERC4626 is ERC20, IERC4626 {
     }
 
     /**
+     * @inheritdoc IERC20Metadata
+     *
      * @dev Decimals are computed by adding the decimal offset on top of the underlying asset's decimals. This
      * "original" value is cached during construction of the vault contract. If this read operation fails (e.g., the
      * asset has not been created yet), a default of 18 is used to represent the underlying asset's decimals.
```
ernestognw, line 104: This doesn't work in the docs. It just takes the next `@dev` tag. I would keep the old NatSpec here instead.

```suggestion
     * @inheritdoc IERC20Metadata
     *
     * @dev Decimals are computed by adding the decimal offset on top of the underlying asset's decimals. This
     * "original" value is cached during construction of the vault contract. If this read operation fails (e.g., the
     * asset has not been created yet), a default of 18 is used to represent the underlying asset's decimals.
     *
     * See {IERC20Metadata-decimals}.
```

ernestognw (approved):

ernestognw:
On second thought (thanks @cairoeth), I don't think the changelog entry is enough. It seems like there are [multiple cases](https://sourcegraph.com/search?q=context:global+repo:tintinweb/smart-contract-sanctuary+file:/mainnet/+/%5CbpreviewRedeem%5C%28.*override/&patternType=standard&sm=0) of ERC4626 vaults that override `previewRedeem`, which would affect the logic in `maxWithdraw`.

To me, it's not a huge deal because I can't see how it may affect the vault critically. If we continue with this change, we should discuss whether a changelog entry under the "Breaking Changes" section is good enough, or whether the change will cause an issue in some projects.

cairoeth:
Following up on ^, the initial problem is that `maxRedeem` and `maxWithdraw` do not consider fees in the `ERC4626Fees` implementation -- I was thinking that a solution is just overriding `maxRedeem` in `ERC4626Fees` with something like this PR, instead of making changes in `ERC4626`. I mainly saw it as a problem in the `ERC4626Fees` implementation rather that in `ERC4626`. Based on the examples @ernestognw provided, I think this PR would make it easier for developers to mistakenly modify `maxWithdraw` by overriding `previewRedeem`, whereas before they weren't really connected as `maxWithdraw` just used `_convertToAssets`.

ernestognw:
Opened #5130 as an alternative. It would make sense to include the `maxWithdraw` override in `ERC4626` given that anyone adding fees would need to override `maxWithdraw` as well, and it's perhaps not obvious.

I'm not sure that's worth documenting. The "raw" ERC4626 shouldn't have any concerns about fees, and overriding functions are already assumed under the developer's risk.

pcaversaccio:
So @cairoeth and myself quickly discussed this change, and if you read the `maxWithdraw` section of the [EIP-4626](https://eips.ethereum.org/EIPS/eip-4626#maxwithdraw) carefully, it states:

image

Now given the changeset for this PR, you could override `previewRedeem` with a function that reverts and thus it disables `maxWithdraw` and is hence not anymore EIP-4626 compliant. Note, that `previewRedeem` is allowed to revert:
> MAY revert due to other conditions that would also cause `redeem` to revert.

Under these circumstances, I would not recommend implementing this change.

Amxx:
> Now given the changeset for this PR, you could override `previewRedeem` with a function that reverts and thus it disables `maxWithdraw` and is hence not anymore EIP-4626 compliant.

My deep feeling is that `withdraw` and `redeem` are two ways of expresdsing the same operation. Its like saying "change 1$ into €" vas saying "change 0.91€ worth of $ into €". You give the number differently, but its essentially the same thing. I'd be curious why anyone would want to disable `withdraw` while keeping `redeem` functionnal. IMO it makes no sens.

This PR would increass the coupling between the two.

To me the consequence of this coupling are the following:
- Someone that want to change the behvarior of both functions only has one function to override, and both behavior would be consistent with one another. We've example of code where one of the override was missed, and you end up with undesired inconsistent behavior.
- Someone that wants to implement an inconsistency between the two function can still do it, but it may require 2 overrides instead of one.

It is my personal opinion that the first point (wanting consistency) is way more common (and desirable), and this is the one we should favor in our designs. We also identified instances of buggy vaults that would not have been buggy if this PR had been implemented.
I also think that the second point (implement inconsistent behavior between redeem and withdraw) is going to be an edge case that should be well documented. As such, I think asking the dev to override both side so that the inconsistency appears clearly is a good thing.

That is why I continue to support this change. 

If you can find any example of real production code (not POCs) that this would break, I'd love to study them!

Amxx:
This is likelly going to be being delayed, from 5.2 to 5.3.

So far:
1. We do have examples of production contracts where this change would have prevented issues/inconstency.
2. We don't have concrete example of contracts where this change would have caused issues.

We are delaying this because we are waiting for example of the second. At some point we need to stop delaying that more. If `2.` doesn't exist, we should merge this PR to benefit from `1.`.

@frangio @ernestognw @arr00

contracts/token/ERC20/extensions/ERC4626.sol
james-toussaint: It would be useful to add doc stating that only `previewRedeem(..)` needs to be override, something close to [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5130/files#r1696866993) or close to a previous comment from you "_Someone that want to change the behvarior of both functions only has one function to override, and both behavior would be consistent with one another_" into the top documentation of the contract (before or after _[CAUTION]_).

james-toussaint (dismissed):

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -107,67 +107,67 @@ abstract contract ERC4626 is ERC20, IERC4626 {
...
     }
 
-    /** @dev See {IERC4626-previewWithdraw}. */
+    /// @inheritdoc IERC4626
     function previewWithdraw(uint256 assets) public view virtual returns (uint256) {
         return _convertToShares(assets, Math.Rounding.Ceil);
     }
 
-    /** @dev See {IERC4626-previewRedeem}. */
+    /// @inheritdoc IERC4626
```
arr00, line 185: We should add a note here that if a developer chooses to make `previewRedeem` revert under certain scenarios, they should ensure `maxWithdraw` won't revert.

arr00, line 185:
```suggestion
    /**
     * @dev See {IERC4626-previewRedeem}.
     * NOTE: If this function is overridden to introduce reverting logic, ensure that `maxWithdraw` is overridden as necessary to return successfully in all situations.
     */
```

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: does that look good ? https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5130/commits/fa67c3faf99d41325f34bba5e3a016e921624e83

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: This messes up the documentation. Instead, I'm proposing to add an "override guide" at the contract level: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5130/commits/fa67c3faf99d41325f34bba5e3a016e921624e83

Fell free to propose adding to it.

Amxx:
@coderabbitai pause

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -44,6 +44,26 @@ import {Math} from "../../../utils/math/Math.sol";
  *
  * To learn more, check out our xref:ROOT:erc4626.adoc[ERC-4626 guide].
  * ====
+ *
+ * [NOTE]
+ * ====
+ * When overriding this contract, some elements must to be considered:
+ *
+ * * When overriding the behavior of the deposit or withdraw mechanism, it is recommended to override the internal
```
arr00:
```suggestion
 * * When overriding the behavior of the deposit or withdraw mechanisms, it is recommended to override the internal
```

arr00 (approved):

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -44,6 +44,26 @@ import {Math} from "../../../utils/math/Math.sol";
...
+ * functions since that could lead to inconsistent behaviors between the {deposit} and {mint} or between {withdraw} and
+ * {redeem}, which is documented to have lead to loss of funds.
+ *
+ * * Overrides to the deposit or withdraw mechanism must be reflected in the preview functions as well.
+ *
+ * * {maxWithdraw} depends on {maxRedeem}. Therefore, overriding {maxRedeem} only is enough. On the other hand,
+ * overriding {maxWithdraw} only would have no effect on {maxRedeem}, and could create an inconsistency between the two
+ * functions.
+ *
+ * * If {previewRedeem} is overriden to revert, {maxWithdraw} must be overridden as necessary to ensure it
```
arr00:
```suggestion
 * * If {previewRedeem} is overridden to revert, {maxWithdraw} must be overridden as necessary to ensure it
```

arr00 (approved):

contracts/token/ERC20/extensions/ERC4626.sol
james-toussaint: Clear & useful note yes, thank you!

PR #5135, opened 2024-07-30
Refactor ERC4626Fees example to account for fees in `maxWithdraw`

ernestognw:
Alternative to #5130

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/mocks/docs/ERC4626Fees.sol
+++ b/contracts/mocks/docs/ERC4626Fees.sol
@@ -8,6 +8,10 @@ import {SafeERC20} from "../../token/ERC20/utils/SafeERC20.sol";
 import {Math} from "../../utils/math/Math.sol";
 
 /// @dev ERC-4626 vault with entry/exit fees expressed in https://en.wikipedia.org/wiki/Basis_point[basis point (bp)].
+///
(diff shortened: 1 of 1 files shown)


contracts/mocks/docs/ERC4626Fees.sol
```diff
@@ -21,6 +21,11 @@ abstract contract ERC4626Fees is ERC4626 {
 
     // === Overrides ===
 
+    /// @dev Maximum amount of the underlying asset that can be withdrawn from the owner's balance.
```
cairoeth, line 24:
```suggestion
    /// @dev Maximum amount of the underlying asset that can be withdrawn from the owner's balance after subtracting fees.
```

cairoeth (commented):
looks good -- as we have the testing suite for `ERC4626Fees` [already set up](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/test/token/ERC20/extensions/ERC4626.test.js#L616), maybe we want to add a unit test for `maxWithdraw`? alternatively, we can add the tests on the community repo

Amxx:
IMO, we should do the change in ERC4626, not in ERC4626Fees.

PR #5136, opened 2024-07-31
Fix natspec comment missing *

Amxx:


ericglau:
Would it make sense to add a custom Solidity linting rule for this?

Amxx:
> Would it make sense to add a custom Solidity linting rule for this?

yes!

If there is a `@dev` or similar tag in a comment that is not `///` or `/**`, then raise an error.

ernestognw (approved):

ericglau:
Looks like Solhint doesn't support parsing comments since it's not part of the AST, so the linting suggestion above may be a future consideration.

pedrobergamini:
lol I was looking for a Heap lib and just found this recently merged. Congrats and thanks for your work!

PR #5140, opened 2024-08-02
Get leaves from memory in processMultiProofCalldata

Amxx:
Following the audit discussion.

Note that is reverting a (undocumented breaking) change that we introduced in #4887
5.0 does use memory for leaves in the calldata variants:
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/utils/cryptography/MerkleProof.sol#L88-L95
- https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v5.0/contracts/utils/cryptography/MerkleProof.sol#L88-L95

Edit: I also took this opportunity to fix some missing `{suffix}` in the code generation.

### DIFF
--- a/contracts/utils/cryptography/MerkleProof.sol
+++ b/contracts/utils/cryptography/MerkleProof.sol
@@ -333,7 +333,7 @@ library MerkleProof {
         bytes32[] calldata proof,
         bool[] calldata proofFlags,
         bytes32 root,
-        bytes32[] calldata leaves
+        bytes32[] memory leaves
     ) internal pure returns (bool) {
         return processMultiProofCalldata(proof, proofFlags, leaves) == root;
     }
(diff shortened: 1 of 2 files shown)


ernestognw (commented):
Looks good, but I'm wondering whether it's still useful to have the following two versions:

```solidity
function multiProofVerifyCalldata(
    bytes32[] calldata proof,
    bool[] calldata proofFlags,
    bytes32 root,
    bytes32[] calldata leaves,
    function(bytes32, bytes32) view returns (bytes32) hasher
    return processMultiProofCalldata(proof, proofFlags, leaves, hasher) == root;
}

function multiProofVerifyCalldata(
    bytes32[] calldata proof,
    bool[] calldata proofFlags,
    bytes32 root,
    bytes32[] calldata leaves
    return processMultiProofCalldata(proof, proofFlags, leaves) == root;
}
```

They seem rather unnecessary given that we're copying leaves to memory when using the underlying `processMultiProof`

Amxx:
Right, I forgot about the leaves location in `multiProofVerifyCalldata`. Fixed.

ernestognw (approved):
Thanks @Amxx

PR #5141, opened 2024-08-04
🔨 scripts: ignore prepare errors

cruzdanilo:
# issue
since #4962, the project has a `prepare` script that can only run inside a git repository. this is an issue when using the package as a git dependency in an npm-based project. when installing the package, the `prepare` script will run, but at that point, there will be no `.git` folder and the script will fail with the following error:
```log
> openzeppelin-solidity@5.0.2 prepare
> git config --local core.hooksPath .githooks
fatal: --local can only be used inside a git repository
```
i can confirm that this issue prevents the package from being used as a git dependency in npm and pnpm, and will probably cause similar issues with other npm-based package managers.

considering that some unreleased features are already being used in production by other libraries, using the package as a git dependency can be a requirement for some projects. example of `Base64.encodeURL` being used by base's `webauthn-sol` library:

https://github.com/base-org/webauthn-sol/blob/main/src/WebAuthn.sol#L123

# fix
i suggest the simplest and cleanest solution, which is just ignoring any error in the script with `|| true`. to be more strict, there's another more verbose and maybe less compatible solution:
```bash
# suggested fix
git config --local core.hooksPath .githooks || true

# another option
if git status &>/dev/null; then git config --local core.hooksPath .githooks; fi
```


ernestognw (approved):
Thanks for reporting @cruzdanilo, we discussed and decided to go for the `if git status &>/dev/null; then git config core.hooksPath .githooks; fi` since it may be the best to get it working on every OS

PR #5139, opened 2024-08-02
Release v5.1 updates

Amxx:
This PR contains commits cherry picked from master that should be applied to release-v5.1.

---

Any PR done for https://github.com/OpenZeppelin/openzeppelin-contracts/milestone/21 should first be merge into master then cherry picked here. Once we are happy with them, we should fast forward the release branch to this

PR #5142, opened 2024-08-05
Chore/add empty set note merkle proof

ernestognw:
Makes it explicit that users verifying multiproofs should consider that the empty set is valid.
Most implementations do check the validity of the leaves or build them within the contract, which makes it impossible to validate a proof incorrectly.

This is consistent with our no-ops policy

#### PR Checklist

- [ ] Tests
- [x] Documentation


cairoeth (approved):
lgtm, would be cherry-picked for 5.1 update?

ernestognw:
> lgtm, would be cherry-picked for 5.1 update?

Yes, I'll pick it up to #5139

PR #5144, opened 2024-08-08
Update MerkleProof note clarifying empty set definition

ernestognw:
In the last commit ([bcd4beb](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/bcd4beb5e7fd8bd8edf160fbffb5d5b03804efdb)), the definition of the empty set was incorrect as it needs to include the root within the proof. This PR updates it to the correct definition.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/cryptography/MerkleProof.sol
+++ b/contracts/utils/cryptography/MerkleProof.sol
@@ -169,9 +169,8 @@ library MerkleProof {
      *
      * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
      *
-     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
-     * and therefore a valid multiproof (i.e. it returns `true`). Consider disallowing this case if you're not
-     * validating the leaves elsewhere.
+     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return true.
+     * The `leaves` should be validated independently. See {processMultiProof}.
      */
     function multiProofVerify(
         bytes32[] memory proof,
@@ -193,6 +192,10 @@ library MerkleProof {
      * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
      * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
      * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+     * 
+     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
+     * and therefore a valid multiproof (i.e. it returns `proof[0]`). Consider disallowing this case if you're not
+     * validating the leaves elsewhere.
      */
     function processMultiProof(
         bytes32[] memory proof,
@@ -252,9 +255,8 @@ library MerkleProof {
      *
      * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
      *
-     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
-     * and therefore a valid multiproof (i.e. it returns `true`). Consider disallowing this case if you're not
-     * validating the leaves elsewhere.
+     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return true.
+     * The `leaves` should be validated independently. See {processMultiProof}.
      */
     function multiProofVerify(
         bytes32[] memory proof,
@@ -277,6 +279,10 @@ library MerkleProof {
      * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
      * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
      * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+     * 
+     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
+     * and therefore a valid multiproof (i.e. it returns `proof[0]`). Consider disallowing this case if you're not
+     * validating the leaves elsewhere.
      */
     function processMultiProof(
         bytes32[] memory proof,
@@ -337,9 +343,8 @@ library MerkleProof {
      *
      * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
      *
-     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
-     * and therefore a valid multiproof (i.e. it returns `true`). Consider disallowing this case if you're not
-     * validating the leaves elsewhere.
+     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return true.
+     * The `leaves` should be validated independently. See {processMultiProofCalldata}.
      */
     function multiProofVerifyCalldata(
         bytes32[] calldata proof,
@@ -361,6 +366,10 @@ library MerkleProof {
      * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
      * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
      * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+     * 
+     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
+     * and therefore a valid multiproof (i.e. it returns `proof[0]`). Consider disallowing this case if you're not
+     * validating the leaves elsewhere.
      */
     function processMultiProofCalldata(
         bytes32[] calldata proof,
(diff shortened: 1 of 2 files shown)


pcaversaccio (commented):
I would clarify that `proof` must have the same value as `root` otherwise it doesn't work (will return `false`). Also, since you make an equality check, it's actually not a no-op (at least for my understanding)

ernestognw:
Right, good point. I'm considering whether the note should be in `processMultiproof` too. wdyt about the following?

```
 * NOTE: Consider the case where \`root == proof[0] && leaves.length == 0\` as it will return true. See {processMultiProof}.
 * Consider disallowing this case if you're not validating the leaves elsewhere.
 function multiProofVerify(...) { ... }

 * NOTE: The _empty set_ (i.e. the case where \`proof.length == 1 && leaves.length == 0\`) is considered a noop,
 * and therefore a valid multiproof (i.e. it returns \`proof[0]\`). Consider disallowing this case if you're not
 * validating the leaves elsewhere.
 function processMultiProof(...) { ... }
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -169,9 +169,8 @@ library MerkleProof {
      *
      * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
      *
-     * NOTE: The _empty set_ (i.e. the case where `proof.length == 0 && leaves.length == 0`) is considered a noop,
-     * and therefore a valid multiproof (i.e. it returns `true`). Consider disallowing this case if you're not
-     * validating the leaves elsewhere.
+     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return true.
```
pcaversaccio, line 172:
```suggestion
     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return `true`.
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -193,6 +192,10 @@ library MerkleProof {
      * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
      * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
      * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+     *
+     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
```
pcaversaccio, line 196:
```suggestion
     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a no-op,
```

pcaversaccio: I think the correct way to write this is NOP, no-op or NOOP (see [here](https://en.wikipedia.org/wiki/NOP_(code))). Feel free to ignore this nit comment.

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -252,9 +255,8 @@ library MerkleProof {
      *
      * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
      *
-     * NOTE: The _empty set_ (i.e. the case where `proof.length == 0 && leaves.length == 0`) is considered a noop,
-     * and therefore a valid multiproof (i.e. it returns `true`). Consider disallowing this case if you're not
-     * validating the leaves elsewhere.
+     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return true.
```
pcaversaccio, line 258:
```suggestion
     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return `true`.
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -277,6 +279,10 @@ library MerkleProof {
      * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
      * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
      * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+     *
+     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
```
pcaversaccio, line 283:
```suggestion
     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a no-op,
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -337,9 +343,8 @@ library MerkleProof {
      *
      * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
      *
-     * NOTE: The _empty set_ (i.e. the case where `proof.length == 0 && leaves.length == 0`) is considered a noop,
-     * and therefore a valid multiproof (i.e. it returns `true`). Consider disallowing this case if you're not
-     * validating the leaves elsewhere.
+     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return true.
```
pcaversaccio, line 346:
```suggestion
     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return `true`.
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -361,6 +366,10 @@ library MerkleProof {
      * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
      * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
      * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+     *
+     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
```
pcaversaccio, line 370:
```suggestion
     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a no-op,
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -420,9 +429,8 @@ library MerkleProof {
      *
      * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
      *
-     * NOTE: The _empty set_ (i.e. the case where `proof.length == 0 && leaves.length == 0`) is considered a noop,
-     * and therefore a valid multiproof (i.e. it returns `true`). Consider disallowing this case if you're not
-     * validating the leaves elsewhere.
+     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return true.
```
pcaversaccio, line 432:
```suggestion
     * NOTE: Consider the case where `root == proof[0] && leaves.length == 0` as it will return `true`.
```

contracts/utils/cryptography/MerkleProof.sol
```diff
@@ -445,6 +453,10 @@ library MerkleProof {
      * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
      * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
      * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+     *
+     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a noop,
```
pcaversaccio, line 457:
```suggestion
     * NOTE: The _empty set_ (i.e. the case where `proof.length == 1 && leaves.length == 0`) is considered a no-op,
```

scripts/generate/templates/MerkleProof.js
```diff
@@ -89,9 +89,8 @@ const templateMultiProof = ({ suffix, location, visibility, hash }) => `\
  *
  * CAUTION: Not all Merkle trees admit multiproofs. See {processMultiProof} for details.
  *
- * NOTE: The _empty set_ (i.e. the case where \`proof.length == 0 && leaves.length == 0\`) is considered a noop,
- * and therefore a valid multiproof (i.e. it returns \`true\`). Consider disallowing this case if you're not
- * validating the leaves elsewhere.
+ * NOTE: Consider the case where \`root == proof[0] && leaves.length == 0\` as it will return true.
```
pcaversaccio, line 92:
```suggestion
 * NOTE: Consider the case where \`root == proof[0] && leaves.length == 0\` as it will return \`true\`.
```

scripts/generate/templates/MerkleProof.js
```diff
@@ -114,6 +113,10 @@ function multiProofVerify${suffix}(${formatArgsMultiline(
  * CAUTION: Not all Merkle trees admit multiproofs. To use multiproofs, it is sufficient to ensure that: 1) the tree
  * is complete (but not necessarily perfect), 2) the leaves to be proven are in the opposite order they are in the
  * tree (i.e., as seen from right to left starting at the deepest layer and continuing at the next layer).
+ *
+ * NOTE: The _empty set_ (i.e. the case where \`proof.length == 1 && leaves.length == 0\`) is considered a noop,
```
pcaversaccio, line 117:
```suggestion
 * NOTE: The _empty set_ (i.e. the case where \`proof.length == 1 && leaves.length == 0\`) is considered a no-op,
```

pcaversaccio (commented):
LGTM, left some nit formatting suggestions; feel free to ignore. I wanted to ask if it makes sense for OZ contracts to use the NatSpec tag `custom:security` for general, security-related code comments (at the function- and at the contract-level)? I use them extensively in my snekmate contracts; some examples:
- Function-level: https://github.com/pcaversaccio/snekmate/blob/main/src/snekmate/governance/timelock_controller.vy#L496-L499
- Contract-level: https://github.com/pcaversaccio/snekmate/blob/main/src/snekmate/tokens/erc20.vy#L40-L50

cairoeth (approved):

ernestognw:
Thanks for your suggestions @pcaversaccio, I see they're already included.

Re: NatSpecTag -- We avoid these custom tags since they don't appear in the docs. Regardless, I think it'll be a good idea if they provide any specific feature (e.g. how we read custom natspec tags in OZ's upgrades plugins).

A similar idea we haven't implemented but considered is to have a `@custom:override-level` to make clear how risky a specific override is. If we were using a `@custom:security`, I guess it would be mostly about overrides.

PR #5159, opened 2024-08-21
Remove double declaration of `P` in P256

cairoeth:
Removes the double declaration of `P` in P256 to save some gas on `_jAdd`, `isValidPublicKey`, and `_jDouble` operations. Also helps give more room to avoid stack too deep issues, like for #5148.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ernestognw (approved):
LGTM

PR #5160, opened 2024-08-22
Delegate override vote

Amxx:


contracts/governance/utils/VotesOverridable.sol
```diff
@@ -0,0 +1,85 @@
...
+    mapping(address account => Checkpoints.Trace208) private _balanceOfCheckpoints;
+
+    function _delegate(address account, address delegatee) internal virtual override {
+        address oldDelegate = delegates(account);
+
+        _delegateCheckpoints[account].push(clock(), uint160(delegatee));
+
+        emit DelegateChanged(account, oldDelegate, delegatee);
+        _moveDelegateVotes(oldDelegate, delegatee, _getVotingUnits(account));
+    }
```
Amxx, line 27: This doesn't call super._delegate.

My understanding is that this completelly bypasses the storage of `Votes.sol`, and calling super would cause Vote's storage to be used.

This creates several issues though:
- any other extention that override `_delegate` and calls super would be dependant on the linearisation order.
- storage from Votes is still allocated, and is accessible through `Votes._xxx` functions.

contracts/governance/utils/VotesOverridable.sol
```diff
@@ -0,0 +1,85 @@
...
+import {SafeCast} from "../../utils/math/SafeCast.sol";
+
+/**
+ * @dev Extension of {Votes} that adds support for checkpointed delegations and balances. This is required
+ * to use the `GovernorOverrideDelegateVote` extension.
+ */
+abstract contract VotesOverridable is Votes {
+    using Checkpoints for Checkpoints.Trace208;
+
+    error VotesOverridableFutureLookup(uint256 timepoint, uint256 currentTimepoint);
```
Amxx, line 15: For the futur: maybe we should have this extension be an ERC (that depend on ERC-5805) and use a custom error that is similar to `ERC5805FutureLookup`

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,194 @@
...
+import {IERC5805} from "../../interfaces/IERC5805.sol";
+import {SafeCast} from "../../utils/math/SafeCast.sol";
+import {Time} from "../../utils/types/Time.sol";
+import {VotesOverridable} from "../utils/VotesOverridable.sol";
+
+/**
+ * @dev Extension of {Governor} which enables delegatees to override the vote of their delegates. This module requires a
+ * token token that inherits `VotesOverridable`.
+ */
+abstract contract GovernorOverrideDelegateVote is GovernorVotes {
```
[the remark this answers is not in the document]
Amxx: Also, usually counting module does not inherit from GovernorVotes. Could we avoid that in heritance?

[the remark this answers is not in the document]
arr00: It is both a voting module and a counting module. Maybe it could be something like `GovernorCountingAndVotesOverrideDelegate`?

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    function _countVote(
+        uint256 proposalId,
+        address account,
+        uint8 support,
+        uint256 totalWeight,
+        bytes memory params
+    ) internal virtual override returns (uint256) {
+        if (keccak256(params) == keccak256(hex"23b70c8d0000000000000000000000000000000000000000")) {
```
Amxx, line 117: What does "0x23b70c8d" represent. Can we do something more readable ?

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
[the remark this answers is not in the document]
arr00: Yes it can be removed. The reason to inherit `GovernorVotes` is that this is strictly additional functionality to it and it makes inheriting `GovernorVotesQuorumFraction` simpler since it also inherits `GovernorVotes`.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,194 @@
...
+        uint8 support,
+        bytes memory params
+    ) private returns (uint256) {
+        ProposalVote storage proposalVote = _proposalVotes[proposalId];
+        uint256 proposalSnapshot = proposalSnapshot(proposalId);
+        address delegate = _getPastDelegate(account, proposalSnapshot);
+
+        if (proposalVote.overrideVoteReceipt[account].hasVoted) {
+            revert GovernorAlreadyCastVoteOverride(account);
+        }
```
Amxx, line 155: Not being able top override before the delegate has voted is a limitation:
- if a delegate vote at the last minute, you don't have time to override
- if a delegate doesn't vote, the token holder can't do anything (meaning the delegate can prevent users to use their votes by not voting at all)

Can we find a way so that is possible?

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,194 @@
...
+            } else if (delegateVoteReceipt.support == uint8(VoteType.Abstain)) {
+                proposalVote.abstainVotes -= overrideWeight;
+            } else {
+                revert GovernorInvalidVoteType();
+            }
+
+            // Write delegate into the params for event
+            assembly {
+                mstore(add(params, 0x20), or(mload(add(params, 0x20)), shl(64, delegate)))
+            }
```
Amxx, line 187: Not sure we should touch the params like that. There are many positions where another module could read them. If module start tampering with them, that may be confusing.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
arr00: This is a check to ensure that the delegatee does not override vote twice. An override vote can come at any time when voting is allowed.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
arr00: I accidently removed a comment explaining this--sorry about that.

```
// If the params are the selector `0x23b70c8d` (keccak("override")[0:8]) followed by address 0, it is an override vote.
```

The formatting of the params and how the override vote is communicated to indexers is something that definitely could be further discussed.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,194 @@
...
+        bool hasVoted;
+        uint8 support;
+    }
+
+    struct ProposalVote {
+        uint256 againstVotes;
+        uint256 forVotes;
+        uint256 abstainVotes;
+        mapping(address voter => VoteReceipt) voteReceipt;
+        mapping(address voter => VoteReceipt) overrideVoteReceipt;
```
Amxx, line 36: Nobody is going to override the override ... so we should not need to track the support of the override, do we?
Feels to me that the override part of the Receipt only needs to be a boolean.

Also, could we use a single mapping for receipt, with the receipt structure being
- bool hasVoted
- uint8 voteSupport
- bool hasOverriden
- uint256 (or smaller?) overrideWeight

That would give better datalocality when we move to verkle.

Note, you could even merge hasVoted + support by storing `support + 1` (here we control that support is never going to be 255, so overflow are not an issue)

arr00: > Nobody is going to override the override ... so we should not need to track the support of the override, do we? Feels to me that the override part of the Receipt only needs to be a boolean.

Yes this is correct. Also, given that snapshots within `Votes` are stored as `uint208`, we only need to allocate that to `overrideWeight`. So we could fit it all into one slot.

Amxx:
I'd like to discuss the use of vote with params, versus using dedicated functions `caseOverrideVote`, `caseOverrideVoteBySig`

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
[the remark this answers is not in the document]
ernestognw: I agree it's not necessary to inherit from `GovernorVotes`, a developer building on Governor would pick a votes module and a counting module, while the quorum fraction is optional.

In such case, I would name it `GovernorCountingOverride`

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
ernestognw: Good idea to use something like a selector to identify the type of parameters. I agree it requires more discussion but in GovernorCountingFractional [we used `support == 255` as a discriminator](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/c1d6ad5a307945c7825cb3c4c911f2fd017c8725/contracts/governance/extensions/GovernorCountingFractional.sol#L34).

If we were to use a selector, then I wouldn't compute it in runtime, and instead save a constant for it:

```solidity
/// keccak256(abi.encodePacked(bytes4(keccak256('override')), bytes20(address(0x0000000000000000000000000000000000000000))))
bytes32 constant DISCRIMINATOR = 0x4b59cd2c312bedbbf68f0e4aa2fcd2c8476080b4bce6a8cd9a550e8834b2219c
```

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,194 @@
...
+        proposalVote.overrideVoteReceipt[account] = VoteReceipt({hasVoted: true, support: support});
+        if (support == uint8(VoteType.Against)) {
+            proposalVote.againstVotes += overrideWeight;
+        } else if (support == uint8(VoteType.For)) {
+            proposalVote.forVotes += overrideWeight;
+        } else if (support == uint8(VoteType.Abstain)) {
+            proposalVote.abstainVotes += overrideWeight;
+        } else {
+            revert GovernorInvalidVoteType();
+        }
```
ernestognw, line 168: This block is repeated a couple of times, I would try to encapsulate it somehow (may be in a private function that receives an `operator` function pointer, or the `proposalVote` might save votes in an array so that we only do `proposalVote[support] += ...`)

contracts/governance/utils/VotesOverridable.sol
ernestognw: We might use `ERC5805FutureLookup` here too. It's not officially part of the ERC so it doesn't have a strict definition. I think it's a good fit here

contracts/governance/utils/VotesOverridable.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * @dev Extend functionality of the function by checkpointing balances.
+     */
+    function _transferVotingUnits(address from, address to, uint256 amount) internal virtual override {
+        super._transferVotingUnits(from, to, amount);
+        if (from != to) {
+            if (from != address(0)) {
+                Checkpoints.Trace208 storage store = _balanceOfCheckpoints[from];
+                store.push(clock(), uint208(store.latest() - amount));
```
ernestognw, line 60: We may need to use `_getVotingUnits` to keep consistency in how votes are counted:

```solidity
_balanceOfCheckpoints[from].push(clock(), _getVotingUnits(from))
```

Same for the other

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
arr00: Agree with this. I'll use a private function that tallies the votes.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,194 @@
...
+    function _getPastDelegate(address account, uint256 timepoint) internal view virtual returns (address) {
+        return VotesOverridable(address(token())).getPastDelegate(account, timepoint);
+    }
+
+    /**
+     * @dev Fetch the past `balanceOf` for an `account` at a given `timepoint` from the token.
+     */
+    function _getPastBalanceOf(address account, uint256 timepoint) internal view virtual returns (uint256) {
+        return VotesOverridable(address(token())).getPastBalanceOf(account, timepoint);
+    }
```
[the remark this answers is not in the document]
arr00: Given they are only called once each it makes sense to inline them.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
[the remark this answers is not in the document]
arr00: > I agree it's not necessary to inherit from `GovernorVotes`, a developer building on Governor would pick a votes module and a counting module, while the quorum fraction is optional.
> 
> In such case, I would name it `GovernorCountingOverride`

We could theoretically isolate this to be just a counting module and have a different voting module which implements `_getPastDelegate` and `_getPastBalanceOf`. This module would then be dependent on the usage of that voting module. Since they would be so intertwined and relatively useless on their own, I chose this approach which is a voting and counting module in one.

contracts/governance/utils/VotesOverridable.sol
arr00: Agree with this. I changed it to call super.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,181 @@
...
+import {IERC5805} from "../../interfaces/IERC5805.sol";
+import {SafeCast} from "../../utils/math/SafeCast.sol";
+import {Time} from "../../utils/types/Time.sol";
+import {VotesOverridable} from "../utils/VotesOverridable.sol";
+
+/**
+ * @dev Extension of {Governor} which enables delegatees to override the vote of their delegates. This module requires a
+ * token token that inherits `VotesOverridable`.
+ */
+abstract contract GovernorOverrideDelegateVote is GovernorVotes {
```
Amxx, line 16: If we agree this is a voting module, then we should probably name it in a way that is succint yet explicit about what it achieves. I think the current name may be a bit too long.

This is just a proposition:
```suggestion
abstract contract GovernorCountOverridable is GovernorVotes {
```

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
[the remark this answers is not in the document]
Amxx: I think inheriting from GovernorVotes to get acces to `token()` is fine.

For the name I think it should be succinct yet explicit about what it achieves. I think the current name may be a bit too long.

This is just a proposition:
```suggestion
abstract contract GovernorCountOverridable is GovernorVotes {
```

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,181 @@
...
+    enum VoteType {
+        Against,
+        For,
+        Abstain
+    }
+
+    struct VoteReceipt {
+        uint8 support;
+        bool hasOverriden;
+        uint208 overrideWeight;
```
Amxx, line 29: With the two other args, we could make it uint240 and still fit in one slot. Any reason to use 208?

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    struct ProposalVote {
+        uint256 againstVotes;
+        uint256 forVotes;
+        uint256 abstainVotes;
```
Amxx, line 35: idea: if we make this a `uint256[3] votes` then we can access the value dirrectly by doing `votes[support]` without needing to use `if`s.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,181 @@
...
+        mapping(address voter => VoteReceipt) voteReceipt;
+    }
+
+    error GovernorAlreadyCastVoteOverride(address account);
```
Amxx, line 39: IMO that would be more "natural" to read, what do you think ?
```suggestion
    error GovernorVoteAlreadyOverriden(address account);
```

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,181 @@
...
+     * @dev Supported vote types. Matches Governor Bravo ordering.
+     */
+    enum VoteType {
+        Against,
+        For,
+        Abstain
+    }
+
+    struct VoteReceipt {
+        uint8 support;
```
Amxx, line 27: if this stores "support + 1", maybe we should not name it suppor, but something different.

most basic name would be `supportPlusOne`, but it could be anything that makes the distinction clear.

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,181 @@
...
+        uint256 forVotes;
+        uint256 abstainVotes;
+        mapping(address voter => VoteReceipt) voteReceipt;
+    }
+
+    error GovernorAlreadyCastVoteOverride(address account);
+
+    mapping(uint256 proposalId => ProposalVote) private _proposalVotes;
+
+    constructor(VotesOverridable tokenAddress) GovernorVotes(tokenAddress) {}
```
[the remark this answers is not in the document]
Amxx:
```suggestion
```

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
Amxx: For example, some tally operation could become:

if (support > uint8(Abstain)) {
    proposalVote.votes[support] += delta;
} else {
    revert GovernorInvalidVoteType();
}

contracts/governance/utils/VotesOverridable.sol
```diff
@@ -0,0 +1,75 @@
...
+     * Requirements:
+     *
+     * - `timepoint` must be in the past. If operating using block numbers, the block must be already mined.
+     */
+    function getPastDelegate(address account, uint256 timepoint) public view returns (address) {
+        uint48 currentTimepoint = clock();
+        if (timepoint >= currentTimepoint) {
+            revert VotesOverridableFutureLookup(timepoint, currentTimepoint);
+        }
+        return address(uint160(_delegateCheckpoints[account].upperLookupRecent(SafeCast.toUint48(timepoint))));
```
Amxx, line 39: that is already a uint160. 
```suggestion
        return address(_delegateCheckpoints[account].upperLookupRecent(SafeCast.toUint48(timepoint)));
```

contracts/governance/utils/VotesOverridable.sol
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @dev Extend functionality of the function by checkpointing balances.
+     */
+    function _transferVotingUnits(address from, address to, uint256 amount) internal virtual override {
+        super._transferVotingUnits(from, to, amount);
+        if (from != to) {
+            if (from != address(0)) {
+                Checkpoints.Trace208 storage store = _balanceOfCheckpoints[from];
+                store.push(clock(), uint208(_getVotingUnits(from)));
```
Amxx, line 50:
```suggestion
                _balanceOfCheckpoints[from].push(clock(), uint208(_getVotingUnits(from)));
```

contracts/governance/utils/VotesOverridable.sol
```diff
@@ -0,0 +1,75 @@
...
+            }
+            if (to != address(0)) {
+                Checkpoints.Trace208 storage store = _balanceOfCheckpoints[to];
+                store.push(clock(), uint208(_getVotingUnits(to)));
```
Amxx, line 54:
```suggestion
                _balanceOfCheckpoints[to].push(clock(), uint208(_getVotingUnits(to)));
```

contracts/governance/extensions/GovernorOverrideDelegateVote.sol
Amxx: Following discussion  with Tally, override should ahve its own public facing function(s). This removes the need for params checks here.

PR #5166, opened 2024-08-28
Add toUint, toInt and hexToUint to Strings

Amxx:
As discussed. Usefull for parsing addresses from CAIP10 identifier (strings)

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/utils/Strings.sol
```diff
@@ -115,4 +120,85 @@ library Strings {
...
+        bytes memory buffer = bytes(input);
+
+        // check presence of a negative sign.
+        uint256 offset = bytes1(buffer) == 0x2d ? 1 : 0;
+        int8 factor = bytes1(buffer) == 0x2d ? int8(-1) : int8(1);
+
+        int256 result = 0;
+        for (uint256 i = offset; i < buffer.length; ++i) {
+            result *= 10; // will revert if overflow
+            result += factor * int8(_parseChr(buffer[i], 10)); // parseChr is at most 35, it fits into an int8
```
Amxx, line 159:
```suggestion
            result += factor * int8(_parseChr(buffer[i], 10)); // parseChr is at most 9, it fits into an int8
```

.changeset/eighty-hounds-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Strings`: Add `toUint`, `toInt` and `hexToUint` to parse strings into numbers.`
```
cairoeth, line 5:
```suggestion
`Strings`: Add `toUint`, `toInt` and `hexToUint` to parse strings into numbers.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +120,85 @@ library Strings {
...
+        }
+        return result;
+    }
+
+    /**
+     * @dev Parse an decimal string and returns the value as a `int256`.
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9].
+     * - the result does not fit in a int256.
```
cairoeth, line 147:
```suggestion
     * @dev Parse a string in decimal (i.e. base 10) and returns the value as a `int256`.
     *
     * This function will revert if:
     * - the string contains any character (outside the prefix) that is not in [0-9].
     * - the result does not fit in a `int256`.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +120,85 @@ library Strings {
     function equal(string memory a, string memory b) internal pure returns (bool) {
         return bytes(a).length == bytes(b).length && keccak256(bytes(a)) == keccak256(bytes(b));
     }
+
+    /**
+     * @dev Parse an decimal string and returns the value as a `uint256`.
+     *
+     * This function will revert if:
+     * - the string contains any character that is not in [0-9].
+     * - the result does not fit in a uint256.
```
cairoeth, line 129:
```suggestion
     * @dev Parse a string in decimal (i.e. base 10) and returns the value as a `uint256`.
     *
     * This function will revert if:
     * - the string contains any character that is not in [0-9].
     * - the result does not fit in a `uint256`.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +120,85 @@ library Strings {
...
+        }
+        return result;
+    }
+
+    /**
+     * @dev Parse an hexadecimal string (with or without "0x" prefix), and returns the value as a `uint256`.
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9a-fA-F].
+     * - the result does not fit in a uint256.
```
cairoeth, line 169:
```suggestion
     * @dev Parse a hexadecimal string (with or without "0x" prefix), and returns the value as a `uint256`.
     *
     * This function will revert if:
     * - the string contains any character (outside the prefix) that is not in [0-9a-fA-F].
     * - the result does not fit in a `uint256`.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +120,85 @@ library Strings {
...
+    function toUint(string memory input) internal pure returns (uint256) {
+        bytes memory buffer = bytes(input);
+
+        uint256 result = 0;
+        for (uint256 i = 0; i < buffer.length; ++i) {
+            result *= 10; // will revert if overflow
+            result += _parseChr(buffer[i], 10);
+        }
+        return result;
+    }
```
cairoeth, line 140: this would reduce gas by 3.36% (444 gas) on avg. for `toUint`
```suggestion
    function toUint(string memory input) internal pure returns (uint256 result) {
        bytes memory buffer = bytes(input);

        uint256 bufferLength = buffer.length;
        for (uint256 i = 0; i < bufferLength; ++i) {
            result *= 10; // will revert if overflow
            result += _parseChr(buffer[i], 10);
        }
    }
```

contracts/utils/Strings.sol
19714, line 1: contracts/utils/Strings.sol

19714: contracts/utils/Strings.sol

contracts/utils/Strings.sol
```diff
@@ -115,4 +120,85 @@ library Strings {
...
+        // - Case 1: [0-9]
+        // - Case 2: [a-z]
+        // - Case 2: [A-Z]
+        // - otherwise not supported
+        unchecked {
+            if (uint8(chr) > 47 && uint8(chr) < 58) result = uint8(chr) - 48;
+            else if (uint8(chr) > 96 && uint8(chr) < 123) result = uint8(chr) - 87;
+            else if (uint8(chr) > 64 && uint8(chr) < 91) result = uint8(chr) - 55;
+            else revert StringsInvalidChar(chr, base);
+        }
```
cairoeth, line 198: we can do casting once and just use `result`
```suggestion
        uint8 result = uint8(chr);

        // Try to parse `chr`:
        // - Case 1: [0-9]
        // - Case 2: [a-z]
        // - Case 2: [A-Z]
        // - otherwise not supported
        unchecked {
            if (result > 47 && result < 58) result -= 48;
            else if (result > 96 && result < 123) result -= 87;
            else if (result > 64 && result < 91) result -= 55;
            else revert StringsInvalidChar(chr, base);
        }
```

contracts/utils/Strings.sol
Amxx: Interresting. The compiler should optimize "length caching" by itself.

cairoeth: i'm not aware it caches the length yet -- it does the optimization to `i++` unchecked as part of Solidity 0.8.22 tho. here's the gas diff on `toUint` before and after using the buffer length cached "manually":

before:
| test/utils/Strings.t.sol:Gas contract |                 |       |        |       |         |
|---------------------------------------|-----------------|-------|--------|-------|---------|
| Deployment Cost                       | Deployment Size |       |        |       |         |
| 220588                                | 803             |       |        |       |         |
| Function Name                         | min             | avg   | median | max   | # calls |
| toUint                                | 1133            | 14165 | 10693  | 37951 | 256     |

after:
Suite result: ok. 4 passed; 0 failed; 0 skipped; finished in 514.46ms (1.41s CPU time)
| test/utils/Strings.t.sol:Gas contract |                 |       |        |       |         |
|---------------------------------------|-----------------|-------|--------|-------|---------|
| Deployment Cost                       | Deployment Size |       |        |       |         |
| 221452                                | 807             |       |        |       |         |
| Function Name                         | min             | avg   | median | max   | # calls |
| toUint                                | 1141            | 11482 | 4941   | 37728 | 256     |


function:
```solidity
    function toUint(string memory input) internal pure returns (uint256) {
        bytes memory buffer = bytes(input);
        uint256 bufferLength = buffer.length;

        uint256 result = 0;
        for (uint256 i = 0; i < bufferLength; ++i) {
            result *= 10; // will revert if overflow
            result += _parseChr(buffer[i], 10);
        }
        return result;
    }
```

contracts/utils/Strings.sol
Amxx: When looking online, It looks like "decimal string" is the correct way. Maybe "string in decimal format" would be ok, but "string in decimal" doesn't look commonly used.

Amxx: https://books.google.com/ngrams/graph?content=decimal+string%2Cstring+in+decimal&year_start=1800&year_end=2019&corpus=en-2019&smoothing=0

contracts/utils/Strings.sol
Amxx: I'd rename `result` to `value` if we use that trick

Amxx:
```suggestion
        uint8 value = uint8(chr);

        // Try to parse `chr`:
        // - Case 1: [0-9]
        // - Case 2: [a-z]
        // - Case 2: [A-Z]
        // - otherwise not supported
        unchecked {
            if (value > 47 && value < 58) value -= 48;
            else if (value > 96 && value < 123) value -= 87;
            else if (value > 64 && value < 91) value -= 55;
            else revert StringsInvalidChar(chr, base);
        }
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +120,83 @@ library Strings {
...
+
+        uint256 bufferLength = buffer.length;
+        for (uint256 i = 0; i < bufferLength; ++i) {
+            result *= 10; // will revert if overflow
+            result += _parseChr(buffer[i], 10);
+        }
+    }
+
+    /**
+     * @dev Parse an decimal string and returns the value as a `int256`.
```
Amxx, line 142:
```suggestion
     * @dev Parse a decimal string and returns the value as a `int256`.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +121,89 @@ library Strings {
...
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9].
+     * - the result does not fit in a `int256`.
+     */
+    function toInt(string memory input) internal pure returns (int256) {
+        bytes memory buffer = bytes(input);
+
+        // check presence of a negative sign.
+        bool isNegative = bytes1(buffer) == 0x2d;
```
cairoeth, line 155: what about the true negative char "−" (0x2212)?

Amxx: is that comonly used/supported? 
- In my local node instance, using it causes a `Uncaught SyntaxError: Invalid or unexpected token`
- In my local python instance, using it causes a `SyntaxError: invalid character '−' (U+2212)`

I honestly don't think we should bother with non ascii characters.

cairoeth (approved):

contracts/utils/Strings.sol
```diff
@@ -115,4 +121,89 @@ library Strings {
...
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9a-fA-F].
+     * - the result does not fit in a `uint256`.
+     */
+    function hexToUint(string memory input) internal pure returns (uint256) {
+        bytes memory buffer = bytes(input);
+
+        // skip 0x prefix if present. Length check doesn't appear to be critical
+        uint256 offset = bytes2(buffer) == 0x3078 ? 2 : 0;
```
Amxx, line 178:
```suggestion
        uint256 offset = (bytes2(buffer) == 0x3078).ternary(2, 0);
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +121,89 @@ library Strings {
...
+     * @dev Parse a hexadecimal string (with or without "0x" prefix), and returns the value as a `uint256`.
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9a-fA-F].
+     * - the result does not fit in a `uint256`.
+     */
+    function hexToUint(string memory input) internal pure returns (uint256) {
+        bytes memory buffer = bytes(input);
+
+        // skip 0x prefix if present. Length check doesn't appear to be critical
```
cairoeth, line 177: second sentence feels out of place
```suggestion
        // skip 0x prefix if present
```

Amxx: Just for the record, this is what the comment was about:

The string length may be less than 2. String could be empty, of just `"1"`. In some cases, doing a `input[1]` or even a `input[0]` would revert (out of bound acces). Doing `bytes2(input)` does not revert if the string is to short.

So we check the length to verify that it ok to read the prefix ? It turn out no.

- If the string is empty, then regardless of the result of that lookup (that could read dirty bytes), the loop will not run (because length == 0), and the result will be 0 -> that is ok
- If the string has length 1 then we have two options
  - the check identifies the prefix
    - that means the string is `"0"`, and there is a dirty `x` after. 
    - In that case we have an offset of 2, and the length is 1, so the for loop does not run and the function returns 0 -> that is ok
  - the check does not find the prefix
    - the only "digit" is read by the loop, and the result should be just fine
- If the string has length >= 2, then the prefix lookup is in the bounds

That is the long explanation (I'm happy its visible in the PR 😃) to something that is not really trivial, and can be missed, but missing it is not a risk. 

We may get questions about it though ...

cairoeth (approved):

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+        bool isNegative = bytes1(unsafeReadBytesOffset(buffer, begin)) == 0x2d;
+        int8 factor = isNegative ? int8(-1) : int8(1);
+        uint256 offset = isNegative.toUint();
+
+        int256 result = 0;
+        for (uint256 i = begin + offset; i < end; ++i) {
+            uint8 chr = _tryParseChr(buffer[i]);
+            if (chr > 9) return (false, 0);
+            result *= 10;
+            result += factor * int8(chr);
```
ernestognw, line 238: I guess we can apply the factory until the end, but I know that might increase the scope right now since we'd need to check for the limits `[-(2**255), 2**255)`.

It's ok rn but we can consider for an optimization

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+        }
+        return (true, result);
+    }
+
+    /**
+     * @dev Parse a decimal string and returns the value as a `int256`.
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9].
+     * - the result does not fit in a `int256`.
```
ernestognw, line 190:
```suggestion
     * Requirements:
     * - The provided string must contain only ASCII characters in the range of [0-9] (excluding the negative sign prefix).
     * - The result must fit in an `int256` type.

File: contracts/utils/Strings.sol [hunk 14]
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+     */
+    function tryParseInt(
+        string memory input,
+        uint256 begin,
+        uint256 end
+    ) internal pure returns (bool success, int256 value) {
+        bytes memory buffer = bytes(input);
+
+        // check presence of a negative sign.
+        bool isNegative = bytes1(unsafeReadBytesOffset(buffer, begin)) == 0x2d;
```
ernestognw, line 229: Is it intentional that we're not considering the `+` sign (`0x2b`)? In my opinion ignoring the plus sign is fine as long as we are explicit in the Natspec

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+        return (true, result);
+    }
+
+    /**
+     * @dev Parse a decimal string and returns the value as a `int256`.
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9].
+     * - the result does not fit in a `int256`.
+     */
```
ernestognw, line 191: We might like to note this

```suggestion
     *
     * NOTE: This function does not accept strings with a plus sign prefix (i.e. `+`)
     */
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+    function parseInt(string memory input, uint256 begin, uint256 end) internal pure returns (int256) {
+        (bool success, int256 value) = tryParseInt(input, begin, end);
+        if (!success) revert StringsInvalidChar();
+        return value;
+    }
+
+    /**
+     * @dev Variant of {parseInt-string} that returns false if the parsing fails because of an invalid character.
+     *
+     * This function will still revert if the result does not fit in a `int256`
```
ernestognw, line 209:
```suggestion
     * Requirements:
     * - The result must fit into an `int256` type.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+        }
+        return (true, result);
+    }
+
+    /**
+     * @dev Parse a hexadecimal string (with or without "0x" prefix), and returns the value as a `uint256`.
+     *
+     * This function will revert if:
+     * - the string contains any character (outside the prefix) that is not in [0-9a-fA-F].
+     * - the result does not fit in a `uint256`.
```
ernestognw, line 248:
```suggestion
     * Requirements:
     * - The provided string must contain only ASCII characters in [0-9a-fA-F] (excluding the negative sign prefix).
     * - The result must fit in an `uint256` type.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+    function parseHex(string memory input, uint256 begin, uint256 end) internal pure returns (uint256) {
+        (bool success, uint256 value) = tryParseHex(input, begin, end);
+        if (!success) revert StringsInvalidChar();
+        return value;
+    }
+
+    /**
+     * @dev Variant of {parseHex-string} that returns false if the parsing fails because of an invalid character.
+     *
+     * This function will still revert if the result does not fit in a `uint256`
```
ernestognw, line 267:
```suggestion
     * Requirements:
     * - The result must fit into an `uint256` type.
```

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+     */
+    function tryParseHex(string memory input) internal pure returns (bool success, uint256 value) {
+        return tryParseHex(input, 0, bytes(input).length);
+    }
+
+    /**
+     * @dev Variant of {parseHex-string-uint256-uint256} that returns false if the parsing fails because of an
+     * invalid character.
+     *
+     * This function will still revert if the result does not fit in a `uint256`
```
ernestognw, line 277:
```suggestion
     * Requirements:
     * - The result must fit into an `uint256` type.
```

ernestognw (commented):
I see we added more variations of the base functions and understand that we might want to reduce it. However, I like that the interface is pretty versatile (especially by providing pointers).

I'm not too worried about the length of this particular library since it's not procedurally generated

contracts/utils/Strings.sol
```diff
@@ -115,4 +128,247 @@ library Strings {
...
+            // length garantees that this does not overflow, and value2 is at most type(uint160).max
+            (bool s, uint256 v) = tryParseHex(input, begin, end);
+            return (s, address(uint160(v)));
+        } else {
+            return (false, address(0));
+        }
+    }
+
+    // TODO: documentation.
+    function unsafeReadBytesOffset(bytes memory buffer, uint256 offset) internal pure returns (bytes32 value) {
```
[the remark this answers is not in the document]
Amxx: Where should that go ? A Bytes library ?

contracts/utils/Strings.sol
```diff
@@ -115,4 +120,85 @@ library Strings {
...
+
+    /**
+     * @dev Parse an decimal string and returns the value as a `uint256`.
+     *
+     * This function will revert if:
+     * - the string contains any character that is not in [0-9].
+     * - the result does not fit in a uint256.
+     */
+    function toUint(string memory input) internal pure returns (uint256) {
+        bytes memory buffer = bytes(input);
```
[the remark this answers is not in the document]
Amxx: My understanding is that this is copying the reference

Basically:
- `input` is a "pointer" (uint256 value) that points to the location of the string in memory.
- `buffer` is also a "pointer", that points to the same location.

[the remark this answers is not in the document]
Amxx: If you want to try for yourself:
```
contract Test {
    function test(string memory input) public pure returns (uint256 x, uint256 y) {
        assembly { x := mload(0x40) }
        bytes memory buffer = bytes(input);
        buffer;
        assembly { y := mload(0x40) }
    }
}
```

contracts/utils/Strings.sol
Amxx: we can add support for + if we think that feature is usefull.

contracts/utils/Strings.sol
Amxx: `int256` range from
`-2**255  = -57896044618658097711785492504343953926634992332820282019728792003956564819968`
to
`2**255-1 = 57896044618658097711785492504343953926634992332820282019728792003956564819967`

This means MinUint256 would not fit in a int256 if we don't consider the sign from the start. An alternative would be to parse as a `uint256` ... but then I'm not sure about how to do the casting

Amxx: I'd do something like

```solidity
    uint256 private constant ABS_MIN_INT256 = 2 ** 255;

    function tryParseInt(
        string memory input,
        uint256 begin,
        uint256 end
    ) internal pure returns (bool success, int256 value) {
        bytes memory buffer = bytes(input);

        // check presence of a negative sign.
        bool positiveSign = bytes1(unsafeReadBytesOffset(buffer, begin)) == bytes1("+");
        bool negativeSign = bytes1(unsafeReadBytesOffset(buffer, begin)) == bytes1("-");
        uint256 offset = (positiveSign || negativeSign).toUint();

        (bool subSuccess, uint256 absValue) = tryParseUint(input, begin + offset, end);

        if (subSuccess && !negativeSign && absValue < ABS_MIN_INT256) {
            return (true, int256(absValue));
        } else if (subSuccess && negativeSign && absValue <= ABS_MIN_INT256) {
            return (true, int256(-absValue));
        } else {
            return (false, 0);
        }
    }
```
but the negative uint256 -> int256 case is not clear to me.

Amxx: Maybe

```
if (subSuccess && negativeSign && absValue == ABS_MIN_INT256) return (true, type(int256).min);
else if (subSuccess && absValue < ABS_MIN_INT256) return (true, negativeSign ? -int256(absValue) : int256(absValue));
else return (false, 0);

File: contracts/utils/Strings.sol (hunk 20 above)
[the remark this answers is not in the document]
Comment (Amxx) [reply contracts/utils/Strings.sol:372]: Note that the same function exists in RSA.sol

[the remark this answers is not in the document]
Comment (frangio) [reply contracts/utils/Strings.sol:372]: I believe this function was present in other places but always private.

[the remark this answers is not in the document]
Comment (frangio) [reply contracts/utils/Strings.sol:370]: > Yes. private, and marked as memory safe, because the private calls are all enforcing the safety.

Right... That was the issue. That using memory-unsafe functions disables some optimizations globally, so they were important to avoid.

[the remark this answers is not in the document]
Comment (Amxx) [reply contracts/utils/Strings.sol:370]: So do we want to have:
- multiple private (and memory safe) implementations
- one internal, none memory safe implementation 
- one internal, memory safe (checks memory bound) implementation
?

**Comment by Amxx** [comment]
Added a Bytes.sol library. The CAIP-2/CAIP-10 PR will put more logic into that library.

File: contracts/utils/Bytes.sol [hunk 22]
```diff
@@ -0,0 +1,18 @@
...
+
+import {Math} from "./math/Math.sol";
+
+/**
+ * @dev Bytes operations.
+ */
+library Bytes {
+    /// @dev Reads a bytes32 from a bytes array without bounds checking.
+    function unsafeReadBytesOffset(bytes memory buffer, uint256 offset) internal pure returns (bytes32 value) {
+        // This is not memory safe in the general case
```
ernestognw, line 13: Why is it not memory safe if it's just reading? I think we can safely annotate it:

contracts/utils/Strings.sol
```diff
@@ -113,4 +128,261 @@ library Strings {
...
+     */
+    function tryParseInt(string memory input) internal pure returns (bool success, int256 value) {
+        return tryParseInt(input, 0, bytes(input).length);
+    }
+
+    /**
+     * @dev Variant of {parseInt-string-uint256-uint256} that returns false if the parsing fails because of an invalid
+     * character.
+     *
+     * This function will still revert if the result does not fit in a `int256`
```
ernestognw, line 230: Since this is a `try` function, shouldn't it return `success == false`?

contracts/utils/Strings.sol
```diff
@@ -113,4 +128,261 @@ library Strings {
...
+    function tryParseHex(string memory input) internal pure returns (bool success, uint256 value) {
+        return tryParseHex(input, 0, bytes(input).length);
+    }
+
+    /**
+     * @dev Variant of {parseHex-string-uint256-uint256} that returns false if the parsing fails because of an
+     * invalid character.
+     *
+     * Requirements:
+     * - The result must fit into an `uint256` type.
```
ernestognw, line 295: Same as `tryParseInt`

contracts/utils/Strings.sol
```diff
@@ -113,4 +128,261 @@ library Strings {
...
+     */
+    function parseAddress(string memory input, uint256 begin, uint256 end) internal pure returns (address) {
+        (bool success, address value) = tryParseAddress(input, begin, end);
+        if (!success) revert StringsInvalidAddressFormat();
+        return value;
+    }
+
+    /**
+     * @dev Variant of {parseAddress-string} that returns false if the parsing fails because the input is not a properly
+     * formatted address. See {parseAddress} requirements.
```
ernestognw, line 343: Same as `tryParseInt `

ernestognw (commented):
Almost there

contracts/utils/Strings.sol
```diff
@@ -113,4 +128,261 @@ library Strings {
...
+     * @dev Variant of {parseAddress-string-uint256-uint256} that returns false if the parsing fails because input is not a properly
+     * formatted address. See {parseAddress} requirements.
+     */
+    function tryParseAddress(
+        string memory input,
+        uint256 begin,
+        uint256 end
+    ) internal pure returns (bool success, address value) {
+        // check that input is the correct length
+        bool hasPrefix = bytes2(bytes(input).unsafeReadBytesOffset(begin)) == 0x3078;
```
arr00, line 359: Using prefix `0x` as bytes2 multiple times. Could we declare it as a constant for readability?

contracts/utils/Strings.sol
```diff
@@ -113,4 +128,261 @@ library Strings {
...
+        // skip 0x prefix if present
+        bool hasPrefix = bytes2(buffer.unsafeReadBytesOffset(begin)) == bytes2("0x");
+        uint256 offset = hasPrefix.toUint() * 2;
+
+        uint256 result = 0;
+        for (uint256 i = begin + offset; i < end; ++i) {
+            uint8 chr = _tryParseChr(buffer[i]);
+            if (chr > 15) return (false, 0);
+            result *= 16;
+            result += chr;
```
arr00, line 313:
```suggestion
            assembly ("memory-safe") {
                result := or(shl(4, result), chr)
            }
```

Would something like this be more efficient? Why don't we return a `bytes32`.

.changeset/eighty-hounds-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Strings`: Add `parseUint`, `parseInt`, `parseHex` and `parseAddress` to parse strings into numbers and addresses. Also provide variant of these function that parse substrings, and `tryXxx` variants that do not revert on invalid input.
```
cairoeth, line 5:
```suggestion
`Strings`: Add `parseUint`, `parseInt`, `parseHex` and `parseAddress` to parse strings into numbers and addresses. Also provide variants of these functions that parse substrings, and `tryXxx` variants that do not revert on invalid input.
```

contracts/utils/Bytes.sol
cairoeth: this is not memory safe because it's possible that the block reads from memory outside the bounds allocated by solidity (it's not checking bounds)

.changeset/rude-cougars-look.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Bytes`: Add a library of common operation that operate on `bytes` objects.
```
cairoeth, line 5:
```suggestion
`Bytes`: Add a library for common operations on `bytes` objects.
```

contracts/governance/Governor.sol
```diff
@@ -760,68 +762,24 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
-            //   - (description + 32) + (len - 52) = description + (len - 20)
-            // - Note: Solidity will ignore anything past the first 12 bytes
-            marker := mload(add(description, sub(len, 20)))
-        }
+        // Extract what would be the `#proposer=` marker beginning the suffix
+        bytes10 marker = bytes10(Bytes.unsafeReadBytesOffset(bytes(description), length - 52));
 
         // If the marker is not found, there is no proposer suffix to check
-        if (marker != bytes12("#proposer=0x")) {
+        if (marker != bytes10("#proposer=")) {
```
cairoeth, line 776: doesn't look beautiful but we can save the casting cost with the hardcoded value, and add a comment for visibility
```suggestion
        if (marker != 0x2370726f706f7365723d00000000000000000000000000000000000000000000) {
```

cairoeth: the good thing that it allows to show that if the proposer is zero it returns true as well

contracts/utils/Bytes.sol
ernestognw: Ah got it. I consider it may be better to annotate as `memory-safe` and check bounds. Otherwise, [optimizations get disabled](https://forum.soliditylang.org/t/assembly-and-solidity-optimizations/2053/3), perhaps it's cheaper to check and annotate than not checking (when optimizations are applied)

Amxx: The entier point of this function is to not check the bound. Checking the bound require loading the length, which is one expensive mload.

ernestognw: I thought the entire point was to make it cheaper. I think disabling optimizations we're doing the opposite.

Amxx: Basically this function is already doing stuff as simple as possible. Not sure any further optimization is required from the compiler for this block.

contracts/utils/Strings.sol
Amxx: The revert is caused by 
```solidity
result *= 10;
result += factor * int8(chr);
```
overflowing.

Note that this is done in a loop. Any branch that you add here will be multiplied by the number of characteres in the substring. 

If you can find a nice and cheap way to return false, please feel free to change that

contracts/utils/Bytes.sol
ernestognw: From [here](https://forum.soliditylang.org/t/assembly-and-solidity-optimizations/2053/3):

> Assuming that you’re using the via-ir pipeline, having assembly blocks that are not memory safe (i.e. have memory access within the block, but are not explicitly marked as memory safe) will prevent the ‘stack-to-memory’ mover, and will also not inline as aggressively (inlining is reduced to smaller functions than would be the case with memory safe blocks)

My point is that by having these assembly block that's not memory safe we might be hurting the performance of the whole contract.

contracts/utils/Strings.sol
Amxx: its just twice. I'm not sure having a constant would really help readability here.

contracts/utils/Strings.sol
Amxx: If the result overflow, this will return incorrect values. Also its way less readable

If you want to save gas, you could do just the same logic using unchecked instead of assembly ... but again you would have silent invalid result in case of overflow.

contracts/governance/Governor.sol
Amxx: AFAIK there is no casting cost here. Have you been able to observe any difference?

contracts/utils/Strings.sol
arr00: Overflow occurs if the length is > 64 (assuming no leading 0s) so it’s easy to calculate and revert.

Amxx: Not sure we should assume there are no leading zero.
Also, that won't work for the base 10 variant.

contracts/governance/Governor.sol
Amxx: Capture d’écran du 2024-10-11 21-38-59

I tried these two, they produce the exact same bytecode

cairoeth: okk makes sense. thanks for checking!

Amxx: Capture d’écran du 2024-10-11 21-40-18

I tried these two, they produce the exact same bytecode

contracts/utils/Bytes.sol
Amxx: so: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5166#discussion_r1765350468

contracts/utils/Strings.sol
cairoeth: doesn't the revert also happen for `tryParseUint`?

contracts/utils/Strings.sol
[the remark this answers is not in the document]
Amxx: Extra option is:

- one internal, that is not actually memory safe, but that we mark as `"memory-safe"`. And we document it should only be used with input that are known to be within the bounds

contracts/utils/Strings.sol
Amxx: Yes, it affect all variants (3) of this function. (uint, int, hex)

ernestognw: Thanks all for your comments. I think the contract is currently well-documented anyway, and updating to a version that doesn't revert (in the future) is the kind of breaking change that'd be fine in a minor version.

I wouldn't block merging

contracts/utils/Strings.sol
arr00: Switching to `bytes2("0x")` is preferred given https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5166#discussion_r1797359676. I see that was done.

contracts/utils/Strings.sol
arr00: What you committed makes sense--I prefer the shift as its probably more logical when thinking of bytes32 hex. Still confused why a `uint256` is returned.

Amxx: I think returning uint256 of bytes32 is equivalent. 
@ernestognw do you have a preference ?

ernestognw: Agree is equivalent. I personally prefer `uint256` since parsing string to bytes is already easy, and I seems that the only reason to parse to a `bytes32` would be to use it as a number (otherwise it'd be already `bytes32`)

Amxx: > What you committed makes sense--I prefer the shift as its probably more logical when thinking of bytes32 hex.

Problem with the shift is it gives invalid behavior when result does not fit in 256 bits :/ (no overflow check)

contracts/utils/Strings.sol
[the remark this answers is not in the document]
ernestognw: I'm not sure that the last option is a safe recommendation. I'm too stupid regarding how compiler optimizations work so feel free to disregard, but I wonder if guaranteeing the safetiness with a private function will have a negative effect on the so-called "‘stack-to-memory’ mover"

contracts/utils/Bytes.sol
ernestognw: Although I agree on adding a `Bytes` library, I'm not sure this particular function should go in. Perhaps this whole discussion indicates that the function should remain private in its contract.

I'd be in favor of not adding the Bytes library right now

contracts/utils/Strings.sol
ernestognw: Alternatively, we can check that the length of the string in `tryParseUint` is >= 78, and check for overflow use `Math.tryAdd` in the last digit if the string's length == 78. (78 digits are in 2*256)

Similarly for `tryParseHex` but with 64 digits instead.

contracts/utils/Strings.sol
arr00: > > What you committed makes sense--I prefer the shift as its probably more logical when thinking of bytes32 hex.
> 
> Problem with the shift is it gives invalid behavior when result does not fit in 256 bits :/ (no overflow check)

Checked shift as implemented is good.

contracts/utils/Strings.sol
```diff
@@ -113,4 +128,261 @@ library Strings {
...
+        string memory input,
+        uint256 begin,
+        uint256 end
+    ) internal pure returns (bool success, int256 value) {
+        bytes memory buffer = bytes(input);
+
+        // check presence of a negative sign.
+        bool positiveSign = bytes1(buffer.unsafeReadBytesOffset(begin)) == bytes1("+");
+        bool negativeSign = bytes1(buffer.unsafeReadBytesOffset(begin)) == bytes1("-");
+        uint256 offset = (positiveSign || negativeSign).toUint();
```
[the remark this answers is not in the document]
Amxx:
```suggestion
        // Check presence of a negative sign.
        bytes1 sign = bytes1(buffer.unsafeReadBytesOffset(begin));
        bool positiveSign =  sign == bytes1("+");
        bool negativeSign = sign == bytes1("-");
        uint256 offset = (positiveSign || negativeSign).toUint();
```

contracts/utils/Strings.sol
Amxx: @ernestognw zero prefix would create false positives. I'm not sure we can rule them out.

contracts/utils/Bytes.sol
Amxx: If it was just in this contract I'd agree.

But the truth is that this function is needed:
- in Strings.sol
- in Governor.sol
- in the other Bytes operation that come in #5252

So if we keep it private, we have to define it 3 times, in 3 different contracts/libraries.

Amxx:
I made `unsafeReadBytesOffset` private again, and duplicated it in Governor.sol and Strings.sol.

contracts/utils/README.adoc
```diff
@@ -31,10 +31,11 @@ Miscellaneous contracts and libraries containing utility functions you can use t
  * {Address}: Collection of functions for overloading Solidity's https://docs.soliditylang.org/en/latest/types.html#address[`address`] type.
  * {Arrays}: Collection of functions that operate on https://docs.soliditylang.org/en/latest/types.html#arrays[`arrays`].
  * {Base64}: On-chain base64 and base64URL encoding according to https://datatracker.ietf.org/doc/html/rfc4648[RFC-4648].
+ * {Bytes}: Common bytes operations.
```
ernestognw, line 34:
```suggestion
```

contracts/governance/Governor.sol
```diff
@@ -760,68 +761,24 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
-            // Case a-f
-            else if (96 < c && c < 103) {
-                return (true, c - 87);
-            }
-            // Else: not a hex char
-            else {
-                return (false, 0);
-            }
-        }
+        // Check that the last 42 characters (after the marker) is a properly formatted address.
```
ernestognw, line 779:
```suggestion
        // Check that the last 42 characters (after the marker) are a properly formatted address.
```

contracts/governance/Governor.sol
```diff
@@ -760,68 +761,24 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
+        (bool success, address recovered) = Strings.tryParseAddress(description, length - 42, length);
+        return !success || recovered == proposer;
```
ernestognw, line 781: Sholdn't this be?

```suggestion
        return success && recovered == proposer;
```

ernestognw (commented):
I'd like to reconsider the naming of `parseHex` as raised by @arr00

The function name is ambiguous because it's not clear whether it parses the string as a literal hex (i.e. `hex"1234"`) or as a decimal string (i.e. 1234). I'd suggest renaming to `parseHexUint`

contracts/governance/Governor.sol
Amxx: If the address is not properly formated, then we don't consider it and the proposal should pass. At least that is the current behavior.

For example if you have a description that ends in

```#proposer=NULL&something=xxxxxxxxxxxxxxxxxxxxxxxxxxx```
The "#proposer=" marker is at the right location, but we should not apply any restrictions

**Review by ernestognw — APPROVED** [review]
lfg

File: contracts/governance/Governor.sol (hunk 34 above)
Comment (Amxx) [reply contracts/governance/Governor.sol:781]: For ref: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5166/files#diff-66653f6d39b7c3e38c356d782e05f4871f34553eda930c07a6286344eb44d2e1L791

File: contracts/governance/Governor.sol [hunk 35]
```diff
@@ -760,68 +761,24 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
-            // - Start of the string contents in memory = description + 32
-            // - First character of the marker = len - 52
-            //   - Length of "#proposer=0x0000000000000000000000000000000000000000" = 52
-            // - We read the memory word starting at the first character of the marker:
-            //   - (description + 32) + (len - 52) = description + (len - 20)
-            // - Note: Solidity will ignore anything past the first 12 bytes
-            marker := mload(add(description, sub(len, 20)))
-        }
+        // Extract what would be the `#proposer=` marker beginning the suffix
+        bytes10 marker = bytes10(_unsafeReadBytesOffset(bytes(description), length - 52));
```
cairoeth, line 772: we can remove the private function and just have as it's only used once:
```suggestion
        bytes10 marker;
        assembly ("memory-safe") {
            marker := mload(add(description, sub(length, 52)))
        }
```

cairoeth: length is at least 52 so the `sub` can't underflow

contracts/utils/Strings.sol
```diff
@@ -113,4 +126,275 @@ library Strings {
...
+        bytes1 sign = bytes1(_unsafeReadBytesOffset(buffer, begin));
+        bool positiveSign = sign == bytes1("+");
+        bool negativeSign = sign == bytes1("-");
+        uint256 offset = (positiveSign || negativeSign).toUint();
+
+        (bool absSuccess, uint256 absValue) = tryParseUint(input, begin + offset, end);
+
+        if (absSuccess && absValue < ABS_MIN_INT256) {
+            return (true, negativeSign ? -int256(absValue) : int256(absValue));
+        } else if (absSuccess && negativeSign && absValue == ABS_MIN_INT256) {
```
cairoeth, line 247: curious if there's a reason to not use `type(int256).min`? i think it acts like the constant
```suggestion
        if (absSuccess && absValue < type(int256).max) {
            return (true, negativeSign ? -int256(absValue) : int256(absValue));
        } else if (absSuccess && negativeSign && absValue == type(int256).max) {
```

ernestognw (approved):

contracts/governance/Governor.sol
Amxx: lets mark the function unchecked then

arr00 (approved):

PR #5173, opened 2024-08-29
Add note about 100% royalty in IERC2981

ernestognw:
Fixes L-01

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/interfaces/IERC2981.sol
+++ b/contracts/interfaces/IERC2981.sol
@@ -16,8 +16,8 @@ interface IERC2981 is IERC165 {
      * @dev Returns how much royalty is owed and to whom, based on a sale price that may be denominated in any unit of
      * exchange. The royalty amount is denominated and should be paid in that same unit of exchange.
      *
-     * NOTE: Consider 100% royalty is a valid scenario, therefore contracts performing a royalty transfer with the value
-     * returned by this function may want to check for 0 value transfers.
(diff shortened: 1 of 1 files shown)


contracts/interfaces/IERC2981.sol
```diff
@@ -15,6 +15,9 @@ interface IERC2981 is IERC165 {
     /**
      * @dev Returns how much royalty is owed and to whom, based on a sale price that may be denominated in any unit of
      * exchange. The royalty amount is denominated and should be paid in that same unit of exchange.
+     *
+     * NOTE: Consider 100% royalty is a valid scenario, therefore contracts performing a royalty transfer with the value
+     * returned by this function may want to check for 0 value transfers.
```
Amxx, line 20: suggestion
```suggestion
     * NOTE: ERC-2981 authorizes setting the royalty to 100% of the price. In that scenario all the price would go 
     * toward the royalty receiver, and no tokens would go the the seller. Contracts dealing with royalty should 
     * consider that the transfer of value toward the seller might be empty.
```

ernestognw: I'm keeping it, but changing it to two lines, lmk if it doesn't work

suggestion
```suggestion
     * NOTE: ERC-2981 allows setting the royalty to 100% of the price. In that case, all the price would be sent to the 
     * royalty receiver and 0 tokens to the seller. Contracts dealing with royalty should consider empty transfers.
```

Amxx (approved):

PR #5176, opened 2024-08-29
Make padding skip clear in Base64URL encoding

ernestognw:
Fixes N-03

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Base64.sol
+++ b/contracts/utils/Base64.sol
@@ -22,7 +22,8 @@ library Base64 {
     }
 
     /**
-     * @dev Converts a `bytes` to its Bytes64Url `string` representation. Skips `=` padding.
+     * @dev Converts a `bytes` to its Bytes64Url `string` representation.
+     * Output is not padded with `=` as specified in https://www.rfc-editor.org/rfc/rfc4648[rfc4648].
      */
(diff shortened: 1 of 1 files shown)


contracts/utils/Base64.sol
```diff
@@ -22,7 +22,7 @@ library Base64 {
     }
 
     /**
-     * @dev Converts a `bytes` to its Bytes64Url `string` representation.
+     * @dev Converts a `bytes` to its Bytes64Url `string` representation. Skips `=` padding.
```
Amxx, line 25:
```suggestion
     * @dev Converts a `bytes` to its Bytes64Url `string` representation. Output is not padded with `=` as authorized 
     * by https://www.rfc-editor.org/rfc/rfc4648[rfc4648].
```

ernestognw: changed `authorized` to `specified` since I think its meaning is more accurate in this context

Amxx (approved):

pcaversaccio:
Hmm, I was just reading https://www.rfc-editor.org/rfc/rfc4648#section-4. Standard Base64 includes padding (not strictly enforced tho). I still think OZ should add padding IMO.

PR #5177, opened 2024-08-30
Consistently name multiple returned values

ernestognw:
Fixes N-05

Took the time to look for every case of multiple return values that weren't named. We can reduce the scope to those of the audit if needed.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -250,7 +250,7 @@ library P256 {
         JPoint[16] memory points,
         uint256 u1,
         uint256 u2
-    ) private view returns (uint256 ax, uint256 ay) {
+    ) private view returns (uint256 rx, uint256 ry) {
         uint256 x = 0;
         uint256 y = 0;
         uint256 z = 0;
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -120,7 +120,7 @@ library P256 {
      * IMPORTANT: This function disallows signatures where the `s` value is above `N/2` to prevent malleability.
      * To flip the `s` value, compute `s = N - s` and `v = 1 - v` if (`v = 0 | 1`).
      */
-    function recovery(bytes32 h, uint8 v, bytes32 r, bytes32 s) internal view returns (bytes32, bytes32) {
+    function recovery(bytes32 h, uint8 v, bytes32 r, bytes32 s) internal view returns (bytes32 x, bytes32 y) {
         if (!_isProperSignature(r, s) || v > 1) {
             return (0, 0);
         }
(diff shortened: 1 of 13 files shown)


contracts/utils/cryptography/P256.sol
```diff
@@ -246,7 +246,11 @@ library P256 {
      * the individual points for a single pass are precomputed
      * overall this reduces the number of additions while keeping the same number of doublings
      */
-    function _jMultShamir(JPoint[16] memory points, uint256 u1, uint256 u2) private view returns (uint256, uint256) {
+    function _jMultShamir(
+        JPoint[16] memory points,
+        uint256 u1,
+        uint256 u2
+    ) private view returns (uint256 ax, uint256 ay) {
```
Amxx, line 253: why `ax`, `ay` ? I confused by the `a` part 

- Could be `rx` ,ad `ry` for "result"
- Could be `vx` and `vy` (because v comes after u)

Amxx:
I'm concerned with the performance impact of this change, because yes, there is an impact.

Lets compare
```solidity
function double_v1(uint256 X) public returns (uint256) { 
    return 2 * x;
}
function double_v2(uint256 X) public returns (uint256 result) { 
    return 2 * x;
}
function double_v3(uint256 X) public returns (uint256 result) { 
    result = 2 * x;
}
```

- In the first case, the input is doubled, and put to a temporary (unamed) variable on the stack. This is returned.
- In the second case, the result variable is declared (and reserved) on the stack (that is a push0). Then the input is doubled and put onto a temporary (unamed) variable on the stack, then the temporary variable is moved to the return variable (on the stack) using a `swap` or something similar. Cleaning the stack might also require a pop. In the end, the value is returned.
- In the third case, the result variable is declared (and reserved) on the stack (that is a push0). Then the input is doubled and put into the reserved space. In the end, the value is returned.

I'm not exactly sure what the compiler optimizes, but option 2 is clearly the worst in terms of stack space usage. Naming the returns creates a stack object in between the arguments of the function and the internal variable of the function. This makes stack too deep happen "faster".

  Here is looks like the impact will be negligeable. Probably less then 10/20 gas per function. Possibly 0 with optimisations ?

contracts/utils/cryptography/P256.sol
ernestognw: Named them after "**a**ffine **x**" and "**a**ffine **y**". I'm fine with `rx` and `ry`

ernestognw:
The second case is more expensive, but I'd be fine with the gas increase. Generally, we prefer to follow guidelines over gas savings and here the order of the savings is low. We're not always naming variables, only when there are multiple of them, which doesn't happen often.

Alternative, we can use the 3rd option, but I remember we never liked it for readability. We might agree to do it whenever returned values are named for whatever reason.

Amxx:
I'd be ok with doing option 2 when that doesn't cause stack too deep errors. In cases where options 2 causes stack to deep error, then I would go for either option 3, or with 

```solidity
function double_v2_inline_commented(uint256 X) public returns (uint256 /*result*/) { 
    return 2 * x;
}
```

ernestognw:
In cases where a stack too deep error is caused I'd just go for option 3. We've done that in the past in [AccessManager#schedule](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3547cdce21f0f907b93d1e1ebc2f0945c23bd166/contracts/access/manager/AccessManager.sol#L473)

cairoeth (commented):
should the return values of `getFull` and `_getFullAt` functions in `Time` be named as well?

PR #5178, opened 2024-08-30
Guidelines for named return values

ernestognw:
Makes official our internally agreed policy of naming return values when they're more than one

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -131,6 +131,16 @@ In addition to the official Solidity Style Guide we have a number of other conve
   abstract contract AccessControl is ..., {
   ```
 
+* Return values should be explicitly named if there are more than one.
+
+  ```solidity
+  function expiration() public view returns (uint256) { // Good
+  function hasRole() public view returns (bool isMember, uint32 currentDelay) { // Good
+
+  function expiration() public view returns (uint256 expiration) { // Bad
+  function hasRole() public view returns (bool, uint32) { // Bad
+  ```
+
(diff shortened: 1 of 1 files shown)


GUIDELINES.md
```diff
@@ -131,6 +131,16 @@ In addition to the official Solidity Style Guide we have a number of other conve
   abstract contract AccessControl is ..., {
   ```
 
+* Return values should be explicitly named if there are more than one.
+
+  ```solidity
+  function expiration() public view returns (uint256) { // Good
+  function hasRole() public view returns (bool isMember, uint32 currentDelay) { // Good
+
+  function expiration() public view returns (uint256 expiration) { // Bad
```
Amxx, line 140: I don't think naming a return value here is bad. It avoid having to explicitelly declare it in the body when the value is set in assembly.

For example, I think
```solidity
function getAddressSlot(bytes32 slot) internal pure returns (AddressSlot storage r) {
    assembly ("memory-safe") { r.slot := slot }
}
```
is better than 
```solidity
function getAddressSlot(bytes32 slot) internal pure returns (AddressSlot storage) {
    AddressSlot storage r;
    assembly ("memory-safe") { r.slot := slot }
    return r;
}
```

ernestognw: Good point, I remember we discussed this when we mentioned this guideline.
I'll rewrite it

Amxx (approved):

PR #5182, opened 2024-09-03
Implement ERC-4337 Account and ERC-7579 modules

ernestognw:
Alternative to #5119 

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/account/signers/SignerRSA.sol
```diff
@@ -3,12 +3,14 @@
...
-import {Clones} from "../../proxy/Clones.sol";
 import {EIP712ReadableSigner} from "./EIP712ReadableSigner.sol";
 
 abstract contract SignerRSA is EIP712ReadableSigner {
+    bytes private _e;
+    bytes private _n;
+
     function signer() public view virtual returns (bytes memory e, bytes memory n) {
-        return abi.decode(Clones.fetchCloneArgs(address(this)), (bytes, bytes));
+        return (_e, _n);
```
Amxx: This is going in storage. That is not complying with 4337 storage accesses.

contracts/account/signers/SignerP256.sol
```diff
@@ -3,12 +3,14 @@
...
-import {Clones} from "../../proxy/Clones.sol";
 import {EIP712ReadableSigner} from "./EIP712ReadableSigner.sol";
 
 abstract contract SignerP256 is EIP712ReadableSigner {
+    bytes32 private immutable _qx;
+    bytes32 private immutable _qy;
+
     function signer() public view virtual returns (bytes32 qx, bytes32 qy) {
-        return abi.decode(Clones.fetchCloneArgs(address(this)), (bytes32, bytes32));
+        return (_qx, _qy);
```
Amxx: Any reason for that change ? My understanding is that it requires redeploying an entier instnace for every new account (that include redeploying the P256 library).

IMO clonesWithArgs are way better here.

contracts/account/signers/SignerRSA.sol
ernestognw: Right, even in immutable arguments it accesses storage via EIP712Readable. It also discards upgradeability and most other account providers I've reviewed don't identify this as an issue or just don't care about it.

I asked about this in [Alto (Pimlico)](https://github.com/pimlicolabs/alto/issues/260) and it seems like these transactions will be processed privately but not broadcasted to the decentralized mempool (whatever that is). So far, Accounts can read their own storage or external "associated storage" (i.e. `keccak(account_address||bytes32)+n` where `n` is up to 128) so that enables to validate multiple contract signatures using `IERC1271` through a validator.

With that approach, a MultisigValidator is just a multiplexer of multiple accounts checked with `SignatureChecker.isValidSignatureNow` and is still compatible during the validation phase. However, any upgradeable version of the account or associated modules breaks the rules again (checks non-associated storage when reading the implementation slot).

From what I've seen (e.g. Safe), Multisigs frequently work by collecting the signatures and then packing them into a transaction that someone else's execute, for that reason I think it's fine that multisig-approved calls are not validated during the execution phase, and rather executed by another account or relayer.

Bottom line is: Our default recommendation is an `AccountERC7579` that requires at least 1 validator module. Validator modules shouldn't be upgradeable or they'll break storage access rules although the account can still be executed from another account or through a private bundler (e.g Pimlico whitelists Safe proxy accounts even though they read the master copy from storage).

contracts/account/signers/SignerP256.sol
ernestognw: Yes. I still think we should document how to use signer clones since they are usable for upgradeable accounts where the validation is in its own storage (compliant with the storage validation rules). However, the easiest approach is to use a validator module if it's not upgradeable.

Imo the clones with immutable arguments shouldn't be packaged into the library and rather a section in our documentation where we describe what "immutable signers" are useful for.

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+```solidity
+include::api:example$account/MyAccountECDSAPlainSigner.sol[]
+```
+
+NOTE: Using xref:api:utils.adoc#MessageHashUtils-toEthSignedMessageHash-bytes32-[`MessageHashUtils.toEthSignedMessageHash`] wraps the message with an `\x19Ethereum Signed Message:\n` prefix as most wallets do. However, this is not the most recommended approach since it requires the private key owner to sign an opaque message without contextual information. See xref:account-abstraction.adoc#readable_typed_signatures[readable typed signatures].
+
+==== ERC-7562 Validation Rules
+
+To process a bundle of `UserOperations`, bundlers call xref:api:account.adoc#AccountBase-validateUserOp-struct-PackedUserOperation-bytes32-uint256-[`validateUserOp`] on each operation sender to check whether the operation can be executed. However, the bundler has no guarantee that the state of the blockchain will remain the same after the validation phase. To overcome this problematic, https://eips.ethereum.org/EIPS/eip-7562[ERC-7562] proposes a set of limitations to EVM code so that bundlers (or node operators) are protected from unexpected state changes.
```
frangio, line 94:
```suggestion
To process a bundle of `UserOperations`, bundlers call xref:api:account.adoc#AccountBase-validateUserOp-struct-PackedUserOperation-bytes32-uint256-[`validateUserOp`] on each operation sender to check whether the operation can be executed. However, the bundler has no guarantee that the state of the blockchain will remain the same after the validation phase. To overcome this problem, https://eips.ethereum.org/EIPS/eip-7562[ERC-7562] proposes a set of limitations to EVM code so that bundlers (or node operators) are protected from unexpected state changes.
```

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+=== Signers
+
+An Account controlled by a signer is a type of account that depends on a https://en.wikipedia.org/wiki/Digital_signature[digital signature] verification algorithm and is similar to Externally Owned Accounts (EOAs) in that they require a private key signature to authenticate it. These accounts can be used with ERC-4337 infrastructure normally and do not violate xref:account-abstraction.adoc#erc_7562_validation_rules[ERC-7562 Validation Rules] when used to validate signatures externally in their upgradeable version.
+
+Generally speaking, blockchains use the properties of https://en.wikipedia.org/wiki/Digital_signature[digital signatures] to authenticate the sender of each transaction. With this in mind, developers can xref:utilities.adoc#checking_signatures_on_chain[verify signatures on-chain] as a way to authenticate user operations for a given off-chain signer (i.e. a private key).
```
frangio, line 110: What kinds of accounts are not controlled by signers? I found this section confusing.

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+To setup an account signer, start by inheriting from `AccountSigner` and xref:api:account.adoc#ERC1271TypedSigner[`ERC1271TypedSigner`]:
```
frangio, line 112: I think this is outdated because `AccountSigner` is not present.

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,194 @@
...
+include::api:example$account/MyAccountECDSATypedSigner.sol[]
+```
+
+As you would note, the Account requires to implement both `_validateUserOp` and `_validateSignature`. These are 2 different functions because xref:api:account.adoc#ERC1271TypedSigner[`ERC1271TypedSigner`] adds a wrapper to the `userOpHash` and any other hash received through the `isValidSignature` function the account exposes. The wrapper makes sure that the signature isn't replayable across different account implementations and that the user saw the types associated to the signed message.
+
+IMPORTANT: Accounts must be deployed via an xref:account-abstraction.adoc#account_factory[account factory].
+
+==== Readable Typed Signatures
+
+A common security practice to prevent user operation https://mirror.xyz/curiousapple.eth/pFqAdW2LiJ-6S4sg_u1z08k4vK6BCJ33LcyXpnNb8yU[replayability across smart contract accounts controlled by the same private key] (i.e. multiple accounts for the same signer) is to link the signature to the `address` and `chainId`. This can be done by asking the user to sign the hash of the user operation along with these values.
```
frangio, line 124:
```suggestion
A common security practice to prevent user operation https://mirror.xyz/curiousapple.eth/pFqAdW2LiJ-6S4sg_u1z08k4vK6BCJ33LcyXpnNb8yU[replayability across smart contract accounts controlled by the same private key] (i.e. multiple accounts for the same signer) is to link the signature to the `address` and `chainId` of the account. This can be done by asking the user to sign the hash of the user operation along with these values.
```

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+The following example demonstrates this concept with an xref:api:account.adoc#AccountERC7579[`AccountERC7579`]
+
+```solidity
+include::api:example$account/MyModularAccountClonable.sol[]
+```
+
+TIP: xref:upgradeable.adoc[OpenZeppelin Contracts upgradeable] are initializable out of the box.
+
+NOTE: For upgradeable accounts, you might want to check out the xref:api:proxy.adoc#BeaconProxy[`BeaconProxy`] pattern to have a single implementation contract from which upgradeable clones can be created.
```
frangio, line 176: I believe beacon proxies will run into validation rule violations so it is not a good recommendation.

contracts/account/modules/P256Validator.sol
```diff
@@ -0,0 +1,80 @@
...
+        bytes32 userOpHash
+    ) public view virtual returns (uint256) {
+        return
+            _validateSignature(userOpHash, userOp.signature)
+                ? ERC4337Utils.SIG_VALIDATION_SUCCESS
+                : ERC4337Utils.SIG_VALIDATION_FAILED;
+    }
+
+    /// @inheritdoc IERC7579Validator
+    function isValidSignatureWithSender(
```
frangio, line 35: This should have the mitigations from ERC1271TypedSigner.

On the other hand, ERC1271TypedSigner defines an external `isValidSignature` function that is not needed for a validator module.

contracts/account/AccountERC7579.sol
```diff
@@ -0,0 +1,241 @@
...
+        bytes memory deInitData
+    ) public virtual onlyEntryPointOrSelf {
+        _uninstallModule(moduleTypeId, module, deInitData);
+    }
+
+    function executeUserOp(
+        PackedUserOperation calldata userOp,
+        bytes32 /*userOpHash*/
+    ) public virtual override onlyEntryPointOrSelf {
+        Address.functionDelegateCall(address(this), userOp.callData[4:]);
```
frangio, line 101: This looks weird to me. Is it specified anywhere that `executeUserOp` should behave like this? Can you provide an example?

contracts/account/AccountERC7579.sol
frangio, line 1: ERC-7579 is Draft so I believe related files should be prefixed with `draft-`, right?

contracts/account/AccountERC7579.sol
frangio: I would consider removing this function if it's non-standard.

contracts/account/AccountERC7579.sol
```diff
@@ -0,0 +1,241 @@
...
+    mapping(bytes4 => address) private _fallbacks;
+
+    modifier onlyModule(uint256 moduleTypeId) {
+        _checkModule(moduleTypeId, msg.sender);
+        _;
+    }
+
+    /// @inheritdoc IERC1271
+    function isValidSignature(bytes32 hash, bytes calldata signature) public view virtual override returns (bytes4) {
+        address module = abi.decode(signature[0:20], (address));
```
frangio, line 36: This is not standardized. Do existing modules implement this behavior? What about the off-chain "signer clients"? Do they generate this kind of prefixed signature automatically?

docs/modules/ROOT/pages/account-abstraction.adoc
ernestognw: Rephrased. I'm solving since this is still kinda present but written differently.

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,194 @@
...
+
+The Account Contract is a type of smart contract implements the logic required to validate an `UserOperation` in the context of ERC-4337. Any smart contract account should conform with the `IAccount` interface to validate operations.
+
+```solidity
+interface IAccount {
+    function validateUserOp(PackedUserOperation calldata, bytes32, uint256) external returns (uint256 validationData);
+}
+```
+
+Similarly, an Account should have a way to execute these operations by either handling arbitrary calldata on its `fallback` or implementing the `IAccountExecute` interface:
```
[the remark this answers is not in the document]
ernestognw: Wouldn't be the same as a `fallback` essentially? We may rephrase it and remove the specific examples like `fallback` or `customExecute`

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,194 @@
+= Account Abstraction
+
+Unlike Externally Owned Accounts (EOAs), smart contracts may contain arbitrary verification logic based on authentication mechanisms different to Ethereum's native xref:api:utils.adoc#ECDSA[ECDSA] and have execution advantages such as batching or gas sponsorship. To leverage these properties of smart contracts, the community has widely adopted https://eips.ethereum.org/EIPS/eip-4337[ERC-4337], a standard to process user operations through an alternative mempool.
+
+OpenZeppelin Contracts provides multiple contracts for Account Abstraction following this standard as it enables more flexible and user-friendly interactions with applications. Account Abstraction use cases include wallets in novel contexts (e.g. embedded wallets), more granular configuration of accounts, recovery mechanisms and modularity with standards such as xref:erc7579.adoc#ERC7579[ERC-7579].
```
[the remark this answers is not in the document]
ernestognw: Rephrased

contracts/account/AccountERC7579.sol
ernestognw: True. This is so that there are no other entrypoint in the ERC7579 account other than those defined by the standard (that are also decorated with the `withHook` modifier). I'm not sure what to do now but for now I'd rather keep the delegate call.

ernestognw:
Closing in favor of #5242 #5243 #5244 and upcoming.

PR #5189, opened 2024-09-04
Add `Memory` utility library

ernestognw:
#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/Memory.sol
+++ b/contracts/utils/Memory.sol
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/// @dev Memory utility library.
+library Memory {
+    type Pointer is bytes32;
+
+    /// @dev Returns a memory pointer to the current free memory pointer.
+    function getFreePointer() internal pure returns (Pointer ptr) {
+        assembly ("memory-safe") {
+            ptr := mload(0x40)
+        }
+    }
+
+    /// @dev Sets the free memory pointer to a specific value.
+    ///
+    /// WARNING: Everything after the pointer may be overwritten.
+    function setFreePointer(Pointer ptr) internal pure {
+        assembly ("memory-safe") {
+            mstore(0x40, ptr)
+        }
+    }
+
+    /// @dev Pointer to `bytes32`.
+    function asBytes32(Pointer ptr) internal pure returns (bytes32) {
+        return Pointer.unwrap(ptr);
+    }
+
+    /// @dev `bytes32` to pointer.
+    function asPointer(bytes32 value) internal pure returns (Pointer) {
+        return Pointer.wrap(value);
+    }
+}
--- a/test/utils/Memory.t.sol
+++ b/test/utils/Memory.t.sol
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Test} from "forge-std/Test.sol";
+import {Memory} from "@openzeppelin/contracts/utils/Memory.sol";
+
+contract MemoryTest is Test {
+    using Memory for *;
+
+    function testSymbolicGetSetFreePointer(bytes32 ptr) public {
+        Memory.Pointer memoryPtr = ptr.asPointer();
+        Memory.setFreePointer(memoryPtr);
+        assertEq(Memory.getFreePointer().asBytes32(), memoryPtr.asBytes32());
+    }
+}
--- a/test/utils/Memory.test.js
+++ b/test/utils/Memory.test.js
@@ -1,0 +1,41 @@
+const { ethers } = require('hardhat');
+const { expect } = require('chai');
+const { loadFixture } = require('@nomicfoundation/hardhat-network-helpers');
+
+async function fixture() {
+  const mock = await ethers.deployContract('$Memory');
+
+  return { mock };
+}
+
+describe('Memory', function () {
+  beforeEach(async function () {
+    Object.assign(this, await loadFixture(fixture));
+  });
+
+  describe('free pointer', function () {
+    it('sets memory pointer', async function () {
+      const ptr = '0x00000000000000000000000000000000000000000000000000000000000000a0';
+      expect(await this.mock.$setFreePointer(ptr)).to.not.be.reverted;
+    });
+
+    it('gets memory pointer', async function () {
+      expect(await this.mock.$getFreePointer()).to.equal(
+        // Default pointer
+        '0x0000000000000000000000000000000000000000000000000000000000000080',
+      );
+    });
+
+    it('asBytes32', async function () {
+      const ptr = ethers.toBeHex('0x1234', 32);
+      await this.mock.$setFreePointer(ptr);
+      expect(await this.mock.$asBytes32(ptr)).to.equal(ptr);
+    });
+
+    it('asPointer', async function () {
+      const ptr = ethers.toBeHex('0x1234', 32);
+      await this.mock.$setFreePointer(ptr);
+      expect(await this.mock.$asPointer(ptr)).to.equal(ptr);
+    });
+  });
+});
--- a/.changeset/dull-students-eat.md
+++ b/.changeset/dull-students-eat.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Memory`: Add library with utilities to manipulate memory
--- a/contracts/utils/README.adoc
+++ b/contracts/utils/README.adoc
@@ -40,6 +40,7 @@ Miscellaneous contracts and libraries containing utility functions you can use t
  * {Packing}: A library for packing and unpacking multiple values into bytes32
  * {Panic}: A library to revert with https://docs.soliditylang.org/en/v0.8.20/control-structures.html#panic-via-assert-and-error-via-require[Solidity panic codes].
  * {Comparators}: A library that contains comparator functions to use with with the {Heap} library.
+ * {Memory}: A utility library to manipulate memory.
 
 [NOTE]
 ====


contracts/utils/Memory.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/// @dev Memory utility library.
+library Memory {
+    type Pointer is bytes32;
```
Amxx, line 7: Should we call that type `type FMP is bytes32;` ?

test/utils/Memory.t.sol
```diff
@@ -0,0 +1,16 @@
...
+import {Test} from "forge-std/Test.sol";
+import {Memory} from "@openzeppelin/contracts/utils/Memory.sol";
+
+contract MemoryTest is Test {
+    using Memory for *;
+
+    function testSymbolicGetSetFreePointer(bytes32 ptr) public {
+        Memory.Pointer memoryPtr = ptr.asPointer();
+        Memory.setFreePointer(memoryPtr);
+        assertEq(Memory.getFreePointer().asBytes32(), memoryPtr.asBytes32());
```
Amxx, line 14:
```suggestion
        Memory.setFreePointer(ptr.asPointer());
        assertEq(Memory.getFreePointer().asBytes32(), ptr);
```

Amxx:
IMO this PR should include the usage of this lib

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -386,3 +386,37 @@ await instance.multicall([
...
+  bytes memory callData = abi.encodeWithSelector(
+    bytes4(keccak256("foo()"))
+  )
+  (bool success, /* bytes memory returndata */) = target.call(callData);
+  require(success);
+  Memory.setFreePointer(ptr); // Reset pointer
+}
+----
+
+In this way, new memory will be allocated in the space where the `returndata` and `callData` used to be, potentially reducing memory expansion costs by shrinking the its size at the end of the transaction and resulting in gas savings.
```
Amxx, line 420:
```suggestion
In this way, memory allocated to accomodate the `callData` and `returndata` is freed, potentially reducing memory expansion costs of future memory operations that will be able to re-use that space.
```

Amxx:
Its sad we missed that in 5.3, as it is not controversial. Lets fix conflicts and make sure it gets into 5.4

contracts/utils/Memory.sol
```diff
@@ -0,0 +1,40 @@
...
+
+pragma solidity ^0.8.20;
+
+/// @dev Utilities to manipulate memory.
+///
+/// Memory is a contiguous and dynamic byte array in which Solidity stores non-primitive types.
+/// This library provides functions to manipulate pointers to this dynamic array.
+///
+/// WARNING: When manipulating memory, make sure to follow the Solidity documentation
+/// guidelines for https://docs.soliditylang.org/en/v0.8.20/assembly.html#memory-safety[Memory Safety].
```
arr00, line 11:
```suggestion
/**
 *  @dev Utilities to manipulate memory.
 *
 * Memory is a contiguous and dynamic byte array in which Solidity stores non-primitive types.
 * This library provides functions to manipulate pointers to this dynamic array.
 *
 * WARNING: When manipulating memory, make sure to follow the Solidity documentation
 * guidelines for https://docs.soliditylang.org/en/v0.8.20/assembly.html#memory-safety[Memory Safety].
 */

File: contracts/utils/Memory.sol (hunk 1 above)
Comment (arr00) [reply contracts/utils/Memory.sol:15]: Re amxx here. Why is `getFreePointer` preferred over `getFreeMemoryPointer`?

File: contracts/utils/Memory.sol [hunk 5]
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/**
+ *  @dev Utilities to manipulate memory.
```
arr00, line 6:
```suggestion
 * @dev Utilities to manipulate memory.
```

contracts/utils/Memory.sol
ernestognw: No strong opinion, but I just think is redundant to refer to "Memory" when that's the name of the library

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -84,6 +85,7 @@ abstract contract ERC4626 is ERC20, IERC4626 {
      * @dev Attempts to fetch the asset decimals. A return value of false indicates that the attempt failed in some way.
      */
     function _tryGetAssetDecimals(IERC20 asset_) private view returns (bool ok, uint8 assetDecimals) {
+        Memory.Pointer ptr = Memory.getFreePointer();
```
ernestognw: Should be fixed at the level of the call instead. A compliant ERC is supposed to return a valid uints8 for the decimals

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -31,15 +32,19 @@ library SafeERC20 {
      * non-reverting calls are assumed to be successful.
      */
     function safeTransfer(IERC20 token, address to, uint256 value) internal {
+        Memory.Pointer ptr = Memory.getFreePointer();
```
ernestognw: Should be fixed at the level of `_callOptionalReturn`. Same for all others

contracts/access/manager/AuthorityUtils.sol
```diff
@@ -17,6 +18,7 @@ library AuthorityUtils {
         address target,
         bytes4 selector
     ) internal view returns (bool immediate, uint32 delay) {
+        Memory.Pointer ptr = Memory.getFreePointer();
```
ernestognw: Not needed anymore

contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -50,12 +51,15 @@ library SignatureChecker {
         bytes32 hash,
         bytes memory signature
     ) internal view returns (bool) {
+        Memory.Pointer ptr = Memory.getFreePointer();
```
ernestognw: Should be solved at the call level too

test/utils/Memory.test.js
ernestognw, line 1: Wondering whether keeping the js tests is worth

contracts/utils/Memory.sol
```diff
@@ -0,0 +1,104 @@
...
+    /// @dev Sets the free `Pointer` to a specific value.
+    ///
+    /// WARNING: Everything after the pointer may be overwritten.
+    function setFreePointer(Pointer ptr) internal pure {
+        assembly ("memory-safe") {
+            mstore(0x40, ptr)
+        }
+    }
+
+    /// @dev Returns a `Pointer` to the content of a `bytes` buffer. Skips the length word.
```
Amxx: Everything from here onward (line 33 to 103) was recently added with no discussion about the usecase. I was personally happy with having utils to "cleanup" memory manually, which AFAIK was the original idea behind this PR. 

Some of the things here ressemble `Bytes._unsafeReadBytesOffset, which was left private (and not internal) because we could not guarantee the memory safety. This library goes way furter, with assembly marked memory safe when its not something we know (depends on the user input).

Using the copy function, its quite easy to break stuff by writting to the wrong locations (for example to 0x60).

ernestognw: The background for these changes is that they're used by the RLP library and we previously kept private:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5680/commits/0332ffe69e737a0aaef5d611463a6de0d6306f10#diff-0399261b2cd954212dda0147b86044380b879ab21e1d1b89c564cb48f3552216L304-L328

I agree with your take. However I think it's worth pursuing this pattern since it composes pretty well imo (as long as we add proper WARNINGS and docs). See how it looks in some places in RLP:

```solidity
function readRawBytes(Item memory item) internal pure returns (bytes memory) {
    uint256 itemLength = item.length;
    bytes memory result = new bytes(itemLength);
    result.contentPointer().copy(item.ptr, itemLength);

    return result;
}
```

I would be fine removing these functions. Want to hear your thoughts on the pattern

contracts/utils/Memory.sol
0xClandestine: Would prefer `Memory.Pointer` over an acronym like `FMP`.

ernestognw: > Would prefer Pointer or FreeMemoryPointer

Any preference over these 2? I'd prefer `Pointer` but I understand that `FreeMemoryPointer` could be considered more explicit and less error-prone, in the sense that it's clearer for the developer that the pointer must be "free"

Also, any preference between `getFreePointer` and `getFreeMemoryPointer`? I like the first so that users can do `Memory.getFreePointer()`

0xClandestine: I think `Pointer` is better, you may be handling multiple pointers at one time in a given scope, thus not all are "free memory pointers".

Continuing this line of thought, I'd opt for `Memory.getFreeMemoryPointer`.

ernestognw: Thanks! implemented in: [e67e8b4](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5189/commits/e67e8b4fd020806f3c77ea68dab22020c0459e58)

I'll use you as a deal breaker and solve this convo 😄 . Others can reopen if they strongly disagree

contracts/utils/Memory.sol
0xClandestine: Rather than exposing these in `Memory` when they're only used in the RLP library, maybe make them private within the RLP library? 

Would also consider constraining `destPtr` to be greater than fmp offset.

0xClandestine: @Amxx Worth noting the original proveth implementation does essentially the same, but manually, without `mcopy`.

```solidity
  /*
  * @param src Pointer to source
  * @param dest Pointer to destination
  * @param len Amount of memory to copy from the source
  */
  function copy(uint src, uint dest, uint len) private pure {
      if (len == 0) return;

      // copy as many word sizes as possible
      for (; len >= WORD_SIZE; len -= WORD_SIZE) {
          assembly {
              mstore(dest, mload(src))
          }

          src += WORD_SIZE;
          dest += WORD_SIZE;
      }

      // left over bytes. Mask is used to remove unwanted bytes from the word
      uint mask = 256 ** (WORD_SIZE - len) - 1;
      assembly {
          let srcpart := and(mload(src), not(mask)) // zero out src
          let destpart := and(mload(dest), mask) // retrieve the bytes
          mstore(dest, or(destpart, srcpart))
      }
  }
```

ernestognw: I removed all new controversial functions. Opened https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5792 instead. Also updated the RLP PR (#5680) accordingly as suggested @0xClandestine

contracts/utils/Memory.sol
```diff
@@ -0,0 +1,58 @@
...
+    /// @dev Returns a `Pointer` to the current free `Pointer`.
+    function getFreeMemoryPointer() internal pure returns (Pointer ptr) {
+        assembly ("memory-safe") {
+            ptr := mload(0x40)
+        }
+    }
+
+    /// @dev Sets the free `Pointer` to a specific value.
+    ///
+    /// WARNING: Everything after the pointer may be overwritten.
```
arr00:
```suggestion
    /**
     * @dev Sets the free `Pointer` to a specific value.
     *
     * WARNING: Everything after the pointer may be overwritten.
     **/
```

contracts/utils/Memory.sol
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * WARNING: Everything after the pointer may be overwritten.
+     **/
+    function setFreeMemoryPointer(Pointer ptr) internal pure {
+        assembly ("memory-safe") {
+            mstore(0x40, ptr)
+        }
+    }
+
+    /// @dev `Pointer` to `bytes32`. Expects a pointer to a properly ABI-encoded `bytes` object.
```
arr00, line 35: Pointer is bytes32? What do you mean by abi-encoded `bytes` object?

ernestognw: Even `bytes` memory are encoded with length 0. With this I mean that if you point to wherever you want it may have unexpected results lol

arr00: Ah I was thinking of the pointer value itself. Maybe we can clarify that it should point to properly encoded data?

arr00: i guess rereading it makes sense

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -461,6 +461,38 @@ await instance.multicall([
...
+    bytes memory tempData = abi.encode(items[i], block.timestamp);
+    // Process tempData...
+    Memory.setFreeMemoryPointer(ptr); // Reset pointer for reuse
+  }
+}
+----
+
+This way, memory allocated for `tempData` in each iteration is reused, significantly reducing memory expansion costs when processing many items.
+
+IMPORTANT: Manual memory management increases gas costs and prevents compiler optimizations. Only use these functions after profiling confirms they're necessary. By default, Solidity handles memory safely - using this library without understanding memory layout and safety may be dangerous. See the https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_memory.html[memory layout] and https://docs.soliditylang.org/en/v0.8.20/assembly.html#memory-safety[memory safety] documentation for details.
```
arr00:
```suggestion
IMPORTANT: Only use these functions after carefully confirming they're necessary. By default, Solidity handles memory safely. Using this library without understanding memory layout and safety may be dangerous. See the https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_memory.html[memory layout] and https://docs.soliditylang.org/en/v0.8.20/assembly.html#memory-safety[memory safety] documentation for details.
```

arr00 (dismissed):

Amxx:
Should be covered by #5094

Amxx (approved):

PR #5190, opened 2024-09-04
Refactor Heap.sol to remove `index` and `lookup`

Amxx:
As suggested during the 5.1 audit, the Heap can be refactored to drop the index/lookup part.

- This greatly simplify the code, which should reflect in the gas metrics.
- This makes each node just data (no room for index and lookup values)
  - we can have a heap of uint256 where all nodes/items occupy just one slot. This removes the need for a uint208 heap (which only purpose was to be more stolrage efficient then the uint256 variant)
  
- dropping the procedural generation since we only have one version

This is a breaking change (changing storage layout) that we can only do in 5.1 before Heap is released.
  
#### PR Checklist

- [x] Tests


Amxx:
Capture d’écran du 2024-09-04 21-56-39

contracts/utils/structs/Heap.sol
```diff
@@ -283,294 +227,18 @@ library Heap {
      */
     function _siftUp(
         Uint256Heap storage self,
-        uint64 pos,
+        uint64 index,
```
ernestognw, line 230: We'll need to rename `_siftDown` arguments as well and remove references to `pos`

ernestognw (dismissed):

ernestognw (commented):
Speaking about challenging the design, we might prefer to use an storage array now that the structure is way simpler:

```solidity
library Heap {
   ...

    function peek(uint256[] storage heap) internal view returns (uint256) {
        // heap[0] will `ARRAY_ACCESS_OUT_OF_BOUNDS` panic if heap is empty.
        return heap[0];
    }

    function pop(uint256[] storage heap) internal returns (uint256) {
        return pop(heap, Comparators.lt);
    }
  
   ...
}
```

wdyt @Amxx ?

EDIT: Opened a [PR](https://github.com/Amxx/openzeppelin-contracts/pull/4)

Amxx:
If we use a simple `uint256[]`, then all the function would work on array that are not proper heaps. This would result in unexpected behavior. Also, if someone does `using Heaps for uint256[]`, then `array.pop()` becomes ambigous (override an exissting solidity function).

In order to avoid confusion, I think its better to have a specific type, just like we use `Trace224` instead of `Checkpoint224[]` or we `ShortString` instead of `bytes32` In both cases we have expectation about the format (trace is sorted by key, ShhortString starts by the length).

ernestognw (approved):
LGTM

contracts/utils/structs/Heap.sol
```diff
@@ -181,396 +140,108 @@ library Heap {
...
+        // Check if there is a risk of overflow when commputing the indices of the child nodes. If that is the case,
+        // there cannot be child nodes in the tree, so sifting is done.
+        if (index >= type(uint256).max / 2) return;
+
+        // Compute the indices of the potential child nodes
+        uint256 lIndex = 2 * index + 1;
+        uint256 rIndex = 2 * index + 2;
+
+        // Three cases:
+        // 1. Both childs exist: sifting may continue on one of the branch (selection required)
```
Amxx, line 203:
```suggestion
        // 1. Both children exist: sifting may continue on one of the branch (selection required)
```

contracts/utils/structs/Heap.sol
```diff
@@ -191,9 +191,18 @@ library Heap {
         uint256 value,
         function(uint256, uint256) view returns (bool) comp
     ) private {
-        uint256 lIndex = 2 * index + 1; // this could overflow uint256
-        uint256 rIndex = 2 * index + 2; // this could overflow uint256
-
+        // Check if there is a risk of overflow when commputing the indices of the child nodes. If that is the case,
+        // there cannot be child nodes in the tree, so sifting is done.
+        if (index >= type(uint256).max / 2) return;
```
ernestognw, line 196: What was the rationale for introducing this change? In the past implementation it would've taken really long to overflow

Amxx: Indeed, this change avoids an overflow in cases that will, in practice, never happen.

For an overflow to happen, the heap would have to include 2**255 items, which will, in practice never happen.
Unit testing will never notice the different. Formal verification will.

Would you remove that check with some documentation? Without documentation

ernestognw (dismissed):

ernestognw (approved):
Added a small note about the comparator function, which has been brought up in other places where we use function pointers.

hcheng826:
Is there an example of modifying it to work with not only `uint256` but also custom struct? Is it sensible to do so?

hcheng826:
I think the possible problem is that many operations from `Array.sol` only support `uint256`. A hacky workaround I can think of is to decode the `uint256` into 2 `uint128`s. Maintain a `MyStruct` array separately. And first `uint128` stores the index of the array and the other `uint128` stores the priority. Override the comparator functions to decode and read the 2nd `uint128`. Could that work? Or is it actually the expected way to use this?

PR #5192, opened 2024-09-09
Delegate override vote

Amxx:
Replaces #5160 (was not able to push commit to arr00's repository.

#### To discuss

- Contract naming
  - `VotesOverridable`: this is an extension of `Votes` that provides more features. It can be used to implement overrides, but it may also be used for other stuff. Maybe the name should not be so tightly associated with override.
  - `GovernorOverrideDelegateVote`: this is a counting module. IMO the name should reflect that.
    - `GovernorCountingOverridable` ?
- Errors naming
  - `GovernorAlreadyCastVoteOverride`: I think we can do better.
- Function naming
  - `overrideVote`, `overrideVoteBySig`: should it include "cast"
- Events
  - events to emit when `overrideVotes` is executed.
  - Is it possible to differentiate between a normal vote and an override vote in the `VoteCast` event?
- Extension Compatibility
  - The PreventLateQuorum extension expects that all votes call the internal `_castVote` function. We need to ensure these two extensions are compatible. Some options:
    1. Call `_castVote` from this function. This would necessitate the usage of vote params.
    2. Create a new internal function that is called every time the vote tally is updated--similar to the idea of the `_update` function on the ERC20 token.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/governance/extensions/GovernorOverrideDelegateVote.sol
```diff
@@ -0,0 +1,209 @@
...
+        string calldata reason
+    ) internal virtual returns (uint256) {
+        _validateStateBitmap(proposalId, _encodeStateBitmap(ProposalState.Active));
+
+        uint256 overridenWeight = _countOverride(proposalId, account, support);
+
+        emit VoteCast(account, proposalId, support, overridenWeight, reason);
+
+        return overridenWeight;
+    }
```
arr00, line 166: The implementation does not call the `_castVote` internal function. `GovernorPreventLateQuorum` assumes that all votes call this function https://github.com/OpenZeppelin/openzeppelin-contracts/blob/8b591baef460523e5ca1c53712c464bcc1a1c467/contracts/governance/extensions/GovernorPreventLateQuorum.sol#L52-L72. If we do continue with this implementation, we need to ensure there is documentation reflecting this.

Amxx: IMO it is essential that this extension works with the LateQuorum without any hiccup. We'll have to fix that.

arr00: We've decided to move the `GovernorPreventLateQuorum` to a new internal function called `_tallyUpdated` which will run every time the vote tally is updated.

.changeset/pink-wasps-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`GovernorCountingOverridable`: Add a governor counting module that enables delegatees to override the vote of their delegate.
```
Amxx, line 5: IMO that is easier to understand

```suggestion
`GovernorCountingOverridable`: Add a governor counting module that enables token holders to override the vote of their delegate.
```

contracts/governance/Governor.sol
```diff
@@ -260,6 +260,13 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         bytes memory params
     ) internal virtual returns (uint256);
 
+    /**
+     * @dev Hook that should be called every time the tally for a proposal is updated.
+     *
+     * Note: This function must run successfully--reverts will result in the bricking of governance
```
Amxx, line 266:
```suggestion
     * Note: This function must run successfully. Reverts will result in the bricking of governance
```

PR #5195, opened 2024-09-11
Clean dirty addresses and booleans

cairoeth:
Fixes M-01

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/scripts/generate/helpers/sanitize.js
+++ b/scripts/generate/helpers/sanitize.js
@@ -1,0 +1,5 @@
+module.exports = {
+  address: expr => `and(${expr}, shr(96, not(0)))`,
+  bool: expr => `iszero(iszero(${expr}))`,
+  bytes: (expr, size) => `and(${expr}, shl(${256 - 8 * size}, not(0)))`,
+};
--- a/scripts/generate/templates/SlotDerivation.js
+++ b/scripts/generate/templates/SlotDerivation.js
@@ -1,4 +1,5 @@
 const format = require('../format-lines');
+const sanitize = require('../helpers/sanitize');
 const { TYPES } = require('./Slot.opts');
 
 const header = `\
@@ -77,13 +78,7 @@ const mapping = ({ type }) => `\
  */
 function deriveMapping(bytes32 slot, ${type} key) internal pure returns (bytes32 result) {
     assembly ("memory-safe") {
-        ${
-          type == 'address'
-            ? 'mstore(0x00, shr(96, shl(96, key)))'
-            : type == 'bool'
-            ? 'mstore(0x00, iszero(iszero(key)))'
-            : 'mstore(0x00, key)'
-        }
+        mstore(0x00, ${(sanitize[type] ?? (x => x))('key')})
         mstore(0x20, slot)
         result := keccak256(0x00, 0x40)
     }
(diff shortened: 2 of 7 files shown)


test/utils/SlotDerivation.t.sol
```diff
@@ -10,6 +10,13 @@ import {SlotDerivation} from "@openzeppelin/contracts/utils/SlotDerivation.sol";
 contract SlotDerivationTest is Test, SymTest {
     using SlotDerivation for bytes32;
 
+    function testDirtyAddress() public {
+        bytes32 result1 = SlotDerivation.deriveMapping(0, address(uint160(type(uint256).max)));
```
Amxx, line 14: This does not contain dirty upper bits. The casting to `uint160` cleans it.

scripts/generate/templates/SlotDerivation.js
```diff
@@ -77,7 +77,13 @@ const mapping = ({ type }) => `\
  */
 function deriveMapping(bytes32 slot, ${type} key) internal pure returns (bytes32 result) {
     assembly ("memory-safe") {
-        mstore(0x00, key)
+        ${
+          type == 'address'
+            ? 'mstore(0x00, shr(96, shl(96, key)))'
```
Amxx, line 82: In packing we clean by doing

```suggestion
            ? 'mstore(0x00, and(key, shl(96, not(0))))'
```

I'm thinking we should an helper for cleaning

test/utils/SlotDerivation.t.sol
cairoeth: only when stored in memory afaik -- https://x.com/optimizoor/status/1560590413768994816?s=46. test reverts with current master behaviour.

Amxx: Ok

I changed the tests to something that is clearer IMO, and that support dirtying any type (bool, addres, anything else we want to add)

scripts/generate/helpers/sanitize.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  address: expr => `and(${expr}, shr(96, not(0)))`,
```
ernestognw, line 2: Perhaps it makes sense to write the mask down directly. It's arguably similar in terms of readability and may shave a couple of instructions.

```suggestion
  address: expr => `and(${expr}, 0x000000000000000000000000ffffffffffffffffffffffffffffffffffffffff)`,
```

ernestognw (approved):
Minor nit if you're eager to discuss. LGTM

scripts/generate/helpers/sanitize.js
Amxx: It increasses the code length though. Not sure which one is cheaper.

Amxx: I checked, and after optimisation both version produce the same bytecode.

Amxx: Without optimisations, the `0x000000000000000000000000ffffffffffffffffffffffffffffffffffffffff` is bigger (and more expensive to deploy), but slightly less expensive to run.

ernestognw (approved):

PR #5204, opened 2024-09-16
Clarify mapping named parameters

cairoeth:
Fixes N03.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/access/manager/AccessManager.sol
+++ b/contracts/access/manager/AccessManager.sol
@@ -109,8 +109,8 @@ contract AccessManager is Context, Multicall, IAccessManager {
     uint64 public constant PUBLIC_ROLE = type(uint64).max; // 2**64-1
 
     mapping(address target => TargetConfig mode) private _targets;
-    mapping(uint64 roleId => Role) private _roles;
-    mapping(bytes32 operationId => Schedule) private _schedules;
+    mapping(uint64 roleId => Role role) private _roles;
+    mapping(bytes32 operationId => Schedule schedule) private _schedules;
 
     // Used to identify operations that are currently being executed via {execute}.
     // This should be transient storage when supported by the EVM.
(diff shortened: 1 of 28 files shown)


contracts/access/manager/AccessManager.sol
```diff
@@ -109,8 +109,8 @@ contract AccessManager is Context, Multicall, IAccessManager {
     uint64 public constant PUBLIC_ROLE = type(uint64).max; // 2**64-1
 
     mapping(address target => TargetConfig mode) private _targets;
-    mapping(uint64 roleId => Role) private _roles;
-    mapping(bytes32 operationId => Schedule) private _schedules;
+    mapping(uint64 roleId => Role role) private _roles;
+    mapping(bytes32 operationId => Schedule schedule) private _schedules;
```
Amxx, line 113: This is tipically the kind of cases were we did not name the mapping parameters because the naming makes the code longer without adding any information.

variable is called `_roles`, so yes elements in it are roles. I think our guidelines was to name/document the keys (that are not obvious) and rely on the mapping name to describe the values of the mapping.

Amxx: @frangio WDYT ?

frangio: Yes for me the name seems redundant here.

ernestognw: Indeed, we decided not to name the mapping values since they don't provide relevant information, usually the name of the map itself already expresses what the value is (e.g. returning a `schedule` in `_schedules`).

Particularly, the issue reported (N-03) only specifies a concrete case:

> Within GovernorCountingFractional.sol, the [_proposalVotes state variable](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/aba9ff61ac6d610eab12b1528c6f22f98dbb4dd3/contracts/governance/extensions/GovernorCountingFractional.sol#L47) can benefit from a named parameter.

For that one, I'd agree on naming the parameter. For the rest I think we can leave without named values

ernestognw (approved):

Amxx (approved):

PR #5206, opened 2024-09-16
Update and clarify documentation comments

cairoeth:
Fixes N-06, N-08, N-10 and N-13

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/structs/MerkleTree.sol
```diff
@@ -22,8 +22,6 @@ import {Panic} from "../Panic.sol";
  * NOTE: Building trees using non-commutative hashing functions (i.e. `H(a, b) != H(b, a)`) is supported. However,
  * proving the inclusion of a leaf in such trees is not possible with the {MerkleProof} library since it only supports
  * _commutative_ hashing functions.
- *
- * _Available since v5.1._
```
cairoeth, line 26: i believe this is not necessary as the versioning will be under license (line 2)

contracts/utils/cryptography/P256.sol
```diff
@@ -239,7 +239,7 @@ library P256 {
     }
 
     /**
-     * @dev Compute P·u1 + Q·u2 using the precomputed points for P and Q (see {_preComputeJacobianPoints}).
+     * @dev Compute G·u1 + Q·u2 using the precomputed points for G and Q (see {_preComputeJacobianPoints}).
```
Amxx, line 242: `_preComputeJacobianPoints` is expressed in terms of `p` / `px` / `py` (the point provided as parameter) and `g` / `GX` / `GY`, the generator for the curve.

After checking, it looks like this function returns `G·u1 + P·u2`, with G the generator and P the point used to build the table.

Amxx:
```suggestion
     * @dev Compute G·u1 + P·u2 using the precomputed points for G and P (see {_preComputeJacobianPoints}).
```

contracts/utils/structs/MerkleTree.sol
Amxx: It used to be done manually. The release process did not take care of this when this file was first worked on. 

The release process tells you when a file was last edited, but not when it was first available. If we optimize this in 5.2 or 5.3, the versioning line will stop providing information about the minimal version necessary to get this file.

Removing this line removes information. @ernestognw wdyt ?

cairoeth: agreed, i think it provides valuable information (when it was introduced). in that case, we should also add it for heap and the other new ones for consistency

Amxx: maybe. 

@ernestognw WDYT ?

cairoeth: conclusion: generalize this type introduction date note.

contracts/utils/cryptography/P256.sol
```diff
@@ -299,10 +299,10 @@ library P256 {
         points[0x09] = _jAddPoint(points[0x01], points[0x08]); // 1,2 (p+2g)
         points[0x0a] = _jAddPoint(points[0x02], points[0x08]); // 2,2 (2p+2g)
         points[0x0b] = _jAddPoint(points[0x03], points[0x08]); // 3,2 (3p+2g)
-        points[0x0c] = _jAddPoint(points[0x04], points[0x08]); // 0,3 (g+2g)
+        points[0x0c] = _jAddPoint(points[0x04], points[0x08]); // 0,3 (3g)
```
ernestognw, line 302: This comment attempted to signal that the result of the `0,3` cell is that of `points[0x04]` (g) + `points[0x00]` (2g) so I'm afraid that by rewriting it to `3g` it loses its purpose. No strong opinion but I'd rather do:

```suggestion
        points[0x0c] = _jAddPoint(points[0x04], points[0x08]); // 0,3 (g+2g = 3g)
```

contracts/utils/cryptography/RSA.sol
```diff
@@ -30,7 +30,7 @@ library RSA {
      * method described in https://datatracker.ietf.org/doc/html/rfc8017#section-8.2.2[section 8.2.2 of RFC8017].
      *
      * IMPORTANT: Although this function allows for it, using n of length 1024 bits is considered unsafe.
-     * Consider using at least 2048 bits.
+     * Consider using at least 2048 bits. Additionally, this function only supports SHA256 as the hash function.
```
ernestognw, line 33: What do you think of removing this note and rename the function to `pkcs1Sha256` just as its bytes counterpart. That was the original naming iiurc but I proposed renaming it while ignoring the AlgorithmIdentifier is included during signature creation

ernestognw:
According to N-10, we should enforce 2048 bits in RSA. I updated accordingly since I do think we should enforce it.

ernestognw (dismissed):

contracts/utils/cryptography/P256.sol
Amxx: Note: whatever we decide, this line should be consistent with line 295

test/utils/cryptography/SigVer15_186-3.rsp
Amxx, line 1: Changing this file is not ok. The entier point is to take an unmodified reference (standard) and using it for tests. If any entry in this file causes an issue, the parsing/filtering should take care of that in a predictable way, but this should not change.

contracts/utils/cryptography/P256.sol
cairoeth: updated both in lines https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5206/commits/6eb8c245fea5de371a01edc881d55f3c626adc8e

test/utils/cryptography/RSA.test.js
```diff
@@ -44,12 +45,12 @@ describe('RSA', function () {
...
       const exp =
         '0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010001';
       const mod =
-        '0xdf3edde009b96bc5b03b48bd73fe70a3ad20eaf624d0dc1ba121a45cc739893741b7cf82acf1c91573ec8266538997c6699760148de57e54983191eca0176f518e547b85fe0bb7d9e150df19eee734cf5338219c7f8f7b13b39f5384179f62c135e544cb70be7505751f34568e06981095aeec4f3a887639718a3e11d48c240d';
-      expect(await this.mock.$pkcs1Sha256(data, sig, exp, mod)).to.be.true;
+        '0x9d107f072b168812dad11e17258c109ecad9c336d01c7968d030d7f7d1410fa04d5e346d26067de03e55926863b2e80b71d9b0077ff000e9caaccb0c4f5affb6f0b87b532879c986779b2846718b003d1dc8142a625ec858a06675ca8be8c25e4dbba675b8597306742ac68139a79265f943b2fcc0337003b7c7932155d6c757e9c55ce832b33dbfdc3f1c3bf86f4364833248d1d4c7e560a8244d47155d2ca7c2443b3dc381ff3a22d8fe6c461afc0cd7fac8df7c57048ef296f5c12575258f489a68d560a8554524d07549aa2b7328be6a21b609a21bfaedcf3e6647ed5be1489aebf03b07643c7d6fc63ae20ee4da4f1b6b04a376828e0a83c66d11dd8bf3';
```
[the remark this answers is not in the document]
Amxx: Also, If the "old" openssl test is no longer correct, the `returns false if s >= n` must be updated accordingly

Amxx (approved):
I added a comment explaining how rsa key are generated using openssl (in the openssl test).

LGTM

PR #5211, opened 2024-09-18
Start Vesting right after cliff ends in VestingWalletCliff

ernestognw:
Fixes L-02

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/finance/VestingWalletCliff.sol
+++ b/contracts/finance/VestingWalletCliff.sol
@@ -46,6 +46,6 @@ abstract contract VestingWalletCliff is VestingWallet {
         uint256 totalAllocation,
         uint64 timestamp
     ) internal view virtual override returns (uint256) {
-        return timestamp < cliff() ? 0 : super._vestingSchedule(totalAllocation, timestamp);
+        return timestamp <= cliff() ? 0 : super._vestingSchedule(totalAllocation, timestamp);
     }
 }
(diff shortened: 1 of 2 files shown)


contracts/finance/VestingWalletCliff.sol
```diff
@@ -46,6 +46,6 @@ abstract contract VestingWalletCliff is VestingWallet {
         uint256 totalAllocation,
         uint64 timestamp
     ) internal view virtual override returns (uint256) {
-        return timestamp < cliff() ? 0 : super._vestingSchedule(totalAllocation, timestamp);
+        return timestamp <= cliff() ? 0 : super._vestingSchedule(totalAllocation, timestamp);
```
cairoeth, line 49: nit: we could save an opcode 
```suggestion
        return cliff() > timestamp ? 0 : super._vestingSchedule(totalAllocation, timestamp);
```

cairoeth (approved):

contracts/finance/VestingWalletCliff.sol
Amxx, line 49: I don't think this change is a big deal, but I also think its somehow wrong.

IMO the cliff is the moment where things start to unlock. which means you return zero until `cliff - 1`.
Anyway, in both cases, its just a mater of waiting one more second, so I don't see how this is even an issue.

PR #5218, opened 2024-09-20
Fix P256 corner cases

cairoeth:
Fixes CR-01 and L-05.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -297,11 +297,11 @@ library P256 {
                     (x, y, z) = _jDouble(x, y, z);
                     (x, y, z) = _jDouble(x, y, z);
                 }
-                // Read 2 bits of u1, and 2 bits of u2. Combining the two give a lookup index in the table.
+                // Read 2 bits of u1, and 2 bits of u2. Combining the two gives the lookup index in the table.
                 uint256 pos = ((u1 >> 252) & 0xc) | ((u2 >> 254) & 0x3);
                 // Points that have z = 0 are points at infinity. They are the additive 0 of the group
                 // - if the lookup point is a 0, we can skip it
-                // - otherwize
+                // - otherwise:
                 //   - if the current point (x, y, z) is 0, we use the lookup point as our new value (0+P=P)
                 //   - if the current point (x, y, z) is not 0, both points are valid and we can use `_jAdd`
                 if (points[pos].z != 0) {
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -299,6 +299,11 @@ library P256 {
                 }
                 // Read 2 bits of u1, and 2 bits of u2. Combining the two give a lookup index in the table.
                 uint256 pos = ((u1 >> 252) & 0xc) | ((u2 >> 254) & 0x3);
+                // Points that have z = 0 are points at infinity. They are the additive 0 of the group
+                // - if the lookup point is a 0, we can skip it
+                // - otherwize
+                //   - if the current point (x, y, z) is 0, we use the lookup point as our new value (0+P=P)
+                //   - if the current point (x, y, z) is not 0, both points are valid and we can use `_jAdd`
                 if (points[pos].z != 0) {
                     if (z == 0) {
                         (x, y, z) = (points[pos].x, points[pos].y, points[pos].z);
@@ -326,10 +331,10 @@ library P256 {
      * │ 12 │ 3g 3g+p 3g+2p 3g+3p │
      * └────┴─────────────────────┘
      *
-     * Note that `_jAdd` (and thus `_jAddPoint`) does handle the case where one of the inputs is a point at infinity
-     * (z = 0). However, we know that since `N ≡ 1 mod 2` and `N ≡ 1 mod 3`, there is no point P such that 2P = 0 or
-     * 3P = 0. This garantees that g, 2g, 3g, p, 2p, 3p are all non-zero, and that all `_jAddPoint` calls have valid
-     * inputs.
+     * Note that `_jAdd` (and thus `_jAddPoint`) does not handle the case where one of the inputs is a point at
+     * infinity (z = 0). However, we know that since `N ≡ 1 mod 2` and `N ≡ 1 mod 3`, there is no point P such that
+     * 2P = 0 or 3P = 0. This guarantees that g, 2g, 3g, p, 2p, 3p are all non-zero, and that all `_jAddPoint` calls
+     * have valid inputs.
      */
     function _preComputeJacobianPoints(uint256 px, uint256 py) private pure returns (JPoint[16] memory points) {
         points[0x00] = JPoint(0, 0, 0); // 0,0
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -184,11 +184,12 @@ library P256 {
      * @dev Point addition on the jacobian coordinates
      * Reference: https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian.html#addition-add-1998-cmo-2
      *
-     * Note that `addition-add-1998-cmo-2` doesn't support identical inputs points. This version is modified to use
-     * the `h` and `r` values computed by `addition-add-1998-cmo-2` to detect identical inputs, and fallback
-     * `doubling-dbl-1998-cmo-2` if needed.
+     * Note that:
      *
-     * Note if one of the point is at infinity (z=0), result is undefined.
+     * - `addition-add-1998-cmo-2` doesn't support identical input points. This version is modified to use
+     * the `h` and `r` values computed by `addition-add-1998-cmo-2` to detect identical inputs, and fallback to
+     * `doubling-dbl-1998-cmo-2` if needed.
+     * - if one of the points is at infinity (i.e. `z=0`), the result is undefined.
      */
     function _jAdd(
         JPoint memory p1,
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -187,6 +187,8 @@ library P256 {
      * Note that `addition-add-1998-cmo-2` doesn't support identical inputs points. This version is modified to use
      * the `h` and `r` values computed by `addition-add-1998-cmo-2` to detect identical inputs, and fallback
      * `doubling-dbl-1998-cmo-2` if needed.
+     *
+     * Note if one of the point is at infinity (z=0), result is undefined.
      */
     function _jAdd(
         JPoint memory p1,
@@ -205,7 +207,7 @@ library P256 {
 
             // detect edge cases where inputs are identical
             switch and(iszero(r), iszero(h))
-            // case 1: points are different
+            // case 0: points are different
             case 0 {
                 let hh := mulmod(h, h, p) // h²
 
@@ -279,11 +281,6 @@ library P256 {
      * We optimize this for 2 bits at a time rather than a single bit. The individual points for a single pass are
      * precomputed. Overall this reduces the number of additions while keeping the same number of
      * doublings
-     *
-     * Note that some particular values, such as P = -G (private key = N-1) will result in invalid point computation.
-     * This would result in the inability to verify or recover valid signatures. This edge case could potentially
-     * cause a valid signature to fail verification. While specific private keys are known to trigger that issue, it
-     * is almost impossible to happen with randomly generated keys.
      */
     function _jMultShamir(
         JPoint[16] memory points,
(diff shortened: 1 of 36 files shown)


contracts/utils/cryptography/P256.sol
```diff
@@ -183,6 +183,9 @@ library P256 {
     /**
      * @dev Point addition on the jacobian coordinates
      * Reference: https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian.html#addition-add-1998-cmo-2
+     *
+     * Note that if both input points are identical, this function will return `(0, 0, 0)` instead of doubling the point.
+     * The probability of this happening naturally is negligible (approximately 2048 / (2^256 - 1), or about 10^-74).
```
Amxx, line 188: If woth input are identical, the probability of this happening is 100%.

What that probability here ? If the the probability of two random points being identical ? Is it something else ?

```suggestion
     * The probability of this happening naturally is negligible (approximately 2048 / (2^256 - 1), or about 10^-74).
```

contracts/utils/cryptography/P256.sol
```diff
@@ -183,6 +183,9 @@ library P256 {
     /**
      * @dev Point addition on the jacobian coordinates
      * Reference: https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian.html#addition-add-1998-cmo-2
+     *
+     * Note that if both input points are identical, this function will return `(0, 0, 0)` instead of doubling the point.
+     * The probability of identical input points happening naturally is negligible (approximately 2048 / (2^256 - 1), or about 10^-74).
```
Amxx, line 188: I'm not sure what is "natural" here

```suggestion
     * The probability of two random points being identical is negligible (approximately 2048 / (2^256 - 1), or about 10^-74).
```

Amxx: Also, how was that computed ? 

At first glace, it feals like since field of size P, so that probability of two values being identical is 1/P = 8.6e-78

cairoeth: is not only 1 key out of P, as there are a few possible cases, from audit info:

there's two scenarios that lead to `_jAdd`:
1. `_preComputeJacobianPoints` calls `_jAdd`
2. `_jMultShamir` calls `_jAdd`

For the first scenario, there are 7 distinct valid weak private keys (`d`) that would trigger the bug: `d`∈{1,2,3,2<sup>−1</sup>,3<sup>−1</sup>,2 3<sup>−1</sup>,3 2<sup>−1</sup>}. Assuming that `d` is chosen uniformly at random among n<sup>−1</sup> values, where `n` is a 256<sup>256</sup> bit prime, the probability to choose any one of the weak keys is: 7/(2<sup>256</sup>-1).

For the second scenario, at most 2048 valid weak private keys would trigger the bug. Specifically, the probability of randomly drawing one of the weak keys is at most: 2048/(2<sup>256</sup>−1). Used this probability in the comment as it's more likely.

contracts/utils/cryptography/P256.sol
```diff
@@ -285,6 +288,9 @@ library P256 {
      * │  8 │ 2g 2g+p 2g+2p 2g+3p │
      * │ 12 │ 3g 3g+p 3g+2p 3g+3p │
      * └────┴─────────────────────┘
+     *
+     * Note that in the negligible scenario where `P` is `-G`, the addition of `points[0x05]` (`G` + `P`) will result
```
Amxx, line 292: As discussed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5218/files#r1771415748, there are other cases then the obvious `P = G` that would cause this to fail.

Amxx:
```suggestion
     * Note that because of limitation to the `_jAdd` function (which returns invalid results when adding a point
     * with itself), this function could produce invalid values. This would result in the inability to verify or recover
     * valid signatures. It is estimated to have a probability of about 1e-74 to happen.
```

test/utils/cryptography/P256.t.sol
```diff
@@ -31,6 +31,22 @@ contract P256Test is Test {
...
+        //
+        // In particular, the following key pairs are not supported:
+        // * private = 1 (P = G)
+        // * private = 2 (P = 2G)
+        // * private = 3 (P = 3G)
+        // * private = N - 3 (P = -3G)
+        // * private = N - 2 (P = -2G)
+        // * private = N - 1 (P = -G)
+        //
+        // Ideally we could just do `return bound(seed, 1, P256.N - 1);`. Somehow, the hash procudes "good" values.
```
Amxx, line 46:
```suggestion
        // Ideally we could just do `return bound(seed, 1, P256.N - 1);`. Somehow, the hash produces "good" values.
```

contracts/utils/cryptography/P256.sol
```diff
@@ -242,10 +274,16 @@ library P256 {
...
+     * https://stackoverflow.com/questions/50993471/ec-scalar-multiplication-with-strauss-shamir-method
+     *
+     * We optimise on this a bit to do with 2 bits at a time rather than a single bit. The individual points for a
+     * single pass are precomputed. Overall this reduces the number of additions while keeping the same number of
+     * doublings
+     *
+     * Note that some particular values, such as P = -G (private key = N-1) will result in invalid point computation.
+     * This would result in the inability to verify or recover valid signatures. This edge case could potentially
+     * cause a valid signature to fail verification. While specific private keys are known to trigger that issue, is
+     * it extremelly unlikelly to happen with randomly generated keys.
```
Amxx, line 286:
```suggestion
     * it extremely unlikelly to happen with randomly generated keys.
```

contracts/utils/cryptography/P256.sol
```diff
@@ -242,10 +274,16 @@ library P256 {
...
+     * https://stackoverflow.com/questions/50993471/ec-scalar-multiplication-with-strauss-shamir-method
+     *
+     * We optimise on this a bit to do with 2 bits at a time rather than a single bit. The individual points for a
+     * single pass are precomputed. Overall this reduces the number of additions while keeping the same number of
+     * doublings
+     *
+     * Note that some particular values, such as P = -G (private key = N-1) will result in invalid point computation.
+     * This would result in the inability to verify or recover valid signatures. This edge case could potentially
+     * cause a valid signature to fail verification. While specific private keys are known to trigger that issue, is
+     * it extremely unlikelly to happen with randomly generated keys.
```
ernestognw, line 286:
```suggestion
     * https://stackoverflow.com/questions/50993471/ec-scalar-multiplication-with-strauss-shamir-method
     *
     * We optimize this for 2 bits at a time rather than a single bit. The individual points for a single pass are 
     * precomputed. Overall this reduces the number of additions while keeping the same number of
     * doublings
     *
     * Note that some particular values, such as P = -G (private key = N-1) will result in invalid point computation.
     * This would result in the inability to verify or recover valid signatures. This edge case could potentially
     * cause a valid signature to fail verification. While specific private keys are known to trigger that issue, it
     * is almost impossible to happen with randomly generated keys.
```

test/utils/cryptography/P256.t.sol
```diff
@@ -31,6 +31,19 @@ contract P256Test is Test {
         assertTrue((qx0 == x && qy0 == y) || (qx1 == x && qy1 == y));
     }
 
+    function _asPrivateKey(uint256 seed) private pure returns (uint256) {
+        // In theory, all private keys between 1 and P256.N-1 should be supported. However, computation artefacts
+        // caused by the `_preComputeJacobianPoints` precomputation prevent a limited number of private keys from
+        // being used. While these private keys can be constructed on purpose, secure keys that are randomly
+        // generated should not suffer from these limitations
```
ernestognw, line 38:
```suggestion
        // In theory, all private keys between 1 and P256.N-1 should be supported. However, computation artifacts
        // caused by the `_preComputeJacobianPoints` precomputation prevents a limited number of private keys from
        // being used. While these private keys can be constructed on purpose, secure keys that are randomly
        // generated should not suffer from these limitations
```

test/utils/cryptography/P256.test.js
```diff
@@ -120,6 +121,28 @@ describe('P256', function () {
     });
   });
 
+  describe('test special cases', function () {
+    const { n: N } = secp256r1.CURVE;
+
+    // In theory, all private keys between 1 and P256.N-1 should be supported. However, the fast computation method
+    // introduces artefacts that can cause invalid computation for some particular keys
```
ernestognw, line 128:
```suggestion
    // In theory, all private keys between 1 and P256.N-1 should be supported. However, the fast computation method
    // introduces artifacts that can cause invalid computation for some particular keys
```

ernestognw (approved):

contracts/utils/cryptography/P256.sol
```diff
@@ -195,25 +199,53 @@ library P256 {
...
-            let r := addmod(mulmod(y2, mulmod(zz1, z1, p), p), sub(p, s1), p) // r = s2-s1 = y2*z1³-s1
+            let r := addmod(mulmod(y2, mulmod(zz1, z1, p), p), sub(p, s1), p) // r = s2-s1 = y2*z1³-s1 = y2*z1³-y1*z2³
             let u1 := mulmod(mload(p1), mulmod(z2, z2, p), p) // u1 = x1*z2²
-            let h := addmod(mulmod(x2, zz1, p), sub(p, u1), p) // h = u2-u1 = x2*z1²-u1
-            let hh := mulmod(h, h, p) // h²
+            let h := addmod(mulmod(x2, zz1, p), sub(p, u1), p) // h = u2-u1 = x2*z1²-u1 = x2*z1²-x1*z2²
+
+            // detect edge cases where inputs are identical
+            switch and(iszero(r), iszero(h))
+            // case 1: points are different
```
cairoeth, line 208:
```suggestion
            // case 0: points are different
```

cairoeth: or maybe this is case 1 and below is case 2

ernestognw: I'd agree with this being `0`

ernestognw (approved):
Turns out the `configureYulOptimizer` and `solcOptimizerDetails` options are deprecated in solcover 0.8.5. I tried updating to the latest version and use `viaIR` in coverage. However, it raises some issues that require further investigation and we may need access to intermediate configuration.

LGTM, we can discuss upgrading `solidity-coverage` later

contracts/utils/cryptography/P256.sol
```diff
@@ -183,6 +183,12 @@ library P256 {
     /**
      * @dev Point addition on the jacobian coordinates
      * Reference: https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian.html#addition-add-1998-cmo-2
+     *
+     * Note that `addition-add-1998-cmo-2` doesn't support identical inputs points. This version is modified to use
+     * the `h` and `r` values computed by `addition-add-1998-cmo-2` to detect identical inputs, and fallback
+     * `doubling-dbl-1998-cmo-2` if needed.
+     *
+     * Note if one of the point is at infinity (z=0), result is undefined.
```
ernestognw, line 191:
```suggestion
     * Note that:
     *
     * - `addition-add-1998-cmo-2` doesn't support identical input points. This version is modified to use
     * the `h` and `r` values computed by `addition-add-1998-cmo-2` to detect identical inputs, and fallback
     * `doubling-dbl-1998-cmo-2` if needed.
     * - if one of the points is at infinity (i.e. `z=0`), the result is undefined.
```

ernestognw (approved):

contracts/utils/cryptography/P256.sol
```diff
@@ -289,6 +325,11 @@ library P256 {
      * │  8 │ 2g 2g+p 2g+2p 2g+3p │
      * │ 12 │ 3g 3g+p 3g+2p 3g+3p │
      * └────┴─────────────────────┘
+     *
+     * Note that `_jAdd`, and thus `_jAddPoint`, does handle the case where one of the input is a point at infinity
+     * (z = 0). However, we know that since `N ≡ 1 mod 3` and `N ≡ 1 mod 3`, we know that there is no point P such
+     * that 2P = 0 or 3P = 0. This garantees that all inputs of the addition done during this function are well
```
Amxx, line 331:
```suggestion
     * that 2P = 0 or 3P = 0. This guarantees that all inputs of the addition done during this function are well
```

contracts/utils/cryptography/P256.sol
```diff
@@ -263,7 +299,12 @@ library P256 {
                 }
                 // Read 2 bits of u1, and 2 bits of u2. Combining the two give a lookup index in the table.
```
cairoeth, line 300:
```suggestion
                // Read 2 bits of u1, and 2 bits of u2. Combining the two gives the lookup index in the table.
```

contracts/utils/cryptography/P256.sol
```diff
@@ -263,7 +299,12 @@ library P256 {
...
                 uint256 pos = ((u1 >> 252) & 0xc) | ((u2 >> 254) & 0x3);
-                if (pos > 0) {
+                // Points that have z = 0 are points at infinity. They are the additive 0 of the group
+                // - if the lookup point is a 0, we can skip it
+                // - otherwize
```
cairoeth, line 304:
```suggestion
                // - otherwise:
```

ernestognw (approved):

PR #5233, opened 2024-09-30
Simplify content in security documentation

Shaykoo:
While reading the documentation the text was not clear and made me confuse. hence, the solution.

Screenshot 2024-09-30 at 10 53 19 PM
Screenshot 2024-09-30 at 10 53 32 PM


Amxx:
Note: changing the packages while doing documentation changes is a big NO to me. It may not be one, but it looks like an "attack" that would hide harmfull changes to the dependencies

Shaykoo:
> Note: changing the packages while doing documentation changes is a big NO to me. It may not be one, but it looks like an "attack" that would hide harmfull changes to the dependencies

Thanks @Amxx I will revert and commit again.

Amxx (dismissed):

cairoeth (dismissed):

docs/modules/ROOT/pages/extending-contracts.adoc
```diff
@@ -46,6 +46,6 @@ NOTE: The same rule is implemented and extended in xref:api:access.adoc#AccessCo
...
 
-Custom overrides, and those of hooks in particular, may break some important assumptions and introduce vulnerabilities in otherwise secure code. While we try to ensure the contracts remain secure in the face of a wide range of potential customizations, this is done in a best-effort manner. While we try to document all important assumptions, this should not be relied upon. Custom overrides should be carefully reviewed and checked against the source code of the contract they are customizing so as to fully understand their impact and guarantee their security.
+Custom overrides, especially to hooks, can disrupt important assumptions and may introduce security risks in the code that was previously secure. While we try to ensure the contracts remain secure in the face of a wide range of potential customizations, this is done in a best-effort manner. While we try to document all important assumptions, this should not be relied upon. Custom overrides should be carefully reviewed and checked against the source code of the contract they are customizing so as to fully understand their impact and guarantee their security.
```
cairoeth, line 49:
```suggestion
Custom overrides, especially to hooks, can disrupt important assumptions and may introduce security risks in the code that was previously secure. While we try to ensure the contracts remain secure in the face of a wide range of potential customizations, this is done in a best-effort manner. While we try to document all important assumptions, this should not be relied upon. Custom overrides should be carefully reviewed and checked against the source code of the contract they are customizing to fully understand their impact and guarantee their security.
```

cairoeth (approved):

PR #5236, opened 2024-10-03
Optimize `log2` with a lookup table

Lohann:
# 23% more efficient Log2 method

Follow an efficient implementation of `floor(log2(x))` which uses ~23% less gas than the current implementation.

`log2` is quite central for many other math operations, such as:
- `sqrt` uses log2 internally (even so the current `sqrt` uses an specialized log2 implementation).
- Count leading zeros: `255 - log2(value) + toUint(value == 0)`
- Estimate how many bytes or bits are necessary to represent a given number
- I use it a lot to perform [Float-Point arithmetic in this library](https://github.com/Analog-Labs/analog-gmp/blob/4abad64209e29b57fd3f29a3b066702b30cbafd0/src/utils/Float9x56.sol#L355-L356).

This implementation uses **lookup tables** to calculate `floor(log2(x))` using `218 gas`,  which  is `65 gas` units less than the current implementation.
Obs: the code in this PR have a more detailed documentation on how the algorithm works:
```solidity
function log2(uint256 x) internal pure returns (uint256) {
    unchecked {
        // Round `x` down to the closest power of 2 using the Seander's algorithm.
        // Reference: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
        x |= x >> 1;
        x |= x >> 2;
        x |= x >> 4;
        x |= x >> 8;
        x |= x >> 16;
        x |= x >> 32;
        x |= x >> 64;
        x |= x >> 128;
        // Note `x = 0` results in 1 here, but that's ok once floor(log2(0)) == floor(log2(1)) anyway.
        x = (x >> 1) + 1;

        uint256 prod0;
        {
            // Compute `n = x mod 255`, the resulting `n` can only be one of the following 
            // values: 1, 2, 4, 8, 16, 32, 64 or 128.
            uint256 n = x % 255;

            // compute `n % 11`, this maps all `n` values to an unique index between 0~31.
            uint256 index = n % 11;

            // Perform a table lookup, the table stores the result of `log2(n)` in
            // the corresponding byte index.
            assembly {
                prod0 := byte(index, 0x0000010002040007030605000000000000000000000000000000000000000000)
            }
        }

        // Compute `prod1 = log2(x / n)`, once `prod1` is always multiple of 8, we can
        // use an interesting lookup technique:  (x / n * table) >> 248 == log(x / n)
        uint256 prod1 = x >> prod0;
        prod1 *= 0x0008101820283038404850586068707880889098a0a8b0b8c0c8d0d8e0e8f0f8;
        prod1 >>= 248;

        // log2(x) = log2(x / n) + log2(n) = prod0 + prod1
        return prod0 + prod1;
    }
}
```

Lookup tables are quite powerful and should be explored more in Solidity, I use it in another project for validate 6 different method `selectors` in constant gas, using less than `100 gas` units, way smaller and more efficient than the **nested ifs** or **binary search** generated by default by the Solidity Compiler:
https://github.com/Lohann/universal-factory/blob/e6e3e6acc3f3b8a9bf0d1495c0f8fdb47d88a801/src/UniversalFactory.sol#L305-L346

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,82 @@ library Math {
...
+            x |= x >> 8;
+            x |= x >> 16;
+            x |= x >> 32;
+            x |= x >> 64;
+            x |= x >> 128;
+            x = (x >> 1) + 1;
+
+            uint256 prod0;
+            assembly {
+                // 1. Compute `n ≡ x mod 255`, because `x` is power of two, the resulting `n` can only be one of the
```
Amxx, line 578: I'd write it as
```suggestion
                // 1. Compute `x ≡ n mod 255`, because `x` is power of two, the resulting `n` can only be one of the
```
or
```suggestion
                // 1. Compute the remainder of `x` modulo 255, because `x` is power of two, the resulting `n` can only be one of the
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,82 @@ library Math {
...
-        uint256 result = 0;
-        uint256 exp;
+    function log2(uint256 x) internal pure returns (uint256) {
+        // Efficient branchless algorithm for compute floor(log2(x)), the algorithm works as follow:
+        //
+        // 1. First round down `x` to the closest power of 2 using an modified version of the Seander's `Round up
+        //    to the next power of 2` algorithm, the version used here is modified to round down instead of up.
+        //    ref: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+        //
+        // 2. Compute `n ≡ x mod 255`, we use this `n` value to split the equation in two parts that can be more
```
Amxx, line 547: Usually, notation `x ≡ y mod n` is used with x being anything, and y being between 0 (included) and n (excluded). 
I believe it would be valid to say that about any values x and y, provided that the residue modulo n is the same.

In that case, I think we are explicitelly getting that residue, so I would write it as 
```suggestion
        // 2. Compute `n`, the residue of `x` modulo 255. We use this `n` value to split the equation in two parts that can be more
```
or
```suggestion
        // 2. Compute `n`, the remainder of the division of `x` by 255. We use this `n` value to split the equation in two parts that can be more
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,82 @@ library Math {
...
+                // 2. Compute `prod0 = log2(n)` using the OPCODE BYTE to lookup a 256-bit word, which we use as table. Notice the
+                // maximum size of our table is 32, but their values can be greater than 32, so first we map `n` into an unique
+                // index between 0~31 by computing `n % 11`, this maps all `n` values to an unique index, as demonstrated below.
+
+                // | n ≡ x mod 255 | index ≡ n mod 11 |    prod0 = log2(n)   |
+                // +-+-+-+-+-+-+-+-+-+-+-+-++-+-+-+-+-+-+-+-+-+-+-+-++-+-+-+-+
+                // |        1      |    1 % 11 == 1   |     log2(1) == 0     |
+                // |        2      |    2 % 11 == 2   |     log2(2) == 1     |
+                // |        4      |    4 % 11 == 4   |     log2(4) == 2     |
+                // |        8      |    4 % 11 == 8   |     log2(8) == 3     |
```
Amxx, line 591:
```suggestion
                // |        8      |    8 % 11 == 8   |     log2(8) == 3     |
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,82 @@ library Math {
...
+                // |       16      |   16 % 11 == 5   |    log2(16) == 4     |
+                // |       32      |   32 % 11 == 10  |    log2(32) == 5     |
+                // |       64      |   64 % 11 == 9   |    log2(64) == 6     |
+                // |      128      |  128 % 11 == 7   |   log2(128) == 7     |
+                let index := mod(n, 11)
+
+                // Below we perform the table lookup, the table stores the result of `log2(n)`
+                // in the corresponding byte at `index`.
+                prod0 := byte(index, 0x0000010002040007030605000000000000000000000000000000000000000000)
```
Amxx, line 600: would that work? In would use a `push11` instead of `push32`, and reduce the bytecode size/deployment cost
```suggestion
                prod0 := byte(index, 0x0000010002040007030605)
```

contracts/utils/math/Math.sol
Amxx: alternative would be to use `=` instead of `≡`, which would be more like a programming assignment (right to left) than a mathematical equivalent check.

contracts/utils/math/Math.sol
Amxx, line 578: See discussion above
```suggestion
                // 1. Compute `n = x mod 255`, because `x` is power of two, the resulting `n` can only be one of the
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,82 @@ library Math {
...
+            assembly {
+                // 1. Compute `n ≡ x mod 255`, because `x` is power of two, the resulting `n` can only be one of the
+                // following values: 1, 2, 4, 8, 16, 32, 64 or 128.
+                let n := mod(x, 255)
+
+                // 2. Compute `prod0 = log2(n)` using the OPCODE BYTE to lookup a 256-bit word, which we use as table. Notice the
+                // maximum size of our table is 32, but their values can be greater than 32, so first we map `n` into an unique
+                // index between 0~31 by computing `n % 11`, this maps all `n` values to an unique index, as demonstrated below.
+
+                // | n ≡ x mod 255 | index ≡ n mod 11 |    prod0 = log2(n)   |
```
Amxx, line 586:
```suggestion
                // | n = x mod 255 | index = n mod 11 |    prod0 = log2(n)   |
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,82 @@ library Math {
...
+            x |= x >> 32;
+            x |= x >> 64;
+            x |= x >> 128;
+            x = (x >> 1) + 1;
+
+            uint256 prod0;
+            assembly {
+                // 1. Compute `n ≡ x mod 255`, because `x` is power of two, the resulting `n` can only be one of the
+                // following values: 1, 2, 4, 8, 16, 32, 64 or 128.
+                let n := mod(x, 255)
```
Amxx, line 580: I would compute n outside of the assembly block.

contracts/utils/math/Math.sol
Lohann: Good catch!

contracts/utils/math/Math.sol
Lohann: Will do, was concerned about solidity check division by zero, but in this case the divisor is constant.

contracts/utils/math/Math.sol
Lohann: You are right, was using this symbol wrongly,  I decide to use `n = x mod 255` once seems more readable and intuitive for developers.

contracts/utils/math/Math.sol
Lohann: Confirmed, using Solc `0.8.27` when set optimizer runs below `600`, it outputs:
```
PUSH9 0x010002040007030605 PUSH1 0xa8 SHL
```
But for any value equal or greater than `600` optimizer runs, solidity seems to prioritize execution cost:
```
PUSH30 0x010002040007030605000000000000000000000000000000000000000000
```

Code:
```solidity
pragma solidity >=0.7.0 <0.9.0;

contract Log2 {
    fallback() payable external {
        uint256 x;
        assembly {
            x := calldataload(0)
        }
        unchecked {
            x |= x >> 1;
            x |= x >> 2;
            x |= x >> 4;
            x |= x >> 8;
            x |= x >> 16;
            x |= x >> 32;
            x |= x >> 64;
            x |= x >> 128;
            x = (x >> 1) + 1;
            uint256 prod0;
            {
                uint256 n = x % 255;
                uint256 index = n % 11;
                assembly {
                    prod0 := byte(index, 0x0000010002040007030605000000000000000000000000000000000000000000)
                }
            }
            uint256 prod1 = x >> prod0;
            prod1 *= 0x0008101820283038404850586068707880889098a0a8b0b8c0c8d0d8e0e8f0f8;
            prod1 >>= 248;
            assembly {
                mstore(0, add(prod0, prod1))
                return (0, 0x20)
            }
        }
    } 
}
```
Obs: when comparing algorithms I like to use `fallback() payable external` to remove the selector overhead from the final bytecode, makes it easier to analyze. For example if I don't manually scope `n` and `index`, solidity outputs worst code than without manual scoping.

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,101 @@ library Math {
...
+            x |= x >> 2;
+            x |= x >> 4;
+            x |= x >> 8;
+            x |= x >> 16;
+            x |= x >> 32;
+            x |= x >> 64;
+            x |= x >> 128;
+            // note `x = 0` results in 1 here, when the closest power of two is actually zero given
+            // `2**-infinity == 0` then we should do `(x >> 1) + toUint(x > 0)` instead, but this is not
+            // necessary given `floor(log2(0)) == floor(log2(1))` anyway.
```
Amxx, line 578:
```suggestion
            // necessary given `floor(log2(0)) == floor(log2(1))` anyway.
```
Strictly speaking, `floor(log2(0))` is minux infinity. I'm guessing were `log2` represent the outcome of this function, and not the mathematical log2. Maybe we should rephrase it a bit.

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,101 @@ library Math {
...
+            // | n = x % 255 | index = n % 11 | table[index] = log2(n) |
+            // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+            // |       1     |   1 % 11 == 1  |   table[1] = log2(1)   |
+            // |       2     |   2 % 11 == 2  |   table[2] = log2(2)   |
+            // |       4     |   4 % 11 == 4  |   table[4] = log2(4)   |
+            // |       8     |   8 % 11 == 8  |   table[8] = log2(8)   |
+            // |      16     |  16 % 11 == 5  |   table[5] = log2(16)  |
+            // |      32     |  32 % 11 == 10 |  table[10] = log2(32)  |
+            // |      64     |  64 % 11 == 9  |   table[9] = log2(64)  |
+            // |     128     | 128 % 11 == 7  |   table[7] = log2(128) |
```
Amxx, line 599:
```suggestion
```suggestion
            //                   log2(n) Lookup Table
            // | n = x % 255 | index = n % 11 |  table[index] = log2(n)   |
            // |-------------|----------------|---------------------------|
            // |       1     |   1 % 11 ==  1 | table[ 1] = log2(  1) = 0 |
            // |       2     |   2 % 11 ==  2 | table[ 2] = log2(  2) = 1 |
            // |       4     |   4 % 11 ==  4 | table[ 4] = log2(  4) = 2 |
            // |       8     |   8 % 11 ==  8 | table[ 8] = log2(  8) = 3 |
            // |      16     |  16 % 11 ==  5 | table[ 5] = log2( 16) = 4 |
            // |      32     |  32 % 11 == 10 | table[10] = log2( 32) = 5 |
            // |      64     |  64 % 11 ==  9 | table[ 9] = log2( 64) = 6 |
            // |     128     | 128 % 11 ==  7 | table[ 7] = log2(128) = 7 |
```
```

File: contracts/utils/math/Math.sol [hunk 9]
```diff
@@ -537,41 +537,101 @@ library Math {
...
+            // |         x        | x / n | prod1 = log2(x/n) | index = log2(x/n) / 8 |
+            // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
+            // |  1   =< x < 2⁸   |  1    |    log2(1) == 0   |      0 / 8 == 0       |
+            // |  2⁸  =< x < 2¹⁶  |  2⁸   |   log2(2⁸) == 8   |      8 / 8 == 1       |
+            // |  2¹⁶ =< x < 2²⁴  |  2¹⁶  |  log2(2¹⁶) == 16  |     16 / 8 == 2       |
+            // |  2²⁴ =< x < 2³²  |  2²⁴  |  log2(2²⁴) == 24  |     24 / 8 == 3       |
+            // |        ...       |  ...  |        ...        |          ...          |
+            // | 2²³² =< x < 2²⁴⁰ |  2²³² | log2(2²³²) == 232 |    232 / 8 == 29      |
+            // | 2²⁴⁰ =< x < 2²⁴⁸ |  2²⁴⁰ | log2(2²⁴⁰) == 240 |    240 / 8 == 30      |
+            // | 2²⁴⁸ =< x < 2²⁵⁶ |  2²⁴⁸ | log2(2²⁴⁸) == 248 |    248 / 8 == 31      |
```
Amxx, line 620:
```suggestion
            // |        x        | x / n | prod1 = log2(x/n) | index = log2(x/n) / 8 |
            // |-----------------|-------|-------------------|-----------------------|
            // | 1    ≤ x < 2⁸   |  1    | log2(1   ) == 0   |       0 / 8 == 0      |
            // | 2⁸   ≤ x < 2¹⁶  |  2⁸   | log2(2⁸  ) == 8   |       8 / 8 == 1      |
            // | 2¹⁶  ≤ x < 2²⁴  |  2¹⁶  | log2(2¹⁶ ) == 16  |      16 / 8 == 2      |
            // | 2²⁴  ≤ x < 2³²  |  2²⁴  | log2(2²⁴ ) == 24  |      24 / 8 == 3      |
            // |       ...       |  ...  |        ...        |          ...          |
            // | 2²³² ≤ x < 2²⁴⁰ |  2²³² | log2(2²³²) == 232 |     232 / 8 == 29     |
            // | 2²⁴⁰ ≤ x < 2²⁴⁸ |  2²⁴⁰ | log2(2²⁴⁰) == 240 |     240 / 8 == 30     |
            // | 2²⁴⁸ ≤ x < 2²⁵⁶ |  2²⁴⁸ | log2(2²⁴⁸) == 248 |     248 / 8 == 31     |
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,101 @@ library Math {
...
+            // moves the result to the most significant byte, finally extract `log2(x/n) by shift right the table.
+            //
+            // prod1 = log2(x / n) = (x / n * table) >> 248
+            uint256 prod1 = x >> prod0;
+            prod1 *= 0x0008101820283038404850586068707880889098a0a8b0b8c0c8d0d8e0e8f0f8;
+            prod1 >>= 248;
+
+            // 5. Sum both values to get the final result:
+            // log2(x) = log2(x / n) + log2(n) = prod0 + prod1
+            return prod0 + prod1;
```
Amxx, line 633:
```suggestion
            // log2(x) = log2(x / n) + log2(n) = prod1 + prod0
            return prod1 + prod0;
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,101 @@ library Math {
...
+            // | 1    ≤ x < 2⁸   |  1    |  log2(1   ) == 0    |       0 / 8 == 0      |
+            // | 2⁸   ≤ x < 2¹⁶  |  2⁸   |  log2(2⁸  ) == 8    |       8 / 8 == 1      |
+            // | 2¹⁶  ≤ x < 2²⁴  |  2¹⁶  |  log2(2¹⁶ ) == 16   |      16 / 8 == 2      |
+            // | 2²⁴  ≤ x < 2³²  |  2²⁴  |  log2(2²⁴ ) == 24   |      24 / 8 == 3      |
+            // |       ...       |  ...  |         ...         |          ...          |
+            // | 2²³² ≤ x < 2²⁴⁰ |  2²³² |  log2(2²³²) == 232  |     232 / 8 == 29     |
+            // | 2²⁴⁰ ≤ x < 2²⁴⁸ |  2²⁴⁰ |  log2(2²⁴⁰) == 240  |     240 / 8 == 30     |
+            // | 2²⁴⁸ ≤ x < 2²⁵⁶ |  2²⁴⁸ |  log2(2²⁴⁸) == 248  |     248 / 8 == 31     |
+            //
+            // Notice compute `(x / n) * value` is equivalent to compute `value << (index * 8)`, we use this fact
```
Amxx, line 622:
```suggestion
            // It is important to note that `value * (x / n)` is equal to `value << (8 * index)`. We use this fact
            // to build a single 32-byte word where `table[index] = log2(x/n)`. Multiply the table by `x/n` moves the 
            // result to the most significant byte, which we can then extract `log2(x/n) by shifting right.
```

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,101 @@ library Math {
...
+            // to build a single 32-byte word using `table[index] = log2(x/n)`, then multiply the table by `x/n`
+            // moves the result to the most significant byte, finally extract `log2(x/n) by shift right the table.
```
Amxx, line 624:
```suggestion
            // It is important to note that `value * (x / n)` is equal to `value << (8 * index)`. We use this fact
            // to build a single 32-byte word where `table[index] = log2(x/n)`. Multiply the table by `x/n` moves the 
            // result to the most significant byte, which we can then extract `log2(x/n) by shifting right.
```

contracts/utils/math/Math.sol
Lohann: True, once we are dealing with natural numbers the actual result is not representable, theoretically the correct would be revert like a division by zero, but then this would break a lot of libraries who rely on it.
Will think in a better description here.

cairoeth:
@Lohann could you take a look at https://github.com/Lohann/openzeppelin-contracts/pull/1? cross sharing:

> We can do a much more efficient and simple implementation by employing a binary search to quickly narrow down the most significant bit, followed by a small lookup table for the final 4 bits.

> It's more performant as we avoid using multiple lookup tables and modulo.

Lohann:
Great work @cairoeth ! your solution is cleaner, cheaper and generates a smaller binary, will update this PR with your proposal.

cairoeth:
@Lohann feel free to merge my PR, which should update this PR. thanks!

contracts/utils/math/Math.sol
Lohann: Not required anymore, once switch to @cairoeth implementation.

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,49 @@ library Math {
...
+        // |    1101    |   13    |        table[13] = 3        |
+        // |    1110    |   14    |        table[14] = 3        |
+        // |    1111    |   15    |        table[15] = 3        |
+        //
+        // The lookup table is represented as a 32-byte value with the MSB positions for 0-15 in the last 16 bytes.
+        assembly {
+            x := byte(x, 0x0000010102020202030303030303030300000000000000000000000000000000)
         }
-        return result;
+        r |= x;
```
Amxx, line 582: Maybe not as cheap, but way simpler/more readable
```
        r = SafeCast.toUint(x > 0xffffffffffffffffffffffffffffffff) << 7;
        r |= SafeCast.toUint((x >> r) > 0xffffffffffffffff) << 6;
        r |= SafeCast.toUint((x >> r) > 0xffffffff) << 5;
        r |= SafeCast.toUint((x >> r) > 0xffff) << 4;
        r |= SafeCast.toUint((x >> r) > 0xff) << 3;
        r |= SafeCast.toUint((x >> r) > 0xf) << 2;
        r |= SafeCast.toUint((x >> r) > 0x3) << 1;
        r |= SafeCast.toUint((x >> r) > 0x1) << 0;
```

cairoeth: I think that
```
        r |= SafeCast.toUint((x >> r) > 0x3) << 1;
        r |= SafeCast.toUint((x >> r) > 0x1);
```
is less readable than `x >>= r` -- also the number of operations is way larger, so not totally worth it. @Lohann what do you think?

Amxx:
```
        r |= SafeCast.toUint((x >> r) > 0x3) << 1;
        r |= SafeCast.toUint((x >> r) > 0x1);
```

is not an alternative to `x >>= r`

Its an alternative to
```
        x >>= r;
        assembly {
            x := byte(x, 0x0000010102020202030303030303030300000000000000000000000000000000)
        }
        r |= x;
```
(and all the comments that come with it.

Amxx, line 582: alternativelly (IMO clearer because it doesn't change the semantics of what x represents)

```solidity
assembly ("memory-safe") {
    r := or(r, byte(shl(r, x), 0x0000010102020202030303030303030300000000000000000000000000000000))
}
```

Amxx (approved):

Lohann:
Hi guys, sorry for the late reply got super busy here, but this seems amazing, thanks @cairoeth and @Amxx, this solutions is cleaner and better than my initial idea, and considering the minimal code I implemented [here](https://github.com/Lohann/openzeppelin-contracts/pull/1#issuecomment-2419797794), it both minimizes the bytecode size and gas used 🎉

contracts/utils/math/Math.sol
```diff
@@ -537,41 +537,48 @@ library Math {
...
-            exp = 2 * SafeCast.toUint(value > (1 << 2) - 1);
-            value >>= exp;
-            result += exp;
-
-            result += SafeCast.toUint(value > 1);
+    function log2(uint256 x) internal pure returns (uint256 r) {
+        // log2 implementation based on De Bruijn Multiplication.
+        // ref: https://www.chessprogramming.org/BitScan#De_Bruijn_Multiplication_2
+
+        // If value has upper 128 bits set, log2 result is at least 128
```
cairoeth, line 544: we are not using a de bruijn sequence in this last version (it's just a simple lookup table) so i'd remove this
```suggestion
        // If value has upper 128 bits set, log2 result is at least 128
```

cairoeth (approved):

PR #5239, opened 2024-10-04
Split StorageSlot into TransientSlot

ernestognw:
To avoid disrupting StorageSlot users, we're splitting the transient storage part to another library that's compiled through Solidity ^0.8.24

#### PR Checklist

- [ ] Tests
- [ ] Documentation


scripts/generate/run.js
```diff
@@ -39,9 +39,11 @@ for (const [file, template] of Object.entries({
   'utils/structs/EnumerableMap.sol': './templates/EnumerableMap.js',
   'utils/SlotDerivation.sol': './templates/SlotDerivation.js',
   'utils/StorageSlot.sol': './templates/StorageSlot.js',
+  'utils/StorageSlotTransient.sol': './templates/StorageSlotTransient.js',
```
Amxx, line 42: I'm think "Transient" and "Storage" should go together. Also this is not a "transeient-variant" of the StorageSlot library (because it uses different type, different functions, ...)

So I'd be in vafor of changing the name. Proposal include:
- [ ] `Transient.sol`
- [ ] `TransientSlot.sol`
- [ ] `TransientStorageSlot.sol`
- [ ] Something else ?

contracts/utils/README.adoc
```diff
@@ -34,7 +34,8 @@ Miscellaneous contracts and libraries containing utility functions you can use t
  * {Strings}: Common operations for strings formatting.
  * {ShortString}: Library to encode (and decode) short strings into (or from) a single bytes32 slot for optimizing costs. Short strings are limited to 31 characters.
  * {SlotDerivation}: Methods for deriving storage slot from ERC-7201 namespaces as well as from constructions such as mapping and arrays.
- * {StorageSlot}: Methods for accessing specific storage slots formatted as common primitive types. Also include primitives for reading from and writing to transient storage (only value types are currently supported).
+ * {StorageSlot}: Methods for accessing specific storage slots formatted as common primitive types. 
+ * {TransientSlot}: primitives for reading from and writing to transient storage (only value types are currently supported).
```
Amxx, line 38:
```suggestion
 * {TransientSlot}: Primitives for reading from and writing to transient storage (only value types are currently supported).
```

lib/forge-std
Amxx, line 1: is that change needed here? If not, I'd do it in a separate PR.

Amxx (dismissed):
LGTM

I'd like @frangio opinion, particularly on the UDVT defined and used in TransientSlot.sol

Amxx:
For the record, here is the rational for this change:

- In the upgrade plugins, we compile ERC-1967 targetting paris. The logic is that we don't know which chain is going to be targeted, and weither it supports push0 or not. So we ship "old versions" of the proxy
- This proxy code relies on the StorageSlot library (for reading/writting the implementation slot)
- The compiler will refuse to compile that is we set the evm target to paris and if the library contains references to tstore/tload (even if they are not actually used of the contract being compiled)

Other users may have similar situations where they use `StorageSlot.sol` with 0.8.20~0.8.23 ... or targeting evm version that do not include transient storage. They should be able to continue using `StorageSlot.sol`. 

Overall, no contract should import any reference to `tstore/tload` unless they actually plan on using transient storage.

frangio:
> I'd like @frangio opinion, particularly on the UDVT defined and used in TransientSlot.sol

The type is basically the same, it's just been moved, right?

I've expressed my opinion before that the "transient slot" type should be different from the "(non-transient) storage slot" type. By separating the libraries this PR is closer to that goal, so I think that's good.

arr00 (approved):

arr00 (approved):

ernestognw:
For the record, we agreed with the naming and parameters during a call. The last update only fixed the lining rules.
Thanks for approving @arr00!

PR #5242, opened 2024-10-08
Add Account Abstraction utils and interfaces

ernestognw:
Smaller version of #5182 

This includes:

- Vendored ERC4337 entrypoint
- ERC4337Utils
- ERC7579Utils
- Extra Packing variants

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changeset entry (run `npx changeset add`)

PR #5243, opened 2024-10-09
Add ERC7739 and ERC7739Utils

ernestognw:
Follow up from #5242 (must be reviewed first) and smaller version of #5182

This PR includes the following primitives:

- ERC7739Utils: A wrapped EIP-712 approach based on [@frangio's](https://github.com/frangio/eip712-wrapper-for-eip1271) and @Vectorized's work
- ERC7739Signer: An ERC1271 abstract contract that implements ERC7739Utils

Both of these will be required for the Account contracts

#### PR Checklist

- [x] Tests
- [x] Documentation


ernestognw:
Most of this work is inspired in [ERC-7739](https://ethereum-magicians.org/t/erc-7739-readable-typed-signatures-for-smart-accounts/20513/2) although I couldn't find an official reference to an ERC. We can update if needed

contracts/vendor/erc4337-entrypoint/core/EntryPoint.sol
Amxx, line 1: This file (and many other in the vendor/erc4337-entrypoint directory) has been modified since it was copied from the origin. This is because of a linter run.

IMO the vendored file receive as little modification as possible. We have to rewrite some imports but that should honestly be all. All modification should be clearly documented, and line number should refer to the same logic on both the original and the vendored version.

That is why the `account-abstraction` branch dissables the linter on all the vendored subfolder.

Amxx:
The diff is completelly bugged. Changing base doesn't reset it

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,278 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {IERC5267} from "../../interfaces/IERC5267.sol";
+import {MessageHashUtils} from "./MessageHashUtils.sol";
+
+/**
+ * @dev Utilities to process https://eips.ethereum.org/EIPS/eip-7739[ERC-7739] typed data signatures
```
ernestognw, line 9: The EIP is still not in that URL but I think it should be as soon as @Vectorized merges https://github.com/ethereum/ERCs/pull/441

.changeset/sweet-grapes-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`ERC7739Utils: Add a library that implements a defensive rehashing mechanism to prevent replayability of smart contract signatures based on the ERC-7739.
```
Amxx, line 5:
```suggestion
`ERC7739Utils`: Add a library that implements a defensive rehashing mechanism to prevent replayability of smart contract signatures based on the ERC-7739.
```

contracts/mocks/ERC7739SignerRSA.sol
```diff
@@ -0,0 +1,21 @@
...
+
+pragma solidity ^0.8.20;
+
+import {RSA} from "../utils/cryptography/RSA.sol";
+import {ERC7739Signer} from "../utils/cryptography/draft-ERC7739Signer.sol";
+import {EIP712} from "../utils/cryptography/EIP712.sol";
+
+contract ERC7739SignerRSA is ERC7739Signer {
+    bytes private _e;
+    bytes private _n;
```
Amxx, line 11: Are these storage accesses 4337 compliant ?

Amxx:
IMO this should be in the go in the community repo, with the account work. I feels very new, and weither we want it or not depends on the account design (which I would also put in community repo)

contracts/mocks/ERC7739SignerRSA.sol
ernestognw: This contract is not meant for ERC4337 but for general ERC7739 message signing. When used as part of an ERC4337 Account, storage accesses are allowed to the account storage itself.

The problem is when an account uses another contract's `isValidSignature` during the validation phase. In that case the transaction can still be processed privately (i.e. through any provider), is just that they won't rely it to the canonical mempool

contracts/mocks/ERC7739SignerECDSA.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {ECDSA} from "../utils/cryptography/ECDSA.sol";
+import {ERC7739Signer} from "../utils/cryptography/draft-ERC7739Signer.sol";
+import {EIP712} from "../utils/cryptography/EIP712.sol";
+
+contract ERC7739SignerECDSA is ERC7739Signer {
```
Amxx, line 9:
```suggestion
contract ERC7739SignerECDSAMock is ERC7739Signer {
```

contracts/mocks/ERC7739SignerP256.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {P256} from "../utils/cryptography/P256.sol";
+import {ERC7739Signer} from "../utils/cryptography/draft-ERC7739Signer.sol";
+import {EIP712} from "../utils/cryptography/EIP712.sol";
+
+contract ERC7739SignerP256 is ERC7739Signer {
```
Amxx, line 9:
```suggestion
contract ERC7739SignerP256Mock is ERC7739Signer {
```

contracts/mocks/ERC7739SignerRSA.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {RSA} from "../utils/cryptography/RSA.sol";
+import {ERC7739Signer} from "../utils/cryptography/draft-ERC7739Signer.sol";
+import {EIP712} from "../utils/cryptography/EIP712.sol";
+
+contract ERC7739SignerRSA is ERC7739Signer {
```
Amxx, line 9:
```suggestion
contract ERC7739SignerRSAMock is ERC7739Signer {
```

contracts/mocks/ERC7739SignerRSA.sol
Amxx: Its a mock anyway, lets resolve that.

It will come back if we ever build a 7579 module that relies on 1271 signature that are resolved by 7739 Signers

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,276 @@
...
+        uint256 contentsEnd = contentsTypeEnd - contentsTypeLength;
+        if (contentsEnd < 64) return (signature[0:0], 0, 0, signature[0:0]);
+
+        uint256 appSeparatorEnd = contentsEnd - 32;
+        uint256 originalSigEnd = appSeparatorEnd - 32;
+
+        originalSig = signature[0:originalSigEnd];
+        appSeparator = bytes32(signature[originalSigEnd:appSeparatorEnd]);
+        contents = bytes32(signature[appSeparatorEnd:contentsEnd]);
+        contentsType = signature[contentsEnd:contentsTypeEnd];
```
Amxx, line 76: all that should probably be unchecked

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,276 @@
...
+    )
+        internal
+        pure
+        returns (bytes calldata originalSig, bytes32 appSeparator, bytes32 contents, bytes calldata contentsType)
+    {
+        uint256 sigLength = signature.length;
+        if (sigLength < 66) return (signature[0:0], 0, 0, signature[0:0]);
+
+        uint256 contentsTypeEnd = sigLength - 2; // Last 2 bytes
+        uint256 contentsTypeLength = uint16(bytes2(signature[contentsTypeEnd:sigLength]));
```
Amxx, line 64: specifying the end is optional, but maybe it saves gas by avoiding a reload ?

```suggestion
        uint256 contentsTypeLength = uint16(bytes2(signature[contentsTypeEnd:]));
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
Amxx, line 76:
```suggestion
        originalSig = signature[:originalSigEnd];
        appSeparator = bytes32(signature[originalSigEnd:appSeparatorEnd]);
        contents = bytes32(signature[appSeparatorEnd:contentsEnd]);
        contentsType = signature[contentsEnd:];
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+    ) internal pure returns (bool valid, bytes calldata contentsTypeName) {
+        uint256 contentsTypeLength = contentsType.length;
+        if (contentsTypeLength == 0) return (false, _emptyCalldataBytes());
+
+        // Does not start with a-z or (
+        bytes1 high = contentsType[0];
+        if ((high > 0x60 && high < 0x7b) || high == 0x28) return (false, _emptyCalldataBytes()); // a-z or (
+
+        // Find the start of the arguments
+        uint256 argsStart = contentsType.indexOf(bytes1("("));
```
Amxx, line 236: Here there is an implicit calldata to memory copy. Not sure how bad it is, but maybe we want a `indexOf` that operates on calldata bytes.

Amxx: This could be done in a further optimisation. Not blocking.

contracts/utils/cryptography/draft-ERC7739Signer.sol
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * Linking the signature to the EIP-712 domain separator is a security measure to prevent signature replay across different
+ * EIP-712 domains (e.g. a single offchain owner of multiple contracts).
+ *
+ * This contract requires implementing the {_validateSignature} function, which passes the wrapped message hash,
+ * which may be either an typed data or a personal sign nested type.
+ *
+ * NOTE: {EIP712} uses {ShortStrings} to optimize gas costs for short strings (up to 31 characters).
+ * Consider that strings longer than that will use storage, which may limit the ability of the signer to
+ * be used within the ERC-4337 validation phase (due to ERC-7562 storage access rules).
```
frangio, line 22: Is this accurate? The account is able to read its own storage during validation so I don't think there will be any issues.

```suggestion
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+     */
+    bytes32 internal constant _NESTED_PERSONAL_SIGN_TYPEHASH = keccak256("PersonalSign(bytes prefixed)");
+
+    /**
+     * @dev Error when the contents type is invalid. See {tryValidateContentsType}.
+     */
+    error InvalidContentsType();
+
+    /**
+     * @dev Parses a nested signature into its components. See {nest}.
```
frangio, line 42:
```suggestion
     * @dev Parses a nested signature into its components.
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio, line 1: `wrap` and `unwrap` are more like `encode` and `decode`.

I recommend the following renames:
- `wrapTypedDataSig` → `encodeSignature`
- `unwrapTypedDataSig` → `decodeSignature`

contracts/utils/cryptography/draft-ERC7739Signer.sol
```diff
@@ -0,0 +1,119 @@
...
+        bytes calldata signature
+    ) internal view virtual returns (bool) {
+        (bytes calldata originalSignature, bytes32 nestedHash) = _nestedTypedDataHash(signature);
+        return hash == nestedHash && _validateSignature(nestedHash, originalSignature);
+    }
+
+    /**
+     * @dev Nested personal signature verification.
+     */
+    function _isValidNestedPersonalSigSignature(
```
frangio, line 60: In `ERC7739Utils` the function is named `...PersonalSign...` here it is `...PersonalSig...`

I think the first one is what we want in order to match the `personal_sign` RPC method.

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+                )
+            );
+    }
+
+    /**
+     * @dev Try to validate the contents type is a valid EIP-712 type.
+     *
+     * A valid `contentsType` is considered invalid if it's empty or it:
+     *
+     * - Starts with a-z or (
```
frangio, line 222: I left [a comment](https://ethereum-magicians.org/t/erc-7739-readable-typed-signatures-for-smart-accounts/20513/4?u=frangio) in the ERC discussion. Disallowing starting lowercase letters is way too restrictive. There’s nothing in EIP-712 that says custom types have to start with an uppercase letter, so we have to assume some applications use lowercase letters. The smart account won’t be usable with those applications which is IMO very serious. I'd recommend against shipping this restriction.

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+     *          )
+     *      )
+     * ));
+     * ```
+     */
+    function toNestedPersonalSignHash(bytes32 separator, bytes32 contents) internal pure returns (bytes32) {
+        return
+            MessageHashUtils.toTypedDataHash(
+                separator,
+                keccak256(abi.encode(_NESTED_PERSONAL_SIGN_TYPEHASH, MessageHashUtils.toEthSignedMessageHash(contents)))
```
frangio, line 124: I think there is an issue with this function. See how there is a hardcoded "32" in the documentation. Also see how the contents argument is `bytes32`.

`personal_sign` messages can be arbitrary length. `contents` should be `bytes`.

Note that there are two variants of [`MessageHashUtils.toEthSignedMessageHash`](https://docs.openzeppelin.com/contracts/5.x/api/utils#MessageHashUtils-toEthSignedMessageHash-bytes32-). I think the first one is there for historical purposes. It should probably be deprecated.

I think we need the tests to use the `personal_sign` RPC call to make sure this is correct.

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+ * that are specific to an {EIP712} domain.
+ *
+ * This library provides methods to wrap, unwrap and operate over typed data signatures with a defensive
+ * rehashing mechanism that includes the application's {EIP712-_domainSeparatorV4} and preserves
+ * readability of the signed content using an EIP-712 nested approach.
+ *
+ * A smart contract domain can validate a signature for a typed data structure in two ways:
+ *
+ * - As an application validating a typed data signature. See {toNestedTypedDataHash}.
+ * - As a smart contract validating a raw message signature. See {toNestedPersonalSignHash}.
```
frangio, line 20: What is the difference between an application and a smart contract? I don't understand this comment.

contracts/utils/cryptography/draft-ERC7739Signer.sol
Amxx: should be solved by my rewrite

contracts/utils/cryptography/draft-ERC7739Utils.sol
Amxx: One strange thing to me is that it accepts `(` as a first letter. Is that reasonable to not have a name and only have a description ?

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,288 @@
...
+     *          )
+     *      )
+     *  )
+     *);
+     *
+     * NOTE: The arguments should be those of the verifying application. See {EIP712-_domainSeparatorV4}
+     * and {IERC5267-eip712Domain} for more details of how to obtain these values. Respectively, they
+     * must be obtained from the verifying contract (e.g. an Account) and the application domain.
+     */
+    function toNestedTypedDataHash(
```
Amxx, line 162: Do we really need that function ? Why not use `MessageHashUtils.toTypedDataHash` directly ?

contracts/utils/cryptography/draft-ERC7739Signer.sol
Amxx:
```suggestion
 *
 * NOTE: {EIP712} uses {ShortStrings} to optimize gas costs for short strings (up to 31 characters). Consider that 
 * strings longer than that, and strings of any size for the upgradable variants of this contract, will be placed in 
 * storage. This may limit the ability of the signer to be used within the ERC-4337 validation. If the sender of the
 * user operation performs an ERC-1271 check against this signer, the ERC-7562 storage access rules will be broken.
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -217,36 +214,33 @@ library ERC7739Utils {
...
+     * A valid `contentsType` is considered invalid if it's empty or if:
+     * - it starts with a '('.
+     * - it contains any of the following bytes: , )\x00
      *
-     * - Starts with a-z or (
-     * - Contains any of the following bytes: , )\x00
+     * NOTE: This is a looser take on the ERC very strict restrictions. This part appears to be under discussion, and
+     * therefore the restrictions implemented here may change in a future release.
      */
     function tryValidateContentsType(
```
frangio, line 224: The new logic makes more sense IMO but does not match the ERC so we need the ERC to lift the a-z restriction.

contracts/utils/cryptography/draft-ERC7739Signer.sol
frangio: > That is in case account A relies on another other contract B for validation:

Yes but this is a much larger issue than EIP-712 name and version strings, because the public keys will usually be in storage.

IMO mentioning this here is a distraction.

contracts/utils/cryptography/draft-ERC7739Signer.sol
```diff
@@ -0,0 +1,105 @@
...
+    function isValidSignature(bytes32 hash, bytes calldata signature) public view virtual returns (bytes4 result) {
+        return _isValidSignature(hash, signature) ? IERC1271.isValidSignature.selector : bytes4(0xffffffff);
+    }
+
+    /**
+     * @dev Internal version of {isValidSignature} that returns a boolean.
+     */
+    function _isValidSignature(bytes32 hash, bytes calldata signature) internal view virtual returns (bool) {
+        return
+            _isValidNestedPersonalSignSignature(hash, signature) || _isValidNestedTypedDataSignature(hash, signature);
```
frangio, line 44: It's not clear to me whether this matches the spec: https://ercs.ethereum.org/ERCS/erc-7739#signature-verification-workflow-deduction

> If the signature contains the correct data to reconstruct the `hash`, the `isValidSignature` function MUST perform the `TypedDataSign` workflow. Otherwise, the `isValidSignature` function MUST perform the `PersonalSign` workflow.

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: Yes, I actually think this is wrong.

The `personal_sign` hash is generated by the application contract. There is no need to regenerate it here, and we don't have the message contents so we can't. 

This function should look like this:

```solidity
    function toNestedPersonalSignHash(bytes32 separator, bytes32 prefixed) internal pure returns (bytes32) {
        return
            MessageHashUtils.toTypedDataHash(
                separator,
                keccak256(abi.encode(_NESTED_PERSONAL_SIGN_TYPEHASH, prefixed))
            );
    }
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: I realize now that the "TypedData" part of the name was relevant. The encoding/decoding is only done for TypedData signatures. 

I still believe encode/decode are more accurate than wrap/unwrap, but I think we should also include "TypedData" in the name.

contracts/utils/cryptography/draft-ERC7739Utils.sol
ernestognw: My understanding is that it's a `bytes32 hash` too. I'd say supporting arbitrary `bytes` opens the door to more obfuscated signature attacks.

> personal_sign does support arbitrary content length, but there is not way for this content to pass through the ERC-1271 bottleneck, is there?

Also, I've seen [it used quite frequently](https://sourcegraph.com/search?q=context:global+repo:tintinweb/smart-contract-sanctuary+file:/mainnet/+/%5CbisValidSignature%5C%28bytes+/&patternType=standard&sm=0). Some contracts reference it's a "legacy" `isValidSignature` version.

contracts/utils/cryptography/draft-ERC7739Utils.sol
ernestognw: should be that an application is third-party to the smart contract.

the point here is that a personal signature is "personal" for the smart contract (its own domain), and an application is an external domain.

not sure if it makes sense, though

contracts/utils/cryptography/draft-ERC7739Utils.sol
ernestognw: Good point. It's result of multiple rounds of iteration but I decided to keep it just for the instructions. I'd leave it for the completeness of this library. It'd be fine if you guys decide to remove it.

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,282 @@
...
+                abi.encodePacked(
+                    "TypedDataSign(",
+                    contentsTypeName,
+                    "bytes1 fields,",
+                    "string name,",
+                    "string version,",
+                    "uint256 chainId,",
+                    "address verifyingContract,",
+                    "bytes32 salt,",
+                    "uint256[] extensions",
```
frangio, line 207: Currently this seems like it's doing a lot of concatenation at runtime that can be done statically.

```suggestion
                    "bytes1 fields,string name,string version,uint256 chainId,address verifyingContract,bytes32 salt,uint256[] extensions",
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,282 @@
...
+                    ")",
```
frangio, line 208: Currently this seems like it's doing a lot of concatenation at runtime that can be done statically.

```suggestion
                    "bytes1 fields,string name,string version,uint256 chainId,address verifyingContract,bytes32 salt,uint256[] extensions)",
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: Personal signatures can also be requested and validated by third-party contracts. The distinction is not about the validating domain but the format of the signature.

contracts/utils/cryptography/draft-ERC7739Utils.sol
ernestognw: > The encoding/decoding is only done for TypedData signatures.

Yes, thanks for noting! I'd think `encodeTypedDataSignature` or `encodeTypedDataSig` and their homonymous are fine

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: My initial comment in this thread was misguided. The issue is another one.
> The `personal_sign` hash is generated by the application contract before calling `isValidSignature`.

contracts/utils/cryptography/draft-ERC7739Signer.sol
Amxx: There are ways to put the public keys outside of storage (for example using a Proxy with immutable args) ... but the dependency on ERC712 prevents is just too much.

I don't have a strong opinion either way :/

I'm considering merging with (or without) the message for the audit, and have PR with all the proposed changes (in this case adding or removing) that we'll discuss in parallele of the audit.

contracts/utils/cryptography/draft-ERC7739Utils.sol
Amxx: @frangio wdyt ?

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * @dev Same as {NESTED_TYPED_DATA_TYPEHASH} but with the `contentsTypeName` already validated.
+     *
+     * See {tryValidateContentsType}.
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function NESTED_TYPED_DATA_TYPEHASH(
+        bytes calldata contentsType,
+        bytes calldata contentsTypeName
```
Amxx, line 194: @frangio should these be strings ???

contracts/utils/cryptography/draft-ERC7739Utils.sol
Amxx: I'm confused by the encoding done here.

Shouldn't there by a comma (and a variable name) between "contentsTypeName" and "bytes1 fields" ???

contracts/utils/cryptography/draft-ERC7739Utils.sol
Amxx: SO what do we do, re-add the lowercase restriction?

test/helpers/eip712.js
```diff
@@ -36,10 +36,40 @@ function hashTypedData(domain, structHash) {
   );
 }
 
+function hashNestedTypedDataType(contentsTypeName, contentsType) {
+  return ethers.solidityPackedKeccak256(
+    ['string'],
+    [
+      `TypedDataSign(${contentsTypeName}bytes1 fields,string name,string version,uint256 chainId,address verifyingContract,bytes32 salt,uint256[] extensions)${contentsType}`,
```
Amxx, line 43: THis string is 
```TypedDataSign(SomeTypebytes1 fields,string name,string version,uint256 chainId,address verifyingContract,bytes32 salt,uint256[] extensions)SomeType()```

IMO it sould be
```
TypedDataSign(SomeType something,bytes1 fields,string name,string version,uint256 chainId,address verifyingContract,bytes32 salt,uint256[] extensions)SomeType()```

Amxx: That is documented properly in the reference implemenation https://ethereum-magicians.org/t/erc-7739-readable-typed-signatures-for-smart-accounts/20513#reference-implementation-23

But not in this section https://ethereum-magicians.org/t/erc-7739-readable-typed-signatures-for-smart-accounts/20513#typeddatasign-final-hash-11

Amxx: The more I read the tests, the worst I feel about it all. This should have been catched by some test doing a full `ethers.TypedDataEncoder.hash` or an actual `signTypedData` on the full EIP712 wrapper, but we don't have that. Instead we have some tests taht are overly complexe for testing actually nothing ...

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: I guess I would keep the logic of the ERC until that gets changed. But I think it needs to be changed before this is released.

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: Slight preference for keeping this function. But I also was wondering whether it made sense, so I'd be ok with removing it.

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: Yes, that seems like a bug here and also in the ERC.

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,282 @@
...
+    ) internal pure returns (bytes32) {
+        return
+            keccak256(
+                abi.encodePacked(
+                    "TypedDataSign(",
+                    contentsTypeName,
```
Amxx, line 200: This is incorrect, It is missing a ` contents,` to be a properly formated EIP-712 struct definition

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,282 @@
...
+        bytes32 salt,
+        uint256[] memory extensions
+    ) internal view returns (bytes32) {
+        (, bytes calldata contentsTypeName) = tryValidateContentsType(contentsType);
+        return
+            keccak256(
+                abi.encode(
+                    NESTED_TYPED_DATA_TYPEHASH(contentsType, contentsTypeName),
+                    contents,
+                    keccak256(bytes(name)),
```
Amxx, line 266: This is missing a field. The field is part of the Struct definition, but is missing here. I'm not sure where we should take it from

Amxx (changes requested):
Right now, there are obvious issues in the code that were overlooked because of unapropriate testing.

The testing implements the same bad behavior in javascript as we have in solidity. Trying to write a test that does signing on the JS part using the propoer wallet tools, namely ethers `this.signer.signTypedData` helps surface obvious issues.

Currently, I don't think this is ready!

I'll work on doing proper testing asap, but considering both work constrains (5.1) and personnal ones, I'm dobting this will be ready for monday's audit.

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: I think so, but I don't know if it makes a difference.

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: Correct one:
```suggestion
                    " contents,bytes1 fields,string name,string version,uint256 chainId,address verifyingContract,bytes32 salt,uint256[] extensions)",
```

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: It has to be added as an argument in this function. In `ERC7739Signer._isValidNestedTypedDataSignature` it needs to be taken from the return value of `eip712Domain()`. It's currently ignored.

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,277 @@
...
+        bytes calldata contentsType,
+        bytes32 contents,
+        bytes1 fields,
+        string memory name,
+        string memory version,
+        address verifyingContract,
+        bytes32 salt,
+        uint256[] memory extensions
+    ) internal view returns (bytes32) {
+        (, bytes calldata contentsTypeName) = tryValidateContentsType(contentsType);
```
Amxx, line 254: Is it ok to throw the boolean away, and still compute the hash if the contentsType is invalid ?

contracts/utils/cryptography/draft-ERC7739Utils.sol
```diff
@@ -0,0 +1,277 @@
...
+        return
+            keccak256(
+                abi.encode(
+                    NESTED_TYPED_DATA_TYPEHASH(contentsType, contentsTypeName),
+                    contents,
+                    fields,
+                    keccak256(bytes(name)),
+                    keccak256(bytes(version)),
+                    block.chainid,
```
Amxx, line 265: Is there any use case where this library is used to perform a verification/validate a `typedDataNestedStructHash` for another chain ?

contracts/utils/cryptography/draft-ERC7739Utils.sol
frangio: It's documented above that this function doesn't validate it, but I don't know why.

Amxx:
Code an tests are making good progress, but the underlying ERC is still subject to a lot of discussion. 

Consequently, I don't think we should include that in 5.2. I'd rahter wait for to see how the ERC discussion evolve and target a release of this code in 5.3

PR #5252, opened 2024-10-11
CAIP helpers

Amxx:
Depends on #5166

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changeset entry (run `npx changeset add`)

PR #5255, opened 2024-10-14
Crosschain ERCXXXX interface and receiver

Amxx:
Depends on #5252

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/crosschain/draft-ERCXXXXReceiver.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import {IERCXXXXGatewayDestinationPassive, IERCXXXXReceiver} from "../interfaces/draft-IERCXXXX.sol";
+
+abstract contract ERCXXXXReceiver is IERCXXXXReceiver {
+    address private immutable _gateway;
+
+    event MessageReceived(bytes gatewayMessageKey, string source, string sender, bytes payload, bytes[] attributes);
```
Amxx, line 10: move to mock (remove from here)

contracts/crosschain/draft-ERCXXXXReceiver.sol
```diff
@@ -0,0 +1,55 @@
...
+    error ERCXXXXReceiverInvalidGateway(address gateway);
+
+    constructor(address gateway_) {
+        _gateway = gateway_;
+    }
+
+    function isGateway(address instance) public view virtual returns (bool) {
```
Amxx, line 17: make internal

contracts/crosschain/draft-ERCXXXXReceiver.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import {IERCXXXXGatewayDestinationPassive, IERCXXXXReceiver} from "../interfaces/draft-IERCXXXX.sol";
+
+abstract contract ERCXXXXReceiver is IERCXXXXReceiver {
+    address private immutable _gateway;
```
Amxx, line 8: remove from here

cairoeth:
`Bytes.slice` is using `mcopy` in assembly so it works on 0.8.24, so want to confirm that we are upgrading to 0.8.25 for `mcopy` in code generation

contracts/crosschain/draft-ERC7786Receiver.sol
```diff
@@ -0,0 +1,60 @@
...
+        address gateway,
+        bytes calldata gatewayMessageKey,
+        string calldata source,
+        string calldata sender,
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) public payable virtual {
+        if (_isKnownGateway(msg.sender)) {
+            // Active mode
+            // no extra check
```
cairoeth, line 33: shouldn't this check that `gatewayMessageKey` and `gateway` are zero https://github.com/ethereum/ERCs/pull/673/files#diff-07848a22c16b11582927b37d76b2c54a0617b1444b30d11bca783ac98eaed01fR150? or is that checked by the gateway?

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * @dev Interface for ERC-7786 source gateways.
+ *
+ * See ERC-7786 for more details
+ */
+interface IERC7786GatewaySource {
+    /**
+     * @dev Event emitted when a message is created. If `outboxId` is zero, no further processing is necessary, and
+     * not {MessageSent} event SHOULD be expected. If `outboxId` is not zero, then further (gateway specific, and non
```
arr00, line 13:
```suggestion
     * no {MessageSent} event SHOULD be expected. If `outboxId` is not zero, then further (gateway specific, and non
```

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @dev This event is emitted when a message, for which the {MessageCreated} event contains an non zero `outboxId`,
+     * received the required post processing actions, and was thus sent to the destination chain.
+     */
+    event MessageSent(bytes32 indexed outboxId);
+
+    /// @dev This error is thrown when a message creation fails because of an unsupported attribute being specified.
+    error UnsuportedAttribute(bytes4 selector);
+
```
arr00, line 32:
```suggestion
    error UnsupportedAttribute(bytes4 selector);
```

Amxx:
> `Bytes.slice` is using `mcopy` in assembly so it works on 0.8.24, so want to confirm that we are upgrading to 0.8.25 for `mcopy` in code generation

Right, I misread the changelog, will go back to 0.8.24 for Bytes and all the code that depends on it directly or indirectly.

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+    /// @dev Getter to check weither an attribute is supported.
+    function supportsAttribute(bytes4 selector) external view returns (bool);
```
arr00, line 34:
```suggestion
    /// @dev Getter to check whether an attribute is supported.
```

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * @dev Endpoint for creating a new message. If the message requires further (gateway specific) processing before
+     * it can be sent to the destination chain, then a non-zero `outboxId` must be returned. Otherwise, this the
+     * message MUST be sent and this function must return 0.
+     *
+     * * MUST emit a {MessageCreated} event.
+     * * SHOULD NOT emit a {MessageSent} event.
+     *
+     * If any of the `attributes` is not supported, this function SHOULD revert with an {UnsuportedAttribute} error.
+     * Other errors SHOULD revert with errors not specified in ERC-7786.
```
arr00, line 45:
```suggestion
     * If any of the `attributes` is not supported, this function SHOULD revert with an {UnsupportedAttribute} error.
```

contracts/crosschain/draft-ERC7786Receiver.sol
Amxx: One answer is that any value should come with a corresponding attribute, and that is how it would be visible from the `processMessage` (instead of using msg.value)

Amxx: @frangio

contracts/crosschain/draft-ERC7786Receiver.sol
```diff
@@ -0,0 +1,60 @@
...
+        } else if (_isKnownGateway(gateway)) {
+            // Passive mode
+            IERC7786GatewayDestinationPassive(gateway).validateReceivedMessage(
```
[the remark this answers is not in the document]
Amxx: One answer is that any value should come with a corresponding attribute, and that is how it would be visible from the processMessage (instead of using msg.value)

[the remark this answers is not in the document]
Amxx: @frangio

contracts/interfaces/draft-IERC7786.sol
Amxx: not sure what the suggestion is here.

contracts/interfaces/draft-IERC7786.sol
Amxx: not sure what the suggestion is here.

arr00: Seems like something got borked with the VSCode plugin. Will fix sorry.

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+    );
+
+    /**
+     * @dev This event is emitted when a message, for which the {MessageCreated} event contains an non zero `outboxId`,
+     * received the required post processing actions, and was thus sent to the destination chain.
+     */
+    event MessageSent(bytes32 indexed outboxId);
+
+    /// @dev This error is thrown when a message creation fails because of an unsupported attribute being specified.
+    error UnsuportedAttribute(bytes4 selector);
```
arr00, line 31:
```suggestion
    error Unsupported(bytes4 selector);
```

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    /// @dev Getter to check weither an attribute is supported.
```
arr00, line 33:
```suggestion
    /// @dev Getter to check whether an attribute is supported.
```

.changeset/khaki-keys-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`ERC-7786`: Add interfaces for the ERC-7786 gateways and receiver, and a receiver base contract.
```
Amxx, line 5:
```suggestion
`draft-IERC7786` and `draft-ERC7786Receiver`: Add interfaces for the ERC-7786 gateways and receiver, and a receiver base contract.
```

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+ * @dev Interface for ERC-7786 destination gateways operating in passive mode.
+ *
+ * See ERC-7786 for more details
+ */
+interface IERC7786GatewayDestinationPassive {
+    error InvalidMessageKey(bytes messageKey);
+
+    /**
+     * @dev Endpoint for checking the validity of a message that is being relayed in passive mode. The message
+     * receiver is implicitelly the caller of this method, which guarantees that no-one but the receiver can
```
arr00, line 65:
```suggestion
     * receiver is implicitly the caller of this method, which guarantees that no-one but the receiver can
```

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * @dev Interface for the ERC-7786 client contracts (receiver).
+ *
+ * See ERC-7786 for more details
+ */
+interface IERC7786Receiver {
+    /**
+     * @dev Endpoint for receiving cross-chain message.
```
arr00, line 87: We should be consistent on `cross-chain` or `crosschain` in the docs

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    /// @dev This error is thrown when a message creation fails because of an unsupported attribute being specified.
+    error UnsupportedAttribute(bytes4 selector);
+
+    /// @dev Getter to check whether an attribute is supported or not.
+    function supportsAttribute(bytes4 selector) external view returns (bool);
+
+    /**
+     * @dev Endpoint for creating a new message. If the message requires further (gateway specific) processing before
+     * it can be sent to the destination chain, then a non-zero `outboxId` must be returned. Otherwise, this the
```
arr00, line 38:
```suggestion
     * it can be sent to the destination chain, then a non-zero `outboxId` must be returned. Otherwise, the
```

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+     * message MUST be sent and this function must return 0.
+     *
+     * * MUST emit a {MessageCreated} event.
+     * * SHOULD NOT emit a {MessageSent} event.
```
arr00, line 42: Why should it not emit a `MessageSent` event if above it says that when returning zero the message must be sent.

contracts/crosschain/draft-ERC7786Receiver.sol
frangio: Yeah I specified it as "SHOULD" so this implementation is compliant. But it's worth considering if it should be "MUST". I don't see any concrete issue with allowing any values and ignoring them.

contracts/crosschain/draft-ERC7786Receiver.sol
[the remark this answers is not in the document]
frangio: If we want to make payable compatible with passive mode we would have to add a `value` parameter to `validateReceivedMessage`.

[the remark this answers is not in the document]
Amxx: After discussion it feels like the thing is:

- Value should only be part of active mode calls.
- Passive mode should not support value, because value cannot be easily moved in a passive way
  - In that case we have a choice of doing `require(msg.value == 0);` in the passive mode, or to allow value to be passed by the relayer, without it being related to the semantics of the message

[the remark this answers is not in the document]
frangio: I would do `require(msg.value == 0)`. In terms of the spec I think this check should be recommended but not mandatory.

contracts/interfaces/draft-IERC7786.sol
```diff
@@ -0,0 +1,99 @@
...
+ * See ERC-7786 for more details
+ */
+interface IERC7786GatewaySource {
+    /**
+     * @dev Event emitted when a message is created. If `outboxId` is zero, no further processing is necessary, and
+     * no {MessageSent} event SHOULD be expected. If `outboxId` is not zero, then further (gateway specific, and non
+     * standardized) action is required.
+     */
+    event MessageCreated(
+        bytes32 outboxId,
```
frangio, line 17: I think this should be `indexed`. I know it's not in the spec but we should fix that.

contracts/interfaces/draft-IERC7786.sol
Amxx: That is part of the ERC specification. 

- If the message requires post processing, a unique id is give to that message, otherwize 0 is used
- The MessageSent is used when post processing is done, in that case we say which message was processed using that (non 0) id.
- if no processing is needed, the message is sent automatically when its created. emitting `MessageSent(0)` would be useless (that is implicit by 0 being in MessageCreated) but also confusing (id 0 is not unique)

contracts/interfaces/draft-IERC7786.sol
Amxx: I think we discussed it. 

The MessageSent one is indexed (because we may want to activelly filter a particular MessageSent once we know the id. But for the `MessageCreated` part, we don't have a way to predict the value (at least not in general) so creating a is not that meaningfull IMO (hence why its not indexed)

frangio: My reasoning was that if you see a `MessageSent` event you may want to retrieve the matching `MessageCreated`.

contracts/crosschain/draft-ERC7786Receiver.sol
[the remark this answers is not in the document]
Amxx: Added a check

contracts/interfaces/draft-IERC7786.sol
Amxx: modified to use `cross-chain` everywhere.

contracts/interfaces/draft-IERC7786.sol
Amxx: make sens

contracts/interfaces/draft-IERC7786.sol
arr00: Makes sense. Thanks for the clarification.

arr00 (approved):

PR #5258, opened 2024-10-15
Fuzz tampered tests for `ERC2771Forwarder`

### LINKED ISSUE #4896: ERC2771Forwarder has a flaky tampered signature test
**📝 Details**

In recent CI runs (e.g [this one](https://github.com/OpenZeppelin/openzeppelin-contracts/actions/runs/7884037744/job/21512184063)), the CI has failed because the ERC2771Forwarder tests use a signatured that's directly modified without any validation.

This seems to be causing a problem in `@noble`, the package used for secp256k1 operations:

```
Error: Cannot find square root
      at sqrtMod (node_modules/@noble/secp256k1/lib/index.js:642:15)
      at Function.fromCompressedHex (node_modules/@noble/secp256k1/lib/index.js:315:17)
      at Function.fromHex (node_modules/@noble/secp256k1/lib/index.js:344:25)
      at Function.fromSignature (node_modules/@noble/secp256k1/lib/index.js:364:25)
      at Object.recoverPublicKey (node_modules/@noble/secp256k1/lib/index.js:843:18)
      at Function.recoverPublicKey (node_modules/ethers/src.ts/crypto/signing-key.ts:182:34)
      at recoverAddress (node_modules/ethers/src.ts/transaction/address.ts:27:38)
      at Proxy.verifyTypedData (node_modules/ethers/src.ts/hash/typed-data.ts:614:26)
      at Context.<anonymous> (test/metatx/ERC2771Forwarder.test.js:139:32)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Copying stacktrace for the record.

cairoeth:
Fixes #4896 and helps #4894 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/metatx/ERC2771Forwarder.t.sol
+++ b/test/metatx/ERC2771Forwarder.t.sol
@@ -2,7 +2,7 @@
 
 pragma solidity ^0.8.20;
 
-import {Test, Vm} from "forge-std/Test.sol";
+import {Test} from "forge-std/Test.sol";
 import {ERC2771Forwarder} from "@openzeppelin/contracts/metatx/ERC2771Forwarder.sol";
 import {CallReceiverMockTrustingForwarder, CallReceiverMock} from "@openzeppelin/contracts/mocks/CallReceiverMock.sol";
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
@@ -48,9 +48,7 @@ contract ERC2771ForwarderTest is Test {
     CallReceiverMockTrustingForwarder internal _receiver;
 
     uint256 internal _signerPrivateKey = 0xA11CE;
-    uint256 internal _relayerPrivateKey = 0xB0B;
     address internal _signer = vm.addr(_signerPrivateKey);
-    address internal _relayer = vm.addr(_relayerPrivateKey);
 
     uint256 internal constant _MAX_ETHER = 10_000_000; // To avoid overflow
 
--- a/test/metatx/ERC2771Forwarder.t.sol
+++ b/test/metatx/ERC2771Forwarder.t.sol
@@ -2,7 +2,7 @@
 
 pragma solidity ^0.8.20;
 
-import {Test} from "forge-std/Test.sol";
+import {Test, Vm} from "forge-std/Test.sol";
 import {ERC2771Forwarder} from "@openzeppelin/contracts/metatx/ERC2771Forwarder.sol";
 import {CallReceiverMockTrustingForwarder, CallReceiverMock} from "@openzeppelin/contracts/mocks/CallReceiverMock.sol";
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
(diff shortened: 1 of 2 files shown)


test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -93,6 +96,68 @@ contract ERC2771ForwarderTest is Test {
...
+    ) private returns (ERC2771Forwarder.ForwardRequestData memory) {
+        return
+            ERC2771Forwarder.ForwardRequestData({
+                from: tamper == 0 ? vm.randomAddress() : request.from,
+                to: tamper == 1 ? vm.randomAddress() : request.to,
+                value: tamper == 2 ? vm.randomUint() : request.value,
+                gas: request.gas,
+                deadline: request.deadline,
+                data: tamper == 3 ? vm.randomBytes(4) : request.data,
+                signature: tamper == 4 ? vm.randomBytes(65) : request.signature
```
ernestognw, line 111: I guess we can define the `tamper` uint8 as an enum. This way I think the `bound` wouldn't be needed if the arguments are enums too.

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -162,4 +227,79 @@ contract ERC2771ForwarderTest is Test {
...
+        batchRequestDatas[1] = _tamperedValues(tamper, batchRequestDatas[1]);
+
+        vm.deal(address(this), batchRequestDatas[1].value);
+
+        vm.recordLogs();
+        _erc2771Forwarder.executeBatch{value: batchRequestDatas[1].value}(batchRequestDatas, payable(address(0xebe)));
+
+        Vm.Log[] memory entries = vm.getRecordedLogs();
+        assertEq(entries.length, 2);
+        assertEq(entries[1].topics[0], keccak256("ExecutedForwardRequest(address,uint256,bool)"));
```
ernestognw, line 303: Although this works, I feel there should be a cleaner way

ernestognw (commented):
Leaving small nits since I'm still not convinced 100% of the Foundry tests. They're still good to go in case you decide to merge before 5.2

test/metatx/ERC2771Forwarder.t.sol
cairoeth: not possible currently https://github.com/foundry-rs/foundry/issues/6623

test/metatx/ERC2771Forwarder.t.sol
cairoeth: i'm not aware of a cleaner because `vm.expectEmit` doesn't allow to catch not-expected events (i.e. it's able to catch the lack of events but not the opposite afaik)

cairoeth: fwiw this approach mimicks the behavior from the [hardhat tests](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5258/files#diff-29fad6b26849463f244fab4e002bfb9306c154b53ef8995d5a5d546de6eacc0fL352)

test/metatx/ERC2771Forwarder.t.sol
Amxx: I'd use an enum internally (as argument for `_tamperedValues` and `_tamperedExecute`)

Any in the public function we can use `TamperType(uint8(bound(tamper, 0, 4)))` to create that enum object

test/metatx/ERC2771Forwarder.t.sol
Amxx: The hardhat test you link looks like an old thing from the web3js times. If we were to (re)write this tests today we would use the `expect().to.not.emit()` syntax from the new helpers.

cairoeth: gotcha, replaced with a more concrete check https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5258/commits/598a09ba2f31416e23af1bb5be8db486cd526d54

Amxx (commented):
Why are `testExecuteBatchTamperedValuesZeroReceiver` and `testExecuteBatchTamperedValues` different?

One does `_tamperedValues` then expect a call, the other replace `batchRequestDatas` entirely and relies on the `expectRevert` in `tamperedExecute`.

I'd like to understand and clean all that, but I'm not able to run it locally :/

test/metatx/ERC2771Forwarder.t.sol
```diff
@@ -93,6 +104,67 @@ contract ERC2771ForwarderTest is Test {
...
+                to: tamper == TamperType.TO ? vm.randomAddress() : request.to,
+                value: tamper == TamperType.VALUE ? vm.randomUint() : request.value,
+                gas: request.gas,
+                deadline: request.deadline,
+                data: tamper == TamperType.DATA ? vm.randomBytes(4) : request.data,
+                signature: tamper == TamperType.SIGNATURE ? vm.randomBytes(65) : request.signature
+            });
+    }
+
+    function _tamperedExecute(
```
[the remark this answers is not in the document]
cairoeth: yeah i kinda meant for this function to generate a tampered execute, feel free to rename

Amxx:
@cairoeth I rewrote the test quite heavily. IMO that is more readable. Let me know what you think.

cairoeth:
looks great, thanks!

Amxx (approved):

PR #5272, opened 2024-10-21
Add NoncesKeyed variant

### LINKED ISSUE #5011: Semi abstracted nonce support
**🧐 Motivation**
[See this section of ERC-4337](https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support)

Relevant discussion #4425, #3927

**📝 Details**

ERC-4337 includes a semi abstracted nonce system. We should provide an abstract contract that implements it.

- Implement it now for ERC-4337 (and associated usecases)
- Evaluate replacing the current system with this one (as an internally breaking change) in v6

Amxx:
Fixes #5011

Name
- `NoncesSemiAbstracted`
- `SemiAbstractedNonces`
- `NoncesWithKeys`
- other ?

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/NoncesSemiAbstracted.sol
+++ b/contracts/utils/NoncesSemiAbstracted.sol
@@ -9,7 +9,7 @@ import {Nonces} from "./Nonces.sol";
  * Follows the https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support[ERC-4337's semi-abstracted nonce system].
  */
 abstract contract NoncesSemiAbstracted is Nonces {
-    mapping(address owner => mapping(uint192 key => uint64)) private _nounces;
+    mapping(address owner => mapping(uint192 key => uint64)) private _nonces;
 
     /**
      * @dev Returns the next unused nonce for an address.
@@ -22,7 +22,7 @@ abstract contract NoncesSemiAbstracted is Nonces {
      * @dev Returns the next unused nonce for an address and key. Result contains the key prefix.
      */
     function nonces(address owner, uint192 key) public view virtual returns (uint256) {
-        return (uint256(key) << 64) | _nounces[owner][key];
+        return (uint256(key) << 64) | _nonces[owner][key];
     }
 
     /**
@@ -40,8 +40,12 @@ abstract contract NoncesSemiAbstracted is Nonces {
      * Returns the current value and increments nonce.
      */
     function _useNonce(address owner, uint192 key) internal virtual returns (uint256) {
-        // TODO: use unchecked here? Do we expect 2**64 nonce ever be used for a single owner?
-        return _nounces[owner][key]++;
+        // For each account, the nonce has an initial value of 0, can only be incremented by one, and cannot be
+        // decremented or reset. This guarantees that the nonce never overflows.
+        unchecked {
+            // It is important to do x++ and not ++x here.
+            return _nonces[owner][key]++;
+        }
     }
 
     /**
--- a/contracts/utils/NoncesSemiAbstracted.sol
+++ b/contracts/utils/NoncesSemiAbstracted.sol
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {Nonces} from "./Nonces.sol";
+
+/**
+ * @dev Alternative to {Nonces}, that support key-ed nonces.
+ *
+ * Follows the https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support[ERC-4337's semi-abstracted nonce system].
+ */
+abstract contract NoncesSemiAbstracted is Nonces {
+    mapping(address owner => mapping(uint192 key => uint64)) private _nonce;
+
+    /**
+     * @dev Returns the next unused nonce for an address.
+     */
+    function nonces(address owner) public view virtual override returns (uint256) {
+        return nonces(owner, 0);
+    }
+
+    /**
+     * @dev Returns the next unused nonce for an address and key. Result contains the key prefix.
+     */
+    function nonces(address owner, uint192 key) public view virtual returns (uint256) {
+        return (uint256(key) << 64) | _nonce[owner][key];
+    }
+
+    /**
+     * @dev Consumes a nonce from the default key.
+     *
+     * Returns the current value and increments nonce.
+     */
+    function _useNonce(address owner) internal virtual override returns (uint256) {
+        return _useNonce(owner, 0);
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


arr00:
#5011

contracts/utils/NoncesSemiAbstracted.sol
```diff
@@ -0,0 +1,69 @@
...
+     */
+    function _useCheckedNonce(address owner, uint256 keyNonce) internal virtual override {
+        _useCheckedNonce(owner, uint192(keyNonce >> 64), uint64(keyNonce));
+    }
+
+    /**
+     * @dev Same as {_useNonce} but checking that `nonce` is the next valid for `owner`.
+     *
+     * This version takes a the key and the nonce as two different parameters.
+     */
```
arr00, line 62:
```suggestion
     *
     * This version takes the key and the nonce as two different parameters.
     */
```

contracts/utils/NoncesSemiAbstracted.sol
```diff
@@ -0,0 +1,69 @@
...
+    function _useNonce(address owner, uint192 key) internal virtual returns (uint256) {
+        // TODO: use unchecked here? Do we expect 2**64 nonce ever be used for a single owner?
+        return _nonce[owner][key]++;
+    }
+
+    /**
+     * @dev Same as {_useNonce} but checking that `nonce` is the next valid for `owner`.
+     *
+     * This version takes a the key and the nonce in a single uint256 parameter:
```
arr00, line 50:
```suggestion
     *
     * This version takes the key and the nonce in a single uint256 parameter:
```

contracts/utils/NoncesSemiAbstracted.sol
```diff
@@ -0,0 +1,69 @@
...
+
+import {Nonces} from "./Nonces.sol";
+
+/**
+ * @dev Alternative to {Nonces}, that support key-ed nonces.
+ *
+ * Follows the https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support[ERC-4337's semi-abstracted nonce system].
+ */
+abstract contract NoncesSemiAbstracted is Nonces {
+    mapping(address owner => mapping(uint192 key => uint64)) private _nonce;
```
arr00, line 12: Rename to `nonces` to be consistent with `Nonces`?

arr00 (approved):
Looks good and sufficiently tested. I do agree that the name `NoncesSemiAbstracted` is somewhat meaningless and would support a name change.

Amxx:
I used the name `NoncesSemiAbstracted` so that it is clear that it is a variant of `Nonces`.

Doing `SemiAbstractedNonces` would not have this effect as they next to each other when sorting alphabetically.

How do you feel about `NoncesWithKeys`? Any other proposal?

contracts/utils/NoncesSemiAbstracted.sol
```diff
@@ -0,0 +1,69 @@
...
+
+import {Nonces} from "./Nonces.sol";
+
+/**
+ * @dev Alternative to {Nonces}, that support key-ed nonces.
+ *
+ * Follows the https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support[ERC-4337's semi-abstracted nonce system].
+ */
+abstract contract NoncesSemiAbstracted is Nonces {
+    mapping(address owner => mapping(uint192 key => uint64)) private _nounces;
```
arr00, line 12:
```suggestion
    mapping(address owner => mapping(uint192 key => uint64)) private _nonces;
```

arr00 (approved):
I think `NoncesWithKeys` is more meaningful. I don't have a strong feeling either was since EF coined this as Semi Abstracted (whatever that means).

contracts/utils/NoncesSemiAbstracted.sol
```diff
@@ -0,0 +1,73 @@
...
+pragma solidity ^0.8.20;
+
+import {Nonces} from "./Nonces.sol";
+
+/**
+ * @dev Alternative to {Nonces}, that support key-ed nonces.
+ *
+ * Follows the https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support[ERC-4337's semi-abstracted nonce system].
+ */
+abstract contract NoncesSemiAbstracted is Nonces {
```
frangio, line 11: "Semi-abstracted" refers to the fact that ERC-4337 handles some of the nonce logic, i.e., it does not accept arbitrary nonce logic defined by accounts:
>This approach maintains the guarantee of UserOperation hash uniqueness on-chain on the protocol level while allowing wallets to implement any custom logic they may need operating on a 192-bit “key” field, while fitting the 32 byte word.

It does not refer to the key-sequence separation. A better name for this library might be `NoncesKeyed`.

frangio, line 11: This contract completely changes the logic and storage so I don't think it should inherit and override. It should work as a drop-in replacement if someone wants to swap out `Nonces` for this one.

```suggestion
abstract contract NoncesSemiAbstracted {
```

frangio, line 11: It seems risky to completely override the logic of `Nonces`.

One possible issue would be an upgrade that adds this contract to one that already had `Nonces`. In that case the value of `nonces(owner)` would reset.

A conservative implementation could reuse the built-in mapping in `Nonces` when the key is 0.

contracts/utils/NoncesSemiAbstracted.sol
```diff
@@ -0,0 +1,73 @@
...
+    mapping(address owner => mapping(uint192 key => uint64)) private _nonces;
+
+    /**
+     * @dev Returns the next unused nonce for an address.
```
frangio, line 15:
```suggestion
     * @dev Returns the next unused nonce for an address for key 0.
```

contracts/utils/NoncesSemiAbstracted.sol
frangio: The overrides for `nonces(address)`, `_useNonce(address)` could be removed since the logic would remain the same.

Amxx:
@frangio Do the last cahnges feel better ?

arr00 (approved):

frangio (approved):

frangio:
> I think `NoncesWithKeys` is more meaningful. I don't have a strong feeling either was since EF coined this as Semi Abstracted (whatever that means).

Just seeing that @arr00 had suggested another name. I don't have a strong preference between `NoncesWithKeys` and `NoncesKeyed`.

cairoeth (approved):

PR #5274, opened 2024-10-21
Merge account abstraction work into master

Amxx:


contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -0,0 +1,154 @@
...
+        return packValidationData(success, validAfter, validUntil);
+    }
+
+    /// @dev Returns the aggregator of the `validationData` and whether it is out of time range.
+    function getValidationData(uint256 validationData) internal view returns (address aggregator, bool outOfTimeRange) {
+        if (validationData == 0) {
+            return (address(0), false);
+        } else {
+            (address aggregator_, uint48 validAfter, uint48 validUntil) = parseValidationData(validationData);
+            return (aggregator_, block.timestamp > validUntil || block.timestamp < validAfter);
```
Amxx, line 74:
```suggestion
            return (aggregator_, block.timestamp < validAfter || validUntil < block.timestamp);
```

contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -0,0 +1,154 @@
...
+
+        bool success = aggregator1 == address(0) && aggregator2 == address(0);
+        uint48 validAfter = uint48(Math.max(validAfter1, validAfter2));
+        uint48 validUntil = uint48(Math.min(validUntil1, validUntil2));
+        return packValidationData(success, validAfter, validUntil);
+    }
+
+    /// @dev Returns the aggregator of the `validationData` and whether it is out of time range.
+    function getValidationData(uint256 validationData) internal view returns (address aggregator, bool outOfTimeRange) {
+        if (validationData == 0) {
```
Amxx, line 70: We should not need this branch.

contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -0,0 +1,154 @@
...
+        (address aggregator2, uint48 validAfter2, uint48 validUntil2) = parseValidationData(validationData2);
+
+        bool success = aggregator1 == address(0) && aggregator2 == address(0);
+        uint48 validAfter = uint48(Math.max(validAfter1, validAfter2));
+        uint48 validUntil = uint48(Math.min(validUntil1, validUntil2));
+        return packValidationData(success, validAfter, validUntil);
+    }
+
+    /// @dev Returns the aggregator of the `validationData` and whether it is out of time range.
+    function getValidationData(uint256 validationData) internal view returns (address aggregator, bool outOfTimeRange) {
```
Amxx, line 69: Would it be better to return an "in range" than an "out of range" boolean ?

contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -0,0 +1,150 @@
...
+            );
+    }
+
+    /**
+     * @dev Combines two validation data into a single one.
+     *
+     * The `aggregator` is set to {SIG_VALIDATION_SUCCESS} if both are successful, while
+     * the `validAfter` is the maximum and the `validUntil` is the minimum of both.
+     */
+    function combineValidationData(uint256 validationData1, uint256 validationData2) internal pure returns (uint256) {
```
Amxx, line 58: Do we need this function. AFAIK the original usecase is aggreagation in multisig accounts, but it may not be relevant anymore.

Feels quite opinionated. In particular, maybe if `aggregator1 == aggregator2` that should be ok.

ernestognw: Yes, I'd agree with removing it. Can we still do it now that is merged?

contracts/account/utils/draft-ERC4337Utils.sol
ernestognw: I'd agree with this change too. It feels weird the boolean is indicating an undesired state

PR #5280, opened 2024-10-25
Governor Sequential Proposal id

arr00:
Governance extension that enables sequential proposal ids. This implementation requires making `hashProposal` a write function which may not be acceptable.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -128,7 +128,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         uint256[] memory values,
         bytes[] memory calldatas,
         bytes32 descriptionHash
-    ) public pure virtual returns (uint256) {
+    ) public virtual returns (uint256) {
         return uint256(keccak256(abi.encode(targets, values, calldatas, descriptionHash)));
     }
 
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -110,6 +110,15 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         return "1";
     }
 
+    function _getProposalId(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) internal virtual returns (uint256) {
+        return hashProposal(targets, values, calldatas, descriptionHash);
+    }
+
     /**
      * @dev See {IGovernor-hashProposal}.
      *
@@ -128,7 +137,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         uint256[] memory values,
         bytes[] memory calldatas,
         bytes32 descriptionHash
-    ) public virtual returns (uint256) {
+    ) public pure virtual returns (uint256) {
         return uint256(keccak256(abi.encode(targets, values, calldatas, descriptionHash)));
     }
 
@@ -317,7 +326,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         string memory description,
         address proposer
     ) internal virtual returns (uint256 proposalId) {
-        proposalId = hashProposal(targets, values, calldatas, keccak256(bytes(description)));
+        proposalId = _getProposalId(targets, values, calldatas, keccak256(bytes(description)));
 
         if (targets.length != values.length || targets.length != calldatas.length || targets.length == 0) {
             revert GovernorInvalidProposalLength(targets.length, calldatas.length, values.length);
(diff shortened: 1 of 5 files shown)


arr00:
For reference, [here](https://github.com/OpenZeppelin/openzeppelin-contracts/compare/master...arr00:openzeppelin-contracts:feat/sequential-proposal-id-alt?expand=1) is a rough diff of what is required if we do not want to change the mutability to write... and this still requires internal vars. The approach implemented here seems like the best option.

contracts/governance/Governor.sol
```diff
@@ -128,7 +128,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         uint256[] memory values,
         bytes[] memory calldatas,
         bytes32 descriptionHash
-    ) public pure virtual returns (uint256) {
+    ) public virtual returns (uint256) {
```
Amxx, line 131: Not even view ... I get why its needed, but wow that is a change of behavior !

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,27 @@
...
+    ) public virtual override returns (uint256) {
+        uint256 proposalHash = super.hashProposal(targets, values, calldatas, descriptionHash);
+
+        uint256 storedProposalId = _proposalIds[proposalHash];
+        if (storedProposalId == 0) {
+            _proposalIds[proposalHash] = ++_numberOfProposals;
+            return _numberOfProposals;
+        }
+
+        return storedProposalId;
```
Amxx, line 25:
```suggestion
        return storedProposalId == 0
            ? (_proposalIds[proposalHash] = ++_numberOfProposals);
            : storedProposalId;
        }

        return storedProposalId;
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,24 @@
...
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256) {
+        uint256 proposalHash = super.hashProposal(targets, values, calldatas, descriptionHash);
+
+        uint256 storedProposalId = _proposalIds[proposalHash];
+        return storedProposalId == 0
+            ? (_proposalIds[proposalHash] = ++_numberOfProposals)
+            : storedProposalId;
+    }
```
wildmolasses, line 23: there may be a problem here; we don't want an external caller to call this, as in the case of a proposal that doesn't exist, it will increment `_numberOfProposals`

arr00: Yes this is a good callout thanks.

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,32 @@
...
+
+        uint256 storedProposalId = _proposalIds[proposalHash];
+        return storedProposalId == 0 ? (_proposalIds[proposalHash] = ++_numberOfProposals) : storedProposalId;
+    }
+
+    /**
+     * @dev Internal function to set the sequential proposal ID for the next proposal. This is helpful for transitioning from another governing system.
+     */
+    function _setNextProposalId(uint256 nextProposalId) internal virtual {
+        if (_numberOfProposals != 0) revert NextProposalIdCanOnlyBeSetOnce();
```
wildmolasses, line 29: IMO this works, but there's a slightly annoying attack surface here. if an attacker can create a proposal on this governor before the upgrade occurs, then any `setNextProposalId` call will revert. this `setNextProposalId` call might be part of the upgrade proposal, which would then fail.

we had the same problem. [we managed it](https://github.com/ScopeLift/compound-governance-upgrade/pull/43/files#diff-8dfda4c5f188c565ed922bb6417549df160fa6898ffd3b316c2b8ea8e888c4abR66-R72) by initializing `nextProposalId` to `type(uint256).max`. then, proposals couldn't be created unless `setNextProposalId` had been called.

arr00: Given that the default flow would be proposal ids starting from 1, we don't want to require calls to `_setNextProposalId`. Also note that `_setNextProposalId` is an internal function which would need to be called in a bespoke public function which is where this issue should be addressed. We should add another function to this extension `getNextProposalId` which would make disabling governance until the proposal id is set easier.

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,32 @@
...
+    function _getProposalId(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) internal virtual override returns (uint256) {
+        uint256 proposalHash = super._getProposalId(targets, values, calldatas, descriptionHash);
+
+        uint256 storedProposalId = _proposalIds[proposalHash];
+        return storedProposalId == 0 ? (_proposalIds[proposalHash] = ++_numberOfProposals) : storedProposalId;
```
frangio, line 22: My idea was that this function should be a view getter, it shouldn't automatically generate a new proposal id, and it should revert if there is no proposal id stored (because the proposal was never created).

arr00:
Closing in favor of #5290

PR #5281, opened 2024-10-25
Allow proposer to cancel proposal while pending, active, succeeded, and queued

### LINKED ISSUE #5260: Enable proposal cancelation by proposer after pending period
**🧐 Motivation**
By default, cancelation is only possible by the proposer during the pending period. There have been many times in governance processes that issues with proposals were discovered during the active or even successful/queued phases. Giving the proposer the ability to cancel during these periods reduces the risk of a buggy proposal being executed.

**📝 Details**
This should be default functionality as it is unlikely a team spinning up governance would explicitly select to allow this behavior--once they know they need it, it would be too late.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/fe6249ec2c2dcdbb05e2b6a0d0d3e229f6ffbb37/contracts/governance/Governor.sol#L468

arr00:
Fixes #5260 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/fair-cars-join.md
+++ b/.changeset/fair-cars-join.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`Governor`: Allow the proposer to cancel a proposal in more states. These states now include: pending, active, succeeded, and queued.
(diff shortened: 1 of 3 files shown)


arr00:
Closing in favor of #5301

PR #5282, opened 2024-10-25
Upgrade coverage dependency

### LINKED ISSUE #5231: Update `solidity-coverage`
### Description

On #5218 we're introducing a workaround to coverage issues in the coverage CI for P256. However, this workaround was deprecated in https://github.com/sc-forks/solidity-coverage/releases/tag/v0.8.7

We should update the library and drop the deprecated configuration. However, an attempt of doing so within the PR ended up in the compilation failing with multiple contracts generating unexpected warnings like:

```
Downloading compiler 0.8.24
Warning: Statement has no effect.
  --> contracts/utils/Context.sol:23:68:
   |
23 |     function _msgSender() internal view virtual returns (address) {abi.encode(0x730c7f785d17a153);  /* function */ 
   |                          
```

arr00:
Fixes #5231

Upgrade to `solidity-coverage` 0.8.13 which contains a fix for processing files compiled with IR route enabled. Ignore the known `Statement has no effect` warning.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.solcover.js
+++ b/.solcover.js
@@ -10,12 +10,4 @@ module.exports = {
     fgrep: '[skip-on-coverage]',
     invert: true,
   },
-  // Work around stack too deep for coverage
-  configureYulOptimizer: true,
-  solcOptimizerDetails: {
-    yul: true,
-    yulDetails: {
-      optimizerSteps: '',
-    },
-  },
 };
--- a/.solcover.js
+++ b/.solcover.js
@@ -7,6 +7,7 @@ module.exports = {
     default_balance_ether: '10000000000000000000000000',
   },
   mocha: {
+    enableTimeouts: false,
     fgrep: '[skip-on-coverage]',
     invert: true,
   },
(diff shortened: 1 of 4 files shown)


arr00:
There are other issues with running coverage now. Some tests fail due to time limits and gas issues.

cgewecke:
Hi @arr00, I help maintain solidity-coverage.

I think it was a mistake for solidity-coverage to deprecate the `configureYulOptimizer` and `solcOptimizerDetails` options. These settings are equivalent to forge's `ir-minimum` and they let the coverage plugin work pretty efficiently.

By contrast, the strategy solidity-coverage uses for projects that only compile with full IR is computationally expensive - it has to search for coverage traces across a much larger range of opcodes. It isn't scaling well for a test suite as large as this one.  

I think if coverage is working OK with the current config it's fine to leave it like that. You *could* upgrade to the latest version of the plugin and continue to use these options but there's no obvious benefit to that. 

Apologies for any confusion the coverage docs have caused, I wasn't aware the plugin's viaIR performance was scaling this badly until I looked into this PR.

arr00:
Appreciate the comment @cgewecke. We'll keep coverage as is for now, but feel free to ping me here if there are fixes regarding this in the future.

PR #5290, opened 2024-11-01
Add GovernorSequentialProposalId extension for sequential numbers on proposals

arr00:
Alternative solution to #5280 

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,35 @@
...
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        string memory description,
+        address proposer
+    ) internal virtual override returns (uint256) {
+        uint256 proposalHash = super.hashProposal(targets, values, calldatas, keccak256(bytes(description)));
+        _proposalIds[proposalHash] = ++_numberOfProposals;
+
+        return super._propose(targets, values, calldatas, description, proposer);
```
ernestognw, line 33: I feel this is a bit cleaner

```suggestion
        uint256 proposalId =  super._propose(targets, values, calldatas, description, proposer);
        _proposalIds[proposalId] = ++_numberOfProposals;
        return proposalId;
```

Otherwise, the `super` in hashProposal is not needed:

```suggestion
        uint256 proposalHash = hashProposal(targets, values, calldatas, keccak256(bytes(description)));
        _proposalIds[proposalHash] = ++_numberOfProposals;

        return super._propose(targets, values, calldatas, description, proposer);
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,46 @@
...
+    function getProposalId(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public view virtual override returns (uint256) {
+        uint256 proposalHash = hashProposal(targets, values, calldatas, descriptionHash);
+
+        uint256 storedProposalId = _proposalIds[proposalHash];
+        return storedProposalId == 0 ? (_proposalCount + 1) : storedProposalId;
```
Amxx, line 20: IMO returning `_proposalCount+1` is dangerous here. You might call that function, get this value, assume its good and store it ... only to realize that when it ends up being proposed it received a different id.

I think we should rather return an error.

```suggestion
        uint256 storedProposalId = _proposalIds[proposalHash];
        if (storedProposalId == 0) revert GovernorNonexistentProposal(0);
        return storedProposalId;
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,46 @@
...
+        uint256 proposalHash = hashProposal(targets, values, calldatas, keccak256(bytes(description)));
+        _proposalIds[proposalHash] = ++_proposalCount;
+
+        return super._propose(targets, values, calldatas, description, proposer);
+    }
+
+    /**
+     * @dev Internal function to set the sequential proposal ID for the next proposal. This is helpful for transitioning from another governing system.
+     */
+    function _setProposalCount(uint256 proposalCount) internal virtual {
```
Amxx, line 39: This function should probably check that the new value is an increase over the old value. That way, calling that by mistake will not result in double numbering
```suggestion
    function _setProposalCount(uint256 proposalCount) internal virtual {
        if (proposalCount < _proposalCount) revert GovernorSequentialProposalIdTooLow();
```

contracts/governance/Governor.sol
```diff
@@ -91,7 +91,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
      */
     function supportsInterface(bytes4 interfaceId) public view virtual override(IERC165, ERC165) returns (bool) {
         return
-            interfaceId == type(IGovernor).interfaceId ||
+            interfaceId == type(IGovernor).interfaceId ^ IGovernor.getProposalId.selector ||
```
Amxx, line 94: The changes to `IGovernor.sol` cause the interfaceId to change. 

This leaves us with 3 options:
- don't change the `supportsInterface` logic, which means the old interfaceId is no longer advertised through ERC-165, and we have a new one instead. That is somehow breaking, and tests need to be updated
- don't change the `IGovernor` interface. We can have the new function in `Governor` without it also being in the interface.
- do an xor here to continue exposing the old interfaceId.

Amxx: @ernestognw is in favor of #1.

I don't have strong opinions

@frangio WDYT?

frangio: For me the default choice is option 3, what we did during 4.x:

>do an xor here to continue exposing the old interfaceId.

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,46 @@
...
+        _proposalCount = proposalCount;
+    }
+
+    function getProposalCount() public view virtual returns (uint256) {
```
frangio, line 43: Might make sense to reuse GovernorBravo's function name/signature here?

https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/GovernorBravoInterfaces.sol#L84

arr00: Agree

contracts/governance/extensions/GovernorSequentialProposalId.sol
Amxx: @arr00 metionned that setting the `_proposalCount` to `type(uint256).max` would be a way to disable the propose workflow before configuration/migration is done (and `_proposalCount` is reset to a "correct value")

IMO this enabling/disabling should probably be achieved by changing the proposal threshold.

test/helpers/governance.js
```diff
@@ -35,12 +35,16 @@ class GovernorHelper {
...
 
-  get id() {
+  get hash() {
     return ethers.keccak256(
       ethers.AbiCoder.defaultAbiCoder().encode(['address[]', 'uint256[]', 'bytes[]', 'bytes32'], this.shortProposal),
     );
   }
 
+  get id() {
+    return this.governor.getProposalCount ? this.governor.getProposalId(...this.shortProposal) : this.hash;
```
Amxx, line 45:
```suggestion
    return this.governor.proposalCount ? this.governor.getProposalId(...this.shortProposal) : this.hash;
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
frangio: Just want to confirm that there aren't any other functions we could reuse.

arr00: Took a look to confirm--no others are relevant to this extension.

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,46 @@
...
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        string memory description,
+        address proposer
+    ) internal virtual override returns (uint256) {
+        uint256 proposalHash = hashProposal(targets, values, calldatas, keccak256(bytes(description)));
+        _proposalIds[proposalHash] = ++_proposalCount;
+
+        return super._propose(targets, values, calldatas, description, proposer);
```
frangio, line 33: This is changing `_propose` in a very significant way: it becomes possible to create multiple proposals with the same contents.

Did you consider this? Is it ok?

I don't have an opinion, but I don't know what other consequences this may have.

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * @dev Internal function to set the sequential proposal ID for the next proposal. This is helpful for transitioning from another governing system.
```
frangio, line 37: I would add a big caveat about decreasing the proposal count. I'd consider enforcing it in a require statement too, not sure if there are significant downsides to that.

contracts/governance/extensions/GovernorSequentialProposalId.sol
arr00: Just pushed a commit to patch this. It actually would be a critical bug where whenever you repropose the same proposal it gives it a new id and abandons the previous proposal.

frangio: Wouldn't both old and new proposals just continue to live side by side? In what sense would it be abandoned?

arr00: Proposals IDs are stored in a mapping of hash to int. So each hash can only be assigned to one proposal id. If you repropose the same proposal it would overwrite previously. As such, any functional calling getProposalId (such as queue and execute) would get the new proposal state and previous votes etc wouldn’t count.

.changeset/brave-islands-sparkle.md
19714, line 1: .changeset/brave-islands-sparkle.md

contracts/governance/extensions/GovernorSequentialProposalId.sol
Amxx: The use-case that @arr00 metionned that someone may set the count to type(uint256).max to (temporarily) disable proposing until the correct number is set. 

IMO that usecase (prevent any proposal from being submitted could be more cleanly done using the proposal threshold).

This is the require I proposed: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5290#discussion_r1846592233

arr00: Would like to get @wildmolasses thoughts here. I'm leaning towards pushing the above approach of just using the proposal threshold.

frangio: Yeah I'd keep that use case separate. But I do think it's a concern if setting the count to the max uint permanently blocks the governor from functioning... So this is why I'm not all in on enforcing that the update is an increase. A safer thing would be to allow increases of some reasonable size (like uint32), it seems arbitrary but I think it works.

contracts/governance/extensions/GovernorSequentialProposalId.sol
arr00: added a requirement that the proposal count is strictly increasing

contracts/governance/IGovernor.sol
```diff
@@ -212,6 +212,16 @@ interface IGovernor is IERC165, IERC6372 {
         bytes32 descriptionHash
     ) external pure returns (uint256);
 
+    /**
+     * @dev Function used to get the proposal id from the proposal details
```
ernestognw, line 216:
```suggestion
     * @dev Function used to get the proposal id from the proposal details.
```

.changeset/brave-islands-sparkle.md
ernestognw, line 1: I think we're missing another check to mention the new `getProposalId` function

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,63 @@
...
+     * May only call this function if the current proposal count is 0.
+     */
+    function _setProposalCount(uint256 newProposalCount) internal virtual {
+        if (_proposalCount != 0) {
+            revert GovernorCanNotSetProposalCount();
+        }
+        _proposalCount = newProposalCount;
+    }
+
+    function proposalCount() public view virtual returns (uint256) {
```
ernestognw, line 60: We need to document this getter. Also, I'd put it up before the `_propose` function

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,63 @@
...
+        uint256 proposalHash = hashProposal(targets, values, calldatas, keccak256(bytes(description)));
+        uint256 storedProposalId = _proposalIds[proposalHash];
+        if (storedProposalId == 0) {
+            _proposalIds[proposalHash] = ++_proposalCount;
+        }
+        return super._propose(targets, values, calldatas, description, proposer);
+    }
+
+    /**
+     * @dev Internal function to set the current proposal count--the next proposal id will be `newProposalCount` + 1.
```
ernestognw, line 49: I don't understand this sentence.

```suggestion
     * @dev Internal function to set the current proposal count--the next proposal id will be `newProposalCount` + 1.
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * May only call this function if the current proposal count is 0.
+     */
+    function _setProposalCount(uint256 newProposalCount) internal virtual {
```
ernestognw, line 53: I assume the purpose of including this function is to allow DAOs that are upgrading to set an initial proposal count in their initializers. If this is the case, I think this should be documented somewhere

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Governor} from "../Governor.sol";
+
+abstract contract GovernorSequentialProposalId is Governor {
```
ernestognw, line 7: We're missing NatSpec here

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,63 @@
...
+    ) public view virtual override returns (uint256) {
+        uint256 proposalHash = hashProposal(targets, values, calldatas, descriptionHash);
+        uint256 storedProposalId = _proposalIds[proposalHash];
+        if (storedProposalId == 0) {
+            revert GovernorNonexistentProposal(0);
+        }
+        return storedProposalId;
+    }
+
+    function _propose(
```
ernestognw, line 33: Missing NatSpec here too

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,63 @@
...
+import {Governor} from "../Governor.sol";
+
+abstract contract GovernorSequentialProposalId is Governor {
+    uint256 private _proposalCount;
+    mapping(uint256 proposalHash => uint256 proposalId) private _proposalIds;
+
+    /**
+     * @dev The proposal count may only be set if the current proposal count is 0 in {_setProposalCount}.
+     */
+    error GovernorCanNotSetProposalCount();
```
ernestognw, line 14: I'd rename it "uninitialized" since it better indicates what's wrong. Also, it goes well with the change I'm proposing to rename {_setProposalCount} to {_initializeProposalCount}. I think that instead of explaining the error literally, we should give a hint to solve it.

```suggestion
    /**
     * @dev The proposal count may only be set if the current proposal count is uninitialized.
     */
    error GovernorUninitializedProposalCount();
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
ernestognw, line 53: I would be in favor of changing this function's name to indicate better what to do with it. In this case, it suggests developers that it should be used within an initializer.

```suggestion
    function _initializeProposalCount(uint256 newProposalCount) internal virtual {
```

test/utils/introspection/SupportsInterface.behavior.js
```diff
@@ -87,6 +87,35 @@ const SIGNATURES = {
...
+    'propose(address[],uint256[],bytes[],string)',
+    'queue(address[],uint256[],bytes[],bytes32)',
+    'execute(address[],uint256[],bytes[],bytes32)',
+    'cancel(address[],uint256[],bytes[],bytes32)',
+    'castVote(uint256,uint8)',
+    'castVoteWithReason(uint256,uint8,string)',
+    'castVoteWithReasonAndParams(uint256,uint8,string,bytes)',
+    'castVoteBySig(uint256,uint8,address,bytes)',
+    'castVoteWithReasonAndParamsBySig(uint256,uint8,address,string,bytes,bytes)',
+  ],
```
ernestognw, line 118: I'm not really a fan of repeating all this interface, I'd look for a way to reuse it easily

contracts/governance/extensions/GovernorSequentialProposalId.sol
arr00: As of now the function is only usable when the proposal count is 0 (one time use), but if there are legitimate use-cases for using it beyond that we would open it up to allow any increase in proposal count. Keeping this phrasing enables that to be a non-breaking change.

test/utils/introspection/SupportsInterface.behavior.js
arr00: I updated it to reuse the interface.

contracts/governance/extensions/GovernorSequentialProposalId.sol
arr00: Could you suggest a change?

ernestognw: I forgot to come back and finish the suggestion. Now that I read it, it makes more sense.

I'd only suggest pointing to the definition of `proposalCount`. My confusion came because I thought the proposal count was an internal implementation detail. This should improve with some docs IMO.

```suggestion
     * @dev Internal function to set the value of {proposalCount}--the next proposal id will be `newProposalCount` + 1.
```

contracts/governance/IGovernor.sol
```diff
@@ -212,6 +212,16 @@ interface IGovernor is IERC165, IERC6372 {
         bytes32 descriptionHash
     ) external pure returns (uint256);
 
+    /**
+     * @dev Function used to get the proposal id from the proposal details.
```
Amxx, line 216:
```suggestion
     * @notice module:core
     * @dev Function used to get the proposal id from the proposal details.
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
Amxx: then the error would be `GovernorAlreadyInitializedProposalCount` ?

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @dev Returns the current proposal count.
+     */
+    function proposalCount() public view virtual returns (uint256) {
+        return _proposalCount;
+    }
+
+    /**
+     * @dev Hook into the proposing mechanism to increment proposal count.
```
Amxx, line 41:
```suggestion
     * @dev See {IGovernor-_propose}. 
     * Hook into the proposing mechanism to increment proposal count.
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
Amxx: I'm not a fan of the "Internal function to set the value of {proposalCount}" variant, because it disregard the parameters of `proposalCount`. Also, `proposalCount` reverts if the details don't match an existing proposal.

Therefore I think the current formulation is more correct.

test/governance/extensions/GovernorStorage.test.js
```diff
@@ -101,7 +101,7 @@ describe('GovernorStorage', function () {
           expect(await this.mock.proposalCount()).to.equal(1n);
 
           expect(await this.mock.proposalDetailsAt(0n)).to.deep.equal([
-            this.proposal.id,
+            await this.proposal.id,
```
Amxx, line 104: AFAIK, the mock used here is not using sequential id. Therefore, `this.proposal.id` is not returning a promise, so there is no need to await it.

Amxx:
```suggestion
            this.proposal.id,
```

contracts/governance/extensions/GovernorSequentialProposalId.sol
Amxx:
```suggestion
/**
 * @dev Extension of {Governor} that changes the numbering of proposals from the default hash-based approach to 
 * sequential ids.
 */
abstract contract GovernorSequentialProposalId is Governor {
```

test/governance/extensions/GovernorStorage.test.js
arr00: The test fails without the await. I'm not sure why but it seems like within the chai deep equ it doesn't resolve without await.

contracts/governance/extensions/GovernorSequentialProposalId.sol
arr00: Yes good point

contracts/governance/extensions/GovernorSequentialProposalId.sol
arr00: You've swayed me. Will change over

test/governance/extensions/GovernorStorage.test.js
Amxx: I figured it out. `GovernorStorage` contains a `proposalCount()` function. This is seen by the helper as a sign that we are using the `GovernorSequentialProposalId` module when we are not. So the issue is in the helper hability to detect `GovernorSequentialProposalId`

arr00: Now that `GovernorStorage` is mentioned, I took a look at how these extensions would work together and it would be a bit odd (not necessarily a security issue though). Should we try to make them more compatible or that not necessary? An example of an oddity is that the end developer would have to decide which `proposalCount` to expose which would not be equal if the set proposal count feature is used.

frangio: > An example of an oddity is that the end developer would have to decide which `proposalCount` to expose which would not be equal if the set proposal count feature is used.

Yes this is a problem...

frangio: Honestly the easiest thing would be to rename `proposalCount` in this PR so it doesn't conflict. For example, `nextProposalId`.

Amxx: In both cases `proposalCount` is here to replicate GovernorBravo's interface (that includes `proposalCount`)

arr00: Renaming the function to `nextProposalId` would actually be more accurate than `proposalCount` in this situation given that the proposal id can be offset initially. I think diverging from the `GovernorBravo` interface is fine here.

contracts/governance/extensions/GovernorSequentialProposalId.sol
```diff
@@ -0,0 +1,83 @@
...
+pragma solidity ^0.8.20;
+
+import {Governor} from "../Governor.sol";
+
+/**
+ * @dev Extension of {Governor} that changes the numbering of proposal ids from the default hash-based approach to
+ * sequential ids.
+ */
+abstract contract GovernorSequentialProposalId is Governor {
+    uint256 private _nextProposalId = 1;
```
Amxx, line 12: I'm not a fan of storage variables that have to be initialized in the constructor. This means upgradeable version need an initializer or a re-initializer.

IMO we should use `_lastProposalId` instead. We can then adapt `nextProposalId()` to return `_lastProposalId+1` if we want.

ernestognw (commented):
Almost there, I agree with @Amxx that we should express the `nextProposalId` in terms of a `lastProposalId` to avoid initializing a variable unnecessarily during construction

test/governance/extensions/GovernorSequentialProposalId.test.js
```diff
@@ -0,0 +1,202 @@
...
+      });
+
+      it('can only initialize proposal count from 0', async function () {
+        await this.helper.propose();
+        expect(this.mock.latestProposalId()).to.eventually.equal(1);
+        await expect(this.mock.$_initializeLatestProposalId(2)).to.be.revertedWithCustomError(
+          this.mock,
+          'GovernorAlreadyInitializedLatestProposalId',
+        );
+      });
```
ernestognw, line 164: We need to update this test. Also, we're missing a test for the `GovernorInvalidLatestProposalId` error

arr00: Yes good point. I actually prefer to remove the check for `if (newLatestProposalId == 0)` since its not fully necessary anymore. It was added when there was the nextProposalId concept and reducing it would cause issues. Here, the do nothing scenario is setting to 0 which isn't necessarily an issue.

ernestognw (approved):
Alright, LGTM

(I'm sure we'll go over another 40 comments after this 😂 )

PR #5293, opened 2024-11-05
`memory` --> `calldata` for read-only function params

### LINKED ISSUE #5291: Use `calldata` instead of `memory`
Shouldn't there be `calldata` type instead of `memory` as we are only reading its value?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/448efeea6640bbbc09373f03fbc9c88e280147ba/contracts/utils/cryptography/ECDSA.sol#L91

Found some more in [cryptography/](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography) folder. Fixed in PR #5293 

```[tasklist]
### Tasks
- [x] Add a PR (if the issue is accepted)
```

abhi3700:
Fixes #5291


abhi3700:
The tests are failing because of following error.

```sh
Downloading compiler 0.8.24
TypeError: Member "recover" not found or not visible after argument-dependent lookup in type(library ECDSA).
  --> contracts/mocks/EIP712Verifier.sol:13:35:
   |
13 |         address recoveredSigner = ECDSA.recover(digest, signature);
   |                                   ^^^^^^^^^^^^^


Error HH600: Compilation failed
```
It doesn't make sense to me as in my VSCode, the method is valid. I am not sure what am I missing 🤔 .
image

Amxx (changes requested):
There is a good reason why we use memory instead of calldata here: We don't know how these function will be called or how they will be overidden.

For example, in `contracts/utils/cryptography/MessageHashUtils.sol`, what makes you think the `message` and `data` are in calldata? In most cases, this data will be produced onchain, using some variant of `abi.encode`.

Similarly for `contracts/utils/cryptography/RSA.sol`. The data buffer is likelly to live in memory. Similarly, the public key part (`e` and `n`) are likelly to be in some form of storage.

For these reason I would strongly oppose forcing the use of calldata when no alternative exist, and were there is any doubts about where the value may come from.

Issue #4761, opened 2023-11-27
Implement ERC-7540


Amxx:
[ERC-7540](https://eips.ethereum.org/EIPS/eip-7540) is in the review phase, and looks like a good candidate for inclusion.

mw2000:
@Amxx I'd love to help with this

Amxx:
> I'm still unsure about how to approach overriding deposit, redeem, withdraw, and mint functions.

This is exactly why this issue needs to be discussed and researched BEFORE we make any PR. As the tag shows, this is this only an idea. We may not even do that at all. 

One thing we need to figure out is the process for going from `pendingDepositRequest` to `maxDeposit`. Because the redeem should `deposit` should consume the `maxDeposit`, NOT the `pendingDepositRequest`.

Please do not rush to propose implementations. The ERC is still in draft, and we need to hear from (multiple) potential users before we start designing an actual solution.

0xTimepunk:
Any updates on this?

mw2000:
@Amxx @0xTimepunk more than happy to pickup where we left off on this PR

alphastorm:
The ERC was finalized two days ago 🎉 

<img width="627" alt="Screenshot 2024-06-28 at 12 37 48 PM" src="https://github.com/OpenZeppelin/openzeppelin-contracts/assets/972382/58fae95b-5fdf-4ba2-bb47-38d6dc856ef1">

cc @Joeysantoro

yorhodes:
@alphastorm @Joeysantoro @mw2000 
I am very interested in a reference implementation for this which composes with the rest of the OpenZeppelin ERC20/4626 lifecycle hooks

kevin-pv01:
We'd be interested in using 7540 if there was an OZ reference implementation.
Do we know if any work on this is planned?

TalCrypto:
Any update on this implementation?
I represent logarithm.fi and we are building an asynchronous redeem vault now.
Would be happy if there were some OZ references.

zjesko:
same, would love to have a OZ reference (or any other) for buildling trading vaults

mw2000:
@zjesko @yorhodes @kevin-pv01 
 
There's a reference implementation in this repo (thanks to @TalCrypto for mentioning)  -> https://github.com/ERC4626-Alliance/ERC-7540-Reference

[...]

PR #5294, opened 2024-11-08
ERC7540 design proposal

hieronx:
Implementation proposal for discussion of #4761 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/interfaces/IERC7540.sol
+++ b/contracts/interfaces/IERC7540.sol
@@ -1,0 +1,140 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.0.0) (interfaces/IERC7540.sol)
+
+pragma solidity ^0.8.20;
+
+interface IERC7540Operator {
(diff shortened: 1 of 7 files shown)


ernestognw (commented):
I'll stop here. After thoroughly reviewing the PR and looking at the forum discussion I'm not convinced this could be a contract in the library. As noted in the Ethereum Magicians forum, the initial design didn't intend the contract to be a ["set in stone OpenZeppelin Contract"](https://ethereum-magicians.org/t/eip-7540-asynchronous-erc-4626-tokenized-vaults/16153/9), which I agree with.

As shown in the review, the implementation breaks in a couple ways around the library and includes some antipatterns that, although well-intentioned, become contradictory when trying to pursue an extendable implementation.

Amxx:
Implemented in https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/228

Will eventually move back to this repo.

PR #5303, opened 2024-11-19
Add a governor extension that implements a proposal guardian

### LINKED ISSUE #5301: Governor cancel guardian feature
**🧐 Motivation**
Cancellation needs to be expanded from the current minimal publicly exposed function on governor.

**📝 Details**
Enable the `_cancel` function to be called by this cancel guardian (set by governor) at any point. If the guardian is not set, allow the proposer to cancel at any time (permitted by `_cancel`). The proposer maintains the ability to cancel during the pending period during all cases.

Amxx:
Fixes #5301

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/governance/extensions/GovernorSecurityCouncil.sol
```diff
@@ -0,0 +1,76 @@
...
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256) {
+        address caller = _msgSender();
+        address authority = council();
+
+        if (authority == address(0)) {
+            // if there is no council
+            // ... only the proposer can cancel
+            // ... no restriction on when the proposer can cancel
+            uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
```
Amxx, line 36: To do when #5290 is merged:

```suggestion
            uint256 proposalId = getProposalId(targets, values, calldatas, descriptionHash);
```

arr00:
Looks good other than some comments. Would it make sense to call it something like `GovernorCancelCouncil` to be more explicit? I could imagine daos having multiple different "security" related councils.

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,75 @@
...
+     * * if the proposal guardian is set, proposer keep their ability to cancel during the pending stage (default behavior).
+     */
+    function cancel(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256) {
+        address caller = _msgSender();
+        address authority = proposalGuardian();
```
ernestognw, line 29: The `authority` is a concept we already use in the context of an AccessManager, it's borrowed from the original [DSAuthority from Dappsys](https://github.com/dapphub/ds-auth/blob/master/src/auth.sol#L17C11-L21). I'd rename this variable to `guardian` to avoid confusion

```suggestion
        address guardian = proposalGuardian();
```

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,75 @@
...
+
+        if (authority == address(0)) {
```
ernestognw, line 31:
```suggestion
        if (guardian == address(0)) {
```

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,75 @@
...
+            // if there is no proposal guardian
+            // ... only the proposer can cancel
+            // ... no restriction on when the proposer can cancel
+            uint256 proposalId = getProposalId(targets, values, calldatas, descriptionHash);
+            address proposer = proposalProposer(proposalId);
+            if (caller != proposer) revert GovernorOnlyProposer(caller);
+            return _cancel(targets, values, calldatas, descriptionHash);
+        } else if (authority == caller) {
```
ernestognw, line 39:
```suggestion
        } else if (guardian == caller) {
```

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {Governor} from "../Governor.sol";
+
+/**
+ * @dev Extension of {Governor} which adds a proposal guardian that can cancel proposals at any stage of their lifecycle.
+ *
+ * Note: if the proposal guardian is not configured, then proposers take this role for their proposals.
```
ernestognw, line 9: `NOTE:` renders in the docs, `Note:` doesn't :(

```suggestion
 * NOTE: if the proposal guardian is not configured, then proposers take this role for their proposals.
```

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,75 @@
...
+ */
+abstract contract GovernorProposalGuardian is Governor {
+    address private _proposalGuardian;
+
+    event ProposalGuardianSet(address oldProposalGuardian, address newProposalGuardian);
+
+    /**
+     * @dev Override {IGovernor-cancel} that implements the extended cancellation logic.
+     * * proposal guardian can cancel any proposal at any point in the lifecycle.
+     * * if no proposal guardian is set, proposer can cancel their proposals at any point in the lifecycle.
```
ernestognw, line 19:
```suggestion
     * * if no proposal guardian is set, the {proposalProposer} can cancel their proposals at any point in the lifecycle.
```

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,75 @@
...
+        address authority = proposalGuardian();
+
+        if (authority == address(0)) {
+            // if there is no proposal guardian
+            // ... only the proposer can cancel
+            // ... no restriction on when the proposer can cancel
+            uint256 proposalId = getProposalId(targets, values, calldatas, descriptionHash);
+            address proposer = proposalProposer(proposalId);
+            if (caller != proposer) revert GovernorOnlyProposer(caller);
+            return _cancel(targets, values, calldatas, descriptionHash);
```
ernestognw, line 38: The default behavior in Governor is to only allow the proposer to cancel when the proposal is still pending. I don't think we want to allow proposers to cancel after that.

Also, if we want to keep the default behavior, wouldn't it be better to just?:

```suggestion
            return super.cancel(targets, values, calldatas, descriptionHash);
```

This would simplify this function imo

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,75 @@
...
+abstract contract GovernorProposalGuardian is Governor {
+    address private _proposalGuardian;
+
+    event ProposalGuardianSet(address oldProposalGuardian, address newProposalGuardian);
+
+    /**
+     * @dev Override {IGovernor-cancel} that implements the extended cancellation logic.
+     * * proposal guardian can cancel any proposal at any point in the lifecycle.
+     * * if no proposal guardian is set, proposer can cancel their proposals at any point in the lifecycle.
+     * * if the proposal guardian is set, proposer keep their ability to cancel during the pending stage (default behavior).
```
ernestognw, line 20: Although the default behavior in the Governor is that a proposal can be canceled by the proposer, I would argue the default behavior is to call `super`. The difference is that `super` may call another override up in the inheritance chain.

```suggestion
     * * if the proposal guardian is set, the {proposalProposer} keeps their default rights defined in {IGovernor-cancel} (calling `super`).
```

ernestognw (commented):
What's the rationale to allow a proposer to cancel a proposal at any time?

I think the contract would be simpler if it just focus on allowing the guardian to cancel at any time and otherwise fallback to their original behavior with `super`.

arr00:
> What's the rationale to allow a proposer to cancel a proposal at any time?
> 
> I think the contract would be simpler if it just focus on allowing the guardian to cancel at any time and otherwise fallback to their original behavior with `super`.

The inspiration for this PR is in #5260 which was then replaced by #5301. There is some context missing but will try to give a TLDR.
- The assumption that governance correctly identifies all faulty proposals is flawed.
  - There are times that a proposal needs to be created strategically but then need to be cancelled. 
  - Proposals very often have to be cancelled after the pending period (see compound governance)
- Compound governance allows cancellation by proposers at any point which is what #5260 advocated for
- @Amxx pointed out that this gives a possibly exploitive permission to the proposer over the community as they could manipulate governance by cancelling at any point.
  - Say governance wants to do A but proposer X wants A to be delayed by a week. Proposer X would frontrun another proposer to propose A so other community members wouldn't propose. Proposer X cancels proposal A right before execution, requiring a whole new proposal process.
- While I don't know of any occurrences of the above, the possibility makes it preferred that the cancellation comes from a trusted community proposal guardian (hopefully a multisig)
  - If governance does not actually set the cancel guardian, it is definitely preferred to pass this authority over to the proposers.

ernestognw:
> @Amxx pointed out that this gives a possibly exploitive permission to the proposer over the community as they could manipulate governance by cancelling at any point.

Right, this was my interpretation as well. I also don't have any concrete example of such an exploit.

Getting back to the whole idea of enhancing cancellation capabilities, one concern I have with this design is the pattern:

```solidity
if (...)  {
  ...
  _cancel(...);
} else if (...) {
  ...
  _cancel(...);
} else {
  super.cancel();
}
```

The reasoning is that we're allowing to bypass `cancel` side effects. By not calling `super` in certain situations (i.e. whether the caller is the proposer or the guardian).

Consider a contract that inherits from `GovernorProposalGuardian`, but also inherits from an abstract contract that counts how many cancellations have been made (e.g. `GovernorCancellationCounter`), so that:

```solidity
contract GovernorCancellationCounter {
  uint256 cancellations;

  function cancel(...) public override virtual returns (uint256) {
      cancellations++;
      super.cancel(...);
    }
}

contract MyGovernor is ..., GovernorCancellationCounter, GovernorProposalGuardian { ... } // Order is important
```

In this case, the count would be missing the internal branches that don't call `super`. Although it's easier to override `_cancel`, I would say there may be legitimate reasons to prefer `cancel`.

I'd feel more comfortable if `Governor` implements an internal:

```solidity
function _validateCancel(...) internal virtual {
    // Current logic
    uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
    _validateStateBitmap(proposalId, _encodeStateBitmap(ProposalState.Pending));
    if (_msgSender() != proposalProposer(proposalId)) {
        revert GovernorOnlyProposer(_msgSender());
    }
}

...


function cancel(...) public virtual returns (uint256) {
  _validateCancel();
  return _cancel(targets, values, calldatas, descriptionHash);
}
```

I thin this way users can override `_validateCancel` and we can use it in `GovernorProposalGuardian` to implement the newer branches without missing the `super.cancel`.

contracts/governance/extensions/GovernorProposalGuardian.sol
Amxx: `return super.cancel(targets, values, calldatas, descriptionHash);` is already what in the last case ... so what you are proposing would be simplified further by doing

```solidity
if (guardian == caller) {
            // if there is a proposal guardian, and the caller is the proposal guardian
            // ... just cancel
            return _cancel(targets, values, calldatas, descriptionHash);
} else {
            // The caller is not the proposal guardian
            // ... apply default behavior
            return super.cancel(targets, values, calldatas, descriptionHash);
}            
```

I'd like to understand what this is not what we have right now.

ernestognw: I shared a proposal down below. My main concern is that we're missing supper in the branches where there's no guardian and when there's a guardian.

For example, in this code, users could override `_validateCancel` and call super to 1) add the guardian behavior and 2) extend the canceler permissions:

```solidity
function _validateCancel(...) internal virtual {
    // Current logic
    uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
    _validateStateBitmap(proposalId, _encodeStateBitmap(ProposalState.Pending));
    if (_msgSender() != proposalProposer(proposalId)) {
        revert GovernorOnlyProposer(_msgSender());
    }
}

...


function cancel(...) public virtual returns (uint256) {
  _validateCancel();
  return _cancel(targets, values, calldatas, descriptionHash);
}
```

arr00: In this solution, we will not call super for `_validateCancel`, which may also skip side effects. I don't see how it's materially better and I'm weary of adding another internal function.

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,67 @@
...
+        } else if (guardian == caller) {
+            // if there is a proposal guardian, and the caller is the proposal guardian
+            // ... just cancel
+            return true;
+        }
+
+        // if there is no guardian and the caller isn't the proposer or
+        // there is a proposal guardian, and the caller is not the proposal guardian
+        // ... apply default behavior
+        return super._validateCancel(proposalId, caller);
```
Amxx, line 65: The natspec are already pretty clear, I think a simple "or" is shoud be good enough here.
```suggestion
        return guardian == caller
            || (guardian == address(0) && caller == proposalProposer(proposalId))
            || super._validateCancel(proposalId, caller);
```

Amxx: Note: Before anyone asks, I don't think we need to check that `guardian != address(0)` in the case where `guardian == caller`. IMO caller should never be address(0).

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * Emits a {ProposalGuardianSet} event.
+     */
+    function _setProposalGuardian(address newProposalGuardian) internal virtual {
+        emit ProposalGuardianSet(_proposalGuardian, newProposalGuardian);
+        _proposalGuardian = newProposalGuardian;
+    }
+
+    /**
+     * @dev Override `_validateCancel` to implement the extended cancellation logic.
```
Amxx, line 43:
```suggestion
     * @dev Override {Governor-_validateCancel} to implement the extended cancellation logic.
```

contracts/governance/extensions/GovernorProposalGuardian.sol
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * * The {proposalGuardian} can cancel any proposal at any point.
+     * * If no proposal guardian is set, the {IGovernor-proposalProposer} can cancel their proposals at any point.
+     * * All other conditions are forwarded to super for default cancellation validation (as defined in {Governor-_validateCancel}).
```
Amxx, line 47:
```suggestion
     * * In any case, permissions defined in {Governor-_validateCancel} (or another override) remains valid.
```

test/governance/extensions/GovernorProposalGuardian.test.js
```diff
@@ -0,0 +1,124 @@
...
+            .to.be.revertedWithCustomError(this.mock, 'GovernorOnlyExecutor')
+            .withArgs(this.owner.address);
+        });
+      });
+
+      describe('cancel proposal during active state', function () {
+        beforeEach(async function () {
+          await this.helper.connect(this.proposer).propose();
+          await this.helper.waitForSnapshot(1n);
+          expect(this.mock.state(this.proposal.id)).to.eventually.equal(ProposalState.Active);
```
Amxx, line 95:
```suggestion
          await expect(this.mock.state(this.proposal.id)).to.eventually.equal(ProposalState.Active);
```

test/governance/extensions/GovernorProposalGuardian.test.js
```diff
@@ -0,0 +1,124 @@
...
+          ],
+          '<proposal description>',
+        );
+      });
+
+      it('deployment check', async function () {
+        expect(await this.mock.name()).to.equal(name);
+        expect(await this.mock.token()).to.equal(this.token);
+        expect(await this.mock.votingDelay()).to.equal(votingDelay);
+        expect(await this.mock.votingPeriod()).to.equal(votingPeriod);
```
Amxx, line 72:
```suggestion
        await expect(this.mock.name()).to.eventually.equal(name);
        await expect(this.mock.token()).to.equal(eventually.this.token);
        await expect(this.mock.votingDelay()).to.eventually.equal(votingDelay);
        await expect(this.mock.votingPeriod()).to.eventually.equal(votingPeriod);
```

arr00:
Should the proposal guardian have the ability to give up this role? I could see an argument in either direction but think it's worth a brief discussion before merging.

For: 
The proposal guardian has a critical role in the governance system--they need to catch potentially broken proposals. These proposals can cause significant issues if they go unmediated. The designers of the governance system explicitly defined a proposal guardian functionality so they want these types of proposals to be cancelled. What if for some reason, the proposal guardian is no longer able to act in the role? They need to be able to transition this role to proposers until governance has time to update the role to a new address.

Against:
Proposals should belong to the community after the pending period--if a malicious actor has the ability to cancel a proposal at any time, they may do so at the last second and stall governance. The proposal guardian can signal to governance that they would like to step down and governance should swap them out in a timely manner.

I think I lean towards for since the frequency of broken proposals out in the wild is quite high. If the proposal guardian isn't able to act effectively, the DAO needs the proposers to have this ability.

Amxx:
When there is a guardian, proposer still have the ability to cancel "early".
Additionally, if the guardian is unable to act, malicious proposal can still be "blocked" by voters saying no.

IMO:
- if this feature is present, "disabling" it is possible through an override that always revert, but the function remains present in the ABI (and bytecode)
- if this feature is not present, its quite easy to add
```solidity
function transferProposalGuardian(address newProposalGuardian) public {
    require(_msgSender == proposalGuardian());
    _setProposalGuardian(newProposalGuardian);
}
```

When in doubts, or when we think not everyone needs something, I'm always in favor of not forving the feature. In that particular case, the Governor is already close to the limit en terms of bytecode size, so I would avoid increassing it for something users may not want/need.

ernestognw (approved):
LGTM

arr00 (approved):

PR #5306, opened 2024-11-19
Document VotesExtended assumptions

Amxx:
Fixes L-02


contracts/governance/utils/VotesExtended.sol
```diff
@@ -7,6 +7,28 @@ import {SafeCast} from "../../utils/math/SafeCast.sol";
 
 /**
  * @dev Extension of {Votes} that adds exposes checkpoints for delegations and balances.
+ *
+ * WARNING: While this contract extends {Votes}, it makes additional assumption. Valid usage of {Votes} may not work
```
Amxx, line 11:
```suggestion
 * WARNING: While this contract extends {Votes}, it makes additional assumptions. Valid uses of {Votes} may not work
```

contracts/governance/utils/VotesExtended.sol
```diff
@@ -7,6 +7,28 @@ import {SafeCast} from "../../utils/math/SafeCast.sol";
 
 /**
  * @dev Extension of {Votes} that adds exposes checkpoints for delegations and balances.
+ *
+ * WARNING: While this contract extends {Votes}, it makes additional assumptions. Valid uses of {Votes} may not be
+ * compatbile with {VotesExtended} without additional considerations. The implementation of {_transferVotingUnits}
```
Amxx, line 12:
```suggestion
 * compatible with {VotesExtended} without additional considerations. The implementation of {_transferVotingUnits}
```

contracts/governance/utils/VotesExtended.sol
```diff
@@ -7,6 +7,28 @@ import {SafeCast} from "../../utils/math/SafeCast.sol";
...
+ * expects to run AFTER the voting weight movement is registered in a way that would reflect on {_getVotingUnits}.
+ *
+ * Said differently, {VotesExtended} MUST be integrated in a way such that calls {_transferVotingUnits} AFTER the 
+ * asset transfer is registered adn balances are updated:
```
Amxx, line 16:
```suggestion
 * asset transfer is registered and balances are updated:
```

contracts/governance/utils/VotesExtended.sol
```diff
@@ -7,6 +7,28 @@ import {SafeCast} from "../../utils/math/SafeCast.sol";
 
 /**
  * @dev Extension of {Votes} that adds exposes checkpoints for delegations and balances.
+ *
+ * WARNING: While this contract extends {Votes}, it makes additional assumptions. Valid uses of {Votes} may not be
+ * compatible with {VotesExtended} without additional considerations. The implementation of {_transferVotingUnits}
+ * expects to run AFTER the voting weight movement is registered in a way that would reflect on {_getVotingUnits}.
```
ernestognw, line 13: I would simplify it to:

```suggestion
* WARNING: While this contract extends from {Votes}, the implementation of {_transferVotingUnits} must run 
* after any state change that affects the result of {_getVotingUnits}. Consider that the order of execution may
* cause inconsistent unit tracking.
```

Amxx: "extends from {Votes}" doesn't sound right to me. We should get a native english speaker to double check. @arr00

ernestognw: I'm fine with "extends from" too

arr00: WARNING: While this contract extends {Votes}, valid uses of {Votes} may not be compatible with {VotesExtended}, without additional considerations. This implementation of {_transferVotingUnits} must run AFTER the voting weight movement is registered, such that it is reflected on {_getVotingUnits}.

Agree with @Amxx that "extends {Votes}" is preferred. I modified it slightly to not repeat the idea of changes to the flow of `_transferVotingUnits`.

contracts/governance/utils/VotesExtended.sol
```diff
@@ -7,6 +7,28 @@ import {SafeCast} from "../../utils/math/SafeCast.sol";
 
 /**
  * @dev Extension of {Votes} that adds checkpoints for delegations and balances.
+ *
+ * WARNING: While this contract extends {Votes}, valid uses of {Votes} may not be compatible with 
+ * {VotesExtended} without additional considerations. This implementation of {_transferVotingUnits} must
+ * run AFTER the voting weight movement is registered, such that it is reflected on {_getVotingUnits}.
+ *
+ * Said differently, {VotesExtended} MUST be integrated in a way such that calls {_transferVotingUnits} AFTER the
```
Amxx, line 15:
```suggestion
 * Said differently, {VotesExtended} MUST be integrated in a way that calls {_transferVotingUnits} AFTER the
```

ernestognw (approved):

PR #5309, opened 2024-11-20
Document voting counting until casted in GovernorCountingOverridable

ernestognw:
Fixes L-04

Although we're not changing the code, it's true that more explicit documentation is required

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/governance/extensions/GovernorCountingOverridable.sol
```diff
@@ -9,7 +9,10 @@ import {GovernorVotes} from "./GovernorVotes.sol";
 
 /**
  * @dev Extension of {Governor} which enables delegatees to override the vote of their delegates. This module requires a
- * token that inherits {VotesExtended}.
+ * token that inherits `VotesExtended`.
+ *
+ * NOTE: Overriding a vote doesn't force the vote to be casted. Consider that an overridden vote won't be counted towards
```
Amxx, line 14: That is incorrect. The overriden vote do affect the result, and count toward quroum even if the delegate never casts a vote.

contracts/governance/extensions/GovernorCountingOverridable.sol
```diff
@@ -52,6 +55,8 @@ abstract contract GovernorCountingOverridable is GovernorVotes {
 
     /**
      * @dev See {IGovernor-hasVoted}.
+     *
+     * NOTE: Overridden votes are not considered casted until vote is really casted through {castVote} or similar.
```
Amxx, line 59: I'm not sure I agree with the wording here.

```suggestion
     * NOTE: Vote casting is the operation by which a delegate uses the power that was delegated to him. Casting an 
     * override is an independent operation that uses the power of the voter itself, and not the power that was 
     * delegated to him. Consequently, overriding a vote does not count has voting, and will not affect this getter. 
     * Checking weither a user casted an override is done using {hasVotedOverride}. This function will only return 
     * true if the account performed a "normal" vote through {castVote} or similar.
```

contracts/governance/extensions/GovernorCountingOverridable.sol
```diff
@@ -120,7 +125,12 @@ abstract contract GovernorCountingOverridable is GovernorVotes {
         return totalWeight;
     }
 
-    /// @dev Variant of {Governor-_countVote} that deals with vote overrides.
+    /**
+     * @dev Variant of {Governor-_countVote} that deals with vote overrides.
+     *
+     * NOTE: Overridden votes are not considered casted until the vote is really casted through {castVote} or similar. Consider
+     * that overriding a vote won't count towards quorum nor the vote result until casted.
```
Amxx, line 132:
```suggestion
     * NOTE: Casting an override is the operation by which an account uses its own voting power (and not the power 
     * that was delegated to him by third parties). Vote casting is a different operation by which a delegate uses the 
     * power that was delegated to him. Consequently, overriding a vote does not count has voting.
```

ernestognw:
Right, I removed the NatSpec at the top of the contract and kept the two notes.

contracts/governance/extensions/GovernorCountingOverridable.sol
```diff
@@ -120,7 +125,13 @@ abstract contract GovernorCountingOverridable is GovernorVotes {
         return totalWeight;
     }
 
-    /// @dev Variant of {Governor-_countVote} that deals with vote overrides.
+    /**
+     * @dev Variant of {Governor-_countVote} that deals with vote overrides.
+     *
+     * NOTE: Casting a vote through {castVote} (or similar) uses the power the account has delegated to. On the
+     * opposite, casting an override with {castOverrideVote} (or similar) uses the power of the account itself as a
+     * single voter. Consequently, overriding a vote does not count as voting until the account casts a vote.
```
Amxx, line 133: I'm not sure about the "until the account casts a vote" part. It implies that the using casting a vote would sudently change the fact that the override now counts as voting, when the reality is the override never counts as voting, and the vote casting is what counts.

```suggestion
     * single voter. Consequently, overriding a vote does not count as voting.
```

contracts/governance/extensions/GovernorCountingOverridable.sol
```diff
@@ -52,6 +52,11 @@ abstract contract GovernorCountingOverridable is GovernorVotes {
 
     /**
      * @dev See {IGovernor-hasVoted}.
+     *
+     * NOTE: Casting a vote through {castVote} (or similar) uses the power the account has delegated to. On the
+     * opposite, casting an override with {castOverrideVote} (or similar) uses the power of the account itself as a
+     * single voter. Consequently, overriding a vote does not count as voting and won't be reflected by this getter.
+     * Consider using {hasVotedOverride} to check if an account has overridden a vote instead.
```
Amxx, line 59:
```suggestion
     * NOTE: Calling {castVote} (or similar) casts a vote using the voting power that is delegated to the voter.
     * Conversely, calling {castOverrideVote} (or similar) uses the voting power of the account itself, from its asset 
     * balances. Casting an "override vote" does not count as voting and won't be reflected by this getter. Consider 
     * using {hasVotedOverride} to check if an account has casted an "override vote" for a given proposal id.
```

contracts/governance/extensions/GovernorCountingOverridable.sol
```diff
@@ -120,7 +125,13 @@ abstract contract GovernorCountingOverridable is GovernorVotes {
         return totalWeight;
     }
 
-    /// @dev Variant of {Governor-_countVote} that deals with vote overrides.
+    /**
+     * @dev Variant of {Governor-_countVote} that deals with vote overrides.
+     *
+     * NOTE: Casting a vote through {castVote} (or similar) uses the power the account has delegated to. On the
+     * opposite, casting an override with {castOverrideVote} (or similar) uses the power of the account itself as a
+     * single voter. Consequently, overriding a vote does not count as voting. See {hasVoted}.
```
Amxx, line 133:
```suggestion
     * NOTE: See {hasVoted} for more details about the difference between {castVote} and {castOverrideVote}.
```

arr00 (approved):

Amxx (approved):

PR #5313, opened 2024-11-20
Add `factory()`, `factoryData()` and `paymasterData()` helpers to ERC4337Utils

Amxx:
Fixes #N-01

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/account/utils/draft-ERC4337Utils.sol
+++ b/contracts/account/utils/draft-ERC4337Utils.sol
@@ -140,17 +140,22 @@ library ERC4337Utils {
 
     /// @dev Returns the first section of `paymasterAndData` from the {PackedUserOperation}.
     function paymaster(PackedUserOperation calldata self) internal pure returns (address) {
-        return address(bytes20(self.paymasterAndData[0:20]));
+        return self.paymasterAndData.length < 52 ? address(0) : address(bytes20(self.paymasterAndData[0:20]));
     }
 
     /// @dev Returns the second section of `paymasterAndData` from the {PackedUserOperation}.
     function paymasterVerificationGasLimit(PackedUserOperation calldata self) internal pure returns (uint256) {
-        return uint128(bytes16(self.paymasterAndData[20:36]));
+        return self.paymasterAndData.length < 52 ? 0 : uint128(bytes16(self.paymasterAndData[20:36]));
     }
 
     /// @dev Returns the third section of `paymasterAndData` from the {PackedUserOperation}.
     function paymasterPostOpGasLimit(PackedUserOperation calldata self) internal pure returns (uint256) {
-        return uint128(bytes16(self.paymasterAndData[36:52]));
+        return self.paymasterAndData.length < 52 ? 0 : uint128(bytes16(self.paymasterAndData[36:52]));
+    }
+
+    /// @dev Returns the forth section of `paymasterAndData` from the {PackedUserOperation}.
+    function paymasterData(PackedUserOperation calldata self) internal pure returns (bytes calldata) {
+        return self.paymasterAndData.length < 52 ? _emptyCalldataBytes() : self.paymasterAndData[52:];
     }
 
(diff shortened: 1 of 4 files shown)


contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -98,6 +98,16 @@ library ERC4337Utils {
...
     }
 
+    /// @dev Returns `factory` from the {PackedUserOperation}, or address(0) if the initCode is empty or not properly formatted.
+    function factory(PackedUserOperation calldata self) internal pure returns (address) {
+        return self.initCode.length < 20 ? address(0) : address(bytes20(self.initCode[0:20]));
+    }
+
+    /// @dev Returns `factoryData` from the {PackedUserOperation}, or empty bytes if the initCode is empty or not properly formatted.
+    function factoryData(PackedUserOperation calldata self) internal pure returns (bytes calldata) {
+        return self.initCode.length < 20 ? _emptyCalldataBytes() : self.initCode[20:];
```
cairoeth, line 108: i think doing this is equivalent to the `_emptyCalldataBytes` function
```suggestion
        return self.initCode.length < 20 ? self.initCode[0:0] : self.initCode[20:];
```

contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -130,16 +140,29 @@ library ERC4337Utils {
...
 
     /// @dev Returns the third section of `paymasterAndData` from the {PackedUserOperation}.
     function paymasterPostOpGasLimit(PackedUserOperation calldata self) internal pure returns (uint256) {
-        return uint128(bytes16(self.paymasterAndData[36:52]));
+        return self.paymasterAndData.length < 52 ? 0 : uint128(bytes16(self.paymasterAndData[36:52]));
+    }
+
+    /// @dev Returns the forth section of `paymasterAndData` from the {PackedUserOperation}.
+    function paymasterData(PackedUserOperation calldata self) internal pure returns (bytes calldata) {
+        return self.paymasterAndData.length < 52 ? _emptyCalldataBytes() : self.paymasterAndData[52:];
```
cairoeth, line 158: same as other comment
```suggestion
        return self.paymasterAndData.length < 52 ? self.paymasterAndData[0:0] : self.paymasterAndData[52:];
```

contracts/account/utils/draft-ERC4337Utils.sol
cairoeth: not sure about gas but i dont expect a huge difference

Amxx: self.initCode is not at the begining of the calldata, so likelly doing that involves:
- copying `self` (that is a call data pointer) 
- pushing some offset (computing the offset may includes some calldataload?)
- adding the two, so that your calldata pointer starts at self.initCode, and has length 0

I'm expecting that to be more expensive. 

I also personally prefer having a very explicit "this is an empty bytes object" function.

test/account/utils/draft-ERC4337Utils.test.js
```diff
@@ -176,28 +203,43 @@ describe('ERC4337Utils', function () {
...
 
       it('returns postOpGasLimit', async function () {
-        expect(this.utils.$paymasterPostOpGasLimit(this.userOp.packed)).to.eventually.equal(this.postOpGasLimit);
+        expect(this.utils.$paymasterPostOpGasLimit(this.userOp.packed)).to.eventually.equal(
+          this.userOp.paymasterPostOpGasLimit,
+        );
+        expect(this.utils.$paymasterPostOpGasLimit(this.emptyUserOp.packed)).to.eventually.equal(0n);
+      });
+
+      it('returns paymasterData', async function () {
```
ernestognw, line 240: For consistency

```suggestion
      it('returns data', async function () {
```

ernestognw (approved):

PR #5317, opened 2024-11-25
Various changes to code clarity (Fix N-07)

Amxx:
Fixes N-07

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/NoncesKeyed.sol
```diff
@@ -7,6 +7,10 @@ import {Nonces} from "./Nonces.sol";
  * @dev Alternative to {Nonces}, that supports key-ed nonces.
  *
  * Follows the https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support[ERC-4337's semi-abstracted nonce system].
+ *
+ * Note: This contract extends {Nonces}, and reuses {Nonces}'s storage for key #0. This makes upgrading from {Nonces}
+ * to {NoncesKeyed} safe, as such an upgrade will NOT reset the current state of nonces. Doing otherwise would enable
+ * replay attacks where operation that used nonces before the upgrade would be replayable after the upgrade.
```
ernestognw, line 13: This should be clearer

```suggestion
 * NOTE: This contract extends from {Nonces} and reuses its storage for the first nonce key (i.e. `0`). This
 * makes upgrading from {Nonces} to {NoncesKeyed} safe. Doing so will NOT reset the current state of nonces to
 * avoid replay attacks where a nonce is reused after the upgrade if it resets the nonces for the first key.
```

ernestognw (dismissed):

contracts/utils/NoncesKeyed.sol
Amxx: @arr00 is "to extend from something" correct to you? I'd say "to extend something" or "to inherit from something".

ernestognw: We can just leave it like this:

```suggestion
 * NOTE: This contract extends from {Nonces} and reuses its storage for the first nonce key (i.e. `0`). This
 * makes upgrading from {Nonces} to {NoncesKeyed} safe. Doing so will NOT reset the current state of nonces to
 * avoid replay attacks where a nonce is reused after the upgrade.
```

arr00: > @arr00 is "to extend from something" correct to you? I'd say "to extend something" or "to inherit from something".

Agree with Hadrien on these sounding more natural.

ernestognw (approved):

contracts/governance/utils/VotesExtended.sol
```diff
@@ -34,8 +34,8 @@ abstract contract VotesExtended is Votes {
     using Checkpoints for Checkpoints.Trace160;
     using Checkpoints for Checkpoints.Trace208;
 
-    mapping(address delegator => Checkpoints.Trace160) private _delegateCheckpoints;
-    mapping(address account => Checkpoints.Trace208) private _balanceOfCheckpoints;
+    mapping(address delegator => Checkpoints.Trace160) private _userDelegationCheckpoints;
+    mapping(address account => Checkpoints.Trace208) private _userVotingWeightCheckpoints;
```
arr00, line 38: It's not clear what is represented here based on the name--delegated voting weight (the default way of referencing voting weight), or the "intrinsic" voting weight to an address.

contracts/utils/NoncesKeyed.sol
```diff
@@ -7,6 +7,10 @@ import {Nonces} from "./Nonces.sol";
  * @dev Alternative to {Nonces}, that supports key-ed nonces.
  *
  * Follows the https://eips.ethereum.org/EIPS/eip-4337#semi-abstracted-nonce-support[ERC-4337's semi-abstracted nonce system].
+ *
+ * NOTE: This contract inherits from {Nonces} and reuses its storage for the first nonce key (i.e. `0`). This
+ * makes upgrading from {Nonces} to {NoncesKeyed} safe. Doing so will NOT reset the current state of nonces to
+ * avoid replay attacks where a nonce is reused after the upgrade.
```
arr00, line 13:
```suggestion
 * makes upgrading from {Nonces} to {NoncesKeyed} safe. Doing so will NOT reset the current state of nonces.
```
This sentence is still a run on. Is it necessary to spell out the specific attack vector here?

ernestognw: I don't think is necessary to mention an attack vector. Also, is there any in concrete?
Documenting it here is important for other systems (e.g. frontends or indexers) to know what behavior they should rely on.

ernestognw: To be clear, I'm fine with the suggestion but perhaps @Amxx has a reason to keep the original

ernestognw (dismissed):

contracts/utils/NoncesKeyed.sol
arr00: This sentence wasn't affected.

contracts/governance/utils/VotesExtended.sol
arr00, line 38:
```suggestion
    mapping(address account => Checkpoints.Trace208) private _userVotingUnitsCheckpoints;
```

arr00 (approved):

ernestognw (dismissed):

ernestognw (approved):

PR #5319, opened 2024-11-25
Document canonicalization in CAIP2 and CAIP10 libraries

ernestognw:
Fixes CR-01

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/CAIP10.sol
+++ b/contracts/utils/CAIP10.sol
@@ -13,6 +13,12 @@ import {CAIP2} from "./CAIP2.sol";
  * account_id:        chain_id + ":" + account_address
  * chain_id:          [-a-z0-9]{3,8}:[-_a-zA-Z0-9]{1,32} (See {CAIP2})
  * account_address:   [-.%a-zA-Z0-9]{1,128}
+ *
+ * WARNING: According to [CAIP-10's canonicalization section](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md#canonicalization),
+ * the specificiation remains at the developer's discretion. However, case sensitivity or other transformations
+ * may introduce ambiguity. For example, when building hashes to identify accounts or data associated to them,
+ * multiple representations of the same account would derive to different hashes. Consider using {Strings-toChecksumHexString}
+ * for a consistent format.
  */
 library CAIP10 {
     using Strings for address;
--- a/contracts/utils/CAIP10.sol
+++ b/contracts/utils/CAIP10.sol
@@ -15,7 +15,7 @@ import {CAIP2} from "./CAIP2.sol";
  * account_address:   [-.%a-zA-Z0-9]{1,128}
  *
  * WARNING: According to [CAIP-10's canonicalization section](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md#canonicalization),
- * the specificiation remains at the developer's discretion. However, case sensitivity or other transformations
+ * the specification remains at the developer's discretion. However, case sensitivity or other transformations
  * may introduce ambiguity. For example, when building hashes to identify accounts or data associated to them,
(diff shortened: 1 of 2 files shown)


contracts/utils/CAIP2.sol
```diff
@@ -12,6 +12,11 @@ import {Strings} from "./Strings.sol";
  * chain_id:    namespace + ":" + reference
  * namespace:   [-a-z0-9]{3,8}
  * reference:   [-_a-zA-Z0-9]{1,32}
+ *
+ * WARNING: The CAIP-2 identifier does not require a canonical representation. In the EVM context, it is
```
Amxx, line 16: I'm not sure what "The CAIP-2 identifier does not require a canonical representation" means.

Can you elaborate ?

contracts/utils/CAIP10.sol
```diff
@@ -13,6 +13,12 @@ import {CAIP2} from "./CAIP2.sol";
  * account_id:        chain_id + ":" + account_address
  * chain_id:          [-a-z0-9]{3,8}:[-_a-zA-Z0-9]{1,32} (See {CAIP2})
  * account_address:   [-.%a-zA-Z0-9]{1,128}
+ *
+ * WARNING: According to [CAIP-10's canonicalization section](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md#canonicalization),
+ * the specification remains at the developer's discretion. However, case sensitivity or other transformations
```
Amxx, line 18: "the specification" of what?

I'm not sure However is the right world to use here ... or maybe I'm missunderstanding.

contracts/utils/CAIP2.sol
ernestognw: The point is that CAIP2 does not define a "canonical" representation for each chain. Although EVM chains will use `erc155` as namespace, there may be chains that have more than one representation for a single chain.

Perhaps saying that "The CAIP-2 identifier may not be canonical" instead of "does not require a canonical representation".

Amxx: What about

> In some cases, multiple CAIP-2 identifiers may all be valid representation of a single chain. For EVM chains, it is recommended to use "eip155:xxx` with `xxx` the EIP-155 chain id as the canonical representation.

ernestognw: I like it

contracts/utils/CAIP10.sol
```diff
@@ -13,6 +13,12 @@ import {CAIP2} from "./CAIP2.sol";
  * account_id:        chain_id + ":" + account_address
  * chain_id:          [-a-z0-9]{3,8}:[-_a-zA-Z0-9]{1,32} (See {CAIP2})
  * account_address:   [-.%a-zA-Z0-9]{1,128}
+ *
+ * WARNING: According to [CAIP-10's canonicalization section](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md#canonicalization),
+ * the implementation remains at the developer's discretion. However, case sensitivity or other transformations
+ * may introduce ambiguity. For example, when building hashes to identify accounts or data associated to them,
+ * multiple representations of the same account would derive to different hashes. Consider using {Strings-toChecksumHexString}
+ * for a consistent format.
```
Amxx, line 21:
```suggestion
 * WARNING: According to [CAIP-10's canonicalization section](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md#canonicalization),
 * the implementation remains at the developer's discretion. Please note that case variations may introduce ambiguity. 
 * For example, when building hashes to identify accounts or data associated to them, multiple representations of the 
 * same account would derive to different hashes. For EVM chains, we recommend using checksummed addresses for the 
 * "account_address" part. They can be generated onchain using {Strings-toChecksumHexString}.
```

Amxx (approved):

PR #5323, opened 2024-11-26
Remove unnecessary trailing `/` in prepare-docs.sh

0xwitty:
The script contained a small but impactful issue in this line:

```bash
name="${f/#"$examples_source_dir/"/}"
```

The corrected version is:

```bash
name="${f/#"$examples_source_dir"/}"
```

### Issue Explanation
The extra `/` after `"$examples_source_dir"` in the `bash` string substitution caused incorrect path handling. This could result in errors or malformed paths when processing examples, particularly when attempting to adjust imports and structure.

### Why It Matters
- **Functionality**: This fix ensures the examples are correctly copied and their imports are properly adjusted, which is critical for generating accurate documentation.
- **Maintainability**: Removing this unnecessary character improves script readability and reduces potential confusion for future contributors.
- **Compatibility**: The fix ensures the script behaves reliably across different environments and path configurations.

This small change prevents potential issues during the documentation generation process and guarantees smoother execution.

#### PR Checklist

- [x] Tests
- [ ] Documentation


Amxx:
@ernestognw you worked a lot on docs. Can you please check this ? If you approve, feel free to merge with my second approval.

ernestognw (approved):
Both the old and new versions produce the same output so they're equivalent.

May be relevant since `examples_source_dir` does not have a trailing slash (`/`), but it seems that this line is treating `/` as a literal instead of a character.

PR #5324, opened 2024-11-26
Implement feedback for M-01, L-08, L-09

ernestognw:
Implement feedback from audit review of previous issues

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -163,7 +163,7 @@ library Clones {
      * access the arguments within the implementation, use {fetchCloneArgs}.
      *
      * This function uses the create2 opcode and a `salt` to deterministically deploy the clone. Using the same
-     * `implementation`, `args` and `salt` multiple time will revert, since the clones cannot be deployed twice
+     * `implementation`, `args` and `salt` multiple times will revert, since the clones cannot be deployed twice
      * at the same address.
      */
     function cloneDeterministicWithImmutableArgs(
@@ -229,7 +229,7 @@ library Clones {
     function fetchCloneArgs(address instance) internal view returns (bytes memory) {
         bytes memory result = new bytes(instance.code.length - 0x2d); // revert if length is too short
         assembly ("memory-safe") {
-            extcodecopy(instance, add(result, 0x20), 0x2d, mload(result))
+            extcodecopy(instance, add(result, 32), 0x2d, mload(result))
         }
         return result;
     }
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -229,7 +229,7 @@ library Clones {
     function fetchCloneArgs(address instance) internal view returns (bytes memory) {
         bytes memory result = new bytes(instance.code.length - 0x2d); // revert if length is too short
         assembly ("memory-safe") {
-            extcodecopy(instance, add(result, 32), 0x2d, mload(result))
+            extcodecopy(instance, add(result, 0x20), 0x2d, mload(result))
         }
         return result;
     }
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -227,9 +227,9 @@ library Clones {
      *   function should only be used to check addresses that are known to be clones.
      */
     function fetchCloneArgs(address instance) internal view returns (bytes memory) {
-        bytes memory result = new bytes(instance.code.length - 0x2d); // revert if length is too short
+        bytes memory result = new bytes(instance.code.length - 45); // revert if length is too short
         assembly ("memory-safe") {
-            extcodecopy(instance, add(result, 0x20), 0x2d, mload(result))
+            extcodecopy(instance, add(result, 32), 45, mload(result))
         }
         return result;
     }
@@ -248,11 +248,11 @@ library Clones {
         address implementation,
         bytes memory args
     ) private pure returns (bytes memory) {
-        if (args.length > 0x5fd3) revert CloneArgumentsTooLong();
+        if (args.length > 24531) revert CloneArgumentsTooLong();
         return
             abi.encodePacked(
                 hex"61",
-                uint16(args.length + 0x2d),
+                uint16(args.length + 45),
                 hex"3d81600a3d39f3363d3d373d3d3d363d73",
                 implementation,
                 hex"5af43d82803e903d91602b57fd5bf3",
--- a/contracts/utils/Strings.sol
+++ b/contracts/utils/Strings.sol
@@ -325,7 +325,8 @@ library Strings {
     }
 
     /**
-     * @dev Variant of {tryParseHexUint} that does not check bounds and returns (true, 0) if `begin <= end` excluding `0x`.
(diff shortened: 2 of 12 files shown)


contracts/utils/Strings.sol
```diff
@@ -391,11 +391,11 @@ library Strings {
         uint256 end
     ) internal pure returns (bool success, address value) {
         // check that input is the correct length
-        bool hasPrefix = (begin < end + 1) && bytes2(_unsafeReadBytesOffset(bytes(input), begin)) == bytes2("0x"); // don't do out-of-bound (possibly unsafe) read if sub-string is empty
+        bool hasPrefix = (end > begin + 1) && bytes2(_unsafeReadBytesOffset(bytes(input), begin)) == bytes2("0x"); // don't do out-of-bound (possibly unsafe) read if sub-string is empty
 
         uint256 expectedLength = 40 + hasPrefix.toUint() * 2;
 
-        if (end - begin == expectedLength && end <= bytes(input).length) {
+        if (begin <= end && end - begin == expectedLength && end <= bytes(input).length) {
```
Amxx, line 398: Since the issue here is the overflow, it feels like an equivalent but cheper option would be
```suggestion
        if (begin + expectedLength == end && end <= bytes(input).length) {
```

contracts/proxy/Clones.sol
```diff
@@ -229,7 +229,7 @@ library Clones {
     function fetchCloneArgs(address instance) internal view returns (bytes memory) {
         bytes memory result = new bytes(instance.code.length - 0x2d); // revert if length is too short
         assembly ("memory-safe") {
-            extcodecopy(instance, add(result, 0x20), 0x2d, mload(result))
```
Amxx, line 232: Any rational to changing that 0x20 in 32 but keeping `0x2d` as an hex?

If I had the change something, that would be the second one. There are many places where we do `add(..., 0x20)`: ERC2771Forwarder, Bytes, Create2, SlotDerivation, Arrays, Base64, Checkpoints, Math, RSA, P256, ECDSA, ERC1155, SAfeERC20 ...

There are way fewer places where we do `add(..., 32)`

contracts/proxy/Clones.sol
```diff
@@ -229,7 +229,7 @@ library Clones {
...
+            extcodecopy(instance, add(result, 32), 0x2d, mload(result))
```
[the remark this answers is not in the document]
Amxx:
```suggestion
            extcodecopy(instance, add(result, 0x20), 0x2d, mload(result))
```

contracts/utils/Strings.sol
```diff
@@ -323,7 +323,7 @@ library Strings {
     }
 
     /**
-     * @dev Variant of {tryParseHexUint} that does not check bounds and returns (true, 0) if they are invalid.
+     * @dev Variant of {tryParseHexUint} that does not check bounds and returns (true, 0) if `begin <= end` excluding `0x`.
```
Amxx, line 326: I don't understand.

I most cases (like the normal cases) we have "begin <= end" ... yet we don't necessarily return `(true, 0)` ... we return `(true, result)` with the result possibly being non zero.

It feels to me like the more we try to address audit comments, the more confused we get. I used to be confident in the code, that is no longer the case. IMO we should go back and have a look at what the code really does, instead of adding patches over patches.

In this particular what do we have:
- a function that tryes to parse uints in hex format
- it do not check the bounds
- if the input within bonds and contain a valid hex string, it returns (true, result)
- if the input within bonds and contains an invalid hex string, it returns (false, 0)
- if begin > end, it returns (true, 0)
- if end > input.length, it reads possibly unalocated memory.

Note that this is a private function, and that any documentation will NOT show up in docs.openzeppelin.com. So IMO we should either:
- not document at all (that used to be the rule for private function)
- be very factual and list the cases one by one as I did here.

contracts/utils/Strings.sol
```diff
@@ -390,12 +390,15 @@ library Strings {
         uint256 begin,
         uint256 end
     ) internal pure returns (bool success, address value) {
+        bool inBounds = begin <= end && // Positive length
```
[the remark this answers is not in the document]
Amxx: the `if (end > bytes(input).length || begin > end) return (false, address(0));` check is the same one we have in:
- `tryParseUint(string, uint256, uint256) internal`
- `tryParseInt(string, uint256, uint256) internal`
- `tryParseHexUint(string, uint256, uint256) internal`

contracts/utils/Strings.sol
Amxx:
```suggestion
     * @dev Implementation of {tryParseHexUint} that does not check bounds. Caller should check that 
     * `begin <= end <= input.length`. Other inputs would result in unexpected behavior.
```

contracts/utils/Strings.sol
```diff
@@ -177,7 +177,7 @@ library Strings {
     }
 
     /**
-     * @dev Variant of {tryParseUint} that does not check bounds and returns (true, 0) if they are invalid.
+     * @dev Implementation of {tryParseUint} that does not check bounds and returns (true, 0) if they are invalid.
```
Amxx, line 180:
```suggestion
     * @dev Implementation of {tryParseUint} that does not check bounds. Caller should check that 
     * `begin <= end <= input.length`. Other inputs would result in unexpected behavior.
```

contracts/utils/Strings.sol
```diff
@@ -177,7 +177,8 @@ library Strings {
     }
 
     /**
-     * @dev Variant of {tryParseUint} that does not check bounds and returns (true, 0) if they are invalid.
+     * @dev Implementation of {tryParseUint} that does not check bounds. Caller should make sure that
+     * `begin <= end <= input.length`. Other inputs would result in unexpected behavior.
```
Amxx, line 181:
```suggestion
     * `begin <= end <= input.length`. Other inputs would result in undefined behavior.
```

contracts/utils/Strings.sol
```diff
@@ -249,7 +250,8 @@ library Strings {
     }
 
     /**
-     * @dev Variant of {tryParseInt} that does not check bounds and returns (true, 0) if they are invalid.
+     * @dev Implementation of {tryParseInt} that does not check bounds. Caller should make sure that
+     * `begin <= end <= input.length`. Other inputs would result in unexpected behavior.
```
Amxx, line 254:
```suggestion
     * `begin <= end <= input.length`. Other inputs would result in undefined behavior.
```

contracts/utils/Strings.sol
```diff
@@ -323,7 +325,8 @@ library Strings {
     }
 
     /**
-     * @dev Variant of {tryParseHexUint} that does not check bounds and returns (true, 0) if they are invalid.
+     * @dev Implementation of {tryParseHexUint} that does not check bounds. Caller should make sure that
+     * `begin <= end <= input.length`. Other inputs would result in unexpected behavior.
```
Amxx, line 329:
```suggestion
     * `begin <= end <= input.length`. Other inputs would result in undefined behavior.
```

Amxx (approved):

Amxx (approved):

PR #5332, opened 2024-12-01
Fix return type for decodeSingle function to avoid potential issues w…

mdqst:
**Description:**

There was a minor issue in the `decodeSingle` function in the contract where the return type for `callData` was declared as `bytes calldata`. The `calldata` type is a special, read-only data location in Solidity, and returning it directly could cause unpredictable behavior when the data is manipulated later. 

To fix this, I changed the return type of `callData` to `bytes memory`, which allows for safe manipulation and processing of the decoded data without violating the immutability of `calldata`. 

**Importance:**  
This change is critical because returning `calldata` directly could lead to issues when trying to modify or use the returned `callData` in further operations. By switching to `bytes memory`, we ensure that the data can be safely handled and manipulated within the function, improving reliability and preventing potential runtime errors. 

Here is the corrected version of the function:

```solidity
function decodeSingle(bytes calldata executionCalldata) internal pure returns (address target, uint256 value, bytes memory callData)
```

This small change will help maintain the integrity of the contract and ensure more predictable behavior when working with decoded data.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/account/utils/draft-ERC7579Utils.sol
+++ b/contracts/account/utils/draft-ERC7579Utils.sol
@@ -141,7 +141,7 @@ library ERC7579Utils {
     /// @dev Decodes a single call execution. See {encodeSingle}.
     function decodeSingle(
         bytes calldata executionCalldata
-    ) internal pure returns (address target, uint256 value, bytes calldata callData) {
+    ) internal pure returns (address target, uint256 value, bytes memory callData) {
         target = address(bytes20(executionCalldata[0:20]));
         value = uint256(bytes32(executionCalldata[20:52]));
         callData = executionCalldata[52:];


Amxx:
Hello @mdqst 

### Why me chose calldata here 

Object in calldata can easily be copied to memory. That is just a simple calldatacopy operation. On the other hand, objects in memory cannot be easily converted to calldata.

As a user of the library, if you want to get the callData object in memory, its really is as simple as doing
```solidity
    (address target, uint256 value, bytes memory callData) = decodeSingle(executionCalldata);
```
And you'll be able to modify the callData memory object.

On the other had, some users may want to avoid copying to memory, as allocating memory can be an expensive operation. Overall, we feel like objects in calldata should not be implicitelly moved to memory. 

### About your security concerns

> The calldata type is a special, read-only data location in Solidity, and returning it directly could cause unpredictable behavior when the data is manipulated later.

I'd like to know more about the unpredictable behavior. As you said, calldata is read-only, so there really isn't any manipulation that can happen unless the data is moved to memory (at which point it can be modified). IMO, that argument justifies to NOT use memory, and stick to calldata to preserve integrity.

PR #5336, opened 2024-12-02
Remove token value mention in ERC-1155 documentation

ebulku:
Changed an incorrect statement about the value of the locked tokens

#### PR Checklist

- [ ] Tests
- [x] Documentation


ebulku:
> GNT is the hold version of GLM, which is today marketed around 0.47 .

That is true if you can migrate from GNT to GLM. I just checked the etherscan link where the market cap  and the value of the locked tokens shows $0

ernestognw (approved):

PR #5337, opened 2024-12-03
Add `_initializableStorageSlot`

### LINKED ISSUE #4782: Rigidity in Initializable's INITIALIZABLE_STORAGE slot
**🧐 Motivation**
Most ERC4337/ERC6900 smart contract accounts use the UUPSUpgradeable proxy pattern today. In OZ5.0's Initializable, the same `INITIALIZABLE_STORAGE` slot is used. As such, if more than 1 SCA implementation uses OZ5.0's Initializable, when a proxy is initialized with one of those implementations, it would not upgrade to other SCA implementations that require an `initializer` function to set up - as such, for security reasons, no SCA can use the Initializable library

**📝 Details**
1. Consider switching `_getInitializableStorage()` or `INITIALIZABLE_STORAGE` to `internal virtual` and allow parent contracts to choose which storage slot to use. This also allows child contracts to group storage slots together when verkle trees are adopted without having to worrying about storage collisions across versions.
2. Consider switching to a deterministic, variable `INITIALIZABLE_STORAGE` instead. One possible implementation is: 
```
bytes32 private immutable INITIALIZABLE_STORAGE;
constructor() {
   bytes32 seed = keccak256(abi.encode("openzeppelin.storage.Initializable", address(this) ));
   INITIALIZABLE_STORAGE = keccak256(abi.encode(uint256(seed ) - 1)) & ~bytes32(uint256(0xff));
}
```

Credit: @gpersoon

ernestognw:
Fixes #4782 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/fast-coats-try.md
+++ b/.changeset/fast-coats-try.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Initializable`: Add `_initializableStorageSlot` function that returns a pointer to the storage struct. The function allows to customize with a custom storage slot with an `override`.
--- a/contracts/proxy/utils/Initializable.sol
+++ b/contracts/proxy/utils/Initializable.sol
@@ -217,12 +217,22 @@ abstract contract Initializable {
     }
 
     /**
+     * @dev Pointer to storage slot. Allows to override it with a custom storage location.
+     *
+     * NOTE: Consider following EIP-7201 formula to derive storage locations.
+     */
+    function _initializableStorageSlot() internal pure virtual returns (bytes32) {
+        return INITIALIZABLE_STORAGE;
+    }
+
+    /**
      * @dev Returns a pointer to the storage namespace.
      */
     // solhint-disable-next-line var-name-mixedcase
     function _getInitializableStorage() private pure returns (InitializableStorage storage $) {
+        bytes32 slot = _initializableStorageSlot();
         assembly {
-            $.slot := INITIALIZABLE_STORAGE
+            $.slot := slot
         }
     }
 }


contracts/proxy/utils/Initializable.sol
```diff
@@ -216,13 +216,23 @@ abstract contract Initializable {
         return _getInitializableStorage()._initializing;
     }
 
+    /**
+     * @dev Pointer to storage slot. Allows to override it with a custom storage location.
+     *
+     * NOTE: Consider following EIP-7201 formula to derive storage locations.
```
ernestognw, line 222:
```suggestion
     * NOTE: Consider following the ERC-7201 formula to derive storage locations.
```

.changeset/fast-coats-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Initializable`: Add `_initializableStorageSlot` function that returns a pointer to the storage struct. The function allows to customize with a custom storage slot with an `override`.
```
ernestognw, line 5:
```suggestion
`Initializable`: Add `_initializableStorageSlot` function that returns a pointer to the storage struct. The function allows customizing with a custom storage slot with an `override`.
```

ernestognw:
Alternatively, we can make `_getInitializableStorage` internal. Thoughts?

Amxx (commented):
I'm concerned by line 63

```solidity
     * when using with upgradeable contracts.
     *
     * @custom:storage-location erc7201:openzeppelin.storage.Initializable // <<== HERE !
     */
    struct InitializableStorage {
```

This comment will be wrong if the value is overriden. How will the plugin handle that ?

arr00:
> ```solidity
> function _getInitializableStorage() internal virtual pure returns (InitializableStorage storage $) {
> ```
> 
> Would be a smaller change, that we could easily generalize to all the upgradeable contracts. The downside is that overriding it would likely require the user writting assembly.

I imagine that most teams requiring this functionality won't have an issue writing minimal solidity. Don't think it's worth the extra function call just to bypass that.

Amxx (dismissed):

Amxx:
> ```solidity
> function _getInitializableStorage() internal virtual pure returns (InitializableStorage storage $) {
> ```
> 
> Would be a smaller change, that we could easily generalize to all the upgradeable contracts. The downside is that overriding it would likely require the user writting assembly.

@ernestognw would that work for you ?

Amxx (approved):

contracts/proxy/utils/Initializable.sol
```diff
@@ -216,13 +216,23 @@ abstract contract Initializable {
         return _getInitializableStorage()._initializing;
     }
 
+    /**
+     * @dev Pointer to storage slot. Allows to override it with a custom storage location.
```
arr00, line 220:
```suggestion
     * @dev Pointer to storage slot. Allows integrators to override it with a custom storage location.
```
Prefer something more readable than "Allows to"

arr00 (approved):

ernestognw:
> Would be a smaller change, that we could easily generalize to all the upgradeable contracts. The downside is that overriding it would likely require the user writting assembly.

I agree with this concern. However, I don't see a way around It given you can't override the NatSpec struct annotation.
Theoretically, if a user overrides the storage pointer using ERC-7201 (which the function already recommends), then collision should not be a risk.

I'm merging this PR, but I think we can discuss exploring an special annotation like `@custom:storage-location-renamed-from erc7201:openzeppelin.storage.Initializable <...>`

Amxx (approved):

PR #5340, opened 2024-12-04
Run linter on commit instead of push

Amxx:
Doing that on push means we run lint when pushing tags, deleting remove banches, ... things that really don't need to run the linter.

Using husky and staged-lint to keep the project cleaner, (and stop reinventing the weel)


.githooks/pre-commit
```diff
@@ -0,0 +1,21 @@
...
+	
+	# Lint javascript and typescript
+	for file in $(git diff --cached --name-only | grep -E '\.(js|ts)$')
+	do
+		npx prettier --log-level warn --ignore-path .gitignore $file --check
+		npx eslint $file
+	done
+
+	# Lint solidity
+	for file in $(git diff --cached --name-only | grep -E '(contracts|test)/.*\.sol$')
```
Amxx, line 16:
```suggestion
	for file in $(git diff --cached --name-only | grep -E '\.sol$')
```

frangio:
I have my opinions but I think it's up to you who will deal with this regularly. Technically this change seems fine. You could look into [lint-staged](https://github.com/lint-staged/lint-staged) if you wanted something that does the filtering out of the box.

.husky/pre-commit
```diff
@@ -0,0 +1,2 @@
+npm run test:generation
```
arr00, line 1: What does this do?

ernestognw: Some contracts are automatically generated with the `npm run generate` command, we run some basic Javascript templates (e.g. [Arrays](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/aba42a07da0d7f0645bc487b9a436690b4b5a1c7/scripts/generate/templates/Arrays.js)).

Since we need to make sure that the generated contracts are in sync with the templates, we run `test:generation` to make sure we're indeed in sync

scripts/prepare.sh
```diff
@@ -2,4 +2,4 @@
 
 set -euo pipefail
 
-if git status &>/dev/null; then git config core.hooksPath .githooks; fi
+npx husky
```
ernestognw, line 5: What if we remove this file and run `husky` in the `prepare` command instead?

ernestognw (approved):

scripts/prepare.sh
Amxx: that works for me

PR #5341, opened 2024-12-05
Add v5.2 audit report

Amxx:


arr00:
What are your thoughts on adding a hyperlink to the commit for each audit? Something like this https://github.com/arr00/openzeppelin-contracts/tree/temp/href-commit-audit/audits

Either to the commit details or to a file viewer at that commit.

ernestognw (approved):

PR #5343, opened 2024-12-05
Add interface for ERC6909

arr00:
Add the `ERC6909` interface to the repo.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/token/ERC6909/IERC6909.sol
Amxx, line 1: until this ERC is finalized, I think it the interface should be named `draft-IERC6909.sol`

arr00: Yes good point

contracts/interfaces/draft-IERC6909.sol
Amxx, line 1: Having the interface in both location is something we inherited from the previous version. We can't remove the IERC20.sol in `token/ERC20` as it would be breaking, so we have this "symlink" approach.

Here we don't have anythink to break, so maybe we should consider having the main interfaces directly in `contracts/interfaces/` and not having one in `contracts/token/ERC6909`

@ernestognw WDYT ?

Amxx: In particular in this PR that doesn't include anything other than the interface, it feels strange to have an otherwize empty `contracts/token/ERC6909` folder.

ernestognw: I would keep it in `contracts/interfaces` for now since we don't have an implementations. If we end up implementing, it's easy to move it to `contracts/token/interfaces/ERC6909` and importing it into `contracts/interfaces/ERC6909`

.changeset/long-walls-draw.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`IERC6909`: Add the interface for the ERC6909.
```
ernestognw, line 5:
```suggestion
`IERC6909`: Add the interface for the ERC-6909.
```

arr00:
There are 3 extensions to the base ERC6909 defined in [the ERC](https://eips.ethereum.org/EIPS/eip-6909). Should we add interfaces for all of them? They are ContractURI, Metadata, and TokenSupply.

Amxx:
Yes, I would have all 3 extensions decalred in the same file.

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -0,0 +1,118 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (interfaces/draft-IERC6909.sol)
```
ernestognw, line 2: This gets automatically added [here](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/502e1a0fe5460d5cf9c8ff5bd215a5a50180d730/scripts/release/update-comment.js), when the package gets versioned.

Let's remove it

```suggestion
```

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -0,0 +1,117 @@
...
+import {IERC165} from "../utils/introspection/IERC165.sol";
+
+/**
+ * @dev Required interface of an ERC-6909 compliant contract, as defined in the
+ * https://eips.ethereum.org/EIPS/eip-6909[ERC].
+ */
+interface IERC6909 is IERC165 {
+    /**
+     * @dev Emitted when the allowance of a `spender` for an `owner` is set by a call to `approve` for a
+     * token of type `id`. The new allowance is `amount`.
```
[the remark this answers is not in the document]
arr00: > Note that the ERC states:
> 
> > MUST be logged when the allowance is set by an owner.
> 
> So it may not be strictly done through `approve`

Meaning there may be a non ERC function used to update the allowance by the owner?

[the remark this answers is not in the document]
ernestognw: Correct. A developer may extend the standard and use another function to set the allowance. For example, something like `approveWithSignature`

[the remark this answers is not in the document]
arr00: Ok good point. Will generalize it for that consideration.

ernestognw (approved):
Updated title since we're not implementing the interface.

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -0,0 +1,117 @@
...
+     * Must return true.
+     */
+    function setOperator(address spender, bool isOperator) external returns (bool);
+
+    /**
+     * @dev Transfers `amount` of token type `id` from the caller's account to `to`.
+     *
+     * Must return true.
+     */
+    function transfer(address to, uint256 id, uint256 amount) external returns (bool);
```
Amxx, line 64: ERC uses `receiver`
```suggestion
    function transfer(address receiver, uint256 id, uint256 amount) external returns (bool);
```

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * @dev Transfers `amount` of token type `id` from `from` to `to`.
+     *
+     * Must return true.
+     */
+    function transferFrom(address from, address to, uint256 id, uint256 amount) external returns (bool);
```
Amxx, line 71: ERC naming
```suggestion
    function transferFrom(address sender, address receiver, uint256 id, uint256 amount) external returns (bool);
```

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -0,0 +1,117 @@
...
+     * Must return true.
+     */
+    function approve(address spender, uint256 id, uint256 amount) external returns (bool);
+
+    /**
+     * @dev Grants or revokes unlimited transfer permission of any token id to `spender` for the caller's tokens.
+     *
+     * Must return true.
+     */
+    function setOperator(address spender, bool isOperator) external returns (bool);
```
Amxx, line 57: ERC naming
```suggestion
    function setOperator(address spender, bool approved) external returns (bool);
```

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * @dev Emitted when the allowance of a `spender` for an `owner` is set for a token of type `id`.
+     * The new allowance is `amount`.
+     */
+    event Approval(address indexed owner, address indexed spender, uint256 indexed id, uint256 amount);
+
+    /**
+     * @dev Emitted when `owner` grants or revokes operator status for a `spender`.
+     */
+    event OperatorSet(address indexed owner, address indexed spender, bool isOperator);
```
Amxx, line 21:
```suggestion
    event OperatorSet(address indexed owner, address indexed spender, bool approved);
```

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * @dev Emitted when `amount` tokens of type `id` are moved from `from` to `to` initiated by `caller`.
+     */
+    event Transfer(address caller, address indexed from, address indexed to, uint256 indexed id, uint256 amount);
```
Amxx, line 26:
```suggestion
    event Transfer(address caller, address indexed sender, address indexed receiver, uint256 indexed id, uint256 amount);
```

Amxx (approved):

PR #5350, opened 2024-12-08
Allow pass in custom arguments to `shouldSupportInterfaces` test helper

ernestognw:
### Motivation

These allows to reuse the `shouldSupportInterfaces` function in another context. One concrete example is that ERC-7802 requires ERC-165 in the community repository, where we can't test ERC-165 unless we copy paste the helpers or store the interface in this repository.

To avoid either, I think we should allow passing in more arguments to `shouldSupportInterfaces` so that the helper can be reused elsewhere.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/utils/introspection/SupportsInterface.behavior.js
+++ b/test/utils/introspection/SupportsInterface.behavior.js
@@ -92,7 +92,7 @@ const SIGNATURES = {
 
 const INTERFACE_IDS = mapValues(SIGNATURES, interfaceId);
 
-function shouldSupportInterfaces(interfaces = []) {
+function shouldSupportInterfaces(interfaces = [], signatures = SIGNATURES, interfaceIds = INTERFACE_IDS) {
   interfaces.unshift('ERC165');
 
   describe('ERC165', function () {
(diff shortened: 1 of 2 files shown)


test/utils/introspection/SupportsInterface.behavior.js
```diff
@@ -92,7 +92,7 @@ const SIGNATURES = {
 
 const INTERFACE_IDS = mapValues(SIGNATURES, interfaceId);
 
-function shouldSupportInterfaces(interfaces = []) {
+function shouldSupportInterfaces(interfaces = [], signatures = SIGNATURES, interfaceIds = INTERFACE_IDS) {
```
Amxx, line 95: This allows inconsistent parameters. IMO we should only take the "signatures", and recompute the "interfaceIds" in the function.

Amxx:
```suggestion
function shouldSupportInterfaces(interfaces = [], signatures = SIGNATURES) {
    const interfaceIds = mapValues(SIGNATURES, interfaceId);
```

ernestognw: This make sense and it doesn't increase the test time considerably

Amxx (approved):

PR #5353, opened 2024-12-10
Add checks to ERC7579Utils.decodeBatch

Amxx:
#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,13 +172,55 @@ library ERC7579Utils {
...
         assembly ("memory-safe") {
-            let ptr := add(executionCalldata.offset, calldataload(executionCalldata.offset))
-            // Extract the ERC7579 Executions
-            executionBatch.offset := add(ptr, 32)
-            executionBatch.length := calldataload(ptr)
+            // Pointer to the end of the calldata buffer
+            let end := add(executionCalldata.offset, executionCalldata.length)
+
+            // Check the buffer at least one element (the offset to the array)
+            if slt(executionCalldata.length, 32) {
```
frangio, line 190: Why is this using signed less than?

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,34 @@ library ERC7579Utils {
...
+            if (arrayLength > type(uint64).max) revert ERC7579DecodingError();
+
+            // Get the array as a bytes slice, and check its is long enough
+            bytes calldata executionArray = executionCalldata[offset + 32:];
+            if (executionArray.length < arrayLength * 96) revert ERC7579DecodingError();
+
+            assembly ("memory-safe") {
+                executionBatch.offset := executionArray.offset
+                executionBatch.length := arrayLength
+            }
```
ernestognw, line 202: Wondering if this can go into a "calldata decoding" library or similar. It's essentially a replacement for `abi.decode`

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,34 @@ library ERC7579Utils {
     }
 
     /// @dev Decodes a batch of executions. See {encodeBatch}.
+    ///
+    /// Note: This function runs some checks and will throw a {ERC7579DecodingError} if the input is not properly formatted.
```
ernestognw, line 176:
```suggestion
    /// NOTE: This function runs some checks and will throw a {ERC7579DecodingError} if the input is not properly formatted.
```

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,34 @@ library ERC7579Utils {
...
+            // Since we know executionCalldata is at least 32, we can subtract with no overflow risk.
+            if (offset > bufferLength - 32) revert ERC7579DecodingError();
+
+            // Get the array length
+            uint256 arrayLength = uint256(bytes32(executionCalldata[offset:offset + 32]));
+            if (arrayLength > type(uint64).max) revert ERC7579DecodingError();
+
+            // Get the array as a bytes slice, and check its is long enough
+            bytes calldata executionArray = executionCalldata[offset + 32:];
+            if (executionArray.length < arrayLength * 96) revert ERC7579DecodingError();
```
ernestognw, line 197: We should explain that `96` comes from the length of the `Execution` array. If it's possible to link them it'd be better

frangio: Yes, this should be defined as a constant with an appropriate name.

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,34 @@ library ERC7579Utils {
...
+        unchecked {
+            uint256 bufferLength = executionCalldata.length;
+
+            // Check executionCalldata is not empty
+            if (bufferLength < 32) revert ERC7579DecodingError();
+
+            // Get the offset of the array, and check its valid
+            uint256 offset = uint256(bytes32(executionCalldata[0:32]));
+
+            // Array length is between offset and offset + 32. We check that this is in the buffer
```
frangio, line 187:
```suggestion
            // Array length should be found in offset:offset + 32. We check that this is within buffer bounds.
```

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,34 @@ library ERC7579Utils {
...
+            if (arrayLength > type(uint64).max) revert ERC7579DecodingError();
+
+            // Get the array as a bytes slice, and check its is long enough
+            bytes calldata executionArray = executionCalldata[offset + 32:];
+            if (executionArray.length < arrayLength * 96) revert ERC7579DecodingError();
+
+            assembly ("memory-safe") {
+                executionBatch.offset := executionArray.offset
+                executionBatch.length := arrayLength
+            }
```
ernestognw, line 202: Isn't this missing that each `Execute` offset is within bounds. An `Execute` might have an offset that exceeds the `executionCalldata` space

contracts/account/utils/draft-ERC7579Utils.sol
frangio, line 187:
```suggestion
            // The array length should be found at offset and be 32 bytes long.
            // We check that this is within buffer bounds.
```

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,34 @@ library ERC7579Utils {
...
+            if (bufferLength < 32) revert ERC7579DecodingError();
+
+            // Get the offset of the array, and check its valid
+            uint256 offset = uint256(bytes32(executionCalldata[0:32]));
+
+            // Array length is between offset and offset + 32. We check that this is in the buffer
+            // Since we know executionCalldata is at least 32, we can subtract with no overflow risk.
+            if (offset > bufferLength - 32) revert ERC7579DecodingError();
+
+            // Get the array length
```
frangio, line 191:
```suggestion
            // Get the array length. offset + 32 is bounded by bufferLength so does not overflow.
```

contracts/account/utils/draft-ERC7579Utils.sol
frangio: It would need to be generic on the return type so it's not possible to make a general helper.

The safety checks could be put in a helper, but it would need to be parameterized by the size of the array elements (actually the array element [heads](https://docs.soliditylang.org/en/latest/abi-spec.html#formal-specification-of-the-encoding)) so it's not possible to make it automatically fully safe.

contracts/account/utils/draft-ERC7579Utils.sol
frangio: The offsets inside each array member are apparently lazily validated by Solidity when you index the array. But we need to come up with a test for this to be sure.

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,34 @@ library ERC7579Utils {
...
-            // Extract the ERC7579 Executions
-            executionBatch.offset := add(ptr, 32)
-            executionBatch.length := calldataload(ptr)
+        unchecked {
+            uint256 bufferLength = executionCalldata.length;
+
+            // Check executionCalldata is not empty
+            if (bufferLength < 32) revert ERC7579DecodingError();
+
+            // Get the offset of the array, and check its valid
```
frangio, line 184: There is no validity check here

```suggestion
            // Get the offset of the array
```

frangio:
I added some tests, in particular one to show that it's fine to only do "shallow" validation, because elements deep in the array are validated later by Solidity.

contracts/account/utils/draft-ERC7579Utils.sol
frangio: Update: Added a test for this.

19714:
```suggestion
            assembly ("memory-safe") {
                executionBatch.offset := executionArray.offset
                executionBatch.length := arrayLength
            }
```

contracts/account/utils/draft-ERC7579Utils.sol
adamegyed, line 197: I think the check for the struct's static-sized elements should be performed by solidity lazily, as you access each `Execution` struct. To make this match the loose ABI-decoder exactly, I think at this stage the only necessary check would be:
```solidity
if (executionArray.length < arrayLength * 32)
```

So, essentially just checking that the struct pointers would still fit in the available space. To illustrate, here is a non-standard, but valid, encoding of 3 `Execution`s that would fail this check, but succeed in `abi.decode(...)` or in the built-in ABI decoder at the top of a function:

```
0x
0000000000000000000000000000000000000000000000000000000000000020    // Offset of `Execution[]`
0000000000000000000000000000000000000000000000000000000000000003    // Length of `Execution[]`
0000000000000000000000000000000000000000000000000000000000000060    // Offset of `executions[0]`
0000000000000000000000000000000000000000000000000000000000000060    // Offset of `executions[1]` (same as `executions[0]`)
0000000000000000000000000000000000000000000000000000000000000060    // Offset of `executions[2]` (same as `executions[0]`)
000000000000000000000000dcaccbf48225b2b2af81351fe52fbd0edf02adff 	// `executions[0].to`
0000000000000000000000000000000000000000000000000000000000000001	// `executions[0].value`
0000000000000000000000000000000000000000000000000000000000000060	// Offset of `executions[0].data`
0000000000000000000000000000000000000000000000000000000000000000	// Length of `executions[0].data`
```

It has all 3 of the offsets point to the same location in calldata, making the total size 7 words, with is less than the expected 9 words.

Either way, it's fine if you keep it as-is, just wanted to flag that this check is ever so slightly too strict.

Amxx: Pleas see https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5353/files#r1881380537

test/account/utils/draft-ERC7579Utils.t.sol
```diff
@@ -0,0 +1,327 @@
...
+    address private _owner;
+    uint256 private _ownerKey;
+    address private _account;
+    address private _beneficiary;
+    address private _recipient1;
+    address private _recipient2;
+
+    constructor() {
+        vm.etch(0x0000000071727De22E5E9d8BAf0edAc6f37da032, vm.readFileBinary("test/bin/EntryPoint070.bytecode"));
+        vm.etch(0xEFC2c1444eBCC4Db75e7613d20C6a62fF67A167C, vm.readFileBinary("test/bin/SenderCreator070.bytecode"));
```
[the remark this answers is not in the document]
Amxx: IMO we should not vendor the solidity code, and use the bin like this as a long term solution.

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,40 @@ library ERC7579Utils {
...
+            if (arrayLength > type(uint64).max) revert ERC7579DecodingError();
+
+            // Get the array as a bytes slice, and check its is long enough:
+            // - `96` is the size of `Execution`'s "head" representation. This is composed of
+            //   - 32 bytes for the `target` (address)
+            //   - 32 bytes for the `value` (uint256)
+            //   - 32 bytes for the offset pointer to the `callData` (bytes)
+            // - `arrayLength * 96` does not overflow because `arrayLength` is less than `2**64`.
+            bytes calldata executionArray = executionCalldata[offset + 32:];
+            if (executionArray.length < arrayLength * 96) revert ERC7579DecodingError();
```
Amxx, line 203: See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5353/files#r1881380537

Which I was able to confirm 
```
ethers.AbiCoder.defaultAbiCoder().encode(["(address,uint256,bytes)[]"], [[["0xA0c32d8A1B956Cfa83AeE3E6D2afBd0b9c2a188F", 1, "0x01234567"], ["0xA0c32d8A1B956Cfa83AeE3E6D2afBd0b9c2a188F", 2, "0x89abcdef"]]])


0x
0x0000 0000000000000000000000000000000000000000000000000000000000000020 // offset to the start of the array (0x0000 + 0x20 => 0x0020)
0x0020 0000000000000000000000000000000000000000000000000000000000000002 // length of the array (2)
[begining of the array]
0x0040 0000000000000000000000000000000000000000000000000000000000000040 // offset to element #0 (0x0040 + 0x40 = 0x0080) 
0x0060 00000000000000000000000000000000000000000000000000000000000000e0 // offset to element #1 (0x0040 + 0xe0 = 0x0120)
[element #0]
0x0080 000000000000000000000000a0c32d8a1b956cfa83aee3e6d2afbd0b9c2a188f // target
0x00a0 0000000000000000000000000000000000000000000000000000000000000001 // value
0x00c0 0000000000000000000000000000000000000000000000000000000000000060 // offset to calldata (0x0080 + 0x60 = 0x00e0)
[data for element #0]
0x00e0 0000000000000000000000000000000000000000000000000000000000000004 // length
0x0100 0123456700000000000000000000000000000000000000000000000000000000 // content
[element #1]
0x0120 000000000000000000000000a0c32d8a1b956cfa83aee3e6d2afbd0b9c2a188f // target
0x0140 0000000000000000000000000000000000000000000000000000000000000002 // value
0x0160 0000000000000000000000000000000000000000000000000000000000000060 // offset to calldata (0x0120 + 0x60 = 0x0180)
[data for element #1]
0x0180 0000000000000000000000000000000000000000000000000000000000000004 // length
0x01a0 89abcdef00000000000000000000000000000000000000000000000000000000 // content
```

```suggestion
            // Get the array as a bytes slice, and check its is long enough:
            // - each element of the array is an "offset pointer" to the data
            //   - each offset pointer takes 32 bytes
            //   - validity of the calldata at that location is checked when the array element is accessed.
            // - `arrayLength * 32` does not overflow because `arrayLength` is less than `2**64`.
            bytes calldata executionArray = executionCalldata[offset + 32:];
            if (executionArray.length < arrayLength * 32) revert ERC7579DecodingError();
```

This breaks some of the tests written by @frangio :/

Amxx: Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5353/commits/6aaa56acdebea0c2b113552b3a287619f0f94406

contracts/account/utils/draft-ERC7579Utils.sol
Amxx: fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5353/commits/6aaa56acdebea0c2b113552b3a287619f0f94406

frangio: Confirmed with the ABI spec that `arrayLength * 32` is the correct value.

test/account/utils/draft-ERC7579Utils.t.sol
```diff
@@ -310,18 +314,53 @@ contract ERC7579UtilsTest is Test {
...
+        // 0000000000000000000000000000000000000000000000000000000000000020 (32) element 0 offset
+        // 000000000000000000000000xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx (recipient) target for element #0
+        // 000000000000000000000000000000000000000000000000000000000000002a (42) value for element #0
+        // 0000000000000000000000000000000000000000000000000000000000000060 (96) offset to calldata for element #0
+        // 000000000000000000000000000000000000000000000000000000000000000c (12) length of the calldata for element #0
+        // 48656c6c6f20576f726c64210000000000000000000000000000000000000000 (..) buffer for the calldata for element #0
+        assertEq(
+            bytes("Hello World!"),
+            this.callDecodeBatchAndGetFirstBytes(
+                abi.encode(32, 1, 32, _recipient1, 42, 96, 12, bytes12(0x48656c6c6f20576f726c6421))
```
frangio, line 335:
```suggestion
                abi.encode(32, 1, 32, _recipient1, 42, 96, 12, bytes12("Hello World!"))
```

test/account/utils/draft-ERC7579Utils.t.sol
```diff
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.24;
 
+// Parts of this test file are adapted for Adam Egyed (@adamegyed) proof of concept available at:
```
frangio, line 4:
```suggestion
// Parts of this test file are adapted from Adam Egyed (@adamegyed) proof of concept available at:
```

test/account/utils/draft-ERC7579Utils.t.sol
```diff
@@ -298,8 +301,9 @@ contract ERC7579UtilsTest is Test {
 
         // GOOD
         this.callDecodeBatch(abi.encode(0));
-        // Note: Solidity also supports this even though it's odd
-        uint256[] memory _1 = abi.decode(abi.encode(0), (uint256[])); _1;
+        // Note: Solidity also supports this even though it's odd. Offset 0 means array is at the same location, which
+        // is interpreted as an array of length 0, which doesn't require any more data
+        abi.decode(abi.encode(0), (uint256[]));
```
frangio, line 306: I'm getting a warning "Statement has no effect". This is why I had declared a variable.

frangio (dismissed):

frangio (approved):

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -169,12 +172,38 @@ library ERC7579Utils {
...
-            executionBatch.offset := add(ptr, 32)
-            executionBatch.length := calldataload(ptr)
+        unchecked {
+            uint256 bufferLength = executionCalldata.length;
+
+            // Check executionCalldata is not empty.
+            if (bufferLength < 32) revert ERC7579DecodingError();
+
+            // Get the offset of the array.
+            uint256 offset = uint256(bytes32(executionCalldata[0:32]));
```
Amxx, line 185: Seppi proposed renaming that to avoid confusion. Any idea for a better name @ernestognw @frangio ? he proposed `offsetToArrayLength`, I think its quite long but if you think its the better option I won't oppose it.

ernestognw: I was thinking of `arrayOffset` but just read Seppi's feedback:

> but then arrayOffset points to the length, while array.offset points to the data, which I find confusing.

Let's use `arrayLengthOffset`

ernestognw (approved):

PR #5355, opened 2024-12-10
Corrected wording in the script  Update remove-ignored-artifacts.js

mdqst:
A minor but important typo in the script that removes build artifacts for ignored contracts. 

The message currently reads:
```javascript
console.error(`Removed ${n} mock artifacts`);
```
While the term "mock" is commonly used for test data, it is not the most accurate term in this case. The script is removing artifacts related to ignored contracts, so the word "ignored" better represents the action being performed.

The corrected line now reads:
```javascript
console.error(`Removed ${n} ignored artifacts`);
```

#### PR Checklist

- [ ] Tests
- [x] Documentation

### DIFF
--- a/scripts/remove-ignored-artifacts.js
+++ b/scripts/remove-ignored-artifacts.js
@@ -42,4 +42,4 @@ for (const filename of filenames) {
   }
 }
 
-console.error(`Removed ${n} mock artifacts`);
+console.error(`Removed ${n} ignored artifacts`);


Amxx:
`contracts/package.json` (that is used for releasing) contains
```json
[
  "**/*.sol",
  "/build/contracts/*.json",
  "!/mocks/**/*"
]
```
similarly `package.json` contains
```
[
  "/contracts/**/*.sol",
  "!/contracts/mocks/**/*"
]
```

In practice, the contracts that are ignored are all mocks. In the futur, I believe this feature will only ever be used for testing related contracts. So the wording is accurate in our case.

PR #5367, opened 2024-12-13
Cherrypick #5353 into release-v5.2

Amxx:
Will get into RC.1


Amxx:
DO not merge until #5371 is merged an cherrypicked

arr00 (approved):

Amxx:
Since we changing an unreleased feature, and there is no Changelog difference between the RC (this is not a patch), I don't think we should have a changelog entry for this. 

We'll fix that when we cleanup the changelog, otherwise it will mess up the list of changeset between here and master and it risk showing up again.

PR #5368, opened 2024-12-13
Remove relative path in solhint.config.js in favor of npm virtual package

Amxx:
This will allow us to simlink `solhint.config.js` into the community repo without having to also ship/vendor/link the corresponding module. The dependency will take care of that.

Note: we also need to make it explicit that `solhint-plugin-openzeppelin` uses `minimatch`, otherwize it won't be correctly linked when installed recursivelly as a sub-dependency of `@openzeppelin/contracts`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12203,7 +12203,10 @@
     "scripts/solhint-custom": {
       "name": "solhint-plugin-openzeppelin",
       "version": "0.0.0",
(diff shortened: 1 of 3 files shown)


ernestognw:
I just made sure that the package-lock was correct. LGTM if it passes the CI

ernestognw (approved):

PR #5370, opened 2024-12-13
fix:  Fix invalid usage of import.meta.dirname Update run.js

mdqst:
### Description

The code incorrectly uses `import.meta.dirname` to resolve the path to `specs.json`. However, `import.meta.dirname` does not exist in Node.js, leading to the error:

```
Cannot read property 'dirname' of undefined
```

This update replaces the invalid `import.meta.dirname` usage with the correct approach, combining `path.dirname` and `import.meta.url`. The updated logic ensures compatibility with Node.js and properly resolves the directory path.

### Changes made
- Updated the `default` value for the `spec` option to:
  ```javascript
  path.resolve(path.dirname(new URL(import.meta.url).pathname), 'specs.json')
  ```
  
### Why this change is necessary
The previous implementation would fail in environments using native ES modules in Node.js. The corrected approach leverages `import.meta.url`, which is the proper method to determine the current file's location, and ensures that the directory resolution is handled accurately.

### Testing
Tested the updated logic to confirm the path to `specs.json` resolves correctly in both standard Node.js environments and those using ES module syntax.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/certora/run.js
+++ b/certora/run.js
@@ -26,7 +26,7 @@ const argv = yargs(hideBin(process.argv))
     spec: {
       alias: 's',
       type: 'string',
-      default: path.resolve(import.meta.dirname, 'specs.json'),
+      default: path.resolve(path.dirname(new URL(import.meta.url).pathname), 'specs.json'),
     },
     parallel: {
       alias: 'p',


Amxx:
`import.meta.dirname` is valid since node 20

PR #5379, opened 2024-12-16
Bubble up revert data on failed `Address.sendValue`

### LINKED ISSUE #5184: `Address.sendValue` doesn't bubble errors
**📝 Details**

`Address.sendValue` ignores errors coming from the value recipient and throws a custom error on failure. This doesn't seem very useful, when a smart contract `receive` reverts, it probably has some important details to share about the failure. It seems that the `functionCall` error handling would be better here, bubble the original error and throw a generic one only if the contract has absolutely no details to share.

arr00:
Fixes #5184

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/cyan-taxis-travel.md
+++ b/.changeset/cyan-taxis-travel.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Address`: bubble up revert data on `sendValue` failed call
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -35,9 +35,9 @@ library Address {
             revert Errors.InsufficientBalance(address(this).balance, amount);
         }
 
-        (bool success, ) = recipient.call{value: amount}("");
+        (bool success, bytes memory returndata) = recipient.call{value: amount}("");
         if (!success) {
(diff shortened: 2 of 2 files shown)


CodeSandwich:
Would the solution below be inferior? It seems that it should be exactly equivalent while deduplicating code.
```
function sendValue(address payable recipient, uint256 amount) internal {
    functionCallWithValue(address(recipient), "", amount);
}
```

arr00:
> Would the solution below be inferior? It seems that it should be exactly equivalent while deduplicating code.
> 
> ```solidity
> function sendValue(address payable recipient, uint256 amount) internal {
>     functionCallWithValue(address(recipient), "", amount);
> }
> ```

`functioncallWithValue` in turn calls [`verifyCallResultFromTarget`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0034c302241c4b1a1685272d4df42ca5d64b8c34/contracts/utils/Address.sol#L106-L121) which doesn't allow calling EOAs.

CodeSandwich:
Ahh, good point, I missed that detail :+1:

Amxx (approved):

PR #5382, opened 2024-12-17
Set max old space size on coverage and test

### LINKED ISSUE #5363: Coverage and test runs often result in `JavaScript heap out of memory`
On MacOS, vanilla runs of the test and coverage npm scripts often results in a `JavaScript heap out of memory` error. This is a known issue with hardhat https://github.com/NomicFoundation/hardhat/issues/3471.

Increasing the allowed old memory to using the following env var fixes the issue `NODE_OPTIONS="--max-old-space-size=8192"`. We should consider documenting this or even setting the env var by default.

arr00:
Fixes #5363 

Hardhat has a known issue which results in `JavaScript heap out of memory` error unless the max old space size is set. This PR will automatically set it on required scripts (test and coverage).

Note: this fix only sets the `NODE_OPTIONS` `max-old-space-size` flag if it was not set already. Additional options in `NODE_OPTIONS` are maintained. The new flag is not maintained after the call is completed.

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ernestognw:
From [the docs](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-mib):

> --max-old-space-size=SIZE (in MiB)
> Sets the max memory size of V8's old memory section. As memory consumption approaches the limit, V8 will spend more time on garbage collection in an effort to free unused memory.
>
> On a machine with 2 GiB of memory, consider setting this to 1536 (1.5 GiB) to leave some memory for other uses and avoid swapping.
> 
> `node --max-old-space-size=1536 index.js`

I'm not sure what _swapping_ refers to, but I'm worried setting it to `8192` makes the "old space size" constant and perhaps too much in low-capability devices. It doesn't say whether the default is calculated depending on the host, but I assume so based on [this comment](https://github.com/nodejs/node/issues/7937#issuecomment-269997873).

Amxx (commented):
- why 4096 and not more ? In the past (before the hardhat migration to EDR 4096 was not enough)
- why only on coverage and not on test ?

arr00:
> * why 4096 and not more ? In the past (before the hardhat migration to EDR 4096 was not enough)
> 
>     * why only on coverage and not on test ?

I'm testing on multiple machines (including CI) to see what we need. Ideally it would be below 8192 so that 8gb machines could run coverage. The test run requires significantly less memory to the point that the default is generally sufficient. We could set it to something like 4096 to just be safe though.

package.json
```diff
@@ -25,7 +25,7 @@
     "prepack": "scripts/prepack.sh",
     "generate": "scripts/generate/run.js",
     "version": "scripts/release/version.sh",
-    "test": "hardhat test",
+    "test": "NODE_OPTIONS=\"--max-old-space-size=3072\" hardhat test",
```
Amxx, line 28: this looks very arbitrary

Amxx (commented):
I'm really not a big fan of this change:
- It fixes a value that may be to-much to to little depending on the evolution of the tests and tooling. Hardhat evolution did lower the needed value in the past. Same is true of coverage. Also, as we add more tests this could become outdated
- Different machines (hardware, os) will have different things available.

If our estimation is too tight, there is a risk we have to update it to frequently. If the estimation is not tight enough, it will some "small" machines

IMO this value should be controlled by the host of the machine. Not by us. And an issue with what this PR proposes is that it doesn't allow the machine's host to override the value we fix without editing the package.json.

package.json
```diff
@@ -25,7 +25,7 @@
     "prepack": "scripts/prepack.sh",
     "generate": "scripts/generate/run.js",
     "version": "scripts/release/version.sh",
-    "test": "hardhat test",
+    "test": "NODE_OPTIONS=\"--max-old-space-size=8192\" hardhat test",
```
Amxx, line 28: Can we remove that from the tests? 

My understanding is that it was only really needed for the coverage ... and we now have an elegant integration in coverage.sh.

arr00: What are your thoughts on the current state?

.github/workflows/checks.yml
```diff
@@ -13,9 +13,6 @@ concurrency:
   group: checks-${{ github.ref }}
   cancel-in-progress: true
 
-env:
-  NODE_OPTIONS: --max_old_space_size=8192
-
```
Amxx, line 18: I'd keep this as the CI config. If this is ever modified, the script would not override it.

scripts/checks/coverage.sh
```diff
@@ -4,11 +4,8 @@ set -euo pipefail
 
 export COVERAGE=true
 export FOUNDRY_FUZZ_RUNS=10
-export NODE_OPTIONS="${NODE_OPTIONS:-}"
 
-if [[ $NODE_OPTIONS != *"--max-old-space-size"* ]]; then
-  export NODE_OPTIONS="${NODE_OPTIONS} --max-old-space-size=8192"
-fi
+eval "scripts/set-max-old-space-size.sh"
```
Amxx: do we need an eval here ?
```suggestion
scripts/set-max-old-space-size.sh
```

arr00: This was already updated the line is now
```
. ./scripts/set-max-old-space-size.sh
```

Amxx (approved):

PR #5392, opened 2024-12-19
Fix envvar setting for test and coverage

Amxx:


arr00 (approved):

0xmichalis:
Sourcing ignores the script shebang so in shell terminals this doesn't work. Any arguments against removing bash support altogether (https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5493)?

PR #5393, opened 2024-12-20
fix: fix incorrect usage of import.meta.dirname in Node.js

0xwitty:
The current implementation references `import.meta.dirname`, which is not a valid property in Node.js. Instead, the correct approach is to use `import.meta.url` and convert it using `path`.  

I’ve updated the code to properly handle this conversion, ensuring compatibility and correctness. 

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/certora/run.js
+++ b/certora/run.js
@@ -26,7 +26,7 @@ const argv = yargs(hideBin(process.argv))
     spec: {
       alias: 's',
       type: 'string',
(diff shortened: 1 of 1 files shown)


Amxx:
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5370

PR #5400, opened 2024-12-23
ERC7579.decodeBatch: Perform sanity check if executionCalldata is not at the end of allocated calldata

### LINKED ISSUE #5395: Possible issue in `ERC7579Utils.decodeBatch`
Recently, v5.2-rc.1 included a fix to a known issue in `ERC7579Utils.decodeBatch`. The issue was identified as originating from missing checks that solidity automatically inserts when doing a `decode` operation. These checks were not present in v5.2-rc.0

The way these checks are currently implemented is as follows:

- when decoding `bytes` (in calldata) as an array, we check that the `bytes` is long enough so that all the "first-level elements" of the array fit inside the `bytes`. Because the array contains dynamic objects, the "first-level elements" are only pointers (offsets) to the actual elements.
- during the decoding, it is NOT checked that the the target of the pointers falls inside the calldata `bytes` object. It is only when dereferencing these "pointers" that solidity checks weither they are "valid".

However, solidity does that by checking that the dereferenced object is in the calldata as defined by `calldatasize`. This means that the pointers could target data that is outside the`bytes` object while still being in allocated calldata.

This could potentially cause issues, though it is not 100% clear how this could be exploited and what the consequences would be. Currently no POC of this bad behavior exists.

Signers of ERC-7579 batch operations (usually within the scope of of an ERC-4337 user operation) should verify that the calldata they sign is properly formed. A possible adversarial senario would exist if the signer is a session key with limited access to the account. This issue, if confirmed, could potentially cause the validation and execution phases to missmatch, allowing the session key to perform restricted operations.

Amxx:
For context, the previous work on securing the decodeBatch function https://github.com/OpenZeppelin/openzeppelin-contracts/commit/77d4a73c87ba0421dd2f05cf2137f5e6de239fa3

Fixes #5395

#### PR Checklist

- [x] Tests
- [ ] Documentation


frangio (dismissed):

contracts/account/utils/draft-ERC7579Utils.sol
```diff
@@ -204,9 +209,37 @@ library ERC7579Utils {
...
-                executionBatch.offset := add(add(executionCalldata.offset, arrayLengthOffset), 32)
+                executionBatch.offset := add(add(bufferPtr, arrayLengthOffset), 32)
                 executionBatch.length := arrayLength
             }
+
+            _validateCalldataBound(executionBatch, bufferPtr + bufferLength);
+        }
+    }
+
+    /** @dev Calldata sanity check
```
Amxx, line 220:
```suggestion
    /** 
     * @dev Calldata sanity check
```

ernestognw (approved):

ernestognw:
Let's pull the trigger on this one. A recent audit on ONCHAINID revealed a critical issue because decoding is performed differently between validation and execution, which justifies merging this PR. In the past, we weren't sure it could be a problem

PR #5403, opened 2024-12-24
fix: typos in documentation files

vtjl10:
<Corrected `warrantees` to `warranties`
Corrected `recomment` to `recommend`


ernestognw:
We've decided previous audits won't be altered for typos. Closing

PR #5408, opened 2024-12-27
Await `.eventually` test matchers

ernestognw:
Mirror of https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/54

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/account/utils/draft-ERC4337Utils.test.js
+++ b/test/account/utils/draft-ERC4337Utils.test.js
@@ -27,7 +27,7 @@ describe('ERC4337Utils', function () {
       const validAfter = 0x9abcdef0n;
       const validationData = packValidationData(validAfter, validUntil, authorizer);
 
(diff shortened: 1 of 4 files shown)


arr00:
If we have to await anyways, what's the point in using `eventually`? Is it preferred for the await to be at the very start of the statement instead of inside the expect?

ernestognw:
> If we have to await anyways, what's the point in using `eventually`? Is it preferred for the await to be at the very start of the statement instead of inside the expect?

I'd personally prefer it. Basically we have 2 options:

```
expect(await something()).to.equal(...)
await expect(something).to.eventually.equal(...)
```

One thing we could say is that we might prefer consistency. I agree, but that's for another PR imo

arr00 (approved):

Amxx (approved):

PR #5412, opened 2024-12-30
Remove unused rounding modes `Trunc` and `Expand`

### LINKED ISSUE #5411: Simplify Rounding Modes
**🧐 Motivation**
1. Remove unused API
2. Extend existing API

**📝 Details**
The Rounding-Modes API (in the Math library) currently includes:
```
    enum Rounding {
        Floor, // Toward negative infinity
        Ceil, // Toward positive infinity
        Trunc, // Toward zero
        Expand // Away from zero
    }
```
The last two options are different from the first two options only when dealing with signed input.
However, throughout the codebase, rounding-modes are used strictly for unsigned input.
Hence as it stands, the last two options are redundant in the current version.

My initial motivation was adding a `Nearest` mode, which for non-negative (unsigned) input stands for:
- Round down (towards zero) any value with remainder < 0.5
- Round up (towards infinity) any value with remainder >= 0.5

Adding this new mode is questionable (and a lot more complicated) when the enum at hand is "seemingly" implemented to support both signed and unsigned input. By getting rid of what appears to be redundant (unused) API, it might make it easier to add this new mode. But even regardless of this new mode, it makes sense to clean-up the unused part of this API.

UPDATE:
---

Note that a more complete fix might actually be to get rid of the Rounding-Mode API altogether, and split each function which takes it as input into two functions - a floor function and a ceil function.

Reason being - it would save gas for the caller:
- For the floor functions, checking the rounding mode AND calculating the additional info will be omitted
- For the ceil functions, checking the rounding mode will be omitted

Of course, with 4 different rounding modes, this might "bloat" the Math library's code.
But if you choose to accept my proposal and reduce it to 2 different rounding modes, then taking it one step forward and splitting each one of the associated functions into a pair of functions might seem like an option for you to consider.

barakman:
Fixes #5411

In the Rounding-Modes API (in the Math library):
```
    enum Rounding {
        Floor, // Toward negative infinity
        Ceil, // Toward positive infinity
        Trunc, // Toward zero
        Expand // Away from zero
    }
```
Remove the last two options, which are not in use throughout the codebase.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -11,10 +11,8 @@ import {SafeCast} from "./SafeCast.sol";
  */
 library Math {
     enum Rounding {
-        Floor, // Toward negative infinity
-        Ceil, // Toward positive infinity
-        Trunc, // Toward zero
-        Expand // Away from zero
+        Floor, // Towards zero
+        Ceil // Towards infinity
     }
 
     /**
@@ -226,7 +224,7 @@ library Math {
      * @dev Calculates x * y / denominator with full precision, following the selected rounding direction.
      */
     function mulDiv(uint256 x, uint256 y, uint256 denominator, Rounding rounding) internal pure returns (uint256) {
-        return mulDiv(x, y, denominator) + SafeCast.toUint(unsignedRoundsUp(rounding) && mulmod(x, y, denominator) > 0);
+        return mulDiv(x, y, denominator) + unsignedRoundsUp(rounding, mulmod(x, y, denominator) > 0);
     }
 
     /**
@@ -529,7 +527,7 @@ library Math {
     function sqrt(uint256 a, Rounding rounding) internal pure returns (uint256) {
         unchecked {
             uint256 result = sqrt(a);
-            return result + SafeCast.toUint(unsignedRoundsUp(rounding) && result * result < a);
+            return result + unsignedRoundsUp(rounding, result * result < a);
         }
     }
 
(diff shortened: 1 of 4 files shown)


barakman:
> ### ⚠️ No Changeset found
> Latest commit: [6a441be](https://github.com/OpenZeppelin/openzeppelin-contracts/commit/6a441be40547edc09241c38f4db349b7d70011ad)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> 
> This PR includes no changesets
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/barakman/zeppelin-solidity/new/simplify-rounding-modes?filename=.changeset/tidy-lamps-rule.md&value=---%0A%22openzeppelin-solidity%22%3A%20patch%0A---%0A%0ARemove%20unused%20rounding%20modes%20%60Trunc%60%20and%20%60Expand%60%0A)

This PR introduces an API-breaking change, so in theory, the changeset should include a major-version update.
On the other hand, the changed part in the API is not used anywhere in the current version, so it is possible that a minor-version update or even a patch-version update will suffice.

Amxx:
Hello @barakman 

Rounding mode `Trunc` and `Expand` have been introduced from the start so that we couldn't unambiguously expand the library to signed integers in the future. While this is not (yet) available in our library, there may be code that depends on our library to define these operation.

Removing enum entries is a breaking change, and we don't allow such modification without a very strong reason.

barakman:
> Removing enum entries is a breaking change, and we don't allow such modification without a very strong reason.

The reasons are mention in the associated issue (#5411), as by changing this API to support only unsigned input, it becomes easier (and cleaner) to:
1. Support an additional rounding mode, namely, round-to-nearest
2. Get rid of the enum altogether, thus reducing the gas consumption of all related functions

Whether these are strong reasons is subjected to opinion, which I will leave for your dev team to decide.

Amxx:
Lets discuss all that in the issue before doing any breaking implementation.

PR #5413, opened 2025-01-04
fix: Fixing issue with import.meta.dirname usage Update run.js

annwag:
I noticed a problem with the code where `import.meta.dirname` was used in the following line:

```javascript
default: path.resolve(import.meta.dirname, 'specs.json'),
```

It turns out that `import.meta.dirname` is not supported in JavaScript, and this leads to an error when the code is executed. To resolve this, I replaced `import.meta.dirname` with `__dirname`, which is the correct way to reference the current directory in Node.js.

This should fix the issue and prevent errors related to the incorrect usage of `import.meta.dirname`.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/certora/run.js
+++ b/certora/run.js
@@ -26,7 +26,7 @@ const argv = yargs(hideBin(process.argv))
     spec: {
       alias: 's',
       type: 'string',
-      default: path.resolve(import.meta.dirname, 'specs.json'),
+      default: path.resolve(__dirname, 'specs.json'),
     },
     parallel: {
       alias: 'p',


ernestognw:
Hey @annwag, thanks for your contribution. However, the `import.meta.dirname` is recommended for ES modules and `_dirname` is not available there. We're keeping the current change but we're definitely interested in reviewing an issue if this is not running for some reason.

We'll remove this script eventually too. Closing

PR #5416, opened 2025-01-06
Expose `_isTrustedByTarget` internally in ERC2771Forwarder

### LINKED ISSUE #5375: ERC2771Forwarder: consider not enforcing trustedForwarder
**🧐 Motivation**
When relaying transactions, you often end up in a situation where you need to relay a token `permit` in addition to some additional transaction(s). While the additional transaction(s) can be forwarded via `ERC2771Forwarder` - considering the target contract implements `ERC2771Context`- the permit can not because the `token` is not "trusting" this forwarder.
However, **not** trusting the forwarder is not a problem as the permit does not rely on altering the context.

For any flow that requires an approval in order to forward the rest of a transaction batch this would be very useful.

**📝 Details**
As far as i can tell the `Forwarder` itself is not part of the Spec, so i think it's worth considering to remove the [`isTrustedForwarder`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/metatx/ERC2771Forwarder.sol#L278C13-L278C31) check as I don't actually see any downside of not having it - worst case the txn would revert. This is in line with how the previous [`MinimalForwarder`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3d7a93876a2e5e1d7fe29b5a0e96e222afdc4cfa/contracts/metatx/MinimalForwarder.sol) worked.

edit: I see this was changed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4502#discussion_r1282385480 specifically to prevent accidents caused by approving the forwarder, but the fix as lined out above, can actually cause problems.

ernestognw:
Fixes #5375 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -302,8 +302,12 @@ contract ERC2771Forwarder is EIP712, Nonces {
      *
      * This function performs a static call to the target contract calling the
      * {ERC2771Context-isTrustedForwarder} function.
+     *
+     * NOTE: Consider the execution of this forwarder is permissionless. Without
+     * this check, users may approve token spending to this contract allowing
+     * any relayer to transfer their tokens.
      */
-    function _isTrustedByTarget(address target) private view returns (bool) {
+    function _isTrustedByTarget(address target) internal view virtual returns (bool) {
         bytes memory encodedParams = abi.encodeCall(ERC2771Context.isTrustedForwarder, (address(this)));
 
         bool success;
--- a/contracts/metatx/ERC2771Forwarder.sol
+++ b/contracts/metatx/ERC2771Forwarder.sol
@@ -303,9 +303,8 @@ contract ERC2771Forwarder is EIP712, Nonces {
      * This function performs a static call to the target contract calling the
      * {ERC2771Context-isTrustedForwarder} function.
      *
-     * NOTE: Consider the execution of this forwarder is permissionless. Without
-     * this check, users may approve token spending to this contract allowing
-     * any relayer to transfer their tokens.
+     * NOTE: Consider the execution of this forwarder is permissionless. Without this check, anyone may transfer assets
+     * that are owned by, or are approved to this forwarder.
      */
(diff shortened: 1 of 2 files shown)


contracts/metatx/ERC2771Forwarder.sol
```diff
@@ -302,8 +302,12 @@ contract ERC2771Forwarder is EIP712, Nonces {
      *
      * This function performs a static call to the target contract calling the
      * {ERC2771Context-isTrustedForwarder} function.
+     *
+     * NOTE: Consider the execution of this forwarder is permissionless. Without
+     * this check, users may approve token spending to this contract allowing
+     * any relayer to transfer their tokens.
```
Amxx, line 308:
```suggestion
     * NOTE: Consider the execution of this forwarder is permissionless. Without this check, anyone may transfer asset
     * that are owned by, or are approved to, this forwarder.
```

Amxx: We had that discussion a while back, but I honestly don't think that is such a critical security check.

It protects against "non ERC-2771 aware" contract being called by the forwarder, which could be abused if the forwarder is owner of something (contract, assets, allowance, ...).

But that should never happen. A forwarder should not receive any tokens of special permission, outside of being a trusted forwarder. Moreover:
- if assets on an "non ERC-2771 aware" contract are held by the forwarder, they are currently frozen. Is it worst to have them available to anyone that to have them frozen? I'm not sure.
- if assets on an "ERC-2771 aware" contract are held by the forwarder, they are frozen, and this check presence/absence doesn't change anything about it.
- there is always the case of `contract X is SomethingThatHoldsToken, ERC2771Forwarder {}` but that feels like a crazy footgun to mix purposes in a single contract like that.

So basically, I think the current check does more harm (preventing some batching) than it does good.
Howefully this entire contract goes away in 6.0

ernestognw: We can consider removing the check right away in 6.0. The current version includes just a note so I think that should be just fine.

I'm updating a bit your suggestion but generally agree

Amxx (approved):

PR #5431, opened 2025-01-13
fix: Fix incorrect symbols for success/error Update run.js

0xminds:
I noticed that in the `writeEntry` function, the symbols for displaying the result of a check are used incorrectly. Currently, `:x:` (cross) is shown for success, and `:heavy_check_mark:` (checkmark) is shown for errors.

This is counterintuitive, as a cross typically represents an error, and a checkmark represents success.

I’ve swapped these symbols to align with common conventions and improve clarity.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/certora/run.js
+++ b/certora/run.js
@@ -160,7 +160,7 @@ function writeEntry(spec, contract, success, url) {
     formatRow(
       spec,
       contract,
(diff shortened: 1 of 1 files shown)


Amxx:
ALready fixed in 5445

PR #5439, opened 2025-01-19
fix: Fix variable scope in tests to ensure isolation Update sanity.te…

voronor:
I moved the `blockNumberBefore` variable declaration inside each `it` block. This ensures that every test runs with an isolated state, preventing potential issues caused by shared state between tests. Tests are now independent and more reliable.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/test/sanity.test.js
+++ b/test/sanity.test.js
@@ -12,15 +12,14 @@ describe('Environment sanity', function () {
   });
 
   describe('snapshot', function () {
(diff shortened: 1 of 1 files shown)


Amxx (changes requested):
This change is actually breaking the purpose of the test. 

Isolating the variable would cause the test to pass even if the snapshot mechanism doesn't work.

PR #5440, opened 2025-01-19
fix: Fix mkdirp function Update prepack.sh

voronor:
I noticed that the `mkdirp` function uses `node -e` to create directories, which can fail if the path contains special characters like spaces or quotes. I've updated it to use the built-in `mkdir -p` command instead, which handles these cases correctly. This should resolve the issue and make the function more robust.

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/scripts/prepack.sh
+++ b/scripts/prepack.sh
@@ -5,7 +5,7 @@ shopt -s globstar
 
 # cross platform `mkdir -p`
 mkdirp() {
-  node -e "fs.mkdirSync('$1', { recursive: true })"
+  mkdir -p "$1"
 }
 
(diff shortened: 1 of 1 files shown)


Amxx:
@ernestognw @arr00 I'm ok with that change, but I'd like to confirm that if works on MacOS/other

Amxx (approved):

Amxx (changes requested):
As documented in the code, this function (`mkdirp`) acts as a 

```
# cross platform `mkdir -p`
```

It was added to circumvent issues with `mkdir -p` not working properly on some platforms.

PR #5442, opened 2025-01-20
Add Bytes32x2Set

Amxx:
- Refactor EnumerableSet.opt.js to support arrays of basetype
  - doesn't affect currently generated code (backward compatible)
- Use the refactor to implement Bytes32x2Set

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/ten-peas-mix.md
+++ b/.changeset/ten-peas-mix.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': minor
 ---
 
-`Hashes`: expose `efficientKeccak256` for hashing non-commutative pairs of bytes32.
+`Hashes`: Expose `efficientKeccak256` for hashing non-commutative pairs of bytes32 without allocating extra memory.
--- a/.changeset/ten-peas-mix.md
+++ b/.changeset/ten-peas-mix.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Hashes`: expose `efficientKeccak256` for hashing non-commutative pairs of bytes32.
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -4,6 +4,8 @@
 
 pragma solidity ^0.8.20;
 
+import "../cryptography/Hashes.sol";
+
 /**
  * @dev Library for managing
  * https://en.wikipedia.org/wiki/Set_(abstract_data_type)[sets] of primitive
@@ -481,6 +483,6 @@ library EnumerableSet {
     }
 
     function _hash(bytes32[2] memory value) private pure returns (bytes32) {
-        return keccak256(abi.encode(value));
+        return Hashes.efficientKeccak256(value[0], value[1]);
     }
 }
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -372,4 +372,115 @@ library EnumerableSet {
 
         return result;
     }
+
+    struct Bytes32x2Set {
+        // Storage of set values
+        bytes32[2][] _values;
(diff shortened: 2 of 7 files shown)


contracts/utils/structs/EnumerableSet.sol
```diff
@@ -372,4 +374,115 @@ library EnumerableSet {
 
         return result;
     }
+
+    struct Bytes32x2Set {
+        // Storage of set values
+        bytes32[2][] _values;
+        // Position is the index of the value in the `values` array plus 1.
+        // Position 0 is used to mean a value is not in the self.
+        mapping(bytes32 valueHash => uint256) _positions;
```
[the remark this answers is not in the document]
Amxx: I looked into the history of this piece of code:

1. `EnumerableSet` was introduced in 2.5. In that version, the code only supported addresses, and the structure was defined as:
```solidity
    struct AddressSet {
        // Position of the value in the `values` array, plus 1 because index 0
        // means a value is not in the set.
        mapping (address => uint256) index;
        address[] values;
    }
```

2. In version 3.0, the inner/outer structure was used, with the default being a bytes32 set that is 
```solidity
    struct Set {
        // Storage of set values
        bytes32[] _values;

        // Position of the value in the `values` array, plus 1 because index 0
        // means a value is not in the set.
        mapping (bytes32 => uint256) _indexes;
    }
```

3. In version 5.0, the naming was changed to was we have currently ([this is the PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/4577)). Note that the PR was the result of an audit issue.

4. This PR is just reusing the wording that we have in the other sets.

IMO:
- No user should touch that storage (if we could make it private we would)
- The description is really clear
- Overall, I tend to favor short names to long ones. IMO `_positionsPlusOne` is difficult/long to read and offer little benefit.
- Changing the name in the existing set would be breaking, and using a different name here would be inconsistent.

So I'd personally keep it that way. @ernestognw WDYT ?

scripts/generate/templates/EnumerableSet.js
```diff
@@ -233,14 +235,131 @@ function values(${name} storage set) internal view returns (${type}[] memory) {
...
+ */
+function values(${name} storage self) internal view returns (${type}[] memory) {
+    return self._values;
+}
+`;
+
+const hashes = `\
+function _hash(bytes32[2] memory value) private pure returns (bytes32) {
+    return Hashes.efficientKeccak256(value[0], value[1]);
+}
```
arr00, line 350: Not generalized here. Ow good. Not blocking for me.

Amxx: Yes.

The generic way would be to do keccak256(Abi.encode(value)) ... but that would leak memory.

My plan was to use the efficient hash here, and if we ever need we can add other hashes in the future.

arr00: That makes sense.

arr00 (approved):

.changeset/ten-peas-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Hashes`: expose `efficientKeccak256` for hashing non-commutative pairs of bytes32.
```
ernestognw, line 5:
```suggestion
`Hashes`: Expose `efficientKeccak256` for hashing non-commutative pairs of bytes32 without allocating extra memory.
```

contracts/utils/structs/EnumerableSet.sol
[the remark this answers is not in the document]
ernestognw: I also agree that it's clearer to have `_positions`. Otherwise it makes it difficult to reason the "plus one" since the position is a hash and (pseudo)randomly distributed

ernestognw (dismissed):

arr00 (approved):

PR #1403, opened 2018-10-10
Non-constructor initialization removed

Aniket-Engg:
Fixes #1391

Issue #1391, opened 2018-10-07
Disallow non-constructor initialization


nventuro:
Some of our contracts (like `ReentrancyGuard`) initialize their state variables outside of a constructor (i.e. in their declaration). We should avoid doing this and move all initialization to the constructor, mainly for clarity, but also because this makes it easier to transform said constructors to initialization functions.

PR #5448, opened 2025-01-21
Remove Unnecessary Initialisation of `_paused`

pcaversaccio:
This PR removes the unnecessary initialisation of `_paused`. The state variable is already set at compile-time to `false` here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/332bcb5f4d9cf0ae0f98fe91c77d9c1fb9951506/contracts/utils/Pausable.sol#L18

Not sure if this requires a changeset. Feel free to push one.

### DIFF
--- a/.changeset/green-drinks-report.md
+++ b/.changeset/green-drinks-report.md
@@ -1,0 +1,5 @@
+---
+"openzeppelin-solidity": minor
+---
+
+`Pausable`: Stop explicitly setting `paused` to `false` during construction.
--- a/contracts/utils/Pausable.sol
+++ b/contracts/utils/Pausable.sol
@@ -38,13 +38,6 @@ abstract contract Pausable is Context {
     error ExpectedPause();
 
     /**
-     * @dev Initializes the contract in unpaused state.
-     */
-    constructor() {
-        _paused = false;
-    }
-
-    /**
      * @dev Modifier to make a function callable only when the contract is not paused.
      *
      * Requirements:


ernestognw:
Hi @pcaversaccio,

This change makes sense but the code pre-dates me so I wonder if I'm missing something. Also, this is strictly breaking but I wouldn't worry too much since it's easy to workaround and doesn't break smart contract upgrades.

Let's see if @Amxx has something to add.

Amxx:
It also pre-dates me 😄 

To me there are 2 things to consider:
- "normal" contracts: the storage should definitelly be empty at construction, so setting it to false should have no effect. A good optimiser may remove that sstore, though I'm not sure it is the case. 
  - The price of warming up the slot, an writting 0 to it, when it already contains 0, is 2200 gas (if the slot is not shared)
- upgradeable contracts: if Pausing is added during an upgrade, the slot may be non-zero. In that case an initializer would do the clean up so that there is no risk the contract is paused by default.
  - With namespace storage, the likelyhood of this happening is extremelly low !
  
I'd be ok with that change. But I'm also curious if @frangio knows more about what we have that in the first place ?

Amxx:
Capture d’écran du 2025-01-23 10-44-21

pcaversaccio:
I did some research. The PR that added the change: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/1403 & the issue from Nico: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/1391. Previously it was like:

```solidity
bool private _paused = false;
```

And since we all like memes and we should have more fun in life, I simply drop this here 😄:

image

ernestognw:
> upgradeable contracts: if Pausing is added during an upgrade, the slot may be non-zero. In that case an initializer would do the clean up so that there is no risk the contract is paused by default.
With namespace storage, the likelyhood of this happening is extremelly low !

I was thinking something similar but I also don't see it happening with namespace storage.

ernestognw (approved):

frangio:
>`bool private _paused = false;`


Yeah this was the original code and it was just meant to be explicit and carried over until today. I think it's fine to remove it.

arr00 (approved):

PR #5456, opened 2025-01-23
Remove unused `setBaseURI` tests

RenanSouza2:
- [X] Tests
- [X] Documentation


Amxx:
Note: neither the core ERC721 contract nor any of its extensions include a `setBaseURI` function. It is only available in `contracts/mocks/token/ERC721URIStorageMock.sol` to test the impact of `_baseURI()` on the derived URI of tokens.

This is tested in `test/token/ERC721/extensions/ERC721URIStorage.test.js` directly (without using the behavior).

Therefore, the behavior is always executed on mocks that do NOT include a mechanism to update the `_baseURI` after construction. This means the section of tests deleted in this PR is in fact never executed (always skipped)

Amxx (approved):

ernestognw (approved):
Thanks @RenanSouza2

Amxx:
Capture d’écran du 2025-01-24 17-47-48

PR #5457, opened 2025-01-24
ERC-7540: Asynchronous ERC-4626 Tokenized Vaults

DeeJayElly:
Fixes #???? 

## 📌 Summary  
This PR introduces an **implementation of ERC-7540**, which extends ERC-4626 with **asynchronous deposit and redemption flows**. The contract provides the ability to request deposits and redemptions asynchronously, ensuring compliance with the **EIP-7540 specification**.

---

### 1️⃣ ERC-7540 Implementation  
- Extends `ERC4626` to **support asynchronous deposits and redemptions**.
- Introduces **request-based flows** via `requestDeposit()` and `requestRedeem()`, ensuring **separation between pending, claimable, and claimed states**.
- Implements **claiming of assets/shares** through the `deposit()` and `redeem()` methods.

### 2️⃣ Compliance with ERC-7540 Specification  
✅ **Request Lifecycle**  
- Introduces `pendingDepositRequest()`, `pendingRedeemRequest()`, `claimableDepositRequest()`, and `claimableRedeemRequest()`.  
- Ensures **requests must transition from Pending → Claimable → Claimed**, per EIP-7540.  
- Implements **cumulative requests** (requests in the same block are merged).  

✅ **ERC-4626 Overrides**  
- `deposit()` and `mint()` **do not transfer assets** (they use `requestDeposit()` instead).  
- `redeem()` and `withdraw()` **do not burn shares** (handled by `requestRedeem()`).  
- `previewDeposit()`, `previewMint()`, `previewRedeem()`, and `previewWithdraw()` **revert as required**.  

✅ **ERC-165 Compliance**  
- Implements `supportsInterface()` to declare support for **ERC-7540, ERC-7575, and ERC-165**.  
- Returns `true` for `0xe3bc4e65` (ERC-7540 operators) and `0x2f0a18c5` (ERC-7575 interface).  

✅ **Max Deposit / Redeem Logic**  
- `maxDeposit()` and `maxRedeem()` now return the **claimable amount**, ensuring correct behavior.  

✅ **Operator Management**  
- Implements `setOperator()` and `isOperator()`, allowing controllers to manage requests on behalf of others.  

---

📜 References
[ERC-7540 Proposal](https://eips.ethereum.org/EIPS/eip-7540)
[ERC-4626 Standard](https://eips.ethereum.org/EIPS/eip-4626)
OpenZeppelin Vaults Documentation

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/interfaces/IERC7540.sol
```diff
@@ -0,0 +1,102 @@
...
+     */
+    function pendingRedeemRequest(uint256 requestId, address controller) external view returns (uint256 shares);
+
+    /**
+     * @dev Gets the claimable redeem request amount for a given controller and requestId.
+     * @param requestId The unique identifier for the request.
+     * @param controller The address of the controller.
+     * @return shares The amount of shares in the claimable state.
+     */
+    function claimableRedeemRequest(uint256 requestId, address controller) external view returns (uint256 shares);
```
Amxx, line 85: Ordering could be more consistent. 

Either:
- requests (deposit + redeem)
- pending (deposit + redeem) request
- claimable (deposit + redeem)

Or

- deposit (request + pending request + claimable)
- redeem (request + pending request + claimable)

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,130 @@
...
+import {IERC20} from "../../../interfaces/IERC20.sol";
+import {ERC4626} from "./ERC4626.sol";
+import {SafeERC20} from "../utils/SafeERC20.sol";
+import {Math} from "../../../utils/math/Math.sol";
+import {ReentrancyGuard} from "../../../utils/ReentrancyGuard.sol";
+
+/**
+ * @dev Abstract implementation of the ERC-7540 standard, extending ERC-4626.
+ */
+abstract contract ERC7540 is ERC4626, IERC7540, ReentrancyGuard {
```
Amxx, line 15: Is reentrancy guard really necessary? None of the contract in the codebase use it. We always make sure that behaviour is safe without it.

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,130 @@
...
+    using SafeERC20 for IERC20;
+    using Math for uint256;
+
+    struct Request {
+        uint256 amount;
+        uint256 claimable;
+    }
+
+    // Mappings to track pending and claimable requests
+    mapping(address => mapping(uint256 => Request)) internal _pendingDepositRequests;
```
Amxx, line 25: Please see https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/GUIDELINES.md#solidity-conventions

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * @dev Creates a new deposit request.
+     */
+    function requestDeposit(
+        uint256 assets,
+        address controller,
+        address owner
+    ) external override nonReentrant returns (uint256 requestId) {
+        require(assets > 0, "ERC7540: assets must be greater than zero");
```
Amxx, line 43: It would be better if asset == 0 did not revert, but instead was a no-op.

(see how ERC-20 transfers of amount 0 are supported)

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,130 @@
...
+        require(owner == msg.sender || isOperator(owner, msg.sender), "ERC7540: unauthorized");
```
Amxx, line 44: Two things here:

- we avoid using `msg.sender` and instead use `_msgSender()` provided by `Context`. This is necessary for supporting ERC-2771meta-tx.
- We avoid using revert strings. Instead, we only use custom errors that can give more context. Please see [ERC-6093](https://eips.ethereum.org/EIPS/eip-6093)

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,130 @@
...
+     */
+    function isOperator(address controller, address operator) public view override returns (bool) {
+        return _operators[controller][operator];
+    }
+
+    /**
+     * @dev Internal function to generate a unique request ID.
+     */
+    function _generateRequestId(address controller, uint256 input) internal view returns (uint256) {
+        return uint256(keccak256(abi.encodePacked(block.timestamp, block.number, msg.sender, controller, input)));
```
Amxx, line 123: Using both `block.timestamp` and `block.number` is really a duplication of the same thing. We we ever expect one to change but not the other?

Note: on L2, block.number is more reliable.

Regardless, this function helps with uniqueness, but doesn't provide any real guarantee. While it is more expensive, maybe using a counter would be the better option.

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,130 @@
...
+        address owner
+    ) external override nonReentrant returns (uint256 requestId) {
+        require(assets > 0, "ERC7540: assets must be greater than zero");
+        require(owner == msg.sender || isOperator(owner, msg.sender), "ERC7540: unauthorized");
+
+        requestId = _generateRequestId(controller, assets);
+
+        IERC20(asset()).safeTransferFrom(owner, address(this), assets);
+
+        _pendingDepositRequests[controller][requestId].amount += assets;
```
Amxx, line 50: It feels like we should inverse the two lines and remove `nonReentrant`. WDYT?

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,130 @@
...
+        require(owner == msg.sender || isOperator(owner, msg.sender), "ERC7540: unauthorized");
+
+        requestId = _generateRequestId(controller, shares);
+
+        _burn(owner, shares);
+
+        _pendingRedeemRequests[controller][requestId].amount += shares;
+
+        emit RedeemRequest(controller, owner, requestId, msg.sender, shares);
+    }
```
Amxx, line 73: There is no external call in this function, why do we need `nonReentrant` ?

contracts/mocks/token/ERC7540Mock.sol
```diff
@@ -0,0 +1,44 @@
...
+
+pragma solidity ^0.8.20;
+
+import {IERC20} from "../../token/ERC20/IERC20.sol";
+import {ERC20} from "../../token/ERC20/ERC20.sol";
+import {ERC7540} from "../../token/ERC20/extensions/ERC7540.sol";
+import {ERC4626} from "../../token/ERC20/extensions/ERC4626.sol";
+
+contract ERC7540Mock is ERC7540 {
+    constructor(address asset) ERC20("ERC4626Mock", "E4626M") ERC7540(IERC20(asset)) {}
```
Amxx, line 11: for some reason, it seems this is breakin, the transpiler
```suggestion
    constructor(IERC20 asset) ERC20("ERC4626Mock", "E4626M") ERC7540(asset) {}
```

Amxx (commented):
I don't see any "coupling" between the 7540 function and the 4626 logic. I was expecting deposit/reddem logic to hook into 4626. When there is a "delayed option" (7540), should the "instantaneous option" (4626) be disabled ?

Right now, the `requestDeposit`/`requestReedem` operate on assets (transferFrom/_burn) without using any of the 4626 logic. That means:
- the exchange rate defined in 4626 is not considered
- the assets that are transfered/burn change the overall state of the vault, making 4626 operation use a bad rate.
- its possible/easy to miss doing the secound part (_mint/transfer)

Also, It would be interresting to have mock that show features of 7540 that go beyound what 4626 already support. For example, it would be nice to have a system where the exit (reedem) is hapenning like a vesting (automatic based on clock).

contracts/interfaces/IERC7540.sol
DeeJayElly: Done @Amxx

contracts/token/ERC20/extensions/ERC7540.sol
DeeJayElly: Makes sense, i think its an overhead, so i agree with your suggestion to remove it. @Amxx

DeeJayElly: Updated

contracts/token/ERC20/extensions/ERC7540.sol
DeeJayElly: Fixed @Amxx

contracts/token/ERC20/extensions/ERC7540.sol
DeeJayElly: Correct, i missed this part. Good catch @Amxx . Its fixed now.

contracts/token/ERC20/extensions/ERC7540.sol
DeeJayElly: Updated @Amxx

contracts/mocks/token/ERC7540Mock.sol
DeeJayElly: Hmm strange, i didnt encounted this error on mu local environment, i have updated it anyways.

contracts/token/ERC20/extensions/ERC7540.sol
DeeJayElly: Fixed @Amxx

contracts/mocks/token/ERC7540Mock.sol
Amxx: The transpiler is not usually run locally. That is the kind of things that shows up in CI

contracts/interfaces/IERC7540.sol
```diff
@@ -0,0 +1,137 @@
...
+        uint256 requestId
+    ) external view returns (Request memory request);
+
+    /**
+     * @dev Gets the status of an operator for a given controller.
+     * @param controller The address of the controller.
+     * @param operator The address of the operator.
+     * @return status Whether the operator is approved.
+     */
+    function getOperator(address controller, address operator) external view returns (bool status);
```
Amxx, line 136: This is a duplicate of `isOperator`.

Also, its not defined in the ERC, so it has nothing to do here

contracts/token/ERC20/extensions/ERC7540.sol
DeeJayElly: Ok i didnt know that, tnx for clarifying it. Its fixed @Amxx

contracts/interfaces/IERC7540.sol
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * @dev Gets the details of a pending redeem request.
+     * @param controller The address of the controller.
+     * @param requestId The unique identifier for the redeem request.
+     * @return request The request object containing amount and claimable details.
+     */
+    function getPendingRedeemRequest(
+        address controller,
+        uint256 requestId
+    ) external view returns (Request memory request);
```
Amxx, line 128: These two functions are NOT documented in the ERC and are thus not part of the standard. They have nothing to do here.

contracts/interfaces/IERC7540.sol
```diff
@@ -0,0 +1,137 @@
...
+pragma solidity ^0.8.20;
+
+import {IERC4626} from "../interfaces/IERC4626.sol";
+
+/**
+ * @dev Interface for the ERC-7540 Asynchronous Tokenized Vaults standard.
+ * https://eips.ethereum.org/EIPS/eip-7540[ERC-7540]
+ */
+interface IERC7540 is IERC4626 {
+    struct Request {
```
Amxx, line 12: This is not part of the ERC

contracts/interfaces/IERC7540.sol
```diff
@@ -0,0 +1,137 @@
...
+        uint256 indexed requestId,
+        address sender,
+        uint256 shares
+    );
+
+    event OperatorSet(address indexed controller, address indexed operator, bool approved);
+
+    event DepositProcessed(address indexed controller, uint256 indexed requestId, uint256 amount);
+
+    event RedeemProcessed(address indexed controller, uint256 indexed requestId, uint256 amount);
```
Amxx, line 38: These events are not part of the ERC

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,148 @@
...
+     * @dev Creates a new deposit request.
+     */
+    function requestDeposit(
+        uint256 assets,
+        address controller,
+        address owner
+    ) external override returns (uint256 requestId) {
+        if (assets == 0) {
+            return 0;
+        }
```
Amxx, line 39: The ERC says:
> If a Vault returns 0 for the requestId of any request, it MUST return 0 for all requests.

Returning 0 here, and not for the other request is not compliant.

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,148 @@
...
+     * @dev Function to return pending redeem request.
+     */
+    function getPendingRedeemRequest(address controller, uint256 requestId) public view returns (Request memory) {
+        return _pendingRedeemRequests[controller][requestId];
+    }
+
+    /**
+     * @dev Function to return operator.
+     */
+    function getOperator(address controller, address operator) public view returns (bool) {
```
Amxx, line 133: duplicate of `isOperator`

DeeJayElly: Fixed

contracts/interfaces/IERC7540.sol
DeeJayElly: Fixed

contracts/interfaces/IERC7540.sol
DeeJayElly: Removed

contracts/interfaces/IERC7540.sol
DeeJayElly: Removed

contracts/token/ERC20/extensions/ERC7540.sol
DeeJayElly: Fixed

contracts/interfaces/IERC7540.sol
hieronx: You can copy the interfaces that match the standard from here: https://github.com/foundry-rs/forge-std/pull/644/files#diff-39b5d73aa19abd1df539c3f24017dd6db331f8ef698e36a3495aa7cbe1ffc89f

hieronx:
@DeeJayElly @Amxx nice to see work happening on the ERC7540 implementation.

I wonder though why you decided to not go with the approach proposed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5294, and discussed in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/4761#issuecomment-2449468467? 

Specifically, having the flexibility of all 4 options:
A. Fully synchronous (IERC4626)
B. Asynchronous on deposit (IERC7540Deposit)
C. Asynchronous on redeem (IERC7540Redeem)
D. Fully asynchronous (IERC7540Deposit + IERC7540Redeem)

A lot of projects I've been speaking to require partial asynchronicity rather than full asynchronicity.

This is implemented in #5294 already through inheritance.

Also, this current implementation is generating request IDs which is not supported in standard ERC7540. This requires a future extension.

PR #5468, opened 2025-01-28
Add ERC to `ERC4337` interface names

arr00:
Fixes #???? 

ERC4337 declares interfaces without the ERC present in them. To increase clarity, eliminate chance of interface name collision, and maintain consistancy with [ERC7579 interfaces](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/interfaces/draft-IERC7579.sol), this PR adds the ERC to the specced interface name.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/interfaces/draft-IERC4337.sol
+++ b/contracts/interfaces/draft-IERC4337.sol
@@ -54,7 +54,7 @@ struct PackedUserOperation {
  *
  * See https://eips.ethereum.org/EIPS/eip-7766[ERC-7766]
  */
-interface IAggregator {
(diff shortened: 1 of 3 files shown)


ernestognw (approved):
LGTM, but I'd like @Amxx to be aware

This change is technically breaking but I don't think is a problem and keeps consistency. We should note it in the changelog though.

contracts/interfaces/draft-IERC4337.sol
Nichebiche, line 1: your erlckme

PR #5479, opened 2025-02-03
chore: fix incorrect mapping syntax in Solidity

brawncode:
Noticed an issue where the mapping key was unnecessarily given a variable name (`token`), which Solidity doesn't support. 

Fixed it by using the correct syntax:  

```solidity
mapping(address => uint256) private _erc20Released;
```  

**Now the mapping correctly links an `address` to a `uint256` without attempting to name the key.**

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx:
See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5331#issuecomment-2511574552

PR #5480, opened 2025-02-03
fix: Fix incorrect mapping syntax

asholz:
#### Description 
The `mapping` syntax in the contract was incorrect due to the use of named parameters, which Solidity does not support. This caused compilation errors.  

Fixed the following lines:  
- `mapping(address account => address) private _delegatee;` → `mapping(address => address) private _delegatee;`  
- `mapping(address delegatee => Checkpoints.Trace208) private _delegateCheckpoints;` → `mapping(address => Checkpoints.Trace208) private _delegateCheckpoints;`  

🚀

#### PR Checklist

- [x] Tests
- [] Documentation


Amxx:
See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5331#issuecomment-2511574552

PR #5486, opened 2025-02-03
Add clear function to Enumerable{Set,Map}

arr00:
Adds a new `clear` function to the EnumerableSet for each type of enumerable set. This function clears the set in reverse order (to save on gas). It does run in `O(n)` so developers should use it carefully and only when they are sure a set can not grow unbounded. I put a warning in the natspec for this.

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -13,7 +13,7 @@ pragma solidity ^0.8.20;
  *
  * - Elements are added, removed, and checked for existence in constant time
  * (O(1)).
- * - Elements are enumerated in O(n). No guarantees are made on the ordering.
+ * - Elements are enumerated and cleared in O(n). No guarantees are made on the ordering.
  *
  * ```solidity
  * contract Example {
@@ -115,6 +115,18 @@ library EnumerableSet {
     }
 
     /**
+     * @dev Removes all the values from a set. O(n).
+     *
+     * WARNING: Developers should keep in mind that this function has an unbounded cost and using it may render the
+     * function uncallable if the set grows to the point where clearing it consumes too much gas to fit in a block.
+     */
+    function _clear(Set storage set) private {
+        for (uint256 i = _length(set); i > 0; --i) {
+            _remove(set, _at(set, i - 1));
+        }
+    }
+
+    /**
      * @dev Returns true if the value is in the set. O(1).
      */
     function _contains(Set storage set, bytes32 value) private view returns (bool) {
@@ -178,6 +190,16 @@ library EnumerableSet {
      */
     function remove(Bytes32Set storage set, bytes32 value) internal returns (bool) {
         return _remove(set._inner, value);
+    }
+
+    /**
+     * @dev Removes all the values from a set. O(n).
+     *
+     * WARNING: Developers should keep in mind that this function has an unbounded cost and using it may render the
+     * function uncallable if the set grows to the point where clearing it consumes too much gas to fit in a block.
+     */
+    function clear(Bytes32Set storage set) internal {
+        _clear(set._inner);
     }
 
     /**
@@ -254,6 +276,16 @@ library EnumerableSet {
     }
 
     /**
+     * @dev Removes all the values from a set. O(n).
+     *
+     * WARNING: Developers should keep in mind that this function has an unbounded cost and using it may render the
+     * function uncallable if the set grows to the point where clearing it consumes too much gas to fit in a block.
+     */
+    function clear(AddressSet storage set) internal {
+        _clear(set._inner);
+    }
+
+    /**
      * @dev Returns true if the value is in the set. O(1).
      */
     function contains(AddressSet storage set, address value) internal view returns (bool) {
@@ -324,6 +356,16 @@ library EnumerableSet {
      */
     function remove(UintSet storage set, uint256 value) internal returns (bool) {
         return _remove(set._inner, bytes32(value));
+    }
+
+    /**
+     * @dev Removes all the values from a set. O(n).
+     *
+     * WARNING: Developers should keep in mind that this function has an unbounded cost and using it may render the
+     * function uncallable if the set grows to the point where clearing it consumes too much gas to fit in a block.
+     */
+    function clear(UintSet storage set) internal {
+        _clear(set._inner);
     }
 
     /**
--- a/contracts/utils/structs/EnumerableSet.sol
+++ b/contracts/utils/structs/EnumerableSet.sol
@@ -489,6 +489,24 @@ library EnumerableSet {
     }
 
     /**
+     * @dev Removes all the values from a set. O(n).
+     *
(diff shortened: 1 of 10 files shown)


contracts/utils/structs/EnumerableSet.sol
```diff
@@ -114,6 +114,18 @@ library EnumerableSet {
...
+    /**
+     * @dev Removes all the values from a set. O(n).
+     *
+     * WARNING: Developers should keep in mind that this function has an unbounded cost and using it may render the
+     * function uncallable if the set grows to the point where clearing it consumes too much gas to fit in a block.
+     */
+    function _clear(Set storage set) private {
+        for (uint256 i = _length(set); i > 0; --i) {
+            _remove(set, _at(set, i - 1));
+        }
```
Amxx, line 126: This feels very wastefull:
- for each iteration, there is a `lastIndex` lookup that identifies that the last element is in fact being removed (1 unecessary sload)
- for each iteration, there is a pop operation that does a sstore for the length with an intermediary value. 
  - that pop also zeros the values, which is honestly not necessary.
  
IMO, this would be much cheaper:
   ```suggestion
        uint256 length = _length(set);
        for (uint256 i = 0; i < length; ++i) {
            delete set._positions[set._values[index]];
        }
        Arrays.unsafeSetLength(set._values, 0);
```

File: scripts/generate/templates/EnumerableSet.js [hunk 2]
```diff
@@ -117,6 +117,18 @@ function _remove(Set storage set, bytes32 value) private returns (bool) {
...
+/**
+ * @dev Removes all the values from a set. O(n).
+ *
+ * WARNING: Developers should keep in mind that this function has an unbounded cost and using it may render the
+ * function uncallable if the set grows to the point where clearing it consumes too much gas to fit in a block.
+ */
+function _clear(Set storage set) private {
+    for (uint256 i = _length(set); i > 0; --i) {
+        _remove(set, _at(set, i - 1));
+    }
```
Amxx, line 129:
```suggestion
     uint256 length = _length(set);
     for (uint256 i = 0; i < length; ++i) {
         delete set._positions[set._values[index]];
     }
     Arrays.unsafeSetLength(set._values, 0);
```

Amxx: (this needs to import Arrays.sol)

contracts/utils/structs/EnumerableSet.sol
arr00: Is it not worth it to clear the storage for the gas refund?

Amxx: Also, tx with gas refunds are computed as:
- first we charge the gas for the update (+2900 gas here, because the slot is warm) 
- then at the end we apply a refund (-5000 gas)

The `+2900 * length` part accumulates, and may cause the transaction to get "out of gas", because the refund is not applied immediatly.

So when seeking the refund, you also increase the risk of the tx becoming to expensive to be executable.

Amxx: (There is also a risk that with Verkle the refund may go away)

arr00: Something else to note is we already have `n` refunds from the mapping so as is, the tx would need to consume a minimum of of `n*5*4800` gas to make use of the refunds. You are probably right that the refund would often go unused if we deleted the array contents as well (requires `n*48000` gas to be used).

contracts/utils/structs/EnumerableSet.sol
```diff
@@ -13,7 +15,7 @@ pragma solidity ^0.8.20;
  *
  * - Elements are added, removed, and checked for existence in constant time
  * (O(1)).
- * - Elements are enumerated in O(n). No guarantees are made on the ordering.
+ * - Elements are enumerated and cleared in O(n). No guarantees are made on the ordering.
```
Amxx, line 18:
```suggestion
 * - Elements are enumerated in O(n). No guarantees are made on the ordering.
 * - Set can be cleared (all elements removed) in O(n).
```

test/utils/structs/EnumerableMap.behavior.js
```diff
@@ -117,6 +117,49 @@ function shouldBehaveLikeMap() {
     });
   });
 
+  describe('clear', function () {
+    it('clears a single item', async function () {
```
Amxx, line 121:
```suggestion
    it('clears a single entry', async function () {
```

test/utils/structs/EnumerableMap.behavior.js
```diff
@@ -117,6 +117,49 @@ function shouldBehaveLikeMap() {
...
+      await this.methods.set(this.keyA, this.valueA);
+
+      await this.methods.clear();
+
+      expect(await this.methods.contains(this.keyA)).to.be.false;
+      await expectMembersMatch(this.methods, [], []);
+    });
+
+    it('clears multiple items', async function () {
```
Amxx, line 130:
```suggestion
    it('clears multiple entries', async function () {
```

test/utils/structs/EnumerableMap.behavior.js
```diff
@@ -117,6 +117,49 @@ function shouldBehaveLikeMap() {
...
+
+      await this.methods.clear();
+
+      expect(await this.methods.contains(this.keyA)).to.be.false;
+      expect(await this.methods.contains(this.keyB)).to.be.false;
+      expect(await this.methods.contains(this.keyC)).to.be.false;
+      await expectMembersMatch(this.methods, [], []);
+    });
+
+    it('does not revert on empty set', async function () {
```
Amxx, line 143:
```suggestion
    it('does not revert on empty map', async function () {
```

test/utils/structs/EnumerableMap.behavior.js
```diff
@@ -117,6 +117,49 @@ function shouldBehaveLikeMap() {
...
+      await this.methods.clear();
+    });
+
+    it('clear then add value', async function () {
```
Amxx, line 147:
```suggestion
    it('clear then add entry', async function () {
```

Amxx (dismissed):

contracts/utils/structs/EnumerableSet.sol
```diff
@@ -442,6 +488,24 @@ library EnumerableSet {
...
+    function clear(Bytes32x2Set storage self) internal {
+        bytes32[2][] storage v = self._values;
+
+        uint256 len = length(self);
+        for (uint256 i = 0; i < len; ++i) {
+            delete self._positions[_hash(v[i])];
+        }
+        assembly ("memory-safe") {
+            sstore(v.slot, 0)
+        }
```
Amxx, line 506: We don't have a version of `Arrays.unsafeSetLength` for `bytes32[2][]`.

Do we want to add that (quite specific) version to Arrays.sol, of are we ok with this line of assembly here?

arr00: I'm fine with the assembly here.

Amxx (approved):

PR #5493, opened 2025-02-07
Make set-max-old-space-size.sh compatible with sh

0xmichalis:
The . operator defaults to the current shell and
ignores the shebang so in non-bash shells, it is
not bash that will be used, which can result in
the following script failure:
```
> . scripts/set-max-old-space-size.sh && hardhat test

sh: 8: scripts/set-max-old-space-size.sh: [[: not found
```
The shebang could be respected by avoiding sourcing but a better option is to simply remove the bash dependency from the script since it is not required.


Amxx (approved):

Amxx:
> The shebang could be respected by avoiding sourcing 

If we do that, would the `export` declaration work ?

arr00 (approved):
This works better thank you. Removing the sourcing would possibly result in the `export` not being reflected in the parent process so this is the ideal solution.

PR #5508, opened 2025-02-14
ref: added functionality to escapeJSONstrings (ref: #5251)

### LINKED ISSUE #5251: Add support for escaping JSON to Strings lib
**🧐 Motivation**
Some existing solidity libraries have functions that allow escaping JSON strings--this is crucial for NFT metadata stored on-chain with user input to ensure there can't be injection attacks. We should consider adding a similar feature to the strings library.

DarkLord017:
Fixes #5251  

A function to escape special characters in JSON strings
Handles key characters like quotes ("), backslashes (), forward slashes (/), and control characters (\n, \t, \r, etc.)
Prevents JSON injection attacks in NFT metadata and other use cases

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/Strings.sol
```diff
@@ -438,4 +447,42 @@ library Strings {
...
+            else if (char == 0x0C) output[outputLength++] = 'f';
+            else if (char == 0x0D) output[outputLength++] = 'r';
+            else if (char == 0x22) output[outputLength++] = '"';
+            else if (char == 0x2F) output[outputLength++] = '/';
+            else if (char == 0x5C) output[outputLength++] = '\\';
+        } else {
+            output[outputLength++] = char;
+        }
+
+        unchecked { ++i; }
```
Amxx, line 474: Please lets all stop doing that. The compiler knows how to handle this correctly.

contracts/utils/Strings.sol
```diff
@@ -438,4 +447,42 @@ library Strings {
             value := mload(add(buffer, add(0x20, offset)))
         }
     }
+    
+    function _escapeJsonString(string memory input) private pure returns (string memory) {
+    bytes memory buffer = bytes(input);
+    bytes memory output = new bytes(buffer.length * 2); // Allocate max possible space
```
Amxx, line 453: I'm worried this allocated a lot of memory that is not freed. This is quite expensive. Alternatives are:
- count number of "extra chars" in a loop, and allocate the right size directly
- resize the output buffer without a copy, and move the free memory ptr, using assembly

contracts/utils/Strings.sol
```diff
@@ -15,6 +15,15 @@ library Strings {
 
     bytes16 private constant HEX_DIGITS = "0123456789abcdef";
     uint8 private constant ADDRESS_LENGTH = 20;
+    uint256 constant SPECIAL_CHARS_LOOKUP =
```
Amxx, line 18:
```suggestion
    uint256 private constant SPECIAL_CHARS_LOOKUP = 
```

contracts/utils/Strings.sol
Amxx: On the other hand, its worth putting the entier function body in an unchecked block so that all the `outputLength++` are unchecked.

contracts/utils/Strings.sol
```diff
@@ -15,6 +15,15 @@ library Strings {
...
     uint8 private constant ADDRESS_LENGTH = 20;
+    uint256 constant SPECIAL_CHARS_LOOKUP = 
+    (1 << 8) |  // backspace
+    (1 << 9) |  // tab
+    (1 << 10) | // newline
+    (1 << 12) | // form feed
+    (1 << 13) | // carriage return
+    (1 << 34) | // double quote
+    (1 << 47) | // forward slash
+    (1 << 92);  // backslash
```
Amxx, line 26: IMO it would be cleaner to use hex representation for these, but that is really personal.
```suggestion
    (1 << 0x08) |  // backspace
    (1 << 0x09) |  // tab
    (1 << 0x0a) | // newline
    (1 << 0x0c) | // form feed
    (1 << 0x0d) | // carriage return
    (1 << 0x22) | // double quote
    (1 << 0x2f) | // forward slash
    (1 << 0x5c);  // backslash
```

DarkLord017:
Screenshot 2025-02-16 at 11 37 54 AM
I tried implementing both changes, but the one I pushed used 207 less gas when tested with 'Test "Hello , World!"'. I've uploaded a screenshot of the other implementation. Which one do you think
 I should go with?

DarkLord017:
U can review it now @Amxx

Amxx:
You are doing quite a lot of read/write to the `output` and `0x40`. That can be significantly reduced!
Also, I don't understand why you do `outputLength + 2`

My version is just 
```solidity

**Comment by Amxx** [comment]
``

**Comment by Amxx** [comment]
TODO: 
- testing: using `JSON.stringify` and removing the first and last caracteres (`"`) almost works. It does NOT escape `\`
- linting: prettier wants use to use `'"'`, but solhint wants `"\""`.

**Comment by DarkLord017** [comment]
okay will do 

> TODO:
> 
> * testing: using `JSON.stringify` and removing the first and last caracteres (`"`) almost works. It does NOT escape `\`
> * linting: prettier wants use to use `'"'`, but solhint wants `"\""`.

File: contracts/utils/Strings.sol [hunk 5]
```diff
@@ -426,6 +435,40 @@ library Strings {
         return value;
     }
 
+    /**
+     * @dev Escape special characters in JSON strings. This can be usefull to prevent JSON injection in NFT metadata.
```
Amxx, line 439:
```suggestion
     * @dev Escape special characters in JSON strings. This can be useful to prevent JSON injection in NFT metadata.
```

DarkLord017:
U can review it now @Amxx

contracts/utils/Strings.sol
```diff
@@ -446,20 +445,24 @@ library Strings {
...
+                    output[outputLength++] = bytes1(uint8(0x74)); // t
+                else if (char == 0x0A)
+                    output[outputLength++] = bytes1(uint8(0x6E)); // n
+                else if (char == 0x0C)
+                    output[outputLength++] = bytes1(uint8(0x66)); // f
+                else if (char == 0x0D)
+                    output[outputLength++] = bytes1(uint8(0x72)); // r
+                else if (char == 0x22)
+                    output[outputLength++] = bytes1(uint8(0x22)); // "
+                else if (char == 0x5C) output[outputLength++] = bytes1(uint8(0x5C)); // \
```
Amxx, line 461: that change really doesn't go in the direction of readability. IMO we should keep the old code and fix the linter instead of doing this kind of things

test/utils/Strings.test.js
```diff
@@ -339,4 +339,11 @@ describe('Strings', function () {
       }
     });
   });
+
+  describe('Escape JSON string', function () {
+    for (const input of ['', 'a', '{"a":"b/c"}', 'a\tb\nc\\d"e\rf/g\fh\bi'])
+      it.only(`escape ${JSON.stringify(input)}`, async function () {
```
Amxx, line 345:
```suggestion
      it(`escape ${JSON.stringify(input)}`, async function () {
```

DarkLord017:
> Is it correct to drop the forward slash escaping?

I read this then removed that 
https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped

Amxx (approved):

arr00:
We should have a comment that this can only be used inside double quote strings. Single quotes are not escaped.

arr00 (approved):

contracts/utils/Strings.sol
```diff
@@ -426,6 +434,43 @@ library Strings {
...
+     *
+     * WARNING: This function should only be used in double quoted JSON strings. Single quotes are not escaped.
+     */
+    function escapeJSON(string memory input) internal pure returns (string memory) {
+        bytes memory buffer = bytes(input);
+        bytes memory output = new bytes(2 * buffer.length); // worst case scenario
+        uint256 outputLength = 0;
+
+        for (uint256 i; i < buffer.length; ++i) {
+            bytes1 char = buffer[i];
```
ernestognw, line 448: We may use `unsafeMemoryAccess` from the Arrays.sol library here

contracts/utils/Strings.sol
```diff
@@ -426,6 +434,43 @@ library Strings {
...
+            if (((SPECIAL_CHARS_LOOKUP & (1 << uint8(char))) != 0)) {
+                output[outputLength++] = "\\";
+                if (char == 0x08) output[outputLength++] = "b";
+                else if (char == 0x09) output[outputLength++] = "t";
+                else if (char == 0x0A) output[outputLength++] = "n";
+                else if (char == 0x0C) output[outputLength++] = "f";
+                else if (char == 0x0D) output[outputLength++] = "r";
+                else if (char == 0x5C) output[outputLength++] = "\\";
```
ernestognw, line 456: Most likely the audit will suggest consistency with the `SPECIAL_CHARS_LOOKUP`

```suggestion
                else if (char == 0x09) output[outputLength++] = "t";
                else if (char == 0x0a) output[outputLength++] = "n";
                else if (char == 0x0c) output[outputLength++] = "f";
                else if (char == 0x0d) output[outputLength++] = "r";
                else if (char == 0x5c) output[outputLength++] = "\\";
```

contracts/utils/Strings.sol
```diff
@@ -426,6 +434,43 @@ library Strings {
...
+                    output[outputLength++] = '"';
+                }
+            } else {
+                output[outputLength++] = char;
+            }
+        }
+        // write the actual length and deallocate unused memory
+        assembly ("memory-safe") {
+            mstore(output, outputLength)
+            mstore(0x40, add(output, shl(5, shr(5, add(outputLength, 63)))))
```
ernestognw, line 468: Same here (audit will report consistency)

```suggestion
            mstore(64, add(output, shl(5, shr(5, add(outputLength, 63)))))
```

contracts/utils/Strings.sol
```diff
@@ -426,6 +434,43 @@ library Strings {
...
     }
 
+    /**
+     * @dev Escape special characters in JSON strings. This can be useful to prevent JSON injection in NFT metadata.
+     *
+     * WARNING: This function should only be used in double quoted JSON strings. Single quotes are not escaped.
+     */
+    function escapeJSON(string memory input) internal pure returns (string memory) {
+        bytes memory buffer = bytes(input);
+        bytes memory output = new bytes(2 * buffer.length); // worst case scenario
```
ernestognw, line 444: I don't like that this is allocating worst-case scenario memory. Ideally, it should be just `buffer.length`, iterate and count the characters to escape, then indeed escape them. The tradeoff here is perhaps a bit more computation (i.e. more checks over) so I'm not convinced it's an actual saving.

ernestognw (dismissed):
Technically correct, left some minor suggestions. I would prefer merging rather than addressing those.

contracts/utils/Strings.sol
Amxx: We tested to have two loops:

- one loop that counts the times escaping is needed
- then we allocate exactly the right size
- then we loop a second time.

It ends up being more expensive.

---

An alternative is to do all the allocation manually (basically we move the fmp ourselves)

Amxx (approved):

PR #5510, opened 2025-02-14
chore: fix incorrect 

GooseMatrix:
function with make - function will make

represent - represents

other roles's - other role's


Amxx:
@GooseMatrix you should run `yarn generate` after changing the templates. Otherwise there is an inconsistency.

Amxx (approved):

PR #5519, opened 2025-02-21
Add Arachnid's and Micah's deployer to the common contracts

Amxx:
This is an evolution of the test environment. Shouldn't need a changelog entry.

This is an interresting factory to use with the ERC-4337 entrypoint (in the community repo)

Note: this contract is provided by default in Foundry.

### DIFF
--- a/hardhat/common-contracts.js
+++ b/hardhat/common-contracts.js
@@ -6,6 +6,7 @@ const fs = require('fs');
 const path = require('path');
 
 const INSTANCES = {
+  // Entrypoint v0.7.0
   entrypoint: {
     address: '0x0000000071727De22E5E9d8BAf0edAc6f37da032',
     abi: JSON.parse(fs.readFileSync(path.resolve(__dirname, '../test/bin/EntryPoint070.abi'), 'utf-8')),
@@ -15,6 +16,14 @@ const INSTANCES = {
     address: '0xEFC2c1444eBCC4Db75e7613d20C6a62fF67A167C',
     abi: JSON.parse(fs.readFileSync(path.resolve(__dirname, '../test/bin/SenderCreator070.abi'), 'utf-8')),
     bytecode: fs.readFileSync(path.resolve(__dirname, '../test/bin/SenderCreator070.bytecode'), 'hex'),
+  },
+  // Arachnid's deterministic deployment proxy
+  // See: https://github.com/Arachnid/deterministic-deployment-proxy/tree/master
+  arachnidDeployer: {
+    address: '0x4e59b44847b379578588920ca78fbf26c0b4956c',
+    abi: [],
+    bytecode:
+      '0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3',
   },
 };
 
--- a/hardhat/common-contracts.js
+++ b/hardhat/common-contracts.js
@@ -20,10 +20,16 @@ const INSTANCES = {
   // Arachnid's deterministic deployment proxy
   // See: https://github.com/Arachnid/deterministic-deployment-proxy/tree/master
   arachnidDeployer: {
-    address: '0x4e59b44847b379578588920ca78fbf26c0b4956c',
+    address: '0x4e59b44847b379578588920cA78FbF26c0B4956C',
     abi: [],
     bytecode:
       '0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3',
+  },
+  // Mocah's deployer
+  mocahDeployer: {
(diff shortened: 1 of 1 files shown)


arr00:
Should we add another one too so we aren't pushing a specific deployer? I generally use [this one](https://etherscan.io/address/0x0000000000ffe8b47b3e2130213b802212439497) by 0age which is deployed on a lot of chains (deployed by micha's deployer).

Amxx:
I added this one because it is, afaik, the oldest one, on the the most widely available, and the one used to deploy the 4337 entry point! Its bytecode is also super small.

I'd be open to adding others, but it don't want that to be a rush to adding more and more. You point one that is itself deployed by Micha's deployer. So should we also include Micha's deployer ?

arr00:
> I added arachnid's because it is, afaik, the oldest one, possibly the most widely available, and the one used to deploy the 4337 entry point! Its bytecode is also super small.
> 
> I'd be open to adding others, but it don't want that to be a rush to adding more and more. You point one that is itself deployed by Micha's deployer. So should we also include Micha's deployer ?
> 
> Edit: after looking at the one you shared, it really looks bulky and opinionated compared to Arachnid's simplicity

You, would probably make more sense to simply use micha's deployer which this is derived from. Micha's is older than arachnid's!

Amxx:
Is Micha's deployer this one ?

https://etherscan.io/address/0xcfa3a7637547094ff06246817a35b8333c315196#code

Bytecode is soooo long

arr00 (dismissed):

Amxx:
Fun fact about "the first factory"
- Arachnid's deterministic deployer was deployed on August 16 2020(https://etherscan.io/tx/0xeddf9e61fb9d8f5111840daef55e5fde0041f5702856532cdbb5a02998033d26)
- [this one](https://etherscan.io/address/0xfac100450af66d838250ea25a389d8cd09062629#code) was deployed to mainnet on July 1 2019 (over a year before) ... [there was a talk in devcon about it](https://www.youtube.com/watch?v=8w4zRtkfKT0)
- Mocah's one was deployed on May 29 2019.

I did not know about Mocah's when I deployed mine.

hardhat/common-contracts.js
```diff
@@ -20,11 +20,17 @@ const INSTANCES = {
...
   // See: https://github.com/Arachnid/deterministic-deployment-proxy/tree/master
   arachnidDeployer: {
-    address: '0x4e59b44847b379578588920ca78fbf26c0b4956c',
+    address: '0x4e59b44847b379578588920cA78FbF26c0B4956C',
     abi: [],
     bytecode:
       '0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3',
   },
+  // Mocah's deployer
+  mocahDeployer: {
```
arr00, line 29: [Micah](https://x.com/MicahZoltu)

arr00:
```suggestion
  // Micah's deployer
  micahDeployer: {
```

hardhat/common-contracts.js
```diff
@@ -16,6 +17,20 @@ const INSTANCES = {
...
+  // Arachnid's deterministic deployment proxy
+  // See: https://github.com/Arachnid/deterministic-deployment-proxy/tree/master
+  arachnidDeployer: {
+    address: '0x4e59b44847b379578588920cA78FbF26c0B4956C',
+    abi: [],
+    bytecode:
+      '0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3',
+  },
+  // Mocah's deployer
+  mocahDeployer: {
```
Amxx, line 29:
```suggestion
  // Micah's deployer
  micahDeployer: {
```

arr00 (approved):

pcaversaccio:
FWIW, I built [`CreateX`](https://github.com/pcaversaccio/createx):
- https://createx.rocks/deployments -> it's currently deployed on 133 chains
- Is a predeploy on all OP Stack chains: https://specs.optimism.io/protocol/preinstalls.html
- Next entry point deployment of ERC-4337 will most probably use this factory
- Has way more features than any other factory out there

Amxx:
CreateX is really big ! We'll have to consider the cost/benefit of having it by default.

I'm starting to think this script should be an independent hardhat module which can be configured to enable/disable what is needed

pcaversaccio:
> CreateX is really big ! We'll have to consider the cost/benefit of having it by default.
> 
> I'm starting to think this script should be an independent hardhat module which can be configured to enable/disable what is needed

Sure - I just wanted to highlight it as it's now the common factory used across dev frameworks. E.g. Hardhat Ignition uses it https://hardhat.org/ignition/docs/guides/create2 as well as Foundry will switch to it as well: https://github.com/foundry-rs/foundry/issues/2638

PR #5466, opened 2025-01-28
Group typo fixes

ernestognw:
The #typo-fixes branches accumulated typo fixes. This branch should be merged before freezing a release candidate.
This is done to avoid multiple small commits on that polute the master branch. All typo fix PR should target this branch.

PR #5528, opened 2025-02-25
fix errors in function ERC20Wrapper.test.js

futreall:
inssuficient - insufficient x2


arr00:
Addressed by PR #5466

PR #5529, opened 2025-02-26
chore: fix quotation marks

taozui472:
The unexpected marks might confuse readers and be distracted by them. Also, some editors might not render correctly depending on the software or platform used, appearing as gibberish or causing formatting problems.


Amxx:
@arr00 @ernestognw WDYT?

Personally, I wouldn't change that.

arr00:
Agree, I prefer smart quotation marks in documentation.

PR #5534, opened 2025-02-27
Add Array.uniquifySorted to the Arrays library

### LINKED ISSUE #5520: Add Array.uniquifySorted to the Arrays library
**🧐 Motivation**
With the recent addition of sorting via quickSort to the library, results very convenient for consumers to also have an `Array.uniquifySorted` available for uniquifying sorted arrays. 

**📝 Details**
In some cases, when we cannot utilize mappings, we utilize the Array library to eliminate duplicates on short-sized arrays, which could be easily done by executing Array.sort() + Array.uniquifySorted().

sivaratrisrinivas:
This adds a uniquifySorted function to remove duplicates from sorted arrays. The function has O(n) time complexity and works with uint256, address, and bytes32 array types. This is particularly useful when mappings cannot be utilized, and provides an easy way to combine Array.sort() + Array.uniquifySorted().

Fixes #5520

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Arrays.sol
+++ b/contracts/utils/Arrays.sol
@@ -479,4 +479,69 @@ library Arrays {
             sstore(array.slot, len)
         }
     }
+
+    /**
+     * @dev Removes duplicate values from a sorted array. This function does not check that the array is sorted,
+     * behavior is undefined if the array is not sorted. The resulting array will have no duplicates and will
+     * be shorter or the same length as the input array.
+     *
+     * This operation is performed in-place by moving elements and modifying the length of the array.
+     * Time complexity O(n).
+     *
+     * WARNING: This function is destructive. It will modify the array passed by reference.
+     */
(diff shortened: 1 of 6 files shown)


contracts/utils/Arrays.sol
```diff
@@ -479,4 +479,69 @@ library Arrays {
...
+                ++resultSize;
+            }
+        }
+
+        // Resize the array by creating a new one (can't modify length of memory arrays directly)
+        uint256[] memory result = new uint256[](resultSize);
+        for (uint256 i = 0; i < resultSize; ++i) {
+            result[i] = array[i];
+        }
+        return result;
```
Amxx, line 513: If it is agreed that modifying the array passed by reference is the right approach, then we should not do that copy. Instead, we should just resize (down) the array. `unsafeSetLength` does that.

PR #5552, opened 2025-03-04
fix typo Checkpoints.t.sol

sky-coderay:
test/utils/structs/Checkpoints.t.sol
function with make - function will make

with - will


Amxx:
Already covered by #5466

PR #5574, opened 2025-03-11
Use `NoncesKeyed` for `Governor` votes by sig with `proposalId` as the key.

### LINKED ISSUE #5565: Governor Should Use `NoncesKeyed` for Voting by Sig
**🧐 Motivation**

Using a nonce in addition to proposal id in governor makes it hard to have a functional vote by sig service. This is because votes by sig must have a sequential nonce with no gaps or repeats--if a user decides to take one of their votes by sig on-chain, all other pending votes they have are invalidated. This isn't the desired flow and it isn't necessary.

**📝 Details**

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ca7a4e39de0860bbaadf95824207886e6de9fa64/contracts/governance/Governor.sol#L583
This line asserts that votes by sig have sequential nonces which isn't actually necessary. The OZ governor introduces voting by nonce as it is possible that a user would vote multiple times for a single proposal id (so the proposal id can't be the only source of entropy).

It is, however, useful to ensure that a sequence binds together votes on a specific proposal. We should use NoncesKeyed with the key being the proposal id and allow for the 0 key to continue working (no breaking changes).

arr00:
Fixes #5565

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -12,7 +12,7 @@ import {SafeCast} from "../utils/math/SafeCast.sol";
 import {DoubleEndedQueue} from "../utils/structs/DoubleEndedQueue.sol";
 import {Address} from "../utils/Address.sol";
 import {Context} from "../utils/Context.sol";
-import {Nonces} from "../utils/Nonces.sol";
+import {NoncesKeyed} from "../utils/NoncesKeyed.sol";
 import {Strings} from "../utils/Strings.sol";
 import {IGovernor, IERC6372} from "./IGovernor.sol";
 
@@ -25,7 +25,7 @@ import {IGovernor, IERC6372} from "./IGovernor.sol";
  * - A voting module must implement {_getVotes}
  * - Additionally, {votingPeriod}, {votingDelay}, and {quorum} must also be implemented
  */
-abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC721Receiver, IERC1155Receiver {
+abstract contract Governor is Context, ERC165, EIP712, NoncesKeyed, IGovernor, IERC721Receiver, IERC1155Receiver {
     using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
 
     bytes32 public constant BALLOT_TYPEHASH =
@@ -578,15 +578,8 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         address voter,
         bytes memory signature
     ) public virtual returns (uint256) {
-        bool valid = SignatureChecker.isValidSignatureNow(
-            voter,
-            _hashTypedDataV4(keccak256(abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, _useNonce(voter)))),
-            signature
-        );
-
-        if (!valid) {
-            revert GovernorInvalidSignature(voter);
-        }
+        bytes memory rawSignatureDigestData = abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, nonces(voter));
+        _validateVoteSignature(voter, proposalId, signature, rawSignatureDigestData);
 
         return _castVote(proposalId, voter, support, "");
     }
@@ -602,29 +595,54 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
         bytes memory params,
         bytes memory signature
     ) public virtual returns (uint256) {
-        bool valid = SignatureChecker.isValidSignatureNow(
+        bytes memory rawSignatureDigestData = abi.encode(
+            EXTENDED_BALLOT_TYPEHASH,
+            proposalId,
+            support,
             voter,
-            _hashTypedDataV4(
-                keccak256(
-                    abi.encode(
-                        EXTENDED_BALLOT_TYPEHASH,
-                        proposalId,
-                        support,
-                        voter,
-                        _useNonce(voter),
-                        keccak256(bytes(reason)),
-                        keccak256(params)
-                    )
-                )
-            ),
-            signature
+            nonces(voter),
+            keccak256(bytes(reason)),
+            keccak256(params)
         );
+        _validateVoteSignature(voter, proposalId, signature, rawSignatureDigestData);
+
+        return _castVote(proposalId, voter, support, reason, params);
+    }
+
+    function _validateVoteSignature(
+        address voter,
+        uint256 proposalId,
+        bytes memory signature,
+        bytes memory rawSignatureDigestData
+    ) internal virtual {
+        bool valid;
+        if (
+            SignatureChecker.isValidSignatureNow(voter, _hashTypedDataV4(keccak256(rawSignatureDigestData)), signature)
+        ) {
+            valid = true;
+            _useNonce(voter);
+        } else {
+            // uint192 is sufficient entropy for proposalId within nonce keys.
+            uint256 keyedNonce = nonces(voter, uint192(proposalId));
+            assembly ("memory-safe") {
+                mstore(add(rawSignatureDigestData, 0xA0), keyedNonce)
+            }
+
+            if (
+                SignatureChecker.isValidSignatureNow(
+                    voter,
+                    _hashTypedDataV4(keccak256(rawSignatureDigestData)),
+                    signature
+                )
+            ) {
+                valid = true;
+                _useNonce(voter, uint192(proposalId));
+            }
+        }
 
         if (!valid) {
             revert GovernorInvalidSignature(voter);
         }
-
-        return _castVote(proposalId, voter, support, reason, params);
     }
 
     /**
--- a/test/governance/Governor.test.js
+++ b/test/governance/Governor.test.js
@@ -3,7 +3,7 @@ const { expect } = require('chai');
 const { loadFixture } = require('@nomicfoundation/hardhat-network-helpers');
 
 const { GovernorHelper } = require('../helpers/governance');
-const { getDomain, Ballot } = require('../helpers/eip712');
+const { getDomain, Ballot, ExtendedBallot } = require('../helpers/eip712');
 const { ProposalState, VoteType } = require('../helpers/enums');
 const time = require('../helpers/time');
 
@@ -27,6 +27,8 @@ const value = ethers.parseEther('1');
 
 const signBallot = account => (contract, message) =>
   getDomain(contract).then(domain => account.signTypedData(domain, { Ballot }, message));
+const signExtendedBallot = account => (contract, message) =>
+  getDomain(contract).then(domain => account.signTypedData(domain, { ExtendedBallot }, message));
 
 async function deployToken(contractName) {
   try {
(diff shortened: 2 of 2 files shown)


tushar994:
lgtm

contracts/governance/Governor.sol
```diff
@@ -602,29 +595,54 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
+            SignatureChecker.isValidSignatureNow(voter, _hashTypedDataV4(keccak256(rawSignatureDigestData)), signature)
+        ) {
+            valid = true;
+            _useNonce(voter);
+        } else {
+            // uint192 is sufficient entropy for proposalId within nonce keys.
+            uint256 keyedNonce = nonces(voter, uint192(proposalId));
+            assembly ("memory-safe") {
+                mstore(add(rawSignatureDigestData, 0xA0), keyedNonce)
+            }
```
Amxx, line 629: I'm really not a fan of the assumption made here.

This is an internal function. It could be overriden, of called with different parameters.

arr00: Would you be ok with making it private?

contracts/governance/Governor.sol
```diff
@@ -602,29 +595,54 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
+        return _castVote(proposalId, voter, support, reason, params);
+    }
+
+    function _validateVoteSignature(
+        address voter,
+        uint256 proposalId,
+        bytes memory signature,
+        bytes memory rawSignatureDigestData
+    ) internal virtual {
+        bool valid;
```
Amxx, line 618: valid is not necessary. you can do

```
if (SignatureChecker.isValidSignatureNow(...)) {
    _useNonce
    return;
}

// rebuild data

if (SignatureChecker.isValidSignatureNow(...)) {
    _useNonce
    return
}

revert
```

Amxx: or 
```
if (SignatureChecker.isValidSignatureNow(...)) {
    _useNonce
} else {
    // rebuild data
    if (SignatureChecker.isValidSignatureNow(...)) {
        _useNonce
    } else {
        revert
    }
}
```

contracts/governance/Governor.sol
```diff
@@ -578,15 +578,8 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
...
-            voter,
-            _hashTypedDataV4(keccak256(abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, _useNonce(voter)))),
-            signature
-        );
-
-        if (!valid) {
-            revert GovernorInvalidSignature(voter);
-        }
+        bytes memory rawSignatureDigestData = abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, nonces(voter));
+        _validateVoteSignature(voter, proposalId, signature, rawSignatureDigestData);
```
Amxx, line 582: We could do something like

```suggestion
        _validateVoteSignature(
            voter, 
            proposalId, 
            signature, 
            abi.encode(BALLOT_TYPEHASH, proposalId, support, voter), // encodedBeforeNonce
            "", // encodedAfterNonce
        );
```

And let `_validateVoteSignature` fill the nonce using `abi.encodePacked(before, nonce(...), after)`

contracts/governance/Governor.sol
arr00: Good point. Updated

contracts/governance/Governor.sol
Amxx: From our discussion: 
- use an offset value to replace the hardcoded 0xA0

test/governance/Governor.test.js
```diff
@@ -198,31 +198,35 @@ describe('Governor', function () {
       });
 
       describe('vote with signature', function () {
```
0xy0k, line 200: what about adding stress test edge cases (e.g. signature replay on multiple proposals) or extend the extension to support batch signatures per user?

ernestognw (commented):
I like the idea of abstracting away the signature validation. Essentially, we're trying to add a feature to how the signature digest is generated (i.e. the keyed nonce). 

Currently, the contract has 2 different typed signatures: `BALLOT_TYPEHASH` and `EXTENDED_BALLOT_TYPEHASH`, and given that it's not possible to extend EIP-712 types, then it's impossible to build a unified `castVoteBySig` function, therefore we have `castVoteWithReasonAndParamsBySig`.

Similar to how there are two functions for casting a signature vote, I feel the best abstraction is to abstract away internal `_validateVoteSig` and `_validateExtendedVoteSig`. Essentially, we would have the following in `Governor.sol`:

```solidity
abstract contract Governor is ... {
    function castVoteBySig(
        uint256 proposalId,
        uint8 support,
        address voter,
        bytes memory signature
    ) public virtual returns (uint256) {
        bool valid = _validateVoteSig(proposalId, support, voter, signature);
        if (!valid) {
            revert GovernorInvalidSignature(voter);
        }
        return _castVote(proposalId, voter, support, "");
    }

    function castVoteWithReasonAndParamsBySig(
        uint256 proposalId,
        uint8 support,
        address voter,
        string calldata reason,
        bytes memory params,
        bytes memory signature
    ) public virtual returns (uint256) {
        bool valid = _validateExtendedVoteSig(proposalId, support, voter, reason, params, signature);
        if (!valid) {
            revert GovernorInvalidSignature(voter);
        }
        return _castVote(proposalId, voter, support, reason, params);
    }

    function _validateVoteSig(
        uint256 proposalId,
        uint8 support,
        address voter,
        bytes memory signature
    ) internal virtual returns (bool) {
        return
            SignatureChecker.isValidSignatureNow(
                voter,
                keccak256(abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, _useNonce(voter))),
                signature
            );
    }

    function _validateExtendedVoteSig(
        uint256 proposalId,
        uint8 support,
        address voter,
        string memory reason,
        bytes memory params,
        bytes memory signature
    ) internal virtual returns (bool) {
        return
            SignatureChecker.isValidSignatureNow(
                voter,
                keccak256(
                    abi.encode(
                        EXTENDED_BALLOT_TYPEHASH,
                        proposalId,
                        support,
                        voter,
                        _useNonce(voter),
                        keccak256(bytes(reason)),
                        keccak256(params)
                    )
                ),
                signature
            );
    }
}
```

Then, building a `GovernorNoncesKeyed` becomes just an override to both `_validateVoteSig` and `_validateExtendedVoteSig`:

```solidity
abstract contract GovernorNoncesKeyed is Governor, NoncesKeyed {
    function _useCheckedNonce(address owner, uint256 nonce) internal virtual override(Nonces, NoncesKeyed) {
        super._useCheckedNonce(owner, nonce);
    }

    function _validateVoteSignature(
        uint256 proposalId,
        uint8 support,
        address voter,
        bytes memory signature
    ) internal virtual override returns (bool) {
        bool valid = SignatureChecker.isValidSignatureNow(
            voter,
            keccak256(abi.encode(BALLOT_TYPEHASH, proposalId, support, voter, nonces(voter, uint192(proposalId)))),
            signature
        );
        if (valid) _useNonce(voter, uint192(proposalId));
        else valid = super._validateVoteSignature(proposalId, support, voter, signature);
        return valid;
    }

    function _validateExtendedVoteSignature(
        uint256 proposalId,
        uint8 support,
        address voter,
        string memory reason,
        bytes memory params,
        bytes memory signature
    ) internal virtual override returns (bool) {
        bool valid = SignatureChecker.isValidSignatureNow(
            voter,
            keccak256(
                abi.encode(
                    EXTENDED_BALLOT_TYPEHASH,
                    proposalId,
                    support,
                    voter,
                    nonces(voter, uint192(proposalId)),
                    keccak256(bytes(reason)),
                    keccak256(params)
                )
            ),
            signature
        );
        if (valid) _useNonce(voter, uint192(proposalId));
        else valid = super._validateExtendedVoteSignature(proposalId, support, voter, reason, params, signature);
        return valid;
    }
}
```

I acknowledge that the internal functions would be missing calling `super` if the first signature is valid. However, I think the pattern of internal functions that return booleans is the only pattern where we allow bypassing a call to super as long as it's noted. In this case, I think it's acceptable as well.

contracts/governance/extensions/GovernorNoncesKeyed.sol
```diff
@@ -0,0 +1,46 @@
...
+pragma solidity ^0.8.20;
+
+import {Governor} from "../Governor.sol";
+import {Nonces} from "../../utils/Nonces.sol";
+import {NoncesKeyed} from "../../utils/NoncesKeyed.sol";
+import {SignatureChecker} from "../../utils/cryptography/SignatureChecker.sol";
+
+/**
+ * @dev An extension of {Governor} that extends existing nonce management to use {NoncesKeyed}, where the key is the first 192 bits of the `proposalId`.
+ * This is useful for voting by signature while maintaining separate sequences of nonces for each proposal.
```
ernestognw, line 12: Based on this definition, then it's possible (not sure if feasible) to "mine" a `key` by just adding a salt to the description of a proposal. I wonder if we should document further that a nonce can be invalidated if the user happens to sign a vote for a proposal with colliding first 192 bits.

arr00: This is 4 bytes larger than an ethereum address--it would be significantly harder to mine a proposal nonce than choosing an arbitrary address (assuming fully random proposal id generation). I think we are fine here.

ernestognw: Yeah lol, I underestimated how big this is. Solving

arr00:
@ernestognw

ernestognw:
Just benchmarked this implementation with an alternative proposed in https://github.com/arr00/openzeppelin-contracts/pull/1.

The current implementation:

```
╭--------------------+------------------+-------------------+--------------------+---------------------╮
| Contract           | Runtime Size (B) | Initcode Size (B) | Runtime Margin (B) | Initcode Margin (B) |
+======================================================================================================+
| MyGovernor         | 16,979           | 19,235            | 7,597              | 29,917              |
╰--------------------+------------------+-------------------+--------------------+---------------------╯
```

Proposal

```
╭--------------------+------------------+-------------------+--------------------+---------------------╮
| Contract           | Runtime Size (B) | Initcode Size (B) | Runtime Margin (B) | Initcode Margin (B) |
+======================================================================================================+
| MyGovernor         | 17,222           | 19,478            | 7,354              | 29,674              |
╰--------------------+------------------+-------------------+--------------------+---------------------╯
```

Naturally, the proposal generates more code to handle both different functions. The impact is a 3% decrease in the runtime margin (1-(7354/7597)) so developers can only around 7,354 bytes on top.  And there’s still some run for the constructor.

In my opinion this is good enough. Not sure what others would say

ernestognw (approved):
LGTM. Thanks for your patience @arr00 !

ernestognw (approved):

Amxx (approved):

PR #5584, opened 2025-03-17
Address audit findings (5.3 diff audit)

Amxx:
#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/Strings.sol
```diff
@@ -438,6 +438,10 @@ library Strings {
      * @dev Escape special characters in JSON strings. This can be useful to prevent JSON injection in NFT metadata.
      *
      * WARNING: This function should only be used in double quoted JSON strings. Single quotes are not escaped.
+     *
+     * NOTE: This function escapes all unicode characters, and not just the ones in ranges defined in section 2.5 of
+     * RFC-4627 (U+0000 to U+001F, U+0022 and U+005C). Javascript's `JSON.parse` does recover escaped unicode
```
ernestognw, line 443: I took a look to different Javascript implementations and I think it's safe to reference ECMAScript's. Every Javascript implementation should comply with it.

```suggestion
     * RFC-4627 (U+0000 to U+001F, U+0022 and U+005C). ECMAScript's `JSON.parse` does recover escaped unicode
```

Characters escaped with `JSON.stringify` are defined [here](https://262.ecma-international.org/14.0/index.html#table-json-single-character-escapes), but seems that `JSON.parse` is consistent with RFC-4627 2.5 section:

>  The representation of strings is similar to conventions used in the C
   family of programming languages.  A string begins and ends with
   quotation marks.  All Unicode characters may be placed within the
   quotation marks except for the characters that must be escaped:
   quotation mark, reverse solidus, and the control characters (U+0000
   through U+001F).
>  Any character may be escaped.

contracts/token/ERC6909/draft-ERC6909.sol
```diff
@@ -166,6 +165,9 @@ contract ERC6909 is Context, ERC165, IERC6909 {
      *
      * - `owner` cannot be the zero address.
      * - `spender` cannot be the zero address.
+     *
+     * Note: Setting `amount` to `type(uint256).max` is equivalent to setting an "infinite" allowance that won't
+     * be reduced when used.
```
ernestognw, line 170: This will render it better in the docs page.

```suggestion
     * NOTE: Setting `amount` to `type(uint256).max` is equivalent to setting an "infinite" allowance that won't
     * be reduced when used.
```

Though, we may want to remove it as this is explicit in the ERC.

```suggestion
```

ernestognw (approved):
Approving since I think the comments I left are non-controversial

contracts/token/ERC6909/draft-ERC6909.sol
arr00: Agree on removing in favor of documenting in the interface.

contracts/token/ERC6909/draft-ERC6909.sol
```diff
@@ -93,10 +93,9 @@ contract ERC6909 is Context, ERC165, IERC6909 {
     }
 
     /**
-     * @dev Moves `amount` of token `id` from `from` to `to` without checking for approvals.
-     *
-     * This internal function is equivalent to {transfer}, and can be used to
-     * e.g. implement automatic token fees, slashing mechanisms, etc.
+     * @dev Moves `amount` of token `id` from `from` to `to` without checking for approvals. This function verifies
+     * that neither the sender or the receiver are address(0), which means it cannot mint or burn tokens.
```
arr00, line 97:
```suggestion
     * that neither the sender nor the receiver are address(0), which means it cannot mint or burn tokens.
```

contracts/token/ERC6909/draft-ERC6909.sol
arr00, line 170:
```suggestion
```

contracts/token/ERC6909/draft-ERC6909.sol
```diff
@@ -166,6 +165,7 @@ contract ERC6909 is Context, ERC165, IERC6909 {
      *
      * - `owner` cannot be the zero address.
      * - `spender` cannot be the zero address.
+     *
```
Amxx, line 168:
```suggestion
```

arr00 (approved):

contracts/interfaces/draft-IERC6909.sol
```diff
@@ -49,7 +49,8 @@ interface IERC6909 is IERC165 {
     function isOperator(address owner, address spender) external view returns (bool);
 
     /**
-     * @dev Sets an approval to `spender` for `amount` tokens of type `id` from the caller's tokens.
+     * @dev Sets an approval to `spender` for `amount` of tokens of type `id` from the caller's tokens. An `amount` of
```
Amxx, line 52: `amount` is a number, like 1, 17 or 42. 

I would not say Sets an approval to spender for 17 of tokens.

PR #5586, opened 2025-03-17
Use Entrypoint's provided hashing function to support v0.8.0 change of hash

Amxx:
TODO:

- [x] Add 0.8.0 entrypoint to common contracts 
- [x] Add ENTRYPOINT_V08 constant
- [x] Test `hash` against both entrypoints (v0.7.0 and v0.8.0)

This should be added to the 5.3 release as a fix to the current behavior that won't work with the next entrypoint.

Note: This is a fix to unreleased code. Doesn't need a changelog entry


contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -19,6 +24,9 @@ library ERC4337Utils {
     /// @dev Address of the entrypoint v0.7.0
     IEntryPoint internal constant ENTRYPOINT_V07 = IEntryPoint(0x0000000071727De22E5E9d8BAf0edAc6f37da032);
 
+    /// @dev Address of the entrypoint v0.8.0 - TBD
```
ernestognw, line 27:
```suggestion
    /// @dev Address of the entrypoint v0.8.0
```

contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -77,31 +85,17 @@ library ERC4337Utils {
...
-        return result;
+    /// @dev Get the hash of a user operation for a given entrypoint
+    function hash(PackedUserOperation calldata self, address entrypoint) internal view returns (bytes32) {
+        // NOTE: getUserOpHash is available since v0.4.0
+        //
+        // Prior to v0.8.0, this has was easy to replicate for any entrypoint and any chainId. Since v0.8.0 of the
+        // entrypoint, this depends on the Entrypoint's domain separator, which is cannot be hardcoded and is complex
+        // to recompute. Domain separator could be fetch using the `getDomainSeparatorV4` getter, or recomputed from
+        // the ERC-5267 getter, but both operation would require doing a view call to the entrypoint. Overall it feels
+        // simpler, less error prone, and probably not much more expensive to ask the entrypoint to compute the hash
```
ernestognw, line 96: I wouldn't be too sure that the alternative (i.e. re hashing) is more expensive. I would say the cost of copying `self` to memory and then call another (perhaps warm) address is more expensive.

I'd simply say:

```suggestion
        // simpler and less error prone
```

hardhat/common-contracts.js
```diff
@@ -6,41 +6,58 @@ const fs = require('fs');
...
-    abi: JSON.parse(fs.readFileSync(path.resolve(__dirname, '../test/bin/SenderCreator070.abi'), 'utf-8')),
-    bytecode: fs.readFileSync(path.resolve(__dirname, '../test/bin/SenderCreator070.bytecode'), 'hex'),
+    v07: {
+      address: '0xEFC2c1444eBCC4Db75e7613d20C6a62fF67A167C',
+      abi: JSON.parse(fs.readFileSync(path.resolve(__dirname, '../test/bin/SenderCreator070.abi'), 'utf-8')),
+      bytecode: fs.readFileSync(path.resolve(__dirname, '../test/bin/SenderCreator070.bytecode'), 'hex'),
+    },
+    v08: {
+      address: '0x449ED7C3e6Fee6a97311d4b55475DF59C44AdD33',
+      abi: [], // not verified yet
```
ernestognw: Seems this is already verified:
https://etherscan.io/address/0x449ED7C3e6Fee6a97311d4b55475DF59C44AdD33#code

scripts/fetch-common-contracts.js
ernestognw, line 1: We should add some instructions to this script

contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -77,31 +85,17 @@ library ERC4337Utils {
...
+    /// @dev Get the hash of a user operation for a given entrypoint
+    function hash(PackedUserOperation calldata self, address entrypoint) internal view returns (bytes32) {
+        // NOTE: getUserOpHash is available since v0.4.0
+        //
+        // Prior to v0.8.0, this has was easy to replicate for any entrypoint and any chainId. Since v0.8.0 of the
+        // entrypoint, this depends on the Entrypoint's domain separator, which cannot be hardcoded and is complex
+        // to recompute. Domain separator could be fetch using the `getDomainSeparatorV4` getter, or recomputed from
+        // the ERC-5267 getter, but both operation would require doing a view call to the entrypoint. Overall it feels
+        // simpler and less error prone
+        // directly.
```
Amxx, line 97:
```suggestion
        // simpler and less error prone to get that functionality from the entrypoint directly.
```

scripts/fetch-common-contracts.js
arr00: Echoing this comment. I think a comment at the top with a short explanation is worthwhile given that the script isn't called automatically from anywhere.

contracts/account/utils/draft-ERC4337Utils.sol
```diff
@@ -8,6 +8,11 @@ import {Math} from "../../utils/math/Math.sol";
 import {Calldata} from "../../utils/Calldata.sol";
 import {Packing} from "../../utils/Packing.sol";
 
+/// @dev This is available on all entrypoint since v0.4.0, but is not formally part of the ERC.
+interface IEntryPointExtra {
+    function getUserOpHash(PackedUserOperation calldata userOp) external view returns (bytes32);
```
0xy0k, line 13: what about adding a fuzz test for `getUserOpHash` across versions using known vectors?

Amxx: A fuzz test assumes we have an independant way of checking the produced hash is valid. How would you do that ?

scripts/fetch-common-contracts.js
Amxx: Done. Let me know what you think.

scripts/fetch-common-contracts.js
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env node
+
+// This script snapshots the bytecode and ABI for the `hardhat/common-contracts.js` script.
+// - Bytecode is fetched directly from the blockchain by querying the provided client endpoint. If not endpoint is
```
arr00, line 4:
```suggestion
// - Bytecode is fetched directly from the blockchain by querying the provided client endpoint. If no endpoint is
```

scripts/fetch-common-contracts.js
```diff
@@ -0,0 +1,50 @@
...
+//   provided, ethers default provider is used instead.
+// - ABI is fetched from etherscan's API using the provided etherscan API key. If no api key is provided, ABI will not
```
arr00, line 6:
```suggestion
// - ABI is fetched from etherscan's API using the provided etherscan API key. If no API key is provided, ABI will not
```

scripts/fetch-common-contracts.js
```diff
@@ -0,0 +1,50 @@
...
+// produced:
+// - `<name>.bytecode` containing the contract bytecode (in binary encoding)
+// - `<name>.abi` containing the ABI (in utf-8 encoding)
+
+const fs = require('fs');
+const path = require('path');
+const { ethers } = require('ethers');
+const { request } = require('undici');
+
+const { argv } = require('yargs/yargs')()
```
arr00, line 19:
```suggestion
const { argv } = require('yargs/yargs')(process.argv.slice(2))
```
Don't want to allow passing them as params?

arr00 (approved):

scripts/fetch-common-contracts.js
Amxx: I though Yargs would do it automatically. Added that using the hideBin utils that is more precise. 
`.slice(2)` would not work if calling the file directly (it is `chmod+x` with the `#!`)

PR #5600, opened 2025-03-24
Use new tagged slither action release

arr00:
Follow up to #5594. The new release includes the required changes.

#### PR Checklist

- [ ] Tests
- [ ] Documentation
- [ ] Changeset entry (run `npx changeset add`)

PR #5602, opened 2025-03-25
Implement ERC-7821 calldata compression in ERC7579Utils

Amxx:
ERC-7821 implementation uses the ERC7579Utils library. While ERC-7821 looks like a subset of ERC-7579, it includes a [specific design choice to help compress calldata](https://eips.ethereum.org/EIPS/eip-7821#replacing-address0-with-addressthis) that is not strictly part of ERC-7579.

When asked if it would be fine to add this mechanism to ERC-7579, Konrad (ERC-7579 authors) says:
> Hadrien Croubois, [21/03/2025 09:56]
> [...]. Since we are using the same library for 7821 and 7579 implementations, I was wondering if it would be ok to add the same behavior to 7579.
>> Konrad | rhinestone, [21/03/2025 22:51]
>> I dont see a problem with this but one point to note is that this will cause differing behaviour to existing accounts, so you can't really rely on that feature working across accounts

This PR adds the target rewrite mechanism to be used in both 7821 (where it is required) and 7579 (where is is not mandated, but also not a problem).

---

No need for a changeset entry if this is part of 5.3 audit fixes (unreleased feature).

#### PR Checklist

- [x] Tests

### DIFF
--- a/contracts/account/utils/draft-ERC7579Utils.sol
+++ b/contracts/account/utils/draft-ERC7579Utils.sol
@@ -218,7 +218,9 @@ library ERC7579Utils {
         uint256 value,
         bytes calldata data
     ) private returns (bytes memory) {
-        (bool success, bytes memory returndata) = target.call{value: value}(data);
+        (bool success, bytes memory returndata) = (target == address(0) ? address(this) : target).call{value: value}(
+            data
+        );
         return _validateExecutionMode(index, execType, success, returndata);
     }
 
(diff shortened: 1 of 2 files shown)


ernestognw (approved):
I understand the CI is failing because we would need to implement changes from #5600 into the release-v5.3 branch

Can we point to `master` instead and then cherry pick into `release-v5.3`? We should do this together during tomorrow's sync

Amxx:
updated to target master

ernestognw (approved):

PR #5614, opened 2025-03-31
Add missing 5.3 changesets

ernestognw:
Adding missing changesets and a missing custom error entry in the CHANGELOG

#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx (approved):
There changeset should be in the changelog of a new RC, but should not be part of the final release because they are uopdating the behavior of yet unreleased code.

PR #5620, opened 2025-04-02
Fix issue with detection of RIP7212 precompile

### LINKED ISSUE #5619: P256.verifyNative reverts when signature is invalid
According to [RIP-7212](https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md), when a signature is invalid, no data is returned. Our current implementation expects a false boolean to be returned, which would allow us to distinguish "invalid signature" and "absence of precompile".

This expectation being invalid, the P256 library present a few issues:
- [P256.verifyNative](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/P256.sol#L72-L79) will revert with "MissingPrecompile" if the signature is invalid (and the precompile is present)
- [P256.verify](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/P256.sol#L61C14-L64) will run the expensive solidity version if the signature is invalid (wasting a lot of gas)

Amxx:
Fixes #5619

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/.changeset/mighty-melons-cheer.md
+++ b/.changeset/mighty-melons-cheer.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-`P256`: Fix precompile detection, avoid incorrect reverts in `verifyNative` and reduce cost of `verify` when the signature is invalid.
+`P256`: Adjust precompile detection in `verifyNative` to consider empty `returndata` on invalid verification. Previously, invalid signatures would've reverted with a `MissingPrecompile` error in chains with RIP-7212 support.
--- a/.changeset/mighty-melons-cheer.md
+++ b/.changeset/mighty-melons-cheer.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`P256`: Fix precompile detection, avoiding revert in `verifyNative`, and reducing cost of `verify` when the signature is invalid.
--- a/.changeset/mighty-melons-cheer.md
+++ b/.changeset/mighty-melons-cheer.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-`P256`: Fix precompile detection, avoiding revert in `verifyNative`, and reducing cost of `verify` when the signature is invalid.
+`P256`: Fix precompile detection, avoid incorrect reverts in `verifyNative`, and reduce cost of `verify` when the signature is invalid.
--- a/.changeset/mighty-melons-cheer.md
+++ b/.changeset/mighty-melons-cheer.md
@@ -2,4 +2,4 @@
 'openzeppelin-solidity': patch
 ---
 
-`P256`: Fix precompile detection, avoid incorrect reverts in `verifyNative`, and reduce cost of `verify` when the signature is invalid.
+`P256`: Fix precompile detection, avoid incorrect reverts in `verifyNative` and reduce cost of `verify` when the signature is invalid.
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -90,10 +90,42 @@ library P256 {
     ) private view returns (bool valid, bool supported) {
         if (!_isProperSignature(r, s) || !isValidPublicKey(qx, qy)) {
             return (false, true); // signature is invalid, and its not because the precompile is missing
+        } else if (_erc7212(h, r, s, qx, qy)) {
+            return (true, true); // precompile is present, signature is valid
+        } else if (
+            _erc7212(
+                0x4cee90eb86eaa050036147a12d49004b6b9c72bd725d39d4785011fe190f0b4d,
+                0xa73bd4903f0ce3b639bbbf6e8e80d16931ff4bcf5993d58468e8fb19086e8cac,
+                0x36dbcd03009df8c59286b162af3bd7fcc0450c9aa81be5d10d312af6c66b1d60,
+                0x4aebd3099c618202fcfe16ae7770b0c49ab5eadf74b754204a3bb6060e44eff3,
+                0x7618b065f9832de4ca6ca971a7a1adc826d0f7c00181a5fb2ddf79ae00b4e10e
+            )
+        ) {
+            return (false, true); // precompile is present, signature is invalid
+        } else {
+            return (false, false); // precompile is absent
         }
+    }
 
-        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
-        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
+    /**
+     * @dev Low level helper for {_tryVerifyNative}. Calls the recompile and check if there is a return value.
+     *
+     * Note: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
+     * Getting the success boolean, copying the returndata to memory, and loading it as a boolean, is not strictly
+     * speaking necessary, but it's protects against non standard implementation that would return 0 (false) for
+     * invalid signatures.
+     */
+    function _erc7212(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) private view returns (bool isValid) {
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            mstore(ptr, h)
+            mstore(add(ptr, 0x20), r)
+            mstore(add(ptr, 0x40), s)
+            mstore(add(ptr, 0x60), qx)
+            mstore(add(ptr, 0x80), qy)
+            let success := staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)
+            isValid := and(success, and(eq(returndatasize(), 0x20), eq(mload(0x0), 1)))
+        }
     }
 
     /**
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -110,7 +110,7 @@ library P256 {
     /**
      * @dev Low level helper for {_tryVerifyNative}. Calls the precompile and checks if there is a return value.
      *
-     * Note: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
+     * NOTE: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
      * Getting the success boolean, copying the returndata to memory, and loading it as a boolean, is not strictly
      * necessary, but it protects against non-standard implementations that would return 0 (false) for
      * invalid signatures.
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -124,7 +124,7 @@ library P256 {
             mstore(add(ptr, 0x60), qx)
             mstore(add(ptr, 0x80), qy)
             let success := staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)
-            isValid := and(success, and(eq(returndatasize(), 0x20), eq(mload(0x0), 1)))
+            isValid := and(success, and(eq(returndatasize(), 0x20), eq(mload(0), 1)))
         }
     }
 
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -93,12 +93,15 @@ library P256 {
         } else if (_rip7212(h, r, s, qx, qy)) {
             return (true, true); // precompile is present, signature is valid
         } else if (
+            // Given precompiles have no bytecode (i.e. `address(0x100).code.length == 0`), we use
+            // a valid signature with small `r` and `s` values to check if the compiler is present. Taken from
+            //  https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204
             _rip7212(
-                0x532eaabd9574880dbf76b9b8cc00832c20a6ec113d682299550d7a6e0f345e25,
+                0xbb5a52f42f9c9261ed4361f59422a1e30036e7c32b270c8807a419feca605023, // sha256("123400")
                 0x0000000000000000000000000000000000000000000000000000000000000005,
                 0x0000000000000000000000000000000000000000000000000000000000000001,
-                0x4a03ef9f92eb268cafa601072489a56380fa0dc43171d7712813b3a19a1eb5e5,
-                0x3e213e28a608ce9a2f4a17fd830c6654018a79b3e0263d91a8ba90622df6f2f0
+                0x6627cec4f0731ea23fc2931f90ebe5b7572f597d20df08fc2b31ee8ef16b1572,
+                0x6170ed77d8d0a14fc5c9c3c4c9be7f0d3ee18f709bb275eaf2073e258fe694a5
             )
         ) {
             return (false, true); // precompile is present, signature is invalid
@@ -110,7 +113,7 @@ library P256 {
     /**
      * @dev Low level helper for {_tryVerifyNative}. Calls the precompile and checks if there is a return value.
      *
-     * NOTE: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
+     * NOTE: According to RIP-7212, invalid signatures are indistinguishable from the absence of the precompile.
      * Getting the success boolean, copying the returndata to memory, and loading it as a boolean, is not strictly
      * necessary, but it protects against non-standard implementations that would return 0 (false) for
      * invalid signatures.
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -94,7 +94,7 @@ library P256 {
             return (true, true); // precompile is present, signature is valid
         } else if (
             // Given precompiles have no bytecode (i.e. `address(0x100).code.length == 0`), we use
-            // a valid signature with small `r` and `s` values to check if the compiler is present. Taken from
+            // a valid signature with small `r` and `s` values to check if the precompile is present. Taken from
             //  https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204
             _rip7212(
                 0xbb5a52f42f9c9261ed4361f59422a1e30036e7c32b270c8807a419feca605023, // sha256("123400")
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -95,7 +95,7 @@ library P256 {
         } else if (
             // Given precompiles have no bytecode (i.e. `address(0x100).code.length == 0`), we use
             // a valid signature with small `r` and `s` values to check if the precompile is present. Taken from
-            //  https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204
+            // https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204
             _rip7212(
                 0xbb5a52f42f9c9261ed4361f59422a1e30036e7c32b270c8807a419feca605023, // sha256("123400")
                 0x0000000000000000000000000000000000000000000000000000000000000005,
--- a/contracts/utils/cryptography/P256.sol
+++ b/contracts/utils/cryptography/P256.sol
@@ -94,11 +94,11 @@ library P256 {
             return (true, true); // precompile is present, signature is valid
         } else if (
             _rip7212(
-                0x4cee90eb86eaa050036147a12d49004b6b9c72bd725d39d4785011fe190f0b4d,
-                0xa73bd4903f0ce3b639bbbf6e8e80d16931ff4bcf5993d58468e8fb19086e8cac,
-                0x36dbcd03009df8c59286b162af3bd7fcc0450c9aa81be5d10d312af6c66b1d60,
-                0x4aebd3099c618202fcfe16ae7770b0c49ab5eadf74b754204a3bb6060e44eff3,
-                0x7618b065f9832de4ca6ca971a7a1adc826d0f7c00181a5fb2ddf79ae00b4e10e
+                0x532eaabd9574880dbf76b9b8cc00832c20a6ec113d682299550d7a6e0f345e25,
+                0x0000000000000000000000000000000000000000000000000000000000000005,
+                0x0000000000000000000000000000000000000000000000000000000000000001,
+                0x4a03ef9f92eb268cafa601072489a56380fa0dc43171d7712813b3a19a1eb5e5,
+                0x3e213e28a608ce9a2f4a17fd830c6654018a79b3e0263d91a8ba90622df6f2f0
             )
         ) {
             return (false, true); // precompile is present, signature is invalid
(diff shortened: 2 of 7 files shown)


contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,42 @@ library P256 {
...
+    function _erc7212(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) private view returns (bool isValid) {
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            mstore(ptr, h)
+            mstore(add(ptr, 0x20), r)
+            mstore(add(ptr, 0x40), s)
+            mstore(add(ptr, 0x60), qx)
+            mstore(add(ptr, 0x80), qy)
+            let success := staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)
+            isValid := and(success, and(eq(returndatasize(), 0x20), eq(mload(0x0), 1)))
```
Amxx, line 127:
```suggestion
            isValid := and(success, and(eq(returndatasize(), 0x20), eq(mload(0), 1)))
```

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,42 @@ library P256 {
...
+            return (false, true); // precompile is present, signature is invalid
+        } else {
+            return (false, false); // precompile is absent
         }
+    }
 
-        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
-        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
+    /**
+     * @dev Low level helper for {_tryVerifyNative}. Calls the recompile and check if there is a return value.
```
arr00, line 111:
```suggestion
     * @dev Low level helper for {_tryVerifyNative}. Calls the precompile and checks if there is a return value.
```

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,42 @@ library P256 {
...
+     *
+     * Note: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
+     * Getting the success boolean, copying the returndata to memory, and loading it as a boolean, is not strictly
+     * speaking necessary, but it's protects against non standard implementation that would return 0 (false) for
```
arr00, line 115:
```suggestion
     * necessary, but it protects against non-standard implementations that would return 0 (false) for
```

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,42 @@ library P256 {
...
+     * invalid signatures.
+     */
+    function _erc7212(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) private view returns (bool isValid) {
```
arr00, line 118: RIP = ERC?

Amxx: erc is indeed wrong here

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,42 @@ library P256 {
...
+            return (false, false); // precompile is absent
         }
+    }
 
-        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
-        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
+    /**
+     * @dev Low level helper for {_tryVerifyNative}. Calls the precompile and checks if there is a return value.
+     *
+     * Note: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
```
arr00, line 113:
```suggestion
     * Note: According to RIP-7212, invalid signatures are indistinguishable from the absence of the precompile.
```

.changeset/mighty-melons-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`P256`: Fix precompile detection, avoiding revert in `verifyNative`, and reducing cost of `verify` when the signature is invalid.
```
arr00, line 5:
```suggestion
`P256`: Fix precompile detection, avoid incorrect reverts in `verifyNative`, and reduce cost of `verify` when the signature is invalid.
```

arr00 (approved):

contracts/utils/cryptography/P256.sol
Amxx, line 113:
```suggestion
     * NOTE: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
```

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,42 @@ library P256 {
...
+        } else if (_rip7212(h, r, s, qx, qy)) {
+            return (true, true); // precompile is present, signature is valid
+        } else if (
+            _rip7212(
+                0x4cee90eb86eaa050036147a12d49004b6b9c72bd725d39d4785011fe190f0b4d,
+                0xa73bd4903f0ce3b639bbbf6e8e80d16931ff4bcf5993d58468e8fb19086e8cac,
+                0x36dbcd03009df8c59286b162af3bd7fcc0450c9aa81be5d10d312af6c66b1d60,
+                0x4aebd3099c618202fcfe16ae7770b0c49ab5eadf74b754204a3bb6060e44eff3,
+                0x7618b065f9832de4ca6ca971a7a1adc826d0f7c00181a5fb2ddf79ae00b4e10e
+            )
```
Vectorized, line 102: Could use a Wycheproof vector with smaller bit length to save some bytecode.

Also feels better to know where these numbers came from.

```
{
    "x": "4a03ef9f92eb268cafa601072489a56380fa0dc43171d7712813b3a19a1eb5e5",
    "y": "3e213e28a608ce9a2f4a17fd830c6654018a79b3e0263d91a8ba90622df6f2f0",
    "r": "0000000000000000000000000000000000000000000000000000000000000005",
    "s": "0000000000000000000000000000000000000000000000000000000000000001",
    "hash": "532eaabd9574880dbf76b9b8cc00832c20a6ec113d682299550d7a6e0f345e25",
    "valid": true,
    "msg": "54657374",
    "comment": "wycheproof/ecdsa_webcrypto_test.json EcdsaP1363Verify SHA-256 #118: small r and s"
}
```

https://github.com/AmadiMichael/p256-verifier-huff/blob/3e48c21a731d3ffdcd86985b5cba574dcab4cf78/test-vectors/vectors_wycheproof.jsonl#L682C1-L682C484

https://github.com/C2SP/wycheproof/blob/aca47066256c167f0ce04d611d718cc85654341e/testvectors/ecdsa_webcrypto_test.json#L1197

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,42 @@ library P256 {
...
+        assembly ("memory-safe") {
+            let ptr := mload(0x40)
+            mstore(ptr, h)
+            mstore(add(ptr, 0x20), r)
+            mstore(add(ptr, 0x40), s)
+            mstore(add(ptr, 0x60), qx)
+            mstore(add(ptr, 0x80), qy)
+            let success := staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)
+            isValid := and(success, and(eq(returndatasize(), 0x20), eq(mload(0), 1)))
+        }
```
Vectorized, line 128:
```solidity
// The `invalid` upon `staticcall` failure is solely for gas estimation.
// When given sufficient gas, the precompile will not revert.
if iszero(staticcall(gas(), 0x100, m, 0xa0, 0x00, 0x00)) { invalid() }
// The precompile will only return `uint256(1)` if and only if the signature is valid.
// As per latest spec, verified against the op-geth implementation:
// See: https://github.com/ethereum-optimism/op-geth/blob/02dfe8692a3c606dbabd83c1ced2037aab9753d7/core/vm/contracts.go#L1342
// Otherwise, it will always return no data.
isValid := iszero(iszero(returndatasize()))
```

More conservative (in case the spec and implementations are somehow edited again):

```solidity
// The `invalid` upon `staticcall` failure is solely for gas estimation.
// When given sufficient gas, the precompile will not revert.
mstore(0x00, 0) // Zeroize the slot for the returndata.
if iszero(staticcall(gas(), 0x100, m, 0xa0, 0x00, 0x20)) { invalid() }
// The precompile will only return `uint256(1)` if and only if the signature is valid.
// As per latest spec, verified against the op-geth implementation:
// See: https://github.com/ethereum-optimism/op-geth/blob/02dfe8692a3c606dbabd83c1ced2037aab9753d7/core/vm/contracts.go#L1342
// Otherwise, it will always return no data.
isValid := mload(0x00)
```

contracts/utils/cryptography/P256.sol
Amxx: These values come from the tests in Geth PR for 7212.

But agreed that using one of Wycheproof vector is possibly better, particularly if it makes small r and s values

contracts/utils/cryptography/P256.sol
Amxx: I'd like to avoid the `invalid()` here. I think its better if the function returns false instead of reverting if we are in a strange situation where the precompile acts eraticly

Amxx: also, aren't `if`s quite expensive compared to basic `and` operations?

Amxx: should we worry about the precompile returning something that is neither 0 nor 1 ?

Amxx: If you believe the precompile will not revert, is there any reason to do 
```if iszero(staticcall(gas(), 0x100, m, 0xa0, 0x00, 0x20)) { invalid() }``` over ```pop(staticcall(gas(), 0x100, m, 0xa0, 0x00, 0x20))``` ?

File: contracts/utils/cryptography/P256.sol [hunk 9]
```diff
@@ -90,10 +90,45 @@ library P256 {
...
+        } else if (_rip7212(h, r, s, qx, qy)) {
+            return (true, true); // precompile is present, signature is valid
+        } else if (
+            _rip7212(
+                0x532eaabd9574880dbf76b9b8cc00832c20a6ec113d682299550d7a6e0f345e25,
+                0x0000000000000000000000000000000000000000000000000000000000000005,
+                0x0000000000000000000000000000000000000000000000000000000000000001,
+                0x4a03ef9f92eb268cafa601072489a56380fa0dc43171d7712813b3a19a1eb5e5,
+                0x3e213e28a608ce9a2f4a17fd830c6654018a79b3e0263d91a8ba90622df6f2f0
+            )
```
ernestognw, line 102: I think we need 1) an explanation here and 2) a link to the signature source.

For 2), I'd prefer using [this example](https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204) since we're using [those values for testing](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/05f218fb6617932e56bf5388c3b389c3028a7b73/test/utils/cryptography/P256.test.js#L123)

```suggestion
            // Given precompiles have no bytecode (i.e. `address(0x100).code.length == 0`), we use
            // a valid signature with small `r` and `s` values to check if the compiler is present. Taken from
            //  https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204
            _rip7212(
                0xbb5a52f42f9c9261ed4361f59422a1e30036e7c32b270c8807a419feca605023, // sha256("123400")
                0x0000000000000000000000000000000000000000000000000000000000000005,
                0x0000000000000000000000000000000000000000000000000000000000000001,
                0x6627cec4f0731ea23fc2931f90ebe5b7572f597d20df08fc2b31ee8ef16b1572,
                0x6170ed77d8d0a14fc5c9c3c4c9be7f0d3ee18f709bb275eaf2073e258fe694a5
            )
```

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,45 @@ library P256 {
...
+            return (false, false); // precompile is absent
         }
+    }
 
-        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
-        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
+    /**
+     * @dev Low level helper for {_tryVerifyNative}. Calls the precompile and checks if there is a return value.
+     *
+     * NOTE: According to RIP-7212, invalid signature are indistinguishable from the absence of the precompile.
```
ernestognw, line 113:
```suggestion
     * NOTE: According to RIP-7212, invalid signatures are indistinguishable from the absence of the precompile.
```

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,45 @@ library P256 {
...
+     */
+    function _rip7212(bytes32 h, bytes32 r, bytes32 s, bytes32 qx, bytes32 qy) private view returns (bool isValid) {
+        assembly ("memory-safe") {
+            mstore(0x00, 0)
+            let ptr := mload(0x40)
+            mstore(ptr, h)
+            mstore(add(ptr, 0x20), r)
+            mstore(add(ptr, 0x40), s)
+            mstore(add(ptr, 0x60), qx)
+            mstore(add(ptr, 0x80), qy)
```
ernestognw, line 126: I don't understand this function, aren't these the memory pointer (0x40), the zero slot (0x60) and an actual memory value (0x80)? 

I understand we want to avoid expanding memory for this call but I think this is a mistake

Vectorized: The assembly is correct.

`ptr` points to the start of the free memory that can be used for anything.

contracts/utils/cryptography/P256.sol
Vectorized: I think we don't need to worry about it.

contracts/utils/cryptography/P256.sol
Amxx: @ernestognw ptr is not `0x40`, its `mload(0x40)`

Basically that is the start of the unallocated memory, were we can "work" without having to clean up after.
You can see that pattern used in many places ([example](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MessageHashUtils.sol#L90-L98))

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,48 @@ library P256 {
     ) private view returns (bool valid, bool supported) {
         if (!_isProperSignature(r, s) || !isValidPublicKey(qx, qy)) {
             return (false, true); // signature is invalid, and its not because the precompile is missing
+        } else if (_rip7212(h, r, s, qx, qy)) {
+            return (true, true); // precompile is present, signature is valid
+        } else if (
+            // Given precompiles have no bytecode (i.e. `address(0x100).code.length == 0`), we use
+            // a valid signature with small `r` and `s` values to check if the compiler is present. Taken from
```
Amxx, line 97:
```suggestion
            // a valid signature with small `r` and `s` values to check if the precompile is present. Taken from
```

contracts/utils/cryptography/P256.sol
ernestognw: lmao yeah I'm too dumb. Solving

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,48 @@ library P256 {
     ) private view returns (bool valid, bool supported) {
         if (!_isProperSignature(r, s) || !isValidPublicKey(qx, qy)) {
             return (false, true); // signature is invalid, and its not because the precompile is missing
+        } else if (_rip7212(h, r, s, qx, qy)) {
+            return (true, true); // precompile is present, signature is valid
+        } else if (
+            // Given precompiles have no bytecode (i.e. `address(0x100).code.length == 0`), we use
+            // a valid signature with small `r` and `s` values to check if the precompile is present. Taken from
+            //  https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204
```
ernestognw, line 98:
```suggestion
            // https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1181-L1204
```

contracts/utils/cryptography/P256.sol
Amxx: For some reason, this was was not using the correct values from the linked example. qx and qy values should have been these ones
http://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1178-L1179

Fixed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5620/commits/7478900a9726efe1fcea0e41e1c1edc6f447307f

Amxx: (you basically copied the qx and qy values from the example after that https://github.com/C2SP/wycheproof/blob/4672ff74d68766e7785c2cac4c597effccef2c5c/testvectors/ecdsa_secp256r1_sha256_p1363_test.json#L1206-L1246

contracts/utils/cryptography/P256.sol
ernestognw: I like @Vectorized's idea since the library is designed to work with RIP-7212 so it should be safe to assume that all implementations would be compliant (i.e. returning empty bytes on failed verification). We can add a note so that developers can just override the function and manually check the returndata, but I'm afraid the consequences of missing such note would be pretty bad.

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,48 @@ library P256 {
...
+            mstore(0x00, 0)
+            let ptr := mload(0x40)
+            mstore(ptr, h)
+            mstore(add(ptr, 0x20), r)
+            mstore(add(ptr, 0x40), s)
+            mstore(add(ptr, 0x60), qx)
+            mstore(add(ptr, 0x80), qy)
+            if iszero(staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)) {
+                invalid()
+            }
```
ernestognw, line 132: Replacing the natspec

```suggestion
            // RIP-7212 precompiles return empty bytes when an invalid signature is passed, making it impossible
            // to distinguish the presence of the precompile. Custom precompile implementations may decide to
            // return `bytes32(0)` (i.e. false) without developers noticing, so we decide to evaluate the return value
            // without expanding memory.
            if iszero(staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)) {
                invalid()
            }
```

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,48 @@ library P256 {
...
 
-        (bool success, bytes memory returndata) = address(0x100).staticcall(abi.encode(h, r, s, qx, qy));
-        return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);
+    /**
+     * @dev Low level helper for {_tryVerifyNative}. Calls the precompile and checks if there is a return value.
+     *
+     * NOTE: According to RIP-7212, invalid signatures are indistinguishable from the absence of the precompile.
+     * Getting the success boolean, copying the returndata to memory, and loading it as a boolean, is not strictly
+     * necessary, but it protects against non-standard implementations that would return 0 (false) for
+     * invalid signatures.
```
ernestognw, line 119: This note doesn't provide the developers with any useful context. Instead, this justifies why we're copying the returndata everytime in the implementation, but I wouldn't note it.

```suggestion
```

Alternatively, if we were not copying the returndata and evaluating `iszero(iszero(returndatasize()))`, then I'd replace this note with asking the developer to thoroughly check the returndata if the chain they're deploying at has a custom precompile (though ask btw)

ernestognw (approved):
Approving because I agree with the functionality. However, I'm leaving another nit. Feel free to merge if you don't agree @Amxx.

.changeset/mighty-melons-cheer.md
ernestognw, line 1: Wondering if we need a changeset. The developers would only noticed that the failure case is more costly but invalid verifications remain invalid. The only perceived change is that the error is different and we've never recommended relying on custom errors for on-chain logic

contracts/utils/cryptography/P256.sol
arr00: Agree this note belongs inline and not in the natspec.

.changeset/mighty-melons-cheer.md
arr00: `verifyNative` would result in false being returned whereas previously it would always return true or revert.

arr00 (approved):

.changeset/mighty-melons-cheer.md
ernestognw: `verifyNative` has always had instructions about ensuring the precompile is supported, so I think it's safe to assume nobody would've used it in a chain without the precompile.

I guess practically (please correct me if wrong), the difference for developers is that the precompile response would've been interpreted as `(false, false)` in `return (success && returndata.length == 0x20) ? (abi.decode(returndata, (bool)), true) : (false, false);` (previously line 96), resulting on a revert with `MissingPrecompile`

.changeset/mighty-melons-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+`P256`: Fix precompile detection, avoid incorrect reverts in `verifyNative` and reduce cost of `verify` when the signature is invalid.
```
ernestognw, line 5:
```suggestion
`P256`: Adjust precompile detection in `verifyNative` to consider empty `returndata` on invalid verification. Previously, invalid signatures would've reverted with a `MissingPrecompile` error.
```

.changeset/mighty-melons-cheer.md
ernestognw: Cool, would you say this changeset adjustment would encompass the idea better?

https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5620/files#r2025554000

contracts/utils/cryptography/P256.sol
```diff
@@ -90,10 +90,48 @@ library P256 {
...
+            let ptr := mload(0x40)
+            mstore(ptr, h)
+            mstore(add(ptr, 0x20), r)
+            mstore(add(ptr, 0x40), s)
+            mstore(add(ptr, 0x60), qx)
+            mstore(add(ptr, 0x80), qy)
+            if iszero(staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)) {
+                invalid()
+            }
+            isValid := mload(0x00)
```
ernestognw, line 133: I just noticed that this is a breaking change. We shouldn't revert here and instead just return `isValid = false`. Otherwise, the `MissingPrecompile` error becomes unreachable.

```suggestion
            if staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20) {
                isValid := mload(0x00)
            }
```

I'll go ahead and update since it seems a good reason IMO. I'll add tests as well

ernestognw (changes requested):

contracts/utils/cryptography/P256.sol
ernestognw: Actually, if we keep the expectation that the precompile implementation is compliant with `RIP7212`, then we can just do:

```
            let success := staticcall(gas(), 0x100, ptr, 0xa0, 0x00, 0x20)
            isValid := and(success, eq(mload(0), 1))
```

ernestognw: nvm I was hitting the `invalid()` with a wrong mock, still pushing the tests and comments adjustments

ernestognw (approved):

ernestognw (approved):

PR #5626, opened 2025-04-04
Add v5.3 audit report

Amxx:


audits/README.md
```diff
@@ -2,6 +2,7 @@
 
 | Date          | Version | Commit                                                                           | Auditor      | Scope                | Links                                                       |
 | ------------- | ------- | -------------------------------------------------------------------------------- | ------------ | -------------------- | ----------------------------------------------------------- |
+| April 2025    | v5.3.0  | [`d4b2e98`](https://github.com/openzeppelin/openzeppelin-contracts/tree/d4b2e98) | OpenZeppelin | v5.3 Changes         | [🔗](./2025-04-v5.3.pdf)                                    |
```
ernestognw, line 5: I think we should be using this one instead

```suggestion
| April 2025    | v5.3.0  | [`f3f0a64 `](https://github.com/openzeppelin/openzeppelin-contracts/tree/f3f0a64) | OpenZeppelin | v5.3 Changes         | [🔗](./2025-04-v5.3.pdf)                                    |
```

Amxx: The 5.3 report states:

> We audited the OpenZeppelin/openzeppelin-contracts repository at commit d4b2e98.
> [...]
> Update: All resolutions and the final state of the audited codebase mentioned in this report are
contained at commit f3f0a64

If we look at the 5.2 audit report it says

> We audited the OpenZeppelin/openzeppelin-contracts repository at commit 98d28f9 

98d28f9 is the "start release candidate" commit, debore the fixes.

---

Basically, for the other lines, the commit reference is the one that was audited, before the fixes. Do we want consistency between the lines ?

Amxx: An option would be to have two columns
- Commit reviewed
- Commit after fixes

The second column would be empty for the old lines

ernestognw: I feel that's too much, instead of "commit after fixes" I'd do the "release commit" but we'd need another PR. Also that one is tagged on the release.

ernestognw (approved):

PR #5628, opened 2025-04-05
Add Ownable2StepSign

markolazic01:
This PR contains an extension of `Ownable2Step`, that I found useful for the following reasons:
- Reduces the amount of needed transactions for execution of an ownership transfer from 2 to 1 (for EOAs), while keeping the same level of protection as present in `Ownable2Step`.
- Provides an additional level of safety to the `newOwner` wallet at the time of the ownership transfer, as it can stay offline and therefore reduce the potential security threats.

Ownership over the `newOwner` EOA, as well as the acceptance of ownership transfer are proved via signature.
Call for an ownership transfer must be made by the current owner (as per usual), with the signature provided by the `newOwner`.

Support for 2step ownership transfer to contracts is inherited from `Ownable2Step`.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/access/Ownable2StepSign.sol
+++ b/contracts/access/Ownable2StepSign.sol
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Ownable2Step} from "./Ownable2Step.sol";
+import {ECDSA} from "../utils/cryptography/ECDSA.sol";
+import {EIP712} from "../utils/cryptography/EIP712.sol";
+
+/**
+ * @dev Contract module which provides a basic access control mechanism, where
+ * there is an account (an owner) that can be granted exclusive access to
+ * specific functions.
+ *
+ * This extension of the {Ownable2Step}, which is an extension of {Ownable} contract,
+ * includes a signature protected ownership transfer mechanism,
+ * where the `newOwner` must provide a signature in order to replace the
+ * old one. This can help prevent common mistakes, such as transfers of ownership to
+ * incorrect accounts, and keep the `newOwner` wallet safer, as it can stay offline at a time.
+ * Function `transferOwnership` available through {Ownable2Step} provides a safe way of making
+ * an ownership transfer to a contract.
+ *
+ * The initial owner is specified at deployment time in the constructor for `Ownable`. This
+ * can later be changed with {transferOwnership}.
+ *
+ * This module is used through inheritance. It will make available all functions
+ * from parents (Ownable2Step and Ownable).
+ */
+abstract contract Ownable2StepSign is Ownable2Step, EIP712 {
+    /**
+     * @dev Nonce used for signatures.
+     */
+    uint256 private _nonce;
+
+    bytes32 private constant TRANSFER_OWNERSHIP_TYPEHASH =
+        keccak256("TransferOwnership(uint256 nonce,uint256 deadline)");
+
+    /**
+     * @dev Permit deadline has expired.
+     */
+    error ExpiredSignature(uint256 deadline);
+
+    /**
+     * @dev Mismatched signature.
+     */
+    error InvalidSigner(address signer, address newOwner);
+
+    constructor(string memory name) EIP712(name, "1") {}
+
+    /**
+     * @dev Transfers ownership of the contract to a new account (`newOwner`).
+     * Can only be called by the current owner.
+     * Requires a `newOwner` account's signature.
+     */
+    function transferOwnership(
+        address newOwner,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) public virtual onlyOwner {
+        if (block.timestamp > deadline) {
+            revert ExpiredSignature(deadline);
+        }
+
+        bytes32 hash = _hashTypedDataV4(keccak256(abi.encode(TRANSFER_OWNERSHIP_TYPEHASH, _useNonce(), deadline)));
+
+        address signer = ECDSA.recover(hash, v, r, s);
+        if (signer != newOwner) {
+            revert InvalidSigner(signer, newOwner);
+        }
+
+        super._transferOwnership(newOwner);
+    }
+
+    /**
+     * @dev Consumes a nonce.
+     *
+     * Returns the current value and increments nonce.
+     */
+    function _useNonce() internal virtual returns (uint256) {
+        // Flow guarantees that the nonce never overflows.
+        unchecked {
+            // It is important to do x++ and not ++x here.
+            return _nonce++;
+        }
+    }
+
+    /**
+     * @dev Returns the next unused nonce.
+     */
+    function nonce() public view virtual returns (uint256) {
+        return _nonce;
+    }
+
+    // solhint-disable-next-line func-name-mixedcase
+    function DOMAIN_SEPARATOR() external view virtual returns (bytes32) {
+        return _domainSeparatorV4();
+    }
+}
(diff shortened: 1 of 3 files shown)


contracts/access/Ownable2StepSign.sol
```diff
@@ -0,0 +1,99 @@
...
+     * @dev Transfers ownership of the contract to a new account (`newOwner`).
+     * Can only be called by the current owner.
+     * Requires a `newOwner` account's signature.
+     */
+    function transferOwnership(
+        address newOwner,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
```
Amxx, line 59: New owner could be an ERC-1271 contract. We should use a `bytes calldata signature` and do 
``SignatureChecker.isValidSignatureNow` instead of ecrecover.

contracts/access/Ownable2StepSign.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @dev Returns the next unused nonce.
+     */
+    function nonce() public view virtual returns (uint256) {
+        return _nonce;
+    }
+
+    // solhint-disable-next-line func-name-mixedcase
+    function DOMAIN_SEPARATOR() external view virtual returns (bytes32) {
```
Amxx, line 96: IMO, we should not do that anywhere. EIP712.sol already comes with ERC-5267 support, which is a wayyyyy beter domain discovery method.

contracts/access/Ownable2StepSign.sol
Amxx: And if we start doing signature verification, maybe we can go all the way and use
`TransferOwnership(address newOwner,uint256 nonce,uint256 deadline)` that we get two signature for:
- one by the current owner
- one by the new owner

```solidity
function transferOwnership(
    address newOwner,
    uint256 deadline,
    bytes calldata previousOwnerSignature,
    bytes calldata newOwnerSignature
) public virtual {
    require(block.timestamp > deadline, ExpiredSignature(deadline));
    
    bytes32 hash = _hashTypedDataV4(keccak256(abi.encode(TRANSFER_OWNERSHIP_TYPEHASH, newOwner, _useNonce(), deadline)));
    require(isValidSignatureNow(owner(), hash, previousOwnerSignature), ...);
    require(isValidSignatureNow(newOwner, hash, newOwnerSignature), ...);
    
}
```

Amxx (commented):
Overall, I don't think there is anything terribly wrong with this PR in terms of design ... But I'm also not sure we want it.

That is not something we got many requests for, and its something that may further reduce the discoverability of our many features. 

Maybe we want to have that in the community repository ?

@ernestognw, @arr00, @gonzaotc  WDYT ?

contracts/access/Ownable2StepSign.sol
Amxx: Yes, there is no dependance on any of the `Ownable2Step` features, so there is no reason to depend on it.

contracts/access/Ownable2StepSign.sol
Amxx: We only haved it in `ERC20Permit` its part of the specifications for [ERC-2612](https://eips.ethereum.org/EIPS/eip-2612). If it wasn't for that, we wouldn't have it !

contracts/access/Ownable2StepSign.sol
markolazic01: Thank you! Will make this adaptation then.

Amxx:
Note:
- I would move that to the community repo instead. 
- This contract shouldn't be names "two steps" because its all in one operation. Instead it could be something like `OwnableRecipientSignOff` (to be discussed)
- Right now the contract assumes that its called by the owner, with the recipient signature. Should we also consider the case where the owner provides a signature, and the recipient calls? Or the case where both side provide a signature and anyone does the call?

markolazic01:
> Note:
> 
> * I would move that to the community repo instead.
> * This contract shouldn't be names "two steps" because its all in one operation. Instead it could be something like `OwnableRecipientSignOff` (to be discussed)
> * Right now the contract assumes that its called by the owner, with the recipient signature. Should we also consider the case where the owner provides a signature, and the recipient calls? Or the case where both side provide a signature and anyone does the call?

With the coming changes, especially referring to the inheritance of Ownable2Step being switched to Ownable, I will update the name (at least temporarily) to `OwnableRecipientSignOff`.

When it comes to the cases mentioned in the last point @Amxx, I have thought about that as well. My current conclusion is that existing form is the safest of the mentioned three options. This would be mostly due to the social aspects, as I feel that users are much more prone to giving away signatures then they are to making suspicious transactions. Also, I feel like it is safest to keep everything in the hands of the current owner, even when the owner receives the 'newOwner' signature, it will be possible for him to rethink the ownership transfer if needed (and if 'owner' is the one providing the signature, there will be no possibility to rethink).
This is, of course, just my view on this. I'd be happy to discuss the other aspects.

contracts/access/OwnableRecipientSignOff.sol
```diff
@@ -0,0 +1,85 @@
...
+     * @dev Permit deadline has expired.
+     */
+    error ExpiredSignature(uint256 deadline);
+
+    /**
+     * @dev Mismatched signature.
+     */
+    error InvalidSigner();
+
+    constructor(string memory name) EIP712(name, "1") {}
```
Amxx, line 45:
```suggestion
```

markolazic01: I've added this only to fix the version to `1`, if that is not desirable I will remove the constructor.

markolazic01:
Hey guys @Amxx @ernestognw @arr00 @gonzaotc, sorry to bother you, any chance we can proceed with this?

Amxx:
> Hey guys @Amxx @ernestognw @arr00 @gonzaotc, sorry to bother you, any chance we can proceed with this?

IMO, the code does not follow our standards. In particular, if a nonce is needed, it should come from the `Nonce` contract and not be reimplemented.

Also , I'm not sure what security this extension provides. The "basic" transferOwnership function is still available, so all the signature really can be bypassed by the owner.

Last but definitelly not least, this PR just doesn't follow our [contributing guideline](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/CONTRIBUTING.md) as the issue it tries to fix was not properly discussed in an issue first.

markolazic01:
Hello @Amxx, I'm glad to hear from you again : )

> In particular, if a nonce is needed, it should come from the `Nonce` contract and not be reimplemented.

In regards to this point, I decided to make the implementation as it is due to the `Nonces` contract's [nonce per account](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Nonces.sol#L14) approach. In this case a global nonce suffices so it seemed to me like a more optimized approach.
I can change this of course, if you consider the inheritance of a standard better.

> I'm not sure what security this extension provides. The "basic" transferOwnership function is still available, so all the signature really can be bypassed by the owner.

I've been meaning to ask you about that, but the conversation faded away and I didn't make any changes in regards to this matter. I've considered to override the `transferOwnership()` function implementation with a simple `revert()`, but I wasn't sure if that aligns with the repository standards. 
Please let me know if this solution satisfies the criteria or if there is a better one.

Other than that, this feature should let users safely confirm that they're in possession of a `newOwner` wallet, while making the ownership transfer in a single transaction (and also keep the `newOwner` wallet completely offline).
This seemed to me like an approach that further enhances security and potentially provides a better UX.

> Last but definitely not least, this PR just doesn't follow our [contributing guideline](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/CONTRIBUTING.md) as the issue it tries to fix was not properly discussed in an issue first.

I apologize for this inconvenience, does it help if I create the related issue now?

PR #5630, opened 2025-04-08
Add RelayedCall library to hide the `msg.sender` from a target contract

Amxx:
Some contract are required to perform arbitrary action controlled by user input. This is dangerous if the contract has special permissions, or holds assets. In such cases, using a relay contract can be useful to change the msg.sender of the outgoing call. This pattern is used in the ERC-4337 entrypoint that relies on a helper called the "senderCreator" when calling account factories. Similarly ERC-6942 does factory calls that could be dangerous if performed directly.

This library provides a `relayCall` that can be used to perform dangerous calls. These calls are indirect through a minimal relayer.

Edit: this can also be usefull for some construction that use a singleton with "satellites" contract that hold assets/permissions.

#### PR Checklist

- [x] Tests
- [ ] Documentation


contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Helper contract for performing potentially dangerous calls through a relay the hide the address of the
+ * original sender.
```
ernestognw:
```suggestion
 * @dev Library for performing external calls through dynamically deployed relay contracts that hide the original
 * caller's address.
```

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * Some contract are required to perform arbitrary action controlled by user input. This is dangerous if the contract
+ * has special permissions, or holds assets. In such cases, using a relay contract can be useful to change the
+ * msg.sender of the outgoing call. This pattern is used in the ERC-4337 entrypoint that relies on a helper called the
+ * "senderCreator" when calling account factories. Similarly ERC-6942 does factory calls that could be dangerous if
+ * performed directly.
```
ernestognw:
```suggestion
 * Some contracts need to make arbitrary external calls controlled by user input. When these contracts have special
 * permissions or hold valuable assets, making such calls directly can be risky because the target contract sees
 * the privileged contract as `msg.sender`. Using a relay contract isolates the original caller by changing the
 * `msg.sender` of the outgoing call to an unprivileged relay address.
 * 
 * For example, this pattern is used in ERC-4337's EntryPoint which relies on a "senderCreator" helper
 * for account factory calls, and ERC-6942 for safe factory interactions.
```

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * This contract provides a `indirectCall` that can be used to perform dangerous calls. These calls are indirect
+ * through a minimal relayer.
```
ernestognw:
```suggestion
 * The library dynamically deploys minimal relay contracts using CREATE2 and routes calls through them, ensuring
 * that target contracts only see the relay (not the original caller) as msg.sender.
```

contracts/utils/IndirectCall.sol
ernestognw, line 1: I don't like the "Indirect" naming because "Relay" is also mentioned a couple of times and I think both are trying to express the same. What about calling the library "RelayedCall" and the functions `relayCall`?

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    function getRelayer() internal returns (address) {
+        return getRelayer(bytes32(0));
+    }
+
+    function getRelayer(bytes32 salt) internal returns (address relayer) {
+        // [Relayer details]
+        //
+        // deployment prefix: 5f604780600a5f3981f3
```
ernestognw: The current version leaves a `0` in the stack and also returns data, which I think is not needed. Maybe this prefix is more succinct?

```suggestion
        // deployment prefix: 5f5f6047600a5f39f3
```

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+        // 0x0016 | 14          | eq             | access
+        // 0x0017 | 6013        | push1 0x13     | 0x13 access
+        // 0x0019 | 36          | calldatasize   | cds 0x13 access
+        // 0x001a | 11          | gt             | (cds>0x13) access
+        // 0x001b | 16          | and            | (cds>0x13 && access)
+        // 0x001c | 6022        | push1 0x22     | 0x22 (cds>0x13 && access)
+        // 0x001e | 57          | jumpi          |
+        // 0x001f | 5f          | push0          | 0
+        // 0x0020 | 5f          | push0          | 0 0
+        // 0x0021 | fd          | revert         |
```
ernestognw: [EXPLANATION NOTE]

Access Control. Equivalent to:

```
require(address(this) == factory && msg.data > 0x13)
```

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+        // 0x0022 | 5b          | jumpdest       |
+        // 0x0023 | 6014        | push1 0x14     | 0x14
+        // 0x0025 | 36          | calldatasize   | cds 0x14
+        // 0x0026 | 03          | sub            | (cds-0x14)
+        // 0x0027 | 6014        | push1 0x14     | 0x14 (cds-0x14)
+        // 0x0029 | 5f          | push0          | 0 0x14 (cds-0x14)
+        // 0x002a | 37          | calldatacopy   |
```
ernestognw: [EXPLANATION NOTE]

Since calling the relay address requires calldata to be `abi.encodePacked(target, data)`. This block skips the `target` and copies `data` to memory

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+        // 0x002f | 36          | calldatasize   | cds 0x14 0 0
+        // 0x0030 | 03          | sub            | (cds-0x14) 0 0
+        // 0x0031 | 5f          | push0          | 0 (cds-0x14) 0 0
+        // 0x0032 | 34          | callvalue      | value 0 (cds-0x14) 0 0
+        // 0x0033 | 5f          | push0          | 0 value 0 (cds-0x14) 0 0
+        // 0x0034 | 35          | calldataload   | cd[0] value 0 (cds-0x14) 0 0
+        // 0x0035 | 6060        | push1 0x60     | 0x60 cd[0] value 0 (cds-0x14) 0 0
+        // 0x0037 | 1c          | shr            | target value 0 (cds-0x14) 0 0
+        // 0x0038 | 5a          | gas            | gas target value 0 (cds-0x14) 0 0
+        // 0x0039 | f1          | call           | suc
```
ernestognw: [EXPLANATION NOTE]

Call execution:

1. [0x002c] Sets `retSize` and `retOffset` to 0
2. [0x002d:0x0030] Sets `argsSize` to calldatasize without the target (i.e. `data` in `abi.encodePacked(target, data)`)
3. [0x0031] Sets `argsOffset` to 0
4. [0x0032] Sets `value` to `msg.value`
5. [0x0033:0x0037] Gets first 32 bytes from calldata, then shifts it to the right by 0x60 bits (i.e. 96 bits), cleaning the target address
6. [0x0038] Sets all gas available 
7. [0x0039] Performs the call

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+        // 0x003a | 3d          | returndatasize | rds suc
+        // 0x003b | 5f          | push0          | 0 rds suc
+        // 0x003c | 5f          | push0          | 0 0 rds suc
+        // 0x003d | 3e          | returndatacopy | suc
```
ernestognw: [EXPLANATION NOTE]

Copies returndata from the relayed call.

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+        // 0x003e | 5f          | push0          | 0 suc
+        // 0x003f | 3d          | returndatasize | rds 0 suc
+        // 0x0040 | 91          | swap2          | suc 0 rds
+        // 0x0041 | 6045        | push1 0x45     | 0x45 suc 0 rds
+        // 0x0043 | 57          | jumpi          | 0 rds
+        // 0x0044 | fd          | revert         |
+        // 0x0045 | 5b          | jumpdest       | 0 rds
+        // 0x0046 | f3          | return         |
```
ernestognw: [EXPLANATION NOTE]

Revert if call is not successful. Bubble up returndata otherwise.

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+
+        assembly ("memory-safe") {
+            let fmp := mload(0x40)
+
+            // build initcode at FMP
+            mstore(add(fmp, 0x46), 0x60145f375f5f601436035f345f3560601c5af13d5f5f3e5f3d91604557fd5bf3)
+            mstore(add(fmp, 0x26), 0x331460133611166022575f5ffd5b60143603)
+            mstore(add(fmp, 0x14), address())
+            mstore(add(fmp, 0), 0x5f604780600a5f3981f373)
+            let initcodehash := keccak256(add(fmp, 0x15), 0x51)
```
ernestognw: [EXPLANATION NOTE]

Hash from 0x15 (after address) with bytecodesize length (0x51 = 81 bytes)

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+
+            // compute create2 address
+            mstore(0x40, initcodehash)
+            mstore(0x20, salt)
+            mstore(0x00, address())
+            mstore8(0x0b, 0xff)
+            relayer := and(keccak256(0x0b, 0x55), shr(96, not(0)))
```
ernestognw: Maybe we should use `Create2.computeAddress` here

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+            mstore(add(fmp, 0x46), 0x60145f375f5f601436035f345f3560601c5af13d5f5f3e5f3d91604557fd5bf3)
+            mstore(add(fmp, 0x26), 0x331460133611166022575f5ffd5b60143603)
+            mstore(add(fmp, 0x14), address())
+            mstore(add(fmp, 0), 0x5f604780600a5f3981f373)
+            let initcodehash := keccak256(add(fmp, 0x15), 0x51)
+
+            // compute create2 address
+            mstore(0x40, initcodehash)
+            mstore(0x20, salt)
+            mstore(0x00, address())
```
ernestognw: I think these aren't memory safe:


```suggestion
            mstore(add(fmp, 0x40), initcodehash)
            mstore(add(fmp, 0x20), salt)
            mstore(fmp, address())
```

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,134 @@
...
+            mstore(0x40, fmp)
+        }
+
+        // For reference: equivalent in solidity
+        // bytes memory initcode = abi.encodePacked(hex"5f604780600a5f3981f373", address(this), hex"331460133611166022575f5ffd5b6014360360145f375f5f601436035f345f3560601c5af13d5f5f3e5f3d91604557fd5bf3");
+        // address relayer = Create2.computeAddress(salt, keccak256(initcode));
+        // if (relayer.code.length == 0) {
+        //     Create2.deploy(0, salt, initcode);
+        // }
+        // return relayer;
```
ernestognw: I don't think this is needed

```suggestion
```

contracts/utils/IndirectCall.sol
Amxx: Here, we are using the scratch space + the memory at location 0x40 (FMP).
The FMP is restored on line 125

This should be memory safe according to https://docs.soliditylang.org/en/latest/assembly.html#advanced-safe-use-of-memory

Also, all the stuff at fmp is still necessary for the create2 that comes after, so we should no overwrite it.

contracts/utils/IndirectCall.sol
Amxx: We could, but in order to do so we would need to put the initcode in a proper `bytes memory` object to pass it. 

This would require moving eveything in memory 0x20 bytes to make room for the length part (more mstore + memory expansion).

Also Create2.computeAddress uses the memory at the fmp, so we need to reserve that object we just created. This means we need to move the FMP, and restore it later (more mload/mstore/complexity)

Overall, if would just be more expensive, and it would probably be less readable interlacing a solidity call in the middle of assembly.

contracts/utils/IndirectCall.sol
```diff
@@ -0,0 +1,129 @@
...
+                if iszero(create2(0, add(fmp, 0x15), 0x51, salt)) {
+                    returndatacopy(fmp, 0, returndatasize())
+                    revert(fmp, returndatasize())
+                }
+            }
+
+            // cleanup fmp space used as scratch
+            mstore(0x40, fmp)
+        }
+
```
Amxx, line 127:
```suggestion
        }
```

contracts/utils/IndirectCall.sol
Amxx: > Beyond the strict definition of memory-safety given above, there are cases in which you may want to use more than 64 bytes of scratch space starting at memory offset 0. If you are careful, it can be admissible to use memory up to (and not including) offset 0x80 and still safely declare the assembly block as memory-safe. This is admissible under either of the following conditions:
>
> - By the end of the assembly block, the free memory pointer at offset 0x40 is restored to a sane value (i.e. it is either restored to its original value or an increment of it due to a manual memory allocation), and the memory word at offset 0x60 is restored to a value of zero.
>
> - The assembly block terminates, i.e. execution can never return to high-level Solidity code. This is the case, for example, if your assembly block unconditionally ends in calling the revert opcode.

Here we have two options:
- we get all the way to the end of the function, in that case the line 125 restores the fmp, and we are good (case1)
- we get to the revert on line 120. In that case the fmp is NOT restored, by case 2 says its ok to not cleanup in that case.

contracts/utils/IndirectCall.sol
Amxx: `5f5f6047600a5f39f3` returns (0, 0) instead or (0, 0x47) ... so it would not work. But yes, we may be able to save one instruction here

Amxx: changing to `60475f8160095f39f3`:
- remove the first push0
- update the offset from 0xA to 0x9
- replace the dup with a push0

```suggestion
        // deployment prefix: 60475f8160095f39f3
```

ernestognw: A gotcha, I missed that the dangling (0, 0x47) was for the RETURN

ernestognw (approved):

PR #5636, opened 2025-04-12
Migrate to inheritdoc

### LINKED ISSUE #5563: Natspec is sometimes not explicative
**🧐 Motivation**
Using natspec, the `@dev` specification is sometimes not useful because it redirects to the interface or to other contracts. Users should explore the source code to see what the function does.

Like this (verified contract on etherscan):

Image

Image

**📝 Details**
It could be useful to use the `@inheritdoc` specification and then add a `@custom:something` specification to list additional details. The custom specification should also be used in docs generation.

i.e.

```solidity
    /**
     * @inheritdoc IERC20
     *
     * @custom:requirements
     * - `to` cannot be the zero address.
     * - the caller must have a balance of at least `value`.
     */
    function transfer(address to, uint256 value) public virtual returns (bool) {
        address owner = _msgSender();
        _transfer(owner, to, value);
        return true;
    }
```

vittominacori:
Fixes #5563 

I migrated simple Natspec comments to use `@inheritdoc`.

For instance:

```diff
/**
- * @dev See {IERC165-supportsInterface}.
+ * @inheritdoc IERC165
*/
function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
    // ...
}
```

I would like to get your feedback about more complex comments like:

```solidity
/**
 * @dev See {IERC20-transferFrom}.
 *
 * Skips emitting an {Approval} event indicating an allowance update. This is not
 * required by the ERC. See {xref-ERC20-_approve-address-address-uint256-bool-}[_approve].
 *
 * NOTE: Does not update the allowance if the current allowance
 * is the maximum `uint256`.
 *
 * Requirements:
 *
 * - `from` and `to` cannot be the zero address.
 * - `from` must have a balance of at least `value`.
 * - the caller must have allowance for ``from``'s tokens of at least
 * `value`.
 */
function transferFrom(address from, address to, uint256 value) public virtual returns (bool) {
    address spender = _msgSender();
    _spendAllowance(from, spender, value);
    _transfer(from, to, value);
    return true;
}
```

Do you suggest to use the `@notice` spec for custom comments:

```
/**
 * @inheritdoc IERC20
 * @notice Skips emitting an {Approval} event indicating an allowance update. This is not
 * required by the ERC. See {xref-ERC20-_approve-address-address-uint256-bool-}[_approve].
 *
 * NOTE: Does not update the allowance if the current allowance
 * is the maximum `uint256`.
 *
 * Requirements:
 *
 * - `from` and `to` cannot be the zero address.
 * - `from` must have a balance of at least `value`.
 * - the caller must have allowance for ``from``'s tokens of at least
 * `value`.
 */
```

Or do you suggest using `@custom` specification like:

```
/**
 * @inheritdoc IERC20
 * @notice Skips emitting an {Approval} event indicating an allowance update. This is not
 * required by the ERC. See {xref-ERC20-_approve-address-address-uint256-bool-}[_approve].
 *
 * @custom:note Does not update the allowance if the current allowance is the maximum `uint256`.
 *
 * @custom:requirements
 * - `from` and `to` cannot be the zero address.
 * - `from` must have a balance of at least `value`.
 * - the caller must have allowance for ``from``'s tokens of at least
 * `value`.
 */
```

### DIFF
--- a/contracts/governance/Governor.sol
+++ b/contracts/governance/Governor.sol
@@ -87,7 +87,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
     }
 
     /**
-     * @dev See {IERC165-supportsInterface}.
+     * @inheritdoc IERC165
      */
     function supportsInterface(bytes4 interfaceId) public view virtual override(IERC165, ERC165) returns (bool) {
         return
@@ -98,14 +98,14 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
     }
 
     /**
-     * @dev See {IGovernor-name}.
+     * @inheritdoc IGovernor
      */
     function name() public view virtual returns (string memory) {
         return _name;
     }
 
     /**
-     * @dev See {IGovernor-version}.
+     * @inheritdoc IGovernor
      */
     function version() public view virtual returns (string memory) {
         return "1";
@@ -134,7 +134,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
     }
 
     /**
-     * @dev See {IGovernor-getProposalId}.
+     * @inheritdoc IGovernor
      */
     function getProposalId(
         address[] memory targets,
@@ -146,7 +146,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
     }
 
     /**
-     * @dev See {IGovernor-state}.
+     * @inheritdoc IGovernor
      */
     function state(uint256 proposalId) public view virtual returns (ProposalState) {
         // We read the struct fields into the stack at once so Solidity emits a single SLOAD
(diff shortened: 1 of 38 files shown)


arr00:
Hey @vittominacori, thanks for taking the initiative here! You've discovered the main issue related to inheritdoc for us. Our docs generator only uses the `@dev` tag for docs generation. As such, if we need to have *more* than just the inheritdoc, we can't use it. I think in those cases we should keep it as is with the `see {x}`. 

One nit, we prefer to use single-line comments (`///`) when the comment is only one line.

vittominacori:
Ok @arr00, it's already as you said. When it's different from just the inherited documentation, I didn't make any changes.

I reverted one line comments to use `///` as it was. Anyway there are a lot of one line comments using the `/**/` syntax all around the code, but I think that should be another PR.

contracts/governance/Governor.sol
```diff
@@ -87,7 +87,7 @@ abstract contract Governor is Context, ERC165, EIP712, Nonces, IGovernor, IERC72
     }
 
     /**
-     * @dev See {IERC165-supportsInterface}.
+     * @inheritdoc IERC165
      */
```
arr00, line 91: Could we make all these single line comments?

```
/// @inheritdoc IERC165
```

ernestognw (commented):
Hey @vittominacori, thanks so much for working on this!
I like the idea and I really think we can get to the triple dashes `///` in this PR too. It's not a controversial change

vittominacori:
Hey @arr00, @ernestognw just updated one line inheritdoc to use ///.

vittominacori:
@ernestognw let me know if you would like to change ALL the one line comments with /// in the entire contracts folder.

It can be easily done in a regex.

arr00:
> @ernestognw let me know if you would like to change ALL the one line comments with /// in the entire contracts folder.
> 
> It can be easily done in a regex.

Let's limit the changes to inheritdoc related lines for now.

vittominacori:
> Let's limit the changes to inheritdoc related lines for now.

Great. Done. I will open a new PR once this will be merged to change one line comments to `///` in other cases.

arr00:
> > Let's limit the changes to inheritdoc related lines for now.
> 
> Great. Done. I will open a new PR once this will be merged to change one line comments to `///` in other cases.

I don't think it's necessary at the moment. We prefer single-line comments for new code, but we don't need to enforce it on existing code.

arr00 (approved):

ernestognw:
I agree with @arr00 that replacing all with triple dashes is another discussion. So far it works fine when they're single line comments, but it's true there's not a strict guideline set somewhere.

ernestognw (approved):
LGTM, will merge after https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5643 gets done too

PR #5642, opened 2025-04-21
Add Blockhash library following EIP-2935

### LINKED ISSUE #5639: Create a library to serve historical block hashes from state (EIP-2935)
**🧐 Motivation**
Before Pectra you can only query the last 256 block hashes. With [EIP-2539](https://eips.ethereum.org/EIPS/eip-2935) the EL will hold in a system contract the last 8192 block hashes(~27 hs) of blocks, this is very useful for rollups and other applications that need to query or proof something against older ethereum state.

**📝 Details**
Calling the new contract will be more expensive, so you will probably want to still use the precompile for newer blocks - a simple library that abstracts this away seems pretty useful.
Here's a very simple way to implement this. If we use this in our minimal-rollup I'm happy to contribute upstream
```solidity
/// @dev Reads a block hash up to 8 191 blocks ago after the Pectra fork.
///      Falls back to the BLOCKHASH opcode for <256 blocks.
library BlockhashHistory {
    address constant _HISTORY_ADDR = 0x0000F90827F1C53a10Cb7A02335B175320002935;
    uint256 constant _WINDOW = 8191;

    /// @notice Returns the hash of block `n` or reverts if too old / future.
    function hashOf(uint256 n) internal view returns (bytes32 h) {
        uint256 current = block.number;
        if (n >= current) revert("future block");
        unchecked {
            if (current - n <= 256) {
                return blockhash(n); // cheap path
            }
            if (current - n > _WINDOW) revert("too old");
        }
        bytes memory input = abi.encodePacked(bytes32(n));
        assembly {
            let success := staticcall(gas(), _HISTORY_ADDR, add(input, 32), 32, 0, 32)
            if iszero(success) { revert(0, 0) }
            h := mload(0)
        }
    }
}
```

ernestognw:
Fixes #5639

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ernestognw:
Tests are missing, perhaps there's a way to enable 2935 in either Foundry or Hardhat but requires some research before continue

contracts/utils/Blockhash.sol
```diff
@@ -0,0 +1,53 @@
...
+    /// @dev Internal function to query the EIP-2935 history storage contract.
+    function _historyStorageCall(uint256 blockNumber) private view returns (bytes32 hash) {
+        assembly ("memory-safe") {
+            // Use scratch space to allocate blockNumber
+            mstore(0, 0x20) // Store length and clear potentially dirty scratch space
+            mstore(0x20, blockNumber) // Store the blockNumber in scratch space
+
+            // In case the history storage address is not deployed, the call will succeed
+            // without returndata, so the hash will be 0 just as querying `blockhash` directly.
+            let success := staticcall(gas(), HISTORY_STORAGE_ADDRESS, 0x20, 0, 0, 0x20)
```
arr00, line 46:
```suggestion
            let success := staticcall(gas(), HISTORY_STORAGE_ADDRESS, 0x20, 0x20, 0, 0x20)
```
Shouldn't the input length be `0x20`?

contracts/utils/Blockhash.sol
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        return distance > 256 && distance <= 8191 ? _historyStorageCall(blockNumber) : blockhash(blockNumber);
+    }
+
+    /// @dev Internal function to query the EIP-2935 history storage contract.
+    function _historyStorageCall(uint256 blockNumber) private view returns (bytes32 hash) {
+        assembly ("memory-safe") {
+            // Use scratch space to allocate blockNumber
+            mstore(0, 0x20) // Store length and clear potentially dirty scratch space
```
arr00, line 41: If the contract isn't deployed and there is no return data, this won't be overwritten.

arr00:
Btw https://github.com/NomicFoundation/hardhat/issues/6226

ernestognw:
Yeah I'm also trying to test in Foundry and it doesn't really work 😕

foundry.toml
```diff
@@ -1,6 +1,6 @@
 [profile.default]
 solc_version = '0.8.24'
-evm_version = 'cancun'
+evm_version = 'prague'
```
ernestognw, line 3: Perhaps this is unnecessary since we're `vm.etch`'ing the bytecode anyway. I couldn't find anythign about EIP-2935 in the Foundry docs

contracts/utils/Blockhash.sol
```diff
@@ -0,0 +1,53 @@
...
+            // Use scratch space to allocate blockNumber
+            mstore(0, blockNumber) // Store the blockNumber in scratch space
+
+            // In case the history storage address is not deployed, the call will succeed
+            // without returndata, so the hash will be 0 just as querying `blockhash` directly.
+            let success := staticcall(gas(), HISTORY_STORAGE_ADDRESS, 0, 0x20, 0, 0x20)
+
+            // In case of failure, the returndata might include the revert reason or custom error
+            if success {
+                hash := mload(0)
```
arr00, line 49: If used on an unsupported chain this will return the block number? Why not just use a second zero-ed slot?

ernestognw: Do you mean using `mload(0x20)`? I think we'd need to clear it out too with `mstore(0x20,0)`

I think this would be a better implementation:

```
assembly ("memory-safe") {
    // Use scratch space to allocate blockNumber
    mstore(0, blockNumber) // Store the blockNumber in scratch space

    // In case the history storage address is not deployed, the call will succeed
    // without returndata, so the hash will be 0 just as querying `blockhash` directly.
    if and(staticcall(gas(), HISTORY_STORAGE_ADDRESS, 0, 0x20, 0, 0x20), returndatasize()) {
        hash := mload(0)
    }
}
```

wdyt?

arr00: Yes I this would work

arr00: ⛳️

arr00 (approved):

arr00 (approved):

ernestognw:
Thanks for leading this @arr00 !

contracts/utils/Blockhash.sol
boss1512boss-commits: 1c2fab1a005044df00fbe6b9907f050bc6b5327e

PR #5644, opened 2025-04-21
Add provenance to NPM publish during release cycle

### LINKED ISSUE #5629: Add provenance to `npm` registry
Currently, you don't publish any provenance statements on https://www.npmjs.com/package/@openzeppelin/contracts. I would recommend following the approach outlined here: https://docs.npmjs.com/generating-provenance-statements. This helps a lot verifying what you publish on `npm` and what is on GitHub.

I recently implemented it myself (using `pnpm` however) for my `npm` packages, and it worked smoothly, e.g. https://www.npmjs.com/package/xdeployer#provenance

Image

ernestognw:
Fixes #5629

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/scripts/release/workflow/publish.sh
+++ b/scripts/release/workflow/publish.sh
@@ -9,7 +9,7 @@ PACKAGE_JSON_VERSION="$(tar xfO "$TARBALL" package/package.json | jq -r .version
 echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc
 
 # Actual publish
-npm publish "$TARBALL" --tag "$TAG"
+npm publish --provenance "$TARBALL" --tag "$TAG"
 
 # Clean up tags
 delete_tag() {


frangio (approved):
Looks good. See: 
- https://docs.npmjs.com/generating-provenance-statements
- https://docs.npmjs.com/cli/v11/commands/npm-publish#provenance

pcaversaccio (commented):
Maybe I'm not fully understanding your release pipeline, but you're missing the `id-token: write` here I think:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3a14bcc78379c079ec783e577f1012719ad57fc3/.github/workflows/release-cycle.yml#L133

If you read the following [section](https://docs.npmjs.com/generating-provenance-statements#about-sigstore) carefully:

> The certificate authority federates with any OIDC provider that includes verifiable build information. It acts as an intermediary between build systems and package registries by verifying the integrity of the OIDC token, issues a signing certificate that contains that build information, and then logging the signing certificate to an immutable ledger.

The `id-token: write` permission (how I understand it) in GitHub Actions allows a job to request such an _OpenID Connect (OIDC)_ token from GitHub (see [here](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings)). Thus, you would need to give the permission to mint such an OIDC token in the CI.

Furthermore, since you use environment variables (via `env`) in the `release-cycle.yml` file already, wouldn't it be better to do it like this instead:

```yml
- name: Publish
  run: bash scripts/release/workflow/publish.sh
   env:
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       TARBALL: ${{ steps.pack.outputs.tarball }}
       TAG: ${{ steps.pack.outputs.tag }}
       NPM_CONFIG_PROVENANCE: true
```

Like this, it's tied directly to the environment you run the Bash script in.

frangio:
See also [here](https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry). There is no discussion of the need for `id-token: write` but it is present in the example workflow, so yeah it sounds like it's necessary. Thanks for the heads up @pcaversaccio!

Regarding where to enable provenance, one thing to consider is whether the script is ever used outside of CI (e.g., to deploy the upgradeable package?). In that case it sounds like putting it in an env var in the workflow is a better choice.

ernestognw:
Great catch @pcaversaccio, thanks! 

So if I'm understanding correctly, enabling `id-token: write` would allow Github to request an OIDC token by Github's issuer (`https://token.actions.githubusercontent.com`).

I'm particularly wary about enabling something on that release cycle job since it's probably the most critical, but in this case I guess it makes sense given Github is already a trusted entity in our supply chain

PR #5647, opened 2025-04-24
Show stack trace for failing foundry tests in CI

arr00:


ernestognw (approved):
I wonder if a longer output would slow down tests. Not a big deal anyway

arr00:
> I wonder if a longer output would slow down tests. Not a big deal anyway

Only longer output when failing. Worth it imo

PR #5649, opened 2025-04-25
Minimize Interface Pragmas

arr00:
The pragmas specified on interfaces are generally way more restrictive than necessary. Given that interfaces don't actually have any related bytecode, we just need to ensure that an interface can compile with lower solc versions--we should set the pragmas to be as permissive as possible to allow consumers to use the interfaces in as many ways as possible.


ernestognw (approved):
I understand we're not running it in the CI because it takes too long. I'm worried we may keep missing cases but not a big deal

PR #5658, opened 2025-05-02
Add EnumerableSetExtended and EnumerableMapExtended

ernestognw:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation


Amxx:
The reason for the name `...Extended` was to avoid conflicts if importing botht he vanila version, and the community extension. We would  not want both library to have the same name.

But since we are moving the code from community to here, we should put all that in the existing library and not create new ones.

.changeset/pink-dolls-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`EnumerableSetExtended` and `EnumerableMapExtended`: Extensions of the `EnumerableSet` and `EnumerableMap` libraries with more types, including non-value types.
```
Amxx:
```suggestion
`EnumerableSet` and `EnumerableMap`: Add support for more types, including non-value types.
```

scripts/generate/templates/Enumerable.opts.js
```diff
@@ -0,0 +1,55 @@
...
+]
+  .map(typeDescr)
+  .map(toSetTypeDescr);
+
+const MAP_TYPES = []
+  .concat(
+    // value type maps
+    ['uint256', 'address', 'bytes32']
+      .flatMap((keyType, _, array) => array.map(valueType => ({ key: { type: keyType }, value: { type: valueType } })))
+      .slice(0, -1), // remove bytes32 → byte32 (last one) that is already defined
```
Amxx:
```suggestion
      .slice(0, -1), // remove bytes32 → bytes32 (last one) that is already defined
```

Amxx:
Note: this is actually very close to #5558 that was closed when we decided to put these structs in the community repo.

scripts/generate/templates/Enumerable.opts.js
```diff
@@ -0,0 +1,55 @@
...
+  keySet: toSetTypeDescr(key),
+  key,
+  value,
+});
+
+const SET_TYPES = [
+  { type: 'bytes32' },
+  { type: 'address' },
+  { type: 'uint256' },
+  { type: 'bytes32', size: 2 },
```
Amxx: I'd remove at least this one
```suggestion
```

scripts/generate/templates/Enumerable.opts.js
```diff
@@ -0,0 +1,55 @@
...
+
+const MAP_TYPES = []
+  .concat(
+    // value type maps
+    ['uint256', 'address', 'bytes32']
+      .flatMap((keyType, _, array) => array.map(valueType => ({ key: { type: keyType }, value: { type: valueType } })))
+      .slice(0, -1), // remove bytes32 → bytes32 (last one) that is already defined
+    // non-value type maps
+    { key: { type: 'bytes', memory: true }, value: { type: 'uint256' } },
+    { key: { type: 'string', memory: true }, value: { type: 'string', memory: true } },
```
Amxx: maybe do bytes => bytes instead ... or have both ?
```suggestion
    { key: { type: 'bytes', memory: true }, value: { type: 'bytes', memory: true } },
```

scripts/generate/templates/Enumerable.opts.js
```diff
@@ -0,0 +1,55 @@
...
+  .map(toSetTypeDescr);
+
+const MAP_TYPES = []
+  .concat(
+    // value type maps
+    ['uint256', 'address', 'bytes32']
+      .flatMap((keyType, _, array) => array.map(valueType => ({ key: { type: keyType }, value: { type: valueType } })))
+      .slice(0, -1), // remove bytes32 → bytes32 (last one) that is already defined
+    // non-value type maps
+    { key: { type: 'bytes', memory: true }, value: { type: 'uint256' } },
```
Amxx: do we want to keep this one ?

scripts/generate/templates/EnumerableMap.js
```diff
@@ -40,6 +40,8 @@ import {EnumerableSet} from "./EnumerableSet.sol";
  * - \`address -> address\` (\`AddressToAddressMap\`) since v5.1.0
  * - \`address -> bytes32\` (\`AddressToBytes32Map\`) since v5.1.0
  * - \`bytes32 -> address\` (\`Bytes32ToAddressMap\`) since v5.1.0
+ * - \`bytes -> uint256\` (\`BytesToUintMap\`) since v5.4.0
+ * - \`string -> string\` (\`StringToStringMap\`) since v5.4.0
```
Amxx: eventually needs updating

scripts/generate/templates/EnumerableSet.js
```diff
@@ -29,8 +30,14 @@ import {Arrays} from "../Arrays.sol";
...
- * As of v3.3.0, sets of type \`bytes32\` (\`Bytes32Set\`), \`address\` (\`AddressSet\`)
- * and \`uint256\` (\`UintSet\`) are supported.
+ * The following types are supported:
+ *
+ * - \`bytes32\` (\`Bytes32Set\`) since v3.3.0
+ * - \`address\` (\`AddressSet\`) since v3.3.0
+ * - \`uint256\` (\`UintSet\`) since v3.3.0
+ * - \`string\` (\`StringSet\`) since v5.4.0
+ * - \`bytes\` (\`BytesSet\`) since v5.4.0
+ * - \`bytes32[2]\` (\`Bytes32x2Set\`) since v5.4.0
```
Amxx: TODO

scripts/generate/templates/Enumerable.opts.js
ernestognw: Yes, I also dont like this one, he

scripts/generate/templates/EnumerableSet.js
ernestognw: Remove? Done 😄

scripts/generate/templates/Enumerable.opts.js
ernestognw: Agree with this. Updating

scripts/generate/templates/Enumerable.opts.js
ernestognw: Nah, let's remove it and just keep `bytes` to `bytes` for now given is the most generic version.

We could theoretically do all variants for `['uint256', 'address', 'bytes32', 'bytes', 'string']`, but I think it needs more thought.

ernestognw: Actually, as you mentioned in #5558, Bytes -> Uint would be useful for weights:

> BytesToUintMap (usefull for something like an ERC-7093 contract for guardian->weight)

However, weights are just tracked with a mapping in the current SignerMultiERC7913Weighted and equivalent ERC7579MultisigWeighted module.

```solidity
// Mapping from signer to weight
mapping(bytes signer => uint256) private _weights;
```

Amxx (dismissed):

PR #5659, opened 2025-05-02
Feature/erc7913

ernestognw:
Follow up to #5657

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/cryptography/ERC7913Utils.sol
```diff
@@ -0,0 +1,82 @@
...
+     */
+    function areValidSignaturesNow(
+        bytes32 hash,
+        bytes[] memory signers,
+        bytes[] memory signatures
+    ) internal view returns (bool) {
+        bytes32 previousId = bytes32(0);
+
+        uint256 signersLength = signers.length;
+        for (uint256 i = 0; i < signersLength; i++) {
```
Amxx: Compiler knows how to optimize that now.
```suggestion
        for (uint256 i = 0; i < signers.length; ++i) {
```

contracts/utils/cryptography/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,246 @@
...
+ * (if used standalone) or during initialization (if used as a clone) may leave the contract
+ * either front-runnable or unusable.
+ */
+abstract contract MultiSignerERC7913 is AbstractSigner {
+    using EnumerableSetExtended for EnumerableSetExtended.BytesSet;
+    using ERC7913Utils for *;
+    using SafeCast for uint256;
+
+    EnumerableSetExtended.BytesSet private _signersSet;
+    uint128 private _threshold;
```
Amxx: why uint128 and not uint256 ?

contracts/utils/cryptography/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,246 @@
...
+
+    /// @dev Returns the minimum number of signers required to approve a multisignature operation.
+    function threshold() public view virtual returns (uint256) {
+        return _threshold;
+    }
+
+    /// @dev Returns the set of authorized signers.
+    function _signers() internal view virtual returns (EnumerableSetExtended.BytesSet storage) {
+        return _signersSet;
+    }
```
Amxx: Why use an internal function and not directly use the storage in the other functions. This goes against our policy of having private storage ! using that internal function someone could add/remove signers without doing all the necessary checks, and without emitting events.

contracts/utils/cryptography/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,246 @@
...
+     */
+    function _removeSigners(bytes[] memory oldSigners) internal virtual {
+        uint256 oldSignersLength = oldSigners.length;
+        for (uint256 i = 0; i < oldSignersLength; i++) {
+            bytes memory signer = oldSigners[i];
+            require(_signers().remove(signer), MultiSignerERC7913NonexistentSigner(signer));
+        }
+        _validateReachableThreshold();
+        emit ERC7913SignersRemoved(oldSigners);
+    }
```
Amxx: what is the rational for having internal functions that add in batch, vs function that add one and can be called multiple time (like we have in AccessControl).

For the `_removeSigners` I see it avoids computing `_validateReachableThreshold` multiple times, but I'd still go for the simple version that doesn't deal with arrays.

contracts/utils/cryptography/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,246 @@
...
+     * * The `signature` must be encoded as `abi.encode(signers, signatures)`.
+     */
+    function _rawSignatureValidation(
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
+        if (signature.length == 0) return false; // For ERC-7739 compatibility
+        (bytes[] memory signingSigners, bytes[] memory signatures) = abi.decode(signature, (bytes[], bytes[]));
+        if (signingSigners.length != signatures.length) return false;
+        return _validateThreshold(signingSigners) && _validateSignatures(hash, signingSigners, signatures);
```
Amxx:
```suggestion
        return 
            signingSigners.length == signatures.length && 
            _validateThreshold(signingSigners) && 
            _validateSignatures(hash, signingSigners, signatures);
```

contracts/utils/cryptography/ERC7913Utils.sol
```diff
@@ -0,0 +1,82 @@
...
+     * @dev Verifies multiple `signatures` for a given hash using a set of `signers`.
+     *
+     * The signers must be ordered by their `keccak256` hash to ensure no duplicates and to optimize
+     * the verification process. The function will return `false` if the signers are not properly ordered.
+     *
+     * Requirements:
+     *
+     * * The `signatures` array must be at least the  `signers` array's length.
+     */
+    function areValidSignaturesNow(
```
Amxx: I'm not a huge fan of having this function here. IMO it should be directly in the MultiSigner (in `_validateSignatures`). The ordering is not really part of the standard (you could decide to order differently then by hash).

We don't have an equivalent (batch) function in `SignatureChecker`, and I don't think we should have this one here.

contracts/utils/cryptography/MultiSignerERC7913Weighted.sol
```diff
@@ -0,0 +1,196 @@
...
+ * IMPORTANT: When setting a threshold value, ensure it matches the scale used for signer weights.
+ * For example, if signers have weights like 1, 2, or 3, then a threshold of 4 would require at
+ * least two signers (e.g., one with weight 1 and one with weight 3). See {signerWeight}.
+ */
+abstract contract MultiSignerERC7913Weighted is MultiSignerERC7913 {
+    using EnumerableSetExtended for EnumerableSetExtended.BytesSet;
+    using SafeCast for uint256;
+
+    // Invariant: sum(weights) >= threshold
+    uint128 private _totalWeight;
```
Amxx: same, why uint128 and not uint256 ?

contracts/utils/cryptography/MultiSignerERC7913Weighted.sol
```diff
@@ -0,0 +1,196 @@
...
+        return _totalWeight; // Doesn't need Math.max because it's incremented by the default 1 in `_addSigners`
+    }
+
+    /**
+     * @dev Gets the weight of the current signer. Returns 1 if not explicitly set.
+     *
+     * NOTE: This internal function doesn't check if the signer is authorized.
+     */
+    function _signerWeight(bytes memory signer) internal view virtual returns (uint256) {
+        return Math.max(_weights[signer], 1);
```
Amxx: This feels edgy to me. 

If a signer is present (`isSigner` returns true), but the `_weights` for it is not set, then the max here would give him weight that `_totalWeight` doesn't account. This cannot happen without override, because `_setSignerWeights` checks that new weights are not 0 ... but an override could easily break it.

Amxx: could also happens if someone calls `_signers().add(...)` directly instead of calling `_addSigners`

contracts/utils/cryptography/MultiSignerERC7913.sol
ernestognw: `threshold` and `totalWeight` are accessed together pretty frequently so it's cheaper if they're packed on the same slot. I think it's reasonable to allow for `uint128` as a maximum for these 2

contracts/utils/cryptography/MultiSignerERC7913.sol
ernestognw: If there's not an internal function, users don't have a way to override the `_signersSet` cleanly. So the purpose for this function was to have a single `_signersSet` access. I don't mind removing it, but there are definitely worse things exposed through internal functions.

contracts/utils/cryptography/MultiSignerERC7913.sol
ernestognw: The rationale is that batching is a subset of a single addition removal, it provides the same functionality to call `_addSigners` or `_removeSigners` with 1-element array.

It's true they can batch them, but batching removes readability. I find clarity and readability there pretty important.

I would agree with changing the event, though

contracts/utils/cryptography/MultiSignerERC7913Weighted.sol
ernestognw: Same as the other, packing.

contracts/utils/cryptography/MultiSignerERC7913Weighted.sol
ernestognw: The alternative is a ternary operator here, but I thought it was too much to check if the signer is a member since it reads storage. For the majority of cases I think this is fine and the risk is only user's input, noting should be fine imo

contracts/utils/cryptography/ERC7913Utils.sol
ernestognw: imo it doesn't need to be standard to be a _utility function_

I would feel comfortable about recommending everyone to follow this ordering approach if they're starting from scratch, so the function here does make sense.

> We don't have an equivalent (batch) function in SignatureChecker, and I don't think we should have this one here.

Is the reason only that SignatureChecker doesn't haven an equivalent? I would argue that's on purpose given SignatureChecker didn't generalized well for arbitrary signers. We created 7913 for that, so let's use it!

Amxx: SignatureChecker.isValidSignatureNow generalized to many ECDSA/1271 signers just as well as the 7913 version does.

ernestognw: I proposed these versions a while ago:

- https://github.com/ernestognw/openzeppelin-contracts/blob/f791820e8806912e08214a0266da0314175b5476/contracts/account/modules/TypedERC1271Executor.sol
- https://github.com/ernestognw/openzeppelin-contracts/blob/f791820e8806912e08214a0266da0314175b5476/contracts/account/modules/MultiERC1271Validator.sol

None of them was even taken into consideration because they didn't generalize well in the first place. We also had all that discussion about whether a key has an identity or not. Honestly this just feels like there are always reasons why the work is not properly done, and no matter how hard I try, it's wrong

To me, it's pretty clear now that 1271 didn't generalize well since the beginning. If any, I tried to make the counterpoint that it was enough for generalizing and we still pursued 7913. My opinion is pretty strong on this one

contracts/utils/cryptography/MultiSignerERC7913.sol
Amxx: Interresting. I thought that storage between contract would not pack, just like structs don't pack (they force a new slot being used for what comes after).

I tried, and indeed, storage does pack between contracts !

That saving won't happen on the upgradeable version because it will be different namespaces, but it does save storage for the vanilla version.

Then my question becomes:
- why not uint64 (in case we ever want to add some more things after these two values)
- why not use a uint128 input argument to the `_setThreshold` to avoid the casting ?

contracts/utils/cryptography/MultiSignerERC7913.sol
Amxx: THis is breaking what is possibly the biggest rule in our entire codebase: all storage is private.

I think having direct access to it it a very big footgun. I feel very strongly that we should remove this access pattern.

contracts/utils/cryptography/MultiSignerERC7913.sol
ernestognw: > why not uint64 (in case we ever want to add some more things after these two values)

We can do that, yes.

> why not use a uint128 input argument to the _setThreshold to avoid the casting ?

I think the interface is different to the actual implementation, so the rationale is that the interface shouldn't reflect specific implementation details like using `uint128` or `uint64`. I would agree with changing the interface anyway, just thought it was an antipattern

contracts/utils/cryptography/MultiSignerERC7913.sol
Amxx: actually, I would even remove the length check here and move it to `_validateSignatures`

contracts/utils/cryptography/MultiSignerERC7913.sol
ernestognw: I don't see how a "footgun" like this is different to an internal "_grantRole". Both are comparable in terms of consequences.

I would remove it. But it's another example in which I have a strong clear reason to do something and it's rejected without a proper counterargument other than "this is how we've been doing it"

contracts/utils/cryptography/MultiSignerERC7913Weighted.sol
Amxx: I'm not sure how a ternary operator would change anything to the fact that there can be an inconsistency between totalWeight() and the sum of the signerWeight.

I was thinking about using an "extraWeight" in a mapping:
- totalWeight() becomes `_totalExtraWeight + signerCount()`
- _signerWeight() becomes `1 + _extraWeight[signer]`
- when setting a signer weight, what is stored is `newWeight - 1` (we require that `newWeight > 0` so that is fine).
...

if someone overides `isSigner` they would also have to override `signerCount()`, which is probably similar to overriding `totalWeight()` ... so I'm not sure.

The upside is that storing 0 is cheaper than storing 1, so if weights are set to 1 its cheaper ... but getting the signerCount() is also a read ...

ernestognw: > I'm not sure how a ternary operator would change anything to the fact that there can be an inconsistency between totalWeight() and the sum of the signerWeight.

I was thinking of (which only counts the weight if it's an authorized signer)

```solidity
function _signerWeight(bytes memory signer) internal view virtual returns (uint256) {
  return Math.ternay(isSigner(signer), Math.max(_weights[signer], 1), 0);
}
```

I don't see any other way in which an inconsistency can be created...

> I was thinking about using an "extraWeight" in a mapping:

And use unnecessary storage? It's your call but honestly this is just unnecessary bloating. A note would suffice imo

contracts/utils/cryptography/ERC7913Utils.sol
Amxx, line 1: IMO the functions in that file should go in SignatureChecker.

ernestognw: Agree with this one. Will update in community and update accordingly over here

contracts/utils/cryptography/MultiSignerERC7913Weighted.sol
Amxx: >And use unnecessary storage?

The extraWeight mapping would replace the weight mapping. It would not be storing anothing new, just the same thing with an offset. 

I'll need to think about it tomorrow. To late for this today.

contracts/utils/cryptography/MultiSignerERC7913.sol
ernestognw: We removed it in community and I see it's the check is not necessary. Solving

ernestognw:
We may need to check #5716 first

contracts/mocks/utils/cryptography/ERC7913VerifierMock.sol
Amxx, line 1: I'm wondering if this mock is really necessary. The P256 verifier does the job just fine, and we have all the tooling to use it easily in tests.

contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -47,4 +57,72 @@ library SignatureChecker {
...
+     * the verification process. The function will return `false` if the signers are not properly ordered.
+     *
+     * Requirements:
+     *
+     * * The `signatures` array must be at least the  `signers` array's length.
+     *
+     * NOTE: Unlike ECDSA signatures, contract signatures are revocable, and the outcome of this function can thus
+     * change through time. It could return true at block N and false at block N+1 (or the opposite).
+     */
+    function areValidERC7913SignaturesNow(
```
Amxx, line 105: fragio proposed to add the uniqueness property to the name of the function. I personnaly like the idea

```suggestion
    function areValidUniqueERC7913SignaturesNow(
```

contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -47,4 +57,72 @@ library SignatureChecker {
...
+        bytes32 hash,
+        bytes[] memory signers,
+        bytes[] memory signatures
+    ) internal view returns (bool) {
+        bytes32 previousId = bytes32(0);
```
Amxx: I think we should return false instead of panicking.
```suggestion
        if (signers.length != signatures.length) return false;

        bytes32 previousId = bytes32(0);
```

test/helpers/signers.js
```diff
@@ -144,4 +147,62 @@ class RSASHA256SigningKey extends RSASigningKey {
...
+    if (weights !== null) {
+      assertArgument(
+        Array.isArray(weights) && weights.length === signers.length,
+        'weights must be an array with the same length as signers',
+        'weights',
+        weights.length,
+      );
+    }
+
+    this.#signers = signers;
```
Amxx, line 171: We could do the sorting once at construction, and then we wouldn't have to put it in `sign()`

```suggestion
    this.#signers = signers.sort((s1, s2) => toBigInt(keccak256(signer.bytes ?? signer.address)) < toBigInt(keccak256(signer.bytes ?? signer.address)) ? -1 : 1);
```

contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -47,4 +57,80 @@ library SignatureChecker {
...
+     *
+     * NOTE: Unlike ECDSA signatures, contract signatures are revocable, and the outcome of this function can thus
+     * change through time. It could return true at block N and false at block N+1 (or the opposite).
+     */
+    function areValidERC7913SignaturesNow(
+        bytes32 hash,
+        bytes[] memory signers,
+        bytes[] memory signatures
+    ) internal view returns (bool) {
+        bytes32 lastId = bytes32(0);
```
Amxx, line 112:
```suggestion
        if (signers.length != signatures.length) return false;

        bytes32 lastId = bytes32(0);
```

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,238 @@
...
+     *
+     * Requirements:
+     *
+     * * The `signature` must be encoded as `abi.encode(signers, signatures)`.
+     */
+    function _rawSignatureValidation(
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
+        if (signature.length == 0) return false; // For ERC-7739 compatibility
```
Amxx, line 202: if the signature length is 0, then the abi.decode would panic.

I would support this check if it made the function clean is in "it will return false instead of panicking" ... but if the signature length is 1 or 2 ... or if the signature is not decodable, then the system will still panic.

So I'm wondering how much this check brings. 

Comment says its for 7739 compatibility, but I don't understand why that is. Is 7739 going to bring empty signatures, and do we need to not panic specifically for these ones ?

ernestognw: >  Is 7739 going to bring empty signatures, and do we need to not panic specifically for these ones ?

Yes! Is a bit of a forced bypass but I couldn't find an alternative

ernestognw: Similar to this [situation](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/117#discussion_r2127365326)

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,236 @@
...
+    /**
+     * @dev Returns a slice of the set of authorized signers.
+     *
+     * Using `start = 0` and `end = type(uint64).max` will return the entire set of signers.
+     *
+     * WARNING: Depending on the `start` and `end`, this operation can copy a large amount of data to memory, which
+     * can be expensive. This is designed for view accessors queried without gas fees. Using it in state-changing
+     * functions may become uncallable if the slice grows too large.
+     */
+    function getSigners(uint64 start, uint64 end) public view virtual returns (bytes[] memory) {
```
ernestognw, line 83: We need to expose the total amount of signers somehow. Currently, it's not possible to know the proper `end` argument for this function onchain.

ernestognw: Add a `totalSigners` function. Solving

contracts/utils/cryptography/SignatureChecker.sol
```diff
@@ -47,4 +57,79 @@ library SignatureChecker {
...
+     * Verification is done as follows:
+     *
+     * * If `signer.length < 20`: verification fails
+     * * If `signer.length == 20`: verification is done using {isValidSignatureNow}
+     * * Otherwise: verification is done using {IERC7913SignatureVerifier}
+     *
+     * NOTE: Unlike ECDSA signatures, contract signatures are revocable, and the outcome of this function can thus
+     * change through time. It could return true at block N and false at block N+1 (or the opposite).
+     */
+    function isValidERC7913SignatureNow(
```
Amxx, line 76: I'm thinking that having 
- `isValidSignatureNow(address,bytes32,bytes)` 
- `isValidERC7913SignatureNow(bytes,bytes32,bytes)` 
might imply that the "default" version is the first one, and might push away from the second as an 7913 specific thing ... which it is ... but people not knowing 7913 might think it doesn't apply to them.

What if we named both of them the same, and relied on the input type to distinguish
- `isValidSignatureNow(address,bytes32,bytes)` 
- `isValidSignatureNow(bytes,bytes32,bytes)` 

The logic is to emphasize that they serve the same purpose, and that the second version is a generalisation of the first one.

Approving this and opening another issue/pr to discuss that

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,241 @@
...
+import {SignatureChecker} from "../SignatureChecker.sol";
+import {EnumerableSet} from "../../structs/EnumerableSet.sol";
+
+/**
+ * @dev Implementation of {AbstractSigner} using multiple ERC-7913 signers with a threshold-based
+ * signature verification system.
+ *
+ * This contract allows managing a set of authorized signers and requires a minimum number of
+ * signatures (threshold) to approve operations. It uses ERC-7913 formatted signers, which
+ * concatenate a verifier address and a key: `verifier || key`.
```
```suggestion
 * signatures (threshold) to approve operations. It uses ERC-7913 formatted signers, which
 * makes it nativelly compatible with ECDSA and ERC-1271 signers.
```
```

File: contracts/utils/cryptography/signers/MultiSignerERC7913.sol (hunk 14 above)
Comment (Amxx) [reply contracts/utils/cryptography/signers/MultiSignerERC7913.sol:83]: It is possible.

---

**Lets imagine you have 7 signers [a,b,c,d,e,f,g] that you are trying to get in batches of 3:**

`getSigners(0,3) => [a,b,c]`

You asked a page of size 3 and you got 3 elements ... so it may not be done. So you do

`getSigners(3,6) => [d,e,f]`

You asked a page of size 3 and you got 3 elements ... so it may not be done. So you do

`getSigners(6,9) => [g]`

You asked a page of size 3 and only got 1 element, so you know your at the end.

---

**Now lets imagine you have 6 signers [a,b,c,d,e,f] that you are trying to get in batches of 3:**
 
`getSigners(0,3) => [a,b,c]`

You asked a page of size 3 and you got 3 elements ... so it may not be done. So you do

`getSigners(3,6) => [d,e,f]`

You asked a page of size 3 and you got 3 elements ... so it may not be done. So you do

`getSigners(6,9) => []`

You asked a page of size 3 and  got an got 0 element, so you know your at the end.

File: contracts/utils/cryptography/signers/MultiSignerERC7913.sol [hunk 17]
```diff
@@ -0,0 +1,241 @@
...
+    error MultiSignerERC7913InvalidSigner(bytes signer);
+
+    /// @dev The `threshold` is unreachable given the number of `signers`.
+    error MultiSignerERC7913UnreachableThreshold(uint64 signers, uint64 threshold);
+
+    /// @dev Returns the total number of signers.
+    function totalSigners() public view virtual returns (uint64) {
+        return uint64(_signers.length()); // Safe cast. Extremely costly to overflow.
+    }
+
```
Amxx: See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5659/files#r2128151953

```suggestion
```

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
Amxx: That is how paged request are usually handled when doing graphql requests.

Amxx: also, if you do `getSigners(0, type(uint256).max)` you get everything all at once (just like before). So if we didn't need a `totalSigners()` before, I don't see why we need it now.

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,241 @@
...
+        return _validateThreshold(signers) && _validateSignatures(hash, signers, signatures);
+    }
+
+    /**
+     * @dev Validates the signatures using the signers and their corresponding signatures.
+     * Returns whether whether the signers are authorized and the signatures are valid for the given hash.
+     *
+     * IMPORTANT: For simplicity, this contract assumes that the signers are ordered by their `keccak256` hash
+     * to avoid duplication when iterating through the signers (i.e. `keccak256(signer1) < keccak256(signer2)`).
+     * The function will return false if the signers are not ordered.
```
```suggestion
     * IMPORTANT: Sorting the signers by their `keccak256` hash will improve the gas efficiency of this function.
     * See {SignatureChecker-areValidERC7913SignaturesNow} for more details.
```

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -97,6 +97,80 @@ function _verify(
...
+function _verifyMultipleSignatures(
+    bytes32 hash,
+    bytes[] memory signers,
+    bytes[] memory signatures
+) internal view returns (bool) {
+    return SignatureChecker.areValidERC7913SignaturesNow(hash, signers, signatures);
+}
+----
+
+The signers must be ordered by their `keccak256` hash to ensure no duplicates and optimize verification.
```
Amxx:
```suggestion
This function will reject inputs that contain duplicated signers. Sorting the signers by their `keccak256` hash is recommanded to minimize the gas cost.
```

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,241 @@
...
+import {SignatureChecker} from "../SignatureChecker.sol";
+import {EnumerableSet} from "../../structs/EnumerableSet.sol";
+
+/**
+ * @dev Implementation of {AbstractSigner} using multiple ERC-7913 signers with a threshold-based
+ * signature verification system.
+ *
+ * This contract allows managing a set of authorized signers and requires a minimum number of
+ * signatures (threshold) to approve operations. It uses ERC-7913 formatted signers, which
+ * makes it nativelly compatible with ECDSA and ERC-1271 signers.
```
```suggestion
 * makes it natively compatible with ECDSA and ERC-1271 signers.
```

contracts/utils/cryptography/SignatureChecker.sol
ernestognw: Yeah this is a good point. So my initial proposal was to have `isValidSignatureNow(bytes,bytes32,bytes)` and differentiate by the signature, as you mention

Amxx (dismissed):

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -0,0 +1,235 @@
...
+
+    /**
+     * @dev Validates the current threshold is reachable.
+     *
+     * Requirements:
+     *
+     * * The {signers}'s length must be `>=` to the {threshold}. Throws {MultiSignerERC7913UnreachableThreshold} if not.
+     */
+    function _validateReachableThreshold() internal view virtual {
+        uint64 signersLength = uint64(_signers.length()); // Safe cast. Economically impossible to overflow.
```
```suggestion
        uint256 signersLength = _signers.length();
```

Amxx (approved):

PR #5660, opened 2025-05-02
Docs/account

ernestognw:
Follow up to #5659

#### PR Checklist

- [ ] Tests
- [ ] Documentation


.changeset/proud-tables-sip.md
arr00, line 1: already added

.changeset/lazy-poets-cheer.md
arr00, line 1: Already added in #5664

contracts/utils/cryptography/AbstractSigner.sol
arr00, line 1: Already added in #5664

.changeset/tame-bears-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`AbstractSigner`, `SignerECDSA`, `SignerP256`, and `SignerRSA`: Add an abstract contract and various implementations for contracts that deal with signature verification.
```
arr00, line 5: AbstractSigner already added in #5664

ernestognw: Oh yes I need to rebase this branch

ernestognw: Oh but this one doesn't have a changeset, so I think it's cool adding it here

ernestognw:
It seems that transpilation is [skipping files that use an Initializer](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/scripts/upgradeable/transpile.sh#L33). This makes MyAccountERC7913, MyAccountMultiSigner and MyAccountMultiSigner weighted to not replace the `Account` import with `@openzeppelin/contracts/account/Account.sol`. We may need to update the transpiler too

Amxx:
[This line](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/scripts/upgradeable/transpile.sh#L33) result in:
- The Initializable contract not being transpiled.
- The Initializable contract being kept (as is) in the upgradeable version.
- All transpiled contract get to import it, and use it for their initializer

I'm not exacty sure how contract that already include an Initializable import are affected, but on of my guess would be that transpiling them would result in using two `Initializable` in two different locations.

IMO we should consider:
- Not copying the `Initializable` file to the upgradeable repo (breaking change)
- Use the `Initializable` contract from the vanilla repo in the upgradeable contracts
- For contracts that are already using initializable, assume they are already using an upgrade pattern (that is the case of the mocks you mention) and then do not transpile them at all.

Amxx:
Alternativelly (I think its more simple) we probably don't need an upgradeable version of the `contracts/mocks/docs` contracts. Are they part of the upgradeable docs ?

contracts/mocks/docs/account/MyAccountERC7702.sol
```diff
@@ -0,0 +1,20 @@
+// contracts/MyAccountERC7702.sol
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {Account} from "../../../account/Account.sol";
+import {ERC721Holder} from "../../../token/ERC721/utils/ERC721Holder.sol";
+import {ERC1155Holder} from "../../../token/ERC1155/utils/ERC1155Holder.sol";
+import {ERC7821} from "../../../account/extensions/ERC7821.sol";
```
Amxx: @ernestognw this is the issue here. We renamed that file
```suggestion
import {ERC7821} from "../../../account/extensions/draft-ERC7821.sol";
```

Amxx: (many other occurence of the same thing)

docs/modules/ROOT/pages/multisig.adoc
```diff
@@ -29,7 +29,36 @@ The xref:api:utils.adoc#SignerERC7913[`SignerERC7913`] contract allows a single
...
+import {Account} from "@openzeppelin/community-contracts/account/Account.sol";
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/EIP712.sol";
+import {ERC721Holder} from "@openzeppelin/contracts/token/ERC721/utils/ERC721Holder.sol";
+import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
+import {ERC7739} from "@openzeppelin/community-contracts/utils/cryptography/signers/ERC7739.sol";
+import {ERC7821} from "@openzeppelin/community-contracts/account/extensions/ERC7821.sol";
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import {SignerERC7913} from "@openzeppelin/community-contracts/utils/cryptography/signers/SignerERC7913.sol";
+
+contract MyAccountERC7913 is Account, SignerERC7913, ERC7739, ERC7821, ERC721Holder, ERC1155Holder, Initializable {
```
Amxx, line 46: maybe this should have a 
```solidity
function updateSigner(bytes memory signer_) public onlyEntryPointOrSelf() {
    _setSigner(signer_);
}
```

ernestognw:
`MyAccountERC7913` imports `Account`, so, if we're skipping the transpilation of `Initializer`, then the `MyAccountERC7913` won't be transpiled either because it imports Initializer. The problem lies in that `Account.sol` is stateless and removed from the repository after transpilation, so the `MyAccountERC7913` must import `Account` from `@openzeppelin/contracts` first.

In short: skipping transpilation of `Initializer` not only skips transpiling that contract, but also skips any contract that uses it.

docs/modules/ROOT/pages/multisig.adoc
ernestognw: Yes! Good catch, thanks

Amxx:
> In short: skipping transpilation of Initializer not only skips transpiling that contract, but also skips any contract that uses it.

OK, but `MyAccountERC7913` should be skipped anyway, since it is in contracts/mocks/docs/**

I don't understand how skipping a contract causes an issue with that contract's import.

I guess we should merge this, and then we can try to play arround and extract code from the docs in another PR.

contracts/mocks/docs/account/MyFactoryAccount.sol
```diff
@@ -0,0 +1,47 @@
...
+
+    /// @dev Create clone accounts on demand and return the address. Uses `callData` to initialize the clone.
+    function _cloneAndInitialize(bytes32 salt, bytes calldata callData) internal returns (address) {
+        (address predicted, bytes32 _calldataSalt) = predictAddress(salt, callData);
+        if (predicted.code.length == 0) {
+            _impl.cloneDeterministic(_calldataSalt);
+            predicted.functionCall(callData);
+        }
+        return predicted;
+    }
```
Amxx: For docs, I'd keep the code simple. Having two function here doesn't feel necessary
```suggestion
    function cloneAndInitialize(bytes32 salt, bytes calldata callData) public returns (address) {
        (address predicted, bytes32 _calldataSalt) = predictAddress(salt, callData);
        if (predicted.code.length == 0) {
            _impl.cloneDeterministic(_calldataSalt);
            predicted.functionCall(callData);
        }
        return predicted;
    }
```

contracts/utils/cryptography/signers/SignerERC7913.sol
```diff
@@ -21,6 +21,10 @@ import {SignatureChecker} from "../SignatureChecker.sol";
  *     function initialize(bytes memory signer_) public initializer {
  *       _setSigner(signer_);
  *     }
+ *
+ *     function setSigner(bytes memory signer_) public {
```
Amxx: this function needs protection
```suggestion
 *     function setSigner(bytes memory signer_) public onlyEntryPointOrSelf() {
```

ernestognw:
> I don't understand how skipping a contract causes an issue with that contract's import.

Skipping MyAccountERC7913 avoids replacing:

```diff
+import {Account} from "@openzeppelin/contracts/account/Account.sol";
-import {Account} from "../../.../account/Account.sol";
```

contracts/mocks/docs/account/MyFactoryAccount.sol
```diff
@@ -0,0 +1,47 @@
...
+    using Address for address;
+
+    address private immutable _impl;
+
+    constructor(address impl_) {
+        _impl = impl_;
+    }
+
+    /// @dev Predict the address of the account
+    function predictAddress(bytes32 salt, bytes calldata callData) public view returns (address, bytes32) {
```
Amxx: I find it strange (and possibly confusing) to return the bytes calldataSalt in addition to the address.

IMO this function should return only the address. In `cloneAndInitialize`, if `predicted.code.length == 0` we can recompute the hash, that is not a big deal IMO.

Amxx:
> > I don't understand how skipping a contract causes an issue with that contract's import.
> 
> Skipping MyAccountERC7913 avoids replacing:
> 
> ```diff
> +import {Account} from "@openzeppelin/contracts/account/Account.sol";
> -import {Account} from "../../.../account/Account.sol";
> ```

What I don't understand is that I would expect skipped contract to just not cause anything, no transpiled code generated, nothing to compile, no possible error. I was under the impression that skipping is a way around the issue, not that skipping would cause issues.

docs/modules/ROOT/pages/account-abstraction.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+This process of bundling user operations involves several costs that the bundler must cover, including base transaction fees, calldata serialization, entrypoint execution, and paymaster context costs. To compensate for these expenses, bundlers use the `preVerificationGas` and `gasFees` fields to charge users appropriately.
+
+NOTE: Estimating `preVerificationGas` is not standardized as it varies based on network conditions such as gas prices and the size of the operation bundle.
+
+TIP: Use xref:api:account.adoc#ERC4337Utils[`ERC4337Utils`] to manipulate the `UserOperation` struct and other ERC-4337 related values.
+
+=== Entrypoint
+
+Each `UserOperation` is executed through a contract known as the https://etherscan.io/address/0x0000000071727de22e5e9d8baf0edac6f37da032#code[`EntryPoint`]. This contract is a singleton deployed across multiple networks at the same address although other custom implementations may be used.
```
Amxx: update to entrypoint v08
```suggestion
Each `UserOperation` is executed through a contract known as the https://etherscan.io/address/0x4337084D9E255Ff0702461CF8895CE9E3b5Ff108#code[`EntryPoint`]. This contract is a singleton deployed across multiple networks at the same address although other custom implementations may be used.
```

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -0,0 +1,307 @@
...
+The library implementation supports a single batch mode (`0x01000000000000000000`) and allows accounts to execute multiple calls atomically. The standard includes access control through the xref:api:account.adoc#ERC7821-_erc7821AuthorizedExecutor-address-bytes32-bytes-[`_erc7821AuthorizedExecutor`] function, which by default only allows the contract itself to execute batches.
+
+Here's an example of how to use batched execution:
+
+[source,solidity]
+----
+import {Account} from "@openzeppelin/community-contracts/account/Account.sol";
+import {ERC7821} from "@openzeppelin/community-contracts/account/extensions/draft-ERC7821.sol";
+
+contract MyAccount is Account, ERC7821 {
```
Amxx: This is missing a signer, do we want to add Signer7702 or something similar ?

Amxx: I'm proposing 7702, because the design of 7821, that accepts calls where msg.sender == address(this) by default, is really 7702 oriented IMO

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+=== Signing the UserOp
+
+To sign a UserOperation, you'll need to first calculate its hash using the EntryPoint's `getUserOpHash` function, then sign this hash using your account's signature scheme, and finally encode the resulting signature in the format that your account contract expects for verification.
+
+[source,typescript]
+----
+const userOpHash = await entrypoint.read.getUserOpHash([userOp]);
+userOp.signature = await eoa.sign({ hash: userOpHash });
+----
```
Amxx: Before v08, there was no real guarantee of what hash the account would use. It would be the entrypoint's hash, but could also be something different (remember AccountCore vs Account, with the 712 rehashing).

Since v08, the hash that the entrypoint builds is a 712 hash. There is no strong reason (AFAIK) to not use it. So I would argue that you should NOT do `getUserOpHash` and then do a raw sign of it, but instead, I'd document signing the userOp as a 712 structure (using the entrypoint's domain)

docs/modules/ROOT/pages/multisig.adoc
```diff
@@ -0,0 +1,306 @@
...
+}
+----
+
+This implementation is ideal for standard multisig setups where each signer has equal authority, and a fixed number of approvals is required.
+
+The `MultiSignerERC7913` contract provides several key features for managing multi-signature accounts. It maintains a set of authorized signers and implements a threshold-based system that requires a minimum number of signatures to approve operations. The contract includes an internal interface for managing signers, allowing for the addition and removal of authorized parties.
+
+NOTE: `MultiSignerERC7913` safeguards to ensure that the threshold remains achievable based on the current number of active signers, preventing situations where operations could become impossible to execute.
+
+The contract also provides the public `isSigner(bytes memory signer)` function to check if a given signer is authorized, which is useful when validating signatures or implementing customized access control logic.
```
Amxx: should we mention `getSigners(uint256 start, uint256 end)` that can be used to fetch a (paginated) list of signers ?

docs/modules/ROOT/pages/multisig.adoc
```diff
@@ -0,0 +1,306 @@
...
+----
+
+Where:
+
+* `signers` is an array of the signers participating in this particular signature
+* `signatures` is an array of the individual signatures corresponding to each signer
+
+[NOTE]
+====
+To avoid duplicate signers, the contract uses `keccak256` to generate a unique id for each signer. When providing a multisignature, the `signers` array must be sorted in ascending order by `keccak256`, and the `signatures` array must match the order of their corresponding signers.
```
Amxx:
```suggestion
To avoid duplicate signers, the contract uses `keccak256` to generate a unique id for each signer. When providing a multisignature, the `signers` array should be sorted in ascending order by `keccak256`, and the `signatures` array must match the order of their corresponding signers.
```

contracts/mocks/docs/account/MyFactoryAccount.sol
```diff
@@ -0,0 +1,42 @@
...
+ * @dev A factory contract to create accounts on demand.
+ */
+contract MyFactoryAccount {
+    using Clones for address;
+    using Address for address;
+
+    address private immutable _impl;
+
+    constructor(address impl_) {
+        _impl = impl_;
```
Amxx, line 20:
```suggestion
        require(impl_.code.length > 0);
        _impl = impl_;
```

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -0,0 +1,354 @@
...
+[source,typescript]
+----
+import { signTypedData } from 'viem/actions';
+
+// EntryPoint v0.8 EIP-712 domain
+const domain = {
+  name: 'Account Abstraction',
+  version: '0.8.0',
+  chainId: 1, // Your target chain ID
+  verifyingContract: '0x<ENTRYPOINT_ADDRESS>',
```
ernestognw: TODO: Replace

docs/modules/ROOT/pages/accounts.adoc
```diff
@@ -241,15 +241,61 @@ IMPORTANT: A penalty of 10% (`UNUSED_GAS_PENALTY_PERCENT`) is applied on the amo
...
+Alternatively, developers can get the raw user operation hash by using the Entrypoint's `getUserOpHash` function:
 
 [source,typescript]
 ----
 const userOpHash = await entrypoint.read.getUserOpHash([userOp]);
 userOp.signature = await eoa.sign({ hash: userOpHash });
 ----
 
-NOTE: The previous example assumes that the account is owned by a single ECDSA signer without any particular signature format.
+NOTE: Using `getUserOpHash` directly may provide a poorer user experience as users see an opaque hash rather than structured transaction data.
```
Amxx: Its worst than that, most signers will not have a (raw)sign options.

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -151,7 +151,7 @@ abstract contract MultiSignerERC7913 is AbstractSigner {
      *
      * Requirements:
      *
-     * * The {signers}'s length must be `>=` to the {threshold}. Throws {MultiSignerERC7913UnreachableThreshold} if not.
+     * * The {getSignerCount}'s length must be `>=` to the {threshold}. Throws {MultiSignerERC7913UnreachableThreshold} if not.
```
```suggestion
     * * The {getSignerCount} must be greater or equal than to the {threshold}. Throws {MultiSignerERC7913UnreachableThreshold} if not.
```

docs/modules/ROOT/pages/multisig.adoc
```diff
@@ -0,0 +1,306 @@
...
+
+NOTE: `MultiSignerERC7913` safeguards to ensure that the threshold remains achievable based on the current number of active signers, preventing situations where operations could become impossible to execute.
+
+The contract also provides public functions for querying signer information: xref:api:utils/cryptography.adoc#MultiSignerERC7913-isSigner-bytes-[`isSigner(bytes memory signer)`] to check if a given signer is authorized, xref:api:utils/cryptography.adoc#MultiSignerERC7913-getSigners-uint64-uint64-[`getSigners(uint64 start, uint64 end)`] to retrieve a paginated list of authorized signers, and xref:api:utils/cryptography.adoc#MultiSignerERC7913-getSignerCount[`getSignerCount()`] to get the total number of signers. These functions are useful when validating signatures, implementing customized access control logic, or building user interfaces that need to display signer information.
+
+=== MultiSignerERC7913Weighted
+p
```
Amxx:
```suggestion

```

contracts/mocks/docs/account/MyFactoryAccount.sol
```diff
@@ -0,0 +1,42 @@
...
+            _impl.cloneDeterministic(_saltedCallData(salt, callData));
+            predicted.functionCall(callData);
+        }
+        return predicted;
+    }
+
+    function _saltedCallData(bytes32 salt, bytes calldata callData) internal pure returns (bytes32) {
+        // Scope salt to the callData to avoid front-running the salt with a different callData
+        return keccak256(abi.encodePacked(salt, callData));
+    }
```
Amxx: There is something interresting to notice when we pass arbitrary "initialize" calldata to the factory:
If we add more stuff after the abi encoded call (like a suffix), it won't be processed by the implementation (during initialization) but if will affect the hash. So technically, we can make the salt optional here and do

```suggestion
    /// @dev Predict the address of the account
    function predictAddress(bytes calldata callData) public view returns (address) {
        return _impl.predictDeterministicAddress(keccak256(callData), address(this));
    }

    /// @dev Create clone accounts on demand
    function cloneAndInitialize(bytes calldata callData) public returns (address) {
        address predicted = predictAddress(callData);
        if (predicted.code.length == 0) {
            _impl.cloneDeterministic(keccak256(callData));
            predicted.functionCall(callData);
        }
        return predicted;
    }
```

If someone what to use a salt, they will "just" have to call `cloneAndInitialize` with
```abi.encodePacked(abi.encodeCall(Account.intialize, (<...args...>)), salt);```

Amxx: The question is, given that both versions have similar features, do we want:
- a more simpler version, easy to read, but where some features are "hidden" (and optional)
- a more complex version, harder to read, but where the extra feature are "in your face"

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
ernestognw: I think this makes the line too large, and not sure how to keep a single bullet point in 2 rows

contracts/mocks/docs/account/MyFactoryAccount.sol
ernestognw: No strong preference, I think the factory is illustrative and the documentation already suggests some key considerations like "ensure the account address is deterministically tied to the initial owners". So all good with the current suggestion.

Overall I agree that an automated factory generation in Wizard must do the job better than this, and we may want to go for that option instead of a canonical factory in OZ contracts.

Amxx (approved):

PR #5680, opened 2025-05-10
Add RLP library

ernestognw:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/RLP.sol
+++ b/contracts/utils/RLP.sol
@@ -1,0 +1,228 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.25;
+
+import {Math} from "./math/Math.sol";
+import {Endianness} from "./math/Endianness.sol";
+import {Bytes} from "./Bytes.sol";
+
+library RLP {
+    using Bytes for bytes;
+
+    struct Item {
+        uint256 length;
+        bytes32 ptr;
+    }
+
+    enum ItemType {
+        DATA_ITEM,
+        LIST_ITEM
+    }
+
+    uint8 internal constant SHORT_THRESHOLD = 55;
+
+    uint8 internal constant SHORT_OFFSET = 128;
+    uint8 internal constant LONG_LENGTH_OFFSET = SHORT_OFFSET + SHORT_THRESHOLD + 1; // 184
+    uint8 internal constant LONG_OFFSET = LONG_LENGTH_OFFSET + 8; // 192
+    uint8 internal constant SHORT_LIST_OFFSET = LONG_OFFSET + SHORT_THRESHOLD + 1; // 248
+
+    function encode(bytes memory buffer) internal pure returns (bytes memory) {
+        return _isSingleByte(buffer) ? buffer : bytes.concat(_encodeLength(buffer.length, SHORT_OFFSET), buffer);
+    }
+
+    function encode(bytes[] memory list) internal pure returns (bytes memory) {
+        bytes memory flattened = _flatten(list);
+        return bytes.concat(_encodeLength(flattened.length, LONG_OFFSET), flattened);
+    }
+
+    function encode(string memory str) internal pure returns (bytes memory) {
+        return encode(bytes(str));
+    }
+
+    function encode(address addr) internal pure returns (bytes memory) {
+        return encode(abi.encodePacked(addr));
+    }
+
+    function encode(uint256 value) internal pure returns (bytes memory) {
+        return encode(_toBinaryBuffer(value));
+    }
+
+    function encode(bytes32 value) internal pure returns (bytes memory) {
+        return encode(_toBinaryBuffer(uint256(value)));
+    }
+
+    function encode(bool value) internal pure returns (bytes memory) {
+        bytes memory encoded = new bytes(1);
+        encoded[0] = value ? bytes1(0x01) : bytes1(SHORT_OFFSET); // false is encoded as an empty string
+        return encoded;
+    }
+
+    function toItem(bytes memory value) internal pure returns (Item memory) {
+        require(value.length != 0); // Empty arrays are not RLP items.
+        return Item(value.length, _skippedLengthPtr(value));
+    }
+
+    function readList(Item memory item) internal pure returns (Item[] memory) {
+        (uint256 listOffset, uint256 listLength, ItemType itemType) = _decodeLength(item);
+        require(itemType == ItemType.LIST_ITEM);
+        require(listOffset + listLength == item.length);
+        Item[] memory items = new Item[](32);
+
+        uint256 itemCount = item.length;
+
+        for (uint256 i; listOffset < itemCount; i++) {
+            (uint256 itemOffset, uint256 itemLength, ) = _decodeLength(
+                Item(itemCount - listOffset, bytes32(uint256(item.ptr) + listOffset))
+            );
+            items[i] = Item(itemLength + itemOffset, bytes32(uint256(item.ptr) + listOffset));
+            listOffset += itemOffset + itemLength;
+        }
+
+        // Decrease the array size to match the actual item count.
+        assembly ("memory-safe") {
+            mstore(items, itemCount)
+        }
+        return items;
+    }
+
+    function readList(bytes memory value) internal pure returns (Item[] memory) {
+        return readList(toItem(value));
+    }
+
+    function readBytes(Item memory item) internal pure returns (bytes memory) {
+        (uint256 itemOffset, uint256 itemLength, ItemType itemType) = _decodeLength(item);
+        require(itemType == ItemType.DATA_ITEM);
+        require(item.length == itemOffset + itemLength);
+        return _copy(item.ptr, bytes32(itemOffset), itemLength);
+    }
+
+    function readBytes(bytes memory item) internal pure returns (bytes memory) {
+        return readBytes(toItem(item));
+    }
+
+    function readRawBytes(Item memory item) internal pure returns (bytes memory) {
+        return _copy(item.ptr, 0, item.length);
+    }
+
+    function _isSingleByte(bytes memory buffer) private pure returns (bool) {
+        return buffer.length == 1 && uint8(buffer[0]) <= SHORT_OFFSET - 1;
+    }
+
+    function _encodeLength(uint256 length, uint256 offset) private pure returns (bytes memory) {
+        return
+            length <= SHORT_THRESHOLD
+                ? abi.encodePacked(bytes1(uint8(length) + uint8(offset)))
+                : _encodeLongLength(length, offset);
+    }
+
+    function _encodeLongLength(uint256 length, uint256 offset) private pure returns (bytes memory) {
+        uint256 bytesLength = Math.log256(length) + 1; // Result is floored
+        return
+            abi.encodePacked(
+                bytes1(uint8(bytesLength) + uint8(offset) + SHORT_THRESHOLD),
+                Endianness.reverseUint256(length) // to big-endian
+            );
+    }
+
+    function _toBinaryBuffer(uint256 value) private pure returns (bytes memory) {
+        uint256 leadingZeroes = _countLeadingZeroBytes(value);
+        return abi.encodePacked(value).slice(leadingZeroes);
+    }
+
+    function _countLeadingZeroBytes(uint256 x) private pure returns (uint256) {
+        uint256 r = 0;
+        if (x > 0xffffffffffffffffffffffffffffffff) r = 128; // Upper 128 bits
+        if ((x >> r) > 0xffffffffffffffff) r |= 64; // Next 64 bits
+        if ((x >> r) > 0xffffffff) r |= 32; // Next 32 bits
+        if ((x >> r) > 0xffff) r |= 16; // Next 16 bits
+        if ((x >> r) > 0xff) r |= 8; // Next 8 bits
+        return 31 ^ (r >> 3); // Convert to leading zero bytes count
+    }
+
+    function _flatten(bytes[] memory list) private pure returns (bytes memory) {
+        bytes memory flattened = new bytes(_totalLength(list));
+        bytes32 dataPtr = _skippedLengthPtr(flattened);
+        for (uint256 i = 0; i < list.length; i++) {
+            bytes memory item = list[i];
+            uint256 length = item.length;
+            _copy(dataPtr, _skippedLengthPtr(item), length);
+            dataPtr = bytes32(uint256(dataPtr) + length);
+        }
+        return flattened;
+    }
+
+    function _totalLength(bytes[] memory list) private pure returns (uint256) {
+        uint256 totalLength;
+        for (uint256 i = 0; i < list.length; i++) {
+            totalLength += list[i].length;
+        }
+        return totalLength;
+    }
+
+    function _decodeLength(Item memory item) private pure returns (uint256 offset, uint256 length, ItemType) {
+        require(item.length != 0);
+        bytes32 ptr = item.ptr;
+        uint256 prefix = uint8(_extractMemoryByte(ptr));
+        if (prefix < SHORT_OFFSET) return (0, 1, ItemType.DATA_ITEM); // Single byte.
+        if (prefix < LONG_LENGTH_OFFSET) return _decodeShortString(prefix - SHORT_OFFSET, item);
+        if (prefix < LONG_OFFSET) {
+            (offset, length) = _decodeLong(prefix - LONG_LENGTH_OFFSET, item);
+            return (offset, length, ItemType.DATA_ITEM);
+        }
+        if (prefix < SHORT_LIST_OFFSET) return _decodeShortList(prefix - LONG_OFFSET, item);
+        (offset, length) = _decodeLong(prefix - SHORT_LIST_OFFSET, item);
+        return (offset, length, ItemType.LIST_ITEM);
+    }
+
+    function _decodeShortString(
+        uint256 strLength,
+        Item memory item
+    ) private pure returns (uint256 offset, uint256 length, ItemType) {
+        require(item.length > strLength);
+        require(strLength != 1 || _extractMemoryByte(bytes32(uint256(item.ptr) + 1)) >= bytes1(SHORT_OFFSET));
+        return (1, strLength, ItemType.DATA_ITEM);
+    }
+
+    function _decodeShortList(
+        uint256 listLength,
+        Item memory item
+    ) private pure returns (uint256 offset, uint256 length, ItemType) {
+        require(item.length > listLength);
+        return (1, listLength, ItemType.LIST_ITEM);
+    }
+
+    function _decodeLong(uint256 lengthLength, Item memory item) private pure returns (uint256 offset, uint256 length) {
+        lengthLength += 1; // 1 byte for the length itself
+        require(item.length > lengthLength);
+        require(_extractMemoryByte(item.ptr) != 0x00);
+
+        uint256 len = _extractMemoryWord(bytes32(uint256(item.ptr) + 1)) >> (256 - 8 * lengthLength);
+        require(len > SHORT_OFFSET);
+        require(item.length <= lengthLength + len);
+        return (lengthLength + 1, len);
+    }
+
+    function _copy(bytes32 destPtr, bytes32 srcPtr, uint256 length) private pure returns (bytes memory src) {
+        assembly ("memory-safe") {
+            mcopy(destPtr, srcPtr, length)
+            src := mload(src)
+        }
+    }
+
+    function _skippedLengthPtr(bytes memory buffer) private pure returns (bytes32 ptr) {
+        assembly ("memory-safe") {
+            ptr := add(buffer, 32)
+        }
+    }
+
+    function _extractMemoryByte(bytes32 ptr) private pure returns (bytes1 v) {
+        assembly ("memory-safe") {
+            v := byte(0, mload(ptr))
+        }
+    }
+
+    function _extractMemoryWord(bytes32 ptr) private pure returns (uint256 v) {
+        assembly ("memory-safe") {
+            v := mload(ptr)
+        }
+    }
+}
--- a/contracts/utils/math/Endianness.sol
+++ b/contracts/utils/math/Endianness.sol
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+library Endianness {
+    function reverseUint256(uint256 value) internal pure returns (uint256) {
+        value = // swap bytes
+            ((value >> 8) & 0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF) |
+            ((value & 0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF) << 8);
+        value = // swap 2-byte long pairs
+            ((value >> 16) & 0x0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF) |
+            ((value & 0x0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF) << 16);
+        value = // swap 4-byte long pairs
+            ((value >> 32) & 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF) |
+            ((value & 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF) << 32);
+        value = // swap 8-byte long pairs
+            ((value >> 64) & 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF) |
+            ((value & 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF) << 64);
+        return (value >> 128) | (value << 128); // swap 16-byte long pairs
+    }
+
+    function reverseUint128(uint128 value) internal pure returns (uint256) {
+        value = // swap bytes
+            ((value & 0xFF00FF00FF00FF00FF00FF00FF00FF00) >> 8) |
+            ((value & 0x00FF00FF00FF00FF00FF00FF00FF00FF) << 8);
+        value = // swap 2-byte long pairs
+            ((value & 0xFFFF0000FFFF0000FFFF0000FFFF0000) >> 16) |
+            ((value & 0x0000FFFF0000FFFF0000FFFF0000FFFF) << 16);
+        value = // swap 4-byte long pairs
+            ((value & 0xFFFFFFFF00000000FFFFFFFF00000000) >> 32) |
+            ((value & 0x00000000FFFFFFFF00000000FFFFFFFF) << 32);
+        return (value >> 64) | (value << 64); // swap 8-byte long pairs
+    }
+
+    function reverseUint64(uint64 value) internal pure returns (uint256) {
+        value = ((value & 0xFF00FF00FF00FF00) >> 8) | ((value & 0x00FF00FF00FF00FF) << 8); // swap bytes
+        value = ((value & 0xFFFF0000FFFF0000) >> 16) | ((value & 0x0000FFFF0000FFFF) << 16); // swap 2-byte long pairs
+        return (value >> 32) | (value << 32); // swap 4-byte long pairs
+    }
+
+    function reverseUint32(uint64 value) internal pure returns (uint256) {
+        value = ((value & 0xFF00FF00) >> 8) | ((value & 0x00FF00FF) << 8); // swap bytes
+        return (value >> 16) | (value << 16); // swap 2-byte long pairs
+    }
+
+    function reverseUint16(uint64 value) internal pure returns (uint256) {
+        return (value >> 8) | (value << 8);
+    }
+}


0xClandestine:
will gladly help on this

ernestognw:
Hi @0xClandestine, so in there is [an open discussion](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5189#discussion_r2137265801) in #5189, some open issues in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5724 and https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5725

I would appreciate help in advancing any of these PRs, as of now we're attending the 5.4 audit release and these are 5.5 candidates. For this particular PR, we're missing test, je 😅

ernestognw:
This PR must be updated if any of #5792 or #5726 is merged. Essentially replacing these [RLP private functions](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5680/commits/acb14cb2c30545d7130cab2a1939b87330b8a7f6#diff-0399261b2cd954212dda0147b86044380b879ab21e1d1b89c564cb48f3552216R300-R321) and string's `equal`

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,244 @@
...
+            // must be either a branch or leaf/extension node
+            if (nodeLength != branchNodeLength && nodeLength != LEAF_OR_EXTENSION_NODE_LENGTH)
+                return ("", ProofError.UNPARSEABLE_NODE);
+
+            if (nodeLength == branchNodeLength) {
+                // If we've consumed the entire key, the value must be in the last slot
+                if (keyIndex == key.length) return _validateLastItem(node.decoded[radix], trieProof, i);
+
+                // Otherwise, continue down the branch specified by the next nibble in the key
+                uint8 branchKey = uint8(key[keyIndex]);
```
0xClandestine: On my point before I don't think it's necessary to convert the key to nibbles beforehand, makes more sense to extract the Nth nibble here.

contracts/utils/Bytes.sol
```diff
@@ -99,6 +99,28 @@ library Bytes {
...
 
+    /// @dev Split each byte in `value` into two nibbles (4 bits each).
+    function nibbles(bytes memory value) internal pure returns (bytes memory) {
+        uint256 length = value.length;
+        bytes memory nibbles_ = new bytes(length * 2);
+        for (uint256 i = 0; i < length; i++) {
+            (nibbles_[i * 2], nibbles_[i * 2 + 1]) = (value[i] & 0xf0, value[i] & 0x0f);
+        }
+        return nibbles_;
+    }
```
0xClandestine: The `toNibbles` function converts a bytes array like 0xABCD into a nibble-expanded format like 0xA00BC00D. The existing implementation does this by iterating linearly over the input, which works correctly but is extremely inefficient.

My approach processes the input in parallel, reducing time complexity from O(n) to O(n / 32). This is inspired by bit-hacking techniques for interleaving bits, as described in the following resources:

https://stackoverflow.com/questions/38881877/bit-hack-expanding-bits
https://graphics.stanford.edu/~seander/bithacks.html#InterleaveBMN

```solidity
    // 0xABCD -> 0xA0B0C0D0
    function expandNibbles(uint128 x) internal pure returns (uint256) {
        uint256 y = uint256(x);
        y = (y | (y << 64)) & 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF;
        y = (y | (y << 32)) & 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF;
        y = (y | (y << 16)) & 0x0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF;
        y = (y | (y << 8)) & 0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF;
        y = (y | (y << 4)) & 0x0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F;
        return y;
    }
    
    // Puts nibbles in their respective height (high/low). 
    // 0xA0B0C0D0 -> 0xA00BC00D
    function toNibbles(uint128 x) internal pure returns (uint256) {
        uint256 y = expandNibbles(x);
        uint256 high = y & 0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF;
        uint256 low = y & 0xFF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00;
        high = high >> 4;
        return high | low;
    }
```

0xClandestine: Bringing it all together:

```solidity
function toNibbles(bytes memory input) internal pure returns (bytes memory output) {
    assembly {
        output := mload(0x40)
        mstore(output, mul(mload(input), 2))
        mstore(0x40, add(output, add(0x20, mul(mload(input), 2))))
        
        for { let i := 0 } lt(i, mload(input)) { i := add(i, 16) } {
            let x := shr(128, mload(add(add(input, 0x20), i)))
            x := and(or(x, shl(64, x)), 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF)
            x := and(or(x, shl(32, x)), 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF)
            x := and(or(x, shl(16, x)), 0x0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF)
            x := and(or(x, shl(8, x)),  0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF)
            let mask := 0x0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F0F
            x := and(or(x, shl(4, x)),  mask)
            x := or(shl(4, and(shr(4, x), mask)), and(x, mask))
            mstore(add(add(output, 0x20), mul(i, 2)), x)
        }
    }
}
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,244 @@
...
+     * Used to determine how much of a path matches a key during trie traversal.
+     */
+    function _sharedNibbleLength(bytes memory _a, bytes memory _b) private pure returns (uint256 shared_) {
+        uint256 max = Math.max(_a.length, _b.length);
+        uint256 length;
+        while (length < max && _a[length] == _b[length]) {
+            length++;
+        }
+        return length;
+    }
```
0xClandestine: This can also be much more efficient by comparing 32 bytes at a time. Once a mismatch is found, use xor and clz to count leading matching bytes within the word. 

O(n) -> O(n/32)

contracts/utils/Bytes.sol
0xClandestine: Can probably be simplified further by adjusting the first 4 masks, working on that now.

contracts/utils/cryptography/TrieProof.sol
0xClandestine:
```solidity
function _getSharedNibbleLength(bytes memory _a, bytes memory _b) private pure returns (uint256 shared) {
    uint256 minLen = _a.length < _b.length ? _a.length : _b.length;

    unchecked {
        // Compare 32 bytes at a time
        for (; shared < minLen; shared += 32) {
            uint256 wordA;
            uint256 wordB;

            assembly {
                wordA := mload(add(_a, add(32, shared)))
                wordB := mload(add(_b, add(32, shared)))
            }

            if (wordA != wordB) {
                uint256 diff = wordA ^ wordB;
                uint256 leadingBits = clz(diff); // clz counts bits
                return shared + (leadingBits / 8);
            }
        }

        // Fallback to byte-by-byte for remaining bytes
        while (shared < minLen && _a[shared] == _b[shared]) shared++;
    }
}
```

0xClandestine: Use unchecked obviously

0xClandestine: With proofs in calldata, and the above the optimizations I'm hitting 175k gas for an account proof verification (down from over 400k).

0xClandestine: Mind you I'm working off of optimisms implementation, this version looks a bit more expensive.

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,327 @@
...
+            (offset, length) = _decodeLong(prefix - LONG_LENGTH_OFFSET, item);
+            return (offset, length, ItemType.DATA_ITEM);
+        }
+
+        // Short list
+        if (prefix < SHORT_LIST_OFFSET) return _decodeShortList(prefix - LONG_OFFSET, item);
+
+        // Long list
+        (offset, length) = _decodeLong(prefix - SHORT_LIST_OFFSET, item);
+        return (offset, length, ItemType.LIST_ITEM);
```
0xClandestine: Feel like we should use an if statement with a revert case after.

contracts/utils/cryptography/TrieProof.sol
ernestognw: Tests. Then I optimize

0xClandestine: No worries, just wanted to write this out while I had a minute.

ernestognw: Yeah, thanks for the review though, it's just I haven't put time into making sure the implementation fully works.

It's a priority for 5.5 regardless. Just, you know, the release will take a couple of months 😓

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,327 @@
...
+     * Format: [prefix + length of the length] + [length in big-endian]
+     */
+    function _encodeLongLength(uint256 length, uint256 offset) private pure returns (bytes memory) {
+        uint256 bytesLength = length.log256() + 1; // Result is floored
+        return
+            abi.encodePacked(
+                bytes1(uint8(bytesLength) + uint8(offset) + SHORT_THRESHOLD),
+                _binaryBuffer(length) // already in big-endian, minimal representation
+            );
+    }
```
Amxx: We should be able to merge thewe two functions, `_encodedLongLength` is not used anywhere else (that I can see)

```suggestion
    /**
     * @dev Encodes a length with appropriate RLP prefix.
     *
     * For lengths <= 55 bytes, uses short encoding: [ length (1 byte) ++ offset (1 byte) ]
     * For lengths > 55 bytes, Uses long encoding using a length-of-length prefix: [ prefix ++ length of the length ++ length in big-endian ]
     */
    function _encodeLength(uint256 length, uint256 offset) private pure returns (bytes memory) {
        return
            length <= SHORT_THRESHOLD
                ? abi.encodePacked(bytes1(uint8(length) + uint8(offset)))
                : abi.encodePacked(
                    bytes1(uint8(length.log256() + 1) + uint8(offset) + SHORT_THRESHOLD),
                    _binaryBuffer(length) // already in big-endian, minimal representation
                );
    }
```

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,327 @@
...
+
+    /// @dev Convenience method to encode a uint256 as RLP. See {_binaryBuffer}.
+    function encode(uint256 value) internal pure returns (bytes memory) {
+        return encode(_binaryBuffer(value));
+    }
+
+    /// @dev Same as {encode-uint256-}, but for bytes32.
+    function encode(bytes32 value) internal pure returns (bytes memory) {
+        return encode(uint256(value));
+    }
```
Amxx: This is going to remove leeding zeros. is that desirable ?

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,327 @@
...
+     * @dev Convenience method to encode a boolean as RLP.
+     *
+     * Boolean `true` is encoded as 0x01, `false` as 0x80 (equivalent to encoding integers 1 and 0).
+     * This follows the de facto ecosystem standard where booleans are treated as 0/1 integers.
+     *
+     * NOTE: Both this and {encodeStrict} produce identical encoded bytes at the output level.
+     * Use this for ecosystem compatibility; use {encodeStrict} for strict RLP spec compliance.
+     */
+    function encode(bool value) internal pure returns (bytes memory) {
+        return encode(value ? uint256(1) : uint256(0));
```
Amxx: We have a utility for that in SafeCast that avoids the ternary jumps 
```suggestion
        return encode(value.toUint());
```

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,327 @@
...
+
+    /**
+     * @dev Strict RLP encoding of a boolean following literal spec interpretation.
+     * Boolean `true` is encoded as 0x01, `false` as empty bytes (0x80).
+     *
+     * NOTE: This is the strict RLP spec interpretation where false represents "empty".
+     * Use this for strict RLP spec compliance; use {encode} for ecosystem compatibility.
+     */
+    function encodeStrict(bool value) internal pure returns (bytes memory) {
+        return value ? abi.encodePacked(bytes1(0x01)) : encode(new bytes(0));
```
Amxx: Would that produce the same results with fewer memory allocations/manipulation ?
```suggestion
        return abi.encodePacked(bytes1(value.ternary(0x01, 0x80)))
```

contracts/utils/RLP.sol
ernestognw: Yeah, this is desirable and part of the spec: https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/

> Positive integers must be represented in big-endian binary form with no leading zeroes (thus making the integer value zero equivalent to the empty byte array). Deserialized positive integers with leading zeroes must be treated as invalid by any higher-order protocol using RLP.

contracts/utils/RLP.sol
ernestognw: No. Unfortunately it doesn't work because all inputs of RLP are "strngs" (i.e. items), so the value for `false` should be represented as an empty buffer. Therefore `true == 0x01` and `false == 0x80`.

Amxx: > Therefore true == 0x01 and false == 0x80

doesn't that match what I'm proposing ?

ernestognw: Ah you're right. I thought we were trying to reencode the 0x80. Tests were failing because we would need this to make it work given endianness is different between `uint256` and `bytes32`

```
return abi.encodePacked(bytes1(bytes32(value.ternary(0x01, 0x80))[31]));
```

Will update

ernestognw: Updated in 8928039

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+     * @dev Strict RLP encoding of a boolean following literal spec interpretation.
+     * Boolean `true` is encoded as 0x01, `false` as empty bytes (0x80).
+     *
+     * NOTE: This is the strict RLP spec interpretation where false represents "empty".
+     * Use this for strict RLP spec compliance; use {encode} for ecosystem compatibility.
+     */
+    function encodeStrict(bool value) internal pure returns (bytes memory) {
```
Amxx: From the documentation, this has the same behavior as `encode(bool)`. If there is no difference between the two, I'd only keep one.

test/utils/RLP.t.sol
```diff
@@ -0,0 +1,115 @@
...
+import {Test} from "forge-std/Test.sol";
+import {RLP} from "@openzeppelin/contracts/utils/RLP.sol";
+import {Memory} from "@openzeppelin/contracts/utils/Memory.sol";
+
+contract RLPTest is Test {
+    using RLP for *;
+
+    // To bytes
+
+    function testEncodeDecodeAddress(address input) external pure {
```
james-toussaint, line 18: Should/could we provide same kind of experience of `abi.encode(..)` & `abi.decode(..)`?
- Encode with`uint256 u`, `bytes memory b`, `address a`
`bytes memory encoded = RLP.encode(u, b, a)`

- Decode with `address input`, `data = input.encode()`
`address decoded = RLP.decode(data, address)`

Amxx: `abi.encode` and `abi.decode` are handled by the compiler. They are like macros. I don't think its possible to do the same at our level.

For encoding, its technically possible if we have "all" possible combination available, and the override systems selects it ... but the fact that it would require a ton of code, and there will always be holes makes me say its probably not worth it.
For decoding, types information are not an input param we can manipulate. We would need some kind of variadic templates ...

james-toussaint: I thought the same for encoding, with an (impossible?) `RLP.js` template but I don't see possible options.
For decoding, "types information are not an input param we can manipulate", I didn't know thank you 🙏

Amxx: Basically: the decoding would need the output type depends on the input arguments. That is not something possible in solidity.

For the encoding, one thing to consider is also the size of the solidity file. If we consider 7 input types
`bool`, `address`, `uint256`, `bytes32`, `string`, `bytes`, `bytes[]` (I'm not even considering the int256), we have
- 7 versions with 1 args (we have that today)
- 49 versions with 2 args
- 343 versions with 3 args
- 2401 versions with 4 args.
So if we want everything from 1 to 4 args, its a total of 2800 functions. with each function being ~200 char, that is a 560k file. It would definitelly not align with "keeping the library light".

Note: you may notice that in Packing.sol, not all types are supported. We only support types that are bytes1, bytes2, bytes4 and multiple of bytes4. I don't think we support things like bytes27 or bytes15. The reason for that is, if we support all types, the library grows to much. An option would be to ship the prodecural generation (instead of the solidity) and have it automatically deflate on the client's end ... but it would be a security nightmare.

Amxx:
We should decide weither we are ok with the following feature:
- Memory slices (in memory.sol ?)
- Bytes accumulator (in bytes.sol ? in a dedicated file under struct ?)
- RLP encoder interface (uses Bytes accumulator)

Overall, we should also discuss naming of the functions in RLP.sol.

test/utils/Bytes.t.sol
```diff
@@ -7,7 +7,7 @@ import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
 import {Bytes} from "@openzeppelin/contracts/utils/Bytes.sol";
 
 contract BytesTest is Test {
-    using Bytes for bytes;
+    using Bytes for *;
```
Amxx:
```suggestion
    using Bytes for bytes;
```

.changeset/wise-webs-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Accumulators: A library for merging an arbitrary dynamic number of bytes buffers.
```
Amxx:
```suggestion
`Accumulators`: A library for merging an arbitrary dynamic number of bytes buffers.
```

.changeset/modern-moments-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Memory`: Add a UDVT for handling slices or memory space similarly to calldata slices.
```
ernestognw: Was this a mistake?

```suggestion
`Memory`: Add a UDVT for handling slices on memory space similarly to calldata slices.
```

contracts/utils/Memory.sol
```diff
@@ -41,4 +43,71 @@ library Memory {
     function asPointer(bytes32 value) internal pure returns (Pointer) {
         return Pointer.wrap(value);
     }
+
+    type Slice is bytes32;
+
+    /// @dev Get a slice representation of a bytes object in memory
+    function asSlice(bytes memory self) internal pure returns (Slice result) {
+        assembly ("memory-safe") {
+            result := or(shl(128, mload(self)), add(self, 0x20))
```
ernestognw, line 65: For consistency

```suggestion
            result := or(shl(0x80, mload(self)), add(self, 0x20))
```

ernestognw: Curios if the 0x80 selection, is it just to fit the length and pointer together.

If so, I wonder if 128 bits is the right choice. It may be too large of a number regardless.

contracts/utils/Memory.sol
```diff
@@ -41,4 +43,71 @@ library Memory {
...
+        }
+    }
+
+    /// @dev Private helper: create a slice from raw values (length and pointer)
+    function _asSlice(uint256 len, Memory.Pointer ptr) private pure returns (Slice result) {
+        // TODO: Fail if len or ptr  is larger than type(uint128).max ?
+        assembly ("memory-safe") {
+            result := or(shl(128, len), ptr)
```
ernestognw:
```suggestion
            result := or(shl(0x80, len), ptr)
```

contracts/utils/Memory.sol
```diff
@@ -41,4 +43,71 @@ library Memory {
...
+    /// @dev Get a slice representation of a bytes object in memory
+    function asSlice(bytes memory self) internal pure returns (Slice result) {
+        assembly ("memory-safe") {
+            result := or(shl(128, mload(self)), add(self, 0x20))
+        }
+    }
+
+    /// @dev Private helper: create a slice from raw values (length and pointer)
+    function _asSlice(uint256 len, Memory.Pointer ptr) private pure returns (Slice result) {
+        // TODO: Fail if len or ptr  is larger than type(uint128).max ?
```
ernestognw: Although a value larger than uint128 would corrupt the accumulator, it's difficult to weight between just noting it in the docs or effectively reverting with a panic code.

What do you think of just documenting it? The function is private

contracts/utils/Memory.sol
```diff
@@ -41,4 +43,71 @@ library Memory {
...
+    /// @dev Offset and cut a Slice (equivalent to self[start:start+length] for calldata slices)
+    function slice(Slice self, uint256 offset, uint256 len) internal pure returns (Slice) {
+        if (offset + len > length(self)) Panic.panic(Panic.ARRAY_OUT_OF_BOUNDS);
+        return _asSlice(len, asPointer(bytes32(uint256(asBytes32(_pointer(self))) + offset)));
+    }
+
+    /**
+     * @dev Read a bytes32 buffer from a given Slice at a specific offset
+     *
+     * Note:If offset > length(slice) - 32, part of the return value will be out of bound and should be ignored.
```
ernestognw: I'm not following this note.

Wouldn't be just a matter of cleaning up the value? I'd prefer that over suggesting that the value must be ignored

contracts/utils/Memory.sol
```diff
@@ -41,4 +43,71 @@ library Memory {
...
+     */
+    function load(Slice self, uint256 offset) internal pure returns (bytes32 value) {
+        if (offset >= length(self)) Panic.panic(Panic.ARRAY_OUT_OF_BOUNDS);
+        assembly ("memory-safe") {
+            value := mload(add(and(self, shr(128, not(0))), offset))
```
ernestognw:
```suggestion
            value := mload(add(and(self, shr(0x80, not(0x00))), offset))
```

contracts/utils/Memory.sol
```diff
@@ -41,4 +43,71 @@ library Memory {
...
+        // TODO: Fail if len or ptr  is larger than type(uint128).max ?
+        assembly ("memory-safe") {
+            result := or(shl(128, len), ptr)
+        }
+    }
+
+    /// @dev Returns the memory location of a given slice (equiv to self.offset for calldata slices)
+    function _pointer(Slice self) private pure returns (Memory.Pointer result) {
+        assembly ("memory-safe") {
+            result := and(self, shr(128, not(0)))
```
ernestognw:
```suggestion
            result := and(self, shr(0x80, not(0x00)))
```

contracts/utils/Memory.sol
```diff
@@ -41,4 +43,71 @@ library Memory {
...
+    function _asSlice(uint256 len, Memory.Pointer ptr) private pure returns (Slice result) {
+        // TODO: Fail if len or ptr  is larger than type(uint128).max ?
+        assembly ("memory-safe") {
+            result := or(shl(128, len), ptr)
+        }
+    }
+
+    /// @dev Returns the memory location of a given slice (equiv to self.offset for calldata slices)
+    function _pointer(Slice self) private pure returns (Memory.Pointer result) {
+        assembly ("memory-safe") {
```
ernestognw: I'm worried about the memory safe annotations since a slice may be arbitrary. In regular circumstances it should be fine, but we can't trust everyone will use it correctly. Thoughts?

ernestognw (commented):
Left some comments in Memory, I will continue reviewing the RLP and Accumulators library later

contracts/utils/Memory.sol
Amxx: I want slices to be represented using a bytes32 an not a structure. That way we don't have to allocate memory for slice, they live on the (machine) stack like any other variable.

That means we have to fit 2 values in 32 bits. These values are "length" and "offset/ptr". Both will not realistically ever be bigger than 64 bits, but since I have 256 bits of space, I just slip that in 2 ans say "lets use 128 bits for each value".

contracts/utils/Memory.sol
Amxx: I share this concern, however:

- Slices that are built using `asSlice(bytes memory)` as safe.
- Slices that are build using `slice(Slice, uint256)` or `slice(Slice, uint256, uint256)` are also safe.

So basically, the only way to build an "unsafe" slice is to do it manually by doing `Slice.wrap(...)`, which we don't provide an helper for.

---

Note that is a slice is unsafe (pointer at unallocated location or length extends it beyond the allocated location), then `_pointer` and `length` will be safe. They don't access the memory (they just do arithmetics on the bytes32). The unsafe functions would be `load(Slice, uint256)` that does an `mload` and `toBytes(slice)` that does an `mcopy`

contracts/utils/Memory.sol
Amxx: Addressed in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5680/commits/8534802a59d40348e0c420f8dc2be528c3b4c2d4

contracts/utils/Memory.sol
Amxx: About the decimal to hexidecimal rewrite (applies to multiple comments).

Here, the number are bits offset/slift inside a bytes32. That is pretty similar to what we have in Packing.sol. 

For example
```solidity
    function extract_32_16(bytes32 self, uint8 offset) internal pure returns (bytes16 result) {
        if (offset > 16) revert OutOfRangeAccess();
        assembly ("memory-safe") {
            result := and(shl(mul(8, offset), self), shl(128, not(0)))
        }
    }
```

In Packing.sol, we use decimals to represent number between 0 and 256. IMO we should do the same here. I know we don't have a good guideline for this.

contracts/utils/Memory.sol
Amxx: For the record, our today most real machines use 64bits memory space. That means all the memory (including virtual memory) of all the apps running on our machine (include AI training, video encoding, 3D rendering, ...) is addressed by the OS using only 64bits (ie 8 bytes).

That represents 16777216 Tib of memory space.

32bits machine could handle 4Gib of memory space, which is very small by today's standard (for a PC). I don't expect smart contracts to handle that amount of memory in a TX in the near future, but I guess I could be surprised by this happening sooner than I expect.

contracts/token/ERC20/extensions/ERC4626.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 // OpenZeppelin Contracts (last updated v5.4.0) (token/ERC20/extensions/ERC4626.sol)
 
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.24;
```
ernestognw, line 4: Any reason for this?

contracts/utils/Memory.sol
ernestognw: Right. Thanks for putting #5936 together. I'm solving this thread since it goes according to them

contracts/utils/Memory.sol
```diff
@@ -41,4 +44,79 @@ library Memory {
...
+    /// @dev Offset and cut a Slice (equivalent to self[start:start+length] for calldata slices)
+    function slice(Slice self, uint256 offset, uint256 len) internal pure returns (Slice) {
+        if (offset + len > length(self)) Panic.panic(Panic.ARRAY_OUT_OF_BOUNDS);
+        return _asSlice(len, asPointer(bytes32(uint256(asBytes32(_pointer(self))) + offset)));
+    }
+
+    /**
+     * @dev Read a bytes32 buffer from a given Slice at a specific offset
+     *
+     * Note: If offset > length(slice) - 32, part of the return value will be out of bound of the slice. These bytes are zeroed.
```
ernestognw:
```suggestion
     * NOTE: If offset > length(slice) - 32, part of the return value will be out of bound of the slice. These bytes are zeroed.
```

contracts/utils/structs/Accumulators.sol
```diff
@@ -0,0 +1,104 @@
...
+import {Memory} from "../Memory.sol";
+
+/**
+ * @dev Structure concatenating an arbitrary number of bytes buffers with limited memory allocation.
+ */
+library Accumulators {
+    /**
+     * @dev Bytes accumulator: a linked list of `bytes`.
+     *
+     * Note: This is a memory structure that SHOULD not be put in storage.
```
ernestognw:
```suggestion
     * NOTE: This is a memory structure that SHOULD not be put in storage.
```

contracts/utils/structs/Accumulators.sol
```diff
@@ -0,0 +1,104 @@
...
+        assembly ("memory-safe") {
+            ptr := item
+        }
+    }
+
+    function _asAccumulatorEntry(Memory.Pointer ptr) private pure returns (AccumulatorEntry memory item) {
+        assembly ("memory-safe") {
+            item := ptr
+        }
+    }
```
ernestognw, line 124: I'd suggest a function to make explicit that the use of `0x00` is a null or sentinel.

```suggestion
    }
    
    function _nullPtr() private pure returns (Memory.Pointer) {
        return Memory.asPointer(0x00);
    }
```

contracts/utils/structs/Accumulators.sol
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.24;
+
+import {Memory} from "../Memory.sol";
+
+/**
+ * @dev Structure concatenating an arbitrary number of bytes buffers with limited memory allocation.
```
ernestognw, line 8: We would benefit from some instructions

```suggestion
 * @dev Structure concatenating an arbitrary number of bytes buffers with limited memory allocation.
 * 
 * The Accumulators library provides a memory-efficient alternative to repeated concatenation of bytes.
 * Instead of copying data on each concatenation (O(n**2) complexity), it builds a linked list of references
 * to existing data and performs a single memory allocation during flattening (O(n) complexity).
 *
 * Uses 0x00 as sentinel value for empty state (i.e. null pointers)
 *
 * == How it works
 *
 * 1. Create an empty accumulator with null head/tail pointers
 * 2. Add data using {push} (append) or {shift} (prepend). It creates linked list nodes
 * 3. Each node stores a reference to existing data (no copying)
 * 4. Call {flatten} to materialize the final concatenated result in a single operation
 *
 * == Performance
 *
 * * Addition: O(1) per operation (just pointer manipulation)
 * * Flattening: O(n) single pass with one memory allocation
 * * Memory: Minimal overhead until flattening (only stores references)
```

contracts/utils/Memory.sol
ernestognw, line 1: Since we haven't release this file, I just noticed it may be clever to follow the same pattern as in ERC7579Utils where we define the type at the file level and then we create free functions to overload operators:

```solidity
// Operators
using {eqPointer as ==} for Memory.Pointer global;

/// @dev Compares two `Pointer` values for equality.
function eqPointer(Memory.Pointer a, Memory.Pointer b) pure returns (bool) {
    return Memory.asBytes32(a) == Memory.asBytes32(b);
}
```

Maybe overengineered but just mentioning before it becomes a breaking change. Note it would populate the global namespace for `Pointer`

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,382 @@
...
+                if (strLength == 1) {
+                    require(bytes1(item.load(1)) >= bytes1(SHORT_OFFSET)); // TODO: custom error for sanity checks
+                }
+                return (1, strLength, ItemType.Data);
+            } else {
+                // Case: Long string (>55 bytes)
+                uint256 lengthLength = prefix - SHORT_OFFSET - SHORT_THRESHOLD;
+
+                require(itemLength > lengthLength, RLPInvalidDataRemainder(lengthLength, itemLength));
+                require(bytes1(item.load(0)) != 0x00); // TODO: custom error for sanity checks
```
ernestognw: Maybe panic with `RESOURCE_ERROR` code?

contracts/utils/Memory.sol
ernestognw: Yeah that makes sense. I think the concern should be explicit in the docs. I updated the NatSpec in fb62ca382, which I think it addresses the concer briefly.

contracts/utils/Memory.sol
ernestognw: I agree with representing the slices as a `bytes32` rather than a structure. So I guess the discussion boils down to whether use 128 bits or 64 bits to represent the length and offset. I was also thinking that in practice none will be bigger than 64 bits (or even 48), so I'm reading that we're purposely deciding to use 128 for convenience and also to acknowledge the potential change to 128 bits if it happens sooner.

Let's keep 128 bits.

contracts/token/ERC20/extensions/ERC4626.sol
ernestognw: If we're keeping it we should add it to the changelog. I did in 698221fe7

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    function decodeString(bytes memory item) internal pure returns (string memory) {
+        return readString(item.asSlice());
+    }
+
+    function decodeList(bytes memory value) internal pure returns (Memory.Slice[] memory) {
+        return readList(value.asSlice());
+    }
+
```
ernestognw, line 322: For API consistency

```suggestion
function decodeList(bytes memory value, uint256 maxListLength) internal pure returns (Memory.Slice[] memory) {
        return readList(value.asSlice(), maxListLength);
    }
```

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,382 @@
...
+        // Decrease the array size to match the actual item count.
+        assembly ("memory-safe") {
+            mstore(list, itemCount)
+        }
+        return list;
+    }
+
+    /****************************************************************************************************************
+     *                                            DECODING - FROM BYTES                                             *
```
ernestognw, line 285: This section is missing natspec

ernestognw (commented):
Almost there!

ernestognw:
The only thing left to discuss is the error to use in `_decodeLength`. See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5680#discussion_r2356559020

Once that's done I think the PR should be good. The main problem may be the coverage hit but we may be fine with that (?)

contracts/utils/Memory.sol
Amxx: It is not possible to do `using {...} for <type> global;` when the Type is in a library

> Can only use "global" with types defined in the same source unit at file level.

To do that, we would have to declare Pointer at the file level (so outside the Memory library), which means we would probably need to rename it.

contracts/utils/RLP.sol
```diff
@@ -0,0 +1,381 @@
...
+                mstore(ptr, element)
+            }
+        }
+
+        // write list length and reserve space
+        assembly ("memory-safe") {
+            mstore(list, div(sub(ptr, add(list, 0x20)), 0x20))
+            mstore(0x40, ptr)
+        }
+        return list;
```
Amxx:
```suggestion
```

contracts/utils/Memory.sol
ernestognw: That's why I was noting that it would populate the global namespace for `Pointer`. I agree it would require renaming which making me unsure.

I don't think there's much value in doing this anyway; the only interesting operator for Memory.Pointer may just be `=` and `+`.

ernestognw:
Can't approve my own PR, but LGTM

contracts/utils/Memory.sol
Amxx: I would not do pointer+pointer

I would eventually do Pointer+uint256, but a global operator would not support that

contracts/token/ERC20/extensions/ERC4626.sol
Amxx: This imports memory, and memory needs 0.24 because it does mcopy

ernestognw:
The `forward` looks pretty cool, though!

PR #5684, opened 2025-05-12
Add docs for the `Time` library

gonzaotc:
Added to the docs API plus a new `Time` section inside `Utilities`, including a basic example on how to use `Time.Delay` and update safely via the guaranties provided in this library.

### DIFF
--- a/contracts/utils/README.adoc
+++ b/contracts/utils/README.adoc
@@ -49,6 +49,7 @@ Miscellaneous contracts and libraries containing utility functions you can use t
  * {AbstractSigner}: Abstract contract for internal signature validation in smart contracts.
  * {ERC7739}: An abstract contract to validate signatures following the rehashing scheme from `ERC7739Utils`.
  * {ERC7739Utils}: Utilities library that implements a defensive rehashing mechanism to prevent replayability of smart contract signatures based on ERC-7739.
+ * {Time}: A library that provides helpers for manipulating time-related objects, including a `Delay` type.
  
 [NOTE]
 ====
@@ -170,3 +171,5 @@ Ethereum contracts have no native concept of an interface, so applications must 
 {{CAIP10}}
 
 {{Blockhash}}
+
+{{Time}}
--- a/docs/modules/ROOT/pages/utilities.adoc
+++ b/docs/modules/ROOT/pages/utilities.adoc
@@ -405,3 +405,81 @@ contract L1Inbox {
 ----
 
 IMPORTANT: After EIP-2935 activation, it takes 8,191 blocks to completely fill the history storage. Before that, only block hashes since the fork block will be available.
+
+=== Time
+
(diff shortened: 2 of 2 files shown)


contracts/utils/README.adoc
```diff
@@ -49,6 +49,7 @@ Miscellaneous contracts and libraries containing utility functions you can use t
  * {AbstractSigner}: Abstract contract for internal signature validation in smart contracts.
  * {ERC7739}: An abstract contract to validate signatures following the rehashing scheme from `ERC7739Utils`.
  * {ERC7739Utils}: Utilities library that implements a defensive rehashing mechanism to prevent replayability of smart contract signatures based on ERC-7739.
+ * {Time}: A library that provides helpers for manipulating time-related objects, including a `Delay` type.
```
ernestognw, line 52: I'm really surprised we missed this 😢

docs/modules/ROOT/pages/utilities.adoc
```diff
@@ -405,3 +405,81 @@ contract L1Inbox {
 ----
 
 IMPORTANT: After EIP-2935 activation, it takes 8,191 blocks to completely fill the history storage. Before that, only block hashes since the fork block will be available.
+
+=== Time
+
+The xref:api:utils.adoc#Time[`Time`] library provides helpers for manipulating time-related objects in a type-safe manner. It uses `uint48` for timepoints and `uint32` for durations, helping to reduce gas costs while providing adequate precision.
+
+One of its key features is the `Delay` type, which represents a duration that can automatically change its value at a specified point in the future while maintaining some security guarantees. When reducing a delay value (e.g., from 7 days to 1 day), the change only takes effect after a setback period, preventing an attacker who gains admin access from immediately reducing security timeouts and executing sensitive operations. This is particularly useful for governance and security mechanisms where timelock periods need to be enforced.
```
ernestognw, line 413:
```suggestion
One of its key features is the `Delay` type, which represents a duration that can automatically change its value at a specified point in the future while maintaining delay guarantees. For example, when reducing a delay value (e.g., from 7 days to 1 day), the change only takes effect after the difference between the old and new delay (i.e. a 6 days) or a minimum setback period, preventing an attacker who gains admin access from immediately reducing security timeouts and executing sensitive operations. This is particularly useful for governance and security mechanisms where timelock periods need to be enforced.
```

ernestognw (approved):
Looking good, just a couple nits

gonzaotc:
TBD: I would also like to improve the `Time.sol` NatSpec:
- Improve the top-level contract description to look good in our API docs, since the "* -" lines are not correctly formatted by our docs engine. 
- Provide a more detailed explanation of `withUpdate`, it's parameters and math, since I think it isn't intuitive at first sight.

ernestognw:
> TBD: I would also like to improve the `Time.sol` NatSpec:
> 
> * Improve the top-level contract description to look good in our API docs, since the "* -" lines are not correctly formatted by our docs engine.
> * Provide a more detailed explanation of `withUpdate`, it's parameters and math, since I think it isn't intuitive at first sight.

I think we can do these on a different PR, IMO this one is good to go and better than no docs. Also remember it won't be out until we release 5.4 so feel free to make changes in a different PR if you have bandwidth

PR #5394, opened 2024-12-20
Add ERC6909 Implementation along with extensions

arr00:
Fixes #???? 

#### PR Checklist

- [x] Tests
- [x] Documentation
- [x] Changeset entry (run `npx changeset add`)

PR #5686, opened 2025-05-14
feat: add ERC7572 interface for contracts that expose contract level metadata

coffeexcoin:
This PR introduces the interface for [draft ERC-7572](https://eips.ethereum.org/EIPS/eip-7572) to the set of interfaces in the package.

This interface is useful for contracts (particularly tokens) that wish to expose a set of metadata for the contract including icon, banner image, description, external url, etc. 

#### PR Checklist

- [ ] Tests (No tests are necessary as this is purely an interface spec)
- [x] Documentation (Natspec documentation is applied to the added interface)

### DIFF
--- a/.changeset/fluffy-bananas-add.md
+++ b/.changeset/fluffy-bananas-add.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Add draft ERC-7572 interface
--- a/contracts/interfaces/draft-IERC7572.sol
+++ b/contracts/interfaces/draft-IERC7572.sol
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.3.0) (interfaces/draft-IERC7572.sol)
+
+pragma solidity ^0.8.20;
+
+/// @title ERC-7572 Contract Level Metadata
+/// @dev Required interface of an ERC-7572 compliant contract, as defined in the
+/// https://eips.ethereum.org/EIPS/eip-7572[ERC].
+interface IERC7572 {
+
+  /// @dev This event should be emitted on updates to the contract metadata 
+  /// to indicate to offchain indexers that they should query the contract
+  /// for the latest URI.
+  event ContractURIUpdated();
+
+  /// @dev Returns an offchain resource for contract metadata or onchain JSON
+  /// data string (data:application/json;utf8,{}).
+  function contractURI() external view returns (string memory);
+}


.changeset/fluffy-bananas-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Add draft ERC-7572 interface
```
ernestognw, line 5:
```suggestion
`IERC7572`: Add interface to support contract-level metadata.
```

ernestognw (commented):
Hi @coffeexcoin,

If I remember correctly, this interface was discussed in the context of #5394. I think we would need to use the interface in 6909 if we approve this PR

coffeexcoin:
> Hi @coffeexcoin,
> 
> 
> 
> If I remember correctly, this interface was discussed in the context of #5394. I think we would need to use the interface in 6909 if we approve this PR

Happy to make those changes - what would be the preferred method, just make 6909ContentURI implement this interface?

coffeexcoin:
Also happy to add erc20/721/1155 extensions that implement this if that would be desirable

ernestognw:
I'd keep it only for 6909 for now. Other standards like ERC721 and ERC1155 define their own metadata functions. Probably that's why we decided not to include this interface in the first place

cc @arr00

arr00:
@ernestognw do you think it's ok to add an event to an interface defined in an ERC? `ERC6909` doesn't define the `ContractURIUpdated` event. I think it should (maybe we should open a PR), but we shouldn't add it unilaterally.

coffee-the-dev:
Hey @arr00 @Amxx @ernestognw I seem to have some conflicting guidance here - can you provide some clarity on which direction you would like me to go in with regards to changes to ERC-6909?

arr00:
> Hey @arr00 @Amxx @ernestognw I seem to have some conflicting guidance here - can you provide some clarity on which direction you would like me to go in with regards to changes to ERC-6909?

Given that `contractURI()` is specified by ERC6909 but `ContractURIUpdated()` is not, I don't think there is any role for ERC7572 to play here.

Amxx:
Note: if we do this change, then `type(IERC6909ContentURI).interfaceId` will change, as it will no longer include `contractURI()` in its computation.

ERC165 is not part of ERC6909 specification, but I still this this should be considered.

Amxx:
Given that ERC-6909 does NOT mention ERC-7572 ... and that ERC-7572 does NOT mention ERC-7572 either ... I think we shouldn't do this change. 

Closing this PR. If there are new elements, we can reopen it, or continue this discussion in an issue (following our guidelines)

PR #5687, opened 2025-05-15
docs: document the non-zero delay enforcement

### LINKED ISSUE #5666: AccessControlDefaultAdminRules scheduled delay is configured in a non-standard manner
In [`AccessControlDefaultAdminRules.sol`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9393147db73ae6261b67cb03003370e9a7fa2448/contracts/access/extensions/AccessControlDefaultAdminRules.sol#L369), you can't execute on inclusion block + scheduled delay but rather, inclusion block + scheduled delay + 1. This is a rather strange pattern since time limits are usually enforced as `block.timestamp >= expectedTimestamp`. I presume this is intended to induce a delay when `schedule timestamp == block.timestamp` but has been left undocumented.

**💻 Environment**
v5.3.0 OZ
v1 Foundry

**📝 Details**
I think this should be documented or fixed. I don't strictly view it as a bug but it's a definite pitfall.

**🔢 Code to reproduce bug**
* `multicall([...], [...])`
  * `beginDefaultAdminTransfer(...)`
  * `acceptDefaultAdminTransfer()`

QEDK:
Fixes #5666

* Adds a documentation comment to expose the non-zero delay requirement for the default admin role transfer.

#### PR Checklist

- [X] Tests
- [X] Documentation

### DIFF
--- a/contracts/access/extensions/AccessControlDefaultAdminRules.sol
+++ b/contracts/access/extensions/AccessControlDefaultAdminRules.sol
@@ -24,6 +24,7 @@ import {IERC165} from "../../utils/introspection/ERC165.sol";
  * * Enforces a 2-step process to transfer the `DEFAULT_ADMIN_ROLE` to another account.
  * * Enforces a configurable delay between the two steps, with the ability to cancel before the transfer is accepted.
  * * The delay can be changed by scheduling, see {changeDefaultAdminDelay}.
+ * * There must be a non-zero delay between the default admin role's transfer and acceptance.
  * * It is not possible to use another role to manage the `DEFAULT_ADMIN_ROLE`.
  *
  * Example usage:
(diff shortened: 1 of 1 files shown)


contracts/access/extensions/AccessControlDefaultAdminRules.sol
```diff
@@ -24,6 +24,7 @@ import {IERC165} from "../../utils/introspection/ERC165.sol";
  * * Enforces a 2-step process to transfer the `DEFAULT_ADMIN_ROLE` to another account.
  * * Enforces a configurable delay between the two steps, with the ability to cancel before the transfer is accepted.
  * * The delay can be changed by scheduling, see {changeDefaultAdminDelay}.
+ * * There must be a non-zero delay between the default admin role's transfer and acceptance.
```
ernestognw, line 27: What about?

```suggestion
 * * Role transfers must wait at least one block after scheduling before it can be accepted.
```

ernestognw (approved):
Will require a second review, but LGTM

Issue #5681, opened 2025-05-11
Storage Slot Collisions in Diamond-Compatible Delegatecall Architecture (Upgradeable Contracts)


EricForgy:
**🧐 Motivation**
OpenZeppelin’s upgradeable contracts are widely used in standard proxy-based deployments. However, when used in a **modular delegatecall architecture** (such as the Diamond Standard or similar plugin-style routers), **fixed storage slot constants across multiple modules can cause critical storage collisions.**

Each module is executed in the same storage context (the proxy/Router), so repeated use of hardcoded slot identifiers like `INITIALIZABLE_STORAGE` leads to overlapping writes between otherwise isolated modules. This makes current upgradeable contracts **unsafe** in modular `delegatecall` setups — an increasingly common pattern in modern contract systems.

---

**📝 Details**
I propose a backward-compatible enhancement to support diamond-compatible usage by **deriving storage slot positions uniquely per module** based on their deployed address.

This can be done using the following pattern:

```solidity
address private immutable __self = address(this);

bytes32 private immutable INITIALIZABLE_STORAGE =
    keccak256(
        abi.encode(
            uint256(keccak256(abi.encodePacked(__self, "openzeppelin.storage.Initializable"))) - 1
        )
    ) & ~bytes32(uint256(0xff));
```

### Why this works:

* **Safe**: Each module gets a deterministic, unique slot namespace.
* **Aligned**: The bitmask aligns with EIP-1967 slot layout expectations.
* **Minimal Overhead**: Only requires switching some `pure` functions to `view` due to `__self`, with no storage reads and negligible gas impact.
* **Compatible**: Initialization checks are typically only called once, so runtime cost is not a concern.

### Suggested Implementation

Introduce a new base contract or modifier version (e.g. `DiamondCompatibleInitializable`) that mirrors `Initializable` but uses per-module slot derivation. This could live in an experimental or `diamond` subdirectory to avoid impacting the existing suite.

---

Let me know if you want this as a full PR — happy to help.

EricForgy:
**Follow-up: Slot Override via Virtual Function (Partial Support)**

After further review, I noticed that `Initializable` defines the storage slot via a **virtual function**:

```solidity
    function _initializableStorageSlot() internal view virtual returns (bytes32);
```

This is extremely helpful, as it allows consumers to **override the slot location** safely — enabling compatibility with delegatecall-based architectures like Diamond without forking the entire contract.

[...]

PR #5688, opened 2025-05-16
Add _reentrancyGuardStorageSlot()

EricForgy:
Fixes https://github.com/OpenZeppelin/openzeppelin-contracts/issues/5681 for ReentrancyGuardTransient.

This PR introduces a pure virtual function `_reentrancyGuardStorageSlot()` in `ReentrancyGuardTransient.sol`, mirroring the pattern used in the upgradeable [Initializable.sol](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/proxy/utils/Initializable.sol)

### Motivation

As discussed in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/5681, storage slot collisions can occur when using OpenZeppelin contracts in delegatecall-based modular systems (e.g., the Diamond Standard). While this issue was originally raised in the context of upgradeable contracts, `ReentrancyGuardTransient.sol` in the non-upgradeable repo is in fact **already upgrade-safe** due to its use of a named storage slot.

Adding `_reentrancyGuardStorageSlot()` as a `pure virtual` function allows advanced users to override the default slot location safely in derived contracts, which is essential when using multiple delegatecall modules that each instantiate `ReentrancyGuardTransient`.

### Summary of Changes

* Introduced `_reentrancyGuardStorageSlot()`, marked `internal pure virtual`
* Replaced all direct uses of `REENTRANCY_GUARD_STORAGE` with calls to this new function
* No behavioral or storage layout changes; fully backward-compatible

### DIFF
--- a/contracts/utils/ReentrancyGuardTransient.sol
+++ b/contracts/utils/ReentrancyGuardTransient.sol
@@ -44,11 +44,11 @@ abstract contract ReentrancyGuardTransient {
         }
 
         // Any calls to nonReentrant after this point will fail
-        REENTRANCY_GUARD_STORAGE.asBoolean().tstore(true);
+        _reentrancyGuardStorageSlot().asBoolean().tstore(true);
     }
 
     function _nonReentrantAfter() private {
-        REENTRANCY_GUARD_STORAGE.asBoolean().tstore(false);
+        _reentrancyGuardStorageSlot().asBoolean().tstore(false);
     }
 
     /**
@@ -56,6 +56,10 @@ abstract contract ReentrancyGuardTransient {
      * `nonReentrant` function in the call stack.
      */
     function _reentrancyGuardEntered() internal view returns (bool) {
-        return REENTRANCY_GUARD_STORAGE.asBoolean().tload();
+        return _reentrancyGuardStorageSlot().asBoolean().tload();
+    }
+
+    function _reentrancyGuardStorageSlot() internal pure virtual returns (bytes32) {
+        return REENTRANCY_GUARD_STORAGE;
     }
 }


gonzaotc (approved):

gonzaotc:
Approved, I consider allowing storage slot overridability while keeping the formula `pure` a good tradeoff between flexibility and security, as mentioned in https://github.com/OpenZeppelin/openzeppelin-contracts/issues/5681#issuecomment-2870284233. Additionally, we will work on the transpiler to ensure this same pattern is correctly applied across the upgradeable contracts.

contracts/utils/ReentrancyGuardTransient.sol
```diff
@@ -44,18 +44,22 @@ abstract contract ReentrancyGuardTransient {
...
     /**
      * @dev Returns true if the reentrancy guard is currently set to "entered", which indicates there is a
      * `nonReentrant` function in the call stack.
      */
     function _reentrancyGuardEntered() internal view returns (bool) {
-        return REENTRANCY_GUARD_STORAGE.asBoolean().tload();
+        return _reentrancyGuardStorageSlot().asBoolean().tload();
+    }
+
+    function _reentrancyGuardStorageSlot() internal pure virtual returns (bytes32) {
```
Amxx, line 62: Should that be
- `_reentrancyGuardStorageSlot`
- `_reentrancyGuardTransientSlot`
- `_reentrancyGuardTransientStorageSlot`

Here, the contract only uses transient, but what if a contract combined transient and normal storage ? Would we want the two types to use the same slot, or would we allow separating the two ?

Amxx:
Some notes:
- we agreed on keeping the `_reentrancyGuardStorageSlot()` name
- for some stupid reason, `ReentrancyGuardTransientUpgradeable` is not seen as stateless, and there is an Upgradeable version of it. We should remove it from 6.0
- there is also an upgradeable version of `ReentrancyGuard` that uses namespace storage. If we apply these changes to the "non-transient" version, it would work, but it would also be strange for the upgradeable version. Do we want that ?

Amxx:
Closed in favor of #5892 (cannot modify this PR's code)

PR #5689, opened 2025-05-16
docs: update EIP712 upgradeable version comment

### LINKED ISSUE #5607: docs: misleading outdated comment at EIP712Upgradeable
The EIP712Upgradeable contract seems to have misleading / outdated documentation in the comments.

**💻 Environment**

Latest (5.2.0)

**📝 Details**

At https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/cryptography/EIP712Upgradeable.sol

```solidity
/**
 ...
 * NOTE: In the upgradeable version of this contract, the cached values will correspond to the address, and the domain
 * separator of the implementation contract. This will cause the {_domainSeparatorV4} function to always rebuild the
 * separator from the immutable values, which is cheaper than accessing a cached version in cold storage.
 */
```

There are no use of immutables or constants in the contract. This note should go away.

**🔢 Code to reproduce bug**

Not relevant

crStiv:
Fixes #5607

Replaces outdated comment about the upgradeable version in EIP712 with an accurate explanation, as discussed in the issue.

### DIFF
--- a/contracts/utils/cryptography/EIP712.sol
+++ b/contracts/utils/cryptography/EIP712.sol
@@ -25,9 +25,8 @@ import {IERC5267} from "../../interfaces/IERC5267.sol";
  * NOTE: This contract implements the version of the encoding known as "v4", as implemented by the JSON RPC method
  * https://docs.metamask.io/guide/signing-data.html[`eth_signTypedDataV4` in MetaMask].
  *
- * NOTE: In the upgradeable version of this contract, the cached values will correspond to the address, and the domain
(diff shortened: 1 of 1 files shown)


ernestognw:
This should be fixed in the [upgradeable patch](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/df878c87fcd6bebfc7ffc9f9023134de281d7729/scripts/upgradeable/upgradeable.patch). See [docs](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/51294b7480fd13e716207a621ac1d55a6290d56d/scripts/upgradeable/README.md)

PR #5690, opened 2025-05-17
Update EIP712 upgradeable patch file

### LINKED ISSUE #5607: docs: misleading outdated comment at EIP712Upgradeable

naevern:
This PR fixes issue #5607 by updating the misleading comment in the EIP712 contract that incorrectly refers to "immutable values" in the upgradeable version.

As suggested by @Amxx, I've replaced the comment with:
> NOTE: The upgradeable version of this contract does not use an immutable cache and recomputes the domain separator each time {_domainSeparatorV4} is called. That is cheaper than accessing a cached version in cold storage.

This change is made in the upgradeable.patch file which is used during the transpilation process to generate the upgradeable contract variants, as recommended by @ernestognw.

Fixes #5607

#### PR Checklist

- [x] Tests
- [x] Documentation


naevern:
@ernestognw I've updated the comments as suggested by @Amxx.
Please take a look at this, if I need to change anything, do let me know. Thanks.

.changeset/smart-teams-raise.md
```diff
@@ -0,0 +1,2 @@
+---
```
CoveMB, line 1: Missing changeset

naevern: added.

naevern:
@CoveMB any more changes required? or is this good to go?

.changeset/smart-teams-raise.md
ernestognw: No, this shouldn't get a changeset, sorry. It's not a change that users perceive.

ernestognw (approved):
LGTM, but requires another review. CC @arr00 @luiz-lvj @gonzaotc

luiz-lvj (approved):

ernestognw:
Thanks @naevern!

PR #5691, opened 2025-05-18
Update Address.sol

### LINKED ISSUE #5685: Gas optimization in `Address.sol` impacting `functionCall()`
## Motivation
I recently audited a protocol that made an external call to a contract by using `functionCall()` from the `Address.sol` library. While auditing, I realized there is a significant gas saving that could affect all protocols using `functionCall()`, so I wanted to make a pull-request to address it.  

## Impact
- With this optimization, all calls to `functionCallWithValue()` that pass `value=0` would save 100 gas. 
- Notably, all contract calls using `Address.functionCall()` would benefit from this gas optimization as it calls `functionCallWithValue()` with `value=0` hardcoded.

## Details

The function `functionCall()` calls `functionCallWithValue()` passing `uint256 value=0`:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e34251682bac9c3252af30e91e999f13dd098b9f/contracts/utils/Address.sol#L62-L63

Then inside `functionCallWithValue()`, the library checks that `address(this).balance` is sufficient to cover for the `value` passed. Since `functionCall()` always calls with `value=0`, this will always be true, and the requirement will always pass:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e34251682bac9c3252af30e91e999f13dd098b9f/contracts/utils/Address.sol#L75-L80

However, reading an `address.balance` is a gas-expensive operation (100 gas according to https://www.evm.codes/), so it is inefficient to read the balance when the passed `value` is zero. A  more gas-efficient implementation of `functionCallWithValue()` would only read the `address.balance` when the requested `value` is greater than zero:

```diff
    function functionCallWithValue(address target, bytes memory data, uint256 value) internal returns (bytes memory) {
-       if (address(this).balance < value) {
+       if ((value > 0) && (address(this).balance < value)) {
            revert Errors.InsufficientBalance(address(this).balance, value);
        }
        (bool success, bytes memory returndata) = target.call{value: value}(data);
        return verifyCallResultFromTarget(target, success, returndata);
    }
```

In the above, the EVM compiler is smart enough to know that the `address.balance` opcode is only used when the first if-condition is met. Therefore, any calls to `functionCallWithValue()` where `value==0` would not read the balance unnecessarily.

MKVEERENDRA:
---
name: Gas optimization: skip balance check when value == 0 in Address.functionCallWithValue
about: Improve gas efficiency in Address.sol by avoiding unnecessary balance reads
title: Gas optimization: skip balance check when value == 0 in Address.functionCallWithValue
labels: optimization
assignees: ''

---

## Description

This PR optimizes the `functionCallWithValue` function in `Address.sol` by skipping the balance check when the `value` parameter is zero. Reading `address(this).balance` is a costly operation (~100 gas), and since calls with `value == 0` never require this check to fail, this saves gas on all such calls, including those made through `functionCall`.

This optimization benefits all contracts and protocols that rely on OpenZeppelin's `Address.functionCall` for external contract calls without sending ETH.

Fixes #5685

## Changes

- Updated `functionCallWithValue` to conditionally check balance only if `value > 0`.

```solidity
function functionCallWithValue(address target, bytes memory data, uint256 value) internal returns (bytes memory) {
    if (value > 0 && address(this).balance < value) {
        revert Errors.InsufficientBalance(address(this).balance, value);
    }
    (bool success, bytes memory returndata) = target.call{value: value}(data);
    return verifyCallResultFromTarget(target, success, returndata);
}

### DIFF
--- a/.changeset/rare-otters-work.md
+++ b/.changeset/rare-otters-work.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': patch
+---
+
+Gas optimization: skip balance check when value == 0 in Address.functionCallWithValue
--- a/contracts/utils/Address.sol
+++ b/contracts/utils/Address.sol
@@ -73,7 +73,7 @@ library Address {
      * - the called Solidity function must be `payable`.
      */
     function functionCallWithValue(address target, bytes memory data, uint256 value) internal returns (bytes memory) {
-        if (address(this).balance < value) {
+        if (value > 0 && address(this).balance < value) {
             revert Errors.InsufficientBalance(address(this).balance, value);
         }
         (bool success, bytes memory returndata) = target.call{value: value}(data);


ernestognw:
Although I agree with #5685, I don't think this is a long-term solution and I have a strong opinion this should be solved at the API level rather than in code.

Closing. See https://github.com/OpenZeppelin/openzeppelin-contracts/issues/5685#issuecomment-2889073220

PR #5694, opened 2025-05-22
Update AccessControl.sol

raymondabiola:
In the renouncRole() function, It needs to also check if the caller already has been assigned a role previously and also send them a custom error message if this is not the case

Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ernestognw:
Hey @raymondabiola, if you take a look, the internal `_revokeRole` already behaves as a noop if the role wasn't revoked. On the opposite, reverting disables users from overriding this function.

Issue #5454, opened 2025-01-23
Enhance BeaconProxy Tests for Robustness and Coverage


techvoyagerX:
**🧐 Motivation**  
The current test for `BeaconProxy` lacks [specific improvement or addition you're proposing, e.g., "more robust error handling" or "additional coverage for edge cases"]. Enhancing this will ensure better reliability and coverage of the `BeaconProxy` functionality.

**📝 Details**  
I propose the following changes:  
1. [Briefly describe the key update you're making, e.g., "Refactor the initialization tests to use more descriptive assertions."]  
2. [Add any improvements you're making, e.g., "Include a new test to handle scenarios with non-standard inputs."]  
3. [Explain how this change aligns with the existing code, e.g., "This maintains consistency with other tests by adhering to the same style and format."]

PR #5700, opened 2025-05-26
test: beacon returning invalid address should revert

Dozgor:
This test adds a scenario where the `BeaconProxy` is initialized with a beacon that returns an invalid (non-contract or null) implementation address.

Although the test causes Ethers.js to throw an error before reaching Solidity-level reverts, it helps highlight an edge case worth guarding against, as described in issue #5454.


Dozgor:
Thank you @Amxx for the clarification and for taking the time to explain.
I’m glad to know the case was already covered — it was a good exercise for me to explore Beacon behavior.
I’ll continue contributing and exploring the repo.
Cheers!

PR #5702, opened 2025-05-27
Refactor Blockhash lib

Amxx:
Looking at #5642, I noticed that the if statement was quite complexe. It comes with an implicit expectation that the static call (right part of the if) is executed BEFORE the returndatasize (left part of the if). While that expectation might be correct today, I'm not sure the compiler garantees it will stay that way.

Considering that the if does not trigger a revert if the staticcall fails, I believe the proposed version is just as effective, and IMO easier to read. All tests (unit and fuzzing) pass just fine.

Given that an MSTORE is just 3 gas, this version is also cheaper !

#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Blockhash.sol
+++ b/contracts/utils/Blockhash.sol
@@ -32,7 +32,7 @@ library Blockhash {
             distance = current - blockNumber;
         }
 
-        return distance > 256 && distance < 8192 ? _historyStorageCall(blockNumber) : blockhash(blockNumber);
+        return distance < 257 ? blockhash(blockNumber) : _historyStorageCall(blockNumber);
     }
 
     /// @dev Internal function to query the EIP-2935 history storage contract.


Amxx:
Note: I was also thinking that doing 
```solidity 
return distance < 256 ? blockhash(blockNumber) : _historyStorageCall(blockNumber)
```

Would be more future proof is the history is ever extended beyound 8192 blocks. That passes unit tests, but not fuzzing. Do we know why ?

contracts/utils/Blockhash.sol
```diff
@@ -32,19 +32,21 @@ library Blockhash {
...
+            // Store the blockNumber in scratch space
+            mstore(0x00, blockNumber)
+            mstore(0x20, 0)
 
-            // In case the history storage address is not deployed, the call will succeed
-            // without returndata, so the hash will be 0 just as querying `blockhash` directly.
-            if and(gt(returndatasize(), 0), staticcall(gas(), HISTORY_STORAGE_ADDRESS, 0, 0x20, 0, 0x20)) {
-                hash := mload(0)
-            }
+            // call history storage contract
```
Amxx, line 45:
```suggestion
            // call history storage address
```

arr00:
> Note: I was also thinking that doing
> 
> ```solidity
> return distance < 256 ? blockhash(blockNumber) : _historyStorageCall(blockNumber)
> ```
> 
> Would be more future proof is the history is ever extended beyound 8192 blocks. That passes unit tests, but not fuzzing. Do we know why ?

Should be 
```
return distance <= 256 ? blockhash(blockNumber) : _historyStorageCall(blockNumber)
```

arr00:
> Looking at #5642, I noticed that the if statement was quite complexe. It comes with an implicit expectation that the static call (right part of the if) is executed BEFORE the returndatasize (left part of the if). While that expectation might be correct today, I'm not sure the compiler garantees it will stay that way.

@ernestognw and I discussed this previously and found that this yul behaviour is pretty well documented and not just an implementation detail (documented [here](https://docs.soliditylang.org/en/latest/yul.html#function-calls)). Will defer to ernesto on this.

PR #5709, opened 2025-06-01
Add PairMap, NibbleMap and UintXMap to Bitmaps

ernestognw:
Fixes #???? 

#### PR Checklist

- [x] Tests
- [x] Documentation


Amxx:
I'm wondering if we should have multiple datastructures, versus having just one structure (bitmap) with functionallity to read//write multiple bits at once.

What is the usecase exactly ?

PR #5716, opened 2025-06-04
Increase SignatureChecker minimum pragma to 0.8.24

ernestognw:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Breaking changes
+
+- Update minimum pragma to 0.8.24 in `SignatureChecker`, `Governor` and Governor's extensions. ([#5716](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5716)).
+
 ## 5.3.0 (2025-04-09)
 
 ### Breaking Changes


Amxx:
My understanding is:
- SignatureChecker doesn't need 0.8.24 now
- SignatureChecker will need 0.8.24 when the ERC-7913 validation gets added
- This PR is done ahead of the ERC-7913 addition to avoid the ERC-7913 PR touching too many files?

If this is all correct, then I approve.

Amxx (approved):

ernestognw:
> My understanding is:
> 
> * SignatureChecker doesn't need 0.8.24 now
> * SignatureChecker will need 0.8.24 when the ERC-7913 validation gets added
> * This PR is done ahead of the ERC-7913 addition to avoid the ERC-7913 PR touching too many files?
> 
> If this is all correct, then I approve.

Yes, this is correct. Basically moving 7913Utils to SIgnatureChecker forced this pragma increase.

> Maybe we want to add a "breaking change" entry to the Changelog saying that some contract compiler requirement have evolved, and that users may need to upgrade the compiler version they are using.

Also true, I added a small note I think should be good enough

Amxx (approved):

PR #5722, opened 2025-06-07
Add ERC7803Utils and ERC7803

ernestognw:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/cryptography/ERC7803Utils.sol
+++ b/contracts/utils/cryptography/ERC7803Utils.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.24;
 
 import {MessageHashUtils} from "./MessageHashUtils.sol";
 import {Strings} from "../Strings.sol";
--- a/contracts/utils/cryptography/ERC7803Utils.sol
+++ b/contracts/utils/cryptography/ERC7803Utils.sol
@@ -89,8 +89,7 @@ library ERC7803Utils {
         }
 
         assembly ("memory-safe") {
-            // Can't overflow because `end` is less than `data.length`
-            mstore(data, sub(mload(data), sub(end, start)))
+            mstore(data, sub(end, start)) // Reset the length of the array. Can't overflow because `end` is less than `data.length`.
         }
 
         return data;
--- a/contracts/utils/cryptography/ERC7803Utils.sol
+++ b/contracts/utils/cryptography/ERC7803Utils.sol
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {MessageHashUtils} from "./MessageHashUtils.sol";
+import {Strings} from "../Strings.sol";
+import {Bytes} from "../Bytes.sol";
+import {Math} from "../math/Math.sol";
+
+/**
+ * @dev Utilities to process https://ercs.ethereum.org/ERCS/erc-7803[ERC-7803] signatures with signing domains
+ * for Account Abstraction.
+ *
+ * This library provides methods to encode and decode ERC-7803 signatures that support signing domains
+ * and authentication method coordination. Signing domains prevent replay attacks when private keys
+ * are shared across smart contract accounts, while authentication methods allow dapps and wallets
+ * to coordinate on signature verification approaches.
+ *
+ * The core functionality implements the recursive encoding scheme defined in ERC-7803:
+ *
+ * * If signing domain separators exist: `"\x19\x02" ‖ first ‖ encodeForSigningDomains(others, verifyingDomainSeparator, message)`
+ * * If no signing domain separators: standard EIP-712 encoding
+ */
+library ERC7803Utils {
+    /**
+     * @dev Encodes message for signing domains according to ERC-7803 specification.
+     *
+     * This implements the recursive encoding scheme:
+     *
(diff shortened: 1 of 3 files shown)


contracts/utils/cryptography/ERC7803Utils.sol
```diff
@@ -0,0 +1,97 @@
...
+            signingDomainSeparators.length == 0
+                ? MessageHashUtils.toTypedDataHash(verifyingDomainSeparator, structHash)
+                : MessageHashUtils.toSigningDomainHash(
+                    signingDomainSeparators[0],
+                    // TODO: Make iterative?
+                    encodeForSigningDomains(
+                        _splice(signingDomainSeparators, 1, signingDomainSeparators.length),
+                        verifyingDomainSeparator,
+                        structHash
+                    )
```
Amxx, line 48: I'm not a fan of the recurtion here, particularly because it require a _splice that is memory intensive. A simple for loop should work just fine
```suggestion
        bytes32 result = MessageHashUtils.toTypedDataHash(verifyingDomainSeparator, structHash);
        for (uint256 i = signingDomainSeparators.length; i > 0; --i) {
            result = MessageHashUtils.toSigningDomainHash(signingDomainSeparators[i-1], result);
        }
        return result;
```

contracts/utils/cryptography/ERC7803Utils.sol
```diff
@@ -0,0 +1,97 @@
...
+                );
```
Amxx, line 49: I'm not a fan of the recurtion here, particularly because it require a _splice that is memory intensive. A simple for loop should work just fine
```suggestion
        bytes32 result = MessageHashUtils.toTypedDataHash(verifyingDomainSeparator, structHash);
        for (uint256 i = signingDomainSeparators.length; i > 0; --i) {
            result = MessageHashUtils.toSigningDomainHash(signingDomainSeparators[i-1], result);
        }
        return result;
```

contracts/utils/cryptography/ERC7803Utils.sol
```diff
@@ -0,0 +1,97 @@
...
+        bytes memory methodBytes = bytes(methodId);
+        if (methodBytes.length < 4) return (false, 0);
+
+        // Check if it starts with "ERC-"
+        if (!(methodBytes[0] == "E" && methodBytes[1] == "R" && methodBytes[2] == "C" && methodBytes[3] == "-")) {
+            return (false, 0);
+        }
+
+        // Extract and validate the ERC number
+        return Strings.tryParseUint(string(Bytes.slice(methodBytes, 4)));
```
Amxx, line 68:
```suggestion
        return (methodBytes.length > 4 && bytes4(methodBytes) == 0x4552432d)
            ? Strings.tryParseUint(string(Bytes.slice(methodBytes, 4)))
            : (false, 0);
```

contracts/utils/cryptography/ERC7803Utils.sol
```diff
@@ -0,0 +1,97 @@
...
+                        _splice(signingDomainSeparators, 1, signingDomainSeparators.length),
+                        verifyingDomainSeparator,
+                        structHash
+                    )
+                );
+    }
+
+    /// @dev Checks if an authentication method ID corresponds to ECDSA.
+    function isECDSA(string memory methodId) internal pure returns (bool) {
+        return Strings.equal(methodId, "ECDSA");
```
Amxx, line 54: When we know the string, and when the length is <=32, it is cheaper (but arguably less readable) to compare things directly

```suggestion
        return bytes(methodId).length == 5 && bytes5(bytes(methodId)) == 0x4543445341;
```

contracts/utils/cryptography/signers/ERC7803.sol
```diff
@@ -0,0 +1,71 @@
...
+     * 3. Apply authentication methods in the specified order
+     */
+    function isValidSignature(bytes32 hash, bytes calldata signature) public view virtual returns (bytes4 result) {
+        // For the hash `0x7803780378037803780378037803780378037803780378037803780378037803` and an empty signature,
+        // we return the magic value `0x78030001` as it's assumed impossible to find a preimage for it that can be used
+        // maliciously. Useful for simulation purposes and to validate whether the contract supports ERC-7803.
+        return
+            (_isValidERC7803Signature(hash, signature) || _rawSignatureValidation(hash, signature))
+                ? IERC1271.isValidSignature.selector
+                : (hash == 0x7803780378037803780378037803780378037803780378037803780378037803 && signature.length == 0)
```
Amxx, line 48: AFAIK, ERC-7803 doesn't document anything like this

Amxx (commented):
- `isECDSA` and `isERC` are not (currently) used. I'm not sure we really need them.
- the signature decoding is not standardized in the ERC document. We should probably ask the ERC to standardized that part. Note that the bytes32[] could be packed encoded.

One thing I'm concerned with is that passing the `signingDomainSeparators` through the signature is probably not the right way to proceed. We need some sort of verification that the domain separator of the actual contract was used. I would expect the implementation to not add anything in the signature (as opposed to ERC-7739). 

On the "onchain" side, ERC-7803 would be an abstract signer layer that would "only" do something like:

```solidity
abstract contract ERC7803 is AbstractSigner, EIP712 {
    using MessageHashUtils from bytes32;

    function _rawSignatureValidation(bytes32 hash, bytes calldata signature) internal view virtual override returns (bool) {
        return super._rawSignatureValidation(_domainSeparatorV4().toSigningDomainHash(hash), signature);
    }
```

Amxx:
As discussed yesterday, closing this PR.

PR #5724, opened 2025-06-09
Add  operations to Math.sol

ernestognw:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -741,6 +741,58 @@ library Math {
     }
 
     /**
+     * @dev Reverses the byte order of a uint256 value, converting between little-endian and big-endian.
+     * Inspired in https://graphics.stanford.edu/~seander/bithacks.html#ReverseParallel[Reverse Parallel]
+     */
+    function reverseBitsUint256(uint256 value) internal pure returns (uint256) {
+        value = // swap bytes
+            ((value >> 8) & 0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF) |
+            ((value & 0x00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF00FF) << 8);
+        value = // swap 2-byte long pairs
+            ((value >> 16) & 0x0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF) |
+            ((value & 0x0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF0000FFFF) << 16);
+        value = // swap 4-byte long pairs
+            ((value >> 32) & 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF) |
+            ((value & 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF) << 32);
+        value = // swap 8-byte long pairs
+            ((value >> 64) & 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF) |
+            ((value & 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF) << 64);
+        return (value >> 128) | (value << 128); // swap 16-byte long pairs
+    }
+
+    /// @dev Same as {reverseBitsUint256} but optimized for 128-bit values.
+    function reverseBitsUint128(uint128 value) internal pure returns (uint256) {
+        value = // swap bytes
+            ((value & 0xFF00FF00FF00FF00FF00FF00FF00FF00) >> 8) |
+            ((value & 0x00FF00FF00FF00FF00FF00FF00FF00FF) << 8);
+        value = // swap 2-byte long pairs
+            ((value & 0xFFFF0000FFFF0000FFFF0000FFFF0000) >> 16) |
+            ((value & 0x0000FFFF0000FFFF0000FFFF0000FFFF) << 16);
+        value = // swap 4-byte long pairs
+            ((value & 0xFFFFFFFF00000000FFFFFFFF00000000) >> 32) |
+            ((value & 0x00000000FFFFFFFF00000000FFFFFFFF) << 32);
+        return (value >> 64) | (value << 64); // swap 8-byte long pairs
+    }
+
+    /// @dev Same as {reverseBitsUint256} but optimized for 64-bit values.
+    function reverseBitsUint64(uint64 value) internal pure returns (uint256) {
+        value = ((value & 0xFF00FF00FF00FF00) >> 8) | ((value & 0x00FF00FF00FF00FF) << 8); // swap bytes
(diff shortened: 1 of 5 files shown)


contracts/utils/math/Math.sol
```diff
@@ -740,6 +740,58 @@ library Math {
...
+    }
+
+    /// @dev Same as {reverseBitsUint256} but optimized for 32-bit values.
+    function reverseBitsUint32(uint32 value) internal pure returns (uint256) {
+        value = ((value & 0xFF00FF00) >> 8) | ((value & 0x00FF00FF) << 8); // swap bytes
+        return (value >> 16) | (value << 16); // swap 2-byte long pairs
+    }
+
+    /// @dev Same as {reverseBitsUint256} but optimized for 16-bit values.
+    function reverseBits16(uint16 value) internal pure returns (uint256) {
```
Amxx: For consistency, this function should be named
```suggestion
    function reverseBitsUint16(uint16 value) internal pure returns (uint256) {
```

However, I'd like to open the discussion on the name, and emplacement, of these reverse function.
- instead of naming them `reverseBitsUint256`, ... `reverseBitsUint16`, I would consider naming them just `reverseBits32` ... `reverseBits2`
- have these function and take bytesXx inputs instead of UintXxx inputs. Implementation would be the same, and bytesXx can easily be casted from/to uints

test/utils/math/Math.test.js
```diff
@@ -710,4 +711,118 @@ describe('Math', function () {
...
+
+  describe('reverseBits', function () {
+    describe('reverseBitsUint256', function () {
+      it('reverses bytes correctly', async function () {
+        await expect(this.mock.$reverseBitsUint256(0)).to.eventually.equal(0n);
+        await expect(this.mock.$reverseBitsUint256(ethers.MaxUint256)).to.eventually.equal(ethers.MaxUint256);
+
+        // Test simple pattern
+        await expect(
+          this.mock.$reverseBitsUint256('0x0000000000000000000000000000000000000000000000000000000000000001'),
```
Amxx: IMO the uint256 and uint128 test pattern should contain more data than "all zero with a one". I like the uint64 test case

contracts/utils/math/Math.sol
```diff
@@ -740,6 +740,58 @@ library Math {
...
+            ((value >> 32) & 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF) |
+            ((value & 0x00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF00000000FFFFFFFF) << 32);
+        value = // swap 8-byte long pairs
+            ((value >> 64) & 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF) |
+            ((value & 0x0000000000000000FFFFFFFFFFFFFFFF0000000000000000FFFFFFFFFFFFFFFF) << 64);
+        return (value >> 128) | (value << 128); // swap 16-byte long pairs
+    }
+
+    /// @dev Same as {reverseBitsUint256} but optimized for 128-bit values.
+    function reverseBitsUint128(uint128 value) internal pure returns (uint256) {
```
Amxx: What is the rational for returning uint256 and not uint128 here ? Same question for the other variants.

Amxx:
Should that be in Math, or should we start a `Byte.sol` file for this, and CLZ (and other) ?

It doesn't feel like this is a math operation strictly speaking.

test/utils/math/Math.t.sol
```diff
@@ -308,6 +308,27 @@ contract MathTest is Test {
         }
     }
 
+    // REVERSE BITS
+    function testSymbolicReverseBitsUint256(uint256 value) public pure {
+        assertEq(Math.reverseBitsUint256(Math.reverseBitsUint256(value)), value);
+    }
+
+    function testSymbolicReverseBitsUint128(uint128 value) public pure {
+        assertEq(Math.reverseBitsUint128(uint128(Math.reverseBitsUint128(value))), value);
```
Amxx: Ideally, we would also inject "dirty upper bits"

- get the value from fuzzing (clean)
- turn it into dirtyValue 
- check that `assertEq(Math.reverseBitsUint128(Math.reverseBitsUint128(dirtyValue)), value);`

contracts/utils/math/Math.sol
ernestognw: Good point. If I remember correctly I wasn't sure about some casting in the RLP library at first. Seems irrelevant now so I updated in 40d7922

contracts/utils/math/Math.sol
ernestognw: > For consistency, this function should be named

Yeah fair, I updated in 33857183a

> - instead of naming them reverseBitsUint256, ... reverseBitsUint16, I would consider naming them just reverseBits32 ... reverseBits2
> - have these function and take bytesXx inputs instead of UintXxx inputs.

Makes a lot of sense. Updated in 89860bc87

ernestognw:
> Should that be in Math, or should we start a `Byte.sol` file for this, and CLZ (and other) ?
> 
> It doesn't feel like this is a math operation strictly speaking.

After the review and changes made I think it makes sense to add it to the `Bytes.sol` library so I updated accordingly. wdyt?

contracts/utils/Bytes.sol
```diff
@@ -99,6 +99,58 @@ library Bytes {
         return result;
     }
 
+    /**
+     * @dev Reverses the byte order of a uint256 value, converting between little-endian and big-endian.
```
Amxx:
```suggestion
     * @dev Reverses the byte order of a bytes32 value, converting between little-endian and big-endian.
```

Amxx:
For the record, I tried seeing is assembly would save gas:
```solidity
    function reverseBits256(bytes32 value) internal pure returns (bytes32) {
        assembly ("memory-safe") {
            let mask
            // swap bytes
            mask := mul(div(not(0), 0xFFFF), 0xFF)
            value := or(and(shr(8, value), mask), shl(8, and(value, mask)))
            // swap 2-bytes long pairs
            mask := mul(div(not(0), 0xFFFFFFFF), 0xFFFF)
            value := or(and(shr(16, value), mask), shl(16, and(value, mask)))
            // swap 4-bytes long pairs
            mask := mul(div(not(0), 0xFFFFFFFFFFFFFFFF), 0xFFFFFFFF)
            value := or(and(shr(32, value), mask), shl(32, and(value, mask)))
            // swap 8-bytes long pairs
            mask := mul(div(not(0), 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF), 0xFFFFFFFFFFFFFFFF)
            value := or(and(shr(64, value), mask), shl(64, and(value, mask)))
            // swap 16-bytes long pairs
            value := or(shr(128, value), shl(128, value))
        }
        return value;
    }
```

It doesn't !

Amxx:
It all looks good to me, except the name of the functions. 

Lets take the example of `reverseBits16`. The name sugest it reverses the 16 bits that compose the object. So I would expect that it turn the binary `0b0000000000000001` (`0x0001`) into `0b1000000000000000` (`0x8000`) ... but the output is `0b0000000100000000` (`0x0100`).

what it reverses are not the inidividual bits, its the bytes. 

So IMO we should either:
- change the behavior to reverse at the bit level
- change the name from `reverseBits16` to `reverseBytes2`

PR #5725, opened 2025-06-09
Add new `equal`, `nibbles` and `clz` functions to Bytes.sol

ernestognw:
Fixes #???? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/Bytes.sol
+++ b/contracts/utils/Bytes.sol
@@ -99,6 +99,35 @@ library Bytes {
         return result;
     }
 
+    /// @dev Split each byte in `value` into two nibbles (4 bits each).
+    function nibbles(bytes memory value) internal pure returns (bytes memory) {
+        uint256 length = value.length;
+        bytes memory nibbles_ = new bytes(length * 2);
+        for (uint256 i = 0; i < length; i++) {
+            (nibbles_[i * 2], nibbles_[i * 2 + 1]) = (value[i] & 0xf0, value[i] & 0x0f);
+        }
+        return nibbles_;
+    }
+
+    /**
+     * @dev Returns true if the two byte buffers are equal.
+     */
+    function equal(bytes memory a, bytes memory b) internal pure returns (bool) {
+        return a.length == b.length && keccak256(a) == keccak256(b);
+    }
+
+    /// @dev Counts the number of leading zero bytes in a uint256.
+    function countLeadingZeroes(uint256 x) internal pure returns (uint256) {
+        if (x == 0) return 32; // All 32 bytes are zero
+        uint256 r = 0;
+        if (x > 0xffffffffffffffffffffffffffffffff) r = 128; // Upper 128 bits
+        if ((x >> r) > 0xffffffffffffffff) r |= 64; // Next 64 bits
+        if ((x >> r) > 0xffffffff) r |= 32; // Next 32 bits
+        if ((x >> r) > 0xffff) r |= 16; // Next 16 bits
+        if ((x >> r) > 0xff) r |= 8; // Next 8 bits
+        return 31 ^ (r >> 3); // Convert to leading zero bytes count
+    }
+
     /**
      * @dev Reads a bytes32 from a bytes array without bounds checking.
      *
--- a/contracts/utils/Bytes.sol
+++ b/contracts/utils/Bytes.sol
@@ -117,7 +117,7 @@ library Bytes {
     }
 
     /// @dev Counts the number of leading zero bytes in a uint256.
-    function countLeadingZeroes(uint256 x) internal pure returns (uint256) {
+    function clz(uint256 x) internal pure returns (uint256) {
         if (x == 0) return 32; // All 32 bytes are zero
         uint256 r = 0;
         if (x > 0xffffffffffffffffffffffffffffffff) r = 128; // Upper 128 bits
--- a/test/utils/Bytes.t.sol
+++ b/test/utils/Bytes.t.sol
@@ -60,7 +60,7 @@ contract BytesTest is Test {
     }
 
     function testSymbolicCountLeadingZeroes(uint256 x) public pure {
-        uint256 result = Bytes.countLeadingZeroes(x);
+        uint256 result = Bytes.clz(x);
         assertLe(result, 32); // [0, 32]
 
         if (x != 0) {
--- a/test/utils/Bytes.t.sol
+++ b/test/utils/Bytes.t.sol
@@ -1,0 +1,77 @@
+pragma solidity ^0.8.20;
+
+import {Test} from "forge-std/Test.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+import {Bytes} from "@openzeppelin/contracts/utils/Bytes.sol";
+
+contract BytesTest is Test {
+    function testIndexOf(bytes memory buffer, bytes1 s) public pure {
+        testIndexOf(buffer, s, 0);
+    }
+
+    function testIndexOf(bytes memory buffer, bytes1 s, uint256 pos) public pure {
+        uint256 result = Bytes.indexOf(buffer, s, pos);
+
+        // Should not be found before result
+        for (uint256 i = pos; result != type(uint256).max && i < result; i++) assertNotEq(buffer[i], s);
+        if (result != type(uint256).max) assertEq(buffer[result], s);
+    }
+
+    function testLastIndexOf(bytes memory buffer, bytes1 s) public pure {
+        testLastIndexOf(buffer, s, 0);
+    }
+
+    function testLastIndexOf(bytes memory buffer, bytes1 s, uint256 pos) public pure {
+        pos = bound(pos, 0, buffer.length);
+        uint256 result = Bytes.lastIndexOf(buffer, s, pos);
+
+        // Should not be found before result
+        for (uint256 i = pos; result != type(uint256).max && i < result; i++) assertNotEq(buffer[i], s);
+        if (result != type(uint256).max) assertEq(buffer[result], s);
+    }
+
+    function testSlice(bytes memory buffer, uint256 start) public pure {
+        testSlice(buffer, start, buffer.length);
+    }
+
+    function testSlice(bytes memory buffer, uint256 start, uint256 end) public pure {
+        bytes memory result = Bytes.slice(buffer, start, end);
+        uint256 sanitizedEnd = Math.min(end, buffer.length);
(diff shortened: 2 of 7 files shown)


contracts/utils/Bytes.sol
```diff
@@ -99,6 +99,35 @@ library Bytes {
...
+    function clz(uint256 x) internal pure returns (uint256) {
+        if (x == 0) return 32; // All 32 bytes are zero
+        uint256 r = 0;
+        if (x > 0xffffffffffffffffffffffffffffffff) r = 128; // Upper 128 bits
+        if ((x >> r) > 0xffffffffffffffff) r |= 64; // Next 64 bits
+        if ((x >> r) > 0xffffffff) r |= 32; // Next 32 bits
+        if ((x >> r) > 0xffff) r |= 16; // Next 16 bits
+        if ((x >> r) > 0xff) r |= 8; // Next 8 bits
+        return 31 ^ (r >> 3); // Convert to leading zero bytes count
+    }
```
Amxx: We could reuse the `Math.log256` function here
```suggestion
    function clz(uint256 x) internal pure returns (uint256) {
        return Math.ternary(x == 0, 32, 31 - Math.log256(x));
    }
```

Also, echoing on the `reverseByte` PR, but I think clz and reverseBytes should be in the same file (possibly a new one)

ernestognw: Updated. Agree that the `reverseBit` functions and `clz` should be in the same library. imo it's fine that we put them in `Bytes.sol`. Any strong reason not to do so?

test/utils/Bytes.t.sol
```diff
@@ -3,9 +3,80 @@
...
+    function testLastIndexOf(bytes memory buffer, bytes1 s) public pure {
+        testLastIndexOf(buffer, s, 0);
+    }
+
+    function testLastIndexOf(bytes memory buffer, bytes1 s, uint256 pos) public pure {
+        pos = bound(pos, 0, buffer.length);
+        uint256 result = Bytes.lastIndexOf(buffer, s, pos);
+
+        // Should not be found before result
+        for (uint256 i = pos; result != type(uint256).max && i < result; i++) assertNotEq(buffer[i], s);
```
Amxx, line 31:
```suggestion
        for (uint256 i = pos; i < Math.min(result, buffer.length); ++i) assertNotEq(buffer[i], s);
```

test/utils/Bytes.t.sol
```diff
@@ -3,9 +3,80 @@
...
 contract BytesTest is Test {
+    function testIndexOf(bytes memory buffer, bytes1 s) public pure {
+        testIndexOf(buffer, s, 0);
+    }
+
+    function testIndexOf(bytes memory buffer, bytes1 s, uint256 pos) public pure {
+        uint256 result = Bytes.indexOf(buffer, s, pos);
+
+        // Should not be found before result
+        for (uint256 i = pos; result != type(uint256).max && i < result; i++) assertNotEq(buffer[i], s);
```
Amxx, line 18:
```suggestion
        for (uint256 i = pos; i < Math.min(result, buffer.length); ++i) assertNotEq(buffer[i], s);
```

test/utils/Bytes.t.sol
Amxx: `lastIndexOf(buffer, s, pos)` will always return either :
- `type(uint256).max` if the `s` is not in `buffer` (before pos)
- a position `result` that is smaller (or equal) to `pos`.

This loop thazt starts a pos and goes forward will never run.

contracts/utils/Bytes.sol
Amxx: IMO we should distinguish between `bytes` as in "arbitrary length buffer", and `Bytes32` as in "a type of fixed length that has math properties".

I would argue that clz is one of these "local" math property and is closer to the Math library then it is to things like `indexOf` or `slice` that are more "global". By that argument, the reverseBytesXX are also more of a "local" operation (similar to clz)

Amxx: If CLZ operates on `uint256`, I think there is a strong case for putting it in Math.

We could have a variant of CLZ in Bytes that does something like
```solidity
    function clz(bytes memory buffer) internal pure returns (uint256) {
        uint256 i = 0;
        while (i < buffer.length && buffer[i] != bytes1(0)) ++i;
        return 8 * i + (i == buffer.length ? 0 : Math.clz(buffer[i]));
    }
```
(should be optimized for reading in blocks of 32 instead of byte by byte

contracts/utils/Bytes.sol
```diff
@@ -99,6 +99,28 @@ library Bytes {
...
 
+    /// @dev Split each byte in `value` into two nibbles (4 bits each).
+    function nibbles(bytes memory value) internal pure returns (bytes memory) {
+        uint256 length = value.length;
+        bytes memory nibbles_ = new bytes(length * 2);
+        for (uint256 i = 0; i < length; i++) {
+            (nibbles_[i * 2], nibbles_[i * 2 + 1]) = (value[i] & 0xf0, value[i] & 0x0f);
+        }
+        return nibbles_;
+    }
```
0xClandestine: This is pretty inefficient, consider using `unchecked` at minimum.

Also unclear why this is useful:

```solidity
➜ nibbles(hex"ABCD")
Type: dynamic bytes
├ Hex (Memory):
├─ Length ([0x00:0x20]): 0x0000000000000000000000000000000000000000000000000000000000000004
├─ Contents ([0x20:..]): 0xa00bc00d00000000000000000000000000000000000000000000000000000000
├ Hex (Tuple Encoded):
├─ Pointer ([0x00:0x20]): 0x0000000000000000000000000000000000000000000000000000000000000020
├─ Length ([0x20:0x40]): 0x0000000000000000000000000000000000000000000000000000000000000004
└─ Contents ([0x40:..]): 0xa00bc00d00000000000000000000000000000000000000000000000000000000
➜ 
```

0xClandestine: Using `unchecked` would be safe since overflow is impossible given `bytes memory` can realistically only have a length smaller than 256 bits.

```solidity
➜ bytes memory b = new bytes(type(uint32).max);
Traces:
  [942682544] 0xBd770416a3345F91E4B34576cb804a576fa48EB1::run()
    └─ ← [MemoryOOG] EvmError: MemoryOOG
```

ernestognw: I'd fix tests first and then optimize, it's likely that we may leverage other functions of the Bytes library

ernestognw: The initial motivation for this function is to use it in https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5680 for TrieProof, but I agree that reallocating memory is perhaps not the most efficient thing to do. I suspect TrieProof may not require the `nibbles()` function if these are read in place, but I need to experiment a bit with that.

We can add the unchecked, but, is there an alternative you were thinking of? I am more worried about the memory expansion cost

0xClandestine: Wondering if inspecting the nibbles JIT rather than converting to separate nibbles array is worthwhile to explore.

0xClandestine: In regard to memory expansion, we need calldata variants for all methods related to MPT proofs (both RLP decoding and MPT branch verification) since the input data is almost always provided by the user via calldata. There's rarely a reason to copy the full RLP payload into memory, because typical use cases (like verifying a historical blockhash) only require extracting one or two fields (e.g. stateRoot, txRoot). Operating directly on calldata avoids unnecessary memory allocation and is significantly more gas-efficient. Comparing my personal implementation to `RLPReader` I'm saving about 40k gas when parsing every block header field (which should never really be needed but serves as a good gas comparison and unit test).

0xClandestine: On the nibbles method it may make sense to first check the size of the input, if less than or equal to 32 bytes the above calculation could be much simpler (no loop needed)

contracts/utils/Bytes.sol
```diff
@@ -187,6 +200,32 @@ library Bytes {
...
+            unchecked {
+                // Mask out bytes beyond buffer length
+                // left is buffer.length at most, can't overflow at realistic size
+                uint256 left = buffer.length - i;
+                if (left < 32) {
+                    // left is less than 32, can't overflow
+                    uint256 shift = (32 - left) * 8;
+                    value = (value >> shift) << shift; // Clear the lower bits for the last iteration
+                }
+            }
```
Amxx: Is that part necessary ? If we read beyound the buffer limit there are 2 possible cases:
- the value is 0, we don't get into the `if (value !=0)`, loop doesn't go any further, we get the default value.
- the value is 0, we get into the `if(value != 0)`, and the `Math.min` makes sure we don't count zeros after the buffer.

Amxx:
```suggestion
```

test/utils/Bytes.t.sol
```diff
@@ -196,6 +209,47 @@ contract BytesTest is Test {
...
+        uint256 manualCount = 0;
+        for (uint256 i = 0; i < buffer.length; ++i) {
+            if (buffer[i] != 0) {
+                break;
+            }
+            manualCount++;
+        }
+
+        assertEq(result, manualCount);
+    }
```
Amxx: I don't see what this tests brings that is not already covered by `testClz`

contracts/utils/math/Math.sol
```diff
@@ -746,4 +746,11 @@ library Math {
     function unsignedRoundsUp(Rounding rounding) internal pure returns (bool) {
         return uint8(rounding) % 2 == 1;
     }
+
+    /**
+     * @dev Counts the number of leading zeros bits in a uint256.
```
ernestognw:
```suggestion
     * @dev Counts the number of leading zero bits in a uint256.
```

contracts/utils/Bytes.sol
```diff
@@ -187,6 +187,20 @@ library Bytes {
         return (value >> 8) | (value << 8);
     }
 
+    /**
+     * @dev Counts the number of leading zeros in bits a bytes array. Returns `8 * buffer.length`
```
Amxx:
```suggestion
     * @dev Counts the number of leading zero bits a bytes array. Returns `8 * buffer.length`
```

contracts/utils/Bytes.sol
ernestognw: Thanks for your feedback @0xClandestine!

We discussed it internally and we agreed to remove the `nibbles` function for now since it doesn't seem to be useful outside of the context of MPT. I'll reimplement the function privately in MPT

PR #5727, opened 2025-06-09
Minimize Solidity pragma versions for interface compatibility

arr00:
Takes just the pragma changes from https://github.com/arr00/openzeppelin-contracts/pull/2


Amxx:
This is going to conflict with #5728. We should do this one first, and do #5728 after !

PR #5728, opened 2025-06-09
Add draft- prefix to ERC-7579 and ERC-7739 related files

Amxx:


docs/modules/ROOT/pages/backwards-compatibility.adoc
```diff
@@ -13,7 +13,9 @@ Infrequently a patch or minor update will remove or change an API in a breaking
 
 === Draft or Pre-Final ERCs
 
-ERCs that are not Final can change in incompatible ways. For this reason, we avoid shipping implementations of ERCs before they are Final. Some exceptions are made for ERCs that have been published for a long time and that seem unlikely to change. Breaking changes to the ERC specification are still technically possible in those cases, so these implementations are published in files named `draft-*.sol` to make that condition explicit.
+ERCs that are not Final can change in incompatible ways. For this reason, we avoid shipping implementations of ERCs before they are Final. Some exceptions are made for ERCs that have been published for a long time and seem unlikely to change. When breaking changes to the ERC specification remain technically possible, these implementations are published in files named `draft-*.sol` to make that condition explicit.
```
ernestognw: I think this change is worth for clarifying when exceptions to the `draft-` prefix are made

ernestognw (dismissed):

docs/modules/ROOT/pages/backwards-compatibility.adoc
arr00:
```suggestion
ERCs that are not Final can change in incompatible ways. For this reason, we avoid shipping implementations of ERCs before they are Final. Some exceptions are made for ERCs that have been published for a long time and seem unlikely to change. Implementations for ERCs that may have breaking changes are published in files named `draft-*.sol` to make that condition explicit. There is no backwards compatibility guarantee for content in files prefixed with `draft`.
```

PR #5730, opened 2025-06-09
Update pragma check: validate by actually running the compiler

Amxx:
This script replaces the old pragma consistency check with a new pragma validity check.

The old check looked at all the pragma, and verified that the pragma each file are at least as restrictive as the pragma of the file it imports. This enshure the transitivity of restriction. However, this old script did not validate that the pragma was restrictive enough. In the old check it was possible to specify pragma that are too lose, for example using `^0.8.20` in a file that use `mcopy`, and that doesn't import anything, was not raising any flag.

This new check goes further, by actually running the oldest compile version that the pragma allows. For example, if a file specified `^0.8.20`, this new version of the script will try compiling the file using `0.8.20`. If that compilation is not successfull, the check will fail (and CI will be informed)

This PR does NOT include a script to automatically build the pragmas.

---

Note: metadata extraction, and solidity version/compiler binding, will be usefull for the script that automatically builds the pragmas, and was put in separate helpers.

---

It is possible to check a particular file/folder by doing (-v is for verbose, optional)
```bash
yarn test:pragma -p 'contracts/governance/Governor.sol' -v
yarn test:pragma -p 'contracts/token/ERC20/**/*.sol' -v
yarn test:pragma -p 'contracts/token/*.sol' -v
```


Amxx:
We need a plimit on the conmpiler runs

scripts/checks/pragma-validity.js
```diff
@@ -0,0 +1,45 @@
...
+const { compile } = require('../solc-versions');
+
+const {
+  argv: { pattern, skipPatterns, verbose, concurrency, _: artifacts },
+} = yargs(hideBin(process.argv))
+  .env('')
+  .options({
+    pattern: { alias: 'p', type: 'string', default: 'contracts/**/*.sol' },
+    skipPatterns: { alias: 's', type: 'string', default: 'contracts/mocks/**/*.sol' },
+    concurrency: { alias: 'c', type: 'number', default: 4 },
```
Amxx:
```suggestion
    concurrency: { alias: 'c', type: 'number', default: 8 },
```

.github/workflows/checks.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         run: npm run test
       - name: Check linearisation of the inheritance graph
         run: npm run test:inheritance
-      - name: Check pragma consistency between files
+      - name: Check pragma validity
         run: npm run test:pragma
```
Amxx:
```suggestion
        run: npm run test:pragma -- --concurrency 4
```

arr00:
Pragma check seems to be flaky. I would prefer to remove concurrency if we are at risk of having a flaky check. Otherwise looks great. Thanks @Amxx !

.github/workflows/checks.yml
```diff
@@ -41,8 +41,8 @@ jobs:
         run: npm run test
       - name: Check linearisation of the inheritance graph
         run: npm run test:inheritance
-      - name: Check pragma consistency between files
-        run: npm run test:pragma
+      - name: Check pragma validity
+        run: npm run test:pragma -- --concurrency 4
```
Amxx:
```suggestion
        run: npm run test:pragma -- --concurrency 1
```

arr00 (approved):

scripts/checks/pragma-validity.js
```diff
@@ -0,0 +1,45 @@
...
+Promise.all(
+  Object.entries(getContractsMetadata(pattern, skipPatterns, artifacts)).map(([source, { pragma }]) =>
+    limit(
+      (file, version) =>
+        compile(file, version).then(
+          () => {
+            verbose && console.log(`Compile ${file} using solc ${version}: ok`);
+          },
+          error => {
+            console.log(`Failed to compile ${file} using solc ${version}\n${error}`);
```
Amxx:
```suggestion
            console.error(`Failed to compile ${file} using solc ${version}\n${error}`);
```

arr00 (approved):

scripts/checks/pragma-validity.js
```diff
@@ -0,0 +1,45 @@
...
+  argv: { pattern, skipPatterns, verbose, concurrency, _: artifacts },
+} = yargs(hideBin(process.argv))
+  .env('')
+  .options({
+    pattern: { alias: 'p', type: 'string', default: 'contracts/**/*.sol' },
+    skipPatterns: { alias: 's', type: 'string', default: 'contracts/mocks/**/*.sol' },
+    concurrency: { alias: 'c', type: 'number', default: 8 },
+    verbose: { alias: 'v', type: 'count' },
+  });
+
```
ernestognw, line 22: The script is no longer checking for pragma consistency.

Previously, a contract declared (for example) as 0.8.20 but that depends on `Bytes.sol` that uses 0.8.24 would fail the check because it should always be compiled with 0.8.24 and above. So the contract must be published as 0.8.24 at minimum.

Would the following make sense?

```suggestion
const metadata = getContractsMetadata(pattern, skipPatterns, artifacts);

// Check pragma compatibility between imports
for (const [source, { pragma, sources }] of Object.entries(metadata)) {
  if (!pragma) continue;
  const minVersion = semver.minVersion(pragma);
  for (const importPath of sources) {
    const importMetadata = metadata[importPath];
    if (!importMetadata?.pragma) continue;
    if (!semver.satisfies(minVersion, importMetadata.pragma)) {
      console.error(`${source} uses ${pragma} but depends on ${importPath} that requires ${importMetadata.pragma}`);
      process.exitCode = 1;
    }
  }
}
```

ernestognw (commented):
I think it looks good but my main concern is that we're not checking

scripts/checks/pragma-validity.js
Amxx: > we're not checking pragma consistency anymore

I think we do, just not ourselves, we let the compiler do it.

If contract A declares ^0.8.20 and imports contract B that declares ^0.8.24, this script will try to compile A with 0.8.20. This will fail, because the compiler imports a file that has an incompatible pragma.

ernestognw: Oh pretty clever, that's true. Thanks!

PR #5733, opened 2025-06-10
Add Bytes.splice, an inplace variant of Buffer.slice

### LINKED ISSUE #5723: Consider adding `slice` and `splice` function to `Arrays` and `Bytes`
**🧐 Motivation**

Currently, the `Bytes` library includes a function to `slice` a chunk of data. I think this is useful, but for cases when the original array can be discarded, I think developers would benefit from implementing an `splice` function.

We could consider these functions for the `Arrays` library too, so we can offer a generalized library for doing this with any native solidity type.

**📝 Details**

Here's a reference implementation for the Arrays library for `bytes32[] memory` arrays:

```solidity
/**
 * @dev Splices a slice of a bytes32 array. Avoids expanding memory.
 *
 * Replicates https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice[Javascript's `Array.splice`]
 *
 * NOTE: Clears the array if `start` is greater than `end`.
 */
function _splice(bytes32[] memory data, uint256 start, uint256 end) private pure returns (bytes32[] memory) {
    // sanitize
    uint256 length = data.length;
    end = Math.min(end, length);
    start = Math.min(start, end);

    if (start != 0) {
        // allocate and copy
        for (uint256 i = start; i < end; i++) {
            data[i - start] = data[i];
        }
    }

    assembly ("memory-safe") {
        mstore(data, sub(end, start)) // Reset the length of the array. Can't overflow because `end` is less than `data.length`.
    }

    return data;
}
```

Amxx:
Partially fixes #5723 (for the Bytes) case

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/afraid-chicken-attack.md
+++ b/.changeset/afraid-chicken-attack.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Bytes`: Add `splice(bytes,uint256)` and `splice(bytes,uint256,uint256)`, two "in place" variants of the existing slice functions
--- a/contracts/utils/Bytes.sol
+++ b/contracts/utils/Bytes.sol
@@ -100,6 +100,35 @@ library Bytes {
     }
 
     /**
+     * @dev In place slice: moves the content of `buffer`, from `start` (included) to the end of `buffer` to the start of that buffer.
+     *
+     * NOTE: This function modifies the provided buffer in place. If you need to preserve the original buffer, use {slice} instead
+     */
+    function splice(bytes memory buffer, uint256 start) internal pure returns (bytes memory) {
+        return splice(buffer, start, buffer.length);
+    }
+
+    /**
+     * @dev In place slice: moves the content of `buffer`, from `start` (included) to end (excluded) to the start of that buffer.
+     *
+     * NOTE: This function modifies the provided buffer in place. If you need to preserve the original buffer, use {slice} instead
+     */
+    function splice(bytes memory buffer, uint256 start, uint256 end) internal pure returns (bytes memory) {
+        // sanitize
+        uint256 length = buffer.length;
+        end = Math.min(end, length);
+        start = Math.min(start, end);
+
+        // allocate and copy
+        assembly ("memory-safe") {
+            mcopy(add(buffer, 0x20), add(add(buffer, 0x20), start), sub(end, start))
+            mstore(buffer, sub(end, start))
+        }
+
+        return buffer;
+    }
+
+    /**
      * @dev Reads a bytes32 from a bytes array without bounds checking.
      *
      * NOTE: making this function internal would mean it could be used with memory unsafe offset, and marking the
--- a/test/utils/Bytes.test.js
+++ b/test/utils/Bytes.test.js
@@ -56,8 +56,8 @@ describe('Bytes', function () {
     });
   });
 
-  describe('slice', function () {
-    describe('slice(bytes, uint256)', function () {
+  describe('slice & splice', function () {
+    describe('slice(bytes, uint256) & splice(bytes, uint256)', function () {
       for (const [descr, start] of Object.entries({
         'start = 0': 0,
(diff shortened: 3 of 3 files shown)


Amxx:
Would be usefull for #5762

contracts/utils/Bytes.sol
```diff
@@ -99,6 +99,35 @@ library Bytes {
...
+     * @dev In place slice: moves the content of `buffer`, from `start` (included) to the end of `buffer` to the start of that buffer.
+     *
+     * NOTE: This function modifies the provided buffer in place. If you need to preserve the original buffer, use {slice} instead
+     */
+    function splice(bytes memory buffer, uint256 start) internal pure returns (bytes memory) {
+        return splice(buffer, start, buffer.length);
+    }
+
+    /**
+     * @dev In place slice: moves the content of `buffer`, from `start` (included) to end (excluded) to the start of that buffer.
```
james-toussaint:
```suggestion
     * @dev Moves the content of `buffer`, from `start` (included) to the end of `buffer` to the start of that buffer.
     *
     * NOTE: This function modifies the provided buffer in place. If you need to preserve the original buffer, use {slice} instead
     */
    function splice(bytes memory buffer, uint256 start) internal pure returns (bytes memory) {
        return splice(buffer, start, buffer.length);
    }

    /**
     * @dev Moves the content of `buffer`, from `start` (included) to end (excluded) to the start of that buffer.
```
?

james-toussaint (approved):

contracts/utils/Bytes.sol
```diff
@@ -99,6 +99,35 @@ library Bytes {
         return result;
     }
 
+    /**
+     * @dev Moves the content of `buffer`, from `start` (included) to the end of `buffer` to the start of that buffer.
+     *
```
ernestognw, line 104: I would be clear about the following

```suggestion
     * The `end` argument is truncated to the length of the `buffer`.
     *
```

contracts/utils/Bytes.sol
```diff
@@ -99,6 +99,35 @@ library Bytes {
...
+     * NOTE: This function modifies the provided buffer in place. If you need to preserve the original buffer, use {slice} instead
+     */
+    function splice(bytes memory buffer, uint256 start) internal pure returns (bytes memory) {
+        return splice(buffer, start, buffer.length);
+    }
+
+    /**
+     * @dev Moves the content of `buffer`, from `start` (included) to end (excluded) to the start of that buffer.
+     *
```
ernestognw, line 114: Same

```suggestion
     * The `end` argument is truncated to the length of the `buffer`.
     *
```

.changeset/afraid-chicken-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+`Bytes`: Add `splice(bytes,uint256)` and `splice(bytes,uint256,uint256)`, two "in place" variants of the existing slice functions
```
ernestognw:
```suggestion
`Bytes`: Add `splice(bytes,uint256)` and `splice(bytes,uint256,uint256)` functions that move a specified range of bytes to the start of the buffer and truncate it in place, as an alternative to `slice`.
```

ernestognw (approved):

test/utils/Bytes.t.sol
```diff
@@ -0,0 +1,87 @@
...
+        // Verify content matches
+        for (uint256 i = 0; i < result.length; i++) {
+            assertEq(result[i], buffer[start + i]);
+        }
+
+        // Original buffer should remain unchanged
+        assertEq(buffer.length, buffer.length);
+        for (uint256 i = 0; i < buffer.length; i++) {
+            assertEq(buffer[i], buffer[i]);
+        }
```
Amxx: This is not checking anything.

ernestognw (approved):

PR #5735, opened 2025-06-11
ERC20Bridgable (ERC-7802)

### LINKED ISSUE #5734: ERC20Bridgeable
Move https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Bridgeable.sol from the community repo to here. Having it in the main repo means an upgradeable version will be available.

File should be prefixed with `draft-`.

Amxx:
Fixes #5734

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/interfaces/draft-IERC7802.sol
+++ b/contracts/interfaces/draft-IERC7802.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.20;
+pragma solidity >=0.6.2;
 
 import {IERC165} from "./IERC165.sol";
 
--- a/contracts/interfaces/draft-IERC7802.sol
+++ b/contracts/interfaces/draft-IERC7802.sol
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {IERC165} from "./IERC165.sol";
+
+/// @title IERC7802
+/// @notice Defines the interface for crosschain ERC20 transfers.
+interface IERC7802 is IERC165 {
+    /// @notice Emitted when a crosschain transfer mints tokens.
+    /// @param to       Address of the account tokens are being minted for.
+    /// @param amount   Amount of tokens minted.
+    /// @param sender   Address of the caller (msg.sender) who invoked crosschainMint.
+    event CrosschainMint(address indexed to, uint256 amount, address indexed sender);
+
+    /// @notice Emitted when a crosschain transfer burns tokens.
+    /// @param from     Address of the account tokens are being burned from.
+    /// @param amount   Amount of tokens burned.
+    /// @param sender   Address of the caller (msg.sender) who invoked crosschainBurn.
+    event CrosschainBurn(address indexed from, uint256 amount, address indexed sender);
+
+    /// @notice Mint tokens through a crosschain transfer.
+    /// @param _to     Address to mint tokens to.
+    /// @param _amount Amount of tokens to mint.
+    function crosschainMint(address _to, uint256 _amount) external;
+
+    /// @notice Burn tokens through a crosschain transfer.
+    /// @param _from   Address to burn tokens from.
+    /// @param _amount Amount of tokens to burn.
+    function crosschainBurn(address _from, uint256 _amount) external;
+}
--- a/contracts/token/ERC20/extensions/draft-ERC20Bridgeable.sol
+++ b/contracts/token/ERC20/extensions/draft-ERC20Bridgeable.sol
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {ERC20} from "../ERC20.sol";
+import {ERC165, IERC165} from "../../../utils/introspection/ERC165.sol";
+import {IERC7802} from "../../../interfaces/draft-IERC7802.sol";
+
+/**
+ * @dev ERC20 extension that implements the standard token interface according to
+ * https://eips.ethereum.org/EIPS/eip-7802[ERC-7802].
+ *
+ * NOTE: To implement a crosschain gateway for a chain, consider using an implementation if {IERC7786} token
+ * bridge (e.g. {AxelarGatewaySource}, {AxelarGatewayDestination}).
+ */
+abstract contract ERC20Bridgeable is ERC20, ERC165, IERC7802 {
+    /// @dev Modifier to restrict access to the token bridge.
+    modifier onlyTokenBridge() {
+        // Token bridge should never be impersonated using a relayer/forwarder. Using msg.sender is preferable to
+        // _msgSender() here, both for cost and security reasons.
+        _checkTokenBridge(msg.sender);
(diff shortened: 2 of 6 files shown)


test/token/ERC20/extensions/draft-ERC20Bridgeable.test.js
ernestognw, line 1: Is `draft-` needed in tests?

test/token/ERC20/extensions/draft-ERC20Bridgeable.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+    it('mints amount provided by the bridge when calling crosschainMint', async function () {
+      const amount = 100n;
+      await expect(this.token.connect(this.bridge).crosschainMint(this.other, amount))
+        .to.emit(this.token, 'CrosschainMint')
+        .withArgs(this.other, amount, this.bridge)
+        .to.emit(this.token, 'Transfer')
+        .withArgs(ethers.ZeroAddress, this.other, amount);
+
+      expect(await this.token.balanceOf(this.other)).to.equal(amount);
```
ernestognw: Let's use the `.eventually` syntax now that we're merging the file

ernestognw (dismissed):

test/token/ERC20/extensions/draft-ERC20Bridgeable.test.js
ernestognw: I see we're doing it for 4337 and 7579. Let's keep it

contracts/interfaces/draft-IERC7802.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import {IERC165} from "./IERC165.sol";
+
+/// @title IERC7802
+/// @notice Defines the interface for crosschain ERC20 transfers.
+interface IERC7802 is IERC165 {
+    /// @notice Emitted when a crosschain transfer mints tokens.
+    /// @param to       Address of the account tokens are being minted for.
+    /// @param amount   Amount of tokens minted.
+    /// @param sender   Address of the caller (msg.sender) who invoked crosschainMint.
```
arr00, line 12: Lots of non-dev tags here--won't be parsed for documentation.

contracts/token/ERC20/extensions/draft-ERC20Bridgeable.sol
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @dev ERC20 extension that implements the standard token interface according to
+ * https://eips.ethereum.org/EIPS/eip-7802[ERC-7802].
+ */
+abstract contract ERC20Bridgeable is ERC20, ERC165, IERC7802 {
+    /// @dev Modifier to restrict access to the token bridge.
+    modifier onlyTokenBridge() {
+        // Token bridge should never be impersonated using a relayer/forwarder. Using msg.sender is preferable to
+        // _msgSender() here, both for cost and security reasons.
```
arr00:
```suggestion
        // Token bridge should never be impersonated using a relayer/forwarder. Using msg.sender is preferable to
        // _msgSender() for cost and security reasons.
```

Don't we say the costs are 0? Shouldn't contradict ourselves. Also given that we plan on removing context, i dont think we need the comment at all.

Amxx: Costs of using _msgSender() are not zero if context is overridden.

Even if we have plan to remove it, the rule remains to use is everywhere ... unless we have a strong reason not to do so. In that case we should document the reason.

I think here there is a reason, and it's documented. Would not remove the documentation.

arr00: The documentation infers that __msgSender() costs more than msg.sender even if it’s not overridden.

contracts/interfaces/draft-IERC7802.sol
ernestognw: Yes, it's true we have been not too strict with that for interfaces and we try to keep them as defined in their corresponding ERCs (which was the case for 7802 as far as I remember). I would be in favor of rendering `@param` and `@returns` in the interfaces page only.

contracts/token/ERC20/extensions/draft-ERC20Bridgeable.sol
ernestognw: iirc, the concern here was that the bridge shouldn't be changed or impersonated in the first place, so adding `_msgSender` here implied that it could be overridden, which is completely unexpected in the context of a token that's trusting a bridge for minting and burning

Amxx: On the other hand:

- If the bridge is a contract, it should not be overridable using 2771 or similar

- I can imagine cases where the bridge is handled by an off-chain system with MPC or similar. In that case the mint function would come from an EOA (handled in MPC) and allowing relaying of these mint operations would be fine

So it can be argued that using _msgSender() is safe and eventually useful.

ernestognw: No reason to do this post 7702

EDIT: omg there are reasons to do so in chains pre-7702! 😭

ernestognw (dismissed):

ernestognw (dismissed):

contracts/token/ERC20/extensions/draft-ERC20Bridgeable.sol
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @dev See {IERC7802-crosschainMint}. Emits a {CrosschainMint} event.
+     */
+    function crosschainMint(address to, uint256 value) public virtual override onlyTokenBridge {
+        _mint(to, value);
+        emit CrosschainMint(to, value, _msgSender());
+    }
+
+    /**
+     * @dev See {IERC7802-crosschainBurn}. Emits a {CrosschainBurn} event.
```
arr00:
```suggestion
     * @dev See {IERC7802-crosschainBurn}. Emits a {IERC7802-CrosschainBurn} event.
```

Both are referencing the event (not the function). Not sure how to fix that but will look into it.

contracts/token/ERC20/extensions/draft-ERC20Bridgeable.sol
```diff
@@ -0,0 +1,50 @@
...
+        _;
+    }
+
+    /// @inheritdoc ERC165
+    function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165, IERC165) returns (bool) {
+        return interfaceId == type(IERC7802).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @dev See {IERC7802-crosschainMint}. Emits a {CrosschainMint} event.
```
arr00:
```suggestion
     * @dev See {IERC7802-crosschainMint}. Emits a {IERC7802-CrosschainMint} event.
```

Same issue with the event being referenced for both.

arr00 (approved):

ernestognw (approved):

PR #5736, opened 2025-06-11
ERC-7930

Amxx:
This is a machine-readable alternative to CAIP-2 and CAIP-10. It is used by ERC-7786

#### PR Checklist

- [x] Tests
- [x] Documentation


contracts/utils/draft-ERC7930.sol
```diff
@@ -0,0 +1,253 @@
...
+import {Math} from "./math/Math.sol";
+import {SafeCast} from "./math/SafeCast.sol";
+import {Bytes} from "./Bytes.sol";
+import {Calldata} from "./Calldata.sol";
+
+/**
+ * @dev Helper library to format and parse https://github.com/ethereum/ERCs/pull/1002[ERC-7930] interoperable
+ * addresses.
+ */
+library ERC7930 {
```
Amxx: Should we do 
```solidity
library ERC7930 {
```

or 
```solidity
library InteroperableAddresses {
```
?

0xtiti:
hey @Amxx - leaving the [Technical Specification: Solidity Library for Interoperable Addresses (ERC-7930)](https://github.com/ethereum/L2-interop/blob/main/specs/addresses/technical-specification-solidity-library-for-interoperable-addresses.md) here in case you haven't seen it already. Great to see work on top of 7930!

Amxx:
@0xtiti make sure you check out https://www.npmjs.com/package/interoperable-addresses

contracts/utils/draft-ERC7930.sol
luiz-lvj: I think it should be self explanatory to new developers, so:

```solidity
library InteroperableAddresses {
```

Amxx: Singular or plural ?

luiz-lvj: singular is better imo

```solidity
InteroperableAddress.parseV1(
```

contracts/mocks/Stateless.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.24;
+pragma solidity ^0.8.26;
```
ernestognw, line 3: Shouldn't be 25?

```suggestion
pragma solidity ^0.8.25;
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+
+pragma solidity ^0.8.26;
+
+import {Math} from "./math/Math.sol";
+import {SafeCast} from "./math/SafeCast.sol";
+import {Bytes} from "./Bytes.sol";
+import {Calldata} from "./Calldata.sol";
+
+/**
+ * @dev Helper library to format and parse https://github.com/ethereum/ERCs/pull/1002[ERC-7930] interoperable
```
ernestognw: If we know we'll get a link, we may be a good idea to just add it.

```suggestion
 * @dev Helper library to format and parse http://eips.ethereum.org/EIPS/eip-7930[ERC-7930] interoperable
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+                chainType,
+                chainReference.length.toUint8(),
+                chainReference,
+                addr.length.toUint8(),
+                addr
+            );
+    }
+
+    /**
+     * @dev Variant of {formatV1} specific to EVM chains. Returns the ERC-7930 interoperable address (version 1) for
```
ernestognw:
```suggestion
     * @dev Variant of {formatV1-bytes2-bytes-bytes-} specific to EVM chains. Returns the ERC-7930 interoperable address (version 1) for
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+     * a given chainid and ethereum address.
+     */
+    function formatEvmV1(uint256 chainid, address addr) internal pure returns (bytes memory) {
+        bytes memory chainReference = _toChainReference(chainid);
+        return abi.encodePacked(bytes4(0x00010000), uint8(chainReference.length), chainReference, uint8(20), addr);
+    }
+
+    /**
+     * @dev Variant of {formatEvmV1} that specifies an EVM chain without an address.
```
ernestognw:
```suggestion
     * @dev Variant of {formatV1-bytes2-bytes-bytes-} that specifies an EVM chain without an address.
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+     */
+    function formatEvmV1(uint256 chainid) internal pure returns (bytes memory) {
+        bytes memory chainReference = _toChainReference(chainid);
+        return abi.encodePacked(bytes4(0x00010000), uint8(chainReference.length), chainReference, uint8(0));
+    }
+
+    /**
+     * @dev Variant of {formatEvmV1} that specifies an EVM address without a chain reference.
```
ernestognw:
```suggestion
     * @dev Variant of {formatV1-bytes2-bytes-bytes-} that specifies an EVM address without a chain reference.
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+     */
+    function formatEvmV1(address addr) internal pure returns (bytes memory) {
+        return abi.encodePacked(bytes6(0x000100000014), addr);
+    }
+
+    /**
+     * @dev Parse a ERC-7930 interoperable address (version 1) into its different components.
+     *
+     * NOTE: This function will revert if the input is not following version 1 of ERC-7930.
+     */
```
ernestognw: I wouldn't use a note here

```suggestion
     * Reverts if the input is not following a version 1 of ERC-7930
     */
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+    ) internal pure returns (bytes2 chainType, bytes memory chainReference, bytes memory addr) {
+        bool success;
+        (success, chainType, chainReference, addr) = tryParseV1(self);
+        require(success, InteroperableAddressParsingError(self));
+    }
+
+    /**
+     * @dev Variant of {parseV1} that handles calldata slices to reduce memory copy costs.
+     *
+     * NOTE: This function will revert if the input is not following version 1 of ERC-7930.
```
ernestognw: Since we're already stating that it is a variant of {parseV1} and {parseV1} already has the note, does it make sense to keep it?

```suggestion
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+        (success, chainType, chainReference, addr) = tryParseV1Calldata(self);
+        require(success, InteroperableAddressParsingError(self));
+    }
+
+    /**
+     * @dev Parse a ERC-7930 interoperable address (version 1) into its different components.
+     *
+     * NOTE: If the input is not following version 1 of ERC-7930, this function will not revert, and instead will
+     * return a false boolean, indicating the parsing was a failure.
+     */
```
ernestognw: Simpler imo

```suggestion
    /**
     * @dev Variant of {parseV1} that does not revert on invalid input. Instead, it returns `false` as the first
     * return value to indicate parsing failure when the input does not follow version 1 of ERC-7930.
     */
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+
+    function _readBytes2Calldata(bytes calldata buffer, uint256 offset) private pure returns (bytes2 value) {
+        assembly ("memory-safe") {
+            value := shl(240, shr(240, calldataload(add(buffer.offset, offset))))
+        }
+    }
+
+    function _emptyBytesMemory() private pure returns (bytes memory result) {
+        assembly ("memory-safe") {
+            result := 0x60 // mload(0x60) is always 0
```
ernestognw, line 231: I may be confused with this one, but, should we load the word in `0x60` (i.e. the zero slot)?

```suggestion
            result := mload(0x60) // mload(0x60) is always 0
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+                return (false, 0x0000, _emptyBytesMemory(), _emptyBytesMemory());
+            addr = self.slice(0x06 + chainReferenceLength, 0x06 + chainReferenceLength + addrLength);
+        }
+    }
+
+    /**
+     * @dev Variant of {tryParseV1} that handles calldata slices to reduce memory copy costs.
+     *
+     * NOTE: If the input is not following version 1 of ERC-7930, this function will not revert, and instead will
+     * return a false boolean, indicating the parsing was a failure.
```
ernestognw:
```suggestion
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+
+    /**
+     * @dev Parse a ERC-7930 interoperable address (version 1) corresponding to an EIP-155 chain.
+     *
+     * NOTE: This function will revert if the input is not following version 1 of ERC-7930 or if the underlying
+     * chainType is not "eip-155". If the input doesn't include a chainReference, then 0 is returned as a chainId.
+     * Similarly if the input doesn't include an address, then address(0) is returned.
+     */
```
ernestognw: Maybe?

```suggestion
     /**
     * @dev Parse a ERC-7930 interoperable address (version 1) corresponding to an EIP-155 chain.
     * The `chainId` and `addr` return values will be zero if the input doesn't include a chainReference
     * or an address, respectively.
     *
     * Requirements:
     *
     * * The input must be a valid ERC-7930 interoperable address (version 1)
     * * The underlying chainType must be "eip-155"
     */
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+        require(success, InteroperableAddressParsingError(self));
+    }
+
+    /**
+     * @dev Variant of {parseEvmV1} that handles calldata slices to reduce memory copy costs.
+     *
+     * NOTE: This function will revert if the input is not following version 1 of ERC-7930 or if the underlying
+     * chainType is not "eip-155". If the input doesn't include a chainReference, then 0 is returned as a chainId.
+     * Similarly if the input doesn't include an address, then address(0) is returned.
+     */
```
ernestognw:
```suggestion
   /**
     * @dev Variant of {parseEvmV1} that handles calldata slices to reduce memory copy costs.
     */
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /**
+     * @dev Parse a ERC-7930 interoperable address (version 1) corresponding to an EIP-155 chain.
+     *
+     * NOTE: This function will not revert if the input is not following version 1 of ERC-7930 or if the underlying
+     * chainType is not "eip-155". Instead it return a false boolean, indicating the parsing was a failure. If the
+     * input doesn't include a chainReference, then 0 is returned as a chainId. Similarly if the input doesn't include
+     * an address, then address(0) is returned.
+     */
```
ernestognw:
```suggestion
    /**
     * @dev Variant of {parseEvmV1} that does not revert on invalid input. Instead, it returns `false` as the first
     * return value to indicate parsing failure when the input does not follow version 1 of ERC-7930.
     */
```

contracts/utils/draft-InteroperableAddress.sol
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /**
+     * @dev Variant of {tryParseEvmV1} that handles calldata slices to reduce memory copy costs.
+     *
+     * NOTE: This function will not revert if the input is not following version 1 of ERC-7930 or if the underlying
+     * chainType is not "eip-155". Instead it return a false boolean, indicating the parsing was a failure. If the
+     * input doesn't include a chainReference, then 0 is returned as a chainId. Similarly if the input doesn't include
+     * an address, then address(0) is returned.
+     */
```
ernestognw:
```suggestion
    /**
     * @dev Variant of {tryParseEvmV1} that handles calldata slices to reduce memory copy costs.
     */
```

ernestognw (commented):
Looks good on the technical side. Left comments about the docs because I feel too much duplication, though, it's also true that too many references may be indirect. So I wouldn't consider them blocking

contracts/utils/draft-InteroperableAddress.sol
Amxx: When don't know when that link will be good. Would you rather merge with a dead link that we hope will get live "soon", or use that PR link and update it later one?

I was going for the second option, but if you feel like the first is preferable, I'm open to discussion.

contracts/mocks/Stateless.sol
Amxx: `mocks/Stateless.sol` imports `utils/draft-InteroperableAddress.sol` with itself does `require(<condition>, <error>);`

This format of require was introduced in [0.8.26](https://github.com/ethereum/solidity/blob/develop/Changelog.md#0826-2024-05-21)

contracts/utils/draft-InteroperableAddress.sol
Amxx: The `bytes memory result` is a memory pointer. Basically its a `uint256` that correspond to a location somewhere in memory.

Usually, it points to a chunk of data that contains `[data.length, data[0], data[1], ..., data[length-1]]`. This chunk is usually allocated after the free memory pointer (and the fmp is moved to reserve it).

When we create a new `bytes` object of length n, we need to allocation `n+1` words: 1 for the length, and n for the data. the `bytes memory` object is a pointer toward the first slot, which contains the length. If we are allocating an empty buffer, we are still allocating 1 world in memory, where we store 0 (the length). That happens a the FMP location, and the FMP has to be move by 1 slot (to reserve the 1 slot we just used).

Now consider that an empty buffer is just represented by a slot that contains 0 (the length) ... and that we don't care about what comes after ... we can use any location in the memory that we know to store 0, and use it as an "anchor". As long as we don't write to it, we are fine.

The memory space is organised (at least in solidity) as follows:
- `0x00-0x3f`: scratch space (64 bytes)
- `0x40-0x5f`: free memory pointer
- `0x60-0x7f`: always zero
- `0x80-...`: memory space (initialy the fmp is `0x80`)

So when I need an empty buffer, I can just say that `0x60` is the location of my empty buffer. It will return length 0, and it has no data attached to it. We are fine using it as long as we don't start "expanding its length" (which would write to 0x60). Expanding the length of memory buffer is not something you should ever do anyway, because you don't know if you are not overriting stuff that goes after.

---

What you are proposing (`mload(0x60)`) would not work, at least not in general.
- we know that memory 0x60 at location is always zero, so what you are doing is `result := 0`
- that means you consider an array that start at location 0, which is the scratch space.
- if there is anything in the first 32 bytes of the scratch space (there is no reason why that would be 0), this is going to be you length. then you array will span over the rest of the scratch space, the fmp (at 0x40), the zero (at 0x60), and the rest of the memory.

Amxx: Note the comment 

> `mload(0x60)` is always zero

means "when we try to get the length of the buffer, we are always going to load 0, so this will always represent an empty buffer"

contracts/utils/draft-InteroperableAddress.sol
Amxx: An alternative would be to link to https://ethereum-magicians.org/t/erc-7930-interoperable-addresses/23365

ernestognw: It's true, but also our documentation won't get published until 5.5, right? In that case I think we can be more confident that it'll be merged (though not 100%)

I'll keep it to Ethereum magicians rather than to the PR

Amxx:
```suggestion
 * @dev Helper library to format and parse https://ethereum-magicians.org/t/erc-7930-interoperable-addresses/23365[ERC-7930] interoperable
```

contracts/mocks/Stateless.sol
ernestognw: Ah, got confused between 0.8.27 and 0.8.26. Makes sense

ernestognw (approved):

PR #5738, opened 2025-06-11
Fix pragma check for upgradeable test

arr00:


ernestognw:
Is the check necessary for upgradeable contracts? I would assume it's redundant
Genuinely curious @Amxx

Amxx:
The transpiler does add code (to handle the namespaces). It should not add more restriction on the pragma ... but it technically could.

PR #5740, opened 2025-06-12
Add script to automatically minimize pragma

Amxx:
Follow up to #5730, Replaces #5649

This adds a scripts that finds the minimal pragma that can be used for a given file.

Usage:
```bash
npm run pragma # minimize pragma for all solidity files in contracts/ (mocks excluded)
npm run pragma -- -p 'contracts/accounts/**/*.sol' # minimize pragma for a given subfolder
npm run pragma -- -p 'contracts/utils/Address.sol' # minimize pragma for a given file
```

Running this script is slow, and should probably not be run in CI. It should be a maintenance operation run every now and then. The pragma validity check (`yarn test:pragma`) already verifies that the pragma are valid.


arr00:
Why are we updating mocks? Aren't they skipped by default?

Amxx:
> Why are we updating mocks? Aren't they skipped by default?

They are, but just for fun I did run the script on the mocks, and it found two things. 

I said "why not" update that ... but if you prefer we can dismiss that.

---

The thing is, mocks include mocks/docs ... and maybe these should be processed.

arr00 (dismissed):

PR #5754, opened 2025-06-19
Fix upgradeable workflow by not running the linter on transpiled output

Amxx:
image

The transpile workflow currently fails because of some lint issue when commiting the result of transpilation. This is a temporary fix to re-establish the transpilation while we investigate the produced code.


Amxx:
Note: [this lint issue existed in the past](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/cdfb877d9cf59106e1615d38d104cea69e228142/contracts/mocks/StatelessUpgradeable.sol), but the linter was less restrictive ...

PR #5757, opened 2025-06-20
Add signer constructors

Amxx:
When the signer were part of the community repository, we did not have an upgradeable version. It made sens to not have a constructor, because we wanted to be able to use them in an "initializable" manner, without a key set at construction. This was still not upgrade-friendly because of the lack of namespaced storage.

This PR proposes to align the signers with the normal practice in the library:
- have a constructor that calls `_setSigner()`. This will force that the signer is set at construction, no longer possible to "forget calling it". This is fine for "one time deployment" (non upgradeable, non proxy) accounts.
- rely on the upgradeable version (produced by the transpiler) to have a version that use initializers and 7702 storage.

### checklist
- [x] Tests
- [ ] Documentation


contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -141,6 +149,7 @@ abstract contract MultiSignerERC7913 is AbstractSigner {
      * * See {_validateReachableThreshold} for the threshold validation.
      */
     function _setThreshold(uint64 newThreshold) internal virtual {
+        require(newThreshold > 0, MultiSignerERC7913NullThreshold());
```
ernestognw: Isn't this checked by `_validateReachableThreshold`?

Amxx: No. `_validateReachableThreshold` check that `signersLength >= currentThreshold`. If `currentThreshold` is 0 that passes just fine. See the test I added to `test/account/AccountMultiSigner.test.js`. Without this require the tx does not revert.

```suggestion
```

test/account/AccountMultiSigner.test.js
```diff
@@ -180,6 +180,12 @@ describe('AccountMultiSigner', function () {
         this.mock,
         'MultiSignerERC7913UnreachableThreshold',
       );
+
+      // No threshold would open the contract to anyone
+      await expect(this.mock.$_setThreshold(0)).to.revertedWithCustomError(
+        this.mock,
+        'MultiSignerERC7913NullThreshold',
+      );
```
Amxx:
```suggestion
```

contracts/utils/cryptography/signers/MultiSignerERC7913.sol
```diff
@@ -66,9 +66,17 @@ abstract contract MultiSignerERC7913 is AbstractSigner {
     /// @dev The `signer` is less than 20 bytes long.
     error MultiSignerERC7913InvalidSigner(bytes signer);
 
+    /// @dev The new threshold is 0.
+    error MultiSignerERC7913NullThreshold();
+
```
```suggestion
```

Amxx:
Note: the changeset is just for RC-1. Will not be part of the final release.

ernestognw (approved):

PR #5762, opened 2025-06-20
Add Base58 library

### LINKED ISSUE #5760: Consider adding a Base58 library
**🧐 Motivation**

Useful for encoding non-EVM addresses

**📝 Details**

Amxx:
Fixes #5760
Includes (depends on) #5733

This probably deserves some assembly. It feels like the decoding could be optimized a lot. Note that base58 is notoriously loop (and memory ops) heavy. 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

- **New Features**
  - On-chain Base58 encoding/decoding utility added.
  - Bytes utilities enhanced with in-place splice variants and helpers to count leading/consecutive bytes; splice parity with slicing ensured.

- **Documentation**
  - Base58 added to docs; Base64 descriptions clarified.

- **Tests**
  - Comprehensive Base58 tests (round-trip, vectors, invalid-input handling).
  - Expanded Bytes tests to cover new helpers and splice.
  - Updated Base64 tests with improved assertion style.


contracts/utils/Base58.sol
```diff
@@ -0,0 +1,135 @@
...
+                    _mstore8(binu, ptr, bytes1(uint8(outi[j] >> (8 * mask))));
+                    ptr++;
+                }
+                mask = 4;
+            }
+
+            uint256 dataCLZ = _countLeading(data, 0x31);
+            for (uint256 msb = dataCLZ; msb < binu.length; ++msb) {
+                if (_mload8(binu, msb) != 0x00) {
+                    return binu.slice(msb - dataCLZ, ptr);
```
Amxx: candidate for splice

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,135 @@
...
+                }
+            }
+            return binu.slice(0, ptr);
```
Amxx: candidate for splice

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,135 @@
...
+            bytes memory cache = bytes(_TABLE);
+            for (uint256 i = 0; i < resultLength; ++i) {
+                uint256 idx = _mload8i(slot, i + slotCLZ - dataCLZ);
+                bytes1 c = _mload8(cache, idx);
+                _mstore8(slot, i, c);
+            }
+
+            assembly ("memory-safe") {
+                mstore(slot, resultLength)
+            }
```
Amxx: here we could splice slot, and then map using the cache

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,133 @@
...
+ */
+library Base58 {
+    using SafeCast for bool;
+    using Bytes for bytes;
+
+    error InvalidBase56Digit(uint8);
+
+    bytes internal constant _TABLE = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+    bytes internal constant _LOOKUP_TABLE =
+        hex"000102030405060708ffffffffffffff090a0b0c0d0e0f10ff1112131415ff161718191a1b1c1d1e1f20ffffffffffff2122232425262728292a2bff2c2d2e2f30313233343536373839";
```
Amxx: For anyone curious, here is how you build these lookup tables:

```javascript
const { ethers } = require("ethers");

const max = (...values) => values.slice(1).reduce((x, y) => (x > y ? x : y), values.at(0));
const min = (...values) => values.slice(1).reduce((x, y) => (x < y ? x : y), values.at(0));

const buildLookup = (...tables) => {
    const bTables = tables.map(table => Array.from(ethers.toUtf8Bytes(table)));
    const MINIMUM = min(...bTables.flatMap(x => x));
    const MAXIMUM = max(...bTables.flatMap(x => x));
    const lookup = Uint8Array.from(Array.from({ length: MAXIMUM - MINIMUM + 1 }).map((_, i) => bTables.map(table => table.indexOf(i + MINIMUM)).find(x => x != -1) ?? 0xff));
    const valid = tables.every(table => Object.entries(table).every(([ i, c]) => i == lookup.at(c.charCodeAt(0) - MINIMUM)));
    return valid ? { tables, lookup: ethers.hexlify(lookup), MINIMUM, MAXIMUM } : undefined;
}

console.log(buildLookup(
    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/", // base64
    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_", // base64url
));

console.log(buildLookup(
    "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz", // base58
));
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,133 @@
...
+            bytes memory binu = new bytes(size);
+
+            bytes memory cache = _LOOKUP_TABLE;
+            uint256 outiLength = (b58Length + 3) / 4;
+            // Note: allocating uint32[] would be enough, but solidity doesn't pack memory.
+            uint256[] memory outi = new uint256[](outiLength);
+            for (uint256 i = 0; i < data.length; ++i) {
+                // get b58 char
+                uint8 chr = _mload8i(data, i);
+                require(chr > 48 && chr < 123, InvalidBase56Digit(chr));
```
Amxx: 48 and 123 are derived from the minimum and maximum values taken by b58 chars, see https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5762/files#r2160061084

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,184 @@
...
+            let slotLength := add(add(div(mul(sub(dataLength, dataLeadingZeros), 138), 100), dataLeadingZeros), 1)
+
+            // Zero the encoded buffer
+            for {
+                let i := 0
+            } lt(i, slotLength) {
+                i := add(i, 0x20)
+            } {
+                mstore(add(add(encoded, 0x20), i), 0)
+            }
```
Vectorized, line 80:
```suggestion
            calldatacopy(add(encoded, 0x20), calldatasize(), slotLength)
```

Amxx: solady's hacks making it into OZ, one at a time

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+        assembly ("memory-safe") {
+            // Count number of zero bytes at the beginning of `input`. These are encoded using the same number of '1's
+            // at then beginning of the encoded string.
+            let inputLeadingZeros := 0
+            for {} lt(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), lt(inputLeadingZeros, inputLength)) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 8351), 6115))
```
Amxx: FYI:
```8351 / 6115 = 1.365658217497956```
is an estimation of
```Math.log2(256) / Math.log2(58) = 1.365658237309761```

File: contracts/utils/Base58.sol [hunk 8]
```diff
@@ -0,0 +1,207 @@
...
+
+            // This is going to be our "scratch" workspace. Be leave enough room on the left to store length + encoded input.
```
Amxx:
```suggestion
            // This is going to be our "scratch" workspace. We leave enough room after FMP to later store length + encoded output.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+                // unfold first loop, with a different shift.
+                if i {
+                    mstore(ptr, shr(mul(sub(32, i), 8), mload(add(input, 0x20))))
+                    ptr := add(ptr, 0x20)
+                }
+            } lt(i, inputLength) {
+                ptr := add(ptr, 0x20) // next limb
+                i := add(i, 31) // move in buffer
+            } {
+                // Load 32 bytes from the input buffer and shift to only keep the 31 leftmost.
```
Amxx:
```suggestion
                // Load 31 bytes from the input buffer and store then in scratch (at ptr) in a dedicated 32 bytes space.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+
+            // This is going to be our "scratch" workspace. We leave enough room after FMP to later store length + encoded output.
+            let scratch := add(mload(0x40), add(outputLengthEstim, 0x21))
+
+            // Cut the input buffer in section (limbs) of 31 bytes (248 bits). Store in scratch.
+            let ptr := scratch
+            for {
+                // first section is possibly smaller than 31 bytes
+                let i := mod(inputLength, 31)
+                // unfold first loop, with a different shift.
```
Amxx:
```suggestion
                // If inputLength is not a multiple of 31, unfold the first loop:
                // Load the first i bytes from the input, right align them, and store them in scratch (at ptr) in a dedicated 32 bytes space.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+                mstore(ptr, shr(8, mload(add(add(input, 0x20), i))))
+            }
+
+            // Store the encoding table. This overlaps with the FMP that we are going to reset later anyway.
+            // See sections 2 of https://inputtracker.ietf.org/doc/html/draft-msporny-base58-03
+            mstore(0x1f, "123456789ABCDEFGHJKLMNPQRSTUVWXY")
+            mstore(0x3f, "Zabcdefghijkmnopqrstuvwxyz")
+
+            // Encoding the "input" part of the result.
+            // `output` point the the left part of the encoded string. we start from scratch, which means we have
```
Amxx:
```suggestion
            // `output` point the the left part of the encoded string. we start from the `scratch` pointer, which means we have
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+                // check if there are non-zero limbs remaining
+                let i := scratch
+                for {} and(iszero(mload(i)), lt(i, ptr)) {
+                    i := add(i, 0x20)
+                } {}
+                if eq(i, ptr) {
+                    break
+                }
+
+                // base 58 arithmetic on the 248bits limbs
```
Amxx:
```suggestion
                // base 58 arithmetic on the 248bits limbs:
                // we compute the remainder modulo 58 of the value encoded in the limbs, 
                // all while dividing the value encoded in the limbs by 58
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+                let carry := 0
+                for {
+                    i := scratch
+                } lt(i, ptr) {
+                    i := add(i, 0x20)
+                } {
+                    let acc := add(shl(248, carry), mload(i))
```
Amxx:
```suggestion
                    // Load the next limb and add the carry (limbs are 248bits objects)
                    let acc := add(shl(248, carry), mload(i))
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+                    mstore(i, div(acc, 58))
```
Amxx:
```suggestion
                    // store the updated limb (divided by 58)
                    mstore(i, div(acc, 58))
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+                    carry := mod(acc, 58)
```
Amxx:
```suggestion
                    // carry the remainder (modulo 58)
                    carry := mod(acc, 58)
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,207 @@
...
+        return _decode(bytes(input));
+    }
+
+    function _encode(bytes memory input) private pure returns (bytes memory output) {
+        uint256 inputLength = input.length;
+        if (inputLength == 0) return "";
+
+        assembly ("memory-safe") {
+            // Count number of zero bytes at the beginning of `input`. These are encoded using the same number of '1's
+            // at then beginning of the encoded string.
```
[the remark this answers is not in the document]
Amxx:
```suggestion
            // at the beginning of the encoded string.
```

test/utils/Base58.t.sol
```diff
@@ -0,0 +1,24 @@
...
+
+pragma solidity ^0.8.26;
+
+import {Test} from "forge-std/Test.sol";
+import {Base58} from "@openzeppelin/contracts/utils/Base58.sol";
+
+contract Base58Test is Test {
+    function testEncodeDecodeEmpty() external pure {
+        assertEq(Base58.decode(Base58.encode("")), "");
+    }
```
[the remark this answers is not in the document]
Amxx:
```suggestion
    function testEncodeDecodeEmpty() external pure {
        assertEq(Base58.decode(Base58.encode(hex"")), hex"");
    }
```

contracts/utils/Bytes.sol
```diff
@@ -68,6 +68,28 @@ library Bytes {
...
+        assembly ("memory-safe") {
+            for {
+                let ptr := add(add(buffer, 0x20), offset)
+                i := 0
+            } and(iszero(shr(248, xor(mload(add(ptr, i)), search))), lt(i, length)) {
+                i := add(i, 1)
+            } {}
+        }
+    }
+
```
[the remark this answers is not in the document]
Amxx: Note: replacing `iszero(shr(248, xor(mload(add(ptr, i)), search)))` with `iszero(eq(byte(0, mload(add(ptr, i))), search)))` implies that the data in search is in the leftmost 8 bits as opposed to the rightmost 8 bits. Need to check (this is a bytes1, so maybe its the correct approach)

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Provides a set of functions to operate with Base58 strings.
+ *
+ * Initially based on https://github.com/storyicon/base58-solidity/commit/807428e5174e61867e4c606bdb26cba58a8c5cb1[storyicon's implementation] (MIT).
+ * Based on the updated and improved https://github.com/Vectorized/solady/blob/main/src/utils/Base58.sol[Vectorized version] (MIT).
+ */
+library Base58 {
+    error InvalidBase58Digit(bytes1);
```
arr00: "character" or "value" fits better here

contracts/utils/Bytes.sol
```diff
@@ -99,6 +121,35 @@ library Bytes {
...
+     * @dev In place slice: moves the content of `buffer`, from `start` (included) to the end of `buffer` to the start of that buffer.
+     *
+     * NOTE: This function modifies the provided buffer in place. If you need to preserve the original buffer, use {slice} instead
+     */
+    function splice(bytes memory buffer, uint256 start) internal pure returns (bytes memory) {
+        return splice(buffer, start, buffer.length);
+    }
+
+    /**
+     * @dev In place slice: moves the content of `buffer`, from `start` (included) to end (excluded) to the start of that buffer.
```
arr00:
```suggestion
     * @dev In place slice: moves the content of `buffer`, from `start` (inclusive) to end (exclusive) to the start of `buffer`.
```

contracts/utils/Bytes.sol
```diff
@@ -99,6 +121,35 @@ library Bytes {
         return result;
     }
 
+    /**
+     * @dev In place slice: moves the content of `buffer`, from `start` (included) to the end of `buffer` to the start of that buffer.
```
arr00:
```suggestion
     * @dev In place slice: moves the content of `buffer`, from `start` (inclusive) to the end of `buffer` to the start of `buffer`.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+            let inputLeadingZeros := 0
+            for {} lt(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), lt(inputLeadingZeros, inputLength)) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 8351), 6115))
+
+            // This is going to be our "scratch" workspace. We leave enough room after FMP to later store length + encoded output.
+            let scratch := add(mload(0x40), add(outputLengthEstim, 0x21))
```
arr00, line 58: Just confirming that 0x21 is size + rounding up?

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+            // - `data` points to the first (highest) non-empty limb. As limb get nullified by the successive
+            //   divisions by 58, we don't need to reprocess the highest ones. Algorithm ends when all limbs are zeroed
+            //   i.e. when the `data` pointer reaches the `ptr` pointer that correspond to the last limb.
+            // - `output` point the the left part of the encoded string. We start from scratch, which means we have
+            //   outputLengthEstim bytes to work with before hitting the FMP
+            for {
+                let data := scratch
+                output := scratch
+            } 1 {} {
+                // move past the fisrt (highest) zero limbs.
```
Amxx:
```suggestion
                // move past the first (highest) zero limbs.
```

contracts/utils/Base58.sol
arr00: i think this worthwhile to add as a comment

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+                }
+
+                // base 58 arithmetic on the 248bits limbs
+                let carry := 0
+                for {
+                    let i := data
+                } lt(i, ptr) {
+                    i := add(i, 0x20)
+                } {
+                    let acc := add(shl(248, carry), mload(i))
```
arr00:
```suggestion
                    let numerator := add(shl(248, carry), mload(i))
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+                    mstore(i, div(acc, 58))
+                    carry := mod(acc, 58)
```
arr00:
```suggestion
                    remainder := mod(acc, 58)
```
I would call it remainder here

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+                }
+
+                // encode carry using base58 table, and add it to the output
+                output := sub(output, 1)
+                mstore8(output, mload(carry))
+            }
+
+            // Write the input leading zeros at the left of the encoded.
+            // This will spill to the left into the "length" of the buffer.
```
arr00:
```suggestion
            // This may spill to the left into the "length" of the buffer.
```
If its a multiple of 32 it wont? Could be?

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    function _decode(bytes memory input) private pure returns (bytes memory output) {
+        bytes4 errorSelector = InvalidBase58Char.selector;
+
+        uint256 inputLength = input.length;
+        if (inputLength == 0) return "";
+
+        /// @solidity memory-safe-assembly
+        assembly {
```
arr00:
```suggestion
        assembly ("memory-safe") {
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+
+            // This is going to be our "scratch" workspace. Be leave enough room on the left to store length + encoded input.
+            let scratch := add(mload(0x40), add(outputLengthEstim, 0x21))
+
+            // Store the decoding table. This overlaps with the FMP that we are going to reset later anyway.
+            mstore(0x2a, 0x30313233343536373839)
+            mstore(0x20, 0x1718191a1b1c1d1e1f20ffffffffffff2122232425262728292a2bff2c2d2e2f)
+            mstore(0x00, 0x000102030405060708ffffffffffffff090a0b0c0d0e0f10ff1112131415ff16)
+
+            // Decode each char of the input string, and stored that in section (limbs) of 31 bytes. Store in scratch.
```
arr00:
```suggestion
            // Decode each char of the input string, and store it in sections (limbs) of 31 bytes. Store in scratch.
```

contracts/utils/Base58.sol
Amxx: 0x20 is the length part. Not sure about the 1. `outputLengthEstim` should already be over estimated.

contracts/utils/Base58.sol
Amxx: Not sure what you mean by numerator / denominator. To me these for fractions, but there is no fraction here, is thee ?

`carry` really feels like the right word to me. Its what remains of the operation and gets carried to the next step, like in fulladders or other similar operations.

`acc` is for accululator. We might find a better name, but I'm really nkot convinced by `numerator`

contracts/utils/Base58.sol
Amxx: I dont't understand the proposal. Would that be clearer ?
```suggestion
            // This will overflow to the left into the "length" part of the reserved buffer.
```

contracts/utils/Base58.sol
arr00: > In [elementary arithmetic](https://en.wikipedia.org/wiki/Elementary_arithmetic), a carry is a [digit](https://en.wikipedia.org/wiki/Numerical_digit) that is transferred from one [column](https://en.wikipedia.org/wiki/Column) of digits to another column of more significant digits.

From wikipedia. I was confused by the word carry since I expected it to be going to somewhere more significant (not less).

arr00: This is the inverse of add and mul.

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,211 @@
...
+        assembly ("memory-safe") {
+            // Count number of zero bytes at the beginning of `input`. These are encoded using the same number of '1's
+            // at the beginning of the encoded string.
+            let inputLeadingZeros := 0
+            for {} lt(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), lt(inputLeadingZeros, inputLength)) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 8351), 6115))
```
Amxx:
```suggestion
            // This is an (over) estimation of the length ratio between bytes (base 256) and base58
            // 8351 / 6115 = 1.365658217497956 > 1.365658237309761 = Math.log(256) / Math.log(58)
            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 8351), 6115))
```

contracts/utils/Base58.sol
arr00: I'm pointing out that there is a chance the spill will not happen. So want to change "will" to "may".

contracts/utils/Base58.sol
Amxx: I think I see what you mean. We are doing a division, so you think or "partial" remainder at each step. That makes sens.

contracts/utils/Base58.sol
Amxx: right. I'm not reading correctly today :/

Its true that if the number of 0 is a multiple of 32, there may not be a spill. That is very unlikelly though.

contracts/utils/Base58.sol
Amxx: yes, the 1 is a round up. I updated the number (for the encode case) and clarified the comments so that the estimation is always an over estimation and we don't need the +1.

Amxx: Update: fuzzing fails if we don't have the +1. Can't exactly remember why.

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Provides a set of functions to operate with Base58 strings.
+ *
```
ernestognw, line 7: Let's add more context about Base58 in general

```suggestion
 * Base58 is an encoding scheme that converts binary data into a human-readable text format.
 * Similar to {Base64} but specifically designed for better human usability.
 *
 * 1. Human-friendly alphabet: Excludes visually similar characters to reduce human error:
 *    * No 0 (zero) vs O (capital O) confusion
 *    * No I (capital i) vs l (lowercase L) confusion
 *    * No non-alphanumeric characters like + or =
 * 2. URL-safe: Contains only alphanumeric characters, making it safe for URLs without encoding.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+ * Initially based on https://github.com/storyicon/base58-solidity/commit/807428e5174e61867e4c606bdb26cba58a8c5cb1[storyicon's implementation] (MIT).
+ * Based on the updated and improved https://github.com/Vectorized/solady/blob/main/src/utils/Base58.sol[Vectorized version] (MIT).
```
ernestognw: Better to use a permalink imo

```suggestion
 * Based on the updated and improved https://github.com/Vectorized/solady/blob/208e4f31cfae26e4983eb95c3488a14fdc497ad7/src/utils/Base58.sol[Vectorized version] (MIT).
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    function _encode(bytes memory input) private pure returns (bytes memory output) {
+        uint256 inputLength = input.length;
+        if (inputLength == 0) return "";
+
+        assembly ("memory-safe") {
+            // Count number of zero bytes at the beginning of `input`. These are encoded using the same number of '1's
+            // at the beginning of the encoded string.
+            let inputLeadingZeros := 0
+            for {} lt(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), lt(inputLeadingZeros, inputLength)) {} {
```
ernestognw, line 46: This loops seems to work **unless** a malformed bytes input buffer is provided. Let's say:

```solidity
bytes memory malformed;
assembly {
    malformed := mload(0x40)  // Get free memory pointer
    mstore(malformed, 1000)   // Set length to 1000
    mstore(add(malformed, 0x20), 0x0000)  // Only 2 bytes of actual data
    mstore(0x40, add(malformed, 0x40))    // Update free memory pointer minimally
}
```

I don't think it's an issue as long as developers provide a wellformed `bytes` pointer, but I'm not sure if that should be documented

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            // This is an estimation of the length ratio between bytes (base 256) and base58
+            // 9886 / 7239 = 1.36565824008841 > 1.365658237309761 = Math.log(256) / Math.log(58)
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 9886), 7239))
```
ernestognw, line 54: Solady's implementation is using 8351 / 6115, is there any significant difference? Note that on `_decode` we're using the reciprocal 6115 / 8351, I wonder if we should use 7239 / 9886 there

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+            // Count number of zero bytes at the beginning of `input`. These are encoded using the same number of '1's
+            // at the beginning of the encoded string.
+            let inputLeadingZeros := 0
+            for {} lt(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), lt(inputLeadingZeros, inputLength)) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            // This is an estimation of the length ratio between bytes (base 256) and base58
+            // 9886 / 7239 = 1.36565824008841 > 1.365658237309761 = Math.log(256) / Math.log(58)
```
ernestognw: This should be clearer:

```suggestion
            // When converting from base-256 (bytes) to base-58, the theoretical length ratio is ln(256)/ln(58).
            // We use 9886/7239 ≈ 1.3657 as a rational approximation that slightly over-estimates to ensure
            // sufficient memory allocation. (ln = natural logarithm)
```

Also note that `log` doesn't mean the same as natural logarithm in some calculators so that was confusing while reading through this.

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+            } lt(i, inputLength) {
+                ptr := add(ptr, 0x20) // next limb
+                i := add(i, 31) // move in buffer
+            } {
+                // Load 31 bytes from the input buffer and store then in scratch (at ptr) in a dedicated 32 bytes space.
+                mstore(ptr, shr(8, mload(add(add(input, 0x20), i))))
+            }
+
+            // Store the encoding table. This overlaps with the FMP that we are going to reset later anyway.
+            // See sections 2 of https://inputtracker.ietf.org/doc/html/draft-msporny-base58-03
```
ernestognw:
```suggestion
            // See https://datatracker.ietf.org/doc/html/draft-msporny-base58-03#section-2
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+            for {} lt(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), lt(inputLeadingZeros, inputLength)) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            // This is an estimation of the length ratio between bytes (base 256) and base58
+            // 9886 / 7239 = 1.36565824008841 > 1.365658237309761 = Math.log(256) / Math.log(58)
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 9886), 7239))
+
+            // This is going to be our "scratch" workspace. We leave enough room after FMP to later store length + encoded output.
```
ernestognw: Isn't?

```suggestion
            // This is going to be our "scratch" workspace. We leave enough room before FMP to later store length + encoded output.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+            // See sections 2 of https://inputtracker.ietf.org/doc/html/draft-msporny-base58-03
+            mstore(0x1f, "123456789ABCDEFGHJKLMNPQRSTUVWXY")
+            mstore(0x3f, "Zabcdefghijkmnopqrstuvwxyz")
+
+            // Encoding the "input" part of the result.
+            // - `data` points to the first (highest) non-empty limb. As limb get nullified by the successive
+            //   divisions by 58, we don't need to reprocess the highest ones. Algorithm ends when all limbs are zeroed
+            //   i.e. when the `data` pointer reaches the `ptr` pointer that correspond to the last limb.
+            // - `output` point the the left part of the encoded string. We start from scratch, which means we have
+            //   outputLengthEstim bytes to work with before hitting the FMP
```
ernestognw: This is difficult to follow until someone starts reading the loop. I would suggest a small diagram for initialization:


```suggestion
            // Memory layout: [output chars] [limb₁(248 bits)][limb₂(248 bits)][limb₃(248 bits)]...
            //                               ↑ scratch                          ↑ ptr
            //                               ↑ output (moves left)
            //                               ↑ data (moves right)
```

Then, readers can follow the logic down below

ernestognw (commented):
Will keep reviewing

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,214 @@
...
+                eq(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), 49),
+                lt(inputLeadingZeros, inputLength)
+            ) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            // This is an estimation of the length ratio between base58 and bytes (base 256)
+            // 6115 / 8351 = 0.7322476350137708 > 0.7322476243909465 = Math.log(58) / Math.log(256)
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 6115), 8351))
```
ernestognw, line 160: According to my comment [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5762/files#r2379908363)

```suggestion
            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 7239), 9886))
```

contracts/utils/Base58.sol
Amxx: that is not malformed, its a properly allocated buffer of size 1000 that contains 32 bytes of zeros (mstore does write 32 bytes, regardless of the value being `0x0000`) ... and  undefined data after that. If you encode that, you'll get a proper encoding of whatever is at that location.

ernestognw: Hmm yeah that's right. I was trying to make a counterexample to the `lt` comparison instead of using `and(eq(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), 0), lt(inputLeadingZeros, inputLength))`

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,230 @@
...
+ *
+ * Base58 is an encoding scheme that converts binary data into a human-readable text format.
+ * Similar to {Base64} but specifically designed for better human usability.
+ *
+ * 1. Human-friendly alphabet: Excludes visually similar characters to reduce human error:
+ *    * No 0 (zero) vs O (capital O) confusion
+ *    * No I (capital i) vs l (lowercase L) confusion
+ *    * No non-alphanumeric characters like + or =
+ * 2. URL-safe: Contains only alphanumeric characters, making it safe for URLs without encoding.
+ *
```
ernestognw, line 16: Would it be worth to specify the following note?

```suggestion
 *
 * NOTE: This implementation follows the Bitcoin Base58 alphabet and encoding rules, which have become
 * the de facto standard despite the Base58 specification still being in draft status. The algorithm
 * is compatible with popular libraries and Bitcoin/cryptocurrency applications. Though the specification
 * is still in draft status, it has achieved widespread adoption in the cryptocurrency ecosystem.
 * 
```

contracts/utils/Base58.sol
Amxx: If you look at the (old) comments (that you changed), you'll see that the values are very specifically chosen, and different on purpose !

```
9886 / 7239 = 1.36565824008841 > 1.365658237309761 (Math.log(256) / Math.log(58)) > 1.365658217497956 = 8351/6115
```
the faction `9886 / 7239` and `8351/6115` have been chosen specifically to "surround the exact value". 

That is so `9886 / 7239 > Math.log(256) / Math.log(58)` and `6115/8351 > Math.log(58) / Math.log(256)`

That way we have an overestimation in both cases.

contracts/utils/Base58.sol
Amxx: see https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5762/files#r2380126147

contracts/utils/Base58.sol
Amxx: IMO, that is not necessary.

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,230 @@
...
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            // When converting from base-256 (bytes) to base-58, the theoretical length ratio is ln(256)/ln(58).
+            // We use 9886/7239 ≈ 1.3657 as a rational approximation that slightly over-estimates to ensure
+            // sufficient memory allocation. (ln = natural logarithm)
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 9886), 7239))
+
+            // This is going to be our "scratch" workspace. We leave enough room before FMP to later store length + encoded output.
```
Amxx:
```suggestion
            // This is going to be our "scratch" workspace. We leave enough room so that we can store length + encoded output at the FMP location.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,230 @@
...
+                if i {
+                    // Right-shift to align partial limb in high bits of 256-bit word
+                    mstore(ptr, shr(mul(sub(32, i), 8), mload(add(input, 0x20))))
+                    ptr := add(ptr, 0x20) // next limb
+                }
+            } lt(i, inputLength) {
+                ptr := add(ptr, 0x20) // next limb
+                i := add(i, 31) // move in buffer
+            } {
+                // Load 31 bytes from input, right-shift by 8 bits to leave 1 zero byte in low bits
```
Amxx: The space is on the left, which when we do arithmetics is the "high" part of the number
```suggestion
                // Load 31 bytes from input, right-shift by 8 bits to leave 1 zero byte on the left.
```

ernestognw (approved):
lfg 🚀

ernestognw (dismissed):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 - Update minimum pragma to 0.8.24 in `Votes`, `VotesExtended`, `ERC20Votes`, `Strings`, `ERC1155URIStorage`, `MessageHashUtils`, `ERC721URIStorage`, `ERC721Votes`, `ERC721Wrapper`, `ERC721Burnable`, `ERC721Consecutive`, `ERC721Enumerable`, `ERC721Pausable`, `ERC721Royalty`, `ERC721Wrapper`, `EIP712`, and `ERC7739`. ([#5726](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5726))
 
+#### Custom error changes
+
+- Replace `InvalidBase64Digit` with `InvalidBase64Char`.
+
```
Amxx: This error was introduced in 5.5, and was never released. We don't need a changelog entry
```suggestion
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,239 @@
...
+pragma solidity ^0.8.20;
+
+/**
+ * @dev Provides a set of functions to operate with Base58 strings.
+ *
+ * Base58 is an encoding scheme that converts binary data into a human-readable text format.
+ * Similar to {Base64} but specifically designed for better human usability.
+ *
+ * 1. Human-friendly alphabet: Excludes visually similar characters to reduce human error:
+ *    * No 0 (zero) vs O (capital O) confusion
```
Amxx:
```suggestion
 *    * No 0 (zero) vs O (capital o) confusion
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,239 @@
...
+            // at the beginning of the encoded string.
+            let inputLeadingZeros := 0
+            for {} lt(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), lt(inputLeadingZeros, inputLength)) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Start the output offset by an over-estimate of the length.
+            // When converting from base-256 (bytes) to base-58, the theoretical length ratio is ln(256)/ln(58).
+            // We use 9886/7239 ≈ 1.3657 as a rational approximation that slightly over-estimates to ensure
+            // sufficient memory allocation. (ln = natural logarithm)
```
Amxx:
```suggestion
            // When converting from base-256 (bytes) to base-58, the theoretical length ratio is log(256)/log(58).
            // We use 9886/7239 ≈ 1.3657 as a rational approximation that slightly over-estimates to ensure
            // sufficient memory allocation.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,239 @@
...
+                    let i := data
+                } lt(i, ptr) {
+                    i := add(i, 0x20)
+                } {
+                    let acc := add(shl(248, carry), mload(i)) // Combine carry from previous limb with current limb
+                    mstore(i, div(acc, 58)) // Store quotient back in limb
+                    carry := mod(acc, 58) // Remainder becomes next carry
+                }
+
+                // Convert remainder (0-57) to Base58 character and store right-to-left
```
Amxx:
```suggestion
                // Convert remainder (0-57) to Base58 character and store right-to-left in the output space
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,239 @@
...
+                eq(byte(0, mload(add(add(input, 0x20), inputLeadingZeros))), 49),
+                lt(inputLeadingZeros, inputLength)
+            ) {} {
+                inputLeadingZeros := add(inputLeadingZeros, 1)
+            }
+
+            // Estimate the output length using the base conversion ratio.
+            // When converting from base-58 to base-256 (bytes), the theoretical length ratio is ln(58)/ln(256).
+            // We use 6115/8351 ≈ 0.7322 as a rational approximation that slightly over-estimates to ensure
+            // sufficient memory allocation. (ln = natural logarithm)
```
Amxx:
```suggestion
            // When converting from base-58 to base-256 (bytes), the theoretical length ratio is log(58)/log(256).
            // We use 6115/8351 ≈ 0.7322 as a rational approximation that slightly over-estimates to ensure
            // sufficient memory allocation.
```

contracts/utils/Base58.sol
```diff
@@ -0,0 +1,239 @@
...
+            let outputLengthEstim := add(inputLeadingZeros, div(mul(sub(inputLength, inputLeadingZeros), 6115), 8351))
+
+            // Reserve memory: [result space][workspace]. We leave enough room before FMP for the final decoded output.
```
Amxx:
```suggestion
            // This is going to be our "scratch" workspace. We leave enough room so that we can store length + decoded output at the FMP location.
```

PR #5765, opened 2025-06-23
Base64.decode

### LINKED ISSUE #4287: Feature request: `Base64.decode`
The current `Base64` library offers support only for base64 encoding:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5420879d9b834a0579423d668fb60c5fc13b60cc/contracts/utils/Base64.sol

However, with the advent of Foundry, there's also the need for decoding base64-encoded strings in test files.

At the moment, OpenZeppelin users are forced to go look for a solution elsewhere, e.g. in [Solady](https://github.com/Vectorized/solady/blob/8d868a936ec1a45be294e26de1a64ebfb73c6c20/src/utils/Base64.sol), which provides both encoding and decoding.

It would thus be helpful to provide a `decode` function on top of `encode`.

Amxx:
Fixes #4287

Note: this uses a different decoding table than solady. It might not be as efficient, but its a method I 100% understand and can explain. The solady method feels like magic to me ... but maybe its magic we want.

Note that changing the decoding method can always be done at a later stage. It should be backward compatible anyway

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

- New Features
  - Added Base64.decode to convert Base64/Base64URL strings back to bytes, supporting both padded and unpadded inputs.
  - Decoding now provides clear feedback with a custom error for invalid characters.

- Documentation
  - Documented the new decoding capability and updated the package changeset.

- Tests
  - Added round-trip tests ensuring encode/decode consistency for standard and URL-safe variants.
  - Introduced tests validating reverts on invalid Base64 characters.
  - Updated test suite naming and expectations for clarity.

### DIFF
--- a/test/utils/Base64.test.js
+++ b/test/utils/Base64.test.js
@@ -54,14 +54,14 @@ describe('Base64', function () {
     const getHexCode = str => ethers.hexlify(ethers.toUtf8Bytes(str));
     const helper = { interface: ethers.Interface.from(['error InvalidBase64Digit(bytes1)']) };
 
-    // ord('$') < 43
-    await expect(this.mock.$decode('dGVzd$=='))
+    // ord('*') < 43
+    await expect(this.mock.$decode('dGVzd*=='))
       .to.be.revertedWithCustomError(helper, 'InvalidBase64Digit')
-      .withArgs(getHexCode('$'));
-    // ord('~') > 122
-    await expect(this.mock.$decode('dGVzd~=='))
+      .withArgs(getHexCode('*'));
+    // ord('{') > 122
+    await expect(this.mock.$decode('dGVzd{=='))
       .to.be.revertedWithCustomError(helper, 'InvalidBase64Digit')
-      .withArgs(getHexCode('~'));
+      .withArgs(getHexCode('{'));
     // ord('@') in range, but '@' not in the dictionary
     await expect(this.mock.$decode('dGVzd@=='))
       .to.be.revertedWithCustomError(helper, 'InvalidBase64Digit')


contracts/utils/Base64.sol
```diff
@@ -54,9 +73,9 @@ library Base64 {
         // This is equivalent to: Math.ceil((4 * data.length) / 3)
         uint256 resultLength = withPadding ? 4 * ((data.length + 2) / 3) : (4 * data.length + 2) / 3;
 
-        string memory result = new string(resultLength);
-
         assembly ("memory-safe") {
+            result := mload(0x40)
```
Amxx, line 73: this avoids a loop that initializes `result` with zeros.

contracts/utils/Base64.sol
```diff
@@ -3,42 +3,57 @@
...
     function encodeURL(bytes memory data) internal pure returns (string memory) {
-        return _encode(data, _TABLE_URL, false);
+        return string(_encode(data, true));
     }
 
     /**
-     * @dev Internal table-agnostic conversion
+     * @dev Converts a Base64 `string` to the `bytes` it represents.
+     *
+     * * Supports padded an unpadded inputs.
```
arr00:
```suggestion
     * * Supports padded and unpadded inputs.
```

arr00 (dismissed):

Amxx:
For the record, this is how the lookup table are computed: https://gist.github.com/Amxx/ae43dc9dc24427f364c9d5789ad65896

test/utils/Base64.test.js
```diff
@@ -43,11 +44,30 @@ describe('Strings', function () {
...
+    const helper = { interface: ethers.Interface.from(['error InvalidBase64Digit(bytes1)']) };
+
+    // ord('$') < 43
+    await expect(this.mock.$decode('dGVzd$=='))
+      .to.be.revertedWithCustomError(helper, 'InvalidBase64Digit')
+      .withArgs(getHexCode('$'));
+    // ord('~') > 122
+    await expect(this.mock.$decode('dGVzd~=='))
+      .to.be.revertedWithCustomError(helper, 'InvalidBase64Digit')
+      .withArgs(getHexCode('~'));
```
james-toussaint: Very minor, maybe that to check range edges:
```suggestion
    // ord('*') < 43
    await expect(this.mock.$decode('dGVzd*=='))
      .to.be.revertedWithCustomError(helper, 'InvalidBase64Digit')
      .withArgs(getHexCode('*'));
    // ord('{') > 122
    await expect(this.mock.$decode('dGVzd{=='))
      .to.be.revertedWithCustomError(helper, 'InvalidBase64Digit')
      .withArgs(getHexCode('{'));
```

contracts/utils/Base64.sol
```diff
@@ -3,42 +3,57 @@
...
      */
-    function _encode(bytes memory data, string memory table, bool withPadding) private pure returns (string memory) {
+    function decode(string memory data) internal pure returns (bytes memory) {
+        return _decode(bytes(data));
+    }
+
+    /**
+     * @dev Internal table-agnostic encoding
+     *
+     * If padding is enabled, uses the Base64 table, otherwise use the Base64Url table.
```
james-toussaint:
```suggestion
     * Padding is enabled when using the Base64 table, and disabled when using the Base64Url table.
```

PR #5768, opened 2025-06-25
Use hardhat network helper to clear code blockhash test

arr00:


test/utils/Blockhash.test.js
```diff
@@ -16,7 +16,7 @@ async function fixture() {
   };
 }
 
-describe('Blockhash', function () {
+describe.only('Blockhash', function () {
```
arr00:
```suggestion
describe('Blockhash', function () {
```

PR #5771, opened 2025-06-26
Refactor SafeERC20 to save gas

Amxx:
This PR avoids the expensive `abi.encode` that allocates memory. It also processes the output (returnSize, returnValue) directly in assembly. The goal is to optimize the "happy path" when a token returns a true boolean.

No user-visible behavior change

Note: this is a priority for @luiz-lvj's work on OIF. Hopefully we can achieve good results.

#### Versions

- [bf140e4](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5771/commits/bf140e40a93ee94e62a568e6112a58f8aab0df68) saves ~60 gas
- [471e088](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5771/commits/471e088f5929d0b7d4f1f4c4675f384e4392a82b) saves ~170 gas 🥳

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/contracts/token/ERC20/utils/SafeERC20.sol
+++ b/contracts/token/ERC20/utils/SafeERC20.sol
@@ -104,9 +104,8 @@ library SafeERC20 {
      */
     function forceApprove(IERC20 token, address spender, uint256 value) internal {
         if (!_safeApprove(token, spender, value, false)) {
-            if (!_safeApprove(token, spender, 0, true) || !_safeApprove(token, spender, value, true)) {
-                revert SafeERC20FailedOperation(address(token));
-            }
+            if (!_safeApprove(token, spender, 0, true)) revert SafeERC20FailedOperation(address(token));
+            if (!_safeApprove(token, spender, value, true)) revert SafeERC20FailedOperation(address(token));
         }
     }
 
--- a/contracts/token/ERC20/utils/SafeERC20.sol
+++ b/contracts/token/ERC20/utils/SafeERC20.sol
@@ -174,11 +174,8 @@ library SafeERC20 {
      * @param value The amount of token to transfer
      * @param bubble Behavior switch if the transfer call reverts: bubble the revert reason or return a false boolean.
      */
-    function _safeTransfer(IERC20 token, address to, uint256 value, bool bubble) private returns (bool) {
+    function _safeTransfer(IERC20 token, address to, uint256 value, bool bubble) private returns (bool success) {
         bytes4 selector = IERC20.transfer.selector;
-        bool success;
-        uint256 returnSize;
-        uint256 returnValue;
 
         assembly ("memory-safe") {
             let fmp := mload(0x40)
@@ -186,16 +183,23 @@ library SafeERC20 {
             mstore(0x04, and(to, shr(96, not(0))))
             mstore(0x24, value)
             success := call(gas(), token, 0, 0, 0x44, 0, 0x20)
-            if and(bubble, iszero(success)) {
-                returndatacopy(fmp, 0, returndatasize())
-                revert(fmp, returndatasize())
+            // if call success and return is true, all is good.
+            if iszero(and(success, eq(mload(0x00), 1))) {
+                // if the call was a failure and bubble is enabled, bubble the error
+                if and(iszero(success), bubble) {
+                    returndatacopy(fmp, 0, returndatasize())
+                    revert(fmp, returndatasize())
+                }
+                // if the call is successful, and either
+                // - we got returndata (is that case mload(0x00) is a real false)
+                // - the token address doesn't have code
+                if and(success, or(gt(returndatasize(), 0), iszero(extcodesize(token)))) {
+                    // than the transfer is a failure
+                    success := 0
+                }
             }
-            returnSize := returndatasize()
-            returnValue := mload(0)
             mstore(0x40, fmp)
         }
-
-        return success && (returnSize == 0 ? address(token).code.length > 0 : returnValue == 1);
     }
 
     /**
@@ -214,11 +218,8 @@ library SafeERC20 {
         address to,
         uint256 value,
         bool bubble
-    ) private returns (bool) {
+    ) private returns (bool success) {
         bytes4 selector = IERC20.transferFrom.selector;
-        bool success;
-        uint256 returnSize;
-        uint256 returnValue;
 
         assembly ("memory-safe") {
             let fmp := mload(0x40)
@@ -227,17 +228,24 @@ library SafeERC20 {
             mstore(0x24, and(to, shr(96, not(0))))
             mstore(0x44, value)
             success := call(gas(), token, 0, 0, 0x64, 0, 0x20)
-            if and(bubble, iszero(success)) {
-                returndatacopy(fmp, 0, returndatasize())
-                revert(fmp, returndatasize())
+            // if call success and return is true, all is good.
+            if iszero(and(success, eq(mload(0x00), 1))) {
+                // if the call was a failure and bubble is enabled, bubble the error
+                if and(iszero(success), bubble) {
+                    returndatacopy(fmp, 0, returndatasize())
+                    revert(fmp, returndatasize())
+                }
+                // if the call is successful, and either
+                // - we got returndata (is that case mload(0x00) is a real false)
+                // - the token address doesn't have code
+                if and(success, or(gt(returndatasize(), 0), iszero(extcodesize(token)))) {
+                    // than the transfer is a failure
+                    success := 0
+                }
             }
-            returnSize := returndatasize()
-            returnValue := mload(0)
             mstore(0x40, fmp)
             mstore(0x60, 0)
         }
-
-        return success && (returnSize == 0 ? address(token).code.length > 0 : returnValue == 1);
     }
 
     /**
@@ -249,11 +257,8 @@ library SafeERC20 {
      * @param value The amount of token to transfer
      * @param bubble Behavior switch if the transfer call reverts: bubble the revert reason or return a false boolean.
      */
-    function _safeApprove(IERC20 token, address spender, uint256 value, bool bubble) private returns (bool) {
+    function _safeApprove(IERC20 token, address spender, uint256 value, bool bubble) private returns (bool success) {
         bytes4 selector = IERC20.approve.selector;
-        bool success;
-        uint256 returnSize;
-        uint256 returnValue;
 
         assembly ("memory-safe") {
             let fmp := mload(0x40)
@@ -261,15 +266,22 @@ library SafeERC20 {
             mstore(0x04, and(spender, shr(96, not(0))))
             mstore(0x24, value)
             success := call(gas(), token, 0, 0, 0x44, 0, 0x20)
-            if and(bubble, iszero(success)) {
-                returndatacopy(fmp, 0, returndatasize())
-                revert(fmp, returndatasize())
+            // if call success and return is true, all is good.
+            if iszero(and(success, eq(mload(0x00), 1))) {
+                // if the call was a failure and bubble is enabled, bubble the error
+                if and(iszero(success), bubble) {
+                    returndatacopy(fmp, 0, returndatasize())
+                    revert(fmp, returndatasize())
+                }
+                // if the call is successful, and either
+                // - we got returndata (is that case mload(0x00) is a real false)
+                // - the token address doesn't have code
+                if and(success, or(gt(returndatasize(), 0), iszero(extcodesize(token)))) {
+                    // than the transfer is a failure
+                    success := 0
+                }
             }
-            returnSize := returndatasize()
-            returnValue := mload(0)
             mstore(0x40, fmp)
         }
-
-        return success && (returnSize == 0 ? address(token).code.length > 0 : returnValue == 1);
     }
 }
--- a/contracts/token/ERC20/utils/SafeERC20.sol
+++ b/contracts/token/ERC20/utils/SafeERC20.sol
@@ -236,7 +236,7 @@ library SafeERC20 {
                     revert(fmp, returndatasize())
                 }
                 // if the call is successful, and either
-                // - we got returndata (is that case mload(0x00) is a real false)
+                // - we got returndata (in that case mload(0x00) is a real false)
                 // - the token address doesn't have code
                 if and(success, or(gt(returndatasize(), 0), iszero(extcodesize(token)))) {
                     // than the transfer is a failure
--- a/contracts/token/ERC20/utils/SafeERC20.sol
+++ b/contracts/token/ERC20/utils/SafeERC20.sol
@@ -184,6 +184,7 @@ library SafeERC20 {
             mstore(0x24, value)
             success := call(gas(), token, 0, 0, 0x44, 0, 0x20)
             // if call success and return is true, all is good.
+            // otherwize (not success or return is not true), we need to perform further checks
             if iszero(and(success, eq(mload(0x00), 1))) {
                 // if the call was a failure and bubble is enabled, bubble the error
                 if and(iszero(success), bubble) {
@@ -229,6 +230,7 @@ library SafeERC20 {
             mstore(0x44, value)
             success := call(gas(), token, 0, 0, 0x64, 0, 0x20)
             // if call success and return is true, all is good.
+            // otherwize (not success or return is not true), we need to perform further checks
             if iszero(and(success, eq(mload(0x00), 1))) {
                 // if the call was a failure and bubble is enabled, bubble the error
                 if and(iszero(success), bubble) {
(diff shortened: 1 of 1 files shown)


frangio:
For reference:

```mermaid
flowchart TD
    Start([Start]) --> Call[call token]

    Call --> Success{success?}

    Success -->|No| Bubble[Bubble revert]
    Success -->|Yes| Check{returnSize == 0?}

    Check -->|Yes| HasCode{token has code?}
    Check -->|No| RetVal{returnValue == 1?}

    HasCode -->|No| Fail[Revert]
    HasCode -->|Yes| OK[Return]

    RetVal -->|No| Fail
    RetVal -->|Yes| OK

    Bubble --> Revert1([Revert])
    Fail --> Revert2([Revert])
    OK --> End([End])

    style Start fill:#90EE90
    style Revert1 fill:#FFB6C1
    style Revert2 fill:#FFB6C1
    style OK fill:#90EE90
    style End fill:#90EE90
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
-     * @param token The token targeted by the call.
-     * @param data The call data (encoded using abi.encode or one of its variants).
+     * @dev Imitates a Solidity `token.transfer(to, value)` call, relaxing the requirement on the return value: the
+     * return value is optional (but if data is returned, it must not be false).
      *
-     * This is a variant of {_callOptionalReturnBool} that reverts if call fails to meet the requirements.
+     * @param token The token targeted by the call.
+     * @param to The recipient of the tokens
+     * @param value The amount of token to transfer
+     * @param bubble Behavior switch if the transfer call reverts: bubble the revert reason or return a false boolean.
```
arr00, line 174: Reminder that this won't be parsed by our docs.

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
+            success := call(gas(), token, 0, 0, 0x64, 0, 0x20)
+            // if call success and return is true, all is good.
+            if iszero(and(success, eq(mload(0x00), 1))) {
+                // if the call was a failure and bubble is enabled, bubble the error
+                if and(iszero(success), bubble) {
+                    returndatacopy(fmp, 0, returndatasize())
+                    revert(fmp, returndatasize())
+                }
+                // if the call is successful, and either
+                // - we got returndata (is that case mload(0x00) is a real false)
```
arr00:
```suggestion
                // - we got returndata (in that case mload(0x00) is a real false)
```
Is this what you mean? Doesn't make sense right now. Same as other function

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
+                // - the token address doesn't have code
+                if and(success, or(gt(returndatasize(), 0), iszero(extcodesize(token)))) {
```
arr00:
```suggestion
                if and(or(iszero(extcodesize(token)), gt(returndatasize(), 0)), success) {
```
Yul evaluates right to left. We should put the cheaper evaluation on the right.

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
+                    // than the transfer is a failure
```
arr00:
```suggestion
                    // then the approval is a failure
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
```
arr00:
```suggestion
                    // then the transfer is a failure
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
```
arr00:
```suggestion
                    // then the transfer is a failure
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
+            success := call(gas(), token, 0, 0, 0x44, 0, 0x20)
+            // if call success and return is true, all is good.
+            if iszero(and(success, eq(mload(0x00), 1))) {
+                // if the call was a failure and bubble is enabled, bubble the error
+                if and(iszero(success), bubble) {
+                    returndatacopy(fmp, 0, returndatasize())
+                    revert(fmp, returndatasize())
+                }
+                // if the call is successful, and either
+                // - we got returndata (is that case mload(0x00) is a real false)
```
arr00:
```suggestion
                // - we got returndata (in that case mload(0x00) is a real false)
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
```
arr00:
```suggestion
                // - we got returndata (in that case mload(0x00) is a real false)
```

contracts/token/ERC20/utils/SafeERC20.sol
frangio: There is no short circuiting, does order of evaluation make any difference?

arr00: Ah forgot about that good point. Doesn't matter then. Last time I was in this situation, the evaluation mutated data that was necessary for the second one.

contracts/token/ERC20/utils/SafeERC20.sol
Amxx: It's a private function. It won't be in docs anyway

arr00 (dismissed):

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +166,122 @@ library SafeERC20 {
...
-            if iszero(success) {
-                let ptr := mload(0x40)
-                returndatacopy(ptr, 0, returndatasize())
-                revert(ptr, returndatasize())
+            let fmp := mload(0x40)
+            mstore(0x00, selector)
+            mstore(0x04, and(to, shr(96, not(0))))
+            mstore(0x24, value)
+            success := call(gas(), token, 0, 0, 0x44, 0, 0x20)
+            // if call success and return is true, all is good.
```
frangio, line 186: Some comments are not super clear IMO. This one for example describes the negation of the actual condition below.

frangio, line 185: This comment is confusing because it describes the negation of the condition below.

Amxx:
Was (informally) approved during today's call. 

OIF needs that released before end of Q3.

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
+            // otherwise (not success or return is not true), we need to perform further checks
```
ernestognw, line 186: This might be clearer:

```suggestion
// if call failed or return value is not boolean true, we need to perform further checks
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
+
+        assembly ("memory-safe") {
+            let fmp := mload(0x40)
+            mstore(0x00, selector)
+            mstore(0x04, and(from, shr(96, not(0))))
+            mstore(0x24, and(to, shr(96, not(0))))
+            mstore(0x44, value)
+            success := call(gas(), token, 0, 0, 0x64, 0, 0x20)
+            // if call success and return is true, all is good.
+            // otherwise (not success or return is not true), we need to perform further checks
```
ernestognw, line 229:
```suggestion
            // if call failed or return value is not boolean true, we need to perform further checks
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
-            success := call(gas(), token, 0, add(data, 0x20), mload(data), 0, 0x20)
-            returnSize := returndatasize()
-            returnValue := mload(0)
+            let fmp := mload(0x40)
+            mstore(0x00, selector)
+            mstore(0x04, and(spender, shr(96, not(0))))
+            mstore(0x24, value)
+            success := call(gas(), token, 0, 0, 0x44, 0, 0x20)
+            // if call success and return is true, all is good.
+            // otherwise (not success or return is not true), we need to perform further checks
```
ernestognw, line 265:
```suggestion
            // if call failed or return value is not boolean true, we need to perform further checks
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
+    function _safeTransferFrom(
+        IERC20 token,
+        address from,
+        address to,
+        uint256 value,
+        bool bubble
+    ) private returns (bool success) {
+        bytes4 selector = IERC20.transferFrom.selector;
+
+        assembly ("memory-safe") {
```
ernestognw, line 221: The block is not memory safe. For reference, it is what the Solidity documentation explicitly says [not to do](https://docs.soliditylang.org/en/latest/assembly.html#memory-safety):

> Since this is mainly about the optimizer, these restrictions still need to be followed, even if the assembly block reverts or terminates. As an example, the following assembly snippet is not memory safe, because the value of returndatasize() may exceed the 64 byte scratch space:
> ```solidity
> assembly {
>   returndatacopy(0, 0, returndatasize())
>   revert(0, returndatasize())
> }
> ```

```suggestion
        assembly {
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
+     * @param value The amount of token to transfer
+     * @param bubble Behavior switch if the transfer call reverts: bubble the revert reason or return a false boolean.
      */
-    function _callOptionalReturn(IERC20 token, bytes memory data) private {
-        uint256 returnSize;
-        uint256 returnValue;
+    function _safeTransfer(IERC20 token, address to, uint256 value, bool bubble) private returns (bool success) {
+        bytes4 selector = IERC20.transfer.selector;
+
         assembly ("memory-safe") {
```
ernestognw, line 179: The block is not memory safe if it bubbles up.

```suggestion
        assembly {
```

For reference, it does what the Solidity docs say not to do on [_memory safety_](https://docs.soliditylang.org/en/latest/assembly.html#memory-safety):

> [...] Since this is mainly about the optimizer, these restrictions still need to be followed, even if the assembly block reverts or terminates. As an example, the following assembly snippet is not memory safe, because the value of returndatasize() may exceed the 64 byte scratch space:
> ```solidity
> assembly {
>   returndatacopy(0, 0, returndatasize())
>   revert(0, returndatasize())
> }
>```

contracts/token/ERC20/utils/SafeERC20.sol
ernestognw, line 221: Same


```suggestion
        assembly {
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
+     * @param bubble Behavior switch if the transfer call reverts: bubble the revert reason or return a false boolean.
      */
-    function _callOptionalReturnBool(IERC20 token, bytes memory data) private returns (bool) {
-        bool success;
-        uint256 returnSize;
-        uint256 returnValue;
+    function _safeApprove(IERC20 token, address spender, uint256 value, bool bubble) private returns (bool success) {
+        bytes4 selector = IERC20.approve.selector;
+
         assembly ("memory-safe") {
```
ernestognw, line 258: Same

```suggestion
        assembly {
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
+            if iszero(and(success, eq(mload(0x00), 1))) {
+                // if the call was a failure and bubble is enabled, bubble the error
+                if and(iszero(success), bubble) {
+                    returndatacopy(0, 0, returndatasize())
+                    revert(0, returndatasize())
+                }
+                // if the return value is not true, then the call is only successful if:
+                // - the token address has code
+                // - the returndata is empty
+                success := and(success, and(iszero(returndatasize()), gt(extcodesize(token), 0)))
```
ernestognw, line 196: Isn't this equivalent?

```suggestion
                success := and(success, and(iszero(returndatasize()), extcodesize(token))
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
```
ernestognw, line 239: Same

```suggestion
                success := and(success, and(iszero(returndatasize()), extcodesize(token))
```

contracts/token/ERC20/utils/SafeERC20.sol
```diff
@@ -163,50 +165,116 @@ library SafeERC20 {
...
```
ernestognw, line 275:
```suggestion
                success := and(success, and(iszero(returndatasize()), extcodesize(token))
```

contracts/token/ERC20/utils/SafeERC20.sol
ernestognw: I guess the main difference is that `success` is "dirty". Is that the case?

contracts/token/ERC20/utils/SafeERC20.sol
frangio: Since memory-unsafe assembly disables optimizations, we should find a way to make the code memory safe instead, otherwise merging this PR could be counterproductive.

Note that the docs are somewhat contradictory and there [is a later section](https://docs.soliditylang.org/en/latest/assembly.html#advanced-safe-use-of-memory) that softens some of the requirements. `returndatacopy(0, 0, returndatasize())` remains unsafe though as far as I can tell.

contracts/token/ERC20/utils/SafeERC20.sol
arr00: Its bitwise right? Then you would get 0 if codesize is even.

contracts/token/ERC20/utils/SafeERC20.sol
arr00: Isn't this exactly what we are doing? The docs are quite contradictory.

> The assembly block terminates, i.e. execution can never return to high-level Solidity code. This is the case, for example, if your assembly block unconditionally ends in calling the revert opcode.

frangio: That's only up to offset 0x80 according to the text above it:

> If you are careful, it can be admissible to use memory up to (and not including) offset 0x80

contracts/token/ERC20/utils/SafeERC20.sol
Amxx: Nice catch, then we should do the memory safe of it
```
assembly ("memory-safe") {
  let p := mload(0x40)
  returndatacopy(p, 0, returndatasize())
  revert(p, returndatasize())
}
```

contracts/token/ERC20/utils/SafeERC20.sol
Amxx: >Its bitwise right? Then you would get 0 if codesize is even.

Exactly!

Amxx: if
```
success:               0x0x0000000000000000000000000000000000000000000000000000000000000001
returndatasize():      0x0x0000000000000000000000000000000000000000000000000000000000000000
not(returndatasize()): 0x0x0000000000000000000000000000000000000000000000000000000000000001
extcodesize(token):    0x0x00000000000000000000000000000000000000000000000000000000001a6f90
```
then the restult of the and would be false

contracts/token/ERC20/utils/SafeERC20.sol
Amxx: Updated the code to store the revert bubble at the fmp.

ernestognw (approved):
LGTM, though, it may make sense to measure gas again to see what's the impact on reverting using the fmp. It should not impact the happy path

Amxx:
Final numbers image

PR #5785, opened 2025-07-06
Formal verification of Account (7702+7579)

Amxx:


certora/diff/account_extensions_draft-AccountERC7579.sol.patch
```diff
@@ -0,0 +1,14 @@
...
++++ account/extensions/draft-AccountERC7579.sol	2025-07-08 18:43:59.413714889 +0200
+@@ -308,8 +308,9 @@
+      * the ERC-2771 format.
+      */
+     function _fallback() internal virtual returns (bytes memory) {
+-        address handler = _fallbackHandler(msg.sig);
+-        require(handler != address(0), ERC7579MissingFallbackHandler(msg.sig));
++        bytes4 selector = bytes4(msg.data[0:4]);
++        address handler = _fallbackHandler(selector);
++        require(handler != address(0), ERC7579MissingFallbackHandler(selector));
```
Amxx, line 22: This is necessary for the prover to understand that the selector being looked for matches the data being passed the the fallback handler. Don't ask me how many hours I wasted figuring that out !

certora/specs/Account.spec
```diff
@@ -0,0 +1,182 @@
...
+
+hook DELEGATECALL(uint256 gas, address target, uint256 argsOffset, uint256 argsLength, uint256 retOffset, uint256 retLength) uint256 rc {
+    if (executingContract == currentContract) {
+        delegatecall = true;
+        delegatecall_target = target;
+        delegatecall_argsLength = argsLength;
+    }
+}
+
+rule delegatecallOpcodeRule(
```
Amxx, line 157: This rule doesn't work. Some function can trigger CALL that HAVOC the ghosts ...

Amxx:
These specs pass ! https://prover.certora.com/output/71959/de382212ccb3440e92693f187ca74e93?anonymousKey=5f06da9b8474ce1aaa13d57841d641d64db70476

@ernestognw WDYT ?
@arr00 @james-toussaint @gonzaotc, you should try to have a look. Its ok if you don't understand. Ask questions !

certora/specs/Account.spec
```diff
@@ -0,0 +1,369 @@
...
+    assert !isModuleInstalledBefore && isModuleInstalledAfter;
+
+    // No side effect on other modules
+    assert isOtherModuleInstalledBefore != isOtherModuleInstalledAfter => (
+        (
+            moduleTypeId == otherModuleTypeId &&
+            module == otherModule
+        ) || (
+            // when a fallback module is installed, the 0 module is "removed" for that selector
+            otherModuleTypeId == 3 && // fallback
```
Amxx, line 157:
```suggestion
            moduleTypeId == 3 &&
            otherModuleTypeId == 3 && // fallback
```

certora/specs/Account.spec
```diff
@@ -0,0 +1,369 @@
...
+    assert isModuleInstalledBefore && !isModuleInstalledAfter;
+
+    // No side effect on other modules
+    assert isOtherModuleInstalledBefore != isOtherModuleInstalledAfter => (
+        (
+            moduleTypeId == otherModuleTypeId &&
+            module == otherModule
+        ) || (
+            // when a fallback module is uninstalled, the 0 module is "added" for that selector
+            otherModuleTypeId == 3 && // fallback
```
Amxx, line 195:
```suggestion
            moduleTypeId == 3 &&
            otherModuleTypeId == 3 && // fallback
```

certora/specs/Account.spec
```diff
@@ -0,0 +1,369 @@
...
+rule validateUserOpPayment(env e){
+    Account.PackedUserOperation userOp;
+    bytes32 userOpHash;
+    uint256 missingAccountFunds;
+
+    uint256 balanceBefore = nativeBalances[currentContract];
+    validateUserOp(e, userOp, userOpHash, missingAccountFunds);
+    uint256 balanceAfter = nativeBalances[currentContract];
+
+    assert balanceBefore - balanceAfter <= missingAccountFunds;
```
james-toussaint, line 487: Could we check that all slots remain unchanged?

certora/specs/Account.spec
```diff
@@ -0,0 +1,369 @@
...
+rule callInstallModule(env e, uint256 moduleTypeId, address module, bytes initData) {
+    require calls == 0;
+
+    installModule(e, moduleTypeId, module, initData);
+
+    assert calls             == 1;
+    assert lastcall_target   == module;
+    assert lastcall_selector == 0x6d61fe70; // onInstall(bytes)
+    assert lastcall_value    == 0;
+}
```
james-toussaint, line 414: Isn't already verified here?
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/a6ecba897aa8f07d148191c1553690461446a325/certora/specs/Account.spec#L245-L250

certora/specs/Account.spec
```diff
@@ -0,0 +1,369 @@
...
+
+rule moduleManagementRule(env e, method f, calldataarg args, uint256 moduleTypeId, address module, bytes additionalContext)
+    filtered { f -> !f.isView }
+{
+    bytes context;
+    require context.length == 0;
+
+    bool isEntryPoint = e.msg.sender == entryPoint();
+    bool isEntryPointOrSelf = e.msg.sender == entryPoint() || e.msg.sender == currentContract;
+    bool isExecutionModule = isModuleInstalled(2, e.msg.sender, context);
```
james-toussaint: Very minor, maybe we can create 3 definitions and call them here and there: `isEntryPoint()`, `isEntryPointOrSelf()` & `isExecutionModule(context)`

certora/specs/Account.spec
Amxx: I'm not sure if that is possible.

certora/specs/Account.spec
Amxx: That part is not verifying the target (we cannot extract it from the args).

This rule's point is mostly to check the target of the call is the module. All other checks (calls == 1, selector, value) are duplicated. Would you remove them ?

PR #5786, opened 2025-07-07
Update FV specs and script

Amxx:
Fixes:
- [x] ERC20Wrapper
- [x] AccessManager
- [ ] AccessManaged: https://discord.com/channels/795999272293236746/1391760909780914316

Todo:
- [ ] `CERTORAKEY`
- [ ] `wait_for_results`


certora/specs/AccessControlDefaultAdminRules.spec
```diff
@@ -70,8 +70,11 @@ invariant defaultAdminRoleAdminConsistency()
 │ Invariant: owner is the defaultAdmin                                                                                │
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
-invariant ownerConsistency()
+// Writting this as an invariant would be flagged by Certora as trivial. Writting it as a rule is just as valid: we
```
Amxx:
```suggestion
// Writing this as an invariant would be flagged by Certora as trivial. Writing it as a rule is just as valid: we
```

certora/specs/AccessManager.spec
```diff
@@ -130,24 +130,35 @@ rule getAdminRestrictions(env e, bytes data) {
...
-        assert delay == (
-            (restricted && selector == to_bytes4(sig:updateAuthority(address,address).selector              )) ||
-            (restricted && selector == to_bytes4(sig:setTargetClosed(address,bool).selector                 )) ||
-            (restricted && selector == to_bytes4(sig:setTargetFunctionRole(address,bytes4[],uint64).selector))
-            ? getTargetAdminDelay(e, getFirstArgumentAsAddress(data))
-            : 0
-        );
+        assert restricted == false;
+        assert roleId     == getTargetFunctionRole(currentContract, selector);
+        assert delay      == 0;
```
ernestognw, line 161: I think this version is clearer and more succinct

```suggestion
        assert restricted == isOnlyAuthorized(selector);

        assert roleId == (
            (restricted && selector == to_bytes4(sig:grantRole(uint64,address,uint32).selector)) ||
            (restricted && selector == to_bytes4(sig:revokeRole(uint64,address).selector      ))
            ? getRoleAdmin(getFirstArgumentAsUint64(data))
            : (restricted ? ADMIN_ROLE() : getTargetFunctionRole(currentContract, selector))
        );

        assert delay == (
            (restricted && selector == to_bytes4(sig:updateAuthority(address,address).selector              )) ||
            (restricted && selector == to_bytes4(sig:setTargetClosed(address,bool).selector                 )) ||
            (restricted && selector == to_bytes4(sig:setTargetFunctionRole(address,bytes4[],uint64).selector))
            ? getTargetAdminDelay(e, getFirstArgumentAsAddress(data))
            : 0
        );
```

certora/specs/EnumerableMap.spec
ernestognw, line 1: The `atUniqueness` invariant is vacuous because the indexes might be out of bounds, so the static analyzer flags the rule for all those values. We should do:

```
invariant atUniqueness(uint256 index1, uint256 index2)
    (index1 < length() && index2 < length()) =>
    (index1 == index2 <=> key_at(index1) == key_at(index2))
    {
        preserved {
            requireInvariant consistencyIndex(index1);
            requireInvariant consistencyIndex(index2);
        }
        preserved remove(bytes32 key) {
            requireInvariant atUniqueness(index1, require_uint256(length() - 1));
            requireInvariant atUniqueness(index2, require_uint256(length() - 1));
        }
    }
```

certora/specs/EnumerableMap.spec
```diff
@@ -75,19 +75,15 @@ invariant atUniqueness(uint256 index1, uint256 index2)
...
 
 invariant consistencyKey(bytes32 key)
-    contains(key) => (
-        _positionOf(key) > 0 &&
-        _positionOf(key) <= length() &&
-        key_at(require_uint256(_positionOf(key) - 1)) == key
-    )
+    (contains(key) <=> key_at(require_uint256(_positionOf(key) - 1)) == key) && _positionOf(key) <= length()
     {
         preserved remove(bytes32 otherKey) {
```
ernestognw, line 88: This is vacuous too for `key` arguments out of bounds, so we need the following:


```suggestion
        preserved {
            require lengthSanity();
        }
        preserved remove(bytes32 key) {
```

certora/specs/EnumerableMap.spec
```diff
@@ -75,19 +75,15 @@ invariant atUniqueness(uint256 index1, uint256 index2)
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 invariant consistencyIndex(uint256 index)
-    index < length() => to_mathint(_positionOf(key_at(index))) == index + 1
+    index < length() <=> to_mathint(_positionOf(key_at(index))) == index + 1
```
ernestognw: This makes the rule vacuos because when index >= length(), the left side is false but the right side reverts.

```suggestion
    index < length() => to_mathint(_positionOf(key_at(index))) == index + 1
```

certora/specs/EnumerableMap.spec
```diff
@@ -75,19 +75,15 @@ invariant atUniqueness(uint256 index1, uint256 index2)
...
         }
     }
 
 invariant consistencyKey(bytes32 key)
-    contains(key) => (
-        _positionOf(key) > 0 &&
-        _positionOf(key) <= length() &&
-        key_at(require_uint256(_positionOf(key) - 1)) == key
-    )
+    (contains(key) <=> key_at(require_uint256(_positionOf(key) - 1)) == key) && _positionOf(key) <= length()
```
ernestognw: Similar to `consistencyIndex`

```suggestion
    (contains(key) => key_at(require_uint256(_positionOf(key) - 1)) == key) && _positionOf(key) <= length()
```

certora/specs/EnumerableSet.spec
```diff
@@ -58,19 +58,15 @@ invariant atUniqueness(uint256 index1, uint256 index2)
...
         }
     }
 
 invariant consistencyKey(bytes32 key)
-    contains(key) => (
-        _positionOf(key) > 0 &&
-        _positionOf(key) <= length() &&
-        at_(require_uint256(_positionOf(key) - 1)) == key
-    )
+    (contains(key) <=> at_(require_uint256(_positionOf(key) - 1)) == key) && _positionOf(key) <= length()
```
ernestognw:
```suggestion
    (contains(key) => at_(require_uint256(_positionOf(key) - 1)) == key) && _positionOf(key) <= length()
```

certora/specs/EnumerableSet.spec
```diff
@@ -58,19 +58,15 @@ invariant atUniqueness(uint256 index1, uint256 index2)
 └─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 */
 invariant consistencyIndex(uint256 index)
-    index < length() => _positionOf(at_(index)) == require_uint256(index + 1)
+    index < length() <=> _positionOf(at_(index)) == require_uint256(index + 1)
```
ernestognw:
```suggestion
    index < length() => _positionOf(at_(index)) == require_uint256(index + 1)
```

ernestognw:
Merged this PR into #5844. We need the branch of the changes to belong to the repository in order to run the Formal Verification CI properly with access to the CERTORAKEY in secrets

certora/specs/AccessManager.spec
Amxx: I think this is a matter of personal preference. They express the same thing in different ways.

One is saying:

Let take all the possible function selector and group them in categories. For each category we verify that the output (restricted, roleId, delay) is what we expect.

The other is saying "here is a rule for what the roleId should be, depending on a combination of factor (including but not limited to the function selector).

The second option is indeed shorter, but I don't think its clearer or more readable. There is more intricated logic, with nested ternary operators. Option one is longer, but IMO the logic is more clearly segmented.

@james-toussaint How do you feel, given that you are "new" to FV with certora, is there one syntax you find clearer / more readable ?

certora/specs/EnumerableMap.spec
Amxx: when one of the side reverts the rule should not be ignore, shouldn't it ?

In the past we've always ignored cases where (part of) the logic revert. Maybe that was not always a conscious choice.

In that particular case, I would agree that `=>` is all we need, and that the reverse implication is probably not needed, so if it makes the verification clearner, we can do that change !

certora/specs/EnumerableMap.spec
Amxx: I don't observe that. I've been running without that preserve and the prover did not find any issue with this property

Amxx:
Closed in favor of #5844

PR #5790, opened 2025-07-08
Cause _addSigners to revert if it triggers a totalWeight overflow

Amxx:
#### PR Checklist

- [x] Tests
- [ ] Documentation

### DIFF
--- a/contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
+++ b/contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
@@ -140,7 +140,7 @@ abstract contract MultiSignerERC7913Weighted is MultiSignerERC7913 {
         super._addSigners(newSigners);
 
         // This will revert if the new signers cause an overflow
-        _validateReachableThreshold
+        _validateReachableThreshold();
     }
 
     /**
--- a/contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
+++ b/contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
@@ -140,7 +140,7 @@ abstract contract MultiSignerERC7913Weighted is MultiSignerERC7913 {
         super._addSigners(newSigners);
 
         // This will revert if the new signers cause an overflow
-        totalWeight();
+        _validateReachableThreshold
     }
 
     /**
--- a/contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
+++ b/contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
@@ -130,6 +130,20 @@ abstract contract MultiSignerERC7913Weighted is MultiSignerERC7913 {
     }
 
     /**
+     * @dev See {MultiSignerERC7913-_addSigners}.
+     *
+     * In cases where {totalWeight} is almost `type(uint64).max` (due to a large `_totalExtraWeight`), adding new
+     * signers could cause the {totalWeight} computation to overflow. Adding a {totalWeight} calls after the new
+     * signers are added ensures no such overflow happens.
+     */
+    function _addSigners(bytes[] memory newSigners) internal virtual override {
+        super._addSigners(newSigners);
+
+        // This will revert if the new signers cause an overflow
+        totalWeight();
+    }
+
+    /**
      * @dev See {MultiSignerERC7913-_removeSigners}.
      *
      * Just like {_addSigners}, this function does not emit {ERC7913SignerWeightChanged} events. The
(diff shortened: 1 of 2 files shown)


contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
```diff
@@ -129,6 +129,20 @@ abstract contract MultiSignerERC7913Weighted is MultiSignerERC7913 {
...
+     *
+     * In cases where {totalWeight} is almost `type(uint64).max` (due to a large `_totalExtraWeight`), adding new
+     * signers could cause the {totalWeight} computation to overflow. Adding a {totalWeight} calls after the new
+     * signers are added ensures no such overflow happens.
+     */
+    function _addSigners(bytes[] memory newSigners) internal virtual override {
+        super._addSigners(newSigners);
+
+        // This will revert if the new signers cause an overflow
+        totalWeight();
```
ernestognw: I feel it's weird to call `totalWeight`. What if we check that `newSigners.length + totalWeight() < type(uint64).max`?

Amxx: it would be `<=`, but yes, if we would have to do that before the super call. if would be fine (I guess)

Amxx: note that if would do more operations, so it would be (very slightly) more expensive

ernestognw: Yeah true. I would be worried someone may override `totalWeight` and then the SafeCast check won't be performed here. So imo we should either:

1) Document that overriding `totalWeight` has side effects in `_addSigners`

```diff
-    /// @dev Gets the total weight of all signers.
+    /**
+     * @dev Gets the total weight of all signers.
+     *
+     * NOTE: Overriding this function without using {SafeCast-toUint64} may skip overflow validation
+     * of the total weight in {_addSigners} as a side effect.
+     */
     function totalWeight() public view virtual returns (uint64) {
         return (getSignerCount() + _totalExtraWeight).toUint64();
     }
```

2) Abstract away validation into an internal function, so that any override to `totalWeight` clarifies that validation maye be skipped

```diff
-    /// @dev Gets the total weight of all signers.
+    /**
+     * @dev Gets the total weight of all signers.
+     *
+     * NOTE: Overriding this function without calling `super._validateAndReturnTotalWeight`
+     * may skip overflow validation of the total weight.
+     */
     function totalWeight() public view virtual returns (uint64) {
-        return (getSignerCount() + _totalExtraWeight).toUint64();
+        return _validateAndReturnTotalWeight();
     }

     function _addSigners(bytes[] memory newSigners) internal virtual override {
         super._addSigners(newSigners);
-
-        // This will revert if the new signers cause an overflow
-        totalWeight();
+        _validateAndReturnTotalWeight();
     }

+
+    function _validateAndReturnTotalWeight() internal view virtual returns (uint64) {
+        return (getSignerCount() + _totalExtraWeight).toUint64();
+    }
 }
```

wdyt?

Amxx: We can also just call _validateReachableThreshold()

Again, a bit more expensive (but nothing terrible in the context of adding signers) and very explicit as to what is happening!

I'd avoid adding more functions.

ernestognw: I like that option better 👍🏻

Amxx:
```suggestion
        super._addSigners(newSigners);

        // This will revert if the new signers cause an overflow
        _validateReachableThreshold
```

contracts/utils/cryptography/signers/MultiSignerERC7913Weighted.sol
```diff
@@ -129,6 +129,20 @@ abstract contract MultiSignerERC7913Weighted is MultiSignerERC7913 {
...
+     *
+     * In cases where {totalWeight} is almost `type(uint64).max` (due to a large `_totalExtraWeight`), adding new
+     * signers could cause the {totalWeight} computation to overflow. Adding a {totalWeight} calls after the new
+     * signers are added ensures no such overflow happens.
+     */
+    function _addSigners(bytes[] memory newSigners) internal virtual override {
+        super._addSigners(newSigners);
+
+        // This will revert if the new signers cause an overflow
+        _validateReachableThreshold
```
Amxx:
```suggestion
        _validateReachableThreshold();
```

ernestognw (approved):

PR #5792, opened 2025-07-09
Add extra memory manipulation functions

ernestognw:
Depends on #5189

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/Memory.sol
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// @dev Returns a `Pointer` to the `value`'s header (i.e. includes the length word).
+    function asPointer(bytes memory value) internal pure returns (Pointer) {
+        bytes32 ptr;
+        assembly ("memory-safe") {
+            ptr := value
+        }
+        return asPointer(ptr);
+    }
```
0xClandestine, line 96:
```suggestion
    /// @dev Returns a `Pointer` to the `value`'s header (i.e. includes the length word).
    function asPointer(bytes memory value) internal pure returns (Pointer ptr) {
        assembly ("memory-safe") {
            ptr := value
        }
    }
```

Amxx:
In #5680, I introduced an approach based on "memory slices", similar to calldata slice. That may be a (safer?) alternate to pure memory pointers

Amxx:
Replaced by the Memory.Slices from #5680

PR #5497, opened 2025-02-11
fix: improve solhint custom rules and remove TODOs

crStiv:
This commit improves the solhint custom rules by:
1. Expanding visibility check for constant variables to include internal constants
2. Re-enabling and fixing the no-external-virtual rule with:
   - Skip checks for constructors and interfaces
   - Improved error message with actionable suggestion
   - Fixed comparison operator

Both TODO comments have been addressed while maintaining the codebase's
coding standards and improving the overall code quality checks.

PR #5794, opened 2025-07-10
Update OZ solhint custom rules [WIP]

gonzaotc:
1. Removes `interface-names` custom rule which is now covered by solhint included 'interface-starts-with-i'
2. Add `interface-only-external-functions` rule.
3. Prevents leading underscore in `constants` and `immutables`, which was previously only prevented on `private constants`
4. Prevents leading underscore in public an external functions.
5. Prevents external virtual functions. 

Note that 2, 3, and 4 require breaking changes that can be potential candidates for 6.0 

Ideally I would like to add another rule to verify that Interfaces can only inherit interfaces.

### DIFF
--- a/scripts/solhint-custom/index.js
+++ b/scripts/solhint-custom/index.js
@@ -23,11 +23,11 @@ class Base {
 
 module.exports = [
   class extends Base {
-    static ruleId = 'interface-names';
+    static ruleId = 'interface-only-external-functions';
 
-    ContractDefinition(node) {
-      if (node.kind === 'interface' && !/^I[A-Z]/.test(node.name)) {
-        this.error(node, 'Interface names should have a capital I prefix');
+    FunctionDefinition(node) {
+      if (node.parent.kind === 'interface' && node.visibility !== 'external') {
+        this.error(node, 'Interface functions must be external');
       }
     }
   },
@@ -47,20 +47,36 @@ module.exports = [
     static ruleId = 'leading-underscore';
 
     VariableDeclaration(node) {
-      if (node.isDeclaredConst) {
-        // TODO: expand visibility and fix
-        if (node.visibility === 'private' && /^_/.test(node.name)) {
-          this.error(node, 'Constant variables should not have leading underscore');
+        if (node.isDeclaredConst || node.isImmutable) {
+            if (/^_/.test(node.name)) {
+              this.error(node, 'Constant and immutable variables should not have leading underscore');
+            }
         }
-      } else if (node.visibility === 'private' && !/^_/.test(node.name)) {
-        this.error(node, 'Non-constant private variables must have leading underscore');
+        else {
+          if (node.isStateVar) {
+            if (node.visibility === 'private' || node.visibility === 'internal') {
+              if (!/^_/.test(node.name)) {
+                this.error(node, 'Private and internal state variables must have leading underscore');
+              }
+            }
+            else {
+                if (/^_/.test(node.name)) {
+                  this.error(node, 'Public state variables should not have leading underscore');
+                }
+            }
+        }
       }
     }
 
     FunctionDefinition(node) {
-      if (node.visibility === 'private' || (node.visibility === 'internal' && node.parent.kind !== 'library')) {
+      if (node.visibility === 'private') {
         if (!/^_/.test(node.name)) {
-          this.error(node, 'Private and internal functions must have leading underscore');
+          this.error(node, 'Private functions must have leading underscore');
+        }
+      }
+      if (node.visibility === 'internal' && node.parent.kind !== 'library') {
+        if (!/^_/.test(node.name)) {
+          this.error(node, 'Non-library internal functions must have leading underscore');
         }
       }
       if (node.visibility === 'internal' && node.parent.kind === 'library') {
@@ -68,17 +84,21 @@ module.exports = [
           this.error(node, 'Library internal functions should not have leading underscore');
         }
       }
+      if (node.visibility === 'public' || node.visibility === 'external') {
+        if (/^_/.test(node.name)) {
+          this.error(node, 'Public and external functions should not have leading underscore');
+        }
+      }
     }
   },
 
-  // TODO: re-enable and fix
-  // class extends Base {
-  //   static ruleId = 'no-external-virtual';
-  //
-  //   FunctionDefinition(node) {
-  //     if (node.visibility == 'external' && node.isVirtual) {
-  //       this.error(node, 'Functions should not be external and virtual');
-  //     }
-  //   }
-  // },
+  class extends Base {
+    static ruleId = 'no-external-virtual';
+  
+    FunctionDefinition(node) {
+      if (node.visibility == 'external' && node.isVirtual) {
+        this.error(node, 'Functions should not be external and virtual');
+      }
+    }
+  },
 ];


Amxx:
For reference, there is already this PR that deals with linter rules: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5497

Amxx (commented):
You are removing the `interface-names` rule. Why?
I think it should stay

Amxx (commented):
How efficient/explicit is `/^_/.test(node.name)` compared to `node.name.startsWith('_')`

contracts/utils/Multicall.sol
```diff
@@ -23,7 +23,7 @@ abstract contract Multicall is Context {
      * @dev Receives and executes a batch of function calls on this contract.
      * @custom:oz-upgrades-unsafe-allow-reachable delegatecall
      */
-    function multicall(bytes[] calldata data) external virtual returns (bytes[] memory results) {
+    function multicall(bytes[] calldata data) public virtual returns (bytes[] memory results) {
```
ernestognw, line 26: If I remember correctly, this is purposely external since it doesn't have an internal use. Users could just call the functions they want directly

ernestognw (commented):
I'm curious about the `external -> public` changes. So far I thought we already made all functions public when needed, so I would assume they were intentional, can we confirm reviewing the history to make sure we're not missing a deliberate decision?

contracts/utils/Multicall.sol
Amxx: By default, all public facing functions should be public so that overrides can call super.

If we have exceptions, the questions we need to answer are:
1. is there any reason we think overrides won't need to call super?
2. is there any risks in making that internally callable?

I'm not sure what override we would want to do, but I'd say calling super would be reasonnable here ... so really the question that remains is 2.

If function A internally calls `multicall(...)` (instead of doing `this.multicall(...)`) that would allow arbitrary function execution with the same caller as A's initial context. That is also what we would achieve by directly calling the corresponding function (assuming they are public and not external).

But yes, let review all the external->public changes, and we can disable the rule selectivelly if we have a good reason to keep the functions externals

scripts/solhint-custom/index.js
```diff
@@ -14,20 +14,20 @@ class Base {
...
 
-    ContractDefinition(node) {
-      if (node.kind === 'interface' && !/^I[A-Z]/.test(node.name)) {
-        this.error(node, 'Interface names should have a capital I prefix');
+    FunctionDefinition(node) {
+      if (node.parent.kind === 'interface') {
+        this.require(node.visibility === 'external', node, 'Interface functions must be external');
       }
     }
   },
```
arr00: Isn't this enforced by the compiler?

scripts/solhint-custom/index.js
```diff
@@ -47,38 +50,63 @@ module.exports = [
...
+        switch (node.visibility) {
+          case 'private':
+            this.require(node.name.startsWith('_'), node, 'Private state variables must have leading underscore');
+            break;
+          case 'internal':
+            this.require(node.name.startsWith('_'), node, 'Internal state variables must have leading underscore');
+            break;
+          case 'public':
+            this.require(!node.name.startsWith('_'), node, 'Public state variables should not have leading underscore');
+            break;
```
arr00, line 66: how will this ever be reached?

scripts/solhint-custom/index.js
```diff
@@ -47,38 +50,63 @@ module.exports = [
     static ruleId = 'leading-underscore';
 
     VariableDeclaration(node) {
+      // TODO: do we want that rule ? Should no immutable variable have a prefix regardless of visibility ?
```
arr00: I think I like to treat immutables similarly to constants but not a strong feeling.

scripts/solhint-custom/index.js
arr00: ah immutable?

scripts/solhint-custom/index.js
```diff
@@ -47,38 +50,63 @@ module.exports = [
...
+          break;
+        case 'public':
+          this.require(!node.name.startsWith('_'), node, 'Public functions should not have leading underscore');
+          break;
+        case 'internal':
+          this.require(
+            node.name.startsWith('_') !== (node.parent.kind === 'library'),
+            node,
+            node.parent.kind === 'library'
+              ? 'Library internal functions should not have leading underscore'
```
arr00, line 84: do we have a rule for non-internal library functions?

scripts/solhint-custom/index.js
```diff
@@ -47,38 +50,63 @@ module.exports = [
...
-  // },
+  class extends Base {
+    static ruleId = 'no-external-virtual';
+
+    FunctionDefinition(node) {
+      if (node.visibility == 'external') {
+        this.require(
+          node.isReceiveEther || node.isFallback || !node.isVirtual,
+          node,
+          'Functions should not be external and virtual',
```
arr00: we are allowing fallback and receiver to be virtual now?

scripts/solhint-custom/index.js
Amxx: indeed it is already checked.

scripts/solhint-custom/index.js
Amxx: yes

scripts/solhint-custom/index.js
Amxx: non internal library function will be handeled like similar function in contracts:
- if public, line 71 applies and they must not have a `_` prefix
- if private, line 83 applies and they must have a `_` prefix

scripts/solhint-custom/index.js
Amxx: We have virtual receive functions. For example, the receive function in `contracts/governance/Governor.sol` is virtual so that someone could override it (to trigger an event?)

Amxx: What this rule says is:
- external virtual function are not great, because when you override them you cannot call super
- we want to prevent that by forbitting the two to be used together. Either you function is virtual (so it can be overriden) but then you should make it public so super can be called ... or your function is external non virtual and that is good
- there is an exception for receive and fallback functions. These have to be external (they cannot be public) and we still want to have them virtual so they can be customized in some cases

scripts/solhint-custom/index.js
```diff
@@ -47,38 +40,63 @@ module.exports = [
...
-  //     }
-  //   }
-  // },
+  class extends Base {
+    static ruleId = 'no-external-virtual';
+
+    FunctionDefinition(node) {
+      if (node.visibility == 'external') {
+        this.require(
+          node.isReceiveEther || node.isFallback || !node.isVirtual,
```
Amxx:
```suggestion
      if (node.visibility == 'external' && node.isVirtual) {
        this.require(
          node.isReceiveEther || node.isFallback,
```

scripts/solhint-custom/index.js
Amxx: Note, if we enable that rule, we currently get
```
contracts/finance/VestingWallet.sol
  41:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  42:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/finance/VestingWalletCliff.sol
  17:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/governance/extensions/GovernorTimelockAccess.sol
  65:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/governance/extensions/GovernorVotes.sol
  16:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/metatx/ERC2771Context.sol
  23:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/proxy/beacon/BeaconProxy.sol
  25:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/proxy/transparent/TransparentUpgradeableProxy.sol
  67:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/proxy/utils/UUPSUpgradeable.sol
  23:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/token/ERC20/extensions/ERC20Capped.sol
  12:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/token/ERC20/extensions/ERC20Wrapper.sol
  22:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/token/ERC20/extensions/ERC4626.sol
  73:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  74:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/token/ERC721/extensions/ERC721Wrapper.sol
  17:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore

contracts/utils/cryptography/EIP712.sol
  42:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  43:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  44:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  46:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  47:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  49:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
  50:5  error  Immutable variables should not have leading underscore  openzeppelin/leading-underscore
```

Amxx: Just looking at `contracts/finance/VestingWallet.sol`, the two variables are 
```solidity
uint64 private immutable _start;
uint64 private immutable _duration;
```

If we rename them to `start` and `duration`, we would get a conflict with `function start() public view virtual returns (uint256)` and the equivalent `duration` function. We could make the immutable public, and remove the function, but then the value would not be overridable anymore.

IMO we should not do such a big change.

scripts/solhint-custom/index.js
arr00: meaning do we want to error if there is a non-internal lib function?

Amxx: private library function are definitelly ok.

Public ones ? We don't use them right now. I'm not sure if we want to forgid them. If we want to, I'd create a dedicated rule for that instead of pushing that into a "name format rule"

Amxx: added https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5794/commits/01d53b16c8c8f50eea92eb1ed0e4499c4280291a

scripts/solhint-custom/index.js
arr00: I would make them caps snake case too. but yes a large change im ok with leaving as is

scripts/solhint-custom/index.js
```diff
@@ -47,38 +40,73 @@ module.exports = [
...
+      // TODO: do we want that rule ? Should no immutable variable have a prefix regardless of visibility ?
+      //
+      // else if (node.isImmutable) {
+      //   this.require(!node.name.startsWith('_'), node, 'Immutable variables should not have leading underscore');
+      // }
       if (node.isDeclaredConst) {
-        // TODO: expand visibility and fix
-        if (node.visibility === 'private' && /^_/.test(node.name)) {
-          this.error(node, 'Constant variables should not have leading underscore');
+        this.require(!node.name.startsWith('_'), node, 'Constant variables should not have leading underscore');
```
arr00: should we require screaming snake case?

scripts/solhint-custom/index.js
```diff
@@ -47,38 +40,73 @@ module.exports = [
...
-  // },
+  class extends Base {
+    static ruleId = 'no-public-library';
+
+    FunctionDefinition(node) {
+      if (node.parent.kind === 'library') {
+        this.require(
+          node.visibility === 'internal' || node.visibility === 'private',
+          node,
+          'Library functions should not be internal or private',
```
arr00:
```suggestion
          'Library functions should be internal or private',
```

arr00 (approved):

scripts/solhint-custom/index.js
Amxx: There is already a `const-name-snakecase` in solhint. See line 5 of `solhint.config.js`

Amxx: If we want to do immutable snakecase, there is also a `immutable-vars-naming` we could enable.

gonzaotc:
Seems like the consensus on immutables is to be more relaxed and leave them as they are (since there are 20+ cases), and I’m fine with that.

The only pending item I see is:

> But yes, let’s review all the external → public changes, and disable the rule selectively if we have good reasons to keep some functions external.

contracts/proxy/utils/UUPSUpgradeable.sol
```diff
@@ -71,7 +71,7 @@ abstract contract UUPSUpgradeable is IERC1822Proxiable {
      * bricking a proxy that upgrades to it, by delegating to itself until out of gas. Thus it is critical that this
      * function revert if invoked through a proxy. This is guaranteed by the `notDelegated` modifier.
      */
-    function proxiableUUID() public view virtual notDelegated returns (bytes32) {
+    function proxiableUUID() external view notDelegated returns (bytes32) {
         return ERC1967Utils.IMPLEMENTATION_SLOT;
     }
```
arr00, line 76: just confirming this means there is never a valid reason to use a different slot?

Amxx: This is not used as a slot, we use the (constant slot, defined in ERC-1967) as a "magic value". That way, we can identify compatibility. A bit like the bytes4 returns by `IERC721Receiver.onERC721Received`

Amxx: Cf: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/UUPSUpgradeable.sol#L137-L147

arr00: oh interesting. Thanks for the context

PR #5798, opened 2025-07-15
Add warning about Clones pointing to implementation with no code

Amxx:
Related issue: https://github.com/coinbase/smart-wallet/pull/117

### DIFF
--- a/contracts/proxy/Clones.sol
+++ b/contracts/proxy/Clones.sol
@@ -24,6 +24,11 @@ library Clones {
      * @dev Deploys and returns the address of a clone that mimics the behavior of `implementation`.
      *
      * This function uses the create opcode, which should never revert.
+     *
+     * WARNING: This function does NOT check that `implementation` contains code. If that is not the case, creation of
+     * the clone will succeed, and any following initialization call risks silently failing (no code is executed but
+     * call is seen as successful) if it doesn't return any data. This would leave the clone uninitialize, with a risk
+     * of malicious initialization by a third party after the implementation contract is deployed.
      */
     function clone(address implementation) internal returns (address instance) {
         return clone(implementation, 0);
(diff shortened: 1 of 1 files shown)


contracts/proxy/Clones.sol
```diff
@@ -24,6 +24,11 @@ library Clones {
      * @dev Deploys and returns the address of a clone that mimics the behavior of `implementation`.
      *
      * This function uses the create opcode, which should never revert.
+     *
+     * WARNING: This function does NOT check that `implementation` contains code. If that is not the case, creation of
+     * the clone will succeed, and any following initialization call risks silently failing (no code is executed but
+     * call is seen as successful) if it doesn't return any data. This would leave the clone uninitialize, with a risk
+     * of malicious initialization by a third party after the implementation contract is deployed.
```
ernestognw: This is clearer imo

```suggestion
     * WARNING: This function does not check if `implementation` has code. If it doesn't, the clone will
     * be created but uninitialized given the call to an empty implementation will succeed, allowing a
     * third party to initialize it later.
```

Amxx: I think that proposition could be miss interpreted.

In
> This function does not check if `implementation` has code. If it doesn't, [...]

You intent "it" to mean "the implementation not having code" ... but it could also be interpretend as "the function not doing the check".

I also think the second sentence is quite long.

Amxx: WDYT of 
```suggestion
     * WARNING: This function does not check if `implementation` has code. A clone that points to an address
     * without code cannot be initialized. Initialization calls may appear to be successful when, in reality, they
     * have no effect and leave the clone uninitialized, allowing a third party to initialize it later.
```

ernestognw: Yeah that seems more clear to me. Feel free to push and I'll reapprove 😄

PR #5806, opened 2025-07-20
draft-IERC4337: split EntryPoint interfaces into `IEntryPointStake` and `IEntryPointDeposit` [WIP]

gonzaotc:
Currently, IEntryPointStake contains both the methods defined in the ERC for managing stakes and also for managing deposits, whose fundamental difference would be better reflected by this new interface organization.

The ERC-4337 specification explicitly states: _["the deposit (for paying gas fees) is separate from the stake (which is locked)" ](https://eips.ethereum.org/EIPS/eip-4337#entrypoint-interface:~:text=The%20deposit%20(for%20paying%20gas%20fees)%20is%20separate%20from%20the%20stake)_

Note that there are Paymasters which require the IEntryPointDeposit interface (obligatory part of the standard) but may not require the IEntryPointStake interface, which is optional. The new interface organization allows paymaster builders to import only the interface components they need.

### DIFF
--- a/contracts/interfaces/draft-IERC4337.sol
+++ b/contracts/interfaces/draft-IERC4337.sol
@@ -107,21 +107,6 @@ interface IEntryPointNonces {
  */
 interface IEntryPointStake {
     /**
-     * @dev Returns the balance of the account.
-     */
-    function balanceOf(address account) external view returns (uint256);
-
-    /**
-     * @dev Deposits `msg.value` to the account.
-     */
-    function depositTo(address account) external payable;
-
-    /**
-     * @dev Withdraws `withdrawAmount` from the account to `withdrawAddress`.
-     */
-    function withdrawTo(address payable withdrawAddress, uint256 withdrawAmount) external;
-
-    /**
      * @dev Adds stake to the account with an unstake delay of `unstakeDelaySec`.
      */
     function addStake(uint32 unstakeDelaySec) external payable;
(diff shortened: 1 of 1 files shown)


Amxx:
For the record, This entrypoint interface is split in sections that follows the implementation of the canonical entryopoint
- [v0.6](https://github.com/eth-infinitism/account-abstraction/blob/v0.6.0/contracts/core/EntryPoint.sol)
- [v0.7](https://github.com/eth-infinitism/account-abstraction/blob/v0.7.0/contracts/core/EntryPoint.sol)
- [v0.8](https://github.com/eth-infinitism/account-abstraction/blob/v0.8.0/contracts/core/EntryPoint.sol)

The entrypoint inherits from helper contracts that are `StakeManager` and `NonceManager`. The `IEntryPointNonces` and `IEntryPointStake` interfaces correspond to what these modules implement/expose. In the reference implementation, these interfaces are called [`IStakeManager`](https://github.com/eth-infinitism/account-abstraction/blob/v0.8.0/contracts/interfaces/IStakeManager.sol) and [`INonceManager`](https://github.com/eth-infinitism/account-abstraction/blob/v0.8.0/contracts/interfaces/INonceManager.sol).

---

With this context in mind, I think the right move would be to just get rid of these "auxiliary" interfaces, and put everything directly in `IEntrypoint`.

That being said, the current definition of `IEntryPointStake` (and `IEntryPointNonces`) is something we should avoid breaking, unless really necessary. This file being draft means we don't provide the same guarantees as other files, but it also doesn't mean we don't need a strong reason to change things.

I'm not in favor of doing this change now, but I am in favor of discussing what we really want to do with this interface for 6.0

PR #5813, opened 2025-07-23
Update Math.sol (Fixes typo in NatSpec)

aviggiano:
Fixes typo in NatSpec

The NatSpec would state that the `ternary` function from the `Math` library would perform a branchless ternary evaluation for `a ? b : c`, but these variable parameters do not conform with the implementation.

### DIFF
--- a/contracts/utils/math/Math.sol
+++ b/contracts/utils/math/Math.sol
@@ -134,10 +134,10 @@ library Math {
     }
 
     /**
-     * @dev Branchless ternary evaluation for `a ? b : c`. Gas costs are constant.
+     * @dev Branchless ternary evaluation for `condition ? a : b`. Gas costs are constant.
      *
      * IMPORTANT: This function may reduce bytecode size and consume less gas when used standalone.
-     * However, the compiler may optimize Solidity ternary operations (i.e. `a ? b : c`) to only compute
+     * However, the compiler may optimize Solidity ternary operations (i.e. `condition ? a : b`) to only compute
      * one branch when needed, making this function more expensive.
      */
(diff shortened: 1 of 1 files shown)


Amxx (approved):

aviggiano:
Hey @Amxx 

I encountered this typo while refactoring my code in an attempt to reduce bytecode size, so I removed all `condition ? a : b` instances and replaced them with `Math.ternary(condition, a, b)`

However, this caused an issue that was caught by a fuzz test and which was not highlighted in the documentation. The updated NatSpec says:

```
    /**
     * @dev Branchless ternary evaluation for `condition ? a : b`. Gas costs are constant.
     *
     * IMPORTANT: This function may reduce bytecode size and consume less gas when used standalone.
     * However, the compiler may optimize Solidity ternary operations (i.e. `condition ? a : b`) to only compute
     * one branch when needed, making this function more expensive.
     */
```

This led me to believe that the only tradeoff was bytecode size vs gas, but in reality using `Math.ternary` executes both `a` and `b`, while `condition ? a : b` executes only one branch. This can lead to side effects if either `a` or `b` are not pure values.

I believe it would be best to amend the NatSpec to be clearer about that. I suggest the following text:


```
    /**
     * @dev Branchless ternary evaluation for `condition ? a : b`. Gas costs are constant.
     *
     * IMPORTANT: This function may reduce bytecode size and consume less gas when used standalone.
     * However, the compiler may optimize Solidity ternary operations (i.e. `condition ? a : b`) to only compute
     * one branch when needed, making this function more expensive, and potentially causing side-effects due to the evaluation of both `a` and `b`
     */
```

PR #5814, opened 2025-07-23
ECDSA: add parse and tryParse

Amxx:
Following discussion with @frangio, we believe that this could be usefull for someone that wants to re-implement `isValidSignatureNow` in a way taht supports both 65 and 64 bytes signatures.

We should add this support nativelly in v6.0, but for the same reason as the one mentioned in [this GHSA](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories/GHSA-4h98-2769-gh6h) this is not a change that should happen in a minor release.

```solidity
function isValidSignatureNow64or65(address signer, bytes32 hash, bytes memory signature) internal view returns (bool) {
        if (signer.code.length == 0) {
            (uint8 v, bytes32 r, bytes32 s) = ECDSA.parse(signature);
            (address recovered, ECDSA.RecoverError err, ) = ECDSA.tryRecover(hash, v, r, s);
            return err == ECDSA.RecoverError.NoError && recovered == signer;
        } else {
            return isValidERC1271SignatureNow(signer, hash, signature);
        }
    }
```

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -212,7 +212,7 @@ library ECDSA {
      * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (eip-2098)
      * signature formats. Returns 0, 0, 0 is the signature is not in a proper format.
      */
-    function parse(bytes memory signature) internal pure returns (int8 v, bytes32 r, bytes32 s) {
+    function parse(bytes memory signature) internal pure returns (uint8 v, bytes32 r, bytes32 s) {
         assembly ("memory-safe") {
             // Check the signature length
             switch mload(signature)
@@ -240,7 +240,7 @@ library ECDSA {
     /**
      * @dev Variant of {parse} that takes a signature in calldata
      */
-    function parseCalldata(bytes calldata signature) internal pure returns (int8 v, bytes32 r, bytes32 s) {
+    function parseCalldata(bytes calldata signature) internal pure returns (uint8 v, bytes32 r, bytes32 s) {
         assembly ("memory-safe") {
             // Check the signature length
             switch signature.length
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -197,6 +197,63 @@ library ECDSA {
     }
 
     /**
+     * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (eip-2098)
+     * signature formats. Returns 0, 0, 0 is the signature is not in a proper format.
+     */
+    function parse(bytes memory signature) internal pure returns (int8 v, bytes32 r, bytes32 s) {
+        assembly ("memory-safe") {
+            // Check the signature length
+            switch mload(signature)
+            // - case 65: r,s,v signature (standard)
+            case 65 {
+                r := mload(add(signature, 0x20))
+                s := mload(add(signature, 0x40))
+                v := byte(0, mload(add(signature, 0x60)))
+            }
+            // - case 64: r,vs signature (cf https://eips.ethereum.org/EIPS/eip-2098)
+            case 64 {
+                let vs := mload(add(signature, 0x40))
+                r := mload(add(signature, 0x20))
+                s := and(vs, shr(1, not(0)))
+                v := add(shr(255, vs), 27)
+            }
+            default {
+                r := 0
+                s := 0
+                v := 0
+            }
+        }
+    }
+
+    /**
+     * @dev Variant of {parse} that takes a signature in calldata
+     */
+    function parseCalldata(bytes calldata signature) internal pure returns (int8 v, bytes32 r, bytes32 s) {
+        assembly ("memory-safe") {
+            // Check the signature length
+            switch signature.length
+            // - case 65: r,s,v signature (standard)
+            case 65 {
+                r := calldataload(signature.offset)
+                s := calldataload(add(signature.offset, 0x20))
+                v := byte(0, calldataload(add(signature.offset, 0x40)))
+            }
+            // - case 64: r,vs signature (cf https://eips.ethereum.org/EIPS/eip-2098)
+            case 64 {
+                let vs := calldataload(add(signature.offset, 0x20))
+                r := calldataload(signature.offset)
+                s := and(vs, shr(1, not(0)))
+                v := add(shr(255, vs), 27)
+            }
+            default {
+                r := 0
+                s := 0
+                v := 0
+            }
+        }
+    }
+
+    /**
      * @dev Optionally reverts with the corresponding custom error according to the `error` argument provided.
      */
     function _throwError(RecoverError error, bytes32 errorArg) private pure {
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -43,11 +43,9 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
-     * NOTE: This function protects against malleability by only supporting 65 bytes long signatures, and rejecting
-     * EIP-2098 short signatures. This guarantee is DEPRECATED and will be removed in the next major release (v6.0).
-     * Developers SHOULD NOT use signatures as unique identifiers. If an operation must be marked as consumed to
-     * prevent replayability, either the `hash` (or the `hash`/`recovered` pair if multiple accounts are to sign the
-     * same hash) should be invalidated. Nonces are also a viable solution.
+     * NOTE: This function only supports 65-byte signatures. ERC-2098 short signatures are rejected. This restriction
+     * is DEPRECATED and will be removed in v6.0. Developers SHOULD NOT use signatures as unique identifiers; use hash
+     * invalidation or nonces for replay protection.
      *
      * IMPORTANT: `hash` _must_ be the result of a hash operation for the
      * verification to be secure: it is possible to craft signatures that
@@ -112,11 +110,9 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
-     * NOTE: This function protects against malleability by only supporting 65 bytes long signatures, and rejecting
-     * EIP-2098 short signatures. This guarantee is DEPRECATED and will be removed in the next major release (v6.0).
-     * Developers SHOULD NOT use signatures as unique identifiers. If an operation must be marked as consumed to
-     * prevent replayability, either the `hash` (or the `hash`/`recovered` pair if multiple accounts are to sign the
-     * same hash) should be invalidated. Nonces are also a viable solution.
+     * NOTE: This function only supports 65-byte signatures. ERC-2098 short signatures are rejected. This restriction
+     * is DEPRECATED and will be removed in v6.0. Developers SHOULD NOT use signatures as unique identifiers; use hash
+     * invalidation or nonces for replay protection.
      *
      * IMPORTANT: `hash` _must_ be the result of a hash operation for the
      * verification to be secure: it is possible to craft signatures that
@@ -209,7 +205,7 @@ library ECDSA {
     }
 
     /**
-     * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (eip-2098)
+     * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (erc-2098)
      * signature formats. Returns 0, 0, 0 is the signature is not in a proper format.
      */
     function parse(bytes memory signature) internal pure returns (uint8 v, bytes32 r, bytes32 s) {
--- a/contracts/utils/cryptography/ECDSA.sol
+++ b/contracts/utils/cryptography/ECDSA.sol
@@ -43,12 +43,11 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
-     * NOTE: This function also prevents signature malleability by only supporting 65 bytes long signatures, and
-     * rejecting eip-2098 short signatures. While this guarantee is still present, it is DEPRECATED and will be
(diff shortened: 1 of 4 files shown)


contracts/utils/cryptography/ECDSA.sol
```diff
@@ -196,6 +196,63 @@ library ECDSA {
         return recovered;
     }
 
+    /**
+     * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (eip-2098)
+     * signature formats. Returns 0, 0, 0 is the signature is not in a proper format.
+     */
+    function parse(bytes memory signature) internal pure returns (int8 v, bytes32 r, bytes32 s) {
```
frangio: For discussion (I'm undecided): should this have a special name given that, unlike the rest of this library, it allows 64-65 byte malleability? E.g., `parseMalleable`.

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -196,6 +196,63 @@ library ECDSA {
         return recovered;
     }
 
+    /**
+     * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (eip-2098)
+     * signature formats. Returns 0, 0, 0 is the signature is not in a proper format.
```
frangio: If one wants to check if decoding succeeded, is it necessary to check all three return values for 0, or is there one of them that could be checked for 0 to provide indication that decoding failed? It would be useful to document this if so.

Maybe `v` is enough?

Amxx: v should be enough.

Amxx: According to https://www.evm.codes/precompiled, any `v` that is NOT 27 or 28 will be rejected by the precompile. Similarly, `r` and `s` are expected to be in range `]0, secp256k1n[`, so either of those being 0 would also mean "obviously invalid signature".

That mean any individual value being 0 is a good enough check

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -196,6 +196,63 @@ library ECDSA {
...
+     */
```
Amxx, line 210:
```suggestion
     * 
     * NOTE: supporting both format means that any valid signature can be reformated by anyone to produce another
     * equally valid signature. This means an observer can produce a different input that will results in the same
     * output. Therefore, calls to this function should not assume any kind of "signature uniqueness".
     */
```

contracts/utils/cryptography/ECDSA.sol
Amxx: Would [that note](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5814/files#r2231118557) be good enough ?

IMO not using signatures as unique identifier is basic good practices, and I'm not sure we should complexify the name of the functions (creating possible confusion?) as a way to "help enforce" these basic good practices.

frangio: > not using signatures as unique identifier is basic good practices

Yeah I agree and we should move towards this assumption. My point is only about how this breaks from the existing library design.

How about deprecating malleability protection in all of the `ECDSA` library?

Amxx: > How about deprecating malleability protection in all of the ECDSA library?

YES !

Amxx: @frangio how does that look ?
144b0e29d9f0ff91f56f18c2f9b123b7707d2675

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -43,6 +43,13 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
+     * NOTE: This function also prevents signature malleability by only supporting 65 bytes long signature, and
+     * rejecting eip-2098 short signature. While this guarantee is still present, it is DEPRECATED and will be removed
+     * in the next major release (v6.0). Developers SHOULD NOT use signatures as unique identifier. If an operation
```
Amxx:
```suggestion
     * in the next major release (v6.0). Developers SHOULD NOT use signatures as unique identifiers. If an operation
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -106,6 +113,13 @@ library ECDSA {
...
```
Amxx:
```suggestion
     * in the next major release (v6.0). Developers SHOULD NOT use signatures as unique identifiers. If an operation
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -43,6 +43,13 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
+     * NOTE: This function also prevents signature malleability by only supporting 65 bytes long signature, and
```
Amxx:
```suggestion
     * NOTE: This function also prevents signature malleability by only supporting 65 bytes long signatures, and
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -106,6 +113,13 @@ library ECDSA {
...
```
Amxx:
```suggestion
     * NOTE: This function also prevents signature malleability by only supporting 65 bytes long signatures, and
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -43,6 +43,13 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
+     * NOTE: This function also prevents signature malleability by only supporting 65 bytes long signatures, and
+     * rejecting eip-2098 short signatures. While this guarantee is still present, it is DEPRECATED and will be
+     * removed in the next major release (v6.0). Developers SHOULD NOT use signatures as unique identifiers. If an
+     * operation must me marked as consumed to prevent replayability, either the `hash` (or the `hash`/`recovered`
+     * pair if multiple accounts are to sign the same hash) should be invalidated. Nonces are also a viable solution.
+     * Marking the signatures as consumed is very strongly discouraged.
```
frangio: This text is a little too long for my taste but it's ok. Minor changes suggested below.

The deprecation needs to be added to the changelog as well.

```suggestion
     * NOTE: This function protects against malleability by only supporting 65 bytes long signatures, and
     * rejecting EIP-2098 short signatures. This guarantee is DEPRECATED and will be
     * removed in the next major release (v6.0). Developers SHOULD NOT use signatures as unique identifiers. If an
     * operation must be marked as consumed to prevent replayability, either the `hash` (or the `hash`/`recovered`
     * pair if multiple accounts are to sign the same hash) should be invalidated. Nonces are also a viable solution.
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -43,6 +43,12 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
+     * NOTE: This function protects against malleability by only supporting 65 bytes long signatures, and rejecting
+     * EIP-2098 short signatures. This guarantee is DEPRECATED and will be removed in the next major release (v6.0).
+     * Developers SHOULD NOT use signatures as unique identifiers. If an operation must be marked as consumed to
+     * prevent replayability, either the `hash` (or the `hash`/`recovered` pair if multiple accounts are to sign the
+     * same hash) should be invalidated. Nonces are also a viable solution.
```
ernestognw: I think this is clearer and simpler:

```suggestion
     * NOTE: Only supports 65-byte signatures (ERC-2098 short signatures are rejected). This restriction is
     * DEPRECATED and will be removed in v6.0. Don't use signatures as unique identifiers; use hash invalidation
     * or nonces for replay protection.
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -106,6 +112,12 @@ library ECDSA {
...
```
ernestognw: Same

```suggestion
     * NOTE: Only supports 65-byte signatures (ERC-2098 short signatures are rejected). This restriction is
     * DEPRECATED and will be removed in v6.0. Don't use signatures as unique identifiers; use hash invalidation
     * or nonces for replay protection.
```

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -196,6 +208,63 @@ library ECDSA {
         return recovered;
     }
 
+    /**
+     * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (eip-2098)
```
ernestognw: Isn't an ERC?

```suggestion
     * @dev Parse a signature into its `v`, `r` and `s` components. Supports both 65 bytes and 64 bytes (erc-2098)
```

contracts/utils/cryptography/ECDSA.sol
ernestognw: Isn't possible to pass in an invalid signature but only a valid `v` value? I think it's not enough to make sure the parsing succeeded. Also, what does it mean that _decoding failed_?

I agree we should note it if it's useful

contracts/utils/cryptography/ECDSA.sol
```diff
@@ -43,6 +43,12 @@ library ECDSA {
      * this function rejects them by requiring the `s` value to be in the lower
      * half order, and the `v` value to be either 27 or 28.
      *
+     * NOTE: This function protects against malleability by only supporting 65 bytes long signatures, and rejecting
```
arr00, line 46:
```suggestion
     * NOTE: This function protects against malleability by only supporting 65 bytes long signatures and rejecting
```

contracts/utils/cryptography/ECDSA.sol
arr00: Prefer shorter

contracts/utils/cryptography/ECDSA.sol
arr00: Why is v a signed integer?

contracts/utils/cryptography/ECDSA.sol
luiz-lvj: @ernestognw it should be possible to pass as invalid signature but with a valid `v` or `r` for example, so only checking if one of them  is 0 could lead to false positives

luiz-lvj: I think the point of checking if `decoding failed` is to avoid calling `tryRecover` unnecessarily, which would be a lot more expensive

ernestognw: Right, I see what you mean, thanks @luiz-lvj. Imo deserves a note:

```solidity
 * NOTE: A returned `v == 0` means the signature couldn't be parsed. Consider avoid calling `tryRecover` unnecessarily 
```

contracts/utils/cryptography/ECDSA.sol
Amxx: >Why is v a signed integer?

Because I made a mistake. Thanks for spotting it !

contracts/utils/cryptography/ECDSA.sol
Amxx: A few comments:

- False positive on the "parsing success" would only lead to `tryVerify` being called with bad params, which would be identified as an invalid signature, so IMO there is no security risk.

- In case of 65 bytes long signatures, the decoding parsing might be "correct" and v could be 0. That would happen if the signature encodes invalid data (a v=0). I don't think that would quality as a parsing error.

- If we are in a situation where we parse a bytes (of arbitray length) and we get v, r, s returned to us, we have two options: inspect them to try and exit "early", or pass them to `tryRecover`/`recover` and let them handle the invalid signatures. The first options includes check that everyone will pay, including valid sigantures ! The second option makes failure more expensive (they happen later) but doesn't add any cost to the "happy path". **IMO we should optimise the "happy path" cost, which means I wouldn't add any v,r,s check between he parsing and the recovery**.

ernestognw: > we should optimise the "happy path" cost, which means I wouldn't add any v,r,s check between he parsing and the recovery.

Yes, agree with this. Just echoing fran's comment:

> It would be useful to document this if so.

Amxx: > Yes, agree with this. Just echoing fran's comment:
>>It would be useful to document this if so.

👍🏻 

So what do we want to do? I would personally not encourage devs to do such a check. Does anyone feel differently? 

If we feel like 0,0,0 being an invalid signature for all hashes is enough, then what about:

```suggestion
     * signature formats. If the signature format is not recognised, this function returns (0,0,0), which both
     * {recover} and {tryRecover} will see as invalid (for every hash).
```

ernestognw: The block started to sound repetitive. See https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5814/commits/23c393f6b5a75f27cca6b236fb0305a7d849bf9c

ernestognw (approved):

contracts/utils/cryptography/ECDSA.sol
frangio: I think I had something in mind when I asked this question that was a wrong idea. I was thinking that in some cases (like `SignatureChecker`) we'd want to check if we had a valid signature, but looking at it better we never really do that, the decoded values just go straight to `ecrecover`.

Feel free to remove the 

> ```solidity
>  * NOTE: A returned `v == 0` means the signature couldn't be parsed.
> ```

Something like this is what I had in mind, but I think I was motivated by a wrong idea.

ernestognw: Cool. Solving

frangio: > ```solidity
>  * NOTE: A returned `v == 0` means the signature couldn't be parsed.
> ```

Something like this is what I had in mind, but I think I was motivated by a wrong idea. I was thinking that in some cases (like `SignatureChecker`) we'd want to check if we had a valid signature, but looking at it better we never really do that, the decoded values just go straight to `ecrecover`.

So I don't think anyone would ever just look at `v == 0` and make a decision based on that.

PR #5816, opened 2025-07-24
Add ERC-7751 interface

Amxx:
https://eips.ethereum.org/EIPS/eip-7751 is final.

#### PR Checklist

- [x] Documentation


ernestognw:
I'm wondering whether it makes sense to have an interface for the error. Wouldn't make more sense to have a library and the library includes the error?

Amxx:
The goeal is to add support for this errors in libraries in the futur.
I still think the interface can be usefull it projects wants to decode it.

PR #5824, opened 2025-07-31
test: enable some math tests to be run by halmos

daejunpark:
Some of the math library tests are also run by halmos for verification. With the recent switch in halmos's default SMT solver from Z3 to Yices, a few more tests can now be verified as well.

The following additional tests are now supported by halmos (>=v0.3.3):
(Most of these run quickly, except testAverage2, which takes ~100s locally, and ~200s in GitHub CI.)

Math.t.sol:
- testAdd512
- testCeilDiv (requires `--solver cvc5-int`)
- testInvMod2
- testInvMod17
- testInvMod65537 (requires `--solver bitwuzla-abs`)

SignedMath.t.sol:
- testAverage2 (requires `--solver-timeout-assertion 0`)

Bytes.t.sol:
- testIndexOf (the first variant with two parameters)
- testLastIndexOf (the first variant with two parameters)

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/test/utils/math/SignedMath.t.sol
+++ b/test/utils/math/SignedMath.t.sol
@@ -48,7 +48,8 @@ contract SignedMathTest is Test {
     }
 
     // 2. more complex test, full int256 range
-    function testAverage2(int256 a, int256 b) public pure {
+    /// @custom:halmos --solver-timeout-assertion 0
+    function testSymbolicAverage2(int256 a, int256 b) public pure {
         (int256 result, int256 min, int256 max) = (
             SignedMath.average(a, b),
             SignedMath.min(a, b),
--- a/.github/workflows/formal-verification.yml
+++ b/.github/workflows/formal-verification.yml
@@ -84,3 +84,5 @@ jobs:
         run: pip install -r fv-requirements.txt
       - name: Run Halmos
         run: halmos --match-test '^symbolic|^testSymbolic' -vv
+        env:
+          HALMOS_ALLOW_DOWNLOAD: 1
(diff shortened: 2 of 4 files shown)


Amxx:
> (Most of these run quickly, except testAverage2, which takes ~100s locally, and ~3min in GitHub CI.)

I'm not sure if we want to make that change, considering how much it would slow down CI

ernestognw:
> > (Most of these run quickly, except testAverage2, which takes ~100s locally, and ~3min in GitHub CI.)
> 
> I'm not sure if we want to make that change, considering how much it would slow down CI

I think this is fine. It runs in parallel to our other tests and is still around 4 mins
In contrast, coverage is taking around 12 mins

test/utils/math/SignedMath.t.sol
```diff
@@ -48,7 +48,8 @@ contract SignedMathTest is Test {
     }
 
     // 2. more complex test, full int256 range
-    function testAverage2(int256 a, int256 b) public pure {
+    /// @custom:halmos --solver-timeout-assertion 0
```
ernestognw, line 51: Based on [this](https://github.com/a16z/halmos/blob/ba94bdc232c746b61f12eeacef6136eace81c7c2/tests/regression/halmos.toml#L154), I understand that this is just removing the timeout.

Since it's not well documented I would leave a note:


```suggestion
    /// (timeout 0 means no timeout)
    /// @custom:halmos --solver-timeout-assertion 0
    function testSymbolicAverage2(int256 a, int256 b) public pure {
```

ernestognw (approved):

PR #5826, opened 2025-07-31
Add TrieProof library

ernestognw:
Follow up from #5680 

#### PR Checklist

- [ ] Tests
- [ ] Documentation


contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,254 @@
...
+        bytes32 root
+    ) internal pure returns (bool) {
+        return verify(key, value, proof, root, EVM_TREE_RADIX);
+    }
+
+    /// @dev Same as {verify} but with a custom radix.
+    function verify(
+        bytes memory key,
+        bytes memory value,
+        bytes[] memory proof,
```
0xClandestine: Reminder that proofs are almost always provided via calldata, as such calldata should be used here for reduced gas costs.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,254 @@
...
+
+    /// @dev Split each byte in `value` into two nibbles (4 bits each).
+    function _nibbles(bytes memory value) private pure returns (bytes memory) {
+        uint256 length = value.length;
+        bytes memory nibbles_ = new bytes(length * 2);
+        for (uint256 i = 0; i < length; i++) {
+            (nibbles_[i * 2], nibbles_[i * 2 + 1]) = (value[i] & 0xf0, value[i] & 0x0f);
+        }
+        return nibbles_;
+    }
```
0xClandestine: Reminder, a much more efficient approached has been added to Solady.

https://github.com/Vectorized/solady/pull/1461

EDIT: more context https://graphics.stanford.edu/~seander/bithacks.html#InterleaveBMN

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,201 @@
...
+
+                    // Increment keyIndex by the number of nibbles consumed and continue traversal
+                    root = _getNodeId(node.decoded[1]);
+                    keyIndex += shared;
+                } else if (prefix == uint8(Prefix.LEAF_EVEN) || prefix == uint8(Prefix.LEAF_ODD)) {
+                    // Leaf node (terminal) - return its value if key matches completely
+                    if (pathRemainder.getHash() != keyRemainder.getHash()) {
+                        return (bytes32(0), ProofError.MISMATCH_LEAF_PATH_KEY_REMAINDERS);
+                    } else if (keyRemainder.length() == 0) {
+                        return (bytes32(0), ProofError.INVALID_KEY_REMAINDER);
```
james-toussaint: @Amxx I'm not sure we need that, do you have an example?
```suggestion
```

Amxx: I don't have an example. 

Note: this check was part of the code I refactored. I did not introduce it.

Amxx:
Note: I tried run the tests from https://github.com/ethereum-optimism/optimism/blob/ef970556e668b271a152124023a8d6bb5159bacf/packages/contracts-bedrock/test/libraries/trie/MerkleTrie.t.sol and I'm getting a lot of `RLPInvalidEncoding` errors

Amxx:
I'm curious weither we want to deal with returning an error Enum, as opposed to reverting.
- it makes the handling of the processProof function more difficult. We have to get both values, check the error, and then process the result. This is visible when doing foundry unit tests
```solidity
        (bytes32 value, TrieProof.ProofError err) = TrieProof.processProof(key, proof, root);
        assertEq(uint8(err), uint8(TrieProof.ProofError.NO_ERROR));
        assertEq(value, expected);
```
vs
```solidity
        assertEq(expected, TrieProof.processProof(key, proof, root));
```

- This would be nice if it we wanted to build a function that does a `processProof` with guarantee that it doesn't revert. However, there are many think that can cause this function to revert. In particular, the RLP decoding is really sensitive and can revert pretty easily.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /// @dev The radix of the Ethereum trie (hexadecimal = 16)
+    uint256 internal constant EVM_TREE_RADIX = 16;
+
+    /// @dev Number of items in leaf or extension nodes (always 2)
+    uint256 internal constant LEAF_OR_EXTENSION_NODE_LENGTH = 2;
+
+    /// @dev Verifies a `proof` against a given `key`, `value`, `and root` hash.
+    function verify(bytes memory key, bytes32 value, bytes[] memory proof, bytes32 root) internal pure returns (bool) {
```
davidperk: Is this implementation intended to only support the storage trie (where values are always 32 bytes)? Unlike the storage trie, the state, transactions, and receipts tries store arbitrary-length RLP blobs.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,214 @@
...
+        bytes[] memory proof,
+        bytes32 root
+    ) internal pure returns (bytes memory value, ProofError err) {
+        if (key.length == 0) return ("", ProofError.EMPTY_KEY);
+
+        // Expand the key
+        bytes memory keyExpanded = _nibbles(key);
+
+        bytes32 currentNodeId = root;
+        uint256 currentNodeIdLength = 32;
```
james-toussaint:
```suggestion
        bytes32 currentNodeId;
        uint256 currentNodeIdLength;
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,214 @@
...
+            // ensure we haven't overshot the key
+            if (keyIndex > keyExpanded.length) {
+                return ("", ProofError.INDEX_OUT_OF_BOUNDS);
+            }
+
+            // validates the node hashes at different levels of the proof.
+            if (keyIndex == 0) {
+                // Root node must match root hash
+                if (currentNodeIdLength != 32 || keccak256(node.encoded) != currentNodeId)
+                    return ("", ProofError.INVALID_ROOT_HASH);
```
james-toussaint:
```suggestion
                if (keccak256(node.encoded) != root) return ("", ProofError.INVALID_ROOT_HASH);
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,214 @@
...
+                    bytes1 branchKey = keyExpanded[keyIndex];
+                    (currentNodeId, currentNodeIdLength) = _getNodeId(node.decoded[uint8(branchKey)]);
+                    keyIndex += 1;
+                }
+            } else if (nodeLength == LEAF_OR_EXTENSION_NODE_LENGTH) {
+                bytes memory path = _nibbles(node.decoded[0].readBytes());
+                uint8 prefix = uint8(path[0]);
+                Memory.Slice pathRemainder = path.asSlice().slice(2 - (prefix % 2)); // Path after the prefix
+                Memory.Slice keyRemainder = keyExpanded.asSlice().slice(keyIndex); // Remaining key to match
+
```
james-toussaint, line 153:
```suggestion
                uint256 pathRemainderLength = pathRemainder.length();
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,214 @@
...
+                    pathRemainder.getHash() != keyRemainder.slice(0, pathRemainder.length()).getHash()
+                ) {
+                    return ("", ProofError.INVALID_PATH_REMAINDER);
+                }
+
+                if (prefix == uint8(Prefix.EXTENSION_EVEN) || prefix == uint8(Prefix.EXTENSION_ODD)) {
+                    // Increment keyIndex by the number of nibbles consumed and continue traversal
+                    (currentNodeId, currentNodeIdLength) = _getNodeId(node.decoded[1]);
+                    keyIndex += pathRemainder.length();
+                } else if (prefix == uint8(Prefix.LEAF_EVEN) || prefix == uint8(Prefix.LEAF_ODD)) {
```
james-toussaint:
```suggestion
                if (prefix < 2) {
                    // Extension node - Increment keyIndex by the number of nibbles consumed and continue traversal
                    (currentNodeId, currentNodeIdLength) = _getNodeId(node.decoded[1]);
                    keyIndex += pathRemainder.length();
                } else if (prefix < 4) {
```
Perhaps we should convert the `enum Prefix { .. }` declaration to a piece of doc/natspec.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @dev Extracts the node ID (hash or raw data based on size)
+     *
+     * For small nodes (encoded length < 32 bytes) the node ID is the node content itself,
+     * For larger nodes, the node ID is the hash of the encoded node data.
+     */
+    function _getNodeId(Memory.Slice node) private pure returns (bytes32 nodeId, uint256 nodeIdLength) {
+        nodeIdLength = Math.min(node.length(), 32);
+        nodeId = nodeIdLength < 32 ? node.load(0) : node.readBytes32();
```
james-toussaint: Same price than directly returning couple?

Amxx: AFAIK yes

contracts/utils/cryptography/TrieProof.sol
Amxx: If I remember correctly, this caused a stack too deep error :/

contracts/utils/cryptography/TrieProof.sol
Amxx: This change (using one < instead of two == ) would (very slightly) increasse performance but would also decrease readability. @ernestognw WDYT ?

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,213 @@
...
+
+    /// @dev Number of items in leaf or extension nodes (always 2)
+    uint256 internal constant LEAF_OR_EXTENSION_NODE_LENGTH = 2;
+
+    /// @dev Verifies a `proof` against a given `key`, `value`, `and root` hash.
+    function verify(
+        bytes memory key,
+        bytes memory value,
+        bytes[] memory proof,
+        bytes32 root
```
Amxx: Not that there is anything wrong with this order, but once we release it, we cannot change it without it being a breaking change. Are we sure that is the order we want ?

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,213 @@
...
+    ) internal pure returns (bool) {
+        (bytes memory processedValue, ProofError err) = processProof(key, proof, root);
+        return Bytes.equal(processedValue, value) && err == ProofError.NO_ERROR;
+    }
+
+    /// @dev Processes a proof for a given key and returns the processed value.
+    function processProof(
```
Amxx: https://github.com/ethereum-optimism/optimism/blob/ef970556e668b271a152124023a8d6bb5159bacf/packages/contracts-bedrock/src/libraries/trie/MerkleTrie.sol uses:
- `verifyInclusionProof` 
- `get`

IMO `verifyInclusionProof` is to much, and I prefer our `verify`. Do we want to stick with `processProof` to match the `MerkleProof` library (even though they do different things) or use a simpler/shorter name like `get` or `process` ?

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,213 @@
...
+        bytes memory key,
+        bytes[] memory proof,
+        bytes32 root
+    ) internal pure returns (bytes memory value, ProofError err) {
+        if (key.length == 0) return ("", ProofError.EMPTY_KEY);
```
Amxx: In `contracts/utils/draft-InteroperableAddress.sol` we use

```solidity
function _emptyBytesMemory() private pure returns (bytes memory result) {
    assembly ("memory-safe") {
        result := 0x60 // mload(0x60) is always 0
    }
}
```

Do we want to do the same here ?

contracts/utils/cryptography/TrieProof.sol
Amxx: when in doubt: check in remix !

test/utils/cryptography/TrieProof.test.js
```diff
@@ -0,0 +1,526 @@
...
+        key: '0xaa',
+        proof: [
+          '0xe21aa07ea462226a3dc0a46afb4ded39306d7a84d311ada3557dfc75a909fd25530905',
+          '0xf380808080808080808080a027f11bd3af96d137b9287632f44dd00fea1ca1bd70386c30985ede8cc287476e808080c220338080',
+          '0xe48200bba0a6911545ed01c2d3f4e15b8b27c7bfba97738bd5e6dd674dd07033428a4c53af',
+        ],
+        error: ProofError.INVALID_PATH_REMAINDER,
+      },
+      {
+        title: 'test_get_smallerPathThanKey2_reverts',
```
Amxx:
```suggestion
        title: 'test_get_extraProofElements_reverts',
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,198 @@
...
+
+    /// @dev Split each byte in `value` into two nibbles (4 bits each).
+    function _nibbles(bytes memory value) private pure returns (bytes memory) {
+        uint256 length = value.length;
+        bytes memory nibbles_ = new bytes(length * 2);
+        for (uint256 i = 0; i < length; i++) {
+            (nibbles_[i * 2], nibbles_[i * 2 + 1]) = ((value[i] & 0xf0) >> 4, value[i] & 0x0f);
+        }
+        return nibbles_;
+    }
```
Amxx: solady has an assembly trick to make that function cheaper. Do we want to use it. It processes the nibbles in block of 32 (reading the input in chunks of 16 bytes) instead of 2 by 2 (reading each byte individually)

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * @dev Extracts the node ID (hash or raw data based on size)
+     *
+     * For small nodes (encoded length < 32 bytes) the node ID is the node content itself,
+     * For larger nodes, the node ID is the hash of the encoded node data.
+     */
+    function _getNodeId(Memory.Slice node) private pure returns (bytes32 nodeId, uint256 nodeIdLength) {
+        nodeIdLength = Math.min(node.length(), 32);
+        nodeId = nodeIdLength < 32 ? node.load(0) : node.readBytes32();
```
Amxx: This assumes the nodes are properly formated. In practice no node should should be exactly 32 bytes, and no node should be more than 33 bytes. Node length should be in the ranges:

- [1-31] for short node
- [33] (32 + RLP encoding prefix) for long node represented by their hash.

Should be trigger an error for anything else ? The error could be passed through the nodeIdLength and checked on the receiving end.

Amxx: Note: 

- If we get a 32 bytes node, it will be decoded as a `readBytes32` with may fail (if invalid encoding) or contain a bytes32 with a leading `00`. That will be processed in the next iteration of the loop. If we expected a hash that matches (start with a 00, and remaining 31 bytes are identical) then the "issue" is silent, but there is IMO no security concern.

- If we get more than 33 bytes. readBytes32 will call readUint256 that will revert with `RLPInvalidEncoding`.

So overall I don't think there is any concern to have here.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /// @dev Split each byte in `input` into two nibbles (4 bits each).
+    function _nibbles(bytes memory input) private pure returns (bytes memory output) {
```
james-toussaint: This function could be useful outside our usage, we should perhaps move it to something like Bytes.sol as internal and add few tests.

contracts/utils/cryptography/TrieProof.sol
james-toussaint: I would use `verify(..)` for the first one and `get(..)` (or something like `value(..)`) for the second one considering it is a _key->value_ lookup (top down instead of bottom up).

contracts/utils/cryptography/TrieProof.sol
james-toussaint: Good point, here I would use following orders (but open to suggestions):
- `get/process`
  - key (we start with the key)
  - root ("parent" of the proof)
  - proof
- `verify`
  - key
  - value
  - root
  - proof. Is consistent with current natspec and what we have in `utils/cryptography/[verifiers|signers]` where `signature` (sort of proof) is the last arg.

contracts/utils/cryptography/TrieProof.sol
james-toussaint: Very slightly yes. I was also thinking about smthg like:
```
if (prefix < uint8(Prefix.EXTENSION_ODD) + 1)
else if (prefix < uint8(Prefix.LEAF_ODD) + 1)
```
which I guess should be computed/hardcoded during compilation so assuming same cost.

Amxx: or
```
if (prefix <= uint8(Prefix.EXTENSION_ODD)) {
    //...
} else if (prefix <= uint8(Prefix.LEAF_ODD)) {
    // ...
}
```

contracts/utils/cryptography/TrieProof.sol
Amxx: IMO, `verify` should be 
- value
- ... the other args of `get`/`process`/`processProof` in the same order

That way we can do
```solidity
value.verify(...args)
```

which in some case could something a bit like
```solidity
RLP.encoder().push(nonce).push(balance).push(storageHash).push(codehash).encode().verify(...);
```
(which might make a lot of sens when verifying encoded events)

Amxx: In `MerkleTree.sol` we verify a proof ... which makes sens because the proof is "just" a `bytes32[]`, but in our case its more (there is the key).

@ernestognw WDYT ?

Amxx: ALso, @luiz-lvj, how would you like the interface to look like for the OIF broadcasters ?

james-toussaint: Yes moving value to first place is better/cleaner 👍

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,216 @@
...
+ *
+ * Proving a storage slot is usually done in 3 steps:
+ *
+ * * From the stateRoot of a block, process the account proof (see `eth_getProof`) to get the account details.
+ * * RLP decode the account details to extract the storageRoot.
+ * * Use storageRoot of that account to process the storageProof (again, see `eth_getProof`).
+ *
+ * See https://ethereum.org/en/developers/docs/data-structures-and-encoding/patricia-merkle-trie[Merkle-Patricia trie]
+ *
+ * Based on https://github.com/ethereum-optimism/optimism/blob/ef970556e668b271a152124023a8d6bb5159bacf/packages/contracts-bedrock/src/libraries/trie/MerkleTrie.sol[this implementation from by optimism].
```
Amxx:
```suggestion
 * Based on https://github.com/ethereum-optimism/optimism/blob/ef970556e668b271a152124023a8d6bb5159bacf/packages/contracts-bedrock/src/libraries/trie/MerkleTrie.sol[this implementation from optimism].
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,223 @@
...
+        // Process proof
+        uint256 keyIndex = 0;
+        uint256 proofLength = proof.length;
+        for (uint256 i = 0; i < proofLength; ++i) {
+            Node memory node = Node(proof[i], proof[i].decodeList());
+
+            // ensure we haven't overshot the key
+            if (keyIndex > keyExpanded.length) {
+                return (_emptyBytesMemory(), ProofError.INDEX_OUT_OF_BOUNDS);
+            }
```
james-toussaint: Are you able to hit that one @Amxx? I added 2 new errors [Handle empty leaf/ext path and empty ext path remainder](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5826/commits/8d80329e0adadc002b8d886cc87fa8e82e50062d) down the road to `INDEX_OUT_OF_BOUNDS`. Not 100% sure this one can be removed and likely can be moved.

Amxx: Lets try to run through it:

- initially, `keyIndex` is zero (line 120) and `keyExpanded.length` is not empty (its `2*key.length`, which is non zero from line 108)
- at each iteration, `keyIndex` can be updated as follows:
  - on line 153, incremented by one, in a branch where it is not equal to `keyExpanded.length`. The increment might lead to an equality, but not to the `INDEX_OUT_OF_BOUNDS` error.
  - on line 180 we increment it by `pathRemainderLength`. We know from line 167 that `pathRemainderLength` is smaller or equal to `keyRemainder.length()` with `keyRemainder`'s length being `keyExpanded.length - keyIndex` by construction on line 161.
  
This gives us a guarantee that at every iteration of the loop, we have `keyIndex <= keyExpanded.length`

james-toussaint: Agree. Thank you for the removal.

Amxx (dismissed):

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,226 @@
...
+    function _validateLastItem(
+        Memory.Slice item,
+        uint256 trieProofLength,
+        uint256 i
+    ) private pure returns (bytes memory, ProofError) {
+        if (i != trieProofLength - 1) {
+            return (_emptyBytesMemory(), ProofError.INVALID_EXTRA_PROOF_ELEMENT);
+        } else {
+            bytes memory value = item.readBytes();
+            return value.length == 0 ? (_emptyBytesMemory(), ProofError.EMPTY_VALUE) : (value, ProofError.NO_ERROR);
```
ernestognw: Why not just?

```suggestion
            return (value, value.length == 0 ? ProofError.EMPTY_VALUE : ProofError.NO_ERROR);
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,226 @@
...
+                            ? _validateLastItem(decoded[1], proofLength, i)
+                            : (_emptyBytesMemory(), ProofError.MISMATCH_LEAF_PATH_KEY_REMAINDER);
+                } else {
+                    return (_emptyBytesMemory(), ProofError.UNKNOWN_NODE_PREFIX);
+                }
+            } else {
+                return (_emptyBytesMemory(), ProofError.UNPARSEABLE_NODE);
+            }
+
+            // Reset memory before next iteration. Deallocates `decoded` and `path`.
```
ernestognw, line 186:
```suggestion
            // Reset memory before next iteration. Deallocates `decoded`, `path` and `value` (_validateLastItem).
```

contracts/utils/cryptography/TrieProof.sol
ernestognw: > In practice no node should should be exactly 32 bytes, and no node should be more than 33 bytes.

This sounds about right, though, I think the `_getNodeId` function is wrong since it doesn't allow values of 33 bytes (i.e. 32-bytes hash + RLP encoding prefix). Shouldn't we do?:

```suggestion
        nodeIdLength = Math.min(node.length(), 32);
        nodeId = node.length() < 32 ? node.load(0) : node.readBytes32();
```

> Should be trigger an error for anything else ?

I understand everything else is 32, 34, 35, 36... Correct? 

> If we get a 32 bytes node, it will be decoded as a readBytes32 with may fail (if invalid encoding) or contain a bytes32 with a leading 00

Not sure if I follow the leading zeroes here. For the case of length 32, it would be decoded by `readBytes32` successfully if encoded correctly as a short string. For example, a 31-byte string: 0x9f (0x80 + 31) + 31 bytes = 32 bytes total). Am I right?

> That will be processed in the next iteration of the loop. If we expected a hash that matches (start with a 00, and remaining 31 bytes are identical) then the "issue" is silent, but there is IMO no security concern.

I also don't follow what the remaining 31 bytes are in this case, but the proof will fail in the next iteration with `INVALID_LARGE_NODE`. I guess the question is whether we want to return an error for the specific case where the node length is invalid (e.g. `ProofError.INVALID_NODE_LENGTH`), but I personally like `INVALID_LARGE_NODE` since it really is what's happening (e.g. if a node with length 32 was received, it should've been encoded as a large node)

IMO we're fine with the current error.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,226 @@
...
+                    keyIndex += 1;
+                }
+            } else if (decoded.length == LEAF_OR_EXTENSION_NODE_LENGTH) {
+                bytes memory path = decoded[0].readBytes().toNibbles(); // expanded path
+                if (path.length == 0) {
+                    return (_emptyBytesMemory(), ProofError.EMPTY_PATH);
+                }
+                uint8 prefix = uint8(path[0]);
+                Memory.Slice keyRemainder = keyExpanded.asSlice().slice(keyIndex); // Remaining key to match
+                Memory.Slice pathRemainder = path.asSlice().slice(2 - (prefix % 2)); // Path after the prefix
```
ernestognw, line 151: This `.slice` call will revert if the provided `path`'s length is 1 and the prefix is even. Shall we just check the following in Line 145?

```solidity
                if (path.length < 2) {
                    return (_emptyBytesMemory(), ProofError.EMPTY_PATH);
                }
```

Especially if this function is `try*`, I guess it's implicit that it shouldn't revert. Although we mention that the function may still revert due to invalid RLP encoding, I don't think this would classify as invalid RLP encoding.

contracts/utils/Memory.sol
```diff
@@ -112,6 +112,16 @@ library Memory {
...
 
+    /// @dev Return the keccak256 hash of the memory section represented by the slice.
+    /// Equivalent to keccak256(toBytes(self)).
+    function getHash(Slice self) internal pure returns (bytes32 result) {
+        uint256 len = length(self);
+        Memory.Pointer ptr = _pointer(self);
+        assembly ("memory-safe") {
+            result := keccak256(ptr, len)
+        }
+    }
```
ernestognw: If I understand correctly, this function is used only to compare 2 slices. What if instead of `getHash` we create an `equal(Slice a, Slice b)` function?

I personally don't like the `getHash(Slice)` because it can be hashed with sha256 as well. So there's no apparent reason to just call "hash" to the `keccak256` of the `Slice` in particular.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,226 @@
...
+    function _getNodeId(Memory.Slice node) private pure returns (bytes32 nodeId, uint256 nodeIdLength) {
+        nodeIdLength = Math.min(node.length(), 32);
+        nodeId = nodeIdLength < 32 ? node.load(0) : node.readBytes32();
+    }
+
+    function _emptyBytesMemory() private pure returns (bytes memory result) {
+        assembly ("memory-safe") {
+            result := 0x60 // mload(0x60) is always 0
+        }
+    }
```
ernestognw, line 230: I'm unconfortable with this return. We may be unknowingly exposing the zero slot to external users who may later manipulate the value. If so, it would have unexpected consequences. I would feel ok if we just allocate new memory with 0 length.

contracts/utils/cryptography/TrieProof.sol
Amxx: Indeed, we cannot prevent reverts completelly with the current RLP library. Line 132 will revert if the node is not a properly encoded RLP list (among other things that can revert).

contracts/utils/cryptography/TrieProof.sol
Amxx: `value` (_validateLastItem) is only processed/allocated in cases where we return. When such a return happens, the execution flow ends and we don't get to line 186. That means that the last iteration in not deallocated (only the intermediary iterations of the loops are deallocated.

Note that its a good thing that value is not dealocatted. When a function returns a `bytes memory` you want it to be in reserved space.

contracts/utils/cryptography/TrieProof.sol
Amxx: For users to manipulate it, they would have to manually (there is no builtin mechanism for that, either in solidity not in our library) increase the length of a memory array. That operation would be dangerous, regardless of of the value being returned.

Not that by default, solidity uses that mechanism for uninitialized empty buffers. The following test contract returns 0x60 for all the test functions

```solidity
contract Test {
    function test_1() public pure returns (bytes32) {
        return _ptrToBytes32(_empty1());
    }
    
    function test_2() public pure returns (bytes32) {
        return _ptrToBytes32(_empty2());
    }
    
    function test_3() public pure returns (bytes32) {
        bytes memory unitialized;
        return _ptrToBytes32(unitialized);
    }

    function _ptrToBytes32(bytes memory input) internal pure returns (bytes32 output) {
        assembly ("memory-safe") { output := input }
    }

    function _empty1() internal pure returns (bytes memory buffer) {
        // do nothing
    }

    function _empty2() internal pure returns (bytes memory buffer) {
        assembly ("memory-safe") { buffer := 0x60 }
    }
}
```

Amxx: Also, we already use this construction in `contracts/utils/draft-InteroperableAddress.sol`

contracts/utils/cryptography/TrieProof.sol
Amxx: > I think the _getNodeId function is wrong since it doesn't allow values of 33 bytes (i.e. 32-bytes hash + RLP encoding prefix)

If the node is length 33, then `nodeIdLength` is computed to be 32. In that case we return length 32, and we RLP decode the node as a bytes32 (`node.readBytes32`). That is the correct behavior.

Amxx: > I understand everything else is 32, 34, 35, 36... Correct?

Yes. Note that the 34 and above already cause a RLP revert.

Amxx: > For the case of length 32, it would be decoded by readBytes32 successfully if encoded correctly as a short string. For example, a 31-byte string: 0x9f (0x80 + 31) + 31 bytes = 32 bytes total). Am I right?

The whole thing here is "if encoded correctly".

Amxx: My first messages might have been a bit confuse (my understanding of the process clarified) so let me rephrase all this:

We expect two types of nodes:
- "short" (<= 31 bytes) values. These are should not be RLP encoded and the ID is just the "raw" value. We can clearly identify them from the length of the ID being <= 31 bytes
- Hashes of long (>= 32 bytes) values. These hashes are 32 bytes. This mean the hash's length is not longer than the value. In these cases the hash is RLP encoded, and the node is exactly 33 bytes long.

So far, we expect input that are either <=31 bytes, or 33 bytes. The first one are "raw", the second one must be RLP decoded as bytes32, and the actual nodeID length is 32.

---

The question is what if we get something that is NOT standard:

- Anything <= 31 bytes is used raw (not decoded), so there is no invalid input with that length.

- We should never get something that is 32 bytes long. If we do, we are going to try to RLP decode it. It may revert (improper encoding) or return a decoded value (that we know is 31 bytes). Here, the implementation treats it like if we received 33 bytes. Its going to consider the received value to be 32 bytes and it will represent it using a bytes32 that contains 31 bytes of decoded data, and 1 bytes of zero. While this format is not standard, it may be decoded without error, and that decoding may correspond to the hash of a node (if that hash starts with 00). So there is a possibility that this unexpected encoding leads to a verification that passes. I don't think we should be worried about it, but I want to document that

- We could get something that is 33 bytes, but that cannot be decoded (length is not properly encoded or item type is List). In that case we revert with `RLPInvalidEncoding`.

- We should never get something that is 34 bytes long or more. If we do `readBytes32` will revert with `RLPInvalidEncoding`

Amxx: Note: the suggested modification does not change the behavior at all. `node.length() < 32` and `nodeIdLength < 32` are strictly equivalent.

contracts/utils/cryptography/TrieProof.sol
Amxx: Because of the construction of path as a `.toNibbles()`, its length is a multiple of 2. It cannot be one

Therefore, if the length is not zero (as currently verified) it must be 2, 4, 6, or any other even number.
So replacing `== 0` by `< 2` would not make any change to the logic, but it would make for longer bytecode.

contracts/utils/cryptography/TrieProof.sol
ernestognw: We discussed it during a call and agreed that it's ok to return the memory pointer to `0x60` since it seems unlikely that the bytes buffer will be manipulated later in a way that breaks.

Given the InteroperableAddress library hasn't been audited, we should point this out during the next audit to see the research team opinion

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,226 @@
...
+        Memory.Slice item,
+        uint256 trieProofLength,
+        uint256 i
+    ) private pure returns (bytes memory, ProofError) {
+        if (i != trieProofLength - 1) {
+            return (_emptyBytesMemory(), ProofError.INVALID_EXTRA_PROOF_ELEMENT);
+        } else {
+            bytes memory value = item.readBytes();
+            return (value, value.length == 0 ? ProofError.EMPTY_VALUE : ProofError.NO_ERROR);
+        }
```
ernestognw:
```suggestion
        }
        bytes memory value = item.readBytes();
        return (value, value.length == 0 ? ProofError.EMPTY_VALUE : ProofError.NO_ERROR);
```

contracts/utils/cryptography/TrieProof.sol
ernestognw: Added a comment pointing out that checking `== 0` is equivalent to `< 2` due to the path expansion with `toNibbles`

contracts/utils/cryptography/TrieProof.sol
ernestognw: I do like the `<` instead of two `==`

contracts/utils/cryptography/TrieProof.sol
ernestognw: > node.length() < 32 and nodeIdLength < 32 are strictly equivalent.

Yes, I felt for a second we were missing the case where the node is `32` but got confused.

> So there is a possibility that this unexpected encoding leads to a verification that passes. I don't think we should be worried about it, but I want to document that

I'm also not worried about this. I added a NOTE in `_getNodeId`

ernestognw:
I went ahead with [this comment](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5826#discussion_r2608514537) and removed the `getHash` function in favor of a `Memory.equal` for slices.

ernestognw (commented):
Looks good, but I think we need to discuss one last thing.

There's this [PoC](https://github.com/tjade273/optimism-merkle-forgery-poc) that shows it's possible to construct 2 RLP items such that `keccak256(rlp.encode(r1)) == bytes32(rlp.encode(r2))` being `r1` a long item (length >= 33) and `r2` a short item (length < 32). This means that a proof can be forged for `r1` if `r2` exists and viceversa. Although the attack is very unlikely since it apparently requires a branch node close to the leaf's key, it may justify that we expose a variant of `verify` called `secureVerify` that rehashes the key. This is what the [`SecureMerkleTrie` library from optimism](https://github.com/ethereum-optimism/optimism/blob/e7cc17166daf1c55bb3ba138018444de7d29f3c3/packages/contracts-bedrock/src/libraries/trie/SecureMerkleTrie.sol#L47-L49) does

```solidity
function secureVerify(
    bytes memory value,
    bytes32 root,
    bytes memory key,
    bytes[] memory proof
    return verify(value, root, abi.encodePacked(keccak256(key)), proof);
}
```

As far as I understand, both the `verify` and `secureVerify` variants would be useful since the former would serve to verify against a `transactionRoot` or a `receiptsRoot`, but `storageRoot` and `stateRoot` use the secure variant (i.e. this is why we hash the `storage` address and the `slot` in the tests), so it seems worth keeping both.

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,231 @@
...
+    /// @dev The radix of the Ethereum trie
+    uint256 internal constant EVM_TREE_RADIX = 16;
+
+    /// @dev Number of items in a branch node (16 children + 1 value)
+    uint256 internal constant BRANCH_NODE_LENGTH = EVM_TREE_RADIX + 1;
+
+    /// @dev Number of items in leaf or extension nodes (always 2)
+    uint256 internal constant LEAF_OR_EXTENSION_NODE_LENGTH = 2;
+
+    /// @dev Verifies a `proof` against a given `key`, `value`, `and root` hash.
```
ernestognw, line 69: wdyt?

```suggestion
    /**
     * @dev Verifies a `proof` against a given `key`, `value`, and `root` hash.
     *
     * Ethereum introduced key hashing for state and storage tries to prevent DoS
     * attacks by ensuring uniform key distribution and balanced tries. On the opposite,
     * the `transactionsRoot` and `receiptsRoot` don't use key hashing since keys are indexes,
     * thus, short, predictable and not user-controlled. To match Ethereum semantics:
     *
     * * For `storageRoot`/`stateRoot`: Hash the `key` first using `keccak256(key)` before passing it.
     * * For `transactionsRoot`/`receiptsRoot`: Use the `key` directly (RLP-encoded transaction or receipt index).
     *
     * Example usage:
     *
     * ```solidity
     * // For `storageRoot` or `stateRoot` (secure)
     * bytes32 hashedKey = keccak256(abi.encodePacked(slot));
     * bool isValid = verify(value, storageRoot, abi.encodePacked(hashedKey), proof);
     *
     * // For `transactionsRoot` or `receiptsRoot` (non-secure)
     * bool isValid = verify(txData, transactionsRoot, abi.encodePacked(index), proof);
     * ```
     *
     * NOTE: Not hashing the keys make the proof vulnerable to RLP collision forgery attacks, where an attacker constructs
     * two RLP items (one long >=32 bytes, one short <32 bytes) such that `keccak256(rlp.encode(long)) == bytes32(rlp.encode(short))`.
     * This allows proof forgery in custom tries with arbitrary keys. Hashing reduces this risk by enforcing
     * fixed-length keys but breaks non-secure compatibility. If using for custom non-secure tries, consider a secure wrapper.
     */
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     * @dev Extracts the node ID (hash or raw data based on size)
+     *
+     * For small nodes (encoded length < 32 bytes) the node ID is the node content itself,
+     * For larger nodes, the node ID is the hash of the encoded node data.
+     *
+     * NOTE: For exactly 32-byte inputs, this function attempts to RLP decode the value. Standard
+     * RLP encodings will revert with {RLP-RLPInvalidEncoding}. Non-standard encodings that decode
+     * to fewer than 32 bytes will produce a zero-padded bytes32 value. While this could theoretically
+     * match a hash starting with 0x00, verification would fail in subsequent proof traversal steps.
```
Amxx: I don't understand that comment. In particular, I don't understand what `Standard RLP encodings` and `Non-standard encodings` are.

For example, `0x9f5ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba95` is 32 bytes long, and is a completelly valid (canonical) RLP encoding for `0x5ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba95` (that is 31 bytes long). 

If node contains this value (`0x9f5ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba95`), then we will not get a revert.

Amxx: IMO, this note adds more confusion than information.

contracts/utils/cryptography/TrieProof.sol
Amxx: IMO this is too much. People that want to learn about how ethereum trees are constructed should probably read that elsewhere, and not in ABI reference of the function.  Line 25 of the file links to a page about particia-merkle trees on the ethereum.org website. I would think this is enough. If there are other ressources we want to link, we should do it there. Maybe EIP-2718 ?

Amxx:
My understanding is that the choice of key is not ours to make. Its really the tree building process that decides what key is used, and the verifyer must follow the same mechanism otherwize they won't be able to generate a proof.

- When verifying an account details (against the blocks's stateRoot), the key is the keccak256 hash of the account's address
- When verifying a storage slot (against the account's storageHash), the key is the keccak256 hash of the slot
- When verifying a transaction (against the blocks's transactionsRoot), the key is the index of the transaction in the block ?
- When verifying an event (against the blocks's receiptsRoot), the key is the index of the event in the block ?

In all cases its the same function that is used, just with different input (that you prepare differently). I'm not sure I see the point in having wrappers that do that preparation. If the preparation is not done correctly, there should not be a proof, and the verification shoud always fail.

If having a `secureVerify` only does the preparation of the key, and nothing else, I don't think its any more (or less) secure than the "simple" verify. I'd rather not have it than create the impression that one is secure (and should be used) when the other is not.

contracts/utils/cryptography/TrieProof.sol
Amxx: I think there is some inconsistency between the text that says transactions and events use a key that is the "RLP-encoded transaction or receipt index", and the code that does `abi.encodePacked(index)` (not an RLP encoding).

Overall, I would not include that in this PR. Instead I would push the idea (introduced in a call) of adding wrappers specific to account/storage/events/transactions.

ernestognw: I was personally confused about the differences between hashing the key or not, so I would argue that anyone who wants to learn about patricia tries may get the fundamentals from the Ethereum site, but not necessarily will become an expert and understand the security implications.

Maybe here I went too far at explaining, but I was trying to outline the differences for the key in a comment. It seems RLP forgery is not an issue generally, so I'm fine not adding this block

ernestognw:
To be clear, I initially thought the `SecureMerkleTrie` from Optimism was deliberately rehashing the key in a similar fashion to how we prevent a second preimage attack in our merkle tree typescript library, so I considered relevant to have the "secure" variant.

However, I found some research suggesting that the keys are hashed for storage proofs and account proofs to avoid arbitrary disk I/O, so rehashing doesn't really concern to on-chain verification. Still, for transaction or events proof I understand RLP forgery is theoretically possible, just they keys are too low and sequential.

> In all cases its the same function that is used, just with different input (that you prepare differently). I'm not sure I see the point in having wrappers that do that preparation.

Yeah I was trying to avoid users making a mistake rather than "preparing" the verification, but on a second thought I agree there's no point if there's no relevant security consideration.

> If having a `secureVerify` only does the preparation of the key, and nothing else, I don't think its any more (or less) secure than the "simple" verify. I'd rather not have it than create the impression that one is secure (and should be used) when the other is not.


Having specific verify functions may be a good idea but I only see benefit in documenting how to obtain and process each type of proof. For that case I'd rather make a guide in the docs 🤔

contracts/utils/cryptography/TrieProof.sol
ernestognw: I added the docs since I don't think it's worth checking that the value has `length != 32` and you suggested it should be documented.

> `0x9f5ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba95` is 32 bytes long, and is a completelly valid (canonical) RLP encoding for `0x5ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba95` (that is 31 bytes long). 

Yes, but wouldn't that revert in the next traversal iteration? This is what I tried to express with the comment 😅, I can try with a simpler version

Amxx:
> However, I found some research suggesting that the keys are hashed for storage proofs and account proofs to avoid arbitrary disk I/O, so rehashing doesn't really concern to on-chain verification.

For the transaction and receipt tries, the keys are just the index in the block (starting at 0). That makes for overall "shorters" (as in not too deep) trees. We know all values from 0 to N are populated, so there is no serious "unbalancing of the tree here"

For the storage (within account), the slots are controlled by the "user" (the compiler actually). Its easy to populate many consecutive slot at a pseudo-random location. That is what appens when you write a long string to storage. The hashing of the storage slot randomizes all that in the tree, making it more balanced.

For the state tree, I guess the logic is similar to the storage tree, though I don't really understand why. Addresses are hashes already, so I'm not sure what this is preventing. Many its for dealing with consecutive/close addresses that are the result of vanity generation, but an attacker could just as well use a vanity gen that makes the images of addresses "close" to try to unbalance the tree. Still doing the hash makes the state tree have a depth of 32 bytes instead of 20 bytes, which feels counterproductive. Maybe a coredev could explain us why its don't that way.

> Having specific verify functions may be a good idea but I only see benefit in documenting how to obtain and process each type of proof. For that case I'd rather make a guide in the docs 🤔

We don't even know how to generate the proofs for transactions and events. AFAIK the easiest way might be to fetch the entier block (with all txs and events) and rebuild the trees from that. I feel like whoever is going to use this function to prove events likelly knows more about this process than we do, so having mid-documentation (as in documentation that is not perfect) is probably useless. I'd just point to the EIP that stnadardize that (or the yellow paper)

contracts/utils/cryptography/TrieProof.sol
Amxx: > Yes, but wouldn't that revert in the next traversal iteration?

The RLP decoding will not revert, and the system will expect the next iteration to be a "long" value (32 bytes or more) that has a hash `0x005ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba95` (or `0x5ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba9500` I'm not 100% sure where the zero go)

If the next node is indeed a "long" value with this hash, verification will be good.

I don't think that would ever happens, and if it did, it would be more of a "the tree/proof is built in a strange way" then a security issue.

Amxx: And if the next iteration fails, it will be because of an invalid hash, so `ProofError.INVALID_LARGE_NODE` is returned. It won't be a `RLP-RLPInvalidEncoding` as the comment mentions.

Amxx: Going back to the note:
- nodes that are 32 bytes long will be decoded (and not taken raw)
  - if the RLP encoding is invalid, we get a revert
  - if the RLP encoding is valid, we get a value that is 31bytes log of less.
    - that value will be padded with zeros, and treated as if it was a full 32 bytes hash if that bytes is correct, the system will accept it (though, all 32 bytes of this hash should have been rlp encoded, and we should have got a 33 bytes node). This is a strange case that should never happens, and cause no visible security issue.

Amxx: we could also change the code to take the 32 bytes "raw" ... and we another "strange" behavior that is also safe AFAIK.

```solidity
    function _getNodeId(Memory.Slice node) private pure returns (bytes32 nodeId, uint256 nodeIdLength) {
        nodeIdLength = Math.min(node.length(), 32);
        nodeId = node.length() < 33 ? node.load(0) : node.readBytes32();
    }
```

if that is easier to document, we could go for it.

(tested, it passes all unit tests)

ernestognw: > The RLP decoding will not revert, and the system will expect the next iteration to be a "long" value (32 bytes or more) that has a hash `0x005ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba95` (or `0x5ae162fe140cc4e73ea06bea6e98888bbe5531f286d90716c7ffc5de98ba9500` I'm not 100% sure where the zero go)

A non-standard 32 bytes encoding could be (e.g.) `0x9f[31 bytes]`, then `itemLength == 31` and `readUint256` will return `uint256(item.load(itemOffset)) >> (256 - 8 * 31)`, which will produce a 1 byte of zero padding on the left side.

From there, the TrieProof library will check `if (currentNodeIdLength != 32 || keccak256(encoded) != currentNodeId)` and revert with `INVALID_LARGE_NODE`. The note is describing this case, which will not match the next traversal iteration.

> I don't think that would ever happens, and if it did, it would be more of a "the tree/proof is built in a strange way" then a security issue.

I also don't think it's a security issue, but we wanted to document the behavior of `_getNodeId` when a 32-bytes value is passed. We can remove the note completely, I understood you wanted to document clearly, but seems that if it's not a security concern the note is not worth it. Could you clarify?

> And if the next iteration fails, it will be because of an invalid hash

This is what the block says! It reads that "Standard RLP encodings will revert with {RLP-RLPInvalidEncoding}", which applies to 32-bytes where:

1. Prefix `0xA0` (SHORT_OFFSET + 32): Calculates `strLength = 32` and then checks `itemLength > strLength `. Reverts with {RLPInvalidEncoding}
2. Prefix = `0xB8`+ (long string encoding): There isn’t enough space for prefix + length byte + 32 bytes of data. Fails `itemLength > lengthLength + len` and reverts with {RLPInvalidEncoding}
3. Prefix = `0xC0`+ (list encoding): Here, `readUint256` requires `itemType == ItemType.Data`. Reverts with {RLPInvalidEncoding}
4. Prefix < `0x80` (single byte): There's no way to provide a standard RLP encoding for a single value that's 32 bytes

Then, the note mentions non-standard encoding, which will pass with the example that I'm describing above (e.g. `0x9f[31 bytes]`)

> Going back to the note:
> 
> * nodes that are 32 bytes long will be decoded (and not taken raw)
>   
>   * if the RLP encoding is invalid, we get a revert
>   * if the RLP encoding is valid, we get a value that is 31bytes log of less.
>     
>     * that value will be padded with zeros, and treated as if it was a full 32 bytes hash if that bytes is correct, the system will accept it (though, all 32 bytes of this hash should have been rlp encoded, and we should have got a 33 bytes node). This is a strange case that should never happens, and cause no visible security issue.

I think this is what the note reads...

> we could also change the code to take the 32 bytes "raw" ... and we another "strange" behavior that is also safe AFAIK.

I think the current code is fine. The discussion is around the note so I would keep changes to the note itself. It would be helpful if you propose a note that makes sense to you.

ernestognw: Maybe?

```suggestion
     * NOTE: For exactly 32-byte inputs, this function attempts to RLP decode the value. If the RLP encoding
     * is invalid, decoding will revert with {RLP-RLPInvalidEncoding}. If the RLP encoding is valid,
     * the decoded value is used as the node ID for the next traversal step, where verification will
     * fail with {INVALID_LARGE_NODE} if the node ID doesn't match the expected hash.
```

contracts/utils/cryptography/TrieProof.sol
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * NOTE: For exactly 32-byte inputs, this function attempts to RLP decode the value. If the RLP encoding
+     * is invalid, decoding will revert with {RLP-RLPInvalidEncoding}. If the RLP encoding is valid,
+     * the decoded value is used as the node ID for the next traversal step, where verification will
+     * fail with {INVALID_LARGE_NODE} if the node ID doesn't match the expected hash.
+     */
+    function _getNodeId(Memory.Slice node) private pure returns (bytes32 nodeId, uint256 nodeIdLength) {
+        nodeIdLength = Math.min(node.length(), 32);
+        nodeId = nodeIdLength < 32 ? node.load(0) : node.readBytes32();
+    }
```
Amxx, line 224:
```suggestion
    /**
     * @dev Extracts the node ID (hash or raw data based on size)
     *
     * For small nodes (encoded length < 32 bytes) the node ID is the node content itself,
     * For larger nodes, the node ID is the hash of the encoded node data.
     *
     * NOTE: Under normal operation, the input should never be exactly 32-byte inputs. If such an input is provided, 
     * it will be used directly, similarly to how small nodes are processed. The following traversal check whether 
     * the next node is a large one, and whether its hash matches the raw 32 bytes we have here. If that is the case,
     * the value will be accepted. Otherwise, the next step will return an {INVALID_LARGE_NODE} error.
     */
    function _getNodeId(Memory.Slice node) private pure returns (bytes32 nodeId, uint256 nodeIdLength) {
        uint256 nodeLength = node.length();
        return nodeLength < 33
            ? (node.load(0), nodeLength)
            : (node.readBytes32(), 32);
    }
```

PR #5836, opened 2025-08-07
Add ERC20MultiVotes for partial delegations support

### LINKED ISSUE #4963: Add partial delegation to ERC20Votes
**🧐 Motivation**
At Tally, we see a lot of requests for "partial delegation".  "Partial delegation" means letting one address split up voting power between delegates. 

There are several existing workarounds:
- Tokenholders can split their tokens into different wallets
- [Franchiser](https://github.com/NoahZinsmeister/franchiser)
- [ENS's multi delegate](https://github.com/ensdomains/governance-contracts/blob/7d896f62a403a01982452c5543e767af81adc0a6/contracts/ERC20MultiDelegate.sol)

The fact that there are so many workarounds fragments the standard. There's no single one for frontends like ours to support.

The larger problem with these workarounds is that they don't work well for custodians. Custodians have strict legal mandates and security requirements about moving tokens. Generally, they cannot put their tokens into delegation-splitting contracts like Franchiser. Even splitting tokens into multiple wallets is a big operational lift for them.

They would love to have an in-contract method on ERC20Votes that lets them split voting power.

**📝 Details**

To implement this feature, add a method to `ERC20Votes` that supports partial delegation.

I don't have a strong view on implementation details here.  [Franchiser's](https://github.com/NoahZinsmeister/franchiser/blob/962079a092f4db1142473658e1650855831886c5/src/Franchiser.sol#L144C14-L147C6) `subDelegateMany(
        address[] calldata subDelegatees_,
        uint256[] calldata amounts
    )`  or [ENS multi-delegate's](https://github.com/ensdomains/governance-contracts/blob/7d896f62a403a01982452c5543e767af81adc0a6/contracts/ERC20MultiDelegate.sol#L92C1-L97C1)    `delegateMulti(
        uint256[] calldata sources,
        uint256[] calldata targets,
        uint256[] calldata amounts
    ) ` are good starting points. I am happy to chat about details and their tradeoffs or suggest some implementations.

The bookkeeping for partial delegation does increase gas costs.  IMO that overhead is feasible on rollups, but not on Ethereum mainnet. I'd suggest making partial delegation an extension to or alternate implementation of [governance/utils/Votes.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/utils/Votes.sol).

Sp3rick:
Fixes #4963

Implementing MultiVotes as an extension of Votes as first step towards ERC20MultiVotes
MultiVotes adds multiDelegate() to add/change/remove partial delegations, old delegate() is used as defaulted delegate, so all free voting units are assigned to defaulted.

Im currently working on that and maybe I will add ERC20MultiVotes directly trough this pull

I publish that now as a draft to gather preliminary feedback from maintainers and community to proceed in case any refactoring or improvements are advised

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/.changeset/silent-terms-beg.md
+++ b/.changeset/silent-terms-beg.md
@@ -1,0 +1,5 @@
+---
+'openzeppelin-solidity': minor
+---
+
+Add ERC20MultiVotes with MultiVotes for partial delegations support
--- a/contracts/governance/README.adoc
+++ b/contracts/governance/README.adoc
@@ -110,6 +110,8 @@ NOTE: Functions of the `Governor` contract do not include access control. If you
 
 {{VotesExtended}}
 
+{{MultiVotes}}
+
 == Timelock
 
 In a governance system, the {TimelockController} contract is in charge of introducing a delay between a proposal and its execution. It can be used with or without a {Governor}.
--- a/contracts/governance/utils/IMultiVotes.sol
+++ b/contracts/governance/utils/IMultiVotes.sol
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.5.0) (governance/utils/IMultiVotes.sol)
+pragma solidity ^0.8.26;
+
+import {IVotes} from "./IVotes.sol";
+
+/**
+ * @dev Common interface for {ERC20MultiVotes} and other {MultiVotes}-enabled contracts.
+ */
+interface IMultiVotes is IVotes {
+    /**
+     * @dev Requested more units than actually available.
+     */
+    error MultiVotesExceededAvailableUnits(uint256 requested, uint256 available);
+
+    /**
+     * @dev Mismatch between number of given delegates and correspective units.
+     */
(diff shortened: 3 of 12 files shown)


Sp3rick:
I share some design choices that may be worth a look for feedbacks

**Function `multiDelegates`**
multiDelegates is a function aiming to return delegation list of an holder

Currently this function is not fail proof, if a holder has too many delegates, calling multiDelegates from a contract or client may cause an out of gas error


Thus i will write down here some possible solutions to gather a feedback on that:

**Keep it like this**
Clients and contracts that uses multiDelegates should handle the case where out of gas error is thrown, and should be aware that it can happen
Clients that acutally need the list should use indexers

**Limit on partial delegations amount for each account**
It may be hardened on the code, or using a constructor

**Single index or paging for multiDelegates**
The only pros here is that contracts have a chance to always get a portion of multiDelegates, but for security reasons they should be aware that order may change on a different transaction

**Removing multiDelegates**
Thus forcing clients into using indexers, and disabling at all any possibility to gather the list from other contracts

